PR #4, opened 2019-10-26
Alpha application

CedrickFlocon:


app/src/main/java/io/homeassistant/android/launch/LaunchActivity.kt
```diff
@@ -0,0 +1,22 @@
...
+
+class LaunchActivity : AppCompatActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        if (Session.getInstance().url.isNullOrBlank()) {
+            startActivity(OnboardingActivity.newInstance(this))
+        } else {
+            startActivity(WebViewActivity.newInstance(this))
```
balloob: For future, maybe a good idea to fire off token valid check here? That way we don't have to wait until the frontend requests a token but can make sure it's valid.

app/src/main/java/io/homeassistant/android/api/AuthenticationService.kt
```diff
@@ -0,0 +1,38 @@
...
+import retrofit2.Call
+import retrofit2.http.Field
+import retrofit2.http.FormUrlEncoded
+import retrofit2.http.POST
+
+
+interface AuthenticationService {
+
+    companion object {
+        const val CLIENT_ID = "https://home-assistant.io/iOS"
```
balloob:
```suggestion
        const val CLIENT_ID = "https://home-assistant.io/android"
```

app/src/main/java/io/homeassistant/android/webview/WebViewActivity.kt
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    override fun setupJavascriptInterface(token: Token) {
+        webView.addJavascriptInterface(object : Any() {
+            @JavascriptInterface
+            fun getExternalAuth(callback: String) {
+                webView.post {
+                    webView.evaluateJavascript(
+                        "${JSONObject(callback).get("callback")}(true, {\n" +
+                                "  \"access_token\": \"${token.accessToken}\",\n" +
```
balloob: This seems like it is hardcoding the tokens from the first launch. However, if the token is expired in the frontend, this code should refresh the token before giving it to the app.

app/src/main/java/io/homeassistant/android/webview/WebViewActivity.kt
```diff
@@ -0,0 +1,108 @@
...
+                }
+            }
+
+            @JavascriptInterface
+            fun externalBus(message: String) {
+                Log.d(TAG, "External bus $message")
+                webView.post {
+                    when {
+                        JSONObject(message).get("type") == "config/get" -> {
+                            val script = "externalBus({" +
```
balloob: This feels weird, doesn't Kotlin have some JSON dumper for maps ?

app/src/main/java/io/homeassistant/android/api/AuthenticationService.kt
balloob: (I just put it online for you :)

app/src/main/java/io/homeassistant/android/webview/WebViewActivity.kt
balloob: I noticed token is refreshed if necessary when we open the webview. This will work ok if the user is not planning to use the app longer than that the token is currently valid. Better if we refresh it in this function, if necessary. 

Ideally, it would work like this:
 - the moment the app starts up and it has tokens, start a refresh request if they are no longer valid
 - in the JavaScript interface, if a refresh request is in progress (because app got opened), wait for that request to finish, otherwise start own request for refresh.

w1ll1am23:
Noticed you are using findViewById in a few places. I know there isn't going to be a lot of native UI in the app to start with, but might want to look at using Kotlin Synthetic properties.

basically would let you replace, for example
```
findViewById<TextView>(R.id.version_text_view).text
```
with
```
version_text_view.text
```
no need to implicitly provide the view type.

Just a thought.

app/src/main/java/io/homeassistant/android/launch/LaunchActivity.kt
CedrickFlocon: Yes agreed. Could we enable [Github Project](https://github.com/home-assistant/home-assistant-android/projects) to list all the task ?
like listen [connection-status](https://developers.home-assistant.io/docs/en/frontend_external_bus.html#connection-status-update-connection-status) and display authentication screen again in case of `"auth-invalid" | "disconnected"`

app/src/main/java/io/homeassistant/android/webview/WebViewActivity.kt
CedrickFlocon: JSONObject should be able to did it

CedrickFlocon:
@w1ll1am23 As you say we won't have a lot of native UI component.
So do we need to include a dependencies (even if it won't impact the size of APK much) for 5 findViewById ?

CedrickFlocon:
@balloob I will fix these comment

app/src/main/java/io/homeassistant/android/webview/WebViewActivity.kt
CedrickFlocon: To make it quicker I refresh the token (if needed) each time getExternalAuth is call. Is that ok for you ?

w1ll1am23:
@CedrickFlocon it was just a thought, no reason to actually implement it if there are no plans to add additional native UI. I could see the settings menu getting pretty full, but once that happens we can revisit. Thanks for all the hard work here! Looks great.

andriej:
@CedrickFlocon - I'm no android dev, but when there will be possibility to join alpha/beta test, please provide way for people interested in. :-)

CedrickFlocon:
@andriej Sure, I am working on a CI / CD stuff. 
I will probably deploy master commit thought [Firebase App Distribution](https://firebase.google.com/docs/app-distribution)

PR #18, opened 2019-11-09
Add refresh layout to Webview

timmo001:
## Description

Allows you to swipe down to reload the webview

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -59,6 +59,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation "androidx.appcompat:appcompat:1.1.0"
     implementation "androidx.constraintlayout:constraintlayout:1.1.3"
+    implementation "androidx.swiperefreshlayout:swiperefreshlayout:1.0.0"
 
     implementation "com.squareup.retrofit2:retrofit:2.6.2"
     implementation 'com.squareup.okhttp3:logging-interceptor:4.2.2'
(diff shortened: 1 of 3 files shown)


balloob:
I don't think that we should add this if we're targeting to be a native app ?

CedrickFlocon:
In my opinion the frontend is reactive so we don't need to have a pull to refresh.

CedrickFlocon:
Or it could be interesting if we have a network error page
Screen Shot 2019-11-10 at 21 13 36

CedrickFlocon:
Or it could be interesting if we have a network error page

balloob:
So for connection status, since we use service workers, the UI will actually load but be unable to connect. In that case, we'll let the app know via [the connection-status event](https://developers.home-assistant.io/docs/en/frontend_external_bus.html#connection-status-update-connection-status). I don't think that we need pull to refresh.

timmo001:
I somewhat agree. I personally just like to manually refresh the page more for piece of mind when things get updated or the websocket has issues.

Feel free to close :smile:

balloob:
Alright, let's close it. And then let's make sure our webapp deals with things so well, this won't ever be missed 👍

frenck:
There is a use case, e.g., force update for fetching lovelace changes or directly after an HA update (at least those are the moments I need this functionality in the iOS app)

andriej:
I've just encountered problem - part od UI didn't load (weather icons, png files) and I'm not able to reload UI. Changing menus doesn't help.

TheZoker:
I use it quite often within the mobile web app

TheZoker:
So until this is not needed anymore, I would suggest to reopen this and keep it as a workaround for now.
@balloob What do you think?

Heronimonimo:
Another reason to use is to quickly update the graphs in Lovelace after new data is pulled in.

Another partial solution might be to create an option to set the overall refresh interval for Lovelace (not per card). Only problem with this is that it often you just want to manually refresh and not have cards that keep refreshing every few seconds.

balloob:
@Heronimonimo that's a frontend issue. We should not add workarounds to the Android app for this.

PR #20, opened 2019-11-09
Stop view from reloading on orientation change

timmo001:
## Description

The view reloads which causes the webview to reload when the screen orientation changes. This fixes that

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -23,9 +23,13 @@
             </intent-filter>
         </activity>
 
-        <activity android:name=".onboarding.OnboardingActivity" />
+        <activity
+            android:name=".onboarding.OnboardingActivity"
+            android:configChanges="orientation|screenSize" />
 
-        <activity android:name=".webview.WebViewActivity" />
+        <activity
+            android:name=".webview.WebViewActivity"
+            android:configChanges="orientation|screenSize" />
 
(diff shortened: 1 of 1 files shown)


marciogranzotto:
@CedrickFlocon I think this is a good compromise for now. It works and delivers good value to the users. It can be improved in the future.

CedrickFlocon:
This should not be that much complicate ?

marciogranzotto:
For what I understood in the Android Docs, webview's `saveState` no longer saves the display data:

from https://developer.android.com/reference/android/webkit/WebView.html#saveState(android.os.Bundle)
>Please note that this method no longer stores the display data for this WebView. The previous behavior could potentially leak files if restoreState(Bundle) was never called.

also: https://stackoverflow.com/a/2700865/3984316

CedrickFlocon:
Not sure if it's a good idea but we can retain the state with a Fragment

koying:
@CedrickFlocon Just in case, handling config changes through `onConfigurationChanged` still needs this Manifest patch.
Adding "android:configChanges" explicitly says you do that, indeed.

No actually overriding `onConfigurationChanged` in the app means webview is handling it.

nicklasl:
So what is the status of this? I have been waiting for this fix since the launch of the app..

balloob:
Going to close this as this PR has gone stale.

PR #36, opened 2019-11-20
Error Message on URL Parse Exception

BCNelson:
Show a simple error when there is a URL parse exception

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -82,6 +82,7 @@ dependencies {
     kapt "com.google.dagger:dagger-compiler:${daggerVersion}"
 
     implementation "androidx.appcompat:appcompat:$appCompatVersion"
+    implementation 'com.google.android.material:material:1.0.0'
     implementation "androidx.constraintlayout:constraintlayout:$constraintLayoutversion"
 
     implementation("com.jakewharton.threetenabp:threetenabp:$threeTenAbpVersion") {
@@ -95,7 +96,6 @@ dependencies {
     testImplementation "org.assertj:assertj-core:$assertJVersion"
     testImplementation "io.mockk:mockk:$mockkVersion"
     testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutinesVersion"
-    implementation 'com.google.android.material:material:1.0.0'
 }
 
 apply plugin: 'com.google.gms.google-services'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -95,6 +95,7 @@ dependencies {
     testImplementation "org.assertj:assertj-core:$assertJVersion"
     testImplementation "io.mockk:mockk:$mockkVersion"
     testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutinesVersion"
+    implementation 'com.google.android.material:material:1.0.0'
 }
 
 apply plugin: 'com.google.gms.google-services'
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupFragment.kt
@@ -49,15 +49,8 @@ class ManualSetupFragment : Fragment(), ManualSetupView {
         (activity as ManualSetupListener).onSelectUrl()
     }
 
-    override fun displayUrlError(text: String?){
-        view?.findViewById<TextInputLayout>(R.id.url_text_layout)?.error = text
-    }
-
-    override fun displayUrlError(error: URLError) {
-        view?.findViewById<TextInputLayout>(R.id.url_text_layout)?.error = when(error){
-            URLError.NO_PROTOCOL -> context?.getString(R.string.url_error_protocol_missing)
-            else -> context?.getString(R.string.url_error_undefined)
-        }
+    override fun displayUrlError(){
+        view?.findViewById<TextInputLayout>(R.id.url_text_layout)?.error = context?.getString(R.string.url_parse_error)
     }
 
     override fun onDestroy() {
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupFragment.kt
@@ -7,6 +7,7 @@ import android.view.ViewGroup
 import android.widget.Button
 import android.widget.EditText
 import androidx.fragment.app.Fragment
+import com.google.android.material.textfield.TextInputLayout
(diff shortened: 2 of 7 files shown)


app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupFragment.kt
```diff
@@ -48,6 +49,10 @@ class ManualSetupFragment : Fragment(), ManualSetupView {
         (activity as ManualSetupListener).onSelectUrl()
     }
 
+    override fun onErrorUrlParse(text: String?){
```
CedrickFlocon: The method name should be more an action like displayUrlError

app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupPresenterImpl.kt
```diff
@@ -24,6 +24,10 @@ class ManualSetupPresenterImpl @Inject constructor(
             url = URL(urlString)
         } catch (e: MalformedURLException) {
             Log.e(TAG, "Unable to parse url", e)
+            if(e.message != null){
+                view.onErrorUrlParse(e.message)
```
CedrickFlocon: We should use a more friendly message. You can let the view get string from a resource

app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupPresenterImpl.kt
```diff
@@ -24,6 +24,10 @@ class ManualSetupPresenterImpl @Inject constructor(
             url = URL(urlString)
         } catch (e: MalformedURLException) {
             Log.e(TAG, "Unable to parse url", e)
+            if(e.message != null){
```
CedrickFlocon: You have to test this case

BCNelson: I'm not sure what you mean

app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupPresenterImpl.kt
BCNelson: What is the best way to test what the error is in order to display the correct message?

app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupPresenterImpl.kt
```diff
@@ -24,6 +24,10 @@ class ManualSetupPresenterImpl @Inject constructor(
             url = URL(urlString)
```
balloob: Maybe out of scope for this PR, but it would be great if we could cut out the path of the URL, so that a user can copy paste any existing url including `/lovelace/home` in it, and we cut that part off to get the host url.

BCNelson: I can do that but will open a different PR for it

app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupPresenterImpl.kt
BCNelson: Okay where is the best place to have that logic
in the manualSetupPresenter or ManualSetupFragment

app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupPresenterImpl.kt
CedrickFlocon: I mean unit test but you did it

app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupPresenterImpl.kt
CedrickFlocon: You will not be able to getString from the presenter (do not use a context inside the presenter) so in the Fragment do something like that 
`view?.findViewById<TextInputLayout>(R.id.url_text_layout)?.error = getString(R.string.error_url)`

BCNelson: Do you know if URL() throws any exceptions other than No Protocol that we need to catch and display a custom error message for?

BCNelson: The problem that I see here is that it servery limits the Errors that can be shown to the user. right now we show on the No protocol error but there could be other errors in URL Parsing and it will annoy users if they get an irreverent error. 
I think that this solution allows for needed future expandability and communicating as fully as possible what went wrong.

CedrickFlocon: We should not rely on MalformedURLException error message as it's from the JVM.
We are not sure about the implementation on each platform (android version).

balloob: Yeah. We should just have an error message like: Unable to parse your Home Assistant URL. It should look like `https://example.com`.

BCNelson: Okay I changed it to what you suggested.

app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupView.kt
```diff
@@ -1,7 +1,14 @@
 package io.homeassistant.companion.android.onboarding.manual
 
+enum class URLError{
```
CedrickFlocon: you can remove this

BCNelson: Yep. Done

app/build.gradle
```diff
@@ -95,6 +95,7 @@ dependencies {
     testImplementation "org.assertj:assertj-core:$assertJVersion"
     testImplementation "io.mockk:mockk:$mockkVersion"
     testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutinesVersion"
+    implementation 'com.google.android.material:material:1.0.0'
```
CedrickFlocon: sorry I didn't see it early but could you put this line close to the appcompat dependency L84.
this place is for test dependency like mockk, assertJ ...

BCNelson: @CedrickFlocon Done

PR #41, opened 2019-11-21
Add kotlin android extensions and use it

marciogranzotto:
Fixes https://github.com/home-assistant/home-assistant-android/issues/40

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,6 +1,7 @@
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-kapt'
+apply plugin: "kotlin-android-extensions"
 apply plugin: 'com.google.firebase.appdistribution'
 apply plugin: 'com.github.triplet.play'
 apply plugin: 'de.mannodermaus.android-junit5'
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/OnboardingActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/OnboardingActivity.kt
@@ -16,7 +16,8 @@ import io.homeassistant.companion.android.onboarding.manual.ManualSetupListener
 import io.homeassistant.companion.android.webview.WebViewActivity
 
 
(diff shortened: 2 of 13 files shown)


homeassistant:
Hi @marciogranzotto, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant-android#41).

Thanks!

CedrickFlocon:
https://github.com/home-assistant/home-assistant-android/pull/4#issuecomment-549031814

marciogranzotto:
@CedrickFlocon I see this more like a good practice, rather than something we should only use if we have a lot of native UI components.

kotlin-android-extensions also has more features that we can use, lie auto-generate Parcelables with an annotation

I think using it makes the project more scalable and easier to maintain.

JBassett:
@CedrickFlocon I agree with @marciogranzotto we are only going to add more native views for settings  (device location, modifying registration, setting up other sensors we want to pass back, multiple urls for HA instance ...)

CedrickFlocon:
each 2 years we have a new way to do a findViewById (Manual, Butterknife, Databiding, and now Kotlin android extensions). I do not agree and I find it useless to add this dependency unless we have more than 50 findViewById in our project.

marciogranzotto:
@CedrickFlocon with that argument, I say that Dagger is useless on this project too, since there's not a need for such a tool to do so little. But that's a design choice and good practice.

kotlin-android-extensions has been around since the beginning of Android Kolin development and has since become the standard for binding views on Kotlin.
I think most contributors that have experience developing Android apps using Kotlin are used to using it and will most likely find it very strange to use `findViewByID`.

quthla:
Or just use https://developer.android.com/jetpack/androidx/releases/fragment?

Knoxie:
Agree with @JBassett and @marciogranzotto. Better to start now with it then have to refactor later on for consistency. Especially with something like this where it's widely used feature.

balloob:
This PR has gone stale and looks like it won't be accepted at this time anyway. Closing it.

PR #60, opened 2019-11-23
[WIP] Initial work to support multiple instances

w1ll1am23:
This is far from complete, but wanted to get some opinions on it. 

Basically is this something we would be okay with having or am I wasting my time? The goal is to allow users to switch between multiple HA servers using one app, without having to login/out every time. 

This has the ability to list multiple instances in the settings page, with the ability to select and switch to the other instance. You can also delete instances.

Things that I think need done before I would consider this complete.

- [ ] Only allow deleting instances that aren't the current
- [ ] Don't allow selecting the current instance
- [ ] If you are logged out and there are other instances fallback to another one don't go to onboarding
- [ ] Instances get "saved" even if they don't get logged in
- [ ] The settings activity UI should probably be moved to a fragment

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -83,6 +83,7 @@ dependencies {
 
     implementation "androidx.appcompat:appcompat:$appCompatVersion"
     implementation "androidx.constraintlayout:constraintlayout:$constraintLayoutversion"
+    implementation "androidx.recyclerview:recyclerview:$recyclerViewVersion"
 
     implementation("com.jakewharton.threetenabp:threetenabp:$threeTenAbpVersion") {
         exclude group: 'org.threeten'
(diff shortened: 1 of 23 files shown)


balloob:
When we list instances, we might want to make sure we also list the name of the user, since it would be possible to log into the same instance with two different accounts. 

So one thing about multiple instances is that we're going to complicate the hell out of things here. Now we're going to have multiple instances allowing you to send notifications. Now there will be multiple instances that we can report state too.

w1ll1am23:
@balloob I agree the username should be shown good point. I thought about the state reporting for multiple instances but figured we could explain somewhere that only the currently activated instance will be that, active so it would push/receive for only that instance?

I think the shared preference map/object being stored will need to be updated too. Probably something more like the following so we can add whatever in the future.

```json
[ 
   { 
      "host":"192.168.1.2",
      "user":"w1ll1am23",
      "session":{ 
         SESSION_OBJECT
      },
      "isActive":true
   }
]
```

balloob:
Reporting to the activated instance doesn't sound like a good plan. I would want to have my parents instance in it so I can help out, but I wouldn't want to ever report my location to their instance.

CedrickFlocon:
I think it will complicate a lot new feature development.
We should probably focus on the critical path before adding such advance feature(push, geolocation ...).

w1ll1am23:
@balloob @CedrickFlocon fair points.

Think there is any value in changing the structure of the shared preferences now so something like this will be easier in the future or should we just cross that bridge if/when we get there?

balloob:
I think that we could benefit from creating the structure in a way that allows us to add this feature in the future.

balloob:
So I am going to close this PR as we don't intent on merging it anytime soon. We can consider this in the future. For now, I recommend bookmarks 👍

PR #66, opened 2019-11-25
URL format

CedrickFlocon:


app/src/main/res/values/strings.xml
```diff
@@ -6,7 +6,7 @@
     <string name="scan_again">scan again</string>
     <string name="manual_setup">enter address manually</string>
     <string name="input_url">Home Assistant URL</string>
-    <string name="input_url_hint">URL</string>
```
balloob: It's either https or a port. I tend to use https://abcdefgh.nabu.casa or http://hassio.local:8123 as hint, although the latter only works if Android can resolve mdns, which I don't think it does

CedrickFlocon: I use Duck Dns so I have a URL like this : https://subdns.duckdns.org:8123/
But I can change if you think it's could be more helpful.

dshokouhi: My own URL is formatted similarly to @CedrickFlocon although I do not have duckdns, I just have my own domain/cert and just port forward for HA.

balloob: I'm also fine with adding a duckdns url in it.

balloob: If you set the 443 port on your router to point internally at 8123, you can drop the port. It looks a little cleaner, but yes, you can still use https and 8123 together.

dshokouhi: Yea in my own setup I have other things pointed to 443, so I just like to keep HA on its own port :)  But I understand I am not a common use case.  I do suspect that nabu casa and duckdns would be the most popular here.

CedrickFlocon: ok so let's go for this hint `https://homeassistant.duckdns.io:8123`. 
Is That ok for you ?

CedrickFlocon: (if it's ok for you, you can merge it)

balloob: Let's do `example.duckdns.org` and then let's merge.

PR #72, opened 2019-11-26
Add Lokalise SDK

robbiet480:
Adds the Lokalise SDK as [per their docs](https://docs.lokalise.com/en/articles/1400668-lokalise-android-sdk). Project ID and SDK token are baked in. SDK token is safe to be left in. It's in iOS as well.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt
@@ -1,9 +1,10 @@
 package io.homeassistant.companion.android.launch
 
+import android.content.Context
 import android.content.Intent
 import android.os.Bundle
-import android.content.Context
 import androidx.appcompat.app.AppCompatActivity
+import co.lokalise.android.sdk.core.LokaliseContextWrapper
 import io.homeassistant.companion.android.DaggerPresenterComponent
 import io.homeassistant.companion.android.PresenterModule
 import io.homeassistant.companion.android.background.LocationBroadcastReceiver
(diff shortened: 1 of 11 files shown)


robbiet480:
For the future, we should use the pre-release strings if we know that we are in pre-release e.g. development.

app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt
```diff
@@ -41,4 +43,8 @@ class LaunchActivity : AppCompatActivity(), LaunchView {
         presenter.onFinish()
         super.onDestroy()
     }
+
+    override fun attachBaseContext(newBase: Context?) {
```
CedrickFlocon: We should wrap all activity.
Or create a Base Activity which I do not like.

>We will need to inject the Lokalise SDK into the Activity context as well. To do so, we recommend you create a base Activity class and extend all your activities from it. Add the following code into your activity:

robbiet480: Will add to every activity

robbiet480:
@CedrickFlocon Can you just double check there's no hardcoded strings anywhere? I did myself but not sure I looked everywhere.

aleksandrov1995:
After this merge, the app constantly force closes for me Android 9 / Note 9. Tried wiping and reinstalling, but no change. I downgraded to the previous version.

robbiet480:
@aleksandrov1995 Can you try again?

allanpersson:
@robbiet480 same here on v29

CedrickFlocon:
I am on it

aleksandrov1995:
I just did, same thing again. I can't post a screenshot here.

jes1417:
same here v29 force closes on pixel4xl android 10

robbiet480:
Sorry all, @CedrickFlocon is working on a fix. Don't know how I missed this.

Issue #47, opened 2019-11-21
Quality tools


CedrickFlocon:
As all open source project we need to add quality tools. 
It's an open discussion to find out the best setup. 

e.g.
[SonarQube](https://sonarcloud.io/about)
[Detekt](https://github.com/arturbosch/detekt)
[Klint](https://github.com/pinterest/ktlint)

marciogranzotto:
I'm not very familiar with SonarQube or Detekt, but where Ktlint shines for me is that it follows the official Kotlin guidelines and there's no much configuration to be done.

For reference, as said on [this article](https://sourcelevel.io/blog/ktlint-or-detekt-which-linter-to-choose):
> Ktlint gives you less stuff to discuss. It follows the official style. When you let the lint too open for configuration, people create their “official” style, as happened to eslint. Too many style guide versions can lead to inconsistency and counterproductive discussions.

CedrickFlocon:
In my point of view we could use SonarQube for code coverage, bug and code smell.
And KLint for formating.

marciogranzotto:
Makes sense. I'll update my ktlint branch with the master changes and reopen the PR so we can get this part done. Then we can add another PR for SonarQube later.

PR #73, opened 2019-11-26
Code formatting and linter

marciogranzotto:
As discussed on #47 this project adds `ktlint` to the project.
It also auto-formats and solves the linter warnings

Full list of changes:
- Add ktlint to the project
- Auto format and fix linter issues
- After a `ktlint --android --apply-to-idea-project` command, changed the `gitignore` file to include styling idea files. That way any auto-formats that other contributor do after this, will be on the ktlint style.


.travis.yml
```diff
@@ -22,4 +22,5 @@ cache:
 
 script:
   - chmod ugo+x .travis/script.sh
-  - ./.travis/script.sh
\ No newline at end of file
+  - ./.travis/script.sh
+  - ./gradlew ktlintCheck
```
CedrickFlocon: You can put the task in the script.sh just after the lint command

build.gradle
```diff
@@ -6,24 +6,31 @@ buildscript {
...
 
 task clean(type: Delete) {
     delete rootProject.buildDir
 }
+
+ktlint {
+    android = true
+}
+
+apply plugin: "org.jlleitschuh.gradle.ktlint"
```
CedrickFlocon: Could you reorder the file like this :
buildscript
apply (file and plugin)
allprojects 
...

.gitignore
```diff
@@ -1,8 +1,134 @@
-.DS_Store
+# Created by https://www.gitignore.io/api/androidstudio
```
CedrickFlocon: We should not use a template. A lot of stuff is not used.

.gitignore
```diff
@@ -1,8 +1,134 @@
...
+
+# External native build folder generated in Android Studio 2.2 and later
+.externalNativeBuild
+
+# NDK
+obj/
+
+# IntelliJ IDEA
 *.iml
-.idea/
```
CedrickFlocon: commit IDE file is not a good practice. 
We will  link our project to an IDE and also a specific version.

marciogranzotto: ok, I reverted these changes

marciogranzotto:
@CedrickFlocon 
- Reverted `gitignore` changes
- Made the requested changes
- Added a Quality section on README

Issue #53, opened 2019-11-21
Monitor battery status


Nihvel:
Using the app as a kiosk, it will be handy to be able to monitor the status of the battery especially for those tablets that don't always stay attached to the wall and connected to a power source.

The Amazon Fire HD 7 (and probably few others) do not have the GPS module and applications like GPSLogger for logging the location and the battery can't properly work. A work-around I am using is by having a Fake GPS app running in background. Recently the tablet stopped sending messages to HA, don't understand why, therefore the battery status can't be tracked anymore.

The android app could solve this issue even though this means having it running in background. GPSLogger installed in my smartphone works great and the message with location and battery are sent every 15 minutes (take this as a little hint).

Thank you!

balloob:
Custom sensors is supported by the mobile app integration and could be included with location info https://developers.home-assistant.io/docs/en/app_integration_sensors.html

Nihvel:
Thanks, but I meant something more user-friendly: I don't understand a thing from that documentation because I'm a non-developer. And anyway, even if I were one, the issue opened here would like to ease this functionality. Waiting feedback! But thanks again for the project

dshokouhi:
@Nihvel the comment from balloob was more guidance to whomever is going to pick up this feature request to develop it.  You will need to wait for someone to implement it so you can get the battery information.

Nihvel:
@dshokouhi alright, makes more sense now, thanks for explaining ;)

Just as a note: I did enable the _mobile_app_ in _configuration_ and the integration saw the tablet on 2 rows. Not sure about which one was what. In the _developers_ tab it appeared the _device_tracker.my_tablet_ (both of them) but without much information.
Only _name_ and source. Source was set to _GPS_.
Probably, since this tablet does not have a GPS, it just didn't work.

So well, for whomever is going to pick this up: Probably an integrated mqtt client or rest api but anyway that wouldn't depend by the location, is a good solution.

Thanks again!

PR #74, opened 2019-11-26
Device Tracking

JBassett:
WIP
This PR addresses location and battery being sent back to HA.
#8 and #53 

Things remaining:
- [x] Validation on older android models
- [x] Unit Tests


quthla:
Are you planning to add geofencing for the zones defined in HA too instead of just continuous location updates?

andriej:
+1 for geofencing as an option

Knoxie:
Geofencing should work based on the zones already set in homeassistant I believe. The app just needs to report the position.

Unless your referring to editing and creating zones, in witch case I believe it's a different feature.

JBassett:
@Knoxie I think the Geofencing they are referencing this (https://developers.google.com/location-context/geofencing).  I hope to keep this PR small and expand on it in the future.  At some point I plan to add geofencing based on zones.  Need more hours in the day!

Knoxie:
I think duplication of zones in app for Android specific and then zones
already in Hass is a confusing implementation. But I think that's a design
choice for higher powers.

JBassett:
I don't think we would ever implement any custom zones in app.  We would only ever consume the zones that are setup in HA and add geofences around them to get faster updates when the fence is broken.

robbiet480:
@JBassett is correct. You shouldn’t be able to define custom zones in the app. The app should only consume existing zones. This is the same behavior as iOS.

quthla:
Yes, basically just fetch zones from HA, set up geofences, submit location to HA on geofence broadcast.

app/src/main/java/io/homeassistant/companion/android/background/LocationBroadcastReceiver.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    private fun requestUpdates(context: Context) {
+        Log.d(TAG, "Registering for location updates.")
+
+        if (ActivityCompat.checkSelfPermission(
+                context,
+                ACCESS_COARSE_LOCATION
+            ) != PackageManager.PERMISSION_GRANTED
+        ) {
+            Log.w(TAG, "Not starting location reporting because of permissions.")
```
NicoBijl: Maybe a toast message to tell the user that location reporting did not start because of missing permissions?

balloob: And then also ask for the permissions 👍

JBassett: If we make it to this point the user declined to give the permissions, do we really want to ask every time we start the app if they don't want location updates?

balloob: Oh, right, we should not.

quthla: Android will offer to deny any further permission requests after the first one has been denied.
But it'd make more sense if there was an option in the app to toggle location reporting. When turned on, just ask for permission there.

JBassett: Agreed, but out side of the scope of this PR I would say though.  We still need a settings UI for basically everything (urls, device registration changes, location settings)

JBassett:
I know I'm missing tests on LocationBroadcastReceiver.  I ran into a lot of trouble unit testing it because of how differently it was behaving when testing.  For example when the Intent is create it isn't valid and causes exceptions to be thrown.

CedrickFlocon:
In a product point of view, this feature will drain the battery of the user.
I think the good way to go is what @quthla say (fetch HA zone and register geofences on that zone).
BTW if we go on that solution, we should probably add a switch in the settings to turn this feature on/off (and do no rely on the permissions granted).

JBassett:
I agree that the zones are needed.  However I think we should have more than just a toggle.  I would say 3 settings, off, zone only, zone + background.  This was just too get the foundations of that.  The whole point of using the fused client is to preserve the battery.  It will batch the requests, and only give us a location when the phone already has a new location and is awoken.  I've been running this on my phone since I started working on this and it doesn't even register on the battery stats.  As far as the UI for the settings I wasn't sure what kind of style we wanted to go with and thought it would really make this PR huge.  The next PR I planned to work on was the zone based.

allanpersson:
@JBassett what you see suggesting is the ideal setup in my opinion. 

One thought to consider if you want to make a simple solution is make the device tracking, and "trigger" allow/disallow location of the app.
(The setting in Android -  apps - location permission)

Nihvel:
Can the "battery monitor" be not related to the gps location? For gps I mean the gps module on the devices. 
I already noted this in issue https://github.com/home-assistant/home-assistant-android/issues/53 . Tablets that do not have the gps module (amazon fire hd 7) have a bad time sending the location with the battery status in it - meaning that it does not work.

I've been using GPSLogger for a while on my phone and battery tracking is great. On the Fire hd 7 it doesn't work. 
I'm so thankful this issue is being processed, I just wish it won't just be a "clone" of GPSLogger
Thanks guys.

selinium:
I was wondering if the app could detect the HA server on the same network?
This way you could pause location tracking when the app could 'see' the HA server when you were definitely home.

JBassett:
@Nihvel The reason the battery is included with location is because it's how the integration on home assistant is setup.  We can add a completely separate sensor for battery only tracking, but it's outside of the scope of this PR I think.  I would create a new issue/feature request for this.

@selinium If you would like that feature I would say to add that as another issue/feature request.

balloob:
I like the idea of the three options. That way it's up to the user to balance accuracy vs battery impact.

JBassett:
@balloob I would rather do those changes incrementally rather than in one giant PR, but if you want I can add them all here.

balloob:
Oh yes small PRs all around please. I was mainly agreeing with the idea, not pushing to have it in 1 PR .

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
```diff
@@ -38,6 +39,42 @@ class IntegrationRepositoryImpl @Inject constructor(
...
+    // https://developers.home-assistant.io/docs/en/app_integration_sending_data.html#short-note-on-instance-urls
+    private suspend fun getUrls(): Array<String> {
+        val retVal = ArrayList<String>()
+        val webhook = localStorage.getString(PREF_WEBHOOK_ID)
+
+        var url = localStorage.getString(PREF_CLOUD_URL)
+        if (url != null)
+            retVal.add(url)
+
+        url = localStorage.getString(PREF_REMOTE_UI_URL)
```
CedrickFlocon: could you use a other var

data/src/test/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImplSpec.kt
```diff
@@ -110,6 +111,59 @@ object IntegrationRepositoryImplSpec : Spek({
...
+                    "locationName",
+                    arrayOf(45.0, -45.0),
+                    0,
+                    1,
+                    2,
+                    3,
+                    4,
+                    5
+                )
+                val updateLocationRequest = UpdateLocationRequest(
```
CedrickFlocon: no need to create a val

data/src/test/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImplSpec.kt
```diff
@@ -110,6 +111,59 @@ object IntegrationRepositoryImplSpec : Spek({
 
             }
         }
+
+        describe("location updated") {
```
CedrickFlocon: your algorithm is much more complicate, could you add some test case

app/src/main/java/io/homeassistant/companion/android/background/LocationBroadcastReceiver.kt
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Inject
+    lateinit var integrationUseCase: IntegrationUseCase
+
+    private val mainScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
+
+    override fun onReceive(context: Context, intent: Intent) {
+        DaggerReceiverComponent.builder()
+            .appComponent(Graph(context).appComponent)
```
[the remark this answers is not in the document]
JBassett: I will add some better handling around this.  I did this because the API doesn't guarantee that context.applicationContext will return an instance of our Application.  Most times it does, so we can get away with it but I'll handle both cases.

common/src/main/java/io/homeassistant/companion/android/common/dagger/Graph.kt
```diff
@@ -1,13 +1,12 @@
...
 
-import android.app.Application
 import android.content.Context
 import io.homeassistant.companion.android.common.LocalStorageImpl
 import kotlinx.coroutines.runBlocking
 
 
 class Graph(
-    private val application: Application
+    private val context: Context
```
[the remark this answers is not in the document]
JBassett: See other comment response for reasoning behind this.

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
```diff
@@ -38,6 +39,42 @@ class IntegrationRepositoryImpl @Inject constructor(
...
 
+    override suspend fun updateLocation(updateLocation: UpdateLocation) {
+        val updateLocationRequest = createUpdateLocation(updateLocation)
+        var wasSuccess = false
+        getUrls().forEach {
+            try {
+                wasSuccess =
+                    integrationService.updateLocation(it, updateLocationRequest).isSuccessful
+                return@forEach // If we get here update success!
+            } catch (e: Exception) {
```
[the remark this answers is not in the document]
JBassett: https://developers.home-assistant.io/docs/en/app_integration_sending_data.html

> To summarize, here's how requests should be made:
> 
> If you have a Cloudhook URL, use that until a request fails. When a request fails, go to step 2.
> If you have a remote UI URL, use that to construct a webhook URL: <remote_ui_url>/api/webhook/<webhook_id>. When a request fails, go to step 3.
> Construct a webhook URL using the instance URL provided during setup: <instance_url>/api/webhook/<webhook_id>.

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRequestData.kt
```diff
@@ -0,0 +1,13 @@
+package io.homeassistant.companion.android.data.integration
+
+sealed class IntegrationRequestData
```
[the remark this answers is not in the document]
JBassett: As we use more of the requests in the mobile_app integration we will need more subclasses of it.

app/src/main/java/io/homeassistant/companion/android/background/LocationBroadcastReceiver.kt
```diff
@@ -0,0 +1,152 @@
...
+            ACTION_PROCESS_LOCATION -> handleUpdate(context, intent)
+            else -> Log.w(TAG, "Unknown intent action: ${intent.action}!")
+        }
+    }
+
+    private fun ensureInjected(context: Context) {
+        if (context.applicationContext is HomeAssistantApplication) {
+            // If we have the application context then use it
+            DaggerReceiverComponent.builder()
+                .appComponent((context.applicationContext as HomeAssistantApplication).appComponent)
```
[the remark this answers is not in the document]
CedrickFlocon: BTW you can throw an Exception and avoid the else case if the context is not implementing `GraphComponentAccessor`

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
```diff
@@ -37,6 +38,46 @@ class IntegrationRepositoryImpl @Inject constructor(
...
+    private suspend fun getUrls(): Array<String> {
+        val retVal = ArrayList<String>()
+        val webhook = localStorage.getString(PREF_WEBHOOK_ID)
+
+        localStorage.getString(PREF_CLOUD_URL)?.let {
+            retVal.add(it)
+        }
+
+        localStorage.getString(PREF_REMOTE_UI_URL)?.let {
+            val base = if (it.last() == '/') it else "$it/"
```
CedrickFlocon: you can use okhttp to help you build this url 
okhttp3.HttpUrl
String.toHttpUrl()

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRequestData.kt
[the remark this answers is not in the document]
CedrickFlocon: It's easier to add it if we will need it than remove it if we do not need it.

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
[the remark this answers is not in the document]
CedrickFlocon: ok so if I well understand you should return only if it's a success.

app/build.gradle
```diff
@@ -92,11 +92,14 @@ dependencies {
...
     testImplementation "org.spekframework.spek2:spek-runner-junit5:$spek2Version"
     testImplementation "org.assertj:assertj-core:$assertJVersion"
     testImplementation "io.mockk:mockk:$mockkVersion"
     testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutinesVersion"
 }
 
-apply plugin: 'com.google.gms.google-services'
\ No newline at end of file
+apply plugin: 'com.google.gms.google-services'
+apply plugin: 'kotlin-android-extensions'
```
CedrickFlocon: I think you do not use this.
If you use it could you move it with all other plugin

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
JBassett: I migrated to using HttpUrl, but we still need the `/` check to ensure we don't get a double `/` and break the integration.

CedrickFlocon: A good unit test could help you assert this behaviour 🙄 

````
"https://www.home-assistant.io".toHttpUrl().newBuilder().addPathSegment("foo")
"https://www.home-assistant.io/".toHttpUrl().newBuilder().addPathSegment("foo")
"https://www.home-assistant.io/bar".toHttpUrl().newBuilder().addPathSegment("foo")
"https://www.home-assistant.io/bar/".toHttpUrl().newBuilder().addPathSegment("foo")
````

>https://www.home-assistant.io/foo
https://www.home-assistant.io/foo
https://www.home-assistant.io/bar/foo
https://www.home-assistant.io/bar/foo

File: app/build.gradle [hunk 12]
```diff
@@ -5,6 +5,7 @@ apply plugin: 'com.google.firebase.appdistribution'
 apply plugin: 'com.github.triplet.play'
 apply plugin: 'de.mannodermaus.android-junit5'
 apply plugin: 'io.fabric'
+apply plugin: 'com.google.gms.google-services'
```
CedrickFlocon: https://developers.google.com/android/guides/google-services-plugin

>Add dependencies for basic libraries required for the services you have enabled. This step requires that the apply plugin: 'com.google.gms.google-services' line be at the bottom of your app/build.gradle file so that no dependency collisions are introduced. You can see the result of this step by running ./gradlew :app:dependencies.

common/src/main/java/io/homeassistant/companion/android/common/dagger/Graph.kt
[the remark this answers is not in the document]
CedrickFlocon: this could be revert

data/src/test/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImplSpec.kt
```diff
@@ -109,5 +115,196 @@ object IntegrationRepositoryImplSpec : Spek({
...
+                        location.course,
+                        location.verticalAccuracy
+                    )
+                )
+                beforeEachTest {
+                    coEvery { localStorage.getString("cloud_url") } returns "http://best.com/hook/id"
+                    coEvery { localStorage.getString("remote_ui_url") } returns "http://better.com"
+                    coEvery {
+                        integrationService.updateLocation(
+                            any(), // "http://example.com/api/webhook/FGHIJ",
```
CedrickFlocon: Could you test this case :
- [ ] let the `cloud_url` fail but let "remote_ui_url" succeed
- [ ] all url is failing

data/src/test/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImplSpec.kt
```diff
@@ -105,7 +112,300 @@ object IntegrationRepositoryImplSpec : Spek({
...
+
+                beforeEachTest {
+                    coEvery { localStorage.getString("cloud_url") } returns "http://best.com/hook/id"
+                    coEvery { localStorage.getString("remote_ui_url") } returns "http://better.com"
+                    coEvery {
+                        integrationService.updateLocation(
+                            "http://best.com/hook/id".toHttpUrl(),
+                            any() // integrationRequest
+                        )
+                    } throws Exception()
```
CedrickFlocon: Ok so here is my point :
If you replace the `throws Exception()` by 
````
                    } returns mockk {
                        every { isSuccessful } returns false
                    }
````
The test wont pass. 
But I think we should have the same behaviour ? (if it's the case could you add this test and fix the code)

allanpersson:
Device tracking and data seems to be working properly 🙏
I don't know if I was too inpatient (and went in manually and activated) but there was no pop-up asking for permission to track location. (update from v27➡️v28

aleksandrov1995:
No pop - up for me as well. Also the device was added twice. I'm apologizing if this isn't the place to share this. 
I'm sure these will be sorted out as time goes by.
Thanks for the awesome work!

JBassett:
Because we don't have a UI for this yet we don't prompt for permission except on initial onboarding.  If you give permission to location at all times and force close the app and restart it, you should start getting updates.

PR #95, opened 2019-12-04
Initial work on zone based location tracking.

JBassett:


domain/src/main/java/io/homeassistant/companion/android/domain/integration/EntityResponse.kt
```diff
@@ -0,0 +1,12 @@
+package io.homeassistant.companion.android.domain.integration
+
+import java.util.Calendar
+
+data class EntityResponse(
+    val entityId: String,
+    val state: String,
+    val attributes: Map<String, String>,
```
balloob: This is `Map<String, Any>`

balloob: If it helps, in TypeScript we have a base entity and then `ZoneEntity`, `LightEntity` etc where we fill in the attributes with the types we expect.

domain/src/main/java/io/homeassistant/companion/android/domain/integration/EntityResponse.kt
```diff
@@ -0,0 +1,12 @@
...
+    val lastChanged: Calendar,
+    val lastUpdated: Calendar,
+    val context: Map<String, String>
```
balloob: This also is `Any` for values

balloob:
This looks good on the surface. I don't know enough to judge if enough tests / background work is good. So a 👍 from me.

robbiet480:
Closing in favor of just merged #129

PR #125, opened 2019-12-08
Added Troubleshooting section

allanpersson:
With help/information

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,3 +33,14 @@ To run a check with an auto-format:
 ```bash
 ./gradlew ktlintFormat
 ```
+
+## Troubleshooting
+
+The steps can vary depending on the specific device.
+
+**Duplicate devices:**
+- This issue is fixed with the upcoming release of Home Assistant version 0.103.
+
+**Location is not updating:**
+1. Make sure you have granted the Home Assistant app access to your location. (Settings - Apps - Permissions - Location)
+2. Make sure battery optimization for the the Home Assistant app is **off**


balloob:
We should point people at https://companion.home-assistant.io/ instead.

allanpersson:
@ballob good idea, but how do we "seperate" the ios and android app there?
Everything is build up there on the ios app at the moment, so maybe:
- https://companion.home-assistant.io/ landing page with redirect to 2 links (ios and android)
- https://companion.home-assistant.io/ios the existing is renamed to ios
- https://companion.home-assistant.io/android new page build on same frame as ios

Or what do you think?

dshokouhi:
I noticed that some of the instructions on the companion site like around removing config entry and entities are helpful for Android as well.  Its probably time we review that site and find a way to split things up? Maybe make the pages more generalized and have ios and android sections at the bottom of each page for OS specific things?

allanpersson:
@dshokouhi I agree, some stuff is relevant for both Android and ios. The current instructions is very specific ios, and I fear if we don't make two seperate instructions people will be confused and alot off issue/support tickets will come, if we just put android instructions in the same place also.

SeanPM5:
Tom Brien and I discussed this briefly when we were writing all the docs for the companion site. Here is what I wrote back in May. https://github.com/home-assistant/companion.home-assistant/issues/67#issuecomment-491198726

>Another benefit I am noticing just now about this structure, is that it's _mostly_ platform agnostic except for the Integrations section. So theoretically, when the Android app becomes a thing it could possibly share these same docs, and we would simply segment the Integrations section into sub-sections (Integrations/iOS and Integrations/Android).

I think that by sharing the same docs it'll make them better for everyone as the entire community would be contributing to one document. More eyeballs reading over it, improving it and making it easy to understand. One easy to remember URL to point users to. Avoids issues where only one of the docs gets updated and not the other.

This would require some work to generalize the existing pages. And then for the really platform-specific stuff, that would be handled like this:

`/integrations/ios/`
`/integrations/android/`

So there would be an "Integrations" landing page that asks what platform you use, the user clicks the Android logo or the iOS logo and then it takes you to one of the above links.

I don't know if this idea is feasible in the end, or what everyone would prefer though.

allanpersson:
@seanpm5 I think it should be handled with a choice of are you using android og ios, and then to a underpage specific for each integration.
I feel android is more complicated than ios due to almost every big manufactur make adjustments to the way they handle Android, and therefor trying to help people avoid problems is harder in my opinion 🙏

balloob:
Thanks for chiming in @SeanPM5. 

Since both apps are using mobile_app integration, and we're aiming for feature parity for configuration, I think that a lot of trouble shooting can be shared.

robbiet480:
Let's continue the larger conversation about splitting up docs for Android & iOS at https://github.com/home-assistant/companion.home-assistant/issues/137.

As for this PR, I think we should just have a link to the companion docs if anything in troubleshooting.

allanpersson:
Awaiting decision on how to approach this, so for now closing this PR 🙏

PR #129, opened 2019-12-09
Feature/location settings

JBassett:
Building upon https://github.com/home-assistant/home-assistant-android/pull/95.  This PR has settings for location tracking.  Currently just enable or disable background and/or zone based tracking.

Fixes: https://github.com/home-assistant/home-assistant-android/issues/86 and https://github.com/home-assistant/home-assistant-android/issues/108


bramkragten:
Why would you remove the translations?

JBassett:
@bramkragten The Lokalise sdk has been removed from the project for now.  Those translations were added with that service.  But because it is no longer part of the project builds will not pass if you do not translate to all languages that are on file.  I didn't want to use Google Translate because that is a less than reliable way to translate.  Until Lokalise is added back in, we should just stick with no translations.

app/src/main/res/values/strings.xml
```diff
@@ -1,23 +1,28 @@
...
+    <string name="attempting_registration">Attempting to register application…</string>
+    <string name="unable_to_register">Unable to Register Application</string>
+    <string name="error_with_registration">Please check to ensure you have the mobile_app
 integration enabled on your home assistant instance.</string>
-  <string name="url_parse_error">Unable to parse your Home Assistant URL. It should look like https://example.com</string>
+    <string name="url_parse_error">Unable to parse your Home Assistant URL. It should look like https://example.com</string>
+    <string name="pref_location_zone_title">Zone Based Tracking</string>
+    <string name="pref_location_zone_summary">This will take the zones you have in Home Assistant and place geofences around them, updating when you enter or leave.</string>
+    <string name="pref_location_background_title">Background Location Tracking</string>
+    <string name="pref_location_background_summary">This will attempt to update your location no matter what in the background on given intervals.</string>
```
Knoxie: Bette wording needed.

Attempt to update location from background on given interval.

app/src/main/res/values/strings.xml
```diff
@@ -1,23 +1,28 @@
...
+    <string name="skip">Skip</string>
+    <string name="retry">Retry</string>
+    <string name="attempting_registration">Attempting to register application…</string>
+    <string name="unable_to_register">Unable to Register Application</string>
+    <string name="error_with_registration">Please check to ensure you have the mobile_app
 integration enabled on your home assistant instance.</string>
-  <string name="url_parse_error">Unable to parse your Home Assistant URL. It should look like https://example.com</string>
+    <string name="url_parse_error">Unable to parse your Home Assistant URL. It should look like https://example.com</string>
+    <string name="pref_location_zone_title">Zone Based Tracking</string>
+    <string name="pref_location_zone_summary">This will take the zones you have in Home Assistant and place geofences around them, updating when you enter or leave.</string>
```
Knoxie: More concise comment/description:

Import Home Assistant zones for phone geofence service.

bramkragten:
@JBassett Shouldn't we change CI then? I also don't think that adding Lokalise SDK back in will fix this, as that is only for dynamically updating the translations on device and won't give any initial data in the string files? (maybe I'm wrong...)

JBassett:
@bramkragten sounds like a question for @robbiet480 since he added the SDK and translations to begin with.

balloob:
Looking at the screenshot, there seems to be too much of a margin on the left side of the settings?

balloob:
Can you target this PR on the other PR for now, so that we can see what are the changes for just this PR ?

JBassett:
@balloob I can't change the target but this link is the comparison.  https://github.com/JBassett/home-assistant-android/compare/feature/zoneBasedLocationTracking...JBassett:feature/locationSettings
Was hoping to have other PR merged first so I didn't run into this problem, I must be working to fast. :-)

JBassett:
@balloob as far as the margin it appears that this is part of the material design standard:
https://material.io/design/platform-guidance/android-settings.html
However we are missing the action bar at the top with the back button and title.  I can try to add that in, it will definitely make it look better.

balloob:
The change set looks good to me. Nothing stands out to me 👍 Waiting for Cedric for merging.

quthla:
@jbasset the space is meant for icons. If there are none, you can use

https://developer.android.com/reference/androidx/preference/Preference.html#setIconSpaceReserved(boolean)

JBassett:
@quthla if we plan to never have icons then yes I could set that.  But if we plan to add icons to some of the settings then we shouldn't.  I can add the HA icon to the version number at least.  Maybe a GPS Locked icon to zone, and gps not fixed to background.

app/src/main/res/values/strings.xml
gibwar: I disagree, that sounds robotic. Something like "Import existing Home Assistant zones as geofences for zone based tracking." would be better.

app/src/main/res/values/strings.xml
gibwar: How about "Update your location behind the scenes, periodically." since you can't configure the "given interval".

app/src/main/res/values/strings.xml
Knoxie: I was trying to get it to be short enough for one line. No need for extra space on a setting toggle. I just felt the current one didn't read well.

app/src/main/res/values/strings.xml
Knoxie: Sounds good, I thought maybe the interval was going to be settable in future.

Knoxie:
> @quthla if we plan to never have icons then yes I could set that. But if we plan to add icons to some of the settings then we shouldn't. I can add the HA icon to the version number at least. Maybe a GPS Locked icon to zone, and gps not fixed to background.

https://images.app.goo.gl/8D47FHNZqZnaGeTg7

Something available similar to this maybe for zones?

JBassett:
Added icons from https://materialdesignicons.com/
settings-icons

quthla:
Looks good but a proper accent color is missing in colors.xml

robbiet480:
Sorry I missed my pings, wasn't around the last few days. No problem with removing the translations @JBassett. Once this gets merged someone with powers on Lokalise (me, Paulus and Cedrick) needs to upload the English strings and download a new bundle which will contain the English strings in all other languages until they get translated.

PR #140, opened 2019-12-12
Basic Notification Support

### LINKED ISSUE #7: Add push notifications support
The mobile app integration has support for push notifications. This allows this app to register as a target. To do so, we should set `push_token` and `push_url` during the mobile app registration from #5 ([docs](https://developers.home-assistant.io/docs/en/app_integration_notifications.html)). We can use the same one as iOS.

@robbiet480 could you briefly explain how to get `push_token` and `push_url`?

JBassett:
This PR will do the following:
- Implement foreground receiving
- Push token and url to HA for notifications
- Update HA with new tokens

Fixes #7

### DIFF
--- a/domain/src/main/java/io/homeassistant/companion/android/domain/integration/IntegrationUseCase.kt
+++ b/domain/src/main/java/io/homeassistant/companion/android/domain/integration/IntegrationUseCase.kt
@@ -3,6 +3,7 @@ package io.homeassistant.companion.android.domain.integration
 interface IntegrationUseCase {
 
     suspend fun registerDevice(deviceRegistration: DeviceRegistration)
+    suspend fun updateRegistration(deviceRegistration: DeviceRegistration)
 
     suspend fun isRegistered(): Boolean
 
@@ -11,10 +12,8 @@ interface IntegrationUseCase {
     suspend fun getZones(): Array<Entity<ZoneAttributes>>
 
     suspend fun setZoneTrackingEnabled(enabled: Boolean)
-
     suspend fun isZoneTrackingEnabled(): Boolean
 
     suspend fun setBackgroundTrackingEnabled(enabled: Boolean)
-
     suspend fun isBackgroundTrackingEnabled(): Boolean
 }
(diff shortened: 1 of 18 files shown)


robbiet480:
🎉

domain/src/main/java/io/homeassistant/companion/android/domain/integration/IntegrationUseCase.kt
```diff
@@ -3,6 +3,7 @@ package io.homeassistant.companion.android.domain.integration
 interface IntegrationUseCase {
 
     suspend fun registerDevice(deviceRegistration: DeviceRegistration)
+    suspend fun updateRegistration(deviceRegistration: DeviceRegistration)
```
balloob: Not all fields can be updated. From top of my head, I think we can't update encryption and maybe app id?

robbiet480: [Correct](https://developers.home-assistant.io/docs/en/app_integration_sending_data.html#update-registration).

Only `app_data`, `app_version`, `device_name`, `manufacturer`, `model` and `os_version`.

JBassett: I'll update the usecase to only have those fields, and reuse the objects on the backend side.

andriej:
How can I activate those?

dshokouhi:
@andriej restart HA to get the `notify.mobile_app_*` service call registered, also in the future hop on discord in #android we discuss these things there.

dmaciaszek:
After restart HA I get `notify.mobile_app_*` as you suggested but when I call it by dev tools with payload `message: testing notification` nothing happen. In log I have error `Requested entity was not found.`. Am I doing something wrong?

Zixim:
`notify.mobile_app_*`
Replace * with the correct device ?

dmaciaszek:
Yes, selected from dropdown. Problem solved in https://github.com/home-assistant/home-assistant-android/issues/147

PR #155, opened 2019-12-15
Fix dialog button color in webview.

gmatyszczak:
This PR fixes bug with "no button" in dialog https://github.com/home-assistant/home-assistant-android/issues/90

Basically it's not a good idea to set button style globally for whole theme, it can lead to such issues. Better set it explicitly in layouts.


homeassistant:
Hi @gmatyszczak, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant-android#155).

Thanks!

PR #158, opened 2019-12-15
Allow editing of Device Name

JBassett:
Fixes:  https://github.com/home-assistant/home-assistant-android/issues/143

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -65,6 +65,10 @@ android {
             }
         }
     }
+
+    lintOptions {
+        disable 'MissingTranslation'
+    }
 }
 
 play {
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -65,10 +65,6 @@ android {
             }
         }
     }
-
-    lintOptions {
-        disable 'MissingTranslation'
-    }
 }
 
 play {
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
   <string name="app_name">Home Assistant</string>
   <string name="scanning_network">Scanning the network 
 for Home Assistant</string>
@@ -25,8 +25,8 @@ integration enabled on your home assistant instance.</string>
   <string name="pref_location_background_title">Background Location Tracking</string>
   <string name="pref_location_background_summary">Update your location behind the scenes, periodically.</string>
   <string name="application_version">Application Version</string>
-  <string name="location">Location</string>
-  <string name="app_version_info">App Version Info</string>
-  <string name="device_registration">Device Registration</string>
-  <string name="device_name">Device Name</string>
+  <string name="location" tools:ignore="MissingTranslation">Location</string>
+  <string name="app_version_info" tools:ignore="MissingTranslation">App Version Info</string>
+  <string name="device_registration" tools:ignore="MissingTranslation">Device Registration</string>
+  <string name="device_name" tools:ignore="MissingTranslation">Device Name</string>
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -27,4 +27,6 @@ integration enabled on your home assistant instance.</string>
   <string name="application_version">Application Version</string>
   <string name="location">Location</string>
   <string name="app_version_info">App Version Info</string>
+  <string name="device_registration">Device Registration</string>
+  <string name="device_name">Device Name</string>
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -25,4 +25,6 @@ integration enabled on your home assistant instance.</string>
   <string name="pref_location_background_title">Background Location Tracking</string>
   <string name="pref_location_background_summary">Update your location behind the scenes, periodically.</string>
   <string name="application_version">Application Version</string>
+  <string name="location">Location</string>
+  <string name="app_version_info">App Version Info</string>
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<resources xmlns:tools="http://schemas.android.com/tools">
+<resources>
     <string name="app_name">Home Assistant</string>
     <string name="scanning_network">Scanning the network
 for Home Assistant</string>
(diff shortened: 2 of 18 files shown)


app/src/main/res/xml/preferences.xml
```diff
@@ -1,7 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
     <PreferenceCategory
-        android:title="Location">
+        android:title="Device Registration">
+        <EditTextPreference
+            android:key="registration_name"
+            android:title="Device Name"
```
balloob: We should not put strings directly in the UI

app/src/main/res/xml/preferences.xml
```diff
@@ -1,7 +1,14 @@
...
+            android:summary="The name of the device listed in Home Assistant"/>
```
balloob: This summary seems meh, as that's right now a use case, but it could be used for many different things. Better not have any summary ?

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
```diff
@@ -52,6 +65,35 @@ class IntegrationRepositoryImpl @Inject constructor(
...
+        )
+    }
+
+    private suspend fun persistDeviceRegistration(deviceRegistration: DeviceRegistration) {
+        if (deviceRegistration.appVersion != null)
+            localStorage.putString(PREF_APP_VERSION, deviceRegistration.appVersion)
+        if (deviceRegistration.deviceName != null)
+            localStorage.putString(PREF_DEVICE_NAME, deviceRegistration.deviceName)
+        if (deviceRegistration.manufacturer != null)
+            localStorage.putString(PREF_MANUFACTURER, deviceRegistration.manufacturer)
```
balloob: Why would we store these in local storage, can't we just fetch them from Android each time ?

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
```diff
@@ -143,18 +185,23 @@ class IntegrationRepositoryImpl @Inject constructor(
...
         return RegisterDeviceRequest(
             deviceRegistration.appId,
             deviceRegistration.appName,
-            deviceRegistration.appVersion,
-            deviceRegistration.deviceName,
-            deviceRegistration.manufacturer,
-            deviceRegistration.model,
+            deviceRegistration.appVersion ?: oldDeviceRegistration.appVersion,
+            deviceRegistration.deviceName ?: oldDeviceRegistration.deviceName,
+            deviceRegistration.manufacturer ?: oldDeviceRegistration.manufacturer,
```
balloob: This seems weird. Why not just refer to the original source of this data in this function ?

app/build.gradle
```diff
@@ -65,6 +65,10 @@ android {
             }
         }
     }
+
+    lintOptions {
+        disable 'MissingTranslation'
```
balloob: Let's not do this. This is a slippery slope.

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
JBassett: We could request it every time.  The reason I stored them is so that on subsequent updates we only need to pass what has changed rather than everything like the call requires.

app/build.gradle
JBassett: The problem is every time anyone creates a string resource they will need to ignore the missing translation.  Is someone with permissions to Lokalise going to go back through and remove these when they update the translations?

bramkragten: We can't provide translations for every string in every language? That's what we have the community and Lokalise for. So I agree that we should disable the check for translations, just English should be present.

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
JBassett: The request requires all data be passed, however it seemed strange to have to send all the data every time down the whole stack.  It seemed simpler to persist the values and only send potential deltas on update rather than the whole request.  For example if I'm changing device name, why should I need to generate the appData with the push url and token.

balloob: What I meant was, why do we need all data to be passed ? Isn't this value hardcoded per device ?

JBassett: Yes, but the way we have our application structured we don't have access to that in the libraries.  We might be able to inject them via Dagger?  I don't claim to be an expert on that, but I can look into it.

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
JBassett: Migrated this to be injected.

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
JBassett: Injected now.

JBassett:
Screenshot of settings:
Screenshot_20191217-114548

quthla:
Maybe use summary provider for device name?

JBassett:
@quthla: https://github.com/home-assistant/home-assistant-android/pull/158#discussion_r358068061

quthla:
@JBassett Yes, I mean a summary provider, which prints the value of the preference below the label. Not a description

JBassett:
@quthla good call!

Screenshot_20191217-141623

dshokouhi:
so will this update the notification service name for existing integrations or will we need to remove the existing integration and clear data in the app, log back in and then restart HA to see the change?

JBassett:
Should just needed to change setting then restart HA

app/build.gradle
balloob: So I suggest that we not disable global lint option. We either use @CedrickFlocon suggestion and disable it internally. However, I prefer that we just add the string right away.

balloob: I know that for other projects we automatically upload the keys to Lokalise when they are merged to master. Is that something we can set up there too  ?

JBassett: Looks like they have github/git integrations.  https://docs.lokalise.com/en/articles/1684090-github  I don't have lokalise access or repo access to set it up though. 

Even with that setup however the workflow is:
- Merge developers PR
- Lokalise pulls new files from master
- Wait for awesome team to translate
- Lokalise gets updated, PR auto created with new translations.

That still means a dev needs to be able to get a passing build with missing strings.

balloob: Done. Give me your Lokalise username and I'll add you to the project too.

app/src/main/res/values/strings.xml
```diff
@@ -25,4 +25,8 @@ integration enabled on your home assistant instance.</string>
   <string name="pref_location_background_title">Background Location Tracking</string>
   <string name="pref_location_background_summary">Update your location behind the scenes, periodically.</string>
   <string name="application_version">Application Version</string>
+  <string name="location" tools:ignore="MissingTranslation">Location</string>
```
balloob: Okay. Sorry to bring this up again. I think I am trying to understand 😅 

Is this error because the translations of this string is missing? In that case I think that we could just ignore it globally, because it will get uploaded to Lokalise and will get translated when the community does so.

Otherwise we wouldn't be able to add new languages either.

bramkragten: That's what I understood it was, it will fail if any of the included languages doesn't have the string, right?

balloob: Oh, in that case we can disable it globally, because as long as we have English and that's used as a fallback, we're good.

JBassett: I'll revert my revert. :-)

balloob: Sorry!

JBassett: No problem, done.

PR #163, opened 2019-12-16
Fix some styling

bramkragten:
Because `colorControlNormal` was set to white, the underline of a inactive textbox would be white. Making it hard to see where to input text.

Before:
Screenshot_2019-12-17-13-16-31-442_io homeassistant companion android
After:
Screenshot_2019-12-17-13-17-38-500_io homeassistant companion android

(nice and big screenshots 👎 )

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -22,7 +22,7 @@
     </style>
 
     <style name="Widget.HomeAssistant.Button.Colored" parent="Widget.MaterialComponents.Button">
-        <item name="android:textColor">@android:color/white</item>
+        <item name="android:textColor">@color/onPrimary</item>
     </style>
 
     <style name="Widget.HomeAssistant.Button.Outlined" parent="Widget.MaterialComponents.Button.OutlinedButton">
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -31,6 +29,11 @@
         <item name="android:textColor">@color/colorPrimary</item>
     </style>
 
+    <style name="Toolbar">
```
gmatyszczak: According to good practices of styles naming, this should be named `Theme.HomeAssistant.Toolbar`, that would also keep it consistent with the rest of styles.

There's also a good talk from the latest Android Dev Summit about this topic if you are interested https://youtu.be/Owkf8DhAOSo

balloob:
Let's include screenshots when we make UI changes.

app/src/main/res/values/styles.xml
gabrielemariotti: @gmatyszczak  It is true if it is a style.
In this case it is used as a theme overlay (`android:theme="@style/Toolbar"`).

It could be 

`<style name="ThemeOverlay.HomeAssistant.Toolbar"  
       parent="@style/ThemeOverlay.MaterialComponents.Toolbar.Primary">`

bramkragten:
> Let's include screenshots when we make UI changes.

Added screenshots

app/src/main/res/values/styles.xml
gmatyszczak: Correct, it should be ThemeOverlay in this case ;)

bramkragten:
> Maybe it is better to use the name `colorOnPrimary` instead of `<color name="onPrimary">...</color>`.

Not sure I completely understand what you mean

gabrielemariotti:
> mpletely understand what you mean

Sorry, it wasn't clear.
I meant to use the name `colorOnPrimary` instead of `onPrimary`.

PR #168, opened 2019-12-17
Ability to disregard location updates that are not accurate.

JBassett:
Fixes https://github.com/home-assistant/home-assistant-android/issues/166 
Fixes https://github.com/home-assistant/home-assistant-android/issues/145

The default is set to a 50 meter radius.  Anything less accurate than that is ignored and not set back to HA.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/background/LocationBroadcastReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/background/LocationBroadcastReceiver.kt
@@ -12,6 +12,7 @@ import android.util.Log
 import com.google.android.gms.location.Geofence
 import com.google.android.gms.location.GeofencingEvent
 import com.google.android.gms.location.GeofencingRequest
+import com.google.android.gms.location.LocationCallback
 import com.google.android.gms.location.LocationRequest
 import com.google.android.gms.location.LocationResult
 import com.google.android.gms.location.LocationServices
@@ -24,6 +25,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 
 class LocationBroadcastReceiver : BroadcastReceiver() {
 
@@ -49,8 +51,8 @@ class LocationBroadcastReceiver : BroadcastReceiver() {
         when (intent.action) {
             Intent.ACTION_BOOT_COMPLETED -> setupLocationTracking(context)
             ACTION_REQUEST_LOCATION_UPDATES -> setupLocationTracking(context)
-            ACTION_PROCESS_LOCATION -> handleLocationUpdate(context, intent)
-            ACTION_PROCESS_GEO -> handleGeoUpdate(context, intent)
+            ACTION_PROCESS_LOCATION -> runBlocking { handleLocationUpdate(context, intent) }
+            ACTION_PROCESS_GEO -> runBlocking { handleGeoUpdate(context, intent) }
             else -> Log.w(TAG, "Unknown intent action: ${intent.action}!")
         }
     }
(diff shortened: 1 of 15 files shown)


JBassett:
Screenshot_20191217-133036

Screenshot_20191217-133438

app/src/main/res/values/strings.xml
```diff
@@ -25,4 +25,6 @@ integration enabled on your home assistant instance.</string>
   <string name="pref_location_background_title">Background Location Tracking</string>
   <string name="pref_location_background_summary">Update your location behind the scenes, periodically.</string>
   <string name="application_version">Application Version</string>
+  <string name="pref_accuracy_requirement_title" tools:ignore="MissingTranslation">Accuracy Requirement</string>
+  <string name="pref_accuracy_requirement_summary" tools:ignore="MissingTranslation">Disregard location updates that are below this accuracy threshold.</string>
```
dshokouhi: Should we mention if this is in meter or miles etc...?

JBassett: Opps, done. :+1:

quthla:
What about https://github.com/home-assistant/home-assistant-android/issues/145#issuecomment-566491176

Even inaccurate locations might be correct and just ignoring them will probably cause other issues.

ioangogo:
I would argue locations with an inaccuracy of more than 100m is never going to be remotely right most of the time, and typically soon after the phone would be able to get a more accurate location off another source or from locking on to more satellites

quthla:
While this is not an issue for continuous location updates, it is for geofencing. Intents are only sent when entering or exiting a zone.

For example, if the exit intent is ignored but was correct, there will not be another exit intent.

balloob:
I agree with @quthla here. Not being marked as leaving a zone is not good. Would we want to add the filter only to the timed updates then? Makes me wonder if this should be filtered in the app or in Home Assistant ?

quthla:
In my opinion, the best solution would be if app tried to request a more accurate location when it received a geofence intent with an inaccurate location. This should solve the location bouncing issues without causing any negative side effects.

JBassett:
I'll move current logic to only background updates and I'll add a one time high accuracy location request after a non accurate result.

app/src/main/java/io/homeassistant/companion/android/background/LocationBroadcastReceiver.kt
```diff
@@ -126,25 +128,35 @@ class LocationBroadcastReceiver : BroadcastReceiver() {
...
+    private suspend fun handleGeoUpdate(context: Context, intent: Intent) {
         Log.d(TAG, "Received geofence update.")
         val geofencingEvent = GeofencingEvent.fromIntent(intent)
         if (geofencingEvent.hasError()) {
             Log.e(TAG, "Error getting geofence broadcast status code: ${geofencingEvent.errorCode}")
             return
         }
 
-        sendLocationUpdate(geofencingEvent.triggeringLocation, context)
+        if (geofencingEvent.triggeringLocation.accuracy > integrationUseCase.getMinimumAccuracy()) {
```
balloob: I wonder if we should start with just having a hardcoded min accuracy and just launch with just this logic. That should already solve the issue that was reported.

bramkragten:
> I'll move current logic to only background updates and I'll add a one time high accuracy location request after a non accurate result.

How will that effect battery usage? As the reason we don't do high accuracy location requests now is for the battery I guess?

quthla:
> > I'll move current logic to only background updates and I'll add a one time high accuracy location request after a non accurate result.
> 
> How will that effect battery usage? As the reason we don't do high accuracy location requests now is for the battery I guess?

It is only used for geofencing intents so not much of an issue. And if the minimum accuracy is being kept user configurable, everybody can decide if he wants to use it or not.

app/src/main/java/io/homeassistant/companion/android/background/LocationBroadcastReceiver.kt
JBassett: If we want to do that I can close this PR and start a new one that has a hardcoded value for minimum accuracy, it'll be easier than reverting the UI, Data, and Domain changes.  I do think that people are going to want different values based on the issues and some vocal people on Discord.
Just let me know if that is what you would like.

Knoxie:
Having the ability to set is definitely nice. Not everyone will be using
devices that have the same accuracy. Homeassistant is all about
customization in my eyes. No sense in limiting people.

app/src/main/java/io/homeassistant/companion/android/background/LocationBroadcastReceiver.kt
balloob: Let's start by not customizing it. There will always be a minority vocal people that will shout that they want to be able to configure it, and in the end we have an app that only a tiny fraction can use because it's too complicated.

quthla: Are you sure that option is overcomplicating the app? It's a simple value in meters with a description everybody is able to understand.
What would that default value be that accounts for all devices, location providers etc?

So far, only a minority of people asked not to be able to customize it.

dshokouhi: Do the best of both worlds. Keep a good default like we are at 50m and if people want to change it they can. Most users won't need to change an app setting if the default is good enough. People will get confused if they have to change the setting but not if it works out the box. Those who need more accurate location should be able to change it as needed.

balloob: I would just keep it at 200. We literally had 1 or 2 reports that it was not working properly with geofencing and that's it? Seems a bit fast to jump to conclusions for an app with 30k installs.

JBassett:
Closing this PR in favor of @balloob suggestions. I'll reopen a new PR with just a static 200m minimum accuracy and a request for accurate location when geofencing events aren't accurate.

PR #181, opened 2019-12-20
Set minimum accuracy for location updates.

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/166
Fixes: https://github.com/home-assistant/home-assistant-android/issues/145

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/background/LocationBroadcastReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/background/LocationBroadcastReceiver.kt
@@ -12,6 +12,7 @@ import android.util.Log
 import com.google.android.gms.location.Geofence
 import com.google.android.gms.location.GeofencingEvent
 import com.google.android.gms.location.GeofencingRequest
+import com.google.android.gms.location.LocationCallback
 import com.google.android.gms.location.LocationRequest
 import com.google.android.gms.location.LocationResult
 import com.google.android.gms.location.LocationServices
@@ -24,6 +25,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 
 class LocationBroadcastReceiver : BroadcastReceiver() {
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/background/LocationBroadcastReceiver.kt
```diff
@@ -208,6 +221,25 @@ class LocationBroadcastReceiver : BroadcastReceiver() {
...
 
+    private fun requestSingleAccurateLocation(context: Context) {
+        val request = createLocationRequest()
+        request.priority = LocationRequest.PRIORITY_HIGH_ACCURACY
+        request.numUpdates = 2 // Request 2 updates to make sure GPS gets a good lock
+        LocationServices.getFusedLocationProviderClient(context)
+            .requestLocationUpdates(
+                request,
+                object : LocationCallback() {
+                    override fun onLocationResult(locationResult: LocationResult?) {
```
balloob: Will this be called twice?

JBassett: Yes, depending on how fast a GPS lock is acquired the first result is not always as accurate as we would want.  I figured letting it give us two updates should allow long enough for a decently accurate location to be sent.

balloob: So this means we also send 2 location updates to Home Assistant, the first one being potentially inaccurate ?

JBassett: Correct, I thought about checking accuracy before sending this one as well.  However if we get the 2 updates and neither make the accuracy requirement then we really won't be sending an exit zone location at all and figured this would be better.

balloob: I don't know if it's possible, but could we limit to to only send the last update ?

JBassett:
@balloob after you comment I went back and rewrote the logic to be as follows:
- Request 5 updates
- if we find an accurate location, send it and remove further updates
- if we are on our last retry send the location no matter what because otherwise a zone exit or entrance event might not get an update at all.

Issue #105, opened 2019-12-04
Dead end when entering invalid Home Assistant url during onboarding


balloob:
When you enter an invalid url during onboarding, the webview will show a "Cannot connect" page. Pressing back quits the app.

PR #182, opened 2019-12-20
Fix Dead end when entering invalid Home Assistant url during onboarding  #105

Tickaren:


homeassistant:
Hi @HIQOscarStenqvist, @Tickaren, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant-android#182).

Thanks!

balloob:
Can you add a screenshot ?

Tickaren:
@balloob

Tickaren:
Skärmavbild 2019-12-20 kl  10 47 37
@balloob

app/src/main/res/values/strings.xml
```diff
@@ -15,10 +15,12 @@ to your home internet.</string>
     <string name="checking_with_home_assistant">Checking with Home Assistant</string>
     <string name="skip">Skip</string>
     <string name="retry">Retry</string>
+    <string name="ok">Ok</string>
     <string name="attempting_registration">Registering application…</string>
     <string name="unable_to_register">Unable to Register Application</string>
     <string name="error_with_registration">Please check to ensure you have the mobile_app
 integration enabled on your home assistant instance.</string>
+    <string name="error_connection_failed">Unable to connect to Home Assistant, check your URL.</string>
```
balloob: Can we rename this to include onboarding. We wouldn't want to use the "check your URL" part when we can't connect after authentication is done. 

```suggestion
    <string name="error_onboarding_connection_failed">Unable to connect to Home Assistant, check your URL.</string>
```

homeassistant:
Hi @HIQOscarStenqvist, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant-android#182).

Thanks!

app/src/main/res/values/strings.xml
Tickaren: Fixed!

JBassett:
Part of me thinks we should do more than just an alert for this.  I don't think there is a need to display the webview in the background if it's just an error screen.  Perhaps replace the view with an error screen like here with a view flipper?
https://github.com/home-assistant/home-assistant-android/blob/master/app/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationFragment.kt

balloob:
I think ideally we navigate back to the url input screen and then show the alert.

bramkragten:
But only when you are onboarding right? Because reading the code it will now do it for every error in the webview? I don't want to go to onboarding when my HA instance is down because it is updating or something.

JBassett:
The code is specific for the authentication web view not the normal web view.  This should have no affect when an error occurs after you have onboarded.

Tickaren:
Now pops back first then shows alert!

Skärmavbild 2019-12-21 kl  14 53 22

PR #185, opened 2019-12-21
add support for custom root CA

tiagofreire-pt:


tiagofreire-pt:
my mistake. please, delete this.

PR #188, opened 2019-12-22
Make sure mobile_app is required.

JBassett:


app/src/main/java/io/homeassistant/companion/android/launch/LaunchPresenterImpl.kt
```diff
@@ -29,11 +29,12 @@ class LaunchPresenterImpl @Inject constructor(
 
     override fun onViewReady() {
         mainScope.launch {
-            if (authenticationUseCase.getSessionState() == SessionState.CONNECTED) {
+            val sessionValid = authenticationUseCase.getSessionState() == SessionState.CONNECTED
+            if (sessionValid && integrationUseCase.isRegistered()) {
                 resyncRegistration()
```
balloob: Off-topic: are we always resyncing the registration on startup?

JBassett: Currently yes, we can and should add logic that if nothing has changed not to make the http call.

PR #194, opened 2019-12-22
Allow user certs imported into android to be accepted.

JBassett:
This PR will bring all devices to the same base security level: https://developer.android.com/training/articles/security-config#base-config

Fixes: https://github.com/home-assistant/home-assistant-android/issues/184
Fixes: https://github.com/home-assistant/home-assistant-android/issues/171
Fixes: https://github.com/home-assistant/home-assistant-android/issues/48

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -18,6 +18,7 @@
         android:supportsRtl="true"
         android:theme="@style/Theme.HomeAssistant"
         android:usesCleartextTraffic="true"
+        android:networkSecurityConfig="@xml/network_security_config"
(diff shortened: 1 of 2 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -18,6 +18,7 @@
         android:supportsRtl="true"
         android:theme="@style/Theme.HomeAssistant"
         android:usesCleartextTraffic="true"
```
balloob: Do we still need this?

JBassett: No, it was defaulted to false in target API 28 but since we override it here in the network security config we should be ok because that works since API level 24.

PR #195, opened 2019-12-22
Support remoteUiUrl from the mobile_app integration.

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/92

### DIFF
--- a/data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
+++ b/data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
@@ -7,6 +7,7 @@ import io.homeassistant.companion.android.domain.integration.Entity
 import io.homeassistant.companion.android.domain.integration.IntegrationRepository
 import io.homeassistant.companion.android.domain.integration.UpdateLocation
 import io.homeassistant.companion.android.domain.integration.ZoneAttributes
+import java.net.URL
 import javax.inject.Inject
 import javax.inject.Named
 import okhttp3.HttpUrl
@@ -100,6 +101,11 @@ class IntegrationRepositoryImpl @Inject constructor(
 
     override suspend fun isRegistered(): Boolean {
         return localStorage.getString(PREF_WEBHOOK_ID) != null
+    }
+
+    override suspend fun getUiUrl(isInternal: Boolean): URL? {
+        return localStorage.getString(PREF_REMOTE_UI_URL)?.toHttpUrl()?.toUrl()
+            ?: authenticationRepository.getUrl()
     }
 
     override suspend fun updateLocation(updateLocation: UpdateLocation) {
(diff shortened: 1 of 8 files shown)


quthla:
What if the cloud subscription ended or the remote UI is unavailable?

JBassett:
Currently the mobile_app component won't tell us if the subscription is expired.  I don't think currently there is a way to determine if the from the apps point of view.  For now you would need to clear data and sign back in.

As far as the unavailable goes I'm working on that still.

quthla:
And the internal / external feature of the iOS app? It has 2 prefs for the URLs and one for the home wifi. If connected to home wifi, it'll use the internal url and if not, the external.

balloob:
We can add the internal/external feature later. Right now if a cloud user is adding their Home Assistant installation (and especially once we add discovery), it won't work when they leave their house.

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
```diff
@@ -102,6 +103,11 @@ class IntegrationRepositoryImpl @Inject constructor(
         return localStorage.getString(PREF_WEBHOOK_ID) != null
     }
 
+    override suspend fun getUiUrl(isInternal: Boolean): URL? {
+        return localStorage.getString(PREF_REMOTE_UI_URL)?.toHttpUrl()?.toUrl()
```
balloob: We can use `isInternal` here already, by making sure we only return `authenticationRepository.getUrl()` if it's true.

David0331:
O

rajput306:
thanks for this [information](http://www.kztrading.in/).

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
JBassett: Done.

JBassett:
After looking into the auto switching URLs on errors it felt outside the scope of this PR so I deferred to a later time.

quthla:
What I've got in mind is something like this:

Prefs: Home URL, Home Wi-Fi, Remote URL

Discovery: fills Home and Remote URL (if remote_ui_url available)

If only one URL filled: use that
If both URLs filled but no Home Wi-Fi selected: use Remote URL
If all 3 prefs filled: use Home URL when connected to Home Wi-Fi, else use Remote URL
If URL can't be reached on app start: show pref screen with error message so user can make necessary changes

This way, no auto switching remote url on errors is needed and it is ensured that the app never becomes unusable

Your opinions on this?

balloob:
Yeah agree. Just realized that we should also get auth tokens from remote url when not at home. It's not just the web UI.

JBassett:
Agreed, I noticed the auth issue too, I've already started working on extracting the URL handling so it can be reused in all places.

PR #209, opened 2019-12-30
Home Network Settings

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/30

Things left to do:
- [x] Better handling about what error is sending us to settings, currently to sensitive
- [x] Validate that SSID checking is working (I don't think it is yet...)

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationPresenterImpl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationPresenterImpl.kt
@@ -24,7 +24,7 @@ class AuthenticationPresenterImpl @Inject constructor(
 
     override fun onViewReady() {
         mainScope.launch {
-            view.loadUrl(authenticationUseCase.buildAuthenticationUrl(AUTH_CALLBACK).toString())
+            view.loadUrl(authenticationUseCase.buildAuthenticationUrl(false, AUTH_CALLBACK).toString())
         }
     }
 
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationPresenterImpl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationPresenterImpl.kt
@@ -24,7 +24,7 @@ class AuthenticationPresenterImpl @Inject constructor(
 
     override fun onViewReady() {
         mainScope.launch {
-            view.loadUrl(authenticationUseCase.buildAuthenticationUrl(false, AUTH_CALLBACK).toString())
+            view.loadUrl(authenticationUseCase.buildAuthenticationUrl(AUTH_CALLBACK).toString())
         }
     }
 
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
@@ -3,7 +3,7 @@ package io.homeassistant.companion.android.webview
 import android.net.Uri
 import android.util.Log
 import io.homeassistant.companion.android.domain.authentication.AuthenticationUseCase
-import io.homeassistant.companion.android.domain.integration.IntegrationUseCase
+import io.homeassistant.companion.android.domain.url.UrlUseCase
 import javax.inject.Inject
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -13,8 +13,8 @@ import kotlinx.coroutines.launch
 
 class WebViewPresenterImpl @Inject constructor(
     private val view: WebView,
-    private val authenticationUseCase: AuthenticationUseCase,
-    private val integrationUseCase: IntegrationUseCase
+    private val urlUseCase: UrlUseCase,
+    private val authenticationUseCase: AuthenticationUseCase
 ) : WebViewPresenter {
 
     companion object {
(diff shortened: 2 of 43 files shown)


andriej:
Nice to see this PR - one question from an observer: will app be able to ignore 'home'/'remote'?
I have my router setup so it resolves external domain to my local IP when connected with home-wifi.
I'd like to see app not break things and be able to set just 'use one URL always'.

Just my 5 cents.

JBassett:
I too have a router that can handle it.  The ability to use a single URL will always work.

andriej:
Just wanted to make sure.

quthla:
It probably needs [this](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback) to reload the webview if needed

app/src/main/java/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -25,8 +25,8 @@ class WebViewPresenterImpl @Inject constructor(
 
     override fun onViewReady() {
         mainScope.launch {
-            val url = integrationUseCase.getUiUrl(false)
-                ?: throw IllegalStateException("Unable to display the webview if we do not have url")
+            val ssid = view.getCurrentSsid()
+            val url = urlUseCase.getUrl("\"$ssid\"" == urlUseCase.getHomeWifiSsid())
```
balloob: Why the extra quotes ?

quthla: https://developer.android.com/reference/android/net/wifi/WifiInfo.html#getSSID()

data/src/main/java/io/homeassistant/companion/android/data/url/UrlRepositoryImpl.kt
```diff
@@ -0,0 +1,97 @@
...
+import okhttp3.HttpUrl
+import okhttp3.HttpUrl.Companion.toHttpUrl
+import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
+
+class UrlRepositoryImpl @Inject constructor(
+    @Named("url") private val localStorage: LocalStorage
+) : UrlRepository {
+
+    companion object {
+        private const val PREF_CLOUD_URL = "cloud_url"
```
balloob: What is `CLOUD_URL` ? There is a Remote UI cloud url and there is a cloudhook (cloud webhook) url.

balloob: I see now that it's for cloudhook, let's rename it

data/src/main/java/io/homeassistant/companion/android/data/url/UrlRepositoryImpl.kt
```diff
@@ -0,0 +1,97 @@
...
+        localStorage.putString(PREF_WEBHOOK_ID, webhookId)
+    }
+
+    override suspend fun getUrl(isInternal: Boolean): URL? {
+        val internal = localStorage.getString(PREF_LOCAL_URL)?.toHttpUrlOrNull()?.toUrl()
+        val external = localStorage.getString(PREF_REMOTE_URL)?.toHttpUrlOrNull()?.toUrl()
+        return if (isInternal) {
+            internal ?: external
+        } else {
+            external ?: internal
```
balloob: If we're external, would it really help in ever returning the internal url ?

JBassett: It really depends on peoples setups, if someone manually sets up the url during onboarding that gets persisted as the internal address.  However, without the wifi ssid set they are technically external.  So in that case they need the internal address returned.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
JBassett: But I have my logic reversed...

JBassett:
Settings screenshot:
Screenshot_20191230-082941

app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationPresenterImpl.kt
```diff
@@ -24,7 +24,7 @@ class AuthenticationPresenterImpl @Inject constructor(
 
     override fun onViewReady() {
         mainScope.launch {
-            view.loadUrl(authenticationUseCase.buildAuthenticationUrl(AUTH_CALLBACK).toString())
+            view.loadUrl(authenticationUseCase.buildAuthenticationUrl(false, AUTH_CALLBACK).toString())
```
JBassett: - [ ] //TODO: Fix me to use correct url based on WiFi Status.

balloob:
For the screenshot, let's call external url just Home Assistant url. Preferably there would be a new section with a toggle to "Enable different url when at home".

balloob:
Or instead of toggle maybe when no ssid typed in we disable local url field.

JBassett:
Invalid url:
Screenshot_20191231-141441

Disabled preference when no WiFi set:
Screenshot_20191231-131820

balloob:
🎉 

Let's test this for a couple of days and do a new release soon.

dgtal1:
Can you allow multiple SSIDs? I've got 2.4 and 5GHz network at home, both with different SSID. Would be great to see such feature. It may be comma delimited or a list box. Great work guys!

JBassett:
@dgtal1 there is already a request for that. https://github.com/home-assistant/home-assistant-android/issues/250

PR #212, opened 2020-01-01
Basic widget functionality added

KBerstene:
I have created a configurable home screen widget for calling services with a single tap without having to launch the app.  This is an incredibly useful function that I used with the Wink app before I switched to Home Assistant.  As far as I can tell, no Home Assistant android apps have this functionality yet.

This is a very rough widget, however.  Currently, it is configurable to call a single service on a single entity.  The widget is static and is just a picture of the app icon and a label.  Ideally, the widget would have a selectable icon that can respond when pressed.  The response could be as simple as a loading circle and then a checkmark when the service call has been completed like Wink's widgets, or it could be icons that indicate light status upon changes.  Additionally, the labels should be optional (that's something I plan on adding, I just wanted to get this basic widget started first).

Unfortunately, I've never written tests before so I don't even know where to start on those.

I've tested the functionality on a emulated Nexus 5X running Android 7, and my Samsung Galaxy S9 running Android 9.0 (and Nova Launcher).  On my phone there is a minor issue with the size of the icon, but that is something I am planning on working on in the next few days.  I'd prefer to sidestep the issue entirely by adding the responsive icons.


homeassistant:
Hi @KBerstene, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant-android#212).

Thanks!

app/build.gradle
```diff
@@ -114,3 +114,4 @@ dependencies {
 // This plugin must stay at the bottom
 // https://developers.google.com/android/guides/google-services-plugin
 apply plugin: 'com.google.gms.google-services'
+apply plugin: 'kotlin-android-extensions'
```
JBassett: I know others have tried to add this plugin in the past.  I would move it up to the other plugins if you are using it.

app/src/main/java/io/homeassistant/companion/android/widgets/WidgetPrefUtils.kt
```diff
@@ -0,0 +1,37 @@
...
+}
+
+internal fun loadStringPref(context: Context, appWidgetId: Int, key: String): String? {
+    return context.getSharedPreferences(PREFS_NAME, 0)
+        .getString(PREF_PREFIX_KEY + appWidgetId + key, null)
+}
+
+internal fun deleteStringPref(context: Context, appWidgetId: Int, key: String) {
+    saveStringPref(context, appWidgetId, key, null)
+}
```
JBassett: These should be moved into a usecase and repository

app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
```diff
@@ -0,0 +1,135 @@
...
+    private val mainScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
+
+    override fun onUpdate(
+        context: Context,
+        appWidgetManager: AppWidgetManager,
+        appWidgetIds: IntArray
+    ) {
+        // There may be multiple widgets active, so update all of them
+        for (appWidgetId in appWidgetIds) {
+            loadServiceCallData(context, appWidgetId)
```
JBassett: You load this but never use any of it in this method?

app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
```diff
@@ -0,0 +1,135 @@
...
+
+class ButtonWidget : AppWidgetProvider() {
+    companion object {
+        private const val TAG = "ButtonWidget"
+        private const val CALL_SERVICE = "CALL_SERVICE"
+    }
+
+    private lateinit var domain: String
+    private lateinit var service: String
+    private lateinit var serviceData: String
```
JBassett: These shouldn't be a the class level, they should be variables in the methods they are used.

app/src/main/res/values-v14/dimens.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <dimen name="widget_margin">0dp</dimen>
+
+</resources>
```
JBassett: Our minimum sdk is above this?  I don't think this gets used at all and can be removed.

app/src/main/res/layout/widget_button.xml
```diff
@@ -0,0 +1,27 @@
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
JBassett: Screenshots of widget and configuration screen please.

app/src/main/java/io/homeassistant/companion/android/widgets/WidgetPrefUtils.kt
KBerstene: I'm an amateur developer (hence why I didn't write any tests) and I'm not sure how to properly implement this style.  Do you have an article or paper you would recommend I read to make sure I do this correctly?  Even just a style name would be helpful - the best I can find is "Clean Architecture" when doing a cursory search but I'm not sure that's correct.

app/src/main/res/values-v14/dimens.xml
KBerstene: Right, of course that makes sense.  I will remove it in my next commit

app/build.gradle
KBerstene: Was there a particular reason it hasn't been added in the past?

I have moved it up with the other plugins in the latest commit.

app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
KBerstene: Right, this was originally a loading method I had implemented before finding the update ordering bug noted in the ButtonWidgetConfigureActivity.kt lines 40-42.  I have cleaned this up and reworked loading in the latest commit.

app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
KBerstene: Done, these are only in the specific method.

app/build.gradle
JBassett: I'm not sure... it wasn't me against it.

app/src/main/java/io/homeassistant/companion/android/widgets/WidgetPrefUtils.kt
JBassett: That's a good place to start, I would look at the classes you added the ```callService``` method too.  You would want something similar to that.

KBerstene: Okay, I'll work on getting that set up.  I wasn't incredibly confident in how I was modifying the classes for `callService`, but if they look fine to you then I should be able to copy them well enough for a repository/implementation/usecase setup.

app/src/main/res/layout/widget_button.xml
KBerstene: Widget on a home screen:
Widget on a home screen

Widget configuration screen:
Widget configuration screen

app/src/main/java/io/homeassistant/companion/android/widgets/WidgetPrefUtils.kt
KBerstene: This is complete in commit f877070

JBassett: Is this file still used?

app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
```diff
@@ -0,0 +1,210 @@
...
+                                PendingIntent.FLAG_UPDATE_CURRENT
+                            )
+                        )
+                        setTextViewText(
+                            R.id.widgetLabel,
+                            widgetStorage.loadLabel(appWidgetId)
+                        )
+                    }
+                    appWidgetManager.updateAppWidget(appWidgetId, views)
+                } catch (e: Exception) {
```
JBassett: What is this try catch trying to catch?

app/src/main/res/layout/widget_button.xml
```diff
@@ -0,0 +1,49 @@
...
+            />
+
+        <ImageButton
+            android:id="@+id/widgetImageButton"
+            android:layout_width="match_parent"
+            android:layout_height="match_parent"
+            android:layout_margin="0dp"
+            android:contentDescription="@string/widget_image_description"
+            android:background="@null"
+            android:src="@drawable/ic_lightbulb_outline_black_24dp"
```
JBassett: Not sure a lightbulb is generic enough for a widget that calls any service.  Maybe something that indicates a more general action?

app/src/main/res/layout/widget_button_configure.xml
```diff
@@ -0,0 +1,55 @@
...
+        android:layout_height="wrap_content"
+        android:hint="@string/widget_text_hint_service_service"
+        android:inputType="text" />
+
+    <EditText
+        android:id="@+id/widget_text_config_service_data"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:hint="@string/widget_text_hint_service_data"
+        android:inputType="text" />
```
JBassett: I think we should have labels for all these.

domain/src/main/java/io/homeassistant/companion/android/domain/widgets/WidgetUseCaseImpl.kt
```diff
@@ -0,0 +1,63 @@
+package io.homeassistant.companion.android.domain.widgets
+
+import javax.inject.Inject
+
+class WidgetUseCaseImpl @Inject constructor(
```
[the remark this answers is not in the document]
KBerstene: This is interesting--it's how I originally wrote it but I ended up moving the logic into the "use case" because the term made me think that since this is how I'm USING the repository, I should be putting it here.  I think part of the problem is I don't understand why I am essentially rewriting the same general header three times and the logic once instead of a general method once (plus the interface header) and the specific logic once (plus the interface header).

I don't have any problem making the change, I just want to clarify what is confusing to me about why the code is formatted in this manner.

Also, I just wanted to say thank you for being so patient with me, this is the easily the most complex project I've proposed code for, so this type of code structure is entirely new to me.

app/src/main/res/layout/widget_button.xml
KBerstene: Hrm, true.  I am thinking of my personal use case, which is why I chose a light bulb.  I would actually prefer to have a list of icons to choose from, eventually.  For now, I will look at changing it to a more generic "execute action" icon, and perhaps add a few options for specific icons (light bulb, switch, automation icon, etc)

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
```diff
@@ -126,6 +126,26 @@ class IntegrationRepositoryImpl @Inject constructor(
...
+
+        for (it in urlRepository.getApiUrls()) {
+            try {
+                wasSuccess =
+                    integrationService.callService(it.toHttpUrlOrNull()!!, IntegrationRequest("call_service", serviceCallRequest)).isSuccessful
+            } catch (e: Exception) {
+                // Ignore failure until we are out of URLS to try!
+            }
+            // if we had a successful call we can return
+                if (wasSuccess)
```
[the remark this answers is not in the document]
KBerstene: Oof, embarrassing.  Fixed in commit d2fd365

app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
```diff
@@ -0,0 +1,210 @@
...
+
+                // Set a timer to change it back after 1 second
+                Handler().postDelayed({
+                    views.setImageViewResource(
+                        R.id.widgetImageButtonBackground,
+                        R.drawable.ic_circle_white_24dp
+                    )
+                    appWidgetManager.updateAppWidget(appWidgetId, views)
+                }, 1000)
+            } catch (e: Exception) {
```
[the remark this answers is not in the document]
KBerstene: I have narrowed the focus of the try catch to the relevant line of code in commit 38d1d1b.  I have also added negative feedback if the call fails.

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
```diff
@@ -126,6 +126,26 @@ class IntegrationRepositoryImpl @Inject constructor(
...
 
+    override suspend fun callService(domain: String, service: String, serviceData: HashMap<String, String>) {
+        var wasSuccess = false
+
+        val serviceCallRequest = ServiceCallRequest(domain, service, serviceData)
+
+        for (it in urlRepository.getApiUrls()) {
+            try {
+                wasSuccess =
+                    integrationService.callService(it.toHttpUrlOrNull()!!, IntegrationRequest("call_service", serviceCallRequest)).isSuccessful
```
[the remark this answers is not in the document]
KBerstene: `toHttpUrl()` requires the `String` type, whereas `it` is a `URL`.  If you want, I can cast the `URL` to a `String` but I think that will just introduce more complexity.  Also, there are three other methods that use `toHttpUrlOrNull()!!` in this same file--I got this format from the others, so if we change this we should probably change it in those as well.

app/src/main/java/io/homeassistant/companion/android/widgets/WidgetPrefUtils.kt
KBerstene: It is not, since the WidgetRepository/UseCase have been set up.  I have removed it in commit 4f66f48.

app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
KBerstene: Nothing, apparently.  I think I copied it from my other `mainScope.launch {...}` block and didn't think closely enough about why the try catch was there.  Removed in 36cb729.

domain/src/main/java/io/homeassistant/companion/android/domain/widgets/WidgetUseCaseImpl.kt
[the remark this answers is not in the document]
KBerstene: Changed completed as requested in commit 8b2b9db

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
```diff
@@ -126,6 +126,29 @@ class IntegrationRepositoryImpl @Inject constructor(
...
+                        it.toHttpUrlOrNull()!!,
+                        IntegrationRequest("call_service", serviceCallRequest)
+                    ).isSuccessful
+            } catch (e: Exception) {
+                // Ignore failure until we are out of URLS to try!
+            }
+            // if we had a successful call we can return
+            if (wasSuccess)
+                return
+        }
```
marciogranzotto: You don't need to try catch the casting, just first map the URLs using a `mapNotNull` which maps and filters `null`s
```kotlin
urlRepository.getApiUrls().mapNotNull { it.toHttpUrlOrNull() }.forEach {
// ...
```

app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
```diff
@@ -0,0 +1,221 @@
...
+                serviceDataMap["entity_id"] = serviceData
+
+                try {
+                    integrationUseCase.callService(domain, service, serviceDataMap)
+
+                    // If service call does not throw an exception, send positive feedback
+                    feedback = R.drawable.ic_circle_yellow_24dp
+                } catch (e: Exception) {
+                    Log.e(TAG, "Could not send service call.", e)
+                }
```
marciogranzotto: One problem I had in other projects, is that if the Android device is on Power Save mode, all network requests from background fail.
To fix it, I needed to create a `ForegroundService` which then executes the network request on foreground.
You should probably test that

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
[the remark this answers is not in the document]
JBassett: Nope, forgot that method returned urls already.

app/src/main/res/layout/widget_button.xml
KBerstene: I added several options for this in commit 93d1570.  There's now a drop down menu for selecting an icon, and the default is the lightning bolt/flash/'action'/switch icon used by the Lovelace web interface.

Screenshot:
Screenshot_20200117-121930_Home Assistant

app/src/main/res/layout/widget_button_configure.xml
KBerstene: I have changed these to the TextInputEditText layout and changed the labels from the lowercase JSON-type format to a regular text label format.

Screenshot here:
Screenshot_20200117-121930_Home Assistant

domain/src/main/java/io/homeassistant/companion/android/domain/integration/IntegrationUseCase.kt
```diff
@@ -18,6 +18,8 @@ interface IntegrationUseCase {
 
     suspend fun updateLocation(updateLocation: UpdateLocation)
 
+    suspend fun callService(domain: String, service: String, serviceData: HashMap<String, String>)
```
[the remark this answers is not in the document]
KBerstene: I suppose there's no reason for it not to be.  Changed in commit ef757fa.

data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
KBerstene: While I'm not going to disagree that this might be more elegant solution, the formatting for this method was taken from `updateLocation(...)`, `updateRegistration(...)`, and `getZones()`.  If I change it here, it should probably be changed in those as well.

app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
KBerstene: Sorry, it may be unclear here due to the common use of the term "service" between Android and Home Assistant.  The method `callService(...)` is not actually calling an Android service, but is rather calling an available service within Home Assistant.  Specifically, it's using this API: https://developers.home-assistant.io/docs/en/app_integration_sending_data.html#call-a-service

If there is an issue with background data being suppressed within the IntegrationUseCase, I don't think it would be appropriate to attempt to fix that here, as that would be a deeper issue with a component to which I am adding but not significantly altering the core functionality.

KBerstene:
I will attempt to write some tests for the data and domain modules, but looking through them I'm barely comprehending how these work.  I'll do what I can, but I would expect the tests to need major modification and/or completely miss the mark.

domain/src/main/java/io/homeassistant/companion/android/domain/widgets/WidgetRepository.kt
```diff
@@ -0,0 +1,26 @@
...
+    suspend fun loadIcon(appWidgetId: Int): String?
+    suspend fun loadLabel(appWidgetId: Int): String?
+
+    suspend fun saveIcon(appWidgetId: Int, resName: String?)
+    suspend fun saveLabel(appWidgetId: Int, data: String?)
+
+    suspend fun deleteWidgetData(appWidgetId: Int)
+
+    suspend fun saveStringPref(key: String, data: String?)
+    suspend fun loadStringPref(key: String): String?
```
JBassett: Do these get called from anywhere besides WidgetRepositoryImpl.kt?  If they do we should setup another usecase for it rather than just exposing the local storage implementation.

app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
```diff
@@ -0,0 +1,250 @@
...
+                serviceDataMap["entity_id"] = serviceData
+
+                try {
+                    integrationUseCase.callService(domain, service, serviceDataMap)
+
+                    // If service call does not throw an exception, send positive feedback
+                    feedbackColor = R.drawable.ic_circle_green_24dp
+                    feedbackIcon = R.drawable.ic_check_black_24dp
+                } catch (e: Exception) {
+                    Log.e(TAG, "Could not send service call.", e)
```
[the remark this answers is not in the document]
KBerstene: If you look at line 149, the feedback color is actually initialized as a negative feedback color (red). It only changes to positive feedback (green) if it fires correctly.

That being said, I don't have a good way to check for whether or not the service actually did anything. I'll have to check to see if that sort of feedback is even contained within the response itself.

domain/src/main/java/io/homeassistant/companion/android/domain/widgets/WidgetRepository.kt
KBerstene: You mean inside the widget use case? Or a completely separate use case?

JBassett: I wasn't sure if this was called.  I just took a look and they don't need to be in the interface, they can be private functions in the impl.

app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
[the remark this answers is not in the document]
JBassett: Ah, totally missed that.

domain/src/main/java/io/homeassistant/companion/android/domain/widgets/WidgetRepository.kt
KBerstene: Okay, I'll change them to private in the WidgetRepositoryImpl when I get a chance later today.

JBassett: After that we'll get this merged!

PR #236, opened 2020-01-08
Open urls that are outside the application correctly.

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/75
Fixes: https://github.com/home-assistant/home-assistant-android/issues/117

After some testing I didn't see any calls to ```shouldOverrideUrlLoading``` from the application itself.  It appeared that all calls to this method are ones that we should let the OS handle not the webview.


quthla:
Have you tried if this works also for links with target _blank?

JBassett:
> Have you tried if this works also for links with target _blank?

Yes, opens in the default browser.

PR #249, opened 2020-01-10
Add timeout to UI on load.

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/71

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -5,6 +5,7 @@ import android.content.Intent
 import android.net.http.SslError
 import android.os.Build
 import android.os.Bundle
+import android.os.Handler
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -175,8 +178,13 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
                 fun externalBus(message: String) {
                     Log.d(TAG, "External bus $message")
                     webView.post {
-                        when {
-                            JSONObject(message).get("type") == "config/get" -> {
+                        val json = JSONObject(message)
+                        when (json.get("type")) {
+                            "connection-status" -> {
+                                hasConnected = json.getJSONObject("payload")
```
balloob: Let's call it `isConnected` as it can be changed in the future.

PR #251, opened 2020-01-11
Add specific error handling around revoked token.

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/79

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -277,11 +277,14 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
             .setTitle(R.string.error_connection_failed)
             .setMessage(if (isAuthenticationError) R.string.error_auth_revoked else R.string.webview_error)
             .setPositiveButton(R.string.ok) { _, _ ->
-                if (isAuthenticationError)
+                if (isAuthenticationError) {
+                    presenter.clearKnownUrls()
                     openOnBoarding()
-                else
+                } else {
                     startActivity(SettingsActivity.newInstance(this))
+                }
             }
+            .setOnDismissListener { isShowingError = false }
             .show()
     }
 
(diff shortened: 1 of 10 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -259,12 +268,19 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
         super.onDestroy()
     }
 
-    override fun showError() {
+    override fun showError(isAuthenticationError: Boolean) {
+        if (isShowingError)
+            return
+        isShowingError = true
```
balloob: Should this ever be set to `false` ?

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -259,12 +268,19 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
...
+
         AlertDialog.Builder(this)
             .setTitle(R.string.error_connection_failed)
-            .setMessage(R.string.webview_error)
+            .setMessage(if (isAuthenticationError) R.string.error_auth_revoked else R.string.webview_error)
             .setPositiveButton(R.string.ok) { _, _ ->
-                startActivity(SettingsActivity.newInstance(this))
+                if (isAuthenticationError)
```
balloob: We should also clear the existing auth info and location reporting settings.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
JBassett: This is so that when the webview fails to load it doesn't also show an alert.

balloob: Sorry, I wasn't clear. I meant with my comment if we should set it back to `false` once the error is done showing ?

PR #255, opened 2020-01-12
Initial work on automatic discovery of HA on local network.

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/6

TODO List:
- [x]  Selection UI
- [x] Error Handling

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/discovery/DiscoveryFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/discovery/DiscoveryFragment.kt
@@ -2,12 +2,13 @@ package io.homeassistant.companion.android.onboarding.discovery
 
 import android.net.nsd.NsdManager
 import android.os.Bundle
-import android.os.Handler
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.widget.ArrayAdapter
 import android.widget.Button
-import android.widget.ViewFlipper
+import android.widget.ListView
+import androidx.appcompat.widget.AppCompatTextView
 import androidx.core.content.ContextCompat.getSystemService
 import androidx.fragment.app.Fragment
 import io.homeassistant.companion.android.DaggerPresenterComponent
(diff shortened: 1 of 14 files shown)


JBassett:
Screenshot_20200112-204132

app/src/main/java/io/homeassistant/companion/android/onboarding/discovery/DiscoveryFragment.kt
```diff
@@ -1,50 +1,119 @@
...
+        }
     }
 
-    private fun scan() {
-        viewFlipper.displayedChild = LOADING_VIEW
-        Handler().postDelayed({
-            viewFlipper.displayedChild = TIMEOUT_VIEW
-        }, 5000)
+    override fun onScanError() {
+        Toast.makeText(context, R.string.failed_scan, Toast.LENGTH_LONG).show()
```
balloob: When a scan error occurs, we should tell the user, and then show the "manually enter address" view.

app/src/main/java/io/homeassistant/companion/android/onboarding/discovery/HomeAssistantSearcher.kt
```diff
@@ -0,0 +1,86 @@
...
+    // Called as soon as service discovery begins.
+    override fun onDiscoveryStarted(regType: String) {
+        Log.d(TAG, "Service discovery started")
+    }
+
+    override fun onServiceFound(foundService: NsdServiceInfo) {
+        // A service was found! Do something with it.
+        Log.d(TAG, "Service discovery success: $foundService")
+
+        if (foundService.serviceType.startsWith(SERVICE_TYPE)) {
```
balloob: Consider using a guard clause for readability.

```kotlin
if (!foundService.serviceType.startsWith(SERVICE_TYPE)) { return; }
```

However, do we even need this check? We pass `SERVICE_TYPE` to `nsdManager.discoverServices`

app/src/main/java/io/homeassistant/companion/android/onboarding/discovery/DiscoveryFragment.kt
JBassett: Updated so that if the scan failed and we haven't found any instances yet to go to the manual entry, otherwise we should have an instance in the list to select.

PR #268, opened 2020-01-17
Feature/status bar color

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/26


dshokouhi:
Does this grab that color one time or will it update if the theme keeps changing?

JBassett:
It grabs it on app start.  Once the enhancement that was requested to the frontend is complete we should be able to do it when the theme is updated.

dshokouhi:
By app start do you mean initial login or we can force close the app to get it to update?

JBassett:
Force close and reopen will get the latest theme.

PR #270, opened 2020-01-18
Fix get theme because of missing fields in DTO.

JBassett:
Fixes crash introduced with theming application.


data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt
```diff
@@ -167,6 +167,7 @@ class IntegrationRepositoryImpl @Inject constructor(
             try {
                 response = integrationService.getConfig(it.toHttpUrlOrNull()!!, getConfigRequest)
             } catch (e: Exception) {
+                e.printStackTrace()
```
balloob: Is this for debugging or on purpose ?

JBassett: Oops, yep.

PR #273, opened 2020-01-18
Request we be exempt from Doze.

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/119
Fixes: https://github.com/home-assistant/home-assistant-android/issues/192
Fixes: https://github.com/home-assistant/home-assistant-android/issues/192
Fixes: https://github.com/home-assistant/home-assistant-android/issues/227

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationFragment.kt
@@ -30,6 +30,8 @@ class MobileAppIntegrationFragment : Fragment(), MobileAppIntegrationView {
         private const val ERROR_VIEW = 1
         private const val SETTINGS_VIEW = 2
 
+        private const val BACKGROUND_REQUEST = 99
+
         fun newInstance(): MobileAppIntegrationFragment {
             return MobileAppIntegrationFragment()
         }
@@ -79,6 +81,7 @@ class MobileAppIntegrationFragment : Fragment(), MobileAppIntegrationView {
                     presenter.onToggleZoneTracking(it.isSelected)
                 }
                 isEnabled = hasLocationPermission
+                isChecked = hasLocationPermission
             }
             zoneTrackingSummary = findViewById(R.id.location_zone_summary)
             zoneTrackingSummary.isEnabled = hasLocationPermission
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationFragment.kt
```diff
@@ -87,8 +96,32 @@ class MobileAppIntegrationFragment : Fragment(), MobileAppIntegrationView {
...
 
-        (activity as MobileAppIntegrationListener).onIntegrationRegistrationComplete()
+    @SuppressLint("BatteryLife")
+    private fun requestBackgroundAccess() {
+        val packageName = activity?.packageName ?: ""
+        val intent: Intent
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M &&
+            context?.getSystemService(PowerManager::class.java)?.isIgnoringBatteryOptimizations(packageName) == false
+        ) {
+            intent = Intent(
```
balloob: Can we inform the user why we're requesting this ? I wonder if we need to create some checklist page with buttons the user can press to be informed about it:

image

zhulik:
Hello there, thank you for your awesome work! When it's going to be released? Missing tracking is a really annoying issue that breaks all the positive experience of using the app. Feel free to ping me if you need any help with testing

JBassett:
Screenshot_20200122-094530

Screenshot_20200122-094536

Screenshot_20200122-094540

Screenshot_20200122-094549

JBassett:
@balloob just as a heads up about adding the application to the whitelist.  I believe we full under the task automation app category, but if the power to be in the play store think otherwise we might get pulled from the play store and have to fight it.

https://developer.android.com/training/monitoring-device-state/doze-standby#whitelisting-cases

dshokouhi:
@JBassett just reading the description it sounds like we most definitely fall under Task Automation.  Location actions are all automated here, user has 0 control over when we send the data :)

We should make sure the description in the play store reflects upon the automated location reporting too, that might help seal our case?  Because right now we are very vague about what we offer asides from connecting to HA.  We should probably start listing the features we have here so they know what our intentions are?

balloob:
I'll update our description on the next release. 

Based on what permissions the user has granted, can we automatically enable some toggles?

 - Can use location -> enable zone
 - Doze -> enable periodically report

JBassett:
@balloob done

PR #275, opened 2020-01-18
Reload UI url more proactively

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/258

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
@@ -35,6 +35,12 @@ class WebViewPresenterImpl @Inject constructor(
             val oldUrl = url
             url = urlUseCase.getUrl()
 
+            /*
+            We only want to cause the UI to reload if the URL that we need to load has changed.  An
+            example of this would be opening the app on wifi with a local url then loosing wifi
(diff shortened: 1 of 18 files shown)


JBassett:
Removed commits that were part of another PR with force push.  Made me feel dirty...

app/src/main/java/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -27,17 +28,22 @@ class WebViewPresenterImpl @Inject constructor(
...
-                    .buildUpon()
-                    .appendQueryParameter("external_auth", "1")
-                    .build()
-                    .toString()
-            )
+            if (oldUrl != url) {
+                view.loadUrl(
+                    Uri.parse(url.toString())
+                        .buildUpon()
+                        .appendQueryParameter("external_auth", "1")
```
balloob: Can we add a comment to this line that this is to support versions of Home Assistant before 103 released December 2019. That way we can clean it up in the future :)

JBassett:
So as I looked again at the code I realized that if we needed to re-auth on the reload we would fail because the external auth would still be using the old url.  I have removed the need to rebuild all of the dagger components on URL change by introducing an interceptor that will replace the correct host and scheme when making any retrofit calls.

Issue #250, opened 2020-01-10
Home Network WiFi SSID list?


scottismyname:
Home Assistant Android version: 1.3.1 (75)

Android version:10

Phone model: Pixel 2XL

Home Assistant version: 0.103.5

My house has several SSID's that my phone may connect to.  Maybe allow us to specify a list of SSID's that we can consider local?

xaviml:
I have the same problem. I saw in the last update that now we can type the SSID name and I think this is fantastic. However, I have 1 router and 2 extenders in my house, all with different SSID, so I would love to have a list of SSIDs, the same as the iOS app has.

Thank you for the awesome work!

CodeFinder2:
I have the same problem.

BCNelson:
The solution might be to allow the ssid to be a regex that would make is extremely expandable

xaviml:
> The solution might be to allow the ssid to be a regex that would make is extremely expandable

Although it could be an expandable solution, I don't think is very user-friendly. Maybe we could change the text to: "Home Network WiFi SSID contains" and any connected WiFi that contains that text, it will then connect to the internal URL. Just an opinion.

JBassett:
I agree it might be easier for some, but if we are trying to get the largest audience they won't even know what a regular expression is.

dgtal1:
+1 to that. It's wonderfull that the app started supporting local urls via SSIDs. If you add a list of SSIDs instead of just 1 - this will be really fantastic and will solve my problem of 2 home SSIDs for 2.4 and 5GHz network setup.

CodeFinder2:
I agree on simply providing a list. Regular expressions or hacks like "contains" are not a good idea and will confuse many users.

Moreover, using a list would not require any changes in current configs.

SergeyAnokhin:
+1 for list

scop:
Comma separated list would be another interim option, if making it a proper list is too much work.

Delimited SSID lists have a problem that a valid SSID can contain the delimiter. I suppose a comma is rare enough though.

CodeFinder2:
> Comma separated list would be another interim option, if making it a proper list is too much work.
> 
> Delimited SSID lists have a problem that a valid SSID can contain the delimiter. I suppose a comma is rare enough though.

Also thought about that problem. To make it even worse, SSIDs can basically contain any data (conceptually, even unprintable chars but I guess we can surely ignore this).

[...]

PR #278, opened 2020-01-19
Replacing equality for `contains` to Home Network WiFi SSID (#250)

xaviml:
This implementation is a proposal for #250. The desired implementation for the issue was to have a list to edit like the iOS app. However, the issue is that some places have more than one access point with a different SSID name and most of the cases the names share a common name. This is why I suggest this implementation to check if the connected wifi contains the text inputted from the app configuration, instead of checking equality.

I changed from `"\"${getHomeWifiSsid()}\""`  to `getHomeWifiSsid().toString()` since the first one did not work on my phone.

This was tested on my Samsung S9. I added the shared name of my APs on the app and it connected to the internal URL for all my APs.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -34,7 +34,7 @@ integration enabled on your home assistant instance.</string>
     <string name="webview_error">There was an error loading Home Assistant, please review the connection settings and try again.</string>
     <string name="pref_connection_title">Connection Information</string>
     <string name="pref_connection_url">Home Assistant URL</string>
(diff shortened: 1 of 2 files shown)


homeassistant:
Hi @xaviml, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant-android#278).

Thanks!

balloob:
We should not do this, since it's not possible to have a generic name for their wifi endpoint without getting false positives being conntected to other endpoints.

If we want this feature, someone should port it to support a list. We should not merge a hack like this.

PR #291, opened 2020-01-23
Fixup the styling on the Widget configuration.

JBassett:
Screenshot_20200122-231946


balloob:
Icon is not fully aligned on the left side with the text inputs

JBassett:
@balloob I feel like we would be better off mimicking the Developer Tools -> Services UI.  Basically swap the text fields for drop-downs.  I think it would help with the confusion a couple people have already had with setting widgets up.  I'll use this PR to expand on that tonight hopefully.

JBassett:
Screenshot_20200124-001359
Screenshot_20200124-001411
Screenshot_20200124-001648

bramkragten:
Could we use device automation actions for this? And have the service call as an advanced option?

JBassett:
@bramkragten I'm not sure I understand the question.

bramkragten:
Device automations are user-friendly actions (and triggers and conditions) a device can tell us it supports. So you select a device and it gives a list of the action it supports.

JBassett:
@bramkragten I would expect that to be another enhancement request, not part of this one.  Currently I don't think the api the front end is using to generate that is exposed externally.  That means we would require HA changes on top of app changes.

balloob:
This looks amazing.

bramkragten:
@JBassett wasn't meant for this PR, maybe something for the future 😃

KBerstene:
Dang, you were super on top of this.  I have been trying to pull up similar improvements to the widgets over a couple of days and my work looks pretty close to yours, though not exactly the same.  Any thoughts on changing the dropdowns to auto-complete text fields instead?

Also, I do have some improvements to this, specifically creating dynamic boxes based on what service data is applicable based on the service, rather than just blindly sending the entity id alone.  Any objection to me trying to expand this with that in mind?

JBassett:
@KBerstene I was just trying to get something out quickly.  I originally wanted auto completes but hadn't used them such a long time I was able to get the dropdowns done quicker.  By all means please keep iterating on it!

KBerstene:
I'm really happy to see that we both had the same idea about fetching the services and states.  That was my original idea but I didn't want to make too many changes on my initial pull request (or do all the work of figuring out how to do it just to get rejected).  I'll deal with the autocomplete text boxes and see if I can get them to work as semi-dropdowns too (best of both worlds, if I can?)

drthanwho:
Gotta say I'm loving this. One other issues I noticed in terms of styling when the widget icon is added to the home screen the chosen name doesn't fit properly. About 4 letters only show

e.g. Label: Garden Lights and on home screen you see "Gard...."

KBerstene:
@drthanwho Yeah, the amount of text is rather limited, I agree. I will look into seeing if there's a better way to extend the text. In the meantime, I would not be surprised if "Garden" by itself would display fully.

drthanwho:
@KBerstene I did try that too and still got Gard... however "Garage" which is still the same number of letters shows up fine

balloob:
No discussion on old PRs.

PR #298, opened 2020-01-23
Fix Fullscreen always on

neopilou:
Forgot to test if full screen is set or not in setOnSystemuiVisiblityChangeListener (line 231 in WebViewActivity.kt). This listener allows to keep the full screen even after using the virtual keyboard. Without testing the isFullscreen () condition, it always puts the application in full screen, even if the option is not activated.


barrymossel:
Could you also make an option to only hide the on-screen buttons and not the notification bar? The latter wil leave just a black bar on a Galaxy S10 (due to the camera punch hole I guess).

neopilou:
@barrymossel 

Go to Display-->Full screen apps

You'll find all the apps set to "Auto" by default. Find the Home Assistant app and force choose the "Full screen" option.

barrymossel:
Thanks, but that won't work as the punch hole will be in front of the options icon. I rather have the notification bar in place, just without the on screen buttons.

neopilou:
Close and open PR [302](https://github.com/home-assistant/home-assistant-android/pull/302)

PR #302, opened 2020-01-24
Add FullScreen/Hide Navigation Bar

neopilou:
Forgot to test if fullscreen is set or not in setOnSystemuiVisiblityChangeListener (line 231 in WebViewActivity.kt). This listener allows to keep the full screen even after using the virtual keyboard. Without testing the isFullscreen () condition, it always puts the application in full screen, even if the option is not activated.

And add new setting for just hide navigation bar :

Screenshot_20200124-121520_Home Assistant

And as requested all in one commit (almost... one missed indentation in a file... )

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
@@ -47,7 +47,7 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
         val onChangeSwitchFullScreen = Preference.OnPreferenceChangeListener { _, newValue ->
             if (newValue == true)
                 findPreference<SwitchPreference>("hide_nav")?.let {
-                 it.isChecked = false
+                    it.isChecked = false
                 }
             true
         }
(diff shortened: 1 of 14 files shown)


app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -54,6 +70,12 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
             it.summary = "${BuildConfig.VERSION_NAME} (${BuildConfig.VERSION_CODE})"
         }
 
+        findPreference<SwitchPreference>("fullscreen")?.onPreferenceChangeListener =
+            onChangeSwitchFullScreen
+
+        findPreference<SwitchPreference>("hide_nav")?.onPreferenceChangeListener =
```
balloob: Let's not add this option. That's feature creep. You go full screen or not.

balloob: We need to be cautious with the number of options that we show. If it's too many, none will be used. We need to aim for what is right for 80% of the users.

neopilou:
New PR ([304](https://github.com/home-assistant/home-assistant-android/pull/304)) created with fullscreen option only

balloob:
Thanks. Will close this one.

PR #304, opened 2020-01-25
Add Fullscreen option

neopilou:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/230
Add Fullscreen option in app settings :

Screenshot_20200125-113348_Home Assistant

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsPresenterImpl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsPresenterImpl.kt
@@ -29,6 +29,7 @@ class SettingsPresenterImpl @Inject constructor(
             return@runBlocking when (key) {
                 "location_zone" -> integrationUseCase.isZoneTrackingEnabled()
                 "location_background" -> integrationUseCase.isBackgroundTrackingEnabled()
+                "fullscreen" -> integrationUseCase.isFullScreenEnabled()
                 else -> throw Exception()
             }
         }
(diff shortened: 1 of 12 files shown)


JBassett:
I'll validate on a couple devices tonight just to be sure then merge it.

app/src/main/java/io/homeassistant/companion/android/settings/SettingsPresenterImpl.kt
```diff
@@ -39,6 +40,7 @@ class SettingsPresenterImpl @Inject constructor(
             when (key) {
                 "location_zone" -> integrationUseCase.setZoneTrackingEnabled(value)
                 "location_background" -> integrationUseCase.setBackgroundTrackingEnabled(value)
+                "fullscreen" -> integrationUseCase.setFullScreenEnabled(value)
                 else -> throw Exception()
             }
             settingsView.onLocationSettingChanged()
```
JBassett: This is being called when you toggle full screen and it is causing a prompt for location settings.  The `settingsView.onLocationSettingChanged()` should only be called when a location settings has changed, not whenever a boolean setting changed.

app/src/main/res/xml/preferences.xml
```diff
@@ -37,6 +37,13 @@
             android:title="@string/pref_location_background_title"
             android:summary="@string/pref_location_background_summary"/>
     </PreferenceCategory>
+    <PreferenceCategory
+        android:title="@string/other_settings">
+        <SwitchPreference
```
JBassett: I saw the icon added in source but you aren't referencing it here?

JBassett:
The code is functional, just need to fix my comments before we can merge.

app/src/main/java/io/homeassistant/companion/android/settings/SettingsPresenterImpl.kt
neopilou: Add condition for settingsView.onLocationSettingChanged() call

app/src/main/res/xml/preferences.xml
neopilou: Fullscreen icon reference added

PR #326, opened 2020-01-30
Fix Fullscreen for notch

neopilou:
Test if notch or not, if notch just hide nav bar

Test on virtual device : 

2020-01-30_10h02_29

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -248,7 +248,7 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
     }
 
     private fun hideSystemUI() {
-        if (Build.VERSION.SDK_INT >= 28 && isCutout())
+        if (isCutout())
             window.decorView.systemUiVisibility = (View.SYSTEM_UI_FLAG_HIDE_NAVIGATION
                     or View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY)
         else
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -342,7 +342,6 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
         }
     }
 
-    @TargetApi(Build.VERSION_CODES.P)
     private fun isCutout(): Boolean {
         var cutout = false
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && window.decorView.rootWindowInsets.displayCutout != null)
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -247,12 +248,16 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
     }
 
     private fun hideSystemUI() {
-        window.decorView.systemUiVisibility = (View.SYSTEM_UI_FLAG_LAYOUT_STABLE
-                or View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION
-                or View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN
-                or View.SYSTEM_UI_FLAG_HIDE_NAVIGATION
-                or View.SYSTEM_UI_FLAG_FULLSCREEN
-                or View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY)
+        if (Build.VERSION.SDK_INT >= 28 && isCutout())
```
JBassett: Can we use the constant for the SDK version (`Build.VERSION_CODES.P`)?
Also I think it would be better if you move the SDK check inside the `isCutout()` check.

neopilou: I just deleted the SDK check because unnecessary

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -336,4 +341,12 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
             super.onRequestPermissionsResult(requestCode, permissions, grantResults)
         }
     }
+
+    @TargetApi(28)
```
JBassett: This isn't going to keep lower devices from calling this.  It only fixes the linting error.  You should remove this and add the logic that you add above in here.

https://developer.android.com/reference/android/support/annotation/RequiresApi.html

```
private fun isCutout(): Boolean {
        var cutout = false
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && window.decorView.rootWindowInsets.displayCutout != null)
            cutout = true
        return cutout
    }
```

neopilou: en effet, je l'ai ajouté dans isCutout()

JBassett: I'm not sure if you need the `@TargetApi` anymore.  Does it still complain with it removed?  If it does you should still use the constant instead of the number `@TargetApi(Build.VERSION_CODES.P)`

neopilou: No request, i delete it

barrymossel:
Works perfectly for me! Thanks!

Issue #176, opened 2019-12-19
Implement notification with 'request_location_update' functionality


SamJongenelen:
**Home Assistant Android version:**
1.2.0 (42)
**Android version:**
10
**Phone model:**
pixel2

**Home Assistant version:**
0.103.0

**Description of problem:**

~~when sending _request_location_update_ message body to the android app, I get the notification delivered with a message body _request_location_update_. 
Is this the expected behavior ? How can i confirm my location is being asked? Or is this not yet supported on the Android version of the app?~~
not yet supported functionality. This ticket is became the enhancement request

**Traceback (if applicable):**

```

```

**Additional information:**
I am using a HA API call (using NodeRed). The following payload is being sent:

```
{"domain":"notify","service":"mobile_app_pixel_2","data":{"message":"request_location_update"}}
```

balloob:
This is not supported. Is this a feature that is supported on iOS?

SamJongenelen:
Im trying to implement _https://companion.home-assistant.io/en/notifications/location_. 

The feature would be useful for me when trying to locate someone _n_ min after wifi disconnect

dshokouhi:
This could be interesting, does the notification just stay on the device until the user dismisses it?  I wonder what the iOS behavior is like here.

SeanPM5:
The way it works in the iOS app, the notification isn't supposed to show up on the lock screen at all or require any user interaction. It's supposed to be a completely invisible method of updating your location for use in automations and scripts etc.

For certain automations you want to ensure the location data is completely fresh and that's where `request_location_update` is useful. I use it to track precisely where I parked my car for example ([screenshot](https://pbs.twimg.com/media/D-HJ2X9XkAEh2u7?format=jpg&name=4096x4096) from a few months ago). 

I'm not entirely sure how it all works behind the scenes but I believe that @robbiet480  suppresses the notification somehow on the Firebase side of things, so maybe that has to be done for the Android app too?

robbiet480:
This functionality needs to be implemented in the Android app before it will work.

Mic92:
Some special handling could be added here: https://github.com/home-assistant/home-assistant-android/blob/dfb03e32e088b28304dd2bdaa728420b9acf36d4/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt#L52

Mic92:
I also implemented something: https://github.com/home-assistant/home-assistant-android/pull/333

[...]

PR #332, opened 2020-02-01
Actionable Notifications

JBassett:
Current work in progress.  Basic functionality working.

Screenshot_20200201-234943

Example service call (supports up to 3 actions):
```
- service: notify.mobile_app_pixel_2
  data:
    message: 'Something really cool happened'
    title: 'Action Time'
    data:
      data:
        action_1_key: 'EVENT_KEY'
        action_1_title: 'Button Text'
```

Example event in HA:
```
{
    "event_type": "mobile_app_notification_action",
    "data": {
        "action": "KEY_ONE",
        "action_2_title": "Action 2",
        "action_3_title": "Action 3",
        "action_1_title": "Action 1",
        "action_1_key": "KEY_ONE",
        "action_2_key": "KEY_TWO",
        "action_3_key": "KEY_THREE"
    },
    "origin": "REMOTE",
    "time_fired": "2020-02-02T04:45:05.550251+00:00",
    "context": {
        "id": "abc123",
        "parent_id": null,
        "user_id": "123abc"
    }
}
```

Example Consumption in Automation:
```
- id: '1523975760898'
  alias: '[Notification - Action] Close Garage'
  initial_state: 'on'
  trigger:
    - platform: event
      event_type: mobile_app_notification_action
      event_data:
        action: KEY_ONE
  condition: []
  action:
    - service: cover.close_cover
      data:
        entity_id: cover.garage_door_opener

```

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
@@ -31,7 +31,10 @@ class NotificationActionReceiver : BroadcastReceiver() {
 
         intent?.getParcelableExtra<NotificationAction>(EXTRA_NOTIFICATION_ACTION)?.let {
             ioScope.launch {
-                integrationUseCase.fireEvent(it.key, it.data)
+                integrationUseCase.fireEvent(
+                    "mobile_app_notification_action",
+                    it.data.plus(Pair("action", it.key))
+                )
             }
         }
     }
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
@@ -5,17 +5,17 @@ import android.content.Context
 import android.content.Intent
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
 import io.homeassistant.companion.android.domain.integration.IntegrationUseCase
+import javax.inject.Inject
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
-import javax.inject.Inject
 
 class NotificationActionReceiver : BroadcastReceiver() {
 
     companion object {
         const val FIRE_EVENT = "FIRE_EVENT"
-        const val EXTRA_ACTION_KEY = "EXTRA_ACTION_KEY"
+        const val EXTRA_NOTIFICATION_ACTION = "EXTRA_ACTION_KEY"
     }
 
     private val ioScope: CoroutineScope = CoroutineScope(Dispatchers.IO + Job())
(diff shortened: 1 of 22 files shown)


SamJongenelen:
Solves #176

dshokouhi:
This is great! Would it be possible to update the description to include the automation trigger example too?  I tried looking on the iOS side but seems for android it will be different.  Thank you for including examples for the rest of it :)

https://companion.home-assistant.io/en/notifications/actionable#building-automations-for-notification-actions

JBassett:
@dshokouhi done.  If someone wanted to update the actual documentation for android I wouldn't be upset... ;-)
I looked into matching iOS for the actions but they were not compatible.

app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    override fun onReceive(context: Context?, intent: Intent?) {
+        DaggerServiceComponent.builder()
+            .appComponent((context!!.applicationContext as GraphComponentAccessor).appComponent)
+            .build()
+            .inject(this)
+
+        intent?.getParcelableExtra<NotificationAction>(EXTRA_NOTIFICATION_ACTION)?.let {
+            ioScope.launch {
+                integrationUseCase.fireEvent(it.key, it.data)
```
balloob: We should not allow the user to pick the event name to fire. Instead, we should pick an event name like "mobile_app_action" and pass both the key and data as event data.

balloob: Actually, why do we include `it.data` ? We should drop that.

JBassett: We are basically ecoing back the data that the user passed in the notification.  You can put any string->string mapping in there.  We could limit it to a standard event name and only pass the key if you want.  I was just trying to leave it more open, I think the iOS app echos all data back too.

balloob: ok

JBassett: Which option would you like?

JBassett: I updated to use constant event type and added an "action" element to the event data that is the selected option.

balloob: I'm fine with echo'ing it back.

SamJongenelen:
Thanks guys, awesome

bransjr:
Hey, I'm using your example yaml verbatim and get a notification that it is "Action time" and that "something really cool happened" but there are not buttons. any ideas what I'm doing wrong? Might it be a permission with the app?

Chumba:
@bransjr I'm having the same result, did you ever get it sorted out?

dshokouhi:
@bransjr  and @Chumba lets not discuss in a merged PR but please make sure to check the companion docs for the updated example: https://companion.home-assistant.io/docs/notifications/actionable-notifications#building-automations-for-notification-actions

Scroll down for the android example

PR #333, opened 2020-02-01
implement request_location_update via push notifications

Mic92:
I was not able to test this yet, since it seems that I cannot send pull notifications from a development app? https://github.com/home-assistant/home-assistant-android#setup-development-environment

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -13,6 +13,7 @@ import androidx.core.content.ContextCompat
 import com.google.firebase.messaging.FirebaseMessagingService
 import com.google.firebase.messaging.RemoteMessage
 import io.homeassistant.companion.android.R
+import io.homeassistant.companion.android.background.LocationBroadcastReceiver
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
 import io.homeassistant.companion.android.domain.integration.IntegrationUseCase
 import io.homeassistant.companion.android.webview.WebViewActivity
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -50,11 +51,23 @@ class MessagingService : FirebaseMessagingService() {
...
+            } else {
+                Log.d(TAG, "Message Notification: ${it.title} -> ${it.body}")
+                sendNotification(it.title, it.body!!)
+            }
         }
     }
 
+    private fun requestLocationUpdate() {
+        val intent = Intent(this, LocationBroadcastReceiver::class.java)
+        intent.action = LocationBroadcastReceiver.ACTION_REQUEST_LOCATION_UPDATES
```
JBassett: This intent action doesn't request a single high accuracy location update.  If you want that you will need a new action that calls `requestSingleAccurateLocation` and call that from here.

Mic92:
Fixed

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -50,11 +51,23 @@ class MessagingService : FirebaseMessagingService() {
         }
 
         remoteMessage.notification?.let {
-            Log.d(TAG, "Message Notification: ${it.title} -> ${it.body}")
-            sendNotification(it.title, it.body!!)
+            if (it.title == "request_location_update") {
```
JBassett: I just checked what the iOS app is doing, this should be message, not title.  That way we try and stay inline.

Mic92:
All changes implemented.

mdrichardson:
@Mic92 Nice work! Looks like we can do like iOS and just send `message: "request_location_update"` for this?

JBassett:
After looking deeper, this is going to be affected by the same thing as https://github.com/home-assistant/home-assistant-android/issues/337.  I'll be making changes to fix both.

Mic92:
@JBassett did you ever do this fix?

JBassett:
Yes, this is working as expected.

PR #336, opened 2020-02-03
Sensor Support

JBassett:
This is super rough sensor support.  Battery level and charging "working".
https://developers.home-assistant.io/docs/en/app_integration_sensors.html

fixes: https://github.com/home-assistant/home-assistant-android/issues/122
fixes: https://github.com/home-assistant/home-assistant-android/issues/128

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt
@@ -5,8 +5,6 @@ import android.content.Intent
 import android.os.Bundle
 import android.view.MenuInflater
 import androidx.appcompat.app.AppCompatActivity
-import androidx.work.PeriodicWorkRequestBuilder
-import androidx.work.WorkManager
 import com.lokalise.sdk.LokaliseContextWrapper
 import com.lokalise.sdk.menu_inflater.LokaliseMenuInflater
 import io.homeassistant.companion.android.DaggerPresenterComponent
@@ -16,7 +14,6 @@ import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
 import io.homeassistant.companion.android.onboarding.OnboardingActivity
 import io.homeassistant.companion.android.sensors.SensorWorker
 import io.homeassistant.companion.android.webview.WebViewActivity
-import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 
 class LaunchActivity : AppCompatActivity(), LaunchView {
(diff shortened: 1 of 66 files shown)


app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt
```diff
@@ -62,4 +68,12 @@ class LaunchActivity : AppCompatActivity(), LaunchView {
     override fun getMenuInflater(): MenuInflater {
         return LokaliseMenuInflater(this)
     }
+
+    private fun startSensorWorker() {
+        val sensorWorker =
+            PeriodicWorkRequestBuilder<SensorWorker>(15, TimeUnit.MINUTES)
```
balloob: Is this something we want to do periodically or only when:
 - we send a location event
 - we haven't send a location event in 30 min
 - we change charging type (non-charging -> charging)

JBassett: So in this initial implementation I was going for just every 15 ish minutes update all sensors.  I plan on listening for change events wherever possible to trigger an individual sensor to update.  I didn't want to link this functionality to the background location updates initially but there is nothing saying we can't.

dshokouhi:
With this change will we need to redo the integration or restart HA to get the sensors to show up or will they register and appear after enabling?  Also if you could share a screenshot of the new settings that would be great :)

JBassett:
Screenshot_20200208-155536

Screenshot_20200208-154334

JBassett:
image

JBassett:
Once you enable the sensors in settings they will be registered and just appear.  No restarts or resets needed.

balloob:
Not sure if I'm a fan of making the sensors configurable. How will we remove them from Home Assistant if we disable them?

About the timing, could we do:

 - Send sensor data to Home Assistant when the location updates (and we already woke up the device)
 - Set a timer to send a location update 15 minutes after we send the last one. If a zone change triggers a location update before, we reset the timer.

dtorner:
Talking just about the battery sensor, I would prefer to get it updated when it changes to a "charging" or "not_charging" state as well as the other timings, that way users could use the sensors for automations. Right now I use a custom app to send the battery state back to hass, when I wake up and disconnect the device from charging I turn on a light and get a weather telegram message...

dshokouhi:
I think we should definitely trigger an update based on the charging status and when the wifi connection status changes (i.e. connecting/disconnecting/switching AP's).

balloob:
Please do not comment with feature requests on PRs. Also [read all the comments](https://github.com/home-assistant/home-assistant-android/pull/336#discussion_r374112717) 🙄

JBassett:
Removed settings around the sensors and added logic stated above.

PR #341, opened 2020-02-05
Add support for images and actions no matter what the app is doing.

JBassett:
Needs https://github.com/home-assistant/mobile-apps-fcm-push/pull/2 to function.

Still need to think about backwards compatibility...

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -132,7 +132,7 @@ class MessagingService : FirebaseMessagingService() {
         }
 
         // TODO: This message id probably isn't the best
-        val messageId = (messageBody + messageTitle).hashCode()
+        val messageId = (messageBody + messageTitle + System.currentTimeMillis()).hashCode()
 
         actions.forEach {
             val actionIntent = Intent(this, NotificationActionReceiver::class.java).apply {
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -131,9 +131,13 @@ class MessagingService : FirebaseMessagingService() {
                     .bigLargeIcon(null))
         }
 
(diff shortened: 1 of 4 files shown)


balloob:
In what way wouldn't it be backwards compatible ? Actionable notifications have not been released yet.

JBassett:
Because of where we are pulling the title and body of the notification from.  In order to get the data delivered to the `onMessageReceived` function of our code we need to ensure that it is considered a data message.  This means we can't pass the title and message in the normal locations.  If we do pass them normally then we don't get the chance to process the data and the device just displays the notification.
https://firebase.google.com/docs/cloud-messaging/android/receive#handling_messages.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -60,20 +67,28 @@ class MessagingService : FirebaseMessagingService() {
...
 
-        remoteMessage.notification?.let {
-            if (it.body == "request_location_update") {
+            if (!it.containsKey(MESSAGE)) {
+                Log.e(TAG, "Message missing from notification.")
+                return
+            }
+
+            val title = it[TITLE]
+            val message = it[MESSAGE]!!
```
marciogranzotto: Why break the app if it's `null`?

```suggestion
            val message = it[MESSAGE] ?: ""
```

JBassett: The message is a required field from HA.  If it doesn't make it to the application we have a problem.

https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/notify/__init__.py#L45

balloob:
Oh I see the backwards compat issue now. It means that older apps won't be able to receive notifications, because it's in the data channel and that's not supported.

Could we push an update with this PR merged, and then give everyone a week to update their apps before we upgrade the push notifications cloud handling ?

JBassett:
The changes I made to the cloud function adds the new functionality in a new URL/function.  That way we can keep the old functionality around as long as we need.  Because I changed the `push_url` in the app it does need the new function in firebase first.

balloob:
Ok!

PR #355, opened 2020-02-09
Enable receiving broadcast to update location

quthla:
This enables other apps to send a broadcast to the HA app to update the device tracker's location


JBassett:
I think you want to get rid of this line as well:
https://github.com/home-assistant/home-assistant-android/pull/355/files#diff-c8cc2dd42271f2bf39c6aa81eb6a5529R59

Otherwise anyone calling this will need to have the `android.permission.RECEIVE_BOOT_COMPLETED` permission.

```
The name of a permission that broadcasters must have to send a message to the broadcast receiver. If this attribute is not set, the permission set by the <application> element's permission attribute applies to the broadcast receiver. If neither attribute is set, the receiver is not protected by a permission.
```

quthla:
Actually, my app holds that permission anyway, so I'll be fine but maybe others won't. 

I wasn't sure why you put it there, so I asked first before I would remove it.

PR #370, opened 2020-02-11
Make sure registration is required.

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/348

### DIFF
--- a/data/src/main/java/io/homeassistant/companion/android/data/url/UrlRepositoryImpl.kt
+++ b/data/src/main/java/io/homeassistant/companion/android/data/url/UrlRepositoryImpl.kt
@@ -28,6 +28,11 @@ class UrlRepositoryImpl @Inject constructor(
         val retVal = ArrayList<URL>()
         val webhook = localStorage.getString(PREF_WEBHOOK_ID)
 
+        // If we don't have a webhook id we don't have any urls.
+        if (webhook.isNullOrBlank()) {
+            return arrayOf()
+        }
+
         localStorage.getString(PREF_CLOUDHOOK_URL)?.let {
             retVal.add(it.toHttpUrl().toUrl())
         }


data/src/main/java/io/homeassistant/companion/android/data/url/UrlRepositoryImpl.kt
```diff
@@ -28,6 +28,11 @@ class UrlRepositoryImpl @Inject constructor(
         val retVal = ArrayList<URL>()
         val webhook = localStorage.getString(PREF_WEBHOOK_ID)
 
+        // If we don't have a webhook id we don't have any urls.
```
balloob: This is technically correct. But can't we abort onboarding if it fails ? And what happens if we return an empty array here?

JBassett: The check for an empty apiUrl list is already in place during registration, this defect is a regression.
Check for is registered: https://github.com/home-assistant/home-assistant-android/blob/master/app/src/main/java/io/homeassistant/companion/android/launch/LaunchPresenterImpl.kt#L33

isRegistered check: https://github.com/home-assistant/home-assistant-android/blob/master/data/src/main/java/io/homeassistant/companion/android/data/integration/IntegrationRepositoryImpl.kt#L122

erlanger:
I was running home assistant companion in lineageos without problems, but this commit breaks it.
The app gets stuck in a loop "unable to register app". 

* My guess is that this is happening because I am not using google services as I want my data privacy instead of google spying on my home
* In version 1.6 I also had the "unable to register app" message, but I restarted the app and it connected properly

Please keep this option open for those who are willing to give up notification in order to avoid using google services.

Issue #372, opened 2020-02-11
Long notification messages are truncated -- cannot expand


shbatm:
**Home Assistant Android version:**
App 1.6.0 (105)

**Android version:**
10

**Phone model:**
Samsung Galaxy S9+

**Home Assistant version:**
Home Assistant 0.105.2

**Last working Home Assistant release (if known):**
N/A

**Description of problem:**
Notification messages with text longer than 1 line are cut-off and cannot be expanded to show the rest of the text.

**Traceback (if applicable):**

```
N/A
```

**Additional information:**
Looks like (to my non-Android developer ignorant eye) notifications need a BigTextStyle added.
https://developer.android.com/training/notify-user/expanded#large-style

dshokouhi:
Just tested this myself with the following service data and I see the same issue.  Cannot pull down notification to read the rest.

```yaml
message: "This is a very long message text that will push down the notification by one line"
```

![image](https://user-images.githubusercontent.com/1634145/74215100-c4be6f80-4c54-11ea-9cdc-be7eb3410d8f.png)

shbatm:
(disclosure: untrained eye, just trying to help)

Looks like you may be able to add something like:
```java
        .setStyle(new NotificationCompat.BigTextStyle()
                .bigText(data[MESSAGE]))
```

At [this line](https://github.com/home-assistant/home-assistant-android/blob/aacb4ea605c8fd3c59d3b0550c128a47eafdc365/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt#L181) to fix the issue.

CodeFinder2:
I've realized this issue as well.

shbatm:
Tagging @JBassett since you've been doing a lot of great work on the notifications.

robbiet480:
@shbatm Mind submitting a PR so you get credit for the fix? Thanks!

PR #393, opened 2020-02-12
Allow expandable notifications for long text messages.

shbatm:
Proposed resolution for #372 to add support for a large text box in the notification if the message length exceeds one line.

Reference: https://developer.android.com/training/notify-user/expanded#large-style

Please note: I was unable to figure out how to test notifications from the debug app through Android Studio using my own Firebase config. I'll reach out on Discord, but if anyone on here can point me in the right direction, I'd appreciate it.


shbatm:
Tested using Firebase Console.

1) Collapsed notification (as received):
Screenshot_20200212-135808_Home Assistant

2) Expanded notification (with image) -- text will not expand; from what I can tell this may be a limitation of the Android notifications (see link in original PR). If anyone knows different, please let me know.


3) Text only notification expanded:
Screenshot_20200212-135913_Home Assistant

JBassett:
If you put a break point in the receiver do you hit it when you test with firebase?

Issue #363, opened 2020-02-11
Bigger Widgets


angeltonio:
The widget size is 1x1 but it´s possible to add the option for more bigger??...

KBerstene:
What kind of larger widget are you looking for?  Still one that is a single button?

JBassett:
Fixed in https://github.com/home-assistant/home-assistant-android/pull/401.
If you are looking for something more specific please create a new feature request with more details.

Issue #299, opened 2020-01-23
Widget icons and labels out of line and labels cut off


gadgetchnnel:
**Home Assistant Android version:**
1.4.0 (92)

**Android version:**
10

**Phone model:**
Huawei Mate 20 Pro

**Home Assistant version:**
0.104.3

**Last working Home Assistant release (if known):**

**Description of problem:**
Widget icons appear out of line with other icons on the home screen and the labels are cut off after very few characters (the second widget in the screenshot below should say "Computer Room").
![Screenshot_20200123_165600_com teslacoilsw launcher_2](https://user-images.githubusercontent.com/2099542/73013256-deb42180-3e0f-11ea-9fa8-df18eb513718.jpg)

Ideally the icon should be slightly higher up and the label should wrap over at least two lines

**Traceback (if applicable):**

```

```

**Additional information:**

andriej:
Same here (Android 10 / OnePlus 6)

balloob:
A widget is not an icon, so it is not meant to align with the launcher. There are also too many launchers out there, and we would have to support all of them. 

I do agree that we should explore if we can expand the space for the name.

JBassett:
Fixed in https://github.com/home-assistant/home-assistant-android/pull/401

PR #401, opened 2020-02-12
Reworked widget appearance and added resizability.

KBerstene:
I have reworked the appearance of the widgets to not look like app icons.  They're a bit clunkier (imho), but will not be confused with app icons as in #299.  The widgets are also resizeable, though it just increases the size of the icon and doesn't actually add any functionality.  It does mean you can create more "bar"-like icons rather than just square/rectangular buttons.

This also allows for quite a bit more text -- I know the initial design was not great for the amount of text you could put, but this should allow quite a bit more, even without resizing.

Oh, and unlike the last large widget update, this is backwards-compatible and will not break old widgets - when the app updates, they'll just update their shapes.

If it's not obvious, I don't have a ton of graphic design sense, so this is really the best I can do, visually.  If someone else wants to propose a prettier alternative, I would love to see it.

Addresses #299 and (kinda) addresses #363, though not in a real sense for the latter.  I'd' have to be given a direction to go in to create a widget with different functionality, since I am not sure what other people would like to have. This technically can be resized to bigger, though.

Images of a sample home screen are below (SDK 24).  I have also tested on Nova Launcher on Android 9.
Capture1

Image of pressed button displaying failure feedback:
Capture2
(I don't have a successful feedback image, but if you really need one I can get one)

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
@@ -139,8 +139,8 @@ class ButtonWidget : AppWidgetProvider() {
             RemoteViews(context.packageName, R.layout.widget_button)
         val appWidgetManager = AppWidgetManager.getInstance(context)
 
-        loadingViews.setInt(R.id.widgetProgressBar, "setVisibility", View.VISIBLE)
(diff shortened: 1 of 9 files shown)


robbiet480:
@KBerstene this looks awesome! We'll review shortly. In the meantime, [you should really come hang out on Discord with the cool kids](https://discord.gg/R3gmPq) :)

PR #405, opened 2020-02-13
Custom notification sounds

robbiet480:
I have no idea if this is a good idea or a bad idea, especially since it seems since Android O notification sounds are supposed to be tied to a channel.

I propose we add logic to the server side to map the original file names that iOS uses to the Android equivalent format so that we don't have to maintain two sets of lists. For example, `US-EN-Alexa-Son-Is-Arriving.wav` would become `us_en_alexa_son_is_arriving.wav`.

Also, for some reason I can't access the resources in main when debugging. I tested that this works by moving a single file into the debug module res/raw/.


app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -253,19 +257,44 @@ class MessagingService : FirebaseMessagingService() {
...
+            // FIXME: This crashes when applicationID includes the .debug suffix
+            val newSound = Uri.parse(ContentResolver.SCHEME_ANDROID_RESOURCE + "://" + applicationContext.getPackageName() + "/raw/" + soundString)
+            sound = newSound
+        }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val mp = MediaPlayer.create(applicationContext, sound)
+            val dndState = Settings.Global.getInt(getContentResolver(), "zen_mode")
+            if (dndState == 0) {
+                mp.start()
```
JBassett: Not sure how I feel about this.... Feels really hacky.

robbiet480:
Closing as too hacky.

We should just let users manage their channels and put the existing library of tones into [`DIRECTORY_RINGTONES`](https://developer.android.com/reference/android/os/Environment.html#DIRECTORY_RINGTONES) as per [this SO answer](https://stackoverflow.com/a/49605706/486182).

VooPoc:
Was this ever worked out and implemented?

JBassett:
@VooPoc We recently implemented notification channels.  This should allow you to specify per channel notification settings.  This is in the beta currently and when promoted will be in the live app.  https://github.com/home-assistant/android/pull/574

VooPoc:
@JBassett thanks for your guy's efforts on this, I've been following the progress and just implemented it.

PR #413, opened 2020-02-13
Feature/new release flow

JBassett:


README.md
```diff
@@ -24,7 +24,7 @@ If you are looking for documentation around the companion applications check out
 
 ## Testing Dev Releases
 
-We are using [Travis](https://travis-ci.com/home-assistant/home-assistant-android) to perform continuous integration both by unit testing, deploying dev releases to [Firebase App Distribution](https://appdistribution.firebase.dev/i/8zf5W4zz) and final releases to the [Play Store](https://play.google.com/store/apps/details?id=io.homeassistant.companion.android) when we add a git tag.
+We are using [Travis](https://travis-ci.com/home-assistant/home-assistant-android) to perform continuous integration both by unit testing, deploying dev releases to [Play Store Beta](https://play.google.com/apps/testing/io.homeassistant.companion.android) and final releases to the [Play Store](https://play.google.com/store/apps/details?id=io.homeassistant.companion.android) when we release.
```
dshokouhi: So we are just removing the link to firebase because we are at the max 200 users right? Firebase will still be there?

JBassett: Currently it's still configured to deploy to Firebase as well on commit.  Not sure long term if we plan to keep both though.

Issue #329, opened 2020-02-01
New functionnalities


k2v1n58:
Is it possible to have 2 new functionalities:

- Switch screen off (similar to Fully kiosk) that can be used with browser_mod (browser_mod.blackout)
- Keep screen ON to prevent the screen to switch off automatically

balloob:
Duplicate of #325

Issue #325, opened 2020-01-30
Dim screen when not using it like fully kiosk browser


sverleysen:
It would be nice if you could use the Androids device sensors (or camera) to dim the screen when you aren't using the screen, or no motion is detected on the camera.
That is currently the only reason for me not to use the app on my wall mounted tablet.

balloob:
It's not the focus of our app to add a ton of features for tablet use. Our focus is to be used on phones, not as an always on solution. We shouldn't add this.

ngdio:
Would it hurt to have kiosk features though? I also want to use Home Assistant on my tablet and would like features like this one. Or is it too difficult finding someone to develop/maintain such features?

balloob:
Every feature needs to be maintained forever. It needs to be included in every update forever. It needs to be translated. And it can make adding other features more complicated, as all features need to be considered with every other feature toggled on/off. 

And so we need to set a goal for what we want to achieve and focus on that. Other apps can focus on providing a great kiosk experience.

Issue #61, opened 2019-11-23
Screen does not go to sleep if you have video stream showing


mmatesic01:
Hi,
When using this app, if I leave it on tab with video stream(live Picture glance entity ), screen does not turn off after expected inactivity.
When leaving in that state on web page view through any of web browsers this does not happen, screen turn off as expected.
Can you fix this?

NicoBijl:
I am not sure if 'turn off after inactivity' is the expected behaviour. My screen does not turn off either when I watch youtube or netflix for that mather..

And, when for example watching the camera on my daughters room to see if she is asleep, I like the fact that the screen stays on.

robbiet480:
Similar to #277

PR #414, opened 2020-02-13
Add Dim Screen & Keep Screen On 

neopilou:
Add Keep Screen On in webview property 
Add setting to set timeout for dim screen : 
    -> If set to 0 or empty : no timeout, screen keep on
    -> If set above 0 : screen dim after timeout & webview paused (which allows the device to go to sleep even if there is a video stream (camera in picture entity or picture glance for example))

Resolves the following issues : [#61](https://github.com/home-assistant/home-assistant-android/issues/61), [#325](https://github.com/home-assistant/home-assistant-android/issues/325) & [#329](https://github.com/home-assistant/home-assistant-android/issues/329)

Screenshot_20200213-173101_Home Assistant

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -12,7 +12,7 @@
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.RECORD_AUDIO" />
     <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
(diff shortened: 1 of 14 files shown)


klejejs:
Does 0 mean that screen will work based on device's display settings or will it stay on all the time?

neopilou:
Stay on all the time if set to 0

klejejs:
In this case it would be nice if mobile users could disable it as it is not needed for simple day to day mobile use.

PR #423, opened 2020-02-14
Build/Release Changes

JBassett:
Now when we tag we auto migrate from beta to production.


robbiet480:
Shall we test this by recreating 1.7.2 tag in a few?

JBassett:
Sounds like a good plan.

Issue #427, opened 2020-02-14
Allow YAML mode for entering widget service data


gadgetchnnel:
The current auto complete fields and the "Add Field" button when adding a widget work well for services with simple options. However, they don't work for services which have nested properties. For example, it isn't possible to create a widget which will make the equivalents of this service call:

```
service: notify.mobile_app_other_phone
data:
  message: "I have arrived."
  data:
    sticky: true
    actions:
      - action: confirm
         title: "OK"
```
Other, simpler, examples also don't work, such as:

```
service: light.turn_on
data:
  entity_id: light.test_light
  rgb_color: [255,0,0]
```

To allow this sort of thing it would be useful to be able to switch to YAML mode and enter the service data as YAML rather than via the fields.

dshokouhi:
Please do not delete the issue template as all of the information is important.  In this case knowing the app version you are using would be helpful.

Just now I tried to create a widget for `light.turn_on` and I used the autocomplete field to enter the entity ID and I also manually typed in `100` in `brightness_pct` and the widget worked appropriately.  Are you seeing certain fields not working?  It seems to be working as intended.  If possible please share a screenshot of the widget creation screen too so we can see what you are entering.

dshokouhi:
I do see your point on nested properties though, that is indeed something that should be corrected.

gadgetchnnel:
@dshokouhi I'd removed the template as I'd considered this a feature request rather than a bug report (maybe a feature request template is required). My app version is 1.7.2-128.

Here is a cropped screenshot (I cropped out some of the fields I wasn't using for size reasons) of what I tried with the light.turn_on service:

![Screenshot_20200214_184935_io homeassistant companion android](https://user-images.githubusercontent.com/2099542/74559156-a362c980-4f5b-11ea-930a-df5a0225fa74.jpg)

I think the issue there was that rgb_color is actually a nested property, and should take an array of red, green, and blue values (which could be entered if YAML mode were implemented) rather than just a string.

dshokouhi:
Yea RGB should not be sent as a string like you said.  The widget itself reads from services.yaml so when you use something like `notify.mobile_app` there are no values to provide so they must be entered manually, here YAML mode will work.  When you use the `light` domain services we do have values but they should be sent appropriately.  We may have the same issue with boolean values but would need to check that.

https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/light/services.yaml

[...]

PR #443, opened 2020-02-17
Added ability for widget service calls to contain lists

KBerstene:
I have implemented the ability for widget service data to contain lists.  Lists are indicated by comma separations, and they ignore whitespace.

Additionally, the service data will parse into proper numbers, strings, and boolean values.  Any valid YAML boolean string is accepted, including on/off and yes/no.

Partial fix for #427.  It allows lists, but not full JSON objects.  I don't know if you want to allow a raw YAML editor mode or not.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/WidgetDynamicFieldAdapter.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/WidgetDynamicFieldAdapter.kt
@@ -6,9 +6,11 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.AutoCompleteTextView
+import android.widget.MultiAutoCompleteTextView.CommaTokenizer
 import androidx.recyclerview.widget.RecyclerView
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/widgets/WidgetDynamicFieldAdapter.kt
```diff
@@ -118,4 +145,54 @@ class WidgetDynamicFieldAdapter(
...
+    private fun String.toJsonType(): Any? {
+        // Parse the string to one of the following
+        // valid base JSON types:
+        // array, number, boolean, string
+
+        // Check if the text is an array
+        if (this.contains(",")) {
+            val jsonArray = ArrayList<Any>()
+
+            this.split(",").forEach { subString ->
```
JBassett: What happens when the language has `1.1` = `1,1`?

JBassett: After thinking about this more I think we can expect that they use `.` because that's what the JSON format supports.

PR #455, opened 2020-02-22
Add better handling for slow connections.

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/454

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -307,11 +307,7 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
     override fun loadUrl(url: String) {
         loadedUrl = url
         webView.loadUrl(url)
-        Handler().postDelayed({
-            if (!isConnected) {
-                showError()
-            }
-        }, 5000)
+        waitForConnection()
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -385,4 +396,12 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
             cutout = true
         return cutout
     }
+
+    private fun waitForConnection() {
+        Handler().postDelayed({
+            if (!isConnected) {
+                showError()
```
balloob: Can we let this screen automatically hide if `connection-status` event is fired that we're `connected` ?

JBassett: Done.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -222,6 +223,7 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
                             "connection-status" -> {
                                 isConnected = json.getJSONObject("payload")
                                     .getString("event") == "connected"
+                                alertDialog?.cancel()
```
balloob: We should only cancel if it was "connected" (`isConnected` is true)

PR #456, opened 2020-02-22
Better handling of action events.

JBassett:
Show toast when we fail to send event and keep notification so you can
retry.

Fixes: https://github.com/home-assistant/home-assistant-android/issues/453

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
@@ -8,6 +8,7 @@ import android.net.Uri
 import android.os.Handler
 import android.util.Log
 import android.widget.Toast
+import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
 import io.homeassistant.companion.android.domain.integration.IntegrationUseCase
 import javax.inject.Inject
@@ -52,7 +53,7 @@ class NotificationActionReceiver : BroadcastReceiver() {
         }
         val onFailure: () -> Unit = {
             Handler(context.mainLooper).post {
-                Toast.makeText(context, "Unable to send event to HA.", Toast.LENGTH_LONG).show()
+                Toast.makeText(context, R.string.event_error, Toast.LENGTH_LONG).show()
             }
         }
         when (intent.action) {
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
@@ -5,7 +5,9 @@ import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
 import android.net.Uri
+import android.os.Handler
 import android.util.Log
+import android.widget.Toast
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
 import io.homeassistant.companion.android.domain.integration.IntegrationUseCase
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
```diff
@@ -42,32 +45,46 @@ class NotificationActionReceiver : BroadcastReceiver() {
...
-            val notificationService: NotificationManager =
-                context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
-            notificationService.cancel(messageId)
+        val onComplete: () -> Unit = {
+            (context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager)
+                .cancel(messageId)
+        }
+        val onFailure: () -> Unit = {
+            Handler(context.mainLooper).post {
+                Toast.makeText(context, "Unable to send event to HA.", Toast.LENGTH_LONG).show()
```
balloob: I feel like this text can be better. 

Maybe "Failed to communicate with Home Assistant" ?

balloob: Oh wait this is only for events when clicking notification actions.

balloob: This is ok .

app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
```diff
@@ -42,32 +46,46 @@ class NotificationActionReceiver : BroadcastReceiver() {
...
-                action.data.plus(Pair("action", action.key))
-            )
+            try {
+                integrationUseCase.fireEvent(
+                    "mobile_app_notification_action",
+                    action.data.plus(Pair("action", action.key))
+                )
+                onComplete()
+            } catch (e: Exception) {
+                Log.e(TAG, "Unable to fire event.", e)
```
balloob: Should this be the same `event_error` string ?

app/src/main/res/values/strings.xml
```diff
@@ -70,4 +70,5 @@ like to connect to:</string>
   <string name="add_service_data_field">Add Field</string>
   <string name="widget_config_service_error">A custom component is preventing service data from loading.</string>
   <string name="error_ssl">Unable to communicate with Home Assistant because of a SSL error.  Please ensure your certificate is valid.</string>
+    <string name="event_error">Unable to send event to HA.</string>
```
balloob: Thinking about this again, the event really is an implementation detail of how we have designed actionable notifications. Should it be "Unable to notify HA of picked option." ?

app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
JBassett: Since this is just the ADB log it should never get shown to an end user.

app/src/main/res/values/strings.xml
JBassett: Done

app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
robbiet480: This should've been a localizable string :(

robbiet480: Never mind, it ended up being a proper string!

PR #457, opened 2020-02-23
Allow front end to force credential refresh.

JBassett:
Fixes: https://github.com/home-assistant/home-assistant-android/issues/257


JBassett:
I wasn't able to test the full stack because I wasn't sure how to cause the frontend to send the force payload but I validated the rest of the stack.

balloob:
btw to test this, you should be able to revoke the refresh token that is being used by the app.

Issue #447, opened 2020-02-18
Cannot view videos in full screen


a-jackson:
**Home Assistant Android version:**
1.7.2-128
**Android version:**
10
**Phone model:**
Pixel 2 XL
**Home Assistant version:**
0.105.5
**Last working Home Assistant release (if known):**
N/A
**Description of problem:**
Video in a picture-entity card cannot be displayed full screen. I can do it on desktop and on the mobile website in both Chrome and Firefox but the full screen button in the bottom right corner of the video is greyed out in the app.

As a workaround I have been able to login on Chrome and then use Add to Home Screen, when opened like this it does allow full screen.

**Traceback (if applicable):**

```

```

**Screenshot of problem:**

![Screenshot_20200218-191710](https://user-images.githubusercontent.com/2199309/74770203-b212ee80-5283-11ea-8eb3-d4d43ee84765.png)
Full screen button is greyed out

**Additional information:**

I have seen this on 2 other phones, a Galaxy S9 and a Pixel 2 and one was a differnet Home Assistant server however I don't have version information for them.

dshokouhi:
In https://github.com/home-assistant/home-assistant-android/issues/267 I did mention that the full screen button is disabled which prevents PiP from happening so this is either related or a duplicate

dshokouhi:
Closed in #471

Issue #267, opened 2020-01-17
Support PiP when viewing a stream enabled camera


dshokouhi:
**Home Assistant Android version:**
1.3.1 (77)
**Android version:**
10
**Phone model:**
Pixel 4 XL
**Home Assistant version:**
0.103.3
**Last working Home Assistant release (if known):**
N/A
**Description of problem:**
We should support PiP when a camera is opened and the video stream starts.  

It also looks like the full screen icon is disabled, that may be the reason why PiP is not working?

![image](https://user-images.githubusercontent.com/1634145/72574132-da6e8e00-387c-11ea-9ce0-1517965b37af.png)

**Traceback (if applicable):**

```

```

**Additional information:**

dshokouhi:
important to note that this works as expected in chrome so maybe missing permissions or something?

![image](https://user-images.githubusercontent.com/1634145/72575487-74d0d080-3881-11ea-8218-18ab2626db1a.png)

![image](https://user-images.githubusercontent.com/1634145/72575498-7d290b80-3881-11ea-8b0f-b6edb97d4cad.png)

Pe-MaKer:
Same here.
Please take care that camera streaming on the app works the same as it does on chrome. Without scrolling the video due to the useless camera name banner and ideally with pip support.

Zamunda4Eva:
Yes, please add this feature.

a-jackson:
I'm getting the same problem as this as well. Also when turning the phone landscape the header banner with the camera name is pushing down the video chopping off the bottom

dshokouhi:
@a-jackson that sounds like a separate issue. If you experience that issue in Chrome on Android then please open a bug on the front end repo.  If it only exists on the android app then please open a new issue here for it.

dshokouhi:
Closed in #471

PR #458, opened 2020-02-23
Add support for full screen video (#447)

a-jackson:
Adds support for videos to go full screen based on the solution from [here](https://stackoverflow.com/questions/15796661/android-webview-app-wont-let-video-player-go-full-screen)

Have tested on emulator and my Pixel 2 XL

### DIFF
--- a/app/src/main/res/layout/activity_webview.xml
+++ b/app/src/main/res/layout/activity_webview.xml
@@ -1,5 +1,19 @@
 <?xml version="1.0" encoding="utf-8"?>
-<WebView xmlns:android="http://schemas.android.com/apk/res/android"
-    android:id="@+id/webview"
+<ViewFlipper xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/webview_view_flipper"
     android:layout_width="match_parent"
-    android:layout_height="match_parent" />
+    android:layout_height="match_parent">
+
+    <WebView
+        android:layout_width="fill_parent"
+        android:layout_height="fill_parent"
+        android:id="@+id/webview"
+        android:layout_gravity="center"
+        />
+
+    <FrameLayout
+        android:id="@+id/fullScreenContainer"
+        android:layout_width="fill_parent"
+        android:layout_height="fill_parent"
+        />
+</ViewFlipper>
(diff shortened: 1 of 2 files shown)


homeassistant:
Hi @a-jackson, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant-android#458).

Thanks!

dshokouhi:
This will close #447 
 
@a-jackson can you confirm if this will also closes #267? I have a feeling the 2 are interlinked by not having full screen enabled in the video player.

neopilou:
@dshokouhi 
Fullscreen video and pip are two different issues. 
This just enable fullscreen option in video
I tried to have pip and it works like chrome, except that the pip window displays the entire original video in full screen (with black bands, which is not good in vertical orientation) and not only the video

a-jackson:
No it doesn't support picture in picture. It looks like there's some more work for that. I found this https://developer.android.com/guide/topics/ui/picture-in-picture but I'm not very experienced with Android devevelopment. I think it was Android 4 the last time I did any, a lot has changed 😄

app/src/main/res/layout/activity_webview.xml
```diff
@@ -1,5 +1,19 @@
 <?xml version="1.0" encoding="utf-8"?>
-<WebView xmlns:android="http://schemas.android.com/apk/res/android"
-    android:id="@+id/webview"
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
JBassett: Rather than setting visibility it might be cleaner if you use a [ViewFlipper](https://developer.android.com/reference/android/widget/ViewFlipper).  We already use it in the MobileAppIntegrationFragment.

a-jackson:
@JBassett thanks for the review. I've made the changes requested and rebased due to some merge conflicts. You're right ViewFlipper is much cleaner, I didn't know it existed 😄

JBassett:
So it looks like @neopilou got both full screen and PiP implemented in his PR: https://github.com/home-assistant/home-assistant-android/pull/471  Since his covers both cases I'm going to merge his in this case.

PR #493, opened 2020-03-04
Enable entity_picture for device tracker

seamus65:
I have tried to set the entity_picture attribute for the device tracker generated by the app, but this gets cleared everytime the tracker updates.


dshokouhi:
You probably want to clean up the commits so you don't include a bunch of other commits with your change.  Might be easier to start with a clean PR that contains only your commit.

seamus65:
Sorry, it's my first time.

balloob:
You need to create an issue, not a PR.

Setting entity picture is not supported. You need to set this on persons instead.

Issue #391, opened 2020-02-12
Sent battery status update is sent whenever the charging state changes.


Gluwc:
It would be great if a battery status update is sent whenever the charging state changes. Currently it only appears to sends updates to the battery status whenever a GPS update gets triggered.

This will be awesome for some automations like going to bed. Things like plugging in your phone will send an actional notification to ask if you want to enable sleep mode.

Another consideration could be to update all battery status in real-time whenever the phone is charging since it is charging anyway, but less important in my opinion.

Thanks in advance, hugs and kisses, yours sincerely,

Lucas

Twanislas:
Big +1

To me, battery status should be pushed to HA on every state change (charging / not charging) by default.

Actually I would love for that to be a toggle (on/off) setting in the app, along with :
- Update battery in realtime when charging
- Update location in realtime when charging
- Update <whatever sensor> in realtime when charging

We'd need to factor in the data usage as well. I don't care a thousand updates every 30 sec when I'm on wifi but maybe I don't want that on carrier data...

"realtime" being something along 1m or 30s of course.

This would allow (for example) realtime location tracking when driving (and plugged to a power source in the car) and thus fast reaction when coming home (open the garage door, TTS notifications announcing I'm nearing home on all the Google Mini's and such)

balloob:
@Twanislas please don't add feature requests in existing issues but open a new issue.

Twanislas:
@balloob very good point. I was just trying to give ideas.
Still, very good point :D. I'll go ahead and open FRs.
Thanks !

tdejneka:
I can confirm that the reported battery level can lag the battery's true state.

I've been using a Macrodroid macro to report the phone's battery state (every time it changes by 1%) and it's reported as a sensor entity. I noticed this entity indicated 64 whereas the Companion's battery sensor reported 55. 

The phone was plugged in, charging, and connected to WiFi. Location Services are _off_.

A few minutes later the Companion's battery sensor reported 64 but the macro-based sensor had already moved on to report 65.

I don't know what mechanism it uses to report battery level. If it's time-based then it allows for (on my phone) a ~10% discrepancy between reported and actual battery level. I would prefer to see it report by fixed increments of 5% (5, 10, 15, 20, 25, etc).

dshokouhi:
This is now implemented in #525 battery status updates when the phone is goes on/off the charger.

PR #525, opened 2020-03-22
Added ChargingBroadcastReceiver which can instantly update plugged in…

yoxjames:
One thing I ran into when doing my own home automation was the desire to figure create an abstract concept for when everyone in the house is asleep. The simplest way to accomplish this was to see if both me and my partner had our phones plugged in. This is something we always do before bed. I wanted the fan to instantly go on in this scenario as well as the noise machine. 

I was able to set this up using the battery sensor but I noticed there was a lot of lag time... Around 15 minutes which made me immediately think that this was being driven by WorkManager which has a 15 minute minimum periodic work run time. I pulled down the code and found that to be correct as the code was running SensorWorker every 15 minutes as a Periodic Work Request. 

I wanted it to be instant though, and one way I immediately thought to do that was via a BroadcastReceiver. I got this working locally and can now enjoy my sleep automations firing instantly. I thought I should share the little work I have done as I really like this project and would like to contribute more in the future! I figured this would be an interesting way to get my feet wet even if there needs to be some changes.

So what I did was register a Broadcast Receiver for ACTION_POWER_CONNECTED and ACTION_POWER_DISCONNECTED. You might be wondering why I didn't just register a Broadcast Receiver for ACTION_BATTERY_CHANGED. I actually did that in a previous version but found that it was executing _a lot_. This might be totally fine but I wanted to avoid doing something that had serious performance ramifications. I can totally change it back. However my current approach runs a lot less but has one important caveat. The `is_charging` state attribute will not immediately update while the `charger_type` attribute will. This is, presumably, because while Android knows that it was plugged in instantly some time must elapse for Android to be certain that the connection is actually supplying sufficient power to charge the battery. Switching to ACTION_BATTERY_CHANGED remedies this (`is_charging` is updated shortly after the `charger_type`) but again I had some concerns about this.

Furthermore I made one additional structural change you can take or leave. I switched from starting the SensorManager in various activities and fragments and instead moved it to the main App class and also registered another BroadcastReceiver for BOOT events so that the SensorWorker would be started when the device boots. I noticed when playing around with this that two SensorWorkers were actually being enqueued. I changed around a couple of the function calls (such as using `enqueueUniquePeriodicWork` which appears to have remedied this.

This is my first ever open source contribution. I am super open to criticism and being told to redo things. While I have made changes that I think are improvements I recognize there may be things I dont know or failed to consider. I hope that in some form I can contribute this change and hope this starts a conversation about how we might be able to avoid lag time when updating sensors! I look forward to your feedback!

Quick note on testing. I didn't see a great opportunity to add tests to my contribution. I am a champion of the importance of proper unit testing and would be happy to add them if you think that would help. I could probably wedge a presenter into the BroadcastReceivers I have created and do tests that way. However I didn't find that to be super valuable for what I was doing in this PR.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -46,6 +46,13 @@
                 android:resource="@xml/button_widget_info" />
         </receiver>
 
+        <!-- Start things like SensorWorker on device boot -->
+        <receiver android:name=".HomeAssistantBootReceiver">
+            <intent-filter>
+                <action android:name="android.intent.action.BOOT_COMPLETED"/>
+            </intent-filter>
+        </receiver>
+
         <activity android:name=".widgets.ButtonWidgetConfigureActivity">
             <intent-filter>
                 <action android:name="android.appwidget.action.APPWIDGET_CONFIGURE" />
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -64,6 +64,9 @@
             android:enabled="true"
             android:exported="true">
             <intent-filter>
+                <action android:name="android.intent.action.BOOT_COMPLETED" />
+                <action android:name="android.intent.action.QUICKBOOT_POWERON" />
+                <action android:name="com.htc.intent.action.QUICKBOOT_POWERON" />
                 <action android:name="io.homeassistant.companion.android.background.REQUEST_ACCURATE_UPDATE" />
             </intent-filter>
         </receiver>
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -64,9 +64,6 @@
             android:enabled="true"
             android:exported="true">
             <intent-filter>
-                <action android:name="android.intent.action.BOOT_COMPLETED" />
-                <action android:name="android.intent.action.QUICKBOOT_POWERON" />
-                <action android:name="com.htc.intent.action.QUICKBOOT_POWERON" />
                 <action android:name="io.homeassistant.companion.android.background.REQUEST_ACCURATE_UPDATE" />
             </intent-filter>
         </receiver>
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -1,11 +1,14 @@
 package io.homeassistant.companion.android
 
 import android.app.Application
+import android.content.Intent
+import android.content.IntentFilter
 import com.jakewharton.threetenabp.AndroidThreeTen
 import com.lokalise.sdk.Lokalise
 import io.homeassistant.companion.android.common.dagger.AppComponent
 import io.homeassistant.companion.android.common.dagger.Graph
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
+import io.homeassistant.companion.android.sensors.ChargingBroadcastReceiver
 import io.homeassistant.companion.android.sensors.SensorWorker
 
 class HomeAssistantApplication : Application(), GraphComponentAccessor {
@@ -22,6 +25,16 @@ class HomeAssistantApplication : Application(), GraphComponentAccessor {
         graph = Graph(this, 0)
         // Start the sensor worker if they start the app. The only other place we start this ia Boot BroadcastReceiver
         SensorWorker.start(this)
+
+        // This will cause the sensor to be updated every time the OS broadcasts that a cable was plugged/unplugged.
+        // This should be nearly instantaneous allowing automations to fire immediately when a phone is plugged
+        // in or unplugged.
+        registerReceiver(
+            ChargingBroadcastReceiver(appComponent.integrationUseCase()), IntentFilter().apply {
+                addAction(Intent.ACTION_POWER_CONNECTED)
+                addAction(Intent.ACTION_POWER_DISCONNECTED)
+            }
+        )
     }
 
     override val appComponent: AppComponent
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -6,6 +6,7 @@ import com.lokalise.sdk.Lokalise
 import io.homeassistant.companion.android.common.dagger.AppComponent
 import io.homeassistant.companion.android.common.dagger.Graph
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
+import io.homeassistant.companion.android.sensors.SensorWorker
 
 class HomeAssistantApplication : Application(), GraphComponentAccessor {
 
@@ -19,6 +20,8 @@ class HomeAssistantApplication : Application(), GraphComponentAccessor {
 
         AndroidThreeTen.init(this)
         graph = Graph(this, 0)
+        // Start the sensor worker if they start the app. The only other place we start this ia Boot BroadcastReceiver
+        SensorWorker.start(this)
     }
 
     override val appComponent: AppComponent
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -20,7 +20,7 @@ class HomeAssistantApplication : Application(), GraphComponentAccessor {
 
         AndroidThreeTen.init(this)
         graph = Graph(this, 0)
-        // Start the sensor worker if they start the app. The only other place we start this a Boot BroadcastReceiver
+        // Start the sensor worker if they start the app. The only other place we start this ia Boot BroadcastReceiver
         SensorWorker.start(this)
     }
 
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -20,7 +20,7 @@ class HomeAssistantApplication : Application(), GraphComponentAccessor {
 
         AndroidThreeTen.init(this)
         graph = Graph(this, 0)
-        // Start the sensor worker if they start the app. The only other place we start this ia Boot BroadcastReceiver
+        // Start the sensor worker if they start the app. The only other place we start this a Boot BroadcastReceiver
         SensorWorker.start(this)
     }
 
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantBootReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantBootReceiver.kt
@@ -1,0 +1,16 @@
+package io.homeassistant.companion.android
+
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import io.homeassistant.companion.android.sensors.SensorWorker
+
+/**
+ * This class will receive a broadcast intent when the device boots up. This allows us to start the sensor worker
+ * even if the user never starts the HomeAssistant app.
+ */
+class HomeAssistantBootReceiver : BroadcastReceiver() {
+    override fun onReceive(context: Context, intent: Intent) {
+        SensorWorker.start(context)
+    }
+}
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantBootReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantBootReceiver.kt
@@ -4,6 +4,7 @@ import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
 import io.homeassistant.companion.android.sensors.SensorWorker
+import io.homeassistant.companion.android.util.PermissionManager
 
 /**
  * This class will receive a broadcast intent when the device boots up. This allows us to start the sensor worker
@@ -13,6 +14,7 @@ class HomeAssistantBootReceiver : BroadcastReceiver() {
     override fun onReceive(context: Context, intent: Intent) {
         if (intent.action == Intent.ACTION_BOOT_COMPLETED) {
             SensorWorker.start(context)
+            PermissionManager.restartLocationTracking(context)
         }
     }
 }
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantBootReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantBootReceiver.kt
@@ -11,6 +11,8 @@ import io.homeassistant.companion.android.sensors.SensorWorker
  */
 class HomeAssistantBootReceiver : BroadcastReceiver() {
     override fun onReceive(context: Context, intent: Intent) {
-        SensorWorker.start(context)
+        if (intent.action == Intent.ACTION_BOOT_COMPLETED) {
+            SensorWorker.start(context)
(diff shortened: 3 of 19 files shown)


homeassistant:
Hi @yoxjames, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant-android#525).

Thanks!

app/src/main/java/io/homeassistant/companion/android/sensors/SensorWorker.kt
```diff
@@ -27,11 +30,9 @@ class SensorWorker(private val appContext: Context, workerParams: WorkerParamete
             val sensorWorker =
                 PeriodicWorkRequestBuilder<SensorWorker>(15, TimeUnit.MINUTES)
                     .setConstraints(constraints)
-                    .addTag("sensors")
                     .build()
 
-            WorkManager.getInstance(context).cancelAllWorkByTag("sensors")
-            WorkManager.getInstance(context).enqueue(sensorWorker)
+            WorkManager.getInstance(context).enqueueUniquePeriodicWork(TAG, ExistingPeriodicWorkPolicy.REPLACE, sensorWorker)
```
yoxjames: This change appeared to stop duplicate `SensorWorker`s from operating. I was noticing that when doing some debugging myself. The only odd thing here is that I used the `TAG` from the companion object. I could use "sensors" as was done previously if that is desired.

app/src/main/java/io/homeassistant/companion/android/sensors/ChargingBroadcastReceiver.kt
```diff
@@ -0,0 +1,38 @@
...
+            val batterySensorManager = BatterySensorManager().registerSensors(context)
+            integrationUseCase.updateSensors(
+                batterySensorManager.getSensors(context).toTypedArray()
+            )
+        }
+    }
+
+    private suspend fun BatterySensorManager.registerSensors(context: Context) = apply {
+        getSensorRegistrations(context).forEach {
+            try {
```
yoxjames: The other place where we do this (`SensorWorker`) also used this pattern. I had debated finding some way to avoid duplication of logic here but ended up not really being sure where to put that.

app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
```diff
@@ -19,6 +20,8 @@ class HomeAssistantApplication : Application(), GraphComponentAccessor {
 
         AndroidThreeTen.init(this)
         graph = Graph(this, 0)
+        // Start the sensor worker if they start the app. The only other place we start this a Boot BroadcastReceiver
+        SensorWorker.start(this)
```
balloob: We only report sensor values if location reporting is on. I think with this change it will always report ? Also it looks like it was called inside `sendLocationUpdate` because it would restart the 15 minute interval. The idea was that it would send at least once every 15 minutes, but more often if other things cause data to be sent (like connected status changed via this PR)

app/src/main/java/io/homeassistant/companion/android/sensors/SensorWorker.kt
```diff
@@ -43,6 +44,16 @@ class SensorWorker(private val appContext: Context, workerParams: WorkerParamete
...
             .build()
             .inject(this)
+
+        // This will cause the sensor to be updated every time the OS broadcasts that a cable was plugged/unplugged.
+        // This should be nearly instantaneous allowing automations to fire immediately when a phone is plugged
+        // in or unplugged.
+        appContext.registerReceiver(
+            ChargingBroadcastReceiver(integrationUseCase), IntentFilter().apply {
+                addAction(Intent.ACTION_POWER_CONNECTED)
+                addAction(Intent.ACTION_POWER_DISCONNECTED)
```
balloob: Does this also trigger for wireless charging or are those other intents?

balloob:
Hey, thanks for the contribution and welcome to the Home Assistant community. I'm not the main Android dev, but did leave some comments.

app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
JBassett: I like the idea of sending sensors even with location disabled personally.  That start method will send all available sensors then restart the 15 minute timer, it should be safe to call here.

app/src/main/AndroidManifest.xml
```diff
@@ -46,6 +46,13 @@
                 android:resource="@xml/button_widget_info" />
         </receiver>
 
+        <!-- Start things like SensorWorker on device boot -->
+        <receiver android:name=".HomeAssistantBootReceiver">
+            <intent-filter>
+                <action android:name="android.intent.action.BOOT_COMPLETED"/>
```
JBassett: If we are going to add a generic boot receiver then we should remove the boot specific logic from `LocationBroadcastReceiver` and send a broadcast to it instead starting location tracking.

app/src/main/java/io/homeassistant/companion/android/sensors/ChargingBroadcastReceiver.kt
```diff
@@ -0,0 +1,38 @@
...
+    private var updateJob: Job? = null
+    override fun onReceive(context: Context, intent: Intent?) {
+        updateJob?.cancel()
+        updateJob = ioScope.launch {
+            val batterySensorManager = BatterySensorManager().registerSensors(context)
+            integrationUseCase.updateSensors(
+                batterySensorManager.getSensors(context).toTypedArray()
+            )
+        }
+    }
```
JBassett: Rather than duplicating the logic why not call `SensorWorker.start(context)`?  That way we get all sensors updating.

yoxjames: Interesting thought. I'll give that a try. It should work as long as it replaces the existing SensorWorker.

app/src/main/AndroidManifest.xml
yoxjames: I can look into this change!

app/src/main/java/io/homeassistant/companion/android/sensors/SensorWorker.kt
yoxjames: According to the Android docs it should. I can confirm this with an emulator though.

dshokouhi:
This should close #391 however with the caveat that `is_charging` will not immediately update.

app/src/main/java/io/homeassistant/companion/android/sensors/ChargingBroadcastReceiver.kt
yoxjames: I could even add a delay to the restart of SensorWorker which might get around is_charging somewhat. It's using a delay as a solution which generally is not the best idea imo. It probably makes more sense to update what the OS immediately knows which is that the charger_type vs trying to hack around getting is_charging to work right.

In any case I'll mess around with this soon.

balloob: Let's try not to add more features to this PR. Instead, keep it for a future PR. It's easier to review small PRs.

balloob: Also, if you're interested, you can join us on #devs_mobile_apps on the Home Assistant Discord server. It's where we hangout :) https://www.home-assistant.io/join-chat/

yoxjames:
Updated the PR to address the comments everyone had.

app/src/main/AndroidManifest.xml
```diff
@@ -57,9 +64,6 @@
             android:enabled="true"
             android:exported="true">
             <intent-filter>
-                <action android:name="android.intent.action.BOOT_COMPLETED" />
-                <action android:name="android.intent.action.QUICKBOOT_POWERON" />
```
yoxjames: I ended up just removing QUICKBOOT_POWERON (both versions). They did not appear to actually do anything that I could see before.

app/src/main/java/io/homeassistant/companion/android/sensors/SensorWorker.kt
yoxjames: Unfortunately it doesn't look like the emulator actually supports any plug type other than AC which makes testing this tough. However according to the Android docs this approach is sound as any type of power connection should theoretically fire this Intent.

app/src/main/AndroidManifest.xml
JBassett: We added these extra actions because some device manufactures use different actions.
https://stackoverflow.com/a/42622385

app/src/main/java/io/homeassistant/companion/android/HomeAssistantBootReceiver.kt
```diff
@@ -0,0 +1,18 @@
...
+ * This class will receive a broadcast intent when the device boots up. This allows us to start the sensor worker
+ * even if the user never starts the HomeAssistant app.
+ */
+class HomeAssistantBootReceiver : BroadcastReceiver() {
+    override fun onReceive(context: Context, intent: Intent) {
+        if (intent.action == Intent.ACTION_BOOT_COMPLETED) {
+            SensorWorker.start(context)
+        }
+    }
+}
```
JBassett: Now that this is the only boot reciever we need to send a broadcast to restart location tracking.
https://github.com/home-assistant/home-assistant-android/blob/master/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt#L141

app/src/main/AndroidManifest.xml
yoxjames: I should clarify. The LocationBroadcastReceiver appears to me to completely ignore these which is why I thought they could be deleted. It looks like it only processed BOOT_COMPLETED Honestly I am not sure if they are a good idea to include. 

**https://github.com/home-assistant/home-assistant-android/blob/f6618527d4f5f1233e7e0515f9816eb3ac282e5b/app/src/main/java/io/homeassistant/companion/android/background/LocationBroadcastReceiver.kt#L57**

yoxjames:
Pushed another commit to address comments. @JBassett , I have added the restart location tracking stuff into the Boot Receiver. I did have to make one change to `PermissionManager.restartLocationTracking`, I changed it so that it only takes `Context` and not activity. `sendBroadcast` doesn't require a fully fledged activity (it's actually implemented on Context rather than Activity) so this should be fine, however I wanted to ensure I highlighted it.

Also I had to revert just using `SensorWorker.start` in the `ChargingBroadcastReceiver` the reason is that `SensorWorker.start(..)` still entrusts Android to run the work at it's leisure which I assume to mean some time hopefully in 15 minutes. I noticed that unless I had the app in the foreground it would not instantly update the status. The only way to truly ensure that is to take the WorkManager stuff out and directly update the sensors as I had before.

I finally got on the Discord and would be happy to chat about any of this tomorrow!

app/src/main/java/io/homeassistant/companion/android/sensors/ChargingBroadcastReceiver.kt
JBassett: Did you try calling `SensorWorker.start(context)` instead?  That should be included in this PR because we don't want to duplicate this logic.

yoxjames: Yeah sorry see my comment I posted at the bottom of the PR. That wasn't working because it still entrusts Android to actually run the SensorWorker code. Therefore it wouldn't reliably immediately execute and could be delayed by up to 15 minutes which is the problem I had hoped to solve with this PR.

JBassett: Totally missed that.  Can we do `SensorWorker(context, WorkerParameters()).doWork()` instead?  If we can't do that then we should extract that code somewhere that we can call it independently so we don't duplicate it.

yoxjames: I fully agree with the sentiment. I originally didn't love that I was duplicating logic but didn't want to start ripping things apart for a first PR. Essentially I played it safe. However I will definitely look into a way of doing it. The best way, to me, feels like taking what SensorWorker and pulling that into a class that can be called elsewhere and composed into other sources of sensor updates besides a work manager. I'll go ahead and try something out.

yoxjames: Ok I added a new interface/impl `SensorUpdater` which contains that logic in one place. It is not inserted into the dagger graph. In order to do that I would need to make some significant changes that I felt were starting to get out of scope and I didn't want to cause additional delays for this functionality. I can definitely tackle this in a follow up PR where we can focus more exclusively on that. Let me know what you think.

app/src/main/java/io/homeassistant/companion/android/sensors/AllSensorsUpdaterImpl.kt
```diff
@@ -0,0 +1,52 @@
...
+        private const val TAG = "AllSensorsUpdaterImpl"
+    }
+
+    override suspend fun updateSensors() {
+        val sensorManagers = mutableListOf(
+            BatterySensorManager(),
+            NetworkSensorManager()
+        )
+
+        if (integrationUseCase.isBackgroundTrackingEnabled()) {
```
yoxjames: Noticed this change from `SensorWorker` and made sure to implement it here!

app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
```diff
@@ -19,6 +23,18 @@ class HomeAssistantApplication : Application(), GraphComponentAccessor {
 
         AndroidThreeTen.init(this)
         graph = Graph(this, 0)
+        // Start the sensor worker if they start the app. The only other place we start this ia Boot BroadcastReceiver
+        SensorWorker.start(this)
+
+        // This will cause the sensor to be updated every time the OS broadcasts that a cable was plugged/unplugged.
+        // This should be nearly instantaneous allowing automations to fire immediately when a phone is plugged
+        // in or unplugged.
+        registerReceiver(
```
yoxjames: Ended up moving this into `HomeAssistantApplication` rather than `SensorWorker` feels like it belongs here more.

app/src/main/java/io/homeassistant/companion/android/PresenterModule.kt
```diff
@@ -122,5 +123,8 @@ class PresenterModule {
 
         @Binds
         fun bindLockPresenter(presenter: LockPresenterImpl): LockPresenter
+
+        @Binds
+        fun bindSensorUpdater(sensorUpdater: AllSensorsUpdaterImpl): SensorUpdater
```
JBassett: I noticed you never actually inject this anywhere, if that's the case there is no reason to add it here.  If you want to use dagger make sure you always have it injected rather than create it every use.

yoxjames: Ugh yeah this was a relic from when I was putting it into the Dagger graph. The PR got a little large so I backed up from that. Will remove.

yoxjames: Done

Hypfer:
This is the most verbose PR i've ever seen 😄

Xitro01:
Is this something that could be implemented for iOS as well?

Issue #529, opened 2020-03-27
Expand available sensors to mirror iOS functionality more


jeroenseegers:
**Is your feature request related to a problem? Please describe.**
Not directly.

**Describe the solution you'd like**
The iOS companion app enables quite a lot of different sensors to be available when compared to the Android application. It'd be nice to have both companion apps publicize the same sensors.

**Describe alternatives you've considered, if any**
None.

**Additional context**
If compared to the iOS companion app, the following sensors are "missing" from the Android one which I don't believe should be too difficult to implement:
- [x] `sensor.battery_state` (added in #530)
  - This one's currently integrated into the `battery_level` sensor as a `is_charging` property.
- [ ] `sensor.sim1` and `sensor.sim2`
  - If I understand correctly, this information could be requested through the [`SubscriptionManager.getActiveSubscriptionInfoList`](https://developer.android.com/reference/android/telephony/SubscriptionManager.html#getActiveSubscriptionInfoList())
- [ ] `sensor.ssid`
  - This one's currently integrated into the `wifi_connection` sensor as the `state`.
- [ ] `sensor.bssid`
  - This one's currently integrated into the `wifi_connection` sensor as an attribute.
- [ ] `sensor.connection_type`
  - This one could be deducted from the status of the `wifi_connection` sensor, when not connected to a Wi-Fi network this sensor returns `<not connected>` as a state f.e.

I think that making the iOS and Android app align more on these kinds of things would be very nice!

Source of sensor types and availability: https://companion.home-assistant.io/docs/core/sensors

dshokouhi:
I would say maybe we should consider splitting up the requested features but it seems like you have started PR's to add these features :)

jeroenseegers:
We can still split up this task into separate ones per sensor, in that case we'll just edit this one to reflect only the changes to `battery_level` and `battery_state`?

This will also help keeping things a bit more clear in terms of progress I guess.

dshokouhi:
I think that we can leave it as is to avoid breaking changes for users, I think we may be good on this request now with all the new sensors yea?

PR #530, opened 2020-03-27
Add a seperate battery_state sensor

jeroenseegers:
As proposed in #529, this PR adds a `battery_state` sensor.

The current `BatterySensor` enables a single `battery_level` sensor which contains different battery data, example:
Screenshot 2020-03-27 at 14 30 21

With this change, there's now two sensors, `battery_level` and `battery_state`, similar to the sensors available when using the iOS application, example:
Screenshot 2020-03-27 at 14 32 46

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/BatterySensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/BatterySensorManager.kt
@@ -45,8 +45,20 @@ class BatterySensorManager : SensorManager {
         return retVal
     }
 
-    private fun getBatteryIcon(batteryStep: Int): String {
+    private fun getBatteryIcon(batteryStep: Int, isCharging: Boolean = false, chargerType: String? = null, chargingStatus: String? = null): String {
         var batteryIcon = "mdi:battery"
+
+        if (chargingStatus == "unknown") {
+            batteryIcon += "-unknown"
+
+            return batteryIcon
+        }
+
+        if (isCharging)
+            batteryIcon += "-charging"
+
+        if (chargerType == "wireless")
+            batteryIcon += "-wireless"
 
         batteryIcon += when (batteryStep) {
             0 -> "-outline"
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/BatterySensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/BatterySensorManager.kt
@@ -45,6 +45,18 @@ class BatterySensorManager : SensorManager {
         return retVal
     }
 
+    private fun getBatteryIcon(batteryStep: Int): String {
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @jeroenseegers, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant-android#530).

Thanks!

app/src/main/java/io/homeassistant/companion/android/sensors/BatterySensorManager.kt
```diff
@@ -28,60 +29,117 @@ class BatterySensorManager : SensorManager {
...
+
+        val percentage: Int = (level.toFloat() / scale.toFloat() * 100.0f).toInt()
+        val batteryStep: Int = percentage / 10
+
         return Sensor(
             "battery_level",
-            percent,
+            percentage,
+            "sensor",
+            getBatteryIcon(batteryStep),
```
JBassett: Why not display the fancy battery icon for both sensors?

app/src/main/java/io/homeassistant/companion/android/sensors/BatterySensorManager.kt
```diff
@@ -28,60 +29,117 @@ class BatterySensorManager : SensorManager {
...
+    private fun getBatteryLevelSensor(intent: Intent): Sensor<Any>? {
+        val level: Int = intent.getIntExtra(BatteryManager.EXTRA_LEVEL, -1)
+        val scale: Int = intent.getIntExtra(BatteryManager.EXTRA_SCALE, -1)
+
+        if (level == -1 || scale == -1) {
+            Log.e(TAG, "Issue getting battery level!")
+            return null
+        }
+
+        val percentage: Int = (level.toFloat() / scale.toFloat() * 100.0f).toInt()
```
JBassett: Extract out the building of the battery percentage so we don't have to duplicate it.

app/src/main/java/io/homeassistant/companion/android/sensors/BatterySensorManager.kt
jeroenseegers: I figured since this sensor is now purely for the battery level, the icon should also represent just that.

quthla:
It seems most people don't need this and for those who do, can't they just make a template sensor of the attribute and have the same effect?

dshokouhi:
@quthla this will actually help bring iOS and Android closer in terms of features.  iOS already splits up the 2 sensors for users. https://companion.home-assistant.io/docs/core/sensors/#battery-sensors

jeroenseegers:
Correct, I figured it'd be very nice from an architectural point of view to have the same sensors and their respective properties to be published from both the iOS and Android companion apps.

Issue #542, opened 2020-04-02
App in background keeps unused websocket connection open


balloob:
From https://github.com/home-assistant/frontend/issues/5328#issuecomment-607906917

**Home Assistant Android version:** 1.8

**Android version:**

**Phone model:** all?

**Home Assistant version:** 0.108

**Last working Home Assistant release (if known):** -

**Description of problem:**

It looks like when the app is going in the background, the websocket connection of the frontend remains open, but we no longer execute JavaScript to read the messages. This results in warnings in the backend because there are more than 512 pending messages.

If possible, on suspend, we should close the websocket connection and restore it when we're visible again.

**Traceback (if applicable):**

```

```

**Screenshot of problem:**

**Additional information:**

electrofloat:
That is exactly the same thing that is happening when using not the app but Chrome android "add to homescreen" (webapp) feature.

dshokouhi:
Would this be fixed by 0.112?

balloob:
Yep!

Issue #277, opened 2020-01-18
High (mobile) data usage


Wummeke:
**Home Assistant Android version:**
1.4.0(77)

**Android version:**
10

**Phone model:**
Google Pixel 3A

**Home Assistant version:**
0.104.2

**Last working Home Assistant release (if known):**

**Description of problem:**
I noticed that the app uses a lot of data when I do not actively use it. For example, my phone reports that the app has used over 1.5 GB of mobile data since 1-1. I am quite sure that I don't use the app very often when I am not connected to a WiFi network, so this high number surprises me. I suspect that it has something to do with a few camera integrations I use (I cannot think of something else, to be honest).

I thought of disabling background data, but then location tracking doesn't work anymore.

JBassett:
I've noticed that too, for me it happened when I opened a view of the camera and didn't close it before locking the phone.  We should look into freezing the frontend when the UI isn't being displayed.

dshokouhi:
It must be the camera not closing because when I opened my PiP bug I made sure to close the window. My data usage has been super low from the start.

![Screenshot_20200118-082312](https://user-images.githubusercontent.com/1634145/72666971-d7ce7e80-39cb-11ea-98a7-6a05aaea99e1.png)

![Screenshot_20200118-082107](https://user-images.githubusercontent.com/1634145/72666975-ddc45f80-39cb-11ea-9a68-03e95ecffa3e.png)

balloob:
Freezing the web UI should be the way to go. This sounds like it kept streaming the camera in the background

sverleysen:
Is it possible this issue is back again? My app used 7.45gb in a few days in the background

JBassett:
This will be fixed with this change in the frontend: https://github.com/home-assistant/frontend/pull/6149
I'm not sure what release that fix is targeting however.

dshokouhi:
Looks like it'll be in 0.112

dshokouhi:
Fixed in 0.112.x

fiLLLip:
I updated to 0.112.x when it came out. This is still a issue on the android app. The last 24 hours I have had a usage of 1.76GB in background (86MB in foreground). My solution for now is to restrict data usage in background.

DrBlokmeister:
I also suffer from the main issue, albeit a different app

bendikrb:
Same here... 

[...]

PR #545, opened 2020-04-03
Pause the webview by loading a null page, killing javascript/websocket.

JBassett:
Fixes: https://github.com/home-assistant/android/issues/542
Fixes: https://github.com/home-assistant/android/issues/277

This seems to be the only way I can actually get the websocket to be closed.  We basically navigate away when the application gets paused, and reload the app on resume.


balloob:
Is this something we can do on a timer? Say, 5 minutes? 

I also wonder if it might be better to built something into the frontend to help with this. Kill all camera feeds + websocket connection. Not sure if it would be possible.

andriej:
Especially that issue you've linked shows that some other websocket connections may cause it (not related to only android app).

JBassett:
We can delay 5 minutes.  I agree that a hook to the frontend would be the best solution to pause and resume the connection.  I'll add the timer to the PR at some point today.

JBassett:
One day, two months.... What is time anyway?

balloob:
@bramkragten is actually working on a PR for the frontend to implement https://github.com/home-assistant/frontend/issues/5438 / https://github.com/home-assistant/frontend/issues/5437

bramkragten:
The frontend now manages this, so I think this PR is not longer needed.

JBassett:
Closing in favor of the frontend fix

Issue #107, opened 2019-12-04
Add support for notification channels


dshokouhi:
I know we don't have notifications yet but it would be great if we can add support for notification channels once we do. Its important to differentiate notifications so the user can set the priority and sound.  For example you don't want your phone making the same sound for the alarm being triggered as it does for the laundry.

https://developer.android.com/reference/android/app/NotificationChannel

robbiet480:
My understanding is that notification channels are static and must be defined in AndroidManifest.xml which means we can't modify it during runtime. 

Firebase Cloud Messaging allows defining things like [sound](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#AndroidNotification) and [priority](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#androidmessagepriority) on a per notification basis.

quthla:
Not sure if they can be defined in the manifest but they can defnitely be set programmatically.

dshokouhi:
@quthla yup, my ring app does that as i add/remove ring devices.

robbiet480:
My bad, I misread the FCM docs!

RikP:
this will also support different messages being able to be send in do not disturb mode.
So same as critical messages on iphone.

LordBoos:
Even static notification channels would be good. Like 3 channels for priorities like "low", "normal, "critical" so we can customize the channels with different sounds and settings in the android.
I'd like to set up critical notifications for things like "flooded bathroom" that ignore do not disturb mode but at the same time I don't want to low priority notifications like "plant needs water" to wake me up at night.

dshokouhi:
Closed by #574 currently waiting on https://github.com/home-assistant/mobile-apps-fcm-push/pull/13 to be merged so the feature can work properly.  Once its working will get the docs updated.

PR #574, opened 2020-05-02
Implementing Android notification channels

CraftyKoala:
Hi,
I am new here.

Had a quick go on notification channels in android. The solution is quite simple. 
I am checking a cannel parameter in the payload data. 
```
{
  message: "something important",
  channel: "Important"
}
```

This allows users to define channels in their automation. Then they are able to configure them with different sounds and priorities on their phones.

Still testing it, tough. Sofar I just implemented it and send a few test messages to the android emulator from the firebase console. That however worked already.
I am also still working on a way to remove channels.
Anyway, wanted to let you know I am working on this and looking forward to your feedback.

KR

**Update 1:**
Removing notification channels is working aswell. It requires a message like this:
```
{
  message: "remove_channel",
  channel: "Important"
}
```

**Update 2:**
I now tested this in the emulator and on my phone with a custom rest client to send the messages.
Is there a way to rig a test instance of Homeassistant with my server key to do that aswell?
However, so far all tests are successful. And I observed the following on my Samsung:

- The device reports the deleted channels (categories).
- After removing a channel. If a new message arrives with the removed channel name, it is recreated with the same settings configured before deletion.
- Once the channel is created I can configure:
  - Style: Silent/Sound/Popup ...
  - Sound
  - Vibration
  - Light (not the color tough)
  - Badges
  - Lock Screen Appearance
  - Ignore disturb mode

_*Note this is Samsung specific, it will be different on other phones_

I will set this to review now, but still need to update the docs in the Companion Documentation.

Screenshot_1588434371

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -38,6 +38,11 @@ class MessagingService : FirebaseMessagingService() {
         const val TITLE = "title"
         const val MESSAGE = "message"
         const val IMAGE_URL = "image"
+
+        // special action constants
+        const val REQUEST_LOCATION_UPDATE = "request_location_update"
+        const val CLEAR_NOTIFICATION = "clear_notification"
+        const val REMOVE_CHANNEL = "remove_channel"
     }
 
     @Inject
@@ -66,14 +71,20 @@ class MessagingService : FirebaseMessagingService() {
         remoteMessage.data.let {
             Log.d(TAG, "Message data payload: " + remoteMessage.data)
 
-            if (it[MESSAGE] == "request_location_update") {
-                Log.d(TAG, "Request location update")
-                requestAccurateLocationUpdate()
-            } else if (it[MESSAGE] == "clear_notification" && !it["tag"].isNullOrBlank()) {
-                Log.d(TAG, "Clearing notification with tag: ${it["tag"]}")
-                clearNotification(it["tag"]!!)
-            } else {
-                mainScope.launch {
+            when {
+                it[MESSAGE] == REQUEST_LOCATION_UPDATE -> {
+                    Log.d(TAG, "Request location update")
+                    requestAccurateLocationUpdate()
+                }
+                it[MESSAGE] == CLEAR_NOTIFICATION && !it["tag"].isNullOrBlank() -> {
+                    Log.d(TAG, "Clearing notification with tag: ${it["tag"]}")
+                    clearNotification(it["tag"]!!)
+                }
+                it[MESSAGE] == REMOVE_CHANNEL && !it["channel"].isNullOrBlank() -> {
+                    Log.d(TAG, "Removing Notification channel ${it["tag"]}")
+                    removeNotificationChannel(it["channel"]!!)
+                }
+                else -> mainScope.launch {
                     Log.d(TAG, "Creating notification with following data: $it")
                     sendNotification(it)
                 }
@@ -94,6 +105,17 @@ class MessagingService : FirebaseMessagingService() {
         val messageId = tag.hashCode()
 
         notificationManager.cancel(tag, messageId)
+    }
+
+    private fun removeNotificationChannel(channelName: String) {
+        val notificationManager =
+            getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+
+        val channelID: String = createChannelID(channelName)
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            notificationManager.deleteNotificationChannel(channelID)
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @CraftyKoala, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;android#574).

Thanks!

dshokouhi:
This will close #107 I like the workflow you have proposed so far, makes sense to allow creation and deletion on the fly.

app/build.gradle
```diff
@@ -20,7 +20,7 @@ buildscript {
 }
 
 def vCode = "${System.env.VERSION_CODE ?: 1}".toInteger()
-def vName = "1.8.0-${vCode}"
+def vName = "1.9.0-${vCode}"
```
JBassett, line 23: When we go to release to production we bump the version number, no need to do it in this PR.

CraftyKoala: No Problem, I removed the last commit.

CraftyKoala:
When would be the right time to update the companion documentation ?

JBassett:
@CraftyKoala that's a better question for @dshokouhi I would guess we wait until we get it to prod.  With the 2 new features I think we probably have enough to push to prod again.  I'll merge and we can spend the day making sure we don't have any other crashes or horrible errors and we can promote.

dshokouhi:
@CraftyKoala and @JBassett  I have been updating the docs as we merge things actually

dshokouhi:
Trying to test the feature now.  Using the example service data provided (note using YAML in services tab under developer tools):

service: notify.mobile_app_pixel_4_xl

```
message: test
channel: alarm
```

Results in:

```
Traceback (most recent call last):
  File "/srv/homeassistant3.7/lib/python3.7/site-packages/homeassistant/components/websocket_api/commands.py", line 130, in handle_call_service
    connection.context(msg),
  File "/srv/homeassistant3.7/lib/python3.7/site-packages/homeassistant/core.py", line 1209, in async_call
    processed_data = handler.schema(service_data)
  File "/srv/homeassistant3.7/lib/python3.7/site-packages/voluptuous/schema_builder.py", line 272, in __call__
    return self._compiled([], data)
  File "/srv/homeassistant3.7/lib/python3.7/site-packages/voluptuous/schema_builder.py", line 594, in validate_dict
    return base_validate(path, iteritems(data), out)
  File "/srv/homeassistant3.7/lib/python3.7/site-packages/voluptuous/schema_builder.py", line 432, in validate_mapping
    raise er.MultipleInvalid(errors)
voluptuous.error.MultipleInvalid: extra keys not allowed @ data['channel']
```

When I use the service data like:

```
message: test
data:
  channel: "Alarm"
```

the notification shows up but no channel gets created, I still see the default one no HA log error either.  Did I miss something here?  Do we need our FCM server updated to accept the new attribute?

JBassett:
So my bad I didn't notice the difference until now, our service expects `channelId` but the app expects `channel`.  @CraftyKoala you mind making a PR to use `channelId` instead?  That is technically what the FCM documentation uses.

CraftyKoala:
Actually this was intentional as we do not use the FCM framework. FCM is just a data transport.
People should not be confused with their API, since we do not support what is documented there. 

If we would like to use FCM and enable the documented features we need to implement changes in the mobile_app component in the home assistant core.

What the custom implementation expects here is the visible name displayed in the notification settings, the channelId is generated from that.

Should we still change it? (I don't think its a good idea.)

dshokouhi:
@JBassett one other thought on this.....should `remove_channel` be added to the list of notification  messages that are exempt from the rate limit?

smoke007:
I found the new Notification Channels feature documented at https://companion.home-assistant.io/docs/notifications/notifications-basic#notification-channels and tried to implement this feature in HA within my notification.  However, I can't yet create channels on my phone.  After much searching, I found this post and also see that this feature was just merged a few days ago with https://github.com/home-assistant/android/commit/f2f07b025b0ed07841ff7b3e27449a9a2b901037.

My guess is that us users need to wait for a new Home Assistant Companion app to be pushed our to our phones, newer than version 1.8.0-144, correct, or maybe enroll in beta?  If it is true that we need to wait for a new version of the app to be pushed out to our phones, should that be noted on the documents website that this feature isn't rolled out to PROD yet?

Thank you all for all your great work and new features!!  :)

dshokouhi:
@smoke007 the companion docs stay up to date with the beta builds so the docs are easier to manage.  Any user can easily sign up with the beta builds.  Basically as soon as something merges here and is live on the play store beta we try to update the docs.  Feel free to join the beta to test things out!

FigJam23:
I'm unsure on how this works what would a basic .yaml notification look like to push custom sounds to a android 
Script below works for iOS but on Android / pixel phone I just get notifications not custom sounds.

`
- alias: 'DriveWay Alarm Notify'
  trigger:
    - platform: state
      entity_id: 'binary_sensor.driveway_line_crossing'
      to: 'on'
  action:
    service: notify.all_devices
    data:
      message: 'SUP DriveWay Alarm'
      data:
        push:
          sound: US-EN-Morgan-Freeman-Someone-Is-Arriving.wav
`

Issue #559, opened 2020-04-18
Shortcut to Panels (Sidebar)


timmo001:
**Describe the solution you'd like**

Pull the panels from the WebSocket to add [shortcuts](https://developer.android.com/guide/topics/ui/shortcuts/creating-shortcuts#dynamic) to the app.

**Additional context**

Useful for going straight to panels such as other dashboards or ingress views

dshokouhi:
So how is this supposed to work?  When I go to App Configuration > Shortcuts I get nothing.

![image](https://user-images.githubusercontent.com/1634145/80986421-63d2ad80-8de5-11ea-875e-bb04ba257a2b.png)

I even tried to create a new lovelace dashboard but nothing gets populated.

dshokouhi:
Ah looks like we need to wait for the PR in Core to merge?

dshokouhi:
![image](https://user-images.githubusercontent.com/1634145/80986646-b4e2a180-8de5-11ea-92ee-4a735e6b5844.png)

Should we have waited to merge this PR?

JBassett:
We are dependent on the core PR: https://github.com/home-assistant/core/pull/34445.  We might want to hide the settings option though if the get_panels call fails. 
@timmo001 thoughts?

JBassett:
@dshokouhi We need the app to work even if the Core isn't up to date.

timmo001:
Hmm, yeah will be a while before an update is out there for core. Will need to disable this feature until that webhook is available.

timmo001:
Aha looks like you're handling this now

https://github.com/home-assistant/android/pull/576/files

dshokouhi:
Yea the app is working without crashing, its just the screen turns up blank and users will probably open bugs about the error :) .  Hopefully they can merge that PR for 0.110 :)

timmo001:
Well I guess we could cherry-pick the next release to disclude these commits if we release before core

dshokouhi:
We should probably also consider when a user has this version of the app but does not have the expected HA version and will get an error.  Are we able to do something like that?

JBassett:
@timmo001  We should just make the preference not visible if the get_panels call fails.

PR #578, opened 2020-05-04
Disable setting if get_panels is not available or fails

timmo001:
Checks if panels are empty. This can be removed after core PR https://github.com/home-assistant/core/pull/34445 is merged and released to stable

Issue Ref https://github.com/home-assistant/android/issues/559#issuecomment-623579395


timmo001:
OK so core change is not going ahead, but this can still go in for now and be replaced/removed with the WebSocket webhook replacement

PR #579, opened 2020-05-04
Add push notification instructions

JBassett:


CraftyKoala:
I think I got quite far, but I don't know where to find the right URL to use in the APP ?

PR #583, opened 2020-05-08
Stop calling an endpoint that will never come.

JBassett:
Until @timmo001 updates this call to use the webhook, we are going to stop making this request.


timmo001:
Just so you know, I did start setting up the use of the websocket but it may be a little past my knowledge in terms of how the app is setup with kotlin, dagger and the like.

Might be best someone else take this up since I'm back at work after the bank holiday so don't have much time to go into depth on this and I imagine it will take some time to setup

timmo001:
This is a good interim solution though since it will stop trying to call the webhook and return an empty array which is what is happening now.

There are quite a few endpoints that will need to be replaced such as get_config etc. with the websocket

Issue #605, opened 2020-05-28
ClickAction with relative Lovelace Link only working once after fresh App install


dickesW:
**Home Assistant Android version:**
1.9.0-167
**Android version:**
9 / 10
**Phone model:**
Honor View 10 / Huawei P30 Pro
**Home Assistant version:**
0.110.3
**Last working Home Assistant release (if known):**
Never
**Description of problem:**
When clicking on the notification for the first time after a fresh install:
- the requested Lovelace tab defined in `clickAction` is opened (e.g. `'/lovelace/test'`)

When clicking on the notification every other time with a different `clickAction` Lovelace URI:
- always the `clickAction` Lovelace URI sent with the first notification after a fresh install is opened.
- this does not change until a fresh install of the app (not helping: stopping / starting / closing the app, reboot of phone)
- web URIs (e.g. `'https://www.google.com'`) are opened correctly
- sending no `clickAction` also opens the first sent `clickAction` Lovelace URI

Clicking on the same URI in the "Actionable Notification" (see example below) always opens the desired Lovelace tab.

Used notification:
```yaml
title: 'Motion Detected in Backyard'
message: 'Someone might be in the backyard.'
data:
  clickAction: '/lovelace/outside'
  actions:
  - action: 'URI'
    title: 'Open Url'
    uri: '/lovelace/outside'
```

**Traceback (if applicable):**

```

```

**Screenshot of problem:**

**Additional information:**
Might relate to https://github.com/home-assistant/android/issues/486

Issue #593, opened 2020-05-12
Request to rename default notification channel


denniswham:
**Home Assistant Android version:**
1.9.0-167

**Android version:**
Android Version 10

**Phone model:**
Samsung Galaxy S9 (SM-G960F)

**Home Assistant version:**
0.109.6

**Last working Home Assistant release (if known):**
N/A

**Description of problem:**
Is it possible to rename "Default Channel' to 'General'? To my opinion "Default Channel" is too techy and not "end-user language".

**Traceback (if applicable):**

```

```

**Screenshot of problem:**
![image](https://user-images.githubusercontent.com/43110823/81660674-58adfc00-943b-11ea-8b5e-57cb8a75feca.png)

**Additional information:**

balloob:
I agree.

Issue #486, opened 2020-03-02
Application doesn't open the right lovelace view from the relative URL in clickAction


bieniu:
**Home Assistant Android version:** 1.8.0-144

**Android version:** 10

**Phone model:** OnePlus 6

**Home Assistant version:** 0.106.2

**Last working Home Assistant release (if known):** None

**Description of problem:**
The application doesn't open the right lovelace view from the relative URL in `clickAction`. First lovelace view is always opened. This problem doesn't occur if I use action button with URL in `notify` service.

Notify service configuration:
```yaml
message: TEST
data:
  clickAction: /lovelace/energy
```
My lovelace config has energy view:
```yaml
- id: energy
    icon: mdi:power-plug
    title: "Energia"
    path: energy
    panel: true
    cards:
      - ...
      - ...
```
`http` configuration:
```
http:
  ip_ban_enabled: true
  login_attempts_threshold: 5
  use_x_forwarded_for: true
  trusted_proxies: "127.0.0.1"
  base_url: "https://my_domain.org"
```

**Video of problem:**

[![Watch the video](https://i.ytimg.com/vi/UI60fBQTnHQ/hqdefault.jpg?sqp=-oaymwEjCNACELwBSFryq4qpAxUIARUAAAAAGAElAADIQj0AgKJDeAE=&rs=AOn4CLDUSi2MfaJmnpIJkwAh8laMM_bQMQ)](https://youtu.be/UI60fBQTnHQ)

kmlucy:
This isn't fixed for me. Should I open a new issue, or should this one be reopened?

Pixel 3a w/ app version 1.8.0-159 and Home Assistant version 0.107.7.

dshokouhi:
@kmlucy this is an older issue, please try to [start fresh with the app](https://companion.home-assistant.io/docs/troubleshooting/faqs#starting-fresh-with-the-android-app) if you still face an issue please open a new issue. Make sure to fill out the issue template and provide the YAML for the notification service call, the lovelace view YAML as well as the `http` configuration.

McGiverGim:
Maybe this needs some beta version of the Home Assistant core too? I'm facing the same problem, and I can see this in the logs:
```log
2020-05-07 11:58:05 ERROR (MainThread) [homeassistant.components.mobile_app.webhook] Received invalid webhook type: get_panels
```

andriej:
Same issue here

dshokouhi:
[...]

Issue #444, opened 2020-02-18
Notifications do not come through when phone/app asleep


xbmcnut:
**Home Assistant Android version:**
1.7.2-129
**Android version:**
10
**Phone model:**
Nokia 7.1
**Home Assistant version:**
0.105.4
**Last working Home Assistant release (if known):**

**Description of problem:**
Firstly, thank you so much for your time and effort in making the Android version. I’ve been using it from day one and it’s just got better and better. Great job!
I do however have an issue with background notifications. It seems that I don’t get notifications when the app is asleep even though I have battery optimisation turned off. For reference, I also use the Synology Chat application for Home Assistant notifications and when testing persistent notifications this morning, the phone/watch combo sent me my doorbell notification from Synology Chat but the Home Assistant app did not until I opened the app then the notification came though straight away. Once awake, subsequent notifications are instant.

If I leave the phone off for an undetermined amount of time, notifications again stop until the phone/app is woken up.

I did have this issue with the Synology app too when I first configured it but by adding the Chat app to the battery optimisation bypass list fixed this but it does not make a difference with the Home Assistant app.

**Traceback (if applicable):**

```

```

**Screenshot of problem:**

**Additional information:**
May be related to #420?

dshokouhi:
Did you include `ttl` and `priority` in your service call? That is required for notifications to wake up a device in doze mode.  You also want to make sure that background data is enabled and that battery optimizations are turned off for the app. 

https://companion.home-assistant.io/docs/notifications/critical-notifications

xbmcnut:
Thank you pointing that out! No I did not. Updating now to test. Thank you!

xbmcnut:
So it this how I'd format it? What does the TTL flag do?
```
    action:
      - delay: 00:00:02
      - service: notify.mobile_app_nokia_7_1_app
        data:
          message: "The doorbell was pushed at {{ states.sensor.time.state }}"
          # title: "If not stated, it will say Home Assistant anyway"
          data:
            ttl: 0
            priority: high
            data:
              image: http://10.0.1.200:8123/local/images/front_porch.jpg
              actions:
                - action: "unlock_door" # The key you are sending for the event
                  title: "Unlock" # The button title            
                - action: "foyer_on" # The key you are sending for the event
                  title: "Foyer Lamp" # The button title
                - action: "siren_on" # The key you are sending for the event
                  title: "Police Siren" # The button title  
```

dshokouhi:
[...]

PR #612, opened 2020-06-07
Notification improvements (HTML formatting, Grouping, Subject, TTL and priority fix)

hasenbolle:
Hi,

this pull request needs also this pull request:
https://github.com/home-assistant/mobile-apps-fcm-push/pull/15

the current notifications are missing some useful features. Also TTL and priority doesn't seem to work right now.

This pull request fixes TTL, the priority property and adds following features:
- Use a icon for notification without setting a image for the notification (Property: icon).
- Grouping of notification (Property: group).
- Property message supports HTML formatting
- Setting of a subject text for a long message (Property: subject)

About the priority:
Since Android Oreo, the priority setting is deprecated, so the priority setting is only set when system is < Oreo. For System >= Oreo, the priority property is used for the channel importance. If a new channel is created the priority property is used for setting the importance of the channel.

Grouping closed:

Grouping opened:

```
message: "Garage has been opened! Do you want to close the garage?"
title: "Garage"
data:
  actions: 
    - action: "yes"
      title: "Yes"
    - action: "no"
      title: "No"
  group: "Garage"

message: "Garage was closed!"
title: "Garage"
data:
  group: "Garage"
```

Subject for long text (closed):

Subject for long text (opened):

```
message: "Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet."
title: "Long text"
data:
  subject: "Subject for long text"
```

HTML formatting:

```
message: 'This is a <b><span style="color: red">HTML</span></b> <i>text</i><br><br>This is a text after a new line'
title: "Cool HTML formatting"
```

Icon without image:

```
message: 'Mail arrived!'
title: "Mail"
data:
  icon: "https://cdn.pixabay.com/photo/2017/06/28/11/22/letters-2450337_960_720.png"
```

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -13,7 +13,9 @@ import android.net.Uri
 import android.os.Build
 import android.util.Log
 import androidx.core.app.NotificationCompat
+import androidx.core.app.NotificationManagerCompat
 import androidx.core.content.ContextCompat
+import androidx.core.text.HtmlCompat
 import com.google.firebase.messaging.FirebaseMessagingService
 import com.google.firebase.messaging.RemoteMessage
 import io.homeassistant.companion.android.R
@@ -37,7 +39,11 @@ class MessagingService : FirebaseMessagingService() {
         const val TAG = "MessagingService"
         const val TITLE = "title"
         const val MESSAGE = "message"
+        const val SUBJECT = "subject"
+        const val PRIORITY = "priority"
         const val IMAGE_URL = "image"
+        const val TTL = "ttl"
+        const val ICON = "icon"
 
         // special action constants
         const val REQUEST_LOCATION_UPDATE = "request_location_update"
@@ -125,19 +131,26 @@ class MessagingService : FirebaseMessagingService() {
     private suspend fun sendNotification(data: Map<String, String>) {
 
         val tag = data["tag"]
+        val group = data["group"]
         val messageId = tag?.hashCode() ?: System.currentTimeMillis().toInt()
+        val groupId = group?.hashCode() ?: 0
 
         val pendingIntent = handleIntent(data)
 
-        val notificationManager =
-            getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+        val notificationManagerCompat = NotificationManagerCompat.from(this)
 
-        val channelId = handleChannel(notificationManager, data)
+        val channelId = handleChannel(notificationManagerCompat, data)
 
         val notificationBuilder = NotificationCompat.Builder(this, channelId)
             .setContentIntent(pendingIntent)
             .setSmallIcon(R.drawable.ic_stat_ic_notification)
             .setSound(RingtoneManager.getDefaultUri(RingtoneManager.TYPE_NOTIFICATION))
+
+        handleLargeIcon(notificationBuilder, data)
+
+        handleGroup(notificationBuilder, data)
+
+        handleTtl(notificationBuilder, data)
 
         handleColor(notificationBuilder, data)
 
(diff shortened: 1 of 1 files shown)


hasenbolle:
Just read this:
https://github.com/home-assistant/android/issues/444
and this: 
https://firebase.google.com/docs/cloud-messaging/concept-options
about TTL and priority.

Seems that i totally misunderstood the meaning of TTL and priority.

For me TTL means the lifetime of the notification visibility and priority means for me how much intrusive the notification is (ex. high means a head up popup, low means just a symbol in the status bar).

But actually TTL means how long a message is hold on the server and priority means how fast they are delivered.

So than i must take back my statement that TTL and priority isn't working.

But i think many user have the same understanding of TTL and priority (especially users with no technical background). 

My suggestion would be to use new properties for TTL for notification visibility (ex. "notificationTtl" and "importance").

I will commit a change this afternoon.

hasenbolle:
Added #593 
Added #422 
Added #357 
Added #403 

Added new features: 

- Default channel is renamed to General
- Property **ledColor** can set the color of the notification led. Possible values are the same as for property color. eg '#2DF56D' # or 'red'. 
- Property **vibrationPattern** can set the vibration pattern on incoming notifications. Possible values are a list of numbers. eg. "100, 1000, 100, 1000, 100" etc.. The pattern spec. is "off time, on time, off time, on time, off time" etc.
- Property **persistent** can set the notification as ongoing/persistent. I can't be swiped away by the user. But it can be cleared by the sending the message "clear_notification". Because of that, the **tag** property is necessary to set alongside with the **persistent** property.
- Property **importance** is now used for setting the channel importance (Previous it was the **priority** property)
- Property **timeout** is now used for setting the timeout value of a notification in seconds (Previous it was the **ttl** property)

manuelo1983:
Hello, 
thank you very much for the many great features. I'm just about to switch to the app and actually miss exactly these things. It would be great if all these things would come with the next release. It would help me a lot and I think a lot of others would too.
Thanks a lot

hasenbolle:
Committed a fix for #605 and #486

balloob:
(I can't review code, don't know Kotlin)

This is great ! 

Let's not add any more fixes to this PR since it's already quite big. That way it will be easier to review.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -182,17 +238,112 @@ class MessagingService : FirebaseMessagingService() {
...
+        builder: NotificationCompat.Builder,
+        data: Map<String, String>
+    ) {
+        val ledColor = data[LED_COLOR]
+        if (!ledColor.isNullOrBlank()) {
+            builder.setLights(parseColor(ledColor, R.color.colorPrimary), 3000, 3000)
+        }
+    }
+
+    private fun handleLegacyVibrationPattern(
```
balloob, line 266: So I think that we process vibrations like this right now because it aligns both iOS and Android notifications.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -37,7 +39,14 @@ class MessagingService : FirebaseMessagingService() {
         const val TAG = "MessagingService"
         const val TITLE = "title"
         const val MESSAGE = "message"
+        const val SUBJECT = "subject"
+        const val IMPORTANCE = "importance"
+        const val TIMEOUT = "timeout"
         const val IMAGE_URL = "image"
+        const val ICON = "icon"
```
balloob, line 46: I think that `icon` might be a confusing option name for users, since the frontend has an icon set and here it means a url. 

I do understand that it's confusing, as we also already have `IMAGE_URL`. 

Maybe we should update it to be `ICON_URL` so at least it's clear it's a url ?

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
JBassett: It's android legacy not pattern legacy.  Since Android Oreo a channel is needed and that's where you set the vibrate pattern instead of pre Oreo's on the notification itself.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
JBassett: I agree with @balloob, once this is fixed I think the PR looks good!

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
hasenbolle: So this should be ok i guess?

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
hasenbolle: Yes i agree also. But the image property is also called ``image`` without the ``_url``. Maybe the property should also be renamed to ``image_url`` because the property also takes only URLs? Right now just the var is called ``image_url``

hasenbolle:
Just one suggestion that came in my mind:
The vibration pattern is right now passed as a string "1000, 100, 1000, 1000". Maybe it should passed as an array? I see that in Firebase the vibration pattern is passed as an array: https://firebase.google.com/docs/reference/admin/node/admin.messaging.AndroidNotification#vibratetimingsmillis

Instead of:
```yaml
title: "Test Title"
  message: "Test Message"
  data:
     vibrationPattern: "1000, 100, 1000, 1000"
```
It should be:
```yaml
title: "Test Title"
  message: "Test Message"
  data:
     vibrationPattern: 
          - 1000
          - 100
          - 1000
          - 100
```

What are your opinions ?

hasenbolle:
Yes it's a boolean value (false/true). Tested in anyway. For me it's working 👍

dshokouhi:
@hasenbolle ah a boolean value, got it :) working well now thanks for the quick confirmation!

PR #627, opened 2020-06-14
Fix notification groups

chriss158:
First to avoid confusion: I did a name change from hasenbolle to chriss158.

I did some additional tests on notification grouping and found some issues:

- The group of the notifications is left empty, if the last notification of the group is deleted with the `clear_notification` message.
- If a notification with a tag is open and if a new notification with a group which is named equally to the tag of the open notification, grouping/tag doesn't work.

This pull request fixes those issues.


homeassistant:
Hi @chriss158, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;android#627).

Thanks!

chriss158:
There is still an issue with left empty groups. 
If you send a notification  with a group and actions and then click on a button, the group will be left.

I try to fix this today.

Issue #568, opened 2020-04-26
Registration allows setting of Device Name


Nik71git:
**Home Assistant Android version:** 1.8.0-144

**Android version:** 9 PKQ1.180904.001

**Phone model:** Xiaomi Redmi Note 5

**Home Assistant version:** 0.108.9

**Last working Home Assistant release (if known):**

**Description of problem:**Hello. still, with app android release 1.8.0-144 and Home Assistant release 0.108.9, is necessary to rename device whenever you logout from app and login again. People like me and my family who own two or more identical devices (Xiaomi Redmi note 5) need to do that otherwyse automations, script and tracking not working fine anymore since you rename in different ways every device and restart HA. Logout is a security issue since in case of stolen device it has login already done and can take control of devices such alarm and gate opening. The problem is the app and not HA in my humble opinion, it would be nice if device are registered by their integrations not by name but by mac address for instance...

**Traceback (if applicable):**

```

```

**Screenshot of problem:**

**Additional information:**

dshokouhi:
@Nik71git in the android app under app configuration there is a lock option which will require either fingerprint, face unlock or pin/password.  I think this will solve your issue of having to log out each time right?

rradar:
This title is the best :100:

Nik71git:
> This title is the best 💯

Too long? ;-)

dshokouhi:
Closing as duplicate of #385

Issue #548, opened 2020-04-09
Cookies support for html pages in webpage card


radokk:
**Home Assistant Android version:**
1.8.0-144
**Android version:**
10
**Phone model:**
Samsung Note 10
**Home Assistant version:**
0.108.1
**Last working Home Assistant release (if known):**
N/A
**Description of problem:**
Hello, I added simple webpage card to lovelace (it's public transport time table). It's ok on my laptop, but in Companion App (android)  it's telling me that I need to enable cookies. Page loaded, but I have it covered with that cookie notification.  
When I signed to HA via chrome browser on mobile phone, same webcard asked me if I want to accept cookies. Once accepted, webcard loaded page without any notification properly.

Is it possible to add to Companion support for cookies?  

**Traceback (if applicable):**

```

```

**Screenshot of problem:**
Lovelace card in Companion with cookies notification:

![image](https://user-images.githubusercontent.com/60854491/78921498-4a17a200-7a95-11ea-90fe-03c06dcc1afd.png)

Notification in black: 
![image](https://user-images.githubusercontent.com/60854491/78920841-50594e80-7a94-11ea-9963-d144f9cfe6aa.png)
Translation: Enable cookies in browser to get fully functional page. 

Lovelace card in Android chrome web on my mobile phone once cookies were accepted. 
So in browser, it asked me if I want to accept cookies, but in app is telling me cookies support is not enabled. 

![image](https://user-images.githubusercontent.com/60854491/78921567-6287bc80-7a95-11ea-9a28-250afe01a2d3.png)

**Additional information:**
Pages I added to lovelace: 
1) https://bit.ly/2RlAvuw
2) https://bit.ly/39XQ8i6

N1c093:
Would be a great feature :)
I need the cookies to save my authentication credentials for an webpage in the iframe.

smithbill17:
I appreciate this issue is closed, but my experience using Home Assistant 0.115.1 freshly installed on Raspberry Pi 4b, indicates that the issue is still present.

I'm using Chrome 85.0.4183.101 installed on Android 10. I am not using Home Assistant with SSL.

With a Web Page card for  http://www.itv.com/news  it constantly requests allowing Cookies.  I've tried enabling the Chrome Flags for #same-site-by-default-cookies, #improved-cookie-controls & #improved-cookie-controls-for-third-party-cookie-blocking as well as switching Chrome to allow all 3rd Party cookies. Nothing works.

[...]

PR #629, opened 2020-06-16
Add CookieManager

neopilou:
CookieManager added to accept cookies in Webview (Fix issue #568 

Tested with "http://www.html-kit.com/tools/cookietester/" in iFrame card : 

Screenshot_20200616-103012_Home Assistant

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -298,7 +298,7 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
 
         val cookieManager = CookieManager.getInstance()
         cookieManager.setAcceptCookie(true)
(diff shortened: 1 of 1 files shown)


JBassett:
I don't think this will fix the linked issue, I think you ment: https://github.com/home-assistant/android/issues/548?

neopilou:
Yes, it's #548, not #568... (it's bug between chair and keyboard... ;-) )

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -295,6 +296,10 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
             }, "externalApp")
         }
 
+        val cookieManager = CookieManager.getInstance()
+        cookieManager.setAcceptCookie(true)
+        CookieManager.getInstance().setAcceptThirdPartyCookies(webView, true)
```
JBassett, line 301: Why can't this just be `cookieManager.setAcceptThirdPartyCookies(webView, true)`?

PR #633, opened 2020-06-17
Authentication Dialog Enhancement

neopilou:
- Add "Remember" checkbox
- Add button for password visibility

Screenshot_20200617-160827_Home Assistant


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -459,22 +468,71 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
...
+                    }
+                }
+        }
+
+        if (!autoAuth || authError) {
+            AlertDialog.Builder(this)
+                .setView(dialogLayout)
+                .setPositiveButton(android.R.string.ok) { _, _ ->
+                    if (username.text.toString() != "" && password.text.toString() != "") {
+                        val httpAuth = host + "|" + username.text.toString() + "|" + password.text.toString()
```
JBassett, line 511: This feels like a gigantic hack for persisting this information.  We should consider persisting this in a full DB rather than 
a Shared Preference.

data/src/main/java/io/homeassistant/companion/android/data/authentication/AuthenticationRepositoryImpl.kt
```diff
@@ -148,4 +149,18 @@ class AuthenticationRepositoryImpl @Inject constructor(
...
         return localStorage.getBoolean(PREF_BIOMETRIC_ENABLED)
     }
+
+    override suspend fun getHttpAuthList(): Set<String> {
+        val httpAuthList: Set<String>
+        if (localStorage.getStringSet(PREF_HTTP_AUTH_LIST) != null)
+            httpAuthList = localStorage.getStringSet(PREF_HTTP_AUTH_LIST)!!
+        else
+            httpAuthList = emptySet<String>()
+        return httpAuthList
```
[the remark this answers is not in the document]
neopilou: HttpAuthList functions have been removed from shared preference since use of a full DB

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
neopilou: Persisting data is now save in SQLite database, and no longer in Shared Preference.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -459,22 +469,64 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
...
         val dialogLayout = inflater.inflate(R.layout.dialog_authentication, null)
         val username = dialogLayout.findViewById<EditText>(R.id.username)
         val password = dialogLayout.findViewById<EditText>(R.id.password)
-
-        AlertDialog.Builder(this)
-            .setTitle(R.string.auth_request)
-            .setView(dialogLayout)
-            .setPositiveButton(android.R.string.ok) { _, _ ->
-                handler.proceed(username.text.toString(), password.text.toString())
+        val remember = dialogLayout.findViewById<CheckBox>(R.id.checkBox)
```
balloob, line 477: Why is there a remember me checkbox? Isn't that the same as disabling auth completely?

neopilou: I assume this is not the same thing, a user can access a site or a resource in different ways, this application for example, so he will not necessarily want to disable authentication. On the other hand, if the application is already locked (by credential or biometric), he can use this "Remember" function so as not to have to re-authenticate himself on his different sites. It's a bit like a SSO.

app/src/main/res/layout/dialog_authentication.xml
```diff
@@ -1,20 +1,67 @@
 <?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
neopilou: - Left justified checkbox
- Used alert dialog title
- Added padding around user and password

Screenshot_20200619-173159_Home Assistant

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
balloob: That makes no sense to me. You either have biometric authentication turned on, and so you are prompted. If you don't want it, you disable it. 

Let's remove this option.

balloob:
I'm actually completely confused here why we ever have username/password? What is going on… I understand fingerprint, maybe pin as fallback. We should not add a user/pass system ON TOP of Home Assistant which has its own. That is confusing.

neopilou:
We not add a user/pass system on top of Home Assistant which has its own. It's for basic HTTP Authentication Request, for embedded sites in IFrame Card, like Blue Iris, Tautulli, Kodi and others

balloob:
Oh my bad. Ok, that is fine I guess. Can we make the dialog header mention that it is HTTP Auth ? We should also make it look like material design https://material.io/components/dialogs

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
klejejs: I am sorry for intervention, but as we have now confirmed that the authentication dialog is for Basic HTTP auth, shouldn't we keep remember me button? To me it really does make sense to have it. You wouldn't want to login to iFrame every time you open the app.

neopilou:
Replaced title "Authentication Requested" by "Basic HTTP Authentication Requested" :

Screenshot_20200622-105933_Home Assistant

balloob:
Let's drop the `:` from the header.

And our primary color is blue, not purple. `#03a9f4`

balloob:
Let's also make it less wide.

Preferably we make it look exactly like the Chrome on Android dialog. Test it out on Chrome on http://httpbin.org/basic-auth/user/passwd

The "Connection not private" string is added if not over https.

image

JBassett:
@balloob Do we want to be consistent with chrome or the OS?  The [Alert Dialog](https://developer.android.com/reference/android/app/AlertDialog) we are using is the standard for the OS and handles the title and action buttons.  I don't know if we want to implement a completely custom dialog to try to mirror another app.

balloob:
Oh, yes, you're right. Let's use OS level dialog.

neopilou:
Added Dialog .setMessage() (like Chrome) with base URL and "Connection not private" string if not over https : 

Screenshot_20200623-105744_Home Assistant

balloob:
Nice! Can we reduce the width? This seems very wide. Maybe a max width of 300px?

neopilou:
Seted Dialog maximum width at 300dp :

Screenshot_20200624-083932_Home Assistant

balloob:
Can we drop the `:` from the header.

neopilou:
Look in String.xml, the `:` is deleted, but seems not updated in app...

app/src/main/java/io/homeassistant/companion/android/database/DataBaseHelper.kt
```diff
@@ -0,0 +1,87 @@
...
+
+private const val DATABASE_VERSION = 1
+private const val DATABASE_NAME = "HOME_ASSISTANT.db"
+private const val TABLE_NAME = "Authentication_List"
+private const val COLUMN_HOST = "host"
+private const val COLUMN_USERNAME = "username"
+private const val COLUMN_PASSWORD = "password"
+
+class DataBaseHandler(context: Context, factory: SQLiteDatabase.CursorFactory?) :
+    SQLiteOpenHelper(context, DATABASE_NAME, factory, DATABASE_VERSION) {
```
JBassett, line 18: We are going to want to use [Room](https://developer.android.com/training/data-storage/room) long term rather than raw SQL, if you are comfortable taking a first stab setting that up go for it.  If not we can merge this when ready and I can change it over before we promote it.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -296,10 +314,6 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
             }, "externalApp")
         }
 
-        val cookieManager = CookieManager.getInstance()
-        cookieManager.setAcceptCookie(true)
-        cookieManager.setAcceptThirdPartyCookies(webView, true)
-
```
JBassett, line 302: Why are we removing this?

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
JBassett: @balloob this is for saving external authentication a panel or iFrame that requires basic auth.  This shouldn't have any bearing on the authentication to Home Assistant itself.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
neopilou: RE added

app/src/main/java/io/homeassistant/companion/android/database/DataBaseHelper.kt
neopilou: Replaced SQLite by Room

Issue #528, opened 2020-03-26
Better error message when Home Assistant instance has expired SSL Cert


khoberg:
**Is your feature request related to a problem? Please describe.**
Home Assistant Android doesn't give a good error message when the Home Assistant instance it's connecting to has an invalid SSL certificate. It gives a generic 'Unable to connect to home assistant' message, which didn't help trace back the cause of that. It was also opening a browser to the login page which was working fine in my default browser and I was able to log in. My default browser must have cached things or didn't recheck date on cert. I finally figured out the issue when switching my default browser on Android and seeing the cert issue.

It also got into a state when I hit refresh and it opened the browser where if I went back to the app, it would get stuck refreshing forever (didn't actually wait for a timeout, so might have been one) with no way to cancel other than killing the app and reopening

**Describe the solution you'd like**
Have a clearer error message for why connection failed to help troubleshoot the issue.

**Describe alternatives you've considered, if any**

**Additional context**

dshokouhi:
Fixed in #637

PR #637, opened 2020-06-18
ShowError Enhancement

neopilou:
- Fix issue #528
- WebView.ShowError() now show detailed SSL error
- WebView.ShowError() now show error description of onReceivedError()

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -32,6 +32,13 @@ integration enabled on your home assistant instance.</string>
   <string name="error_onboarding_connection_failed">Unable to connect to Home Assistant.</string>
   <string name="error_connection_failed">Unable to connect to Home Assistant.</string>
   <string name="webview_error">There was an error loading Home Assistant, please review the connection settings and try again.</string>
+  <string name="webview_error_description">Encountered error :</string>
+  <string name="webview_error_SSL_DATE_INVALID">The date of the Home Assistant certificate is invalid, please review the Home Assistant certificate or the connection settings and try again.</string>
+  <string name="webview_error_SSL_EXPIRED">The Home Assistant certificate has expired, please review the Home Assistant certificate or the connection settings and try again.</string>
+  <string name="webview_error_SSL_IDMISMATCH">The Home Assistant Hostname certificate mismatch, please review the Home Assistant certificate or the connection settings and try again.</string>
+  <string name="webview_error_SSL_INVALID">A generic SSL error occured when loading Home Assistant, please review the Home Assistant certificate or the connection settings and try again.</string>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -32,6 +32,13 @@ integration enabled on your home assistant instance.</string>
   <string name="error_onboarding_connection_failed">Unable to connect to Home Assistant.</string>
   <string name="error_connection_failed">Unable to connect to Home Assistant.</string>
   <string name="webview_error">There was an error loading Home Assistant, please review the connection settings and try again.</string>
+  <string name="webview_error_description">Encountered error :</string>
+  <string name="webview_error_SSL_DATE_INVALID">The date of the Home Assistant certificate is invalid, please review the Home Assistant certificate or the connection settings and try again.</string>
+  <string name="webview_error_SSL_EXPIRED">The Home Assistant certificate has expired, please review the Home Assistant certificate or the connection settings and try again.</string>
```
dshokouhi, line 37: Should we consider making this more informative? Such as please renew the certificate and don't forget to restart HA for the change to take place.  Maybe even drop a hint to use the `cert_expiry` sensor to prevent this from happening.  I think the more helpful we are here the less frustrated users will be.

app/src/main/res/values/strings.xml
```diff
@@ -32,6 +32,13 @@ integration enabled on your home assistant instance.</string>
...
+  <string name="webview_error_SSL_IDMISMATCH">The Home Assistant Hostname certificate mismatch, please review the Home Assistant certificate or the connection settings and try again.</string>
```
dshokouhi, line 38: Would this be the error a user gets if they try to use SSL with their local IP? Might be good to provide some direction here if that is the case.

app/src/main/res/values/strings.xml
```diff
@@ -32,6 +32,13 @@ integration enabled on your home assistant instance.</string>
...
+  <string name="webview_error_SSL_INVALID">A generic SSL error occured when loading Home Assistant, please review the Home Assistant certificate or the connection settings and try again.</string>
+  <string name="webview_error_SSL_NOTYETVALID">The Home Assistant certificate is not yet valid, please review the Home Assistant certificate or the connection settings and try again.</string>
+  <string name="webview_error_SSL_UNTRUSTED">The Home Assistant certificate authority is not trusted, please review the Home Assistant certificate or the connection settings and try again.</string>
```
dshokouhi, line 41: Would this be a case for the self-signed cert error? or is that a different error?

dshokouhi:
Just some comments, I think it would be good if we can provide more clarity and direction to users and even possibly linking to the appropriate place in our docs to help them out.  We have seen some users get stuck with SSL issues and need some direction.

PR #643, opened 2020-07-01
Fix left empty notifications groups

chriss158:
Some additional fixes for notification groups.

* If user clicked on a notification and this notification was in group, the left empty notification group was not deleted.
* If the notification ran into timeout (using the timeout property) and therefore was deleted and the notification was in group, the left empty notification group was not deleted.
* If a notification with a tag and a group is used and the user decides to remove the group from notification (using the same tag property), the left empty notification group was not deleted.


balloob:
Could you next time please use a clean branch when building a new feature / fixing a bug? There are now a lot of commits that are not in our repo.

chriss158:
Oh i'm sorry, i will do 👍

app/src/main/java/io/homeassistant/companion/android/notifications/NotificationContentReceiver.kt
```diff
@@ -0,0 +1,47 @@
...
+                Intent(Intent.ACTION_VIEW).apply {
+                    this.data = Uri.parse(action)
+                }
+            } else {
+                WebViewActivity.newInstance(context, action)
+            }
+
+            intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
+
+            context.startActivity(intent)
```
JBassett, line 44: Isn't this code duplicated here: https://github.com/home-assistant/android/blob/e16bc38a7436f3c4d9eca93db9596e0451b29e67/app/src/main/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt#L95-L105

and dead code here: https://github.com/home-assistant/android/blob/e16bc38a7436f3c4d9eca93db9596e0451b29e67/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt#L218-L237

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -135,25 +136,37 @@ class MessagingService : FirebaseMessagingService() {
      *
      */
     private suspend fun sendNotification(data: Map<String, String>) {
-        var group = data["group"]
-        if (!group.isNullOrBlank()) group = GROUP_PREFIX + group
+        val notificationManager =
+            getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
```
JBassett, line 140: Why are we moving away from `NotificationManagerCompat`?

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -169,26 +182,75 @@ class MessagingService : FirebaseMessagingService() {
...
 
-        notificationManagerCompat.apply {
+        notificationManager.apply {
             notify(tag, messageId, notificationBuilder.build())
             if (group != null) {
                 notify(group, groupId, getGroupNotificationBuilder(channelId, group, data).build())
+            } else {
+                if (!previousGroup.isNullOrBlank()) {
+                    notificationManager.cancelGroupIfNeeded(previousGroup, previousGroupId)
+                }
```
JBassett, line 204: If we are adding a notification why would we ever need to cancel the group?

app/src/main/java/io/homeassistant/companion/android/notifications/NotificationContentReceiver.kt
chriss158: Yes it is a duplicate code. Didn't noticed that one. And also the dead code. I will remove it.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
chriss158: Because i needed the extension methods of the notification manager. It was easier (for me at least) to use again the notification manager. I'll try to switch back again to the compat one.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
chriss158: If the notification was in a group and it is updated (by the tag property) and it was taken out of the  group by removing the group property, the left empty group needs to be removed (previousGroup).

So first it updates the notification with the notify method, then we check if we had a group in the last notification. (previousGroup). If yes we cancel the group, if its empty.

JBassett: Ok, makes sense.  Think it's a super edge case in practice but good catch.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingService.kt
JBassett: I just worry about older devices, the compat layer adds some safety for us.

PR #650, opened 2020-07-05
Fix widget service calls with numbers (#640)

### LINKED ISSUE #640: Widget sends alarm code as float, fails comparison
1.9.0-167
**Android version:**
10
**Phone model:**
Pixel 3
**Home Assistant version:**
0.111.4

**Description of problem:**
When trying to create a widget to trigger the 'alarm_control_panel.alarm_disarm' service, the widget appears to function correctly but generates an 'invalid code' error in alarm log. Inspecting the logged webhook from the mobile_app shows the code field appears to be sent as a float, and I believe that fails the compare in the alarm panel. 

```
2020-06-28 12:13:37 DEBUG (MainThread) [homeassistant.components.mobile_app.webhook] Received webhook payload for type call_service: {'domain': 'alarm_control_panel', 'service': 'alarm_disarm', 'service_data': {'code': 1234.0, 'entity_id': 'alarm_control_panel.home_alarm'}}
2020-06-28 12:13:37 WARNING (SyncWorker_11) [homeassistant.components.manual.alarm_control_panel] Invalid code given for disarmed
```

**Screenshot of problem:**
Screenshot_20200628-113226

**Additional information:**
Alarm code in this example is: 1234

chriss158:
Fixes #640 

Widget service calls will fail, if a number for any field is used.
Because the GSON parser will parse every integer to a double.
This is especially a problem on the service alarm_control_panel.alarm_disarm
which needs a code to disarm the alarm panel.
So every code will be converted to .0 (eg. 1234 would be 1234.0)
Solution: Use Jackson instead of gson
Jackson parses integers correctly

GSON problem described here:
https://github.com/google/gson/issues/1084

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -115,6 +115,7 @@ dependencies {
     }
 
     implementation(Config.Dependency.Misc.lokalize)
+    implementation(Config.Dependency.Misc.jackson)
 
     implementation(Config.Dependency.Play.location)
     implementation(Config.Dependency.Firebase.core)
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/ButtonWidget.kt
@@ -10,7 +10,8 @@ import android.os.Handler
 import android.util.Log
 import android.view.View
 import android.widget.RemoteViews
-import com.google.gson.Gson
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.fasterxml.jackson.module.kotlin.readValue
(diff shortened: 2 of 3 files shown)


JBassett:
I agree this is a way to fix it on the application's end.  However, I would be curious how @balloob would feel about implementing a fix on the core side.  Technically GSON is following the spec by adding `.0`, the JSON spec doesn't differentiate between Integers and Floats only Numbers.  If there is no appetite to fix on the Core side then we can implement this change.

balloob:
It seems like you have some commits in this branch that are not yours, please clean it up.

chriss158:
Damn, i did accidentally a rebase. Should be ok now.

PR #651, opened 2020-07-05
Hot fix removing authentication from settings.

JBassett:
Was causing crashes and I'm not sure if we need on the settings.  If you have gotten into the application we should assume you are totally authenticated.


neopilou:
Because if you leave the application when you are in the settings and you resume (not restart) the app later, you can access the settings without authentication. This can allow anyone to access the settings and disable authentication. So we have two solutions (maybe others), either we ask for an authentication, or we force the resume of the app in the WebViewActivity, and not in the SettingsActivity.

Issue #558, opened 2020-04-17
Unable to Register Application on Android


ParadoxRus:
Home Assistant Android version:
1.8.0-144

Android version:
Android 10 QP1A.190711.020

Phone model:
Redmi Note 8 Pro

Home Assistant version:
Home Assistant Core 0.107.7
installed on dietPi 6.28.0 on RasberryPi 4

Description of problem:
I've got the same problem as a closed "Unable to Register Application #417"
I am unable to access Home Assistant by Android smartphone, it gets stuck in the screen with the "Unable to Register Application" message.
Plz check to ensure you have the mobile_app integration enabled on your home assistant instance.

Things I have try:
    Restarting HA.
    Reinstalling the app
    Grant all permissions to the app
    Disable battery optimization for the app.
    Enable GPS
    write mobile_app: directly in configuration.yaml vs default_config:

# Configure a default setup of Home Assistant (frontend, api, etc)
#default_config:
automation:
cloud:
config:
frontend:
history:
logbook:
map:
mobile_app:
person:
ssdp:
sun:
system_health:
updater:
zeroconf:

# Uncomment this if you are using SSL/TLS, running in Docker container, etc.
# http:
# base_url: example.duckdns.org:8123

# Text to speech
tts:
- platform: google_translate

group: !include groups.yaml
automation: !include automations.yaml
script: !include scripts.yaml
scene: !include scenes.yaml

discovery:

Additional information:
I have the stock MIUI version with Google Services installed the same problem I also have on Samsung A5. In HA logs I do not see anything.
I connect in my local network with the url like http://serverip:8123 (in my other device this works and in phone browser too)

dshokouhi:
@ParadoxRus #417 was closed because it had to do with HA integration.  It required users to clear data in the app.  Delete the HA integration and restart HA.  Then when you login it would work.  

Did you install the app via google play store?

Does your other device have MIUI on it or a stock ROM?

dshokouhi:
Also this is a possible duplicate of #446

ParadoxRus:
> 
> 
> @ParadoxRus #417 was closed because it had to do with HA integration. It required users to clear data in the app. Delete the HA integration and restart HA. Then when you login it would work.
> 
> Did you install the app via google play store?
> 
> Does your other device have MIUI on it or a stock ROM?

You can see above
_Things I have try:
Restarting HA.
Reinstalling the app
Grant all permissions to the app
Disable battery optimization for the app.
Enable GPS_

[...]

PR #673, opened 2020-07-19
Allow the firebase instance id be optional.

JBassett:
This is discouraged but helpful for contributors.  This way we don't need everyone to setup firebase just to help on the app.

Fixes: #558


JBassett:
@balloob That's correct, however if Google Play Services are having an issue it would also be displayed.  If someone does skip it when they come back up we should get the new ID because we already listen for changes to that ID and it would register after that.  I don't have a device that doesn't have Google Play Services so I'm not sure how it will handle that use case.

Issue #676, opened 2020-07-26
[Feature Request][Bounty] Google Services-free version


anonimou0:
Hello @balloob @dshokouhi @JBassett @homeassistant ,

I know you guys have been very vocal about not maintaining a Google Play Services free version of the app for Android. I guess this is the most recurring opened issue. I would like to approach it in another way and I would discuss with you and, if you do not agree, as you are the maintainer of the project, you will always have the power to close this issue.

My goal is propose a bounty to anyone outside of the current dev team to come up with a pull request with this feature. That way, it wouldn't take your time developing it. Of course, the bounty is applicable to you too if you guys want to jump in. 

I propose 4 bounties and I will discuss them:

1. [$15] Produce a Pull Request with the minimum change possible that just checks if the phone has Google Services on it and disable "push notifications" and "location features", as they require it. That way, code maintenance on the dev team would be, possibly, the minimum. Pretty much what was done by @jleeuwes on https://github.com/jleeuwes/home-assistant-android-without-google , but embedded in the app.

2. [$25] Location features. Many apps can use microG for location and other Nlp services. The goal on this bounty would be check if Google Services is present and if not, use other location providers on the phone. [[ I don't know it's done, but even closed apps that are extremely location dependent like Uber, and browser integration work seamlessly with microG ]]

3. [$35] Push notifications. As pointed out here https://github.com/home-assistant/android/issues/42#issuecomment-581090179 there are alternatives that can be implemented to get this functionality.

4. [$15] Fdroid build with at least bounty number 1. This probably will need the team to be involved in other to avoid previously problems (https://github.com/home-assistant/android/issues/42#issuecomment-584212855). However, once set up, it would automatically generate new builds.

If any of those gets done, I will double my bounty. One share will be for the pull request person and the other share will be for home assistant team.

I'm marking here people that previously requested similar things so they could jump in and donate too. Let's support these guys and support open source projects. @jurgenhaas @rradar @skylarmt @add1989 @nagyrobi @bisby

============================================================

I don't even know how to operationally pay out the bounties. But we can figure something out if this gets traction.

============================================================

**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I open the app without Google Play Services and get the Unable to Register message.

**Describe the solution you'd like**
A possible to use the mainstream Home Assistant app with the latest features, even with some limitations.

**Describe alternatives you've considered, if any**
Alternatives would be using forks for the app or even other platforms for home automation.

**Additional context**
[...]

Issue #600, opened 2020-05-19
Add check for google services


rradar:
**Is your feature request related to a problem? Please describe.**
#231 #358 #412 #599 ... and more to come

**Describe the solution you'd like**
Check for existing google services on the device. If none found show a proper message that this app requires google services and give a information that the last working build of this app for free(d) devices is v1.6

Even better solution provided by @jleeuwes [here](https://github.com/home-assistant/android/issues/231#issuecomment-583729339
) would be to have the error message appearing but the **app still running** with all the functionalities but only stripped from the few (push notification* and ?) which are require the google services.

*Actually even this could be managed. [Signal](https://signal.org/en/) for example uses a automatic fall back for notifications if no google services are available.

**Describe alternatives you've considered, if any**
I installed [HA Client](https://github.com/estevez-dev/ha_client) on the same phone and it works without google on my phone.

**Additional context**
Right now a **not very specific** error message is shown when trying to use this app on a google free phone:
![image](https://user-images.githubusercontent.com/34582688/82384693-4aa63f80-9a28-11ea-897e-a45cd181a91c.png)

jleeuwes:
I created a low-effort fork https://github.com/jleeuwes/home-assistant-android-without-google in which I just take a stable version from time to time and rip out all Google Play / FCM related code. I use it on my LineageOS phone. The ripping-out commits like https://github.com/jleeuwes/home-assistant-android-without-google/commit/246a42ff6efd2d9fa0a2241ebc3fd970ca73a0b1 might be helpful when implementing this issue.

balloob:
This is the third time you have opened an issue @rradar. You even link to old issues. You do not respect us and our time. You can do with your time what you want, but you won't be able to spam us anymore as I have just blocked you.

Issue #599, opened 2020-05-19
Does this app has hard coded dependencies on google services?


rradar:
**Home Assistant Android version:**
1.8.0

**Android version:**
8.1

**Phone model:**
Samsung S7

**Home Assistant version:**
0.109

**Last working Home Assistant release (if known):**
-

**Description of problem:**
Is their any known/wanted hard coded dependencies of google services? 

I having trouble to get this app running in my phone (check #558 for details).

**Traceback (if applicable):**

```

```

**Screenshot of problem:**
![image](https://user-images.githubusercontent.com/34582688/82384693-4aa63f80-9a28-11ea-897e-a45cd181a91c.png)

**Additional information:**
I installed [HA Client](https://github.com/estevez-dev/ha_client) on the same phone on the other hand it worked immediately without problems

dshokouhi:
This app requires google play services to work:

https://companion.home-assistant.io/docs/getting_started/getting-started

Closing as duplicate of #358 #231

rradar:
What a shame :cry:

JonnyHaystack:
Yes, this is most disappointing :disappointed: this really goes against the spirit of Home Assistant imo. I went to the effort of using Home Assistant and other open-source solutions for home automation and avoided any cloud-based/proprietary products, and yet we are expected to install Google's proprietary rootkit on our phones? Why was it done this way?

balloob:
Have a look at the other issues for your answer.

Issue #570, opened 2020-04-27
Release APK


rradar:
**Is your feature request related to a problem? Please describe.**
I run a FOSS OS on my phone and don't use any big brother services which includes the play store

**Describe the solution you'd like**
A apk can be released for all who like to try/use the app but don't have a contract with google

**Describe alternatives you've considered, if any**
As already requested a couple of times it can('t) be included to F-Droid #42 #124 #532 

**Additional context**

balloob:
Duplicate

PR #682, opened 2020-07-31
Minimal Android Flavor

JBassett:
Address the following: #231, #358, #570, #599, #600, #676

I understand many people do not like Google and the services they ship on most Android devices.  This is me extending an olive branch to those people.  This pull request adds two flavors to our wonderful application.  

First up is `Full`, this should be exactly the same as we ship today.  It's full of Google services, but in return has full support for everything we had before (Push notifications via FCM, Google Location services, and Crashlytics to help us).  

Second we have `minimal`.  This is a stripped down version that has no connections or dependencies on Google services.  By using this version you are paying the price of not having push notifications, no location tracking, and most importantly no `priority` support.  My main effort for now is still on the version of the application we publish via the Play Store.  I'm not saying every PR will break this version but it will be less thoroughly tested.

I hope that by doing the leg work to keep the code in a single repo we can have more developers who are excited about developing for the application in general.  A rising tide lifts all boats.  Hopefully this will help drive developers to continue to give back to the community and help move this application forward!


tadly:
Will this be a official release in the play store at some point?

JBassett:
@tadly No, the APKs will be hosted on Github for now.  Publishing to the play store would be ironic since the whole reason behind the minimal build is to remove Google dependencies.  We will work on surfacing them better but you can find the builds under the actions section: https://github.com/home-assistant/android/actions/runs/193845592 for example

tadly:
Does publishing on the play store imply creating a building with google services? Sorry if that's a dumb question ':D

I assume f-droid might be the way to go then?

The reason I am interested in the minimal version is mostly because I don't need all that additional functionality and would rather have the app not "register" at all with the backend (just be a dumb frontend).
f-droid is kind of a pain as non of the store apps are all that good so I personally would prefer installing it through the play store if at all possible.
Just my 2 cents :)

Thanks for the work though (:

balloob:
The Home Assistant frontend is a progressive web app, that you can install on Android as an app with Firefox or Chrome (probably others too). You don't need the app if you just want the UI.

tadly:
I am aware of this, just never was a fan of it.
While the app is basically just a webview too, it starts quite a bit faster and does give me the "Home Network" ability so I do prefer the app in that sense.

I'm also hopeful for further improvements in regards to widgets, app startup-time (maybe due to more aggressive caching. it *could* be made almost instant), etc. which is why I'd like to stick with it :)

But this is quite off-topic and I don't want to bore you guys with it. :)
A minimal version is really nice to have be it in the play store or not

jleeuwes:
@JBassett thank you for this effort!

@tadly I don't know if you are interested in this, but the `minimal`-flavored app is being added to the F-Droid repository: https://github.com/home-assistant/android/issues/897

Issue #431, opened 2020-02-14
REQUEST: read NFC tags and fire service


pergolafabio:
I really like this app : HassNFC, it's a small Android app that reads NFC tags and fires a service, would be great if this could also be implemented in this great app..

More info: https://community.home-assistant.io/t/hass-nfc-control-home-assistant-with-nfc-tags/54227

pergolafabio:
It's also included in ariela, works different there, they created a sensor... Based on the value you read on nfc, the sensor changes

balloob:
We should do neither, we should get a new webhook event for mobile app `nfc_card_scanned` that allows mobile app integration to build up known NFC cards and their IDs.

dshokouhi:
Sounds like something else to harmonize maybe? iOS uses NFC in universal links: https://companion.home-assistant.io/docs/integrations/universal-links

GrandNewbien:
Universal links would by far be the best thing for Android. You can call any service you want then.

David-Development:
Is anyone working on this already? Maybe I could try looking into how this could be implemented by using the universal links as @GrandNewbien suggested?  If someone wants to help, feel free. This would be my first contribution to this project

pergolafabio:
i can help with testing , not coding,

using NFC tags every day, but still with the Hass NFC app , instead of the official one

David-Development:
@pergolafabio Just created a first draft of how nfc support could look like. There are a few more things to discuss but the current state already works quite well on my phone so feel free to test it. (check out the [branch](https://github.com/David-Development/android-1/tree/nfc-support), build the app, run it on your phone, open the app, go to the app configuration, scroll down, click on "NFC Tags" to start generating new nfc tags). Happy testing 🙂

pergolafabio:
cool, following that one
will it be soon in beta? i prefer to test then

pergolafabio:
gonna, close, not needed anymore now

PR #689, opened 2020-08-01
WIP: Add nfc support

David-Development:
The MR adds basic support for reading and writing of nfc tags as requested in https://github.com/home-assistant/android/issues/431.

As suggested by @dshokouhi and @GrandNewbien I implemented NFC support by adding basic [universal link](https://companion.home-assistant.io/docs/integrations/universal-links) support to the app. However due to the fact that iOS uses `www.home-assistant.io/ios/nfc/` in the universal links, they won't be interchangeable between the platforms. Maybe it would make sense to agree upon a standard so that nfc tags work on both platforms.

The UI-Code inside the `NfcActivity` is heavily influenced by the `ButtonWidgetConfigureActivity`. As I'm new to kotlin development there might be some parts of the code that probably could be optimize. Might also make sense to abstract this even more and make one generic activity that contains the logic to create the definition for service calls.

-----

How to test:

- check out the [branch](https://github.com/David-Development/android-1/tree/nfc-support)
- build and run the app
- open the app
- go to the "app configuration"
- scroll down, click on "NFC Tags" to start generating new nfc tags


homeassistant:
Hi @David-Development, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;android#689).

Thanks!

app/src/main/AndroidManifest.xml
```diff
@@ -96,6 +97,16 @@
...
             android:parentActivityName=".webview.WebViewActivity" />
 
+        <activity android:name=".nfc.NfcActivity"
+            android:parentActivityName=".webview.WebViewActivity">
+            <intent-filter>
+                <action android:name="android.nfc.action.NDEF_DISCOVERED" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:scheme="https"
+                    android:host="www.home-assistant.io"
+                    android:pathPrefix="/android/nfc/" />
```
balloob, line 107: Let's settle on a link format right now and implement it correctly right away.

CC @home-assistant/ios-developers

app/src/main/java/io/homeassistant/companion/android/nfc/NFCUtil.kt
```diff
@@ -0,0 +1,92 @@
...
+                } catch (e: IOException) {
+                    //Failed to format tag
+                    return false
+                }
+            }
+            //NDEF is not supported
+            return false
+
+        } catch (e: Exception) {
+            //Write operation has failed
```
balloob, line 87: Is this something we should tell users?

app/src/main/java/io/homeassistant/companion/android/nfc/NfcActivity.kt
```diff
@@ -0,0 +1,295 @@
...
+        } else {
+            nfc_loading.visibility = View.GONE
+            initUi();
+            mNfcAdapter = NfcAdapter.getDefaultAdapter(this)
+        }
+    }
+
+    private suspend fun handleNFCTag(ndefMessage: NdefMessage?) {
+        val url = ndefMessage?.records?.get(0)?.toUri().toString()
+        //val url = "https://www.home-assistant.io/android/nfc/?url=homeassistant://call_service/light.turn_on?entity_id=light.extended_color_light_2"
```
balloob, line 86: I'm personally not a fan of writing urls that contain the service data in it. 

I would much rather have it work like this:

- Each tag gets a unique number
- When a tag is scanned by a phone or device, the ID of the device is also send in.
- Home Assistant is aware of tags and devices and remembers all tags and devices it encounters
- People can automate with a trigger based on tag ID (and optional scan device ID)

That way we don't have to access tags to update automations and it's clear from an overview which ones we have.

I guess we can keep support for this around, but things like custom event are only allowed for admin users and generally we should not offer to users.

app/src/main/AndroidManifest.xml
zacwest: How about we split the difference and do `/nfc/*`? I thought about adding mobile as a prefix but there's no reason to bake that into NFC.

app/src/main/java/io/homeassistant/companion/android/nfc/NfcActivity.kt
David-Development: @balloob Your proposed solution sounds great to me. It has many advantages and is more flexible.

So the key-points to take away here are: 
- keep the support for universal links (but change the url to `https://www.home-assistant.io/nfc/?url=`)
- add support for scanning tags and sending them directly to home-assistant
  - not sure how this can be done? Does the rest api already support something like tags? Maybe someone can give me some pointers how this could be implemented?
  - we still have to write something onto the nfc tags.. otherwise other apps can pick up the nfc tag discovered event as well. Maybe we can create an url that looks something like this `homeassistant://nfc/random-uuid`. Therefore we are still flexible in terms of what will happen when this nfc-tag is scanned.

app/src/main/java/io/homeassistant/companion/android/nfc/NFCUtil.kt
David-Development: Yes, I'll rewrite this part so that users get a better error feedback 👍

app/src/main/java/io/homeassistant/companion/android/nfc/NfcActivity.kt
zacwest: It's worth noting for iOS at least that _only_ universal link style NFC tags can be read in the background.

My proposal for an ID-based one, would be to fire an event like `mobile_app.nfc_tag_read` and pass in everything after` nfc/` in the URL as a parameter to the event.

`https://www.home-assistant.io/nfc/5f0ba733-172f-430d-a7f8-e4ad940c88d7`
-> `mobile_app.nfc_tag_read` + `{"tag": "5f0ba733-172f-430d-a7f8-e4ad940c88d7"}`

Probably worth including as well is device-specific info.

homeassistant:
Hi @David-Development, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;android#689).

Thanks!

app/src/main/java/io/homeassistant/companion/android/nfc/NfcActivity.kt
David-Development: @zacwest I noticed the android also only supports `http`, `https` and `app` as a prefix for the schema. Therefore I went with your proposed url style.

With my latest changes, the app supports both nfc tags with an UUID and nfc tags containing the universal app link. 

I tried firing an event to `mobile_app.nfc_tag_read` as well as `mobile_app_nfc_tag_read` however I can't seem to see it in the ha ui. Is that correct? Do I need to configure something else on my ha instance first? 

In my android logcat output I can see the following request:
`{"type":"fire_event","data":{"event_type":"mobile_app.nfc_tag_read","event_data":{"device_name":"HD1903","tag":"669441d9-35ea-4c0f-bdbd-7eabf071dd1d"}}}`

> if you made an NFC reader it could fire it too

you mean a standalone component in ha itself?

app/src/main/AndroidManifest.xml
balloob: I wonder if it should be /tag/ because NFC is an implementation detail.

zacwest: It's potentially an issue but we could also be stuck trying to contort some future technology into today's limitations.

The current parallel may be iBeacon being seen (proximity in or out). I wonder if we can extend the tag read event were defining below to account for them. But they aren't really URL-based.

app/src/main/java/io/homeassistant/companion/android/nfc/NfcActivity.kt
zacwest: > you mean a standalone component in ha itself?

Yeah like a USB NFC reader that you could tag into, for example. We could go with `nfc.tag_read` perhaps.

zacwest: > I tried firing an event to `mobile_app.nfc_tag_read` as well as `mobile_app_nfc_tag_read` however I can't seem to see it in the ha ui. Is that correct? Do I need to configure something else on my ha instance first?

In the Developer Tools you can go to Events and listen to them there. Does it not appear if you listen to e.g. `*`?

David-Development: @zacwest Thank you, I thought it would appear as an event in the list on the right side. But if I listen to `*` it did show up, thank you! :)  Just updated the code so that the event is `nfc.tag_read` now.

app/src/main/AndroidManifest.xml
David-Development: @zacwest yes those beacons are not url based and android does not natively support them. So from my understanding we would have to include a third party library that does background scanning using bluetooth. However there are some restrictions on newer devices so that the detection might be delayed by about 300s (https://altbeacon.github.io/android-beacon-library/detection_times.html). However I've never used them in production.. maybe there are easier and better ways to use them however I don't have an iBeacon at hand to test with.

As iBeacons (can?) contain a uuid, it might be possible to use them in the same way as the nfc tags. However that'd require some more research and testing.

app/src/main/res/layout/activity_nfc.xml
```diff
@@ -0,0 +1,136 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="?attr/actionBarSize"
+        android:elevation="4dp"
+        android:theme="@style/ThemeOverlay.HomeAssistant.ActionBar" />
+
+    <ProgressBar
+        android:id="@+id/nfc_loading"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="100dp" />
```
JBassett, line 21: I hope the activity doesn't take that long to create that we can see this loading bar?

app/src/main/res/layout/activity_nfc.xml
```diff
@@ -0,0 +1,136 @@
...
+            android:layout_weight="1"
+            android:text="@string/nfc_button_add_tag_id" />
+
+        <androidx.appcompat.widget.AppCompatButton
+            android:id="@+id/add_universal_link"
+            style="@style/Widget.HomeAssistant.Button.Colored"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_weight="1"
+            android:text="@string/nfc_button_add_universal_app_link" />
```
JBassett, line 44: These should be centered and stacked vertically rather than pinned to the top.

app/src/main/res/layout/activity_nfc.xml
```diff
@@ -0,0 +1,136 @@
...
+    </LinearLayout>
+
+
+    <LinearLayout
+        android:id="@+id/nfc_tag_id_setup"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:gravity="center">
+
+        <androidx.appcompat.widget.AppCompatButton
```
JBassett, line 54: Rather than another button why not just give a spinner with instructions to tap to the NFC tag to write too?

app/src/main/java/io/homeassistant/companion/android/nfc/NfcActivity.kt
```diff
@@ -0,0 +1,339 @@
...
+            val params = serviceDataMap.map {(k, v) -> "${k.utf8()}=${v[0]}"}.joinToString("&")
+
+            nfcUrlToWrite =
+                "https://www.home-assistant.io/nfc/?url=homeassistant://call_service/$domain.$service?$params"
+            Log.d(TAG, "Generated url: $nfcUrlToWrite")
+            Toast.makeText(applicationContext, R.string.nfc_write_tag_info, Toast.LENGTH_LONG)
+                .show()
+        } catch (e: Exception) {
+            Log.e(TAG, "Issue configuring nfc tag", e)
+            Toast.makeText(applicationContext, R.string.widget_creation_error, Toast.LENGTH_LONG)
```
JBassett, line 236: Hope it's not a widget creation error! ;-)
Also we might want to do more than just a toast to let the user know something went wrong.

app/src/main/AndroidManifest.xml
balloob: I'm mainly thinking about RFID cards using non-phone apps as readers. https://github.com/maddox/magic-cards

zacwest: If we want to tack a URL onto an identifier (which is what this path is) I'd imagine we're painting ourselves less into a corner. Requiring a URL format for RFID might be overkill (or exceed maximum capacity, etc.) is pushing today's limitations on tomorrow's technology. I'd imagine we say `home-assistant.io/nfc` is more of a namespace than a standard.

However, the RFID point about the _event_ being fired is very convincing, and (comment below) I think we should take the `nfc` out of it.

app/src/main/java/io/homeassistant/companion/android/nfc/NfcActivity.kt
zacwest: From the discussion above, I think we should abstract _again_! I'm not as confident in it, but how do you feel about `tag.read`?

app/src/main/AndroidManifest.xml
balloob: Sounds good

zacwest: I thought about this more and I think I am wrong. For example, a QR code shouldn't require a new implementation for the apps to handle it. What we have are a class of things (URL-based tags) and other things (raw/limited tags). The latter will require unique approaches by their nature.

So new proposal: `https://www.home-assistant.io/tag/<id>` for all URL-based tags. How the app get launched for a URL matters less -- just that it did.

zacwest:
Summarizing some of the above, this is what implemented in iOS: `/tag/xyz` sends `tag.read` with `{"tag" = "xyz"}` + standard (albeit iOS) device information. This works for both QR Codes and NFC Codes, or really anything which can make the URL open.

It's all readily changeable if there's anything that sways one way or another.

David-Development:
Thank you for all the feedback. I really like the iOS design that @zacwest implemented for reading NFC tags. I am currently refactoring the code to split it more in a native android way instead of having all logic in one single class. In favor of simplicity I will remove the support to create universal links in the app as it is quite a lot of overhead and the iOS app does not support it either.

zacwest:
> I will remove the support to create universal links in the app as it is quite a lot of overhead and the iOS app does not support it either.

This is the only NFC support in iOS -- writing/reading `https://www.home-assistant.io/tag/<id>` universal link tags.

David-Development:
> > I will remove the support to create universal links in the app as it is quite a lot of overhead and the iOS app does not support it either.
> 
> This is the only NFC support in iOS -- writing/reading `https://www.home-assistant.io/tag/<id>` universal link tags.

From what I understood from the docs is that universal links look like this `https://www.home-assistant.io/ios/nfc/?url=<a URL you could use with the existing URL handler>`. So basically urls that contain the `call_service` and `fire_event`
Ref: https://companion.home-assistant.io/docs/integrations/universal-links/

I wasn't sure if the new url `https://www.home-assistant.io/tag/<id>` is also considered a "universal link". Thanks for the clarification!

zacwest:
> From what I understood from the docs is that universal links look like this `https://www.home-assistant.io/ios/nfc/?url=<a URL you could use with the existing URL handler>`. So basically urls that contain the `call_service` and `fire_event`
> Ref: https://companion.home-assistant.io/docs/integrations/universal-links/
> 
> I wasn't sure if the new url `https://www.home-assistant.io/tag/<id>` is also considered a "universal link". Thanks for the clarification!

Oh sorry yup, I honestly forgot about that format - universal links are the only way to have background scanning of the NFC tag, so they're both implemented that way. For simplicity, I didn't allow creating the "old style" URL, but it's still handled if there's an existing tag that implements it, so that I totally understand.

David-Development:
I tried implementing a similar workflow as the iOS app has for handling nfc tags (https://github.com/home-assistant/iOS/pull/863). The code makes use of the new androidx `LiveData` and fragment navigation. Reading and writing nfc tags is quite smooth now.

app/src/main/java/io/homeassistant/companion/android/nfc/NfcActivity.kt
David-Development: Ahah oh no.. my bad, fixed it! Thanks for noticing!

app/src/main/res/layout/activity_nfc.xml
David-Development: Great idea, changed it in the latest commit :)

app/src/main/res/layout/activity_nfc.xml
David-Development: I'm using a different layout now which uses the `ConstraintLayout`

app/src/main/res/layout/activity_nfc.xml
David-Development: The progressbar is shown when an nfc tag is scanned and the request is send to the server 🙈 The activity is faster than that! 😅

dshokouhi:
@David-Development just an FYI since I noticed you mentioned the testing bit in your PR.  Once your PR passes the validation the APKs will be available in the actions section in github, in fact every passing commit will have its own set.  Makes testing a bit easier for users :)  Looks I'll be needing some NFC tags to play with this PR!

zacwest:
@David-Development Once you've got the linting/actions settled I'll install the .apk to give it a try too!

One question: I've made iOS support `/tag/xyz` deeplinks from e.g. QR codes the same as from NFC tags. Does this happen for this implementation? I'm writing up some documentation about the iOS version and I can include details about Android at the same time.

pergolafabio:
Is this already in beta to test it out?

dshokouhi:
> Is this already in beta to test it out?

new beta's get cut when a PR is merged.....debug APKs get generated when the build passes and can be found on the actions tab on the github site.

David-Development:
@dshokouhi Oh, I did not know that! Thank you, that is a handy feature!

@zacwest Nice! Looking forward to some feedback! 

> One question: I've made iOS support /tag/xyz deeplinks from e.g. QR codes the same as from NFC tags. Does this happen for this implementation? I'm writing up some documentation about the iOS version and I can include details about Android at the same time.

Honestly, I'm not sure. I just tested it on my phone but from my camera app (oneplus phone) it redirects me to the browser. Since camera apps are quite fragmented across platforms I'm not sure if there is a universal way like there is on ios.

JBassett:
@David-Development should be possible: https://developer.android.com/training/app-links

That should let us register the app as a handler for those urls and android will auto open it.

David-Development:
@JBassett Just ran a quick test on deep links.. however chrome is still opened by default. I think we'll have to use "android app links". However those require that we upload some certificate on the home-assistant.io website for validation (Reference https://developer.android.com/studio/write/app-link-indexing#associatesite)

JBassett:
@David-Development yep the iOS guys had to do something similar: https://github.com/home-assistant/home-assistant.io/blob/c0691b1166a2867ddad8f8272fceb37eda827f82/source/.well-known/apple-app-site-association

app/src/main/AndroidManifest.xml
```diff
@@ -96,6 +97,21 @@
...
+        <activity
+            android:name=".nfc.NfcActivity"
+            android:label="@string/nfc_title_nfc_processing">
+            <intent-filter>
+                <action android:name="android.nfc.action.NDEF_DISCOVERED" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data
+                    android:scheme="https"
+                    android:host="www.home-assistant.io"
+                    android:pathPrefix="/nfc/" />
```
zacwest, line 109: `/tag/`?

app/src/main/java/io/homeassistant/companion/android/nfc/NfcEditSetupActivity.kt
```diff
@@ -0,0 +1,101 @@
...
+                // somehow the protocol gets stripper when scanning using foreground service
+                val nfcTagId = UrlHandler.splitNfcTagId(url)
+                if (nfcTagId == null) {
+                    viewModel.postNewUUID()
+                } else {
+                    viewModel.nfcReadEvent.postValue(nfcTagId)
+                }
+            } else {
+                try {
+                    val nfcTagUrl = "https://www.home-assistant.io/nfc/$nfcTagToWriteUUID"
```
zacwest, line 85: `/tag/`?

zacwest:
I'll admit I'm not super familiar with Android these days, but I don't see a way to manage the NFC tags in the pane. Is there a button or something hiding here? NFC is working on the device otherwise (tested scanning an empty tag, Android picked it up). [Screenshot_20200809-130444](https://user-images.githubusercontent.com/74188/89740830-2e230d80-da41-11ea-808b-06509a6022df.png)

David-Development:
@JBassett That's good to know, then I'll try integrating android app links as well. 

@zacwest That is correct. From my understanding the pane is coming from the home-assistant website, right? I ended up moving the nfc button inside the settings activity as I didn't know how to add it to the pane. So in order to open up the nfc menu, you need to open the "App configuration" and then go to "nfc tags" (almost at the bottom of the settings page).

zacwest:
That screenshot is of the screen you're mentioning, but now that it's night time I realize: the buttons for 'read' and 'write' are white text and do not appear visually in light mode.

David-Development:
@zacwest Oh, that is quite interesting! It works on my device (Android 10). Which Android Version are you on? I guess I'll have to test again why the colors are not working on your device.

zacwest:
Android 10 on a Pixel 2.

app/src/main/java/io/homeassistant/companion/android/nfc/NfcEditFragment.kt
```diff
@@ -0,0 +1,110 @@
...
+            findNavController().navigate(R.id.action_NFC_WRITE)
+        }
+
+        btn_tag_fire_event.setOnClickListener {
+            mainScope.launch {
+                val uuid: String = viewModel.nfcReadEvent.value.toString()
+                val deviceName = integrationUseCase.getRegistration().deviceName!!
+                try {
+                    integrationUseCase.fireEvent(
+                        "nfc.tag_read",
```
balloob, line 82: I made a design doc for the tag integration the other day, [it's here](https://docs.google.com/document/d/1b33WrKvNUFqxB-Q7yv4g05uZ_4v7IS3lQY7yDgSUnL4/edit), I suggest we add a new webhook `tag_scanned` to the mobile app and let it work with that. I will open a PR now so it's in the beta/next release.

balloob: Update: realized we shouldn't do anything until the tag integration is added, or else we need to make a breaking change.

balloob: Ok, made PR https://github.com/home-assistant/core/pull/38721, we can do it without a breaking change.


Send webhook type `scan_tag` with data `{"tag_id": "abcd"}`

David-Development: @balloob Thank you! How do I call the webhook? 🤔 Is the following correct?

`integrationUseCase.fireEvent("scan_tag", hashMapOf("tag_id" to uuid))`

JBassett: You'll need to add a new function for it all the way down the stack (in IntegrationRepositoryImpl).

David-Development:
@JBassett okay, I just generated a DAL file. How do I get it onto the website? 😅 

It needs to be placed here: https://www.home-assistant.io/.well-known/assetlinks.json

And it needs to have the following content:
```json
[{
  "relation": ["delegate_permission/common.handle_all_urls"],
  "target": {
    "namespace": "android_app",
    "package_name": "io.homeassistant.companion.android.debug",
    "sha256_cert_fingerprints":
    ["C4:B8:BD:44:98:21:86:1B:FF:66:BF:02:65:F6:87:48:25:38:3A:B0:E8:3F:97:94:CD:38:48:B9:8E:B6:34:3A"]
  }
}]
```

Note that this is my debug fingerprint. Maybe someone can give me the cert fingerprints for the signing config that is used for the dev builds on github and the production version that is on google play?

JBassett:
> How do I get it onto the website?

Just add a PR here with the new file: https://github.com/home-assistant/home-assistant.io/tree/current/source/.well-known.  We will need to have the 4 packages in that file that we produce.

> Maybe someone can give me the cert fingerprints for the signing config that is used for the dev builds on github and the production version that is on google play?

@robbiet480 or @balloob I don't have access to our release keystore.  If someone wants to send it to me I can handle it or if you would rather get the `sha256_cert_fingerprints` from it.

David-Development:
@JBassett okay, I just created the PR (https://github.com/home-assistant/home-assistant.io/pull/14216). Feel free to edit the file accordingly and replace the namespaces alongside the package names.

JBassett:
Oops, just noticed you created a PR.  I did too: https://github.com/home-assistant/home-assistant.io/pull/14217  Mine has the values we should need.

app/src/main/java/io/homeassistant/companion/android/nfc/NfcEditFragment.kt
David-Development: Oh okay, I see, thank you!. So I added the new function. Just to double check.. this logcat output looks correct, right?

``` 
--> POST https://home-assistant.xxx.com/api/webhook/bcc461...
D/OkHttp: Content-Type: application/json; charset=UTF-8
D/OkHttp: Content-Length: 76
D/OkHttp: {"type":"scan_tag","data":{"tag_id":"123e4567-e89b-12d3-a456-426614174000"}}
```

JBassett: Yep that looks good!  We'll also want to make sure we only do that for Core versions 114 and above since that's the first version that has the new webhook.  @balloob also suggested we fire the event for versions below 114.

David-Development: @JBassett Sorry I'm a little confused by your sentence. Do I need to make a check or are we just firing the event anyways? It's no harm to send it if it's not supported, right?

app/src/main/AndroidManifest.xml
David-Development: I added `/tag/` as well. The `/nfc/` is still needed for the universal links with the `url?=`, right?

David-Development:
Great news, both nfc tags and QR-Codes are working now! 🎉 

- nfc tags and qr codes can be either one of 
   - `https://www.home-assistant.io/tag/123e4567-e89b-12d3-a456-426614174000`
   - `https://www.home-assistant.io/nfc/?url=<URL-Handler>` ([URL-Handler Specs](https://companion.home-assistant.io/docs/integrations/url-handler))

app/src/main/AndroidManifest.xml
zacwest: I think that might be up to you - the iOS app uses `/ios/nfc?url=…` for those. I did not include support for `/nfc/?tag=xyz` in the latest app, but we could add it remotely via the url config.

JBassett:
> * https://www.home-assistant.io/nfc/?url=<URL-Handler>

So correct me if I'm wrong @zacwest but iOS is deprecating this format.  No point in adding it to Android at all if you ask me.

app/src/main/AndroidManifest.xml
David-Development: Okay, I think I'll use the `/ios/nfc/` url then. Otherwise it would require changes for you too. Plus you can't generate those URLs in the Android app itself.

David-Development:
> So correct me if I'm wrong @zacwest but iOS is deprecating this format. No point in adding it to Android at all if you ask me.

Okay, just dropped it. Code is cleaner and shorter now 👍

app/src/main/java/io/homeassistant/companion/android/nfc/NFCUtil.kt
```diff
@@ -0,0 +1,79 @@
...
+                    it.format(nfcMessage)
+                    it.close()
+                    // The data is written to the tag
+                } catch (e: IOException) {
+                    // Failed to format tag
+                    throw Exception("Failed to format tag", e)
+                }
+            }
+        return true
+    }
```
JBassett, line 78: Indentation looks strange through here.

app/src/main/java/io/homeassistant/companion/android/nfc/TagReaderActivity.kt
```diff
@@ -0,0 +1,90 @@
...
+            val rawMessages = intent.getParcelableArrayExtra(NfcAdapter.EXTRA_NDEF_MESSAGES)
+            val ndefMessage = rawMessages[0] as NdefMessage?
+            mainScope.launch {
+                try {
+                    val url = ndefMessage?.records?.get(0)?.toUri().toString()
+                    handleTag(url)
+                } catch (e: Exception) {
+                    val message = R.string.nfc_processing_tag_error
+                    Toast.makeText(this@TagReaderActivity, message, Toast.LENGTH_LONG).show()
+                    Log.e(TAG, e.message)
```
JBassett, line 54: We should add the whole exception and add a custom description.
`Log.e(TAG, "Unable to handle url: $url", e)`

app/src/main/java/io/homeassistant/companion/android/nfc/TagReaderActivity.kt
```diff
@@ -0,0 +1,90 @@
...
+        } else if (Intent.ACTION_VIEW == intent.action) {
+            val data: Uri? = intent?.data
+            val url = data.toString()
+            mainScope.launch {
+                try {
+                    handleTag(url)
+                } catch (e: Exception) {
+                    val message = R.string.qrcode_processing_tag_error
+                    Toast.makeText(this@TagReaderActivity, message, Toast.LENGTH_LONG).show()
+                    Log.e(TAG, e.message)
```
JBassett, line 67: We should add the whole exception and add a custom description.
`Log.e(TAG, "Unable to handle url: $url", e)`

app/src/main/java/io/homeassistant/companion/android/nfc/NfcSetupActivity.kt
```diff
@@ -0,0 +1,100 @@
...
+                    Log.d(TAG, "Wrote nfc tag with url: $nfcTagUrl")
+                    val message = R.string.nfc_write_tag_success
+                    Toast.makeText(applicationContext, message, Toast.LENGTH_LONG).show()
+
+                    viewModel.nfcReadEvent.value = nfcTagToWriteUUID
+                    viewModel.nfcWriteTagDoneEvent.value = nfcTagToWriteUUID
+                } catch (e: Exception) {
+                    val message = R.string.nfc_write_tag_error
+                    Toast.makeText(applicationContext, message, Toast.LENGTH_LONG).show()
+                    Log.e(TAG, e.message)
```
JBassett, line 95: We should add the whole exception and add a custom description.
`Log.e(TAG, "Unable to write tag.", e)`

app/src/main/java/io/homeassistant/companion/android/nfc/NfcEditFragment.kt
```diff
@@ -0,0 +1,108 @@
...
+                try {
+                    integrationUseCase.scanTag(
+                        hashMapOf("tag_id" to uuid)
+                    )
+                    Toast.makeText(activity, R.string.nfc_event_fired_success, Toast.LENGTH_SHORT)
+                        .show()
+                } catch (e: Exception) {
+                    Toast.makeText(activity, R.string.nfc_event_fired_fail, Toast.LENGTH_LONG)
+                        .show()
+                    Log.e(TAG, e.message)
```
JBassett, line 88: We should add the whole exception and add a custom description.
`Log.e(TAG, "Unable to send tag to Home Assistant.", e)`

David-Development:
@JBassett thanks for the feedback, I implemented the suggested changes.

JBassett:
Lets get this out there and see what people think!

PR #702, opened 2020-08-04
Entity State Widget bugfix

colincachia:
- Safely typing Any? attribute to String
- Removing "N/A" when attribute not found


colincachia:
@dshokouhi The application crash issue should be fixed. Amateur mistake as I wasn't safely typing attributes to String.

With regards to the "N/A" issue - it is quite baffling as it should have been showing in cases where the entity's state was null. Same for the label not showing ... somehow it was not stored. Let me know how this fix goes.

Thanks!

JBassett:
Debug APKs are paying off already!

Issue #703, opened 2020-08-05
ACTION_REQUEST_ACCURATE_LOCATION_UPDATE doesn't work as expected


anyuta1166:
**Home Assistant Android version:**

2.0.0-3-g11982f9-238-full

**Android version:**

9

**Phone model:**

Motorola G7 Plus

**Home Assistant version:**

0.113

**Last working Home Assistant release (if known):**

**Description of problem:**
ACTION_REQUEST_ACCURATE_LOCATION_UPDATE doesn't work as expected

In the following code it was called twice from the sensors update routine. The same happens when you send the intent from Tasker.

```

08-05 03:31:50.831 12957 12957 D LocBroadcastReceiver: Got single accurate location update: Location[fused 55,637371,37,357052 hAcc=12 et=+2d11h35m11s629ms alt=201.8000030517578 vAcc=2 sAcc=??? bAcc=??? {Bundle[mParcelledData.dataSize=52]}]
08-05 03:31:50.831 12957 12957 W LocBroadcastReceiver: Location not accurate enough on retry 1 of 5
08-05 03:33:51.050 12957 12957 D LocBroadcastReceiver: Got single accurate location update: Location[fused 55,637371,37,357052 hAcc=12 et=+2d11h37m11s843ms alt=201.8000030517578 vAcc=2 sAcc=??? bAcc=??? {Bundle[mParcelledData.dataSize=52]}]
08-05 03:33:51.050 12957 12957 W LocBroadcastReceiver: Location not accurate enough on retry 2 of 5
08-05 03:46:56.353 16786 16786 D LocBroadcastReceiver: Got single accurate location update: Location[fused 55,637371,37,357052 hAcc=12 et=+2d11h50m17s153ms alt=201.8000030517578 vAcc=2 sAcc=??? bAcc=??? {Bundle[mParcelledData.dataSize=52]}]
08-05 03:46:56.353 16786 16786 W LocBroadcastReceiver: Location not accurate enough on retry 1 of 5
08-05 03:49:07.816 16786 16786 D LocBroadcastReceiver: Got single accurate location update: Location[fused 55,637371,37,357052 hAcc=12 et=+2d11h52m28s603ms alt=201.8000030517578 vAcc=2 sAcc=??? bAcc=??? {Bundle[mParcelledData.dataSize=52]}]
08-05 03:49:07.816 16786 16786 W LocBroadcastReceiver: Location not accurate enough on retry 2 of 5
```
[...]

PR #706, opened 2020-08-06
Correct accuracy for location tracking

dshokouhi:
It was mentioned in https://github.com/home-assistant/android/issues/703#issuecomment-669636182 that we should be using `MINIMUM_ACCURACY` in place of `1` so just switching it back to what was previously expected.  Possibly introduced in #682 at least from what I can tell. This isn't tested but the accuracy of `<= 1` is probably a bit extreme 😆


dshokouhi:
Tested with the debug APK on my phone, gotta love those Actions!  Location is still being reported so I don't see any issues for me.

PR #707, opened 2020-08-06
Mention app flavors and actions APK in the read me file

dshokouhi:
Just thought we should make mention of these new things in the read me file


balloob:
We should probably mention this in the app config at the top if a user is running the lite version. This will help when people start filing issues.

dshokouhi:
@balloob `full` and `minimal` are actually included in the app version at the bottom of the app config screen :)

PR #713, opened 2020-08-08
Auto detect device dark theme

### LINKED ISSUE #710: 0.114b1 - Dark Mode Not Respected
2.1.0

**Android version:**
10.0

**Phone model:**
Samsung Galaxy S9+

**Home Assistant version:**
0.114.0b1

**Description of problem:**

Dark mode is a new HA feature but its not respected with "auto" setting in the HA app. I suspect its an issue with android webview needing a special flag set. It works in Chrome for Android.

```

```

**Screenshot of problem:**

**Additional information:**

dshokouhi:
Fixes #710 

Tested locally in the android emulator: https://drive.google.com/file/d/1hOGL1an0HHNRyKpiF0lBcVVCzHmvkpEC/view?usp=sharing

Also tested on my Pixel 4 XL against HA 0.113.x and 0.115dev

The manifest changes in the webview file seem to be from gradle? Not sure :)

As dark theme was recently added this only applies to applicable devices that are on API level 29

In case you are wondering why we need webkit RC version is due to `FORCE_DARK_STRATEGY` being added then, `FORCE_DARK_AUTO` does not work as intended.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -93,7 +93,7 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
     private var resourceURL: String = ""
     private var unlocked = false
 
-    @SuppressLint("SetJavaScriptEnabled", "RestrictedApi")
+    @SuppressLint("SetJavaScriptEnabled")
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_webview)
@@ -336,10 +336,8 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
             }, "externalApp")
         }
 
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
-            if (WebViewFeature.isFeatureSupported(WebViewFeature.FORCE_DARK_STRATEGY)) {
+        if (WebViewFeature.isFeatureSupported(WebViewFeature.FORCE_DARK_STRATEGY)) {
                 WebSettingsCompat.setForceDarkStrategy(webView.getSettings(), WebSettingsCompat.DARK_STRATEGY_WEB_THEME_DARKENING_ONLY)
-            }
         }
 
         val cookieManager = CookieManager.getInstance()
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -37,9 +37,9 @@ import android.widget.ImageView
 import android.widget.Toast
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
+import androidx.room.Room
 import androidx.webkit.WebSettingsCompat
 import androidx.webkit.WebViewFeature
-import androidx.room.Room
 import eightbitlab.com.blurview.RenderScriptBlur
 import io.homeassistant.companion.android.BuildConfig
 import io.homeassistant.companion.android.DaggerPresenterComponent
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -1,5 +1,6 @@
 package io.homeassistant.companion.android.webview
 
+import android.Manifest
 import android.annotation.SuppressLint
 import android.app.PictureInPictureParams
 import android.content.Context
@@ -9,7 +10,6 @@ import android.content.pm.ShortcutManager
 import android.content.res.Configuration
 import android.graphics.Rect
 import android.graphics.drawable.Icon
-import android.Manifest
 import android.net.http.SslError
 import android.os.Build
 import android.os.Bundle
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -37,8 +37,8 @@ import android.widget.ImageView
 import android.widget.Toast
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
-import androidx.webkit.WebViewFeature
 import androidx.webkit.WebSettingsCompat
+import androidx.webkit.WebViewFeature
 import androidx.room.Room
 import eightbitlab.com.blurview.RenderScriptBlur
 import io.homeassistant.companion.android.BuildConfig
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -1,5 +1,6 @@
 package io.homeassistant.companion.android.webview
 
+import android.Manifest
 import android.annotation.SuppressLint
 import android.app.PictureInPictureParams
 import android.content.Context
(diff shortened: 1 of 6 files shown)


dshokouhi:
Tested this on my own device. It seems to have unintended consequences for HA version 0.113 and lower as it forces the entire default theme into dark as well which we know isn't optimal.  Its very similar to the chrome flag that forces everything to dark mode.

Maybe we can add a condition to only apply to HA version 0.114 or higher?

dshokouhi:
It also appears the entire app is now in a dark theme too including app configuration and onboarding 🤔

dshokouhi:
Light mode does not work as intended here because its going by device theme, so if the device theme is set to dark then it wont look right because light mode will be forced to be dark.  At least when dark mode on the device is selected it does detect and display as expected in 0.114 and looks good.

dshokouhi:
Made the switch from `FORCE_DARK_AUTO` to `FORCE_DARK_STRATEGY` and now previous HA versions are working as intended.  Onboarding and App configuration will follow the device theme so they are now dark when needed. Let me know if I should attach screenshots.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -90,7 +93,7 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
     private var resourceURL: String = ""
     private var unlocked = false
 
-    @SuppressLint("SetJavaScriptEnabled")
+    @SuppressLint("SetJavaScriptEnabled", "RestrictedApi")
```
balloob, line 96: Usually not a good sign if we have to disable the linter from warning about using a restricted api. Which one is a restricted api here ?

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -332,6 +336,12 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
             }, "externalApp")
         }
 
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            if (WebViewFeature.isFeatureSupported(WebViewFeature.FORCE_DARK_STRATEGY)) {
+                WebSettingsCompat.setForceDarkStrategy(webView.getSettings(), WebSettingsCompat.DARK_STRATEGY_WEB_THEME_DARKENING_ONLY)
```
balloob, line 341: From https://developer.android.com/reference/androidx/webkit/WebSettingsCompat#DARK_STRATEGY_WEB_THEME_DARKENING_ONLY

> In this mode WebView content will always be darkened using dark theme provided by web page. 

This will always do dark mode, and user can't do normal mode ? Shouldn't we do `FORCE_DARK_AUTO`? https://developer.android.com/reference/androidx/webkit/WebSettingsCompat#FORCE_DARK_AUTO

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
dshokouhi: It was for `FORCE_DARK_STRATEGY` I assume because its new?

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
dshokouhi: So in my testing the default theme pre HA 0.113 had issues with `FORCE_DARK_AUTO` if the device theme was set to dark, it did not look pretty there.

With this the theme the user sets in their profile is always respected, `auto` from the profile respects the devices theme.

dshokouhi: If you want you can compare the 2 implementations on your device.

this build uses `FORCE_DARK_AUTO`: https://github.com/home-assistant/android/actions/runs/199936179

this build uses `DARK_STRATEGY`: https://github.com/home-assistant/android/actions/runs/199960719

You will see what I am talking about if you set your device theme to dark and load HA 0.113.x and also on 0.114b if you set the user profile to `light` it completely darkens the background when `FORCE_DARK_AUTO` is used.

balloob: Ok, just wanted to check.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
dshokouhi: Would github actions fail if we submitted without `RestrictedApi`? It does run fine locally when I remove it.  I only added it so Android Studio wouldn't complain :)

JBassett: @dshokouhi I removed it locally and I didn't get the linter complaining.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -332,6 +336,12 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
             }, "externalApp")
         }
 
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            if (WebViewFeature.isFeatureSupported(WebViewFeature.FORCE_DARK_STRATEGY)) {
```
JBassett, line 340: Based on the documentation around `isFeatureSupported` I don't think you need the `SDK_INT` check before it, `isFeatureSupported` should handle it all.

dshokouhi: Ok I think the only part needed was the custom styles for v29 then :) I was beating my head around getting this to work properly so I tried lots of things lol

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
dshokouhi: Yup it didn't complain, probably stale code on my end...my bad :)

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
dshokouhi: In my testing it seems to continue to work as desired, so its cleaner now thanks for suggestion!

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
bramkragten: I believe you can choose strategy between theme only and invert? We would want theme and not invert. Sounds like you are now getting the invert logic.

bramkragten: https://stackoverflow.com/a/61643614

The force dark feature for webview content has two so-called strategies:

User agent darkening: The webview will set its content to dark mode by automatically inverting or darkening colors of its content.

Theme based darkening: The webview will change to dark mode according to the theme of the content (which includes the @media (prefers-color-scheme: dark) query).

To set which strategy the webview should use to apply force dark you can use the following code:

if(WebViewFeature.isFeatureSupported(WebViewFeature.FORCE_DARK_STRATEGY)) { WebSettingsCompat.setForceDarkStrategy(myWebView.getSettings(), WebSettingsCompat.DARK_STRATEGY_WEB_THEME_DARKENING_ONLY); }

dshokouhi: @bramkragten to my understanding the current strategy should only pick the dark theme if the website provides it.  Thats at least according to the stack overflow thread.  So if my device is set to dark mode, 0.114 auto detects it displays dark, if i switch device theme to light auto detection changes. If I pick a light theme like #715 and my device is on dark it still works as desired it seems.

```
Only use the dark theme of the content itself to set the page to dark mode. If the content doesn't have a dark theme, the webview won't apply any darkening and show it "as is".
```

bramkragten: Yeah by checking again you seem to be using the code I linked to 😅

dshokouhi:
Ugh I messed this one up when my branch fell behind master :( 

@chriss158 should we just add this to yours now? :P  I think I need to resubmit otherwise :(

dshokouhi:
Just going to resubmit on a clean branch

PR #714, opened 2020-08-08
Fix statusbar color if theme used rgb color instead of hex color

chriss158:
If a theme is using rgb instead of hex for color, the status wont be colored.

**Example:**
Does work:

```
Google` Dark Theme:
  # Header:
  app-header-background-color: #171717
```
Does not work, because color is defined as rgb
```
Google` Dark Theme:
  # Header:
  app-header-background-color: rgb(23, 23, 23)
```

**Solution:** 
Try to parse rgb color first. If not possible, parse hex color


miguelcaravantes:
Using google dark theme and happens to me as well

Issue #719, opened 2020-08-09
Skip biometric auth prompt


noam148:
**Home Assistant Android version:**
2.1.0

**Android version:**
10.0

**Phone model:**
Xiaomi A2 Lite

**Home Assistant version:**
0.113.3

**Description of problem:**
When the biometric auth prompt shows up and you leave the app by switching to another app. After returning to HomeAssistant the biometric auth promt is gone and have access to the app (with blue enabled)

**Screenrecord of problem:**
[https://youtu.be/x2CQS05EcQM](https://youtu.be/x2CQS05EcQM)

**Additional information:**
`biometricPrompt.authenticate(promptDialog)` inside `Authenticater.kt` will be called but something in 
`BiometricPrompt.java` goes something wrong at: `private void authenticateInternal(@NonNull PromptInfo info, @Nullable CryptoObject crypto)`. But not enough time to find out

JBassett:
Should be fixed by #722

PR #722, opened 2020-08-09
fix: Biomatric authentication skip issue

noam148:
Fix for #719
It's not the best solution because the problem is still there but it fixes for now the issue to skip the authentication.


homeassistant:
Hi @noam148, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;android#722).

Thanks!

PR #723, opened 2020-08-09
Fix missing message if notification is collapsed

chriss158:
The message is missing in collapsed state of the notification. The message is only shown in expanded state of the notification.

For some devices (at least for oneplus), we need to set also the content text additionally to the big text to show the message also in collapsed notification state.

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -423,7 +423,9 @@ class MessagingService : FirebaseMessagingService() {
             builder.setContentTitle(getSpannedTextFromHtml(it))
         }
         data[MESSAGE]?.let {
-            builder.setStyle(NotificationCompat.BigTextStyle().bigText(getSpannedTextFromHtml(it)))
+            val text = getSpannedTextFromHtml(it)
+            builder.setContentText(text)
+            builder.setStyle(NotificationCompat.BigTextStyle().bigText(text))
         }
     }
 


app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -423,7 +423,9 @@ class MessagingService : FirebaseMessagingService() {
             builder.setContentTitle(getSpannedTextFromHtml(it))
         }
         data[MESSAGE]?.let {
-            builder.setStyle(NotificationCompat.BigTextStyle().bigText(getSpannedTextFromHtml(it)))
+            val text = getSpannedTextFromHtml(it)
+            builder.setContentText(text)
```
JBassett, line 427: Won't this override what is set in `handleSubject` above?

chriss158: handleSubject is called after handleText. So the subject should overwrite the content text, if defined

gregg098:
I have my push messages setup through Node Red and its been working fine. All messages only have text in the "message" field and I do not use titles at all. With the app release that has this fix, all of my messages now come in with a title of "Missing." Is this intended?

gregg098:
Attached a screenshot. One message has a title and a message. The other just a message.

foto_no_exif

dshokouhi:
@gregg098 first check that you are indeed running `2.1.0-9-ge7e4a25-248-full` then try to force stop the app and open it again.  I just sent `message: test` to my device and it works as expected and I received nothing in the `title`

gregg098:
> @gregg098 first check that you are indeed running `2.1.0-9-ge7e4a25-248-full` then try to force stop the app and open it again. I just sent `message: test` to my device and it works as expected and I received nothing in the `title`

Weird. I just downloaded the newest update from the Play Store less than an hour ago so I assumed that was it. You mentioned this and I checked the play store and it updated again. Guess it cached a beta version from the last few days or something? Anyways, messages do work correctly now. Sorry for the inconvenience.

dshokouhi:
There multiple commits today actually so it could've been one of those :)

JBassett:
One of the downsides of the Play Store version is that it takes a little longer to get to devices.  It should just update your existing app if you don't want to wait.

PR #726, opened 2020-08-10
Initial commit of next alarm sensor

### LINKED ISSUE #56: Mobile Alarm Sensor
The mobile alarm time would be a great to have in home assistant! This works in other apps so should be possible for this app also.

dshokouhi:
This PR adds a next alarm sensor that utilizes androids [`AlarmManager`](https://developer.android.com/reference/android/app/AlarmManager) and fixes #56

The state is provided in UTC date format so the front end can translate it properly with device class `timestamp`

There are also attributes for local time and time in milliseconds.

The sensor will update anytime the alarm is updated.  The state will become `unavailable` when there is no next alarm.

image

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NextAlarmManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NextAlarmManager.kt
@@ -15,9 +15,6 @@ class NextAlarmManager : SensorManager {
     companion object {
 
         private const val TAG = "NextAlarm"
-        var triggerTime = 0L
-        var local = ""
-        var utc = ""
     }
 
     override fun getSensorRegistrations(context: Context): List<SensorRegistration<Any>> {
@@ -48,6 +45,10 @@ class NextAlarmManager : SensorManager {
 
     private fun getNextAlarm(context: Context): Sensor<Any>? {
 
+        var triggerTime = 0L
+        var local = ""
+        var utc = "unavailable"
+
         try {
             val alarmManager: AlarmManager = context.getSystemService(Context.ALARM_SERVICE) as AlarmManager
 
@@ -64,12 +65,9 @@ class NextAlarmManager : SensorManager {
                 val sdf = SimpleDateFormat(dateFormat)
                 sdf.timeZone = TimeZone.getTimeZone("UTC")
                 utc = sdf.format(Date(triggerTime))
-            } else {
-                utc = "unavailable"
             }
         } catch (e: Exception) {
             Log.e(TAG, "Error getting the next alarm info", e)
-            return null
         }
 
         val icon = "mdi:alarm"
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NextAlarmManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NextAlarmManager.kt
@@ -67,9 +67,8 @@ class NextAlarmManager : SensorManager {
             } else {
                 utc = "unavailable"
             }
-
         } catch (e: Exception) {
-            Log.e(TAG,"Error getting the next alarm info", e)
+            Log.e(TAG, "Error getting the next alarm info", e)
             return null
         }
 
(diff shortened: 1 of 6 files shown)


dshokouhi:
Tested debug APK locally, it not only works with the Google Clock app but it also works for Sleep as Android.  I think this should cover most of the popular alarm apps :)

app/src/main/java/io/homeassistant/companion/android/sensors/NextAlarmReceiver.kt
```diff
@@ -0,0 +1,39 @@
...
+    @Inject
+    lateinit var integrationUseCase: IntegrationUseCase
+    private val ioScope = CoroutineScope(Dispatchers.IO)
+    var updateJob: Job? = null
+    override fun onReceive(context: Context, intent: Intent) {
+        val isBootIntent = Intent.ACTION_BOOT_COMPLETED.equals(intent.action, ignoreCase = true)
+        val isNextAlarmIntent = AlarmManager.ACTION_NEXT_ALARM_CLOCK_CHANGED.equals(intent.action, ignoreCase = true)
+            if (!isBootIntent && !isNextAlarmIntent) {
+                return
+            }
```
JBassett, line 26: Is it just me or is this spacing a little off?

app/src/main/java/io/homeassistant/companion/android/sensors/NextAlarmManager.kt
```diff
@@ -0,0 +1,88 @@
...
+import java.util.GregorianCalendar
+import java.util.TimeZone
+
+class NextAlarmManager : SensorManager {
+    companion object {
+
+        private const val TAG = "NextAlarm"
+        var triggerTime = 0L
+        var local = ""
+        var utc = ""
```
JBassett, line 20: These should be declared in the method they are used since they are only used there.  The constant should stay.

app/src/main/java/io/homeassistant/companion/android/sensors/NextAlarmManager.kt
```diff
@@ -0,0 +1,88 @@
...
+                val dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
+                val sdf = SimpleDateFormat(dateFormat)
+                sdf.timeZone = TimeZone.getTimeZone("UTC")
+                utc = sdf.format(Date(triggerTime))
+            } else {
+                utc = "unavailable"
+            }
+        } catch (e: Exception) {
+            Log.e(TAG, "Error getting the next alarm info", e)
+            return null
```
JBassett, line 72: Maybe when you move `utc` to be declared inside this function you set it initially to `"unavailable"` then if the `alarmClockInfo != null` it will override it to the correct value.  We also probably want to send the unavailable if we fail to get the alarm info. AKA remove the return null.

app/src/main/java/io/homeassistant/companion/android/sensors/NextAlarmReceiver.kt
dshokouhi: I just manually checked and they seemed to be fine.  Linter complained about one of the spaces being 5 instead of 8 so I would hope it would catch this too :)

JBassett: Tabs vs spaces maybe?

dshokouhi: Just checked, there are no tabs in this method.  Also cross checked in notepad++ that uses dots or arrows for spaces and tabs

image

dshokouhi: In android studio go to File > Settings > Editor > General > Appearance and turn on "show white spaces" it will do the same thing :)

dshokouhi: Yup you were right, too many indents 🤦 lol

PR #730, opened 2020-08-11
Add Exoplayer view and interface

uvjustin:
See also https://github.com/home-assistant/core/pull/38125 and https://github.com/home-assistant/frontend/pull/6591
H.265 videos are not supported through Chromium or Webview on Android. We can work around this for Android by using Exoplayer which supports H.265 if the codecs are available on the device. This PR defines an Exoplayer view for the Android client and provides a Javascript interface to help call it from the frontend.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -7,8 +7,10 @@ import android.content.Intent
 import android.content.pm.ShortcutInfo
 import android.content.pm.ShortcutManager
 import android.content.res.Configuration
+import android.graphics.Color
 import android.graphics.Rect
 import android.graphics.drawable.Icon
+import android.net.Uri
 import android.net.http.SslError
 import android.os.Build
 import android.os.Bundle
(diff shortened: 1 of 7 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -333,7 +367,49 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
                     }
                 }
             }, "externalApp")
-        }
+
+            addJavascriptInterface(object : Any() {
+
+                @JavascriptInterface
+                fun playHLS(url: String) {
```
balloob, line 374: We shouldn't just add JavaScript interfaces. It should be added to the external bus https://developers.home-assistant.io/docs/frontend/external-bus

uvjustin: Got it, thanks.

uvjustin: @balloob With the Javascript interface, it is easy for the frontend to detect whether the client implements the Exoplayer functionality by checking whether the interface is defined or not, and if it is defined we can reroute the hls call to the exoplayer view rather than hls.js or native.
With message passing we can still pass the message and get the information to exoplayer, but the message result is returned asynchronously (or not returned at all if we are on a different client), so the frontend won't know what player to use until the message reply is received. Any pointers on how to get information about the client's capabilities to the frontend either before or in sync with the initial message?

balloob: The external bus allows fetching the config. So we first need to add a boolean to indicate that an external player is supported, and then add a new command to open the external player from the frontend.

balloob: But before we continue, we should get some input from @JBassett if Exoplayer is a feature we would like to add. 

Also would be good to involve @home-assistant/ios-developers as they might have a similar thing (although I think that iOS does this by default)

uvjustin: Ok, thanks. Will wait for their feedback. iOS has no problem playing H265 in html5, so there's no need for this there.
See my post here for context: https://community.home-assistant.io/t/add-support-for-h-265-in-stream-component/205661/9

uvjustin: > 
> 
> The external bus allows fetching the config. So we first need to add a boolean to indicate that an external player is supported, and then add a new command to open the external player from the frontend.

Oh ExoPlayer is not an external player...it's just the library which many Android apps use for displaying video (including the YouTube app itself). Basically I am just trying to get around the limitations of video support in Android WebView by sending a call from the frontend with the video url and video coordinates and then using Exoplayer to play the video in an overlayed view at those coordinates. The result is pretty much what the current stream component video player looks/acts like. You stay in the HA app the whole time. I've been using it myself for about two weeks and it works pretty well.

Issue #441, opened 2020-02-17
Add support for telephony states


tiagofreire-pt:
**Home Assistant Android version:** 
Last commit available on github.

**Android version:**
9.0

**Phone model:**
Samsung Galaxy J7 2017

**Home Assistant version:**
0.105.4

**Last working Home Assistant release (if known):**
N/A

**Description of problem:**

The full integration of an android smartphone with its main status reported to Home Assistant would improve the data available for automations regarding the phone states:

For in-out call/ring state (https://developer.android.com/reference/android/telephony/TelephonyManager):

- android.intent.action.ACTION_PHONE_STATE_CHANGED (https://developer.android.com/reference/android/telephony/TelephonyManager.html#ACTION_PHONE_STATE_CHANGED)

For most of these intents I have integration through MQTT, except the android.intent.action.ACTION_PHONE_STATE_CHANGED that requires some permissions on android manifest non-existant in the PlayStore app available.

This possibility would be awesome for automations regarding routines and voice assistants.

This full integration on the HA app would require less complexity, resources used by a bunch of apps and generally more efficient workflow and HA configuration.

**Comments** 

Related to https://github.com/home-assistant/home-assistant-android/issues/201#issuecomment-584741995.

PR #731, opened 2020-08-11
Telephony state support

skynetua:
Fixes: https://github.com/home-assistant/android/issues/441

This PR adds support of telephony states requested at #441 

There is a Phone State sensor added. It may have three states:
idle - no active calls at that moment
ringing - for incoming call
offhook - basically for outgoing call

image

I've made it optional through settings. I think a lot of people don't like idea of "something" tracking their calls.

There is a few things I want to add and need recommendation how to do it better.
1. Register broadcast receiver in application class after initial installation when permissions hasn't yet granted.
2. Register\unregister receiver after corresponding setting change


homeassistant:
Hi @skynetua, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;android#731).

Thanks!

app/src/main/res/layout/fragment_mobile_app_integration.xml
```diff
@@ -140,6 +140,35 @@
...
+            <androidx.appcompat.widget.AppCompatTextView
+                style="@style/TextAppearance.HomeAssistant.Title"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_marginTop="@dimen/activity_margin"
+                android:text="@string/permission_explanation_calls" />
+
+            <androidx.appcompat.widget.AppCompatButton
+                android:id="@+id/phone_state_perms"
+                style="@style/Widget.HomeAssistant.Button.Colored"
```
dshokouhi, line 152: We may or may not need to remove this, can you test locally with both dark mode and light mode enabled in the system UI? If the button is missing in light mode just remove this line and it should correct it.

dshokouhi:
Could you also attach a screenshot of the setting you created?  Just so we can see what it looks like.

skynetua:
It is visible at light mode enabled
SmartSelect_20200811-234708_Home Assistant
SmartSelect_20200811-234747_Home Assistant

app/src/main/res/layout/fragment_mobile_app_integration.xml
skynetua: Oh, i've got what you mean after repository update)

Issue #729, opened 2020-08-11
Creating widget action - "ADD FIELD" and "ADD WIDGET" buttons are invisible


juicejuice:
**Home Assistant Android version:** 2.1.1-249-full

**Android version:** Reproducible on 9.0 and 10.0

**Phone model:** Samsung Galaxy Note 10+ 5G and Galaxy Tab A 7

**Home Assistant version:** 0.113.3

**Last working Home Assistant release (if known):** HA Android 2.1.0-239-full

**Description of problem:** After updating to latest release of HA Android app, it becomes difficult to add a new widget as the two action buttons are invisible. It seems you can still tap the empty space where the button should be and it will perform the relevant action.

**Traceback (if applicable):**

```

```

**Screenshot of problem:** 
![Screenshot_20200811-164144_Home Assistant](https://user-images.githubusercontent.com/5318048/89865881-e0013c00-dbf1-11ea-9729-d8abb6815d2b.jpg)

**Additional information:**

dshokouhi:
Seems to be an issue when the device theme is set to light, in dark mode it looks to be fine.

PR #732, opened 2020-08-11
Fix configure widget theme colors

dshokouhi:
Fixes: #729 

Dark mode - service call widget
image

Light mode - service call widget
image

Dark mode - entity static widget
image

Light mode - entity static widget
image

Recommendation per: https://developer.android.com/guide/topics/ui/look-and-feel/darktheme#widgets_and_custom_notification_views

### DIFF
--- a/app/src/main/res/layout/fragment_mobile_app_integration.xml
+++ b/app/src/main/res/layout/fragment_mobile_app_integration.xml
@@ -58,7 +58,6 @@
 
         <androidx.appcompat.widget.AppCompatButton
             android:id="@+id/retry"
-            style="@style/Widget.HomeAssistant.Button.Colored"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:layout_marginTop="40dp"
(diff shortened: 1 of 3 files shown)


JBassett:
Isn't that style used during onboarding as well?  Are we affected by the same situation there?

dshokouhi:
Onboarding appears to be ok as the buttons have the blue background where they did not previously for the configure widget layouts

Light theme:
image

image

Dark theme:
image

image

dshokouhi:
@JBassett you are right, firebase retry screen has the same issue.  Let me remove from there.

dshokouhi:
@JBassett do you happen to know what screen this is found on? https://github.com/home-assistant/android/blob/master/app/src/main/res/layout/fragment_manual_setup.xml#L32

I was expecting to see a button label :)

JBassett:
> do you happen to know what screen this is found on?

That is the manual URL entry for connecting to a Home Assistant instance that you covered with the second set of screenshots.

Issue #429, opened 2020-02-14
Joining or leaving WiFi network should update wifi sensor immediately


SteveDinn:
**Home Assistant Android version:**
1.7.0 (116)

**Android version:**
10

**Phone model:**
OnePlus6

**Home Assistant version:**
0.105.2

**Last working Home Assistant release (if known):**
n/a

**Description of problem:**
When joining or leaving a WiFi network, the WiFi sensor should update immeidately. Changes in the attributes could remain on their current update schedule. It's useful to use the WiFi network as an indication of presence.

**Traceback (if applicable):**
```
```

**Screenshot of problem:**

**Additional information:**

SteveDinn:
Similar in scope to #391

PR #734, opened 2020-08-11
Update wifi sensor on state changes

dshokouhi:
Use a broadcast receiver to update the wifi sensor whenever the state has changed. This includes enabling/disabling and network changes.

Adds `is_wifi_on` attribute to the sensor

Fixes: #429

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -3,10 +3,12 @@ package io.homeassistant.companion.android
 import android.app.Application
 import android.content.Intent
 import android.content.IntentFilter
+import android.net.wifi.WifiManager
 import io.homeassistant.companion.android.common.dagger.AppComponent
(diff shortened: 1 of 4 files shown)


JBassett:
Was this tested without having location permissions?  I'm not sure what happens when you try to register a receiver when you don't have permissions.

dshokouhi:
@JBassett so in the emulator I cleared storage for the app.  Logged into my dev instance and did not grant location permssions.  Then I proceeded to turn wifi on/off and the app did not crash.

What I see in logcat is `2020-08-11 12:13:59.097 8055-8110/io.homeassistant.companion.android.debug W/NetworkSM: Tried getting wifi info without permission.` followed by an error for updating the sensors however the other sensors are able to update just fine.  The receiver is still working actually which doesn't make sense?  

Should I just add a permission check when we register the receiver to avoid this? I don't see any harm if we are able to trigger sensor updates for everything else when the wifi state changes.

JBassett:
No need for the permission check.  I'm working on a slight rework of the sensors to help eleviate some of the problems we are seeing with the dynamic updates which should take care of them being enabled vs disabled.

PR #739, opened 2020-08-12
Add Exoplayer view and interface

uvjustin:
See also https://github.com/home-assistant/frontend/pull/6606
H.265 videos are not supported through Chromium or Webview on Android. We can work around this for Android by using Exoplayer which supports H.265 if the codecs are available on the device. This PR defines an Exoplayer view for the Android client and uses the externalApp interface to help call it from the frontend.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -569,14 +569,16 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
         isInPictureInPictureMode: Boolean,
         newConfig: Configuration
     ) {
-        if (isInPictureInPictureMode) {
-            (decor.getChildAt(3) as FrameLayout).layoutParams.height =
-                FrameLayout.LayoutParams.MATCH_PARENT
-            decor.requestLayout()
-        } else {
-            if (decor.getChildAt(3) != null) {
-                (decor.getChildAt(3) as FrameLayout).layoutParams.height = videoHeight
+        if (exoPlayerView.visibility != View.VISIBLE) {
+            if (isInPictureInPictureMode) {
+                (decor.getChildAt(3) as FrameLayout).layoutParams.height =
+                    FrameLayout.LayoutParams.MATCH_PARENT
                 decor.requestLayout()
+            } else {
+                if (decor.getChildAt(3) != null) {
+                    (decor.getChildAt(3) as FrameLayout).layoutParams.height = videoHeight
+                    decor.requestLayout()
+                }
             }
         }
     }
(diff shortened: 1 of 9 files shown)


dshokouhi:
Awesome, how does the system handle the player in PiP mode? Not sure if that matters here :)

uvjustin:
Oh, I didn't take into account picture in picture mode. It's my first time looking at mobile app code, so not too familiar with everything. Is picture in picture when another app is playing video on top?

dshokouhi:
No worries and welcome!  Thanks for submitting the PR :) Yup exactly, for PiP just play the video in full screen then go to your home screen and it should activate it.

uvjustin:
Oh I see..didn't even realize that the app had that function....probably shouldn't be too hard to add that functionality in.

uvjustin:
Works with PiP now. Only issue is that the windowed size is not always the exact same size as the original window. I suspect it's because the resize call from javascript is tied to the window resize event, while the actual element sizing might get adjusted shortly after that. If it's an issue we could add a slight delay on the javascript side. From what I can tell, there is no good universal way to monitor the resizing of an element (although in addition to my first time with Kotlin this is my first time playing with JS so I could be wrong on this).

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -328,6 +362,53 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
...
+                                val rect = json.getJSONObject("payload")
+                                val displayMetrics = context.resources.displayMetrics
+                                exoLeft = (rect.getInt("left") * displayMetrics.density).toInt()
+                                exoTop = (rect.getInt("top") * displayMetrics.density).toInt()
+                                exoRight = (rect.getInt("right") * displayMetrics.density).toInt()
+                                exoBottom = (rect.getInt("bottom") * displayMetrics.density).toInt()
+                                runOnUiThread {
+                                    exoResizeLayout()
+                                }
+                            }
```
JBassett, line 411: Break these into functions rather than inline.

uvjustin:
> 
> 
> Just to confirm this is still backwards compatible with the existing streaming?

Yes, if the existing ways of playing (hls.js/native) come through the webview there should be no change in how they are executed. This will only launch if the externalApp interface gets called.

uvjustin:
Made the changes but had to rebase to resolve a merge conflict in Config.kt

JBassett:
@uvjustin once the frontend PR gets merged we can merge this.

uvjustin:
On the frontend PR, @balloob suggested changing the names of the bus calls and adding a status return. Pushed the changes here (also ran a code reformat because some of my lines were too long but it made a few other changes).

uvjustin:
Rebased to resolve merge conflict

balloob:
Frontend has been merged. Will be available in the next nightly.

PR #741, opened 2020-08-12
Add easier way to use of emoji in notifications

chriss158:
With this PR is easier to use emojis in notification. 
Just use the emoji shortcode and it will be parsed to the correct unicode code.
It is also possible to use fitzpatrick modifiers to change the skin tone of an emoji.

| Modifier | Type     |
| :------: | -------- |
|    🏻    | type_1_2 |
|    🏼    | type_3   |
|    🏽    | type_4   |
|    🏾    | type_5   |
|    🏿    | type_6   |

`:santa|type_6:`

Example:

```
message: "Now you can easy use emojis :santa|type_6::frog::sparkling_heart: in your notifications :sunglasses::rocket:"
title: "Easier use of Emojis :grinning::thumbsup:"
```

More on the project site of the library which is used:
https://github.com/vdurmont/emoji-java

Here are a lists of emoji short codes:
https://emojipedia.org/

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -117,6 +117,7 @@ dependencies {
     implementation(project(":domain"))
 
     implementation(Config.Dependency.Misc.blurView)
+    implementation(Config.Dependency.Misc.emoji)
 
     implementation(Config.Dependency.Kotlin.core)
     implementation(Config.Dependency.Kotlin.coroutines)
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -117,7 +117,9 @@ dependencies {
     implementation(project(":domain"))
 
     implementation(Config.Dependency.Misc.blurView)
-    implementation(Config.Dependency.Misc.emoji)
+    implementation(Config.Dependency.Misc.emoji)  {
+        exclude(group = "org.json", module = "json")
+    }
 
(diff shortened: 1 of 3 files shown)


chriss158:
What went wrong with the validate? On my dev pc i have no issues running linting.

JBassett:
```
  Errors found:
  
  /home/runner/work/android/android/app/build.gradle.kts:47: Error: json defines classes that conflict with classes now provided by Android. Solutions include finding newer versions or alternative libraries that don't have the same problem (for example, for httpclient use HttpUrlConnection or okhttp instead), or repackaging the library using something like jarjar. [DuplicatePlatformClasses]
          serviceCredentialsFile = "firebaseAppDistributionServiceCredentialsFile.json"
                                                                                  ~~~~
  ```

chriss158:
@JBassett can you help me here? I did not change anything on that. Just added the emoji library. Also android studio complains about that and suggest to suppress lint on that line. But `gradlew ktlintFormat` doesn't complain.

JBassett:
It's `gradlew lint` that is failing.  The dependency you added is pulling in `org.json:json` which is provided by Android and conflicting.  You need to exclude it.

chriss158:
Thanks for help. Is fixed now 👍

Issue #452, opened 2020-02-21
Add bluetooth status and connection sensor


jokerigno:
**Is your feature request related to a problem? Please describe.**
No problem. Just a new feature.

**Describe the solution you'd like**
Add support for BT status and paired device to use for automation:

- Send location when connected to car BT
- Enable/disable voice messages while using headphones
- Use BT for room location

**Describe alternatives you've considered, if any**
Nope

**Additional context**
That's all folks.

recklessnl:
> Use BT for room location

This is actually a great use case for this and could make room presence detection easier, especially if the sensor provides signal strength as well.

For example, let's say you have two floors. On both floors you have a low energy bluetooth device that your phone connects to (like a raspberry pi). Based on the name of the bluetooth connection AND the signal strength, you could create automations for that and guess in what room the phone is located.

This seems like it will make this type of presence detection more straightforward for users. Would be a great feature!

samueldefilippo:
It is also interesting to use the BT of the cell phone to detect the presence of other BT devices.
Like my machine at work.

uphillbattle:
I would like to make an automation that opens my garage door as I enter the "home" zone - but only if I'm in my car (it would be silly if the garage door opens as I come home after walking the dog). I believe a good way to distinguish between being in the car or on foot would be if my phone's Bluetooth is connected to the car's. This would be easily determined with a Bluetooth sensor in the HA companion app.

jurgenweber:
Just like @uphillbattle I would like to detect what car the person is in, presence automation would be fabulous also.

fti7:
I LOVE IT :heart_eyes: Thx @dshokouhi

PR #743, opened 2020-08-13
Add Bluetooth sensor

dshokouhi:
Fixes: #452 

- The state of the sensor will be the total number of connected devices including both paired and not paired devices
- Attributes will be `connected paired devices` `connected not paired devices` `paired devices` all provided in a list with the devices mac address. `is bt on` will reflect the state of the bluetooth adapter.
- The sensor will update when bluetooth state changes as well as connection changes are broadcasted.

Connected but not paired devices will include devices like a Mi Band or any other device that pairs with their own app instead of the system itself.  Bluetooth headsets are always paired with the system and so we can see whether they are paired as well as if they are connected.  Bluetooth names are not always given for a device so I have not included them.  There are other bluetooth things we can pull in like device battery (only supported on android 10) however we should think about how we want to expose them to the user.  For now lets just show what is connected and/or paired.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -4,6 +4,7 @@
     package="io.homeassistant.companion.android">
 
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+    <uses-permission android:name="android.permission.BLUETOOTH" />
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.VIBRATE" />
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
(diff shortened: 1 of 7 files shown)


jurgenweber:
I am super excited! 👍

Hypfer:
I think i've found a bug.
My Garmin Forerunner 235 appears as both a paired as well as not paired device:

image

Don't ask me why though.

Maybe we just need to filter all duplicates from that list?

dshokouhi:
Thanks @Hypfer I can submit a fix for this. We will only count it as a connected paired device in that case.  The mac address will still show up in both lists.  That is probably caused by the device being paired to the app and the android system itself.

jesserockz:
I got the same with my Mi Band, it is paired and connected to the system and connected via BLE to the app so shows up in both.

PR #746, opened 2020-08-13
Remove multiple sensor receivers

JBassett:
Remove PermissionManager.kt


skynetua:
Looks like you forgot to remove ChargingBroadcastReceiver.kt

JBassett:
@skynetua Fixed in a later PR!

PR #749, opened 2020-08-13
Fix status bar / navigation bar / app colors on dark mode

### LINKED ISSUE #725: Status bar is not readable when the Android Night Light function is enabled
**Home Assistant Android version:** 2.1.0-7

**Android version:** 10

**Phone model:** Samsung Galaxy S9

**Home Assistant version:** 0.113.3

**Last working Home Assistant release (if known):** n/a

**Description of problem:** Status bar is not readable when the Android Night Light function is enabled. 

**Screenshot of problem:**
image

**Additional information:** With the PR [#716](https://github.com/home-assistant/android/pull/716) there has been resolved an issue [#433 ](https://github.com/home-assistant/android/issues/433). However, now the status bar is not readable when using a light theme in HA and the Night Light function on the mobile. It seems like a decision about the font color on the status bar is being made based on the HA theme and not on the status bar background color.

chriss158:
Fixes #725

This PR fixes coloring of status bar / navigation bar. Problem was the force dark mode. Which tries automatically set an light theme to a dark theme. This "conversion" overwrites colors which are set from a custom theme via `setStatusBarAndNavigationBarColor`.

Solution: Use the DayNight theme. Then we must manually set dark theme colors. Positive side effect is, that on dark mode also the status bar / navigation bar is colored in a dark color.
If a custom theme is used in HA, we will need to overwrite the colors of the DayNight App theme by calling `setStatusBarAndNavigationBarColor`. If no custom theme is set, DayNight App theme colors are used.

The only downside of this implementation is to check if a custom theme in HA is used or not.
Right now, HA will return the default theme color of #03A9F4 if no custom theme is used. 
See here:
https://github.com/home-assistant/core/blob/ee64aafc3932ea0a7a76a33d1827db0c78fc0ed3/homeassistant/components/frontend/__init__.py#L362
Sadly not the best way to check this :/


dshokouhi:
The onboarding screen is very dark now

image

Light mode is ok

image

#732  may or may not have impacted this :)

chriss158:
Thanks fixed by defining missing colors 🎉

dshokouhi:
Looks good except for the service call widget, seems the icons need to be defined as well? Also not sure if this will be fixed in #745 

image

The theming issues in dark/light are fixed too! Good job!! 🎉

JBassett:
I wouldn't count on https://github.com/home-assistant/android/pull/745 for this next release.  Once we get the widget activities displaying correctly we should be good too go.

chriss158:
Hopefully now everything is covered for dark mode. Colored all of the alert dialogs, widget config and even the widgets are now colored in dark mode.

chriss158:
Images:

JBassett:
I'll take a look tonight!

dshokouhi:
Tested the debug APK locally and it looks great!!

PR #750, opened 2020-08-13
Add storage sensor

dshokouhi:
Adds a storage sensor, the state will be the percentage of free space so we align with the iOS app.

Attributes will be as follows:

- Free internal storage in GB
- Total internal storage in GB
- Free external storage in GB
- Total external storage in GB

External storage will reflect `No SD Card` if the device does not have a SD card.

This sensor will only update during the sensor update interval.

image

Also a fix for the bluetooth sensor to avoid adding duplicate devices to the total count of connected devices, and a proper `unit_of_measurement`. See: https://github.com/home-assistant/android/pull/743#issuecomment-673578475

This also fixes an issue introduced in #752 where sensors were only updating when first registered.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
@@ -14,7 +14,8 @@ class BluetoothSensorManager : SensorManager {
         private val bluetoothConnection = SensorManager.BasicSensor(
             "bluetooth_connection",
             "sensor",
-            "Bluetooth Connection"
+            "Bluetooth Connection",
+            unitOfMeasurement = "connection(s)"
         )
     }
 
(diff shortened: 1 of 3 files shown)


JBassett:
@dshokouhi hopefully that was the last time I change the interface for SensorManager....

app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
```diff
@@ -14,7 +14,8 @@ class BluetoothSensorManager : SensorManager {
         private val bluetoothConnection = SensorManager.BasicSensor(
             "bluetooth_connection",
             "sensor",
-            "Bluetooth Connection"
+            "Bluetooth Connection",
+            unitOfMeasurement = "connection(s)"
```
dshokouhi, line 18: Did I get this right?

Issue #751, opened 2020-08-14
Fatal crash while using the app debug and beta version


dshokouhi:
**Home Assistant Android version:**
Latest master branch changes
**Android version:**
10
**Phone model:**
Pixel 4 XL
**Home Assistant version:**
0.115dev
**Last working Home Assistant release (if known):**
N/A
**Description of problem:**

While testing other new features I keep seeing the following traceback.  It seems to occur randomly and not related to an intent.  It looks like both apps crash together, maybe this is a side effect of having both apps installed at the same time with the new listeners?

The changes I am testing do not involve new listeners so I don't think its related to anything I have done yet?

**Traceback (if applicable):**

```
2020-08-13 17:16:27.875 8784-8784/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: io.homeassistant.companion.android, PID: 8784
    java.lang.RuntimeException: Unable to start receiver io.homeassistant.companion.android.sensors.SensorReceiver: java.lang.ClassCastException: android.app.ReceiverRestrictedContext cannot be cast to io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
        at android.app.ActivityThread.handleReceiver(ActivityThread.java:3797)
        at android.app.ActivityThread.access$1400(ActivityThread.java:219)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1870)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
     Caused by: java.lang.ClassCastException: android.app.ReceiverRestrictedContext cannot be cast to io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
        at io.homeassistant.companion.android.sensors.SensorReceiver.onReceive(SensorReceiver.kt:48)
        at android.app.ActivityThread.handleReceiver(ActivityThread.java:3788)
        at android.app.ActivityThread.access$1400(ActivityThread.java:219) 
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1870) 
        at android.os.Handler.dispatchMessage(Handler.java:107) 
[...]

PR #761, opened 2020-08-14
Add light sensor

dshokouhi:
This PR adds a light sensor whose state will represent the current `lx` the device sees. If this sensor is enabled on a device that does not offer a light sensor the state will remain `unavailable`.

The light sensor will only update with the normal sensor update interval or when any other sensor that updates on a state change.

I did notice one odd issue in that on first launch the sensor may first show `unavailable` before updating to the actual value or we just need to restart the app.  This is a one time thing that I have noticed and once a real value gets reported those are what remain.  I suspect this is because these sensors require event listeners and we need to wait for them to be ready or we are reporting the value too early. This behavior is unlike what we see with the battery state sensor when we plug in the device.

image

There is also a fix for #751

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/LightSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/LightSensorManager.kt
@@ -1,15 +1,15 @@
 package io.homeassistant.companion.android.sensors
 
 import android.content.Context
+import android.content.Context.SENSOR_SERVICE
 import android.hardware.Sensor
 import android.hardware.SensorEvent
 import android.hardware.SensorEventListener
 import android.hardware.SensorManager.SENSOR_DELAY_NORMAL
-import androidx.appcompat.app.AppCompatActivity
 import io.homeassistant.companion.android.domain.integration.SensorRegistration
 import kotlin.math.roundToInt
 
-class LightSensorManager : SensorManager, SensorEventListener, AppCompatActivity() {
+class LightSensorManager : SensorManager, SensorEventListener {
     companion object {
 
         private const val TAG = "LightSensor"
@@ -75,10 +75,6 @@ class LightSensorManager : SensorManager, SensorEventListener, AppCompatActivity
                 lightReading = event.values[0].roundToInt().toString()
             }
         }
-    }
-
-    override fun onPause() {
-        super.onPause()
         mySensorManager.unregisterListener(this)
     }
 }
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/LightSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/LightSensorManager.kt
@@ -1,69 +1,72 @@
 package io.homeassistant.companion.android.sensors
 
 import android.content.Context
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/LightSensorManager.kt
```diff
@@ -0,0 +1,84 @@
...
+import android.content.Context
+import android.hardware.Sensor
+import android.hardware.SensorEvent
+import android.hardware.SensorEventListener
+import android.hardware.SensorManager.SENSOR_DELAY_NORMAL
+import androidx.appcompat.app.AppCompatActivity
+import io.homeassistant.companion.android.domain.integration.SensorRegistration
+import kotlin.math.roundToInt
+
+class LightSensorManager : SensorManager, SensorEventListener, AppCompatActivity() {
```
JBassett, line 12: I don't think we want this to extend an Activity with a sensor.  An activity should only be used to display things.

app/src/main/java/io/homeassistant/companion/android/sensors/LightSensorManager.kt
```diff
@@ -0,0 +1,84 @@
...
+            if (event.sensor.type == Sensor.TYPE_LIGHT) {
+                lightReading = event.values[0].roundToInt().toString()
+            }
+        }
+    }
+
+    override fun onPause() {
+        super.onPause()
+        mySensorManager.unregisterListener(this)
+    }
```
JBassett, line 83: This will never be called because the framework would normally call this when a user is navigating away from the view (activity) being displayed.  Since we just created this the framework knows nothing about it.

app/src/main/java/io/homeassistant/companion/android/sensors/LightSensorManager.kt
```diff
@@ -0,0 +1,84 @@
...
+        if (event != null) {
+            if (event.sensor.type == Sensor.TYPE_LIGHT) {
+                lightReading = event.values[0].roundToInt().toString()
+            }
+        }
+    }
+
+    override fun onPause() {
+        super.onPause()
+        mySensorManager.unregisterListener(this)
```
JBassett, line 82: Maybe we just unregister when we get the onSensorChanged event?

app/src/main/java/io/homeassistant/companion/android/sensors/LightSensorManager.kt
dshokouhi: Yea now that we don't need `onPause` I will remove this

app/src/main/java/io/homeassistant/companion/android/sensors/LightSensorManager.kt
dshokouhi: I was wondering about that :)

app/src/main/java/io/homeassistant/companion/android/sensors/LightSensorManager.kt
dshokouhi: Sounds perfect and still works as intended :)

Issue #763, opened 2020-08-15
Storage sensor not picking up my SD card


jesserockz:
**Home Assistant Android version:** 2.1.1-14-gf04f846-263-full

**Android version:** 10

**Phone model:** Motorola Moto One Vision

**Home Assistant version:** 0.114.0

**Last working Home Assistant release (if known):**

**Description of problem:**
New storage sensor not picking up my SD card storage.

**Traceback (if applicable):**

```

```

**Screenshot of problem:**

![Screenshot_20200815-131054](https://user-images.githubusercontent.com/3060199/90302286-3d9ace80-def9-11ea-8ff2-7fa4deda9bfe.jpg)

**Additional information:**
I have a 32GB SD card in my phone and FX File manager can see it just fine.

Does it have to do with HA Android not having storage permissions?

Related to #750

dshokouhi:
Could you please tell me the full path to the SD card for your device? I think certain manufacturers may use a different path than what is default.  I was talking to another user and he has a Samsung device with the same issue.  We may need to get a list of SD card paths so we can see which one is mounted.  Unfortunately the android emulator does not have a SD card and I only have Pixel devices here.  This one may be a little bit difficult since Google just deprecated the API that made this easy, for privacy concerns.

https://developer.android.com/reference/android/os/Environment#getExternalStorageDirectory()

Thanks for reporting the issue!

jesserockz:
Here is a screenshot:
![Screenshot_20200815-133201](https://user-images.githubusercontent.com/3060199/90302655-cadf2280-defb-11ea-8ff6-3985ce319d3c.jpg)

dshokouhi:
Do you have ADB access by any chance? I may have a fix for this but need to read the response.

in `adb shell` type in `printenv` then I need the full line that lists the path to your external storage.

You should see something similar to `EXTERNAL_STORAGE=/sdcard`

jesserockz:
Yeah, got these (removed the other unrelated lines)

```
SECONDARY_STORAGE=/storage/sdcard:/storage/usb1:/storage/usb2
ANDROID_STORAGE=/storage
EXTERNAL_STORAGE=/sdcard
```

dshokouhi:
darn didn't give the value we were expecting :( i'll need to spend sometime to think of a proper solution for manufacturers that override the default.

jesserockz:
I was just playing with it on my phone.

First point is that `Environment.isExternalStorageRemovable()` is returning false
[...]

PR #770, opened 2020-08-15
Attempt to fix SD card path in storage sensor

dshokouhi:
We were previously just using the default SD card path, unfortunately not all manufacturers use this path and override it.  To get around this we first grab all of the paths we can get, then we iterate through each one and check if its mounted and removable.  Only then will we send the data over on the SD card.

There is one caveat that if the user has a USB flash drive plugged in and a SD card then the reading may not be correct as the order in the array is never the same when it comes to external paths.  I think most users will only have a SD card inserted, I don't foresee having a flash drive plugged in while the phone is in the pocket.

I left some debug statements as well in case we don't capture all cases here so we can at least find a better solution.  Maybe dynamically add attributes for additional external storage devices?  I am not sure how we can differentiate between what is a USB drive and what is a SD card.

Fixes: #763 (hopefully)

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/StorageSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/StorageSensorManager.kt
@@ -3,6 +3,7 @@ package io.homeassistant.companion.android.sensors
 import android.content.Context
 import android.os.Environment
 import android.os.StatFs
+import android.util.Log
 import io.homeassistant.companion.android.domain.integration.SensorRegistration
 import java.io.File
 import kotlin.math.roundToInt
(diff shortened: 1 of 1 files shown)


jesserockz:
This is not showing my SD card storage still. It shows the same values as my internal storage. 

Log:
```
2020-08-16 10:06:29.073 21281-21350/io.homeassistant.companion.android.debug D/StorageSensor: PATHS SD 2
2020-08-16 10:06:29.079 21281-21350/io.homeassistant.companion.android.debug D/StorageSensor: PATH /storage/emulated/0/Android/data/io.homeassistant.companion.android.debug/files is mounted true and removable is false
2020-08-16 10:06:29.096 21281-21350/io.homeassistant.companion.android.debug D/StorageSensor: PATH /storage/8BF3-12EB/Android/data/io.homeassistant.companion.android.debug/files is mounted true and removable is true
```

app/src/main/java/io/homeassistant/companion/android/sensors/StorageSensorManager.kt
```diff
@@ -59,12 +76,12 @@ class StorageSensorManager : SensorManager {
...
+        externalPath = externalMemoryAvailable(context)
         var totalExternalStorage = "No SD Card"
         var freeExternalStorage = "No SD Card"
 
-        if (hasExternalStorage) {
-            val pathSD = context.getExternalFilesDir(null)
+        if (externalPath != null) {
+            val pathSD = context.getExternalFilesDir(externalPath.toString())
             val statSD = StatFs(pathSD?.path)
             blockSizeSD = statSD.blockSizeLong
```
jesserockz, line 86: While debugging:
`val pathSD = context.getExternalFilesDir(externalPath.toString())` becomes `/storage/emulated/0/Android/data/io.homeassistant.companion.android.debug/files/storage/8BF3-12EB/Android/data/io.homeassistant.companion.android.debug/files`

I was able to get the correct numbers with this: 

```suggestion
            val statSD = StatFs(externalPath.toString())
            blockSizeSD = statSD.blockSizeLong
```

dshokouhi:
Thanks for all your help @jesserockz !!

Issue #756, opened 2020-08-14
Add separator between entity state widget state and attribute


klejejs:
**Is your feature request related to a problem? Please describe.**
Currently entity state and attribute in entity state widget are just appended in one sentence.

**Describe the solution you'd like**
An option in setup screen to choose a separator to put between entity state and attribute in the widget.

**Additional context**
I know all of the widget issues I have submitted are very simple fixes and I would be glad to fix them, but, as I have never used Kotlin before, I trust someone else could do that. It would be much easier and quicker for them, than for me to setup the environment and understand the language. Thank you in advance. ;)

dshokouhi:
@klejejs I actually just got started with Kotlin....its not that bad if you want to dive in :)

klejejs:
@dshokouhi Ok, if you say so, I might try to dive in this weekend.

Issue #754, opened 2020-08-14
Add possibility to resize entity state widget text


klejejs:
**Is your feature request related to a problem? Please describe.**

Currently my entity state widget has to be really large, as the font size is huge.

**Describe the solution you'd like**

Set font size based on width or height of the widget or just add option in setup screen to set font size.

PR #771, opened 2020-08-15
Added multiple entity state widget options

klejejs:
This PR adds entity state widget option to choose text size for the widget #754 and choose separator between entity state and attribute #756


JBassett:
@klejejs  Sorry, I made some changes to use room instead of preferences because it was more of a hack to store the widget info.  Once your resolve the conflicts I can review,

klejejs:
@JBassett is there a way to check if the migration is correct even after reinstalling the app? I think I messed up the DB, so I reinstalled the app and I obviously didn't have any problems after that.

JBassett:
What I've been doing to validate is build and run the debug build from master and log in.  Then I check my branch back out with the migration and run it.  That should cause your migration to run  and you can validate the results.

We should add tests to the migrations to make it even better https://developer.android.com/training/data-storage/room/migrating-db-versions#test.  If you don't want to add the functionality in your PR I understand and I can add it in the future.

klejejs:
@JBassett I manually checked migrations and it should be working now. I also looked into testing them, but concluded that it is out of scope for this PR. I unfortunately am not skilled enough to properly implement it.

app/src/main/java/io/homeassistant/companion/android/database/widget/StaticWidgetEntity.kt
```diff
@@ -13,5 +13,9 @@ data class StaticWidgetEntity(
     @ColumnInfo(name = "attribute_id")
     val attributeId: String?,
     @ColumnInfo(name = "label")
-    val label: String?
+    val label: String?,
+    @ColumnInfo(name = "text_size")
+    val textSize: Int?,
+    @ColumnInfo(name = "separator")
+    val separator: String?
```
JBassett, line 20: Based on your other code it looks like these are required and always have a value, why have them nullable?

app/src/main/java/io/homeassistant/companion/android/migrations/Migrations.kt
```diff
@@ -14,7 +14,7 @@ class Migrations constructor(
         private const val TAG = "Migrations"
         private const val PREF_NAME = "migrations"
         private const val PREF_VERSION = "migration_version"
-        private const val LATEST_VERSION = 3
+        private const val LATEST_VERSION = 4
```
JBassett, line 17: This version just happened to match the DB version, if you aren't adding a non database migration you don't need to increment.

app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidget.kt
```diff
@@ -79,9 +82,18 @@ class StaticWidget : AppWidgetProvider() {
             val entityId: String? = widget?.entityId
```
JBassett, line 82: Maybe we should put the rest of this method inside a null check for `widget`?  Then not everything will be nullable here and make the code easier

app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidgetConfigureActivity.kt
```diff
@@ -163,6 +158,16 @@ class StaticWidgetConfigureActivity : Activity() {
                 label.text.toString()
             )
 
+            intent.putExtra(
+                StaticWidget.EXTRA_TEXT_SIZE,
+                textSize.text.toString()
```
JBassett, line 163: Shouldn't this be a number? `Float` or `Int`?

app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidget.kt
```diff
@@ -158,7 +173,9 @@ class StaticWidget : AppWidgetProvider() {
                 appWidgetId,
                 entitySelection,
                 attributeSelection,
-                labelSelection
+                labelSelection,
+                textSizeSelection?.toInt() ?: 30,
```
JBassett, line 177: Should probably use `toIntOrNull()` so we don't throw and exception if it's not a number somehow.

JBassett, line 177: Should probably use `toIntOrNull()` so we don't throw and exception if it's not a number somehow.

app/src/main/java/io/homeassistant/companion/android/database/widget/StaticWidgetEntity.kt
klejejs: Makes sense. I hadn't fully understood what question mark in Kotlin means. Fixed now.

app/src/main/java/io/homeassistant/companion/android/migrations/Migrations.kt
klejejs: Got it. Fixed.

app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidget.kt
klejejs: Also makes sense. Fixed that.

app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidgetConfigureActivity.kt
klejejs: It is a number and I changed it to float in DB, however we are using a text field for input. Yes, you can only input numbers as specified in xml but as it is text field, it has to be a String. It is later converted to float when saving to DB.

app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidget.kt
klejejs: Fixed.

app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidget.kt
```diff
@@ -158,7 +173,9 @@ class StaticWidget : AppWidgetProvider() {
                 appWidgetId,
                 entitySelection,
                 attributeSelection,
-                labelSelection
+                labelSelection,
+                textSizeSelection?.toFloatOrNull() ?: 30F,
+                if (separatorSelection == "") " " else separatorSelection ?: " "
```
klejejs, line 178: @JBassett I think if user never touches the text field, the default value is an empty string not null, so can I write like this?

klejejs:
@JBassett Thank you for doing the code review. I know my code is not perfect, but I am learning to get better ;)

PR #772, opened 2020-08-16
Support opening NFC Settings from external bus.

JBassett:
Takes advantage of new NFC opening from frontend.
https://github.com/home-assistant/developers.home-assistant/pull/621

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -50,6 +50,7 @@ import io.homeassistant.companion.android.authenticator.Authenticator
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
 import io.homeassistant.companion.android.database.AppDatabase
 import io.homeassistant.companion.android.database.authentication.Authentication
+import io.homeassistant.companion.android.nfc.NfcSetupActivity
 import io.homeassistant.companion.android.onboarding.OnboardingActivity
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/nfc/NfcWelcomeFragment.kt
```diff
@@ -46,7 +46,8 @@ class NfcWelcomeFragment : Fragment() {
         }
 
         btn_nfc_write.setOnClickListener {
-            findNavController().navigate(R.id.action_NFC_WRITE)
+//            findNavController().navigate(R.id.action_NFC_WRITE)
```
balloob, line 49: Commented code

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -328,6 +334,12 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
                             "config_screen/show" -> startActivity(
                                 SettingsActivity.newInstance(this@WebViewActivity)
                             )
+                            "tag/write" -> startActivity(
```
balloob, line 337: Just a note that I just updated the proposal. We're going to respond with an empty object here to indicate when it's done.

JBassett: Done.

balloob (approved):

balloob:
Think this is good to go?

JBassett:
> Think this is good to go?

Would be nice to validate with the frontend changes but I think so.

Issue #774, opened 2020-08-16
Next Alarm State Unavailable, Local Time empty


holdestmade:
**Home Assistant Android version:**
2.1.1-18-grc6610a-267-full
**Android version:**
10
**Phone model:**
Huawei P20
**Home Assistant version:**
0.113.1
**Last working Home Assistant release (if known):**
na
**Description of problem:**
The Next Alarm sensor always shows as `Unavailable`, even with an alarm set. The `Time in Milliseconds` attribute does populate though but `Local Time` and `Package` does not
**Traceback (if applicable):**

```

```

**Screenshot of problem:**

**Additional information:**

dshokouhi:
What app did you use to set a alarm? Also what time for `Time in Milliseconds` do you see? Is it just 0?

holdestmade:
Just the built in clock app. The Time in Milliseconds changes from 0 to a large number of milliseconds to the alarm, this is for 8am tomorrow: 1597647600000

dshokouhi:
Ok I wonder if this is because a manufacturer override.  Are you able to get the logcat details? I think you may be hitting an exception after we grab the trigger time.

holdestmade:
Not sure what logcat is ?

dshokouhi:
Your time in milliseconds is correct...mine is `1597756500000` and its in 2 days..

Logcat is how you get the android logs so we can see what the cause of the error is.  Currently we are getting the correct trigger time however the call to get the package name is failing and thats why we dont see a proper state and other data.  What is the built in clock app? Its not the Google Clock is it?

holdestmade:
It's just the Clock app that came with the phone

holdestmade:
The milliseconds is right if you take the last 4 zeros off

dshokouhi:
We are converting that value to UTC time and not local time.  Local time happens elsewhere.  See below where it converts it properly

![image](https://user-images.githubusercontent.com/1634145/90338153-4e505080-df9c-11ea-93b3-d4b50183e811.png)

holdestmade:
Ah, got ya

dshokouhi:
Can you run a quick test and see if using this app to set the alarm time works better? https://play.google.com/store/apps/details?id=com.google.android.deskclock&hl=en_US

Trying to see why we get one value and not the others.

holdestmade:
Yes, works OK with that one:

`2020-08-17T07:30:00.000Z`

[...]

PR #775, opened 2020-08-16
Change order of package for apps that don't provide it

dshokouhi:
Fixes: #774 

Turns out not all apps provide the `package` in the AlarmClockInfo intent and thus fails the call.  Lets move `package` to the end of the logic so the state and other attributes populate accordingly.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NextAlarmManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NextAlarmManager.kt
@@ -67,8 +67,7 @@ class NextAlarmManager : SensorManager {
                 sdf.timeZone = TimeZone.getTimeZone("UTC")
                 utc = sdf.format(Date(triggerTime))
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/NextAlarmManager.kt
```diff
@@ -67,6 +66,8 @@ class NextAlarmManager : SensorManager {
                 val sdf = SimpleDateFormat(dateFormat)
                 sdf.timeZone = TimeZone.getTimeZone("UTC")
                 utc = sdf.format(Date(triggerTime))
+
+                pendingIntent = alarmClockInfo.showIntent.creatorPackage.toString()
```
JBassett, line 70: Lets make it null safe rather than let it throw an exception every time.
```suggestion
                pendingIntent = alarmClockInfo.showIntent?.creatorPackage?: "Unknown"

```

PR #780, opened 2020-08-16
Add themes selector

### LINKED ISSUE #773: Dark mode is enabled even if light mode was selected
**Home Assistant Android version:** 2.2.0

**Android version:** 10

**Phone model:** Google Pixel 3

**Home Assistant version:** 0.114.1

**Last working Home Assistant release (if known):** None

**Description of problem:** 
Dark mode is enabled even if light mode was selected

```

```

**Screenshot of problem:**
Screenshot_20200816-122502

**Additional information:**

chriss158:
Fixes #773

This PR adds a theme selector to the preferences activity.

Now a user can use a light theme in HA and the dark mode of the system combined. He can easy set the theme of the app to "Light".

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
@@ -8,6 +8,7 @@ import androidx.preference.EditTextPreference
 import androidx.preference.Preference
 import androidx.preference.PreferenceFragmentCompat
 import androidx.preference.SwitchPreference
+import androidx.preference.ListPreference
 import io.homeassistant.companion.android.BuildConfig
 import io.homeassistant.companion.android.DaggerPresenterComponent
 import io.homeassistant.companion.android.PresenterModule
@@ -19,6 +20,7 @@ import io.homeassistant.companion.android.sensors.SensorsSettingsFragment
 import io.homeassistant.companion.android.settings.shortcuts.ShortcutsFragment
 import io.homeassistant.companion.android.settings.ssid.SsidDialogFragment
 import io.homeassistant.companion.android.settings.ssid.SsidPreference
+import io.homeassistant.companion.android.util.ThemeHandler
 import javax.inject.Inject
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 
@@ -48,6 +50,13 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
         preferenceManager.preferenceDataStore = presenter.getPreferenceDataStore()
 
         setPreferencesFromResource(R.xml.preferences, rootKey)
+
+        val onChooseTheme = Preference.OnPreferenceChangeListener { preference, newValue ->
+            val index = (preference as ListPreference).findIndexOfValue(newValue as String?)
+            preference.summary = preference.entries[index]
+            ThemeHandler.setNightModeBaseOnTheme(newValue as String)
+            true
+        }
 
         val onChangeUrlValidator = Preference.OnPreferenceChangeListener { _, newValue ->
             val isValid = newValue.toString().isBlank() || newValue.toString().toHttpUrlOrNull() != null
@@ -99,6 +108,15 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
                 true
             }
         }
+
+        removeSystemFromThemesIfNeeded()
+
+        findPreference<Preference>("themes")?.let {
+            it.summary = (it as ListPreference).entry
+        }
+
+        findPreference<Preference>("themes")?.onPreferenceChangeListener =
+            onChooseTheme
 
         findPreference<EditTextPreference>("connection_internal")?.onPreferenceChangeListener =
             onChangeUrlValidator
@@ -163,5 +181,24 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
     private fun authenticationResult(result: Int) {
         val switchLock = findPreference<SwitchPreference>("app_lock")
         switchLock?.isChecked = result == Authenticator.SUCCESS
+    }
+
+    private fun removeSystemFromThemesIfNeeded() {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
+            val pref = findPreference<ListPreference>("themes")
+            if (pref != null) {
+                val systemIndex = pref.findIndexOfValue("system")
+                if (systemIndex > 0) {
+                    var entries = pref.entries?.toMutableList()
+                    entries?.removeAt(systemIndex)
+                    var entryValues = pref.entryValues?.toMutableList()
+                    entryValues?.removeAt(systemIndex)
+                    if (entries != null && entryValues != null) {
+                        pref.entries = entries.toTypedArray()
+                        pref.entryValues = entryValues.toTypedArray()
+                    }
+                }
+            }
+        }
     }
 }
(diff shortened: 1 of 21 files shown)


app/src/main/res/values-de/strings.xml
```diff
@@ -109,4 +109,22 @@
   <string name="widget_text_hint_service_data">Entitäts-ID</string>
   <string name="widget_text_hint_service_domain">Domain</string>
   <string name="widget_text_hint_service_service">Dienst</string>
+  <string name="themes_title_settings">Theme</string>
```
dshokouhi, line 112: I think localize is supposed to pick these up for us

dshokouhi:
Tested the debug APK locally looks great!

app/src/main/res/values-de/strings.xml
JBassett: @dshokouhi  is correct, you need to add any other language that English to Lokalise otherwise they get overridden when we sync the other languages.

app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -100,6 +109,15 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
             }
         }
 
+        removeSystemFromThemesIfNeeded()
+
+        findPreference<Preference>("themes")?.let {
+            it.summary = (it as ListPreference).entry
+        }
```
JBassett, line 116: Can we just use `app:useSimpleSummaryProvider="true"` in the preferences.xml?

app/src/main/res/values-de/strings.xml
chriss158: Ok. Just done it because it was easy to translate right here. I will remove the the translation.

domain/src/main/java/io/homeassistant/companion/android/domain/themes/ThemesUseCaseImpl.kt
```diff
@@ -0,0 +1,17 @@
+package io.homeassistant.companion.android.domain.themes
+
+import javax.inject.Inject
+
+class ThemesUseCaseImpl @Inject constructor(
+    private val themesRepository: ThemesRepository
+) : ThemesUseCase {
+    override suspend fun getCurrentTheme(): String? {
+        val theme = themesRepository.getCurrentTheme()
+        return if (theme.isNullOrEmpty()) "light"
```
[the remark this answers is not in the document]
chriss158: I thought about that too. But the light theme is available on nearly all systems. Follow system is only available from Android Pie upwards. And even on Pie the switch is hidden in the developer settings.

app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
chriss158: Ah perfect. I've tried to use this, but android studio didn't suggested me this setting in the autocompletion. So i thought, this property is not supported by ListPreference. But according to the android docs it is supperted :+1: I will change that.

domain/src/main/java/io/homeassistant/companion/android/domain/themes/ThemesUseCaseImpl.kt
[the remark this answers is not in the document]
dshokouhi: I also feel that it should be follow system by default, maybe we just need to use follow system on devices  that support it?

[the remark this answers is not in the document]
chriss158: Good idea 👍  Change that to follow system for >= Pie

app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
chriss158: I've changed to `app:useSimpleSummaryProvider`. But i needed also to change to use `onSharedPreferenceChanged `because `OnPreferenceChangeListener `triggers before the the value is written to shared preferences. 
If i use `OnPreferenceChangeListener` to trigger theme change, the activity gets reset and the new value never gets written to share prefrences.

JBassett:
@chriss158  Looks like you still have some conflicts, once those are resolved I'll take a look.

chriss158:
On light theme the widget icon is now not visible. Dark theme is ok. Will have a look on this tonight.

chriss158:
Color is fixed.

data/src/main/java/io/homeassistant/companion/android/data/themes/ThemesRepositoryImpl.kt
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    override suspend fun getCurrentTheme(): String {
+        val theme = localStorage.getString(PREF_THEME)
+        return if (theme.isNullOrEmpty()) {
+            if (osVersion.toInt() >= P) {
+                "system"
+            } else "light"
+        } else theme
+    }
```
JBassett, line 25: Now that we are adding a version check into the default I think we would want the default to come from `SettingsPresenterImpl.kt`.  Either return `default` when unset or make the return nullable then deal with it in the presenter.

app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -165,4 +180,31 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
...
+                    if (entries != null && entryValues != null) {
+                        pref.entries = entries.toTypedArray()
+                        pref.entryValues = entryValues.toTypedArray()
+                    }
+                }
+            }
+        }
+    }
+
+    override fun onSharedPreferenceChanged(sharedPreferences: SharedPreferences?, key: String?) {
```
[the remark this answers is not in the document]
chriss158: `ThemeHandler.setNightModeBaseOnTheme` will reload the activity duo to night mode change. `OnPreferenceChangeListener`
will trigger after change of preference but before write to shared preference. If we change the theme here, the  preference is never written to shared preference, because of the reset of the activity.

See here: https://stackoverflow.com/questions/51256723/onpreferencechangelistener-fires-before-change-is-made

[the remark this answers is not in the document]
JBassett: Can we do it after we persist the setting in the SettingsPresenter here then:
https://github.com/home-assistant/android/pull/780/files#diff-f13f037aed758b996ee8da1d6b628c06R78

data/src/main/java/io/homeassistant/companion/android/data/themes/ThemesRepositoryImpl.kt
chriss158: `getCurrentTheme()` is also called in the `WebViewPresenterImpl`. If i implement the check on `SettingsPresenterImpl`, i also need to implement it in `WebViewPresenterImpl`. Any suggestion how to implement this only one time? Maybe in the ThemesUseCase?

app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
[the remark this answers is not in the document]
chriss158: Yep this does work 👍

data/src/main/java/io/homeassistant/companion/android/data/themes/ThemesRepositoryImpl.kt
JBassett: Put the logic in ThemeHandler maybe?

chriss158:
Implemented your changes with the ThemeManger which does all the theme stuff

PR #786, opened 2020-08-19
Enhance PhoneStateSensorManager with carrier information sensors

### LINKED ISSUE #768: Add carrier information as sensor
Trigger different notifications based on carrier. Useful for roaming and multi carriers such as Google Fi.
sensor with carrier name / ID
since we have phone state this should be able to be retrieved as per https://developer.android.com/reference/android/telephony/SubscriptionManager?hl=pt-br#getActiveSubscriptionInfoList%28%29
**Additional context**

skynetua:
Fixes #768 

PR enhance current PhoneStateSensorManager with 2 new sensors SIM_1 and SIM_2. 
Untitled

Sensors require api level 22 or above. Moreover attributes {"carrier id", "mcc", "mnc", "is opportunistic", "data roaming"} require api 29+

I didn't add OnSubscriptionsChangedListener so sensors will be updated only on regular basis

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
@@ -2,6 +2,9 @@ package io.homeassistant.companion.android.sensors
 
 import android.Manifest
 import android.content.Context
+import android.os.Build
+import android.telephony.SubscriptionInfo
+import android.telephony.SubscriptionManager
 import android.telephony.TelephonyManager
 import io.homeassistant.companion.android.domain.integration.SensorRegistration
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
```diff
@@ -14,13 +17,25 @@ class PhoneStateSensorManager : SensorManager {
...
+            "SIM_2"
+        )
     }
 
     override val name: String
         get() = "Phone Sensors"
 
     override val availableSensors: List<SensorManager.BasicSensor>
-        get() = listOf(phoneState)
+        get() = listOf(phoneState, sim_1, sim_2)
```
JBassett, line 38: We should check the version here and only add `sim_1` and `sim_2` if the version of android supports them.  Otherwise they will show up for older versions but crash the app if they try to enable!

skynetua: Yes, my bad

PR #794, opened 2020-08-20
Add last reboot sensor

dshokouhi:
This PR adds a last reboot sensor that is a timestamp of the time the device last booted up.  It was mentioned in #792 that steps are reset only during the last reboot so I thought this might be helpful.  To match the Next Alarm sensor attributes for local time and time in milliseconds are also provided.

image


Hypfer:
This sensor is flapping for me.

image
image
image
image
image



Not sure what is going on there exactly. Rounding errors?

dshokouhi:
@Hypfer please create a bug, merged PR's are not a place to report issues.

PR #801, opened 2020-08-21
Add proximity sensor

dshokouhi:
This PR adds a proximity sensor which will update during the sensor update interval. Based on the [docs](https://developer.android.com/guide/topics/sensors/sensors_position#sensors-pos-prox) not all devices report the actual distance in a real value. 

```
Note: Some proximity sensors return binary values that represent "near" or "far." In this case, the sensor usually reports its maximum range value in the far state and a lesser value in the near state. Typically, the far value is a value > 5 cm, but this can vary from sensor to sensor. You can determine a sensor's maximum range by using the getMaximumRange() method.
```

 Some devices, like my Pixel 4 XL only report 2 values either `5` or `0` for `far` or `near` respectively. We account for this difference and the state will be `far` or `near` only on those devices.  For devices whose `maximumRange` is not 5 we will continue to show the raw data.

I am not certain how this will work on other devices but on devices that do not have a proximity sensor the state remains as `unavailable`.  We may need to get some feedback from users on this as there very well can be differences between manufacturers.  For my test I basically put my hand over the sensor to get a `near` reading and taking away my hand would generate a `far` reading.  To my understanding some device manufacturers use a combination of this binary reading along with the light sensor value to really judge if something is close by or not.

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/ProximitySensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/ProximitySensorManager.kt
@@ -1,0 +1,87 @@
+package io.homeassistant.companion.android.sensors
+
+import android.content.Context
+import android.content.Context.SENSOR_SERVICE
+import android.hardware.Sensor
+import android.hardware.SensorEvent
+import android.hardware.SensorEventListener
+import android.hardware.SensorManager.SENSOR_DELAY_NORMAL
+import io.homeassistant.companion.android.domain.integration.SensorRegistration
+import kotlin.math.roundToInt
+
+class ProximitySensorManager : SensorManager, SensorEventListener {
+    companion object {
+
+        private const val TAG = "ProximitySensor"
+        private val proximitySensor = SensorManager.BasicSensor(
+            "proximity_sensor",
+            "sensor",
+            "Proximity Sensor"
+        )
+        private var proximityReading: String = "unavailable"
+        lateinit var mySensorManager: android.hardware.SensorManager
+    }
+
+    override val name: String
+        get() = "Proximity Sensors"
+
+    override val availableSensors: List<SensorManager.BasicSensor>
+        get() = listOf(proximitySensor)
+
+    override fun requiredPermissions(): Array<String> {
+        return emptyArray()
+    }
+
+    override fun getSensorData(
+        context: Context,
+        sensorId: String
+    ): SensorRegistration<Any> {
+        return when (sensorId) {
+            proximitySensor.id -> getProximitySensor(context)
+            else -> throw IllegalArgumentException("Unknown sensorId: $sensorId")
+        }
+    }
+
+    private fun getProximitySensor(context: Context): SensorRegistration<Any> {
+
+        mySensorManager = context.getSystemService(SENSOR_SERVICE) as android.hardware.SensorManager
+
+        val proximitySensors = mySensorManager.getDefaultSensor(Sensor.TYPE_PROXIMITY)
+        if (proximitySensors != null) {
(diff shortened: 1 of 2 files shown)


skynetua:
Tested it on my samsung:
Sensor remains `near` all the time while no active calls
It changes to `far` on phone call triggered due to sensors update requested by phone_state_changed intent. But does not change to `near` when you actually pick up the phone. 
I think sensor can be useful only with immediate update.

dshokouhi:
Yea there is no receiver for the proximity sensor, most of the hardware sensors are like that.  Once we can allow the sensor worker to update more frequently I think these sensors will be better. And also if its really 5cm proximity for some devices it may not make sense, devices that report the true proximity will make more sense :)

app/src/main/java/io/homeassistant/companion/android/sensors/ProximitySensorManager.kt
```diff
@@ -0,0 +1,87 @@
...
+
+        val proximitySensors = mySensorManager.getDefaultSensor(Sensor.TYPE_PROXIMITY)
+        if (proximitySensors != null) {
+            mySensorManager.registerListener(
+                this,
+                proximitySensors,
+                SENSOR_DELAY_NORMAL)
+
+            // Some devices only report 2 values, one of which is the max range so lets account for those devices
+            if (proximitySensors.maximumRange.roundToInt() == 5) {
```
JBassett, line 57: We should just check if the value for max == sensor value.  I didn't see anywhere in the documentation that says five is the magic value.  We should just check if they are equal to determine if they're far.

dshokouhi: Ok so I can change this to `proximitySensors.maximumRange.roundToInt().toString() == proximityReading` so we verify the maximum range is the value we just got, this also simplifies the condition statement :)

I got the number 5 from : https://developer.android.com/guide/topics/sensors/sensors_position

At the bottom of the section we see them call out `5` in the note

```
Note: Some proximity sensors return binary values that represent "near" or "far." In this case, the sensor usually reports its maximum range value in the far state and a lesser value in the near state. Typically, the far value is a value > 5 cm, but this can vary from sensor to sensor. You can determine a sensor's maximum range by using the getMaximumRange() method.
```

dshokouhi: @JBassett so the problem with checking if the max range == sensor value is that all devices will only report the binary values then.  I think since google mentioned "typically the far value is a value > 5 cm" is a good indication of what to expect here.  Otherwise we can just keep things binary which doesn't make much sense if the device supports a real reading.  I am ok either way here since my device only reports binary values.  Do you have any other suggestions for this?

pkern:
Would you mind disabling this by default? Proximity sensors are quite often very visible sensors (if they can be accessed at all). If they trigger, it can be very distracting - depending on the device's display technology - and the data is very likely only useful to a tiny fraction of the app's user base. After all on some devices you only get a 1-2 cm reading, which is pretty much only useful to see if you're in a phone call or not. Home Assistant permanently holding onto the proximity sensor then means a permanently flickering dot in any app.

dshokouhi:
@pkern we only query the sensor during the update interval, we do not do any hold on the sensor after the worker is done.  Currently there is no way to keep sensors off by default unless they require device permissions.  

Also, in the future please don't comment on merged PR's and instead open a new issue or feature request.

pkern:
I do have evidence that it is continuously on until the app is force stopped. But unfortunately I cannot report a proper bug, so I opened FR #856.

PR #809, opened 2020-08-25
Sensor Refactor

JBassett:
In this PR we now persist all sensor data to the local DB.  This allows us to keep the UI and Home Assistant better in sync.  This also allows sensors to return data asynchronously and allows for more reliable update with those sensors.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
@@ -4,12 +4,14 @@ import android.content.ContentValues
 import android.content.Context
 import android.util.Log
 import androidx.room.Database
+import androidx.room.OnConflictStrategy
 import androidx.room.Room
 import androidx.room.RoomDatabase
 import androidx.room.migration.Migration
 import androidx.sqlite.db.SupportSQLiteDatabase
 import io.homeassistant.companion.android.database.authentication.Authentication
 import io.homeassistant.companion.android.database.authentication.AuthenticationDao
+import io.homeassistant.companion.android.database.sensor.Attribute
 import io.homeassistant.companion.android.database.sensor.Sensor
 import io.homeassistant.companion.android.database.sensor.SensorDao
 import io.homeassistant.companion.android.database.widget.ButtonWidgetDao
@@ -21,13 +23,14 @@ import io.homeassistant.companion.android.database.widget.TemplateWidgetEntity
 
 @Database(
     entities = [
+        Attribute::class,
         Authentication::class,
         Sensor::class,
         ButtonWidgetEntity::class,
         StaticWidgetEntity::class,
         TemplateWidgetEntity::class
     ],
-    version = 6
+    version = 7
 )
 abstract class AppDatabase : RoomDatabase() {
     abstract fun authenticationDao(): AuthenticationDao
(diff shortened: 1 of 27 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
```diff
@@ -15,6 +16,31 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
...
+                            else
+                                it.summary = sensorEntity.state + " " + basicSensor.unitOfMeasurement
+                            // TODO: Add the icon from mdi:icon?
+                        } else {
+                            it.summary = "Disabled"
+                        }
+                    }
+                }
+            }
+            handler.postDelayed(this, 10000)
```
dshokouhi, line 40: Does this mean if we keep the settings screen open we can get updates every 10 seconds into HA?

JBassett: That is correct!  We can consider doing something similar to the iOS app add add this to the webview activity as well to get more consistent updates when the application is open.

dshokouhi: That would be amazing for wall mounted tablets!!!

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
```diff
@@ -105,18 +110,12 @@ class PhoneStateSensorManager : SensorManager {
...
-            if (slotIndex == 1) {
-                return sim_2.toSensorRegistration(
-                    displayName,
-                    "mdi:sim",
-                    attrs
-                )
-            }
-            return sim_1.toSensorRegistration(
+            onSensorUpdated(context,
+                sim_1,
```
skynetua, line 114: There should be `basicSimSensor` i'm not mistaken

app/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
```diff
@@ -124,5 +128,32 @@ abstract class AppDatabase : RoomDatabase() {
...
+                        it.put("device_class", "")
+                    })
+                }
+                database.execSQL("DROP TABLE IF EXISTS `sensors`")
+                database.execSQL("CREATE TABLE IF NOT EXISTS `sensors` (`id` TEXT NOT NULL, `enabled` INTEGER NOT NULL, `registered` INTEGER NOT NULL, `state` TEXT NOT NULL, `state_type` TEXT NOT NULL, `type` TEXT NOT NULL, `icon` TEXT NOT NULL, `name` TEXT NOT NULL, `device_class` TEXT, `unit_of_measurement` TEXT, PRIMARY KEY(`id`))")
+                sensors.forEach {
+                    database.insert("sensors", OnConflictStrategy.REPLACE, it)
+                }
+
+                database.execSQL("CREATE TABLE IF NOT EXISTS `sensor_attributes` (`sensor_id` TEXT NOT NULL, `name` TEXT NOT NULL, `value` TEXT NOT NULL, PRIMARY KEY(`sensor_id`, `name`))")
```
klejejs, line 155: Shouldn't the cursor be closed or does it close automatically?

JBassett: Good catch, I thought I read somewhere you didn't need to close it but I'll close it anyway.

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
JBassett: Good catch!

Issue #807, opened 2020-08-24
Add DND switch to get & set Do Not Disturb status


os131:
**Is your feature request related to a problem? Please describe.**
I'm setting DND each evening and remove it as soon as I wake up. I wish it could be automated using HA. My phone have no possibilities to schedule DND.

**Describe the solution you'd like**
Just add a switch for the purpose

**Describe alternatives you've considered, if any**
I can use tasker to do it but I do prefere grouping every automatioon in one place

**Additional context**

dshokouhi:
I think for now we can probably get the state of DND on the device, currently we do not have a way for HA to send data to the app so we can perform actions.  I can see it being useful to control some things like to remove the device from DND if the alarm was tripped.  

We do also have notification channels which can allow you to override DND for certain notifications in case you are not aware :)

PR #813, opened 2020-08-25
Add do not disturb sensor

dshokouhi:
This PR adds a Do Not Disturb sensor that will update its state immediately upon a state change on Android 6.0+ devices. The initial state of the sensor will be `unavailable` for some devices that have never set their device into DND mode (I noticed my fire tablet on Android 5.0 experienced this behavior at first). When google first launched DND mode it was called interruptions and over time they had slowly made changes to its behavior.  This PR attempts to get the state from all supported Android versions, even the ones that don't support all the states like on Android 10.

State value are: `off`, `priority_only`, `total_silence`, `alarms_only` or `unknown` in addition to `unavailable` as mentioned up above.

On my pixel 4 xl the state is either `off` or `priority_only` I could not get it to be any of the other states.  On my Fire Tablet which is Android 5 I can see `off` `priority_only` and `total_silence` I suspect the other states are in different versions of Android.  We are just reading the value anyways and interpretting it from there.

https://stackoverflow.com/questions/31387137/android-detect-do-not-disturb-status

Fixes: #807 with the exception that we don't control the device yet

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/DNDSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/DNDSensorManager.kt
@@ -1,0 +1,62 @@
+package io.homeassistant.companion.android.sensors
+
+import android.content.Context
+import android.provider.Settings.Global
+import android.util.Log
+import io.homeassistant.companion.android.domain.integration.SensorRegistration
+
+class DNDSensorManager : SensorManager {
+    companion object {
+        private const val TAG = "DNDSensor"
+
+        private val dndSensor = SensorManager.BasicSensor(
+            "dnd_sensor",
+            "sensor",
+            "Do Not Disturb"
+        )
+        var dndState = "unavailable"
+    }
+
+    override val name: String
+        get() = "Do Not Disturb Sensor"
+
+    override val availableSensors: List<SensorManager.BasicSensor>
+        get() = listOf(dndSensor)
+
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/DNDSensorManager.kt
```diff
@@ -0,0 +1,59 @@
...
+class DNDSensorManager : SensorManager {
+    companion object {
+        private const val TAG = "DNDSensor"
+
+        private val dndSensor = SensorManager.BasicSensor(
+            "dnd_sensor",
+            "sensor",
+            "Do Not Disturb Sensor"
+        )
+        var dndState = "unavailable"
```
JBassett, line 16: We shouldn't store this state here.

app/src/main/java/io/homeassistant/companion/android/sensors/DNDSensorManager.kt
```diff
@@ -0,0 +1,59 @@
...
+                3 -> "alarms_only"
+                else -> "unknown"
+            }
+        } catch (e: Exception) {
+            Log.e(TAG, "Error getting the devices DND mode", e)
+        }
+
+        val icon = "mdi:do-not-disturb"
+
+        onSensorUpdated(context,
```
JBassett, line 52: Rather than always updating maybe move inside the try catch so we only update the value when we don't have an exception?

PR #816, opened 2020-08-26
Hopefully fix the background location updates randomly stopping.

JBassett:


dshokouhi:
Nice so location is now converted to be treated more like a sensor so it can respect its own schedule along with everything else if I am understanding this change correctly?

JBassett:
> Nice so location is now converted to be treated more like a sensor so it can respect its own schedule along with everything else if I am understanding this change correctly?

That's the plan.

PR #828, opened 2020-08-27
CI/CD Changes

JBassett:


dshokouhi:
So these now run in parallel? Looks like it cut down the time by almost 50%!! My last PR took 9 min total, longest one here is 5 minutes!  Great job!!🎉

Issue #830, opened 2020-08-27
Fatal crash for geocoded sensor


dshokouhi:
**Home Assistant Android version:**

2.2.1-29

**Android version:**

Android 7

**Phone model:**

Fire HD8 with google services

**Home Assistant version:**

0.114

**Last working Home Assistant release (if known):**

N/A

**Description of problem:**

After updating to the latest play store beta I am seeing teh following crash

**Traceback (if applicable):**

```
2020-08-27 06:06:47.297 1566-25967/? E/AMZ-MAPS-ERROR-MSP_104: Unknown exception in getFromLocation for AppPackage[io.homeassistant.companion.android]. Reason: No Key data found for package: io.homeassistant.companion.android mapControlVersion: 1.0 while Reverse Geocoding.
2020-08-27 06:06:47.391 7575-7575/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: io.homeassistant.companion.android, PID: 7575
    java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:801)
     Caused by: java.lang.reflect.InvocationTargetException
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:911)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:801) 
     Caused by: java.io.IOException: Geocoding failed, please look into the logs for more details!
        at android.location.Geocoder.getFromLocation(Geocoder.java:136)
        at io.homeassistant.companion.android.sensors.GeocodeSensorManager$updateGeocodedLocation$1.onSuccess(GeocodeSensorManager.kt:57)
        at io.homeassistant.companion.android.sensors.GeocodeSensorManager$updateGeocodedLocation$1.onSuccess(GeocodeSensorManager.kt:11)
        at com.google.android.gms.tasks.zzn.run(Unknown Source)
        at android.os.Handler.handleCallback(Handler.java:751)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6188)
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:911) 
[...]

PR #832, opened 2020-08-27
Fix geocoded crash, report expected minimum accuracy for location

dshokouhi:
Fixes: #830 

Also fixes a edge case where sometimes we were sending an invalid `gps_accuracy` attribute which was less than 0.

Example error from HA: `Received invalid webhook payload: value must be at least 0 for dictionary value @ data['gps_accuracy']. Got -2147478` and `Received invalid webhook payload: value must be at least 0 for dictionary value @ data['gps_accuracy']. Got -2147471`

During these errors all updates do not come through so when we detect that accuracy >= 0 we will send the real value, otherwise we will send `0` which is the bare minimum that HA will accept

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
@@ -59,7 +59,7 @@ class GeocodeSensorManager : SensorManager {
                         .getFromLocation(location.latitude, location.longitude, 1)
                         .firstOrNull()
             } catch (e: Exception) {
-                Log.e(TAG, "Failed to get geocoded location")
+                Log.e(TAG, "Failed to get geocoded location", e)
             }
             val attributes = address?.let {
                 mapOf(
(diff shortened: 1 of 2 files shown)


app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
```diff
@@ -46,17 +47,20 @@ class GeocodeSensorManager : SensorManager {
...
+                    Log.e(TAG, "Somehow location is null even though it was successful")
+                    return@addOnSuccessListener
+                }
 
+                if (location.accuracy <= LocationSensorManager.MINIMUM_ACCURACY)
+                    address = Geocoder(context)
+                        .getFromLocation(location.latitude, location.longitude, 1)
+                        .firstOrNull()
+            } catch (e: Exception) {
+                Log.e(TAG, "Failed to get geocoded location")
```
JBassett, line 62:
```suggestion
                Log.e(TAG, "Failed to get geocoded location", e)
```

dshokouhi: Ah thank you :)

PR #842, opened 2020-08-29
Bump some app dependencies

dshokouhi:
Noticed some of our dependencies had updates recently so lets bump them.  For the mentioned breaking changes I did a global search and I did not see us using those APIs.  I tested this in the emulator and on my Pixel 4 XL and did not see any real issues.  Most of these are bug fixes which are good things :)

Note: Kotlin and Coroutines have updates but were not bumped since I am not sure if they should be done yet.

* Dagger release notes https://github.com/google/dagger/releases
* Webkit is no longer in RC https://developer.android.com/jetpack/androidx/releases/webkit?hl=en
* work manager update https://developer.android.com/jetpack/androidx/releases/work?hl=en

### DIFF
--- a/buildSrc/src/main/kotlin/Config.kt
+++ b/buildSrc/src/main/kotlin/Config.kt
@@ -35,7 +35,7 @@ object Config {
         }
 
         object Google {
-            private const val daggerVersion = "2.27"
+            private const val daggerVersion = "2.28.3"
             const val dagger = "com.google.dagger:dagger:${daggerVersion}"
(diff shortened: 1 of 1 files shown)


dshokouhi:
going to check out the lint change locally to make sure it doesn't break anything for widgets

dshokouhi:
interesting appcombat mentions : `Tinting image views: flags not using app:tint that will lead to incorrect appearance on older platform versions` however it breaks on android 10 :)  I will remove this bump for now.

dshokouhi:
Hmm what other dependency can be causing this error

PR #845, opened 2020-08-30
Add link to companion docs under app configuration

dshokouhi:
Thought it would be nice to link to the companion docs so more users are aware of it.  I ended up moving down the app version info since it looks better with the larger icon separated.  Tapping on it will open a new browser tab in whatever is the default on the device.  The link can also be copied if need be on long press.

image


app/src/main/res/xml/preferences.xml
```diff
@@ -74,6 +74,17 @@
...
+    <PreferenceCategory
+        android:title="@string/need_help">
+        <Preference
+            android:key="docs"
+            android:icon="@drawable/ic_question"
+            android:title="@string/documentation"
+            android:summary="https://companion.home-assistant.io">
+            <intent android:action="android.intent.action.VIEW"
+                android:data="https://companion.home-assistant.io" />
+        </Preference>
```
JBassett, line 86: You can add `app:enableCopying="true"` here rather than setting it in the fragment.

app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -129,6 +129,10 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
             return@setOnPreferenceClickListener true
         }
 
+        findPreference<Preference>("docs")?.let {
+            it.isCopyingEnabled = true
+        }
+
```
JBassett, line 135: If you set this in the `preferneces.xml` you don't need to set it here.

dshokouhi: Thanks for that!

PR #847, opened 2020-08-30
Add description for each sensor on details screen, only show device class and icon if they are set

dshokouhi:
Time to add a description for each sensor so users know what they are getting if they decide to enable/disable it.  Not a lot of details just enough to tell them what to expect.  Made sure to warn the users as well for certain sensors that may not behave as expected (i.e. Next Alarm)  Also only show icon and device class if the sensor has it set.

We could consider linking directly to the docs if users find that to be helpful but I think one link to the docs should suffice for now.


app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
```diff
@@ -72,6 +72,32 @@ class SensorDetailFragment(
...
+                "phone_state" -> resources.getString(R.string.phone_state)
+                "pressure_sensor" -> resources.getString(R.string.pressure_sensor)
+                "proximity_sensor" -> resources.getString(R.string.proximity_sensor)
+                "sim_1" -> resources.getString(R.string.sim_1)
+                "sim_2" -> resources.getString(R.string.sim_2)
+                "steps_sensor" -> resources.getString(R.string.steps_sensor)
+                "storage_sensor" -> resources.getString(R.string.storage_sensor)
+                "wifi_connection" -> resources.getString(R.string.wifi_connection)
+                "zone_background" -> resources.getString(R.string.pref_location_zone_summary)
+                else -> resources.getString(R.string.no_description)
```
JBassett, line 97: Might want to prepend something to these keys so we know what they are, maybe `sensor_description_*`?

JBassett: Also it almost feels like this should be part of the `BasicSensor`.  That way people know to add a description when they add a sensor rather than having to come here too.  You could pass the basic sensor instead of just the `sensorId` in the constructor.

dshokouhi: Yea I honestly wasn't sure what the best approach would be here. After I got it working I figured send it over for review and see what can be improved upon.  I do agree having it as part of `BasicSensor` makes sense I'll have to see how to do that 😃

JBassett: I would just add the `R.string.sensor_description`, which should be an int, in the basic sensor.  Then in the preference just do `getString(sensor.description)`

app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
```diff
@@ -16,16 +16,18 @@ import io.homeassistant.companion.android.database.sensor.SensorDao
...
+    private val basicSensor: SensorManager.BasicSensor,
     private val sensorId: String
 ) :
     PreferenceFragmentCompat() {
 
     companion object {
         fun newInstance(
             sensorManager: SensorManager,
+            basicSensor: SensorManager.BasicSensor,
             sensorId: String
```
JBassett, line 28: We should remove this as a parameter and change any uses to `basicSensor.id` instead

dshokouhi: Ah yup no longer needed now :)

Issue #848, opened 2020-08-31
App closed, phone sensors are not updating


freemann:
**Home Assistant Android version:**
2.3.0-306-full

**Android version:**
10.0.7

**Phone model:**
OnePlus 7pro

**Home Assistant version:**
0.114.3

**Last working Home Assistant release (if known):**
Unknown

**Description of problem:**
When I close the HA companion app on my phone, non of the app related sensors are updating in HA.

**Traceback (if applicable):**

```

```

**Screenshot of problem:**

**Additional information:**

dshokouhi:
This is expected you need to keep the app running in the background so the background services continue to run.  Closing the app also essentially stops these services from running.  Try these steps to make sure the app has proper permissions: https://companion.home-assistant.io/docs/troubleshooting/faqs/#location-is-not-updating-in-android-app

anyuta1166:
As I already said in the comments to the location tracking issue, this is Android 9+ specific behavior.

If the phone is left intact and enters the doze mode, sensors update may be delayed by up to 2 hours **even if the app is running in the background**.

You still don't want to create a service with notification? ^_^

freemann:
Totally agree with @anyuta1166 
It's a bit against the automation principal to work with old data. I was working with "Automate" to send battery and location data. 2 weeks ago I moved from Automate to companion data, but this is just not working as it should.

Thinking to go back to Automate, that just worked like a charm.

Hope you could consider to change this behavior :)

freemann:
@dshokouhi I love the foreground service! 
Please "push" it through!! ❤️

freemann:
Should this also fix #671 or make the updates more frequent evenly spread of time?

dshokouhi:
@freemann the foreground service just needs to be tested a lil longer to make sure things are still good.  It won't fix #671 because the periodic work request that we use has a minimum interval of 15 minutes to remain battery friendly so we won't get each and every percentage change. This change will make sure that when the work request starts the updates will be pulled with a foreground service to ensure they actually get pulled.

dshokouhi:
So somethings to consider when the app is closed and how work manager works: https://stackoverflow.com/questions/50682061/android-is-workmanager-running-when-app-is-closed/52605503#52605503

Notice how it says some devices closing the app (or removing from recent apps) is also the same as Force Stop so in those cases foreground service won't help.

@freemann look into https://dontkillmyapp.com/ for your OnePlus device as those devices are more susceptible to killing the app which will stop the worker.

[...]

Issue #551, opened 2020-04-13
After a few days sensors stop reporting updates until app is force closed and reopened


dshokouhi:
**Home Assistant Android version:**
1.8.0-159
**Android version:**
10
**Phone model:**
Pixel 4 XL
**Home Assistant version:**
0.108.2
**Last working Home Assistant release (if known):**
N/A
**Description of problem:**

After a few days of uptime I noticed that my sensor stopped updating randomly.  I tried to open the app but the sensor did not update.  I then force closed the app and reopened it and the sensors updated as expected.  I charged my phone over night and we can see that the battery state and battery level sensor stopped updating.  We don't even see a gradual incline in battery life over night as I usually see, it randomly jumped to the top.  Surprisingly the `charger_type` attribute still updated as expected but battery state nor battery level itself did not update as expected.  Maybe the service is getting stuck or something?

**Traceback (if applicable):**

```

```

**Screenshot of problem:**
![image](https://user-images.githubusercontent.com/1634145/79153470-c30f4600-7d82-11ea-804b-5dafbfe0169a.png)

![image](https://user-images.githubusercontent.com/1634145/79153493-cc001780-7d82-11ea-9c92-27b4638e5e51.png)

![image](https://user-images.githubusercontent.com/1634145/79153628-04075a80-7d83-11ea-8108-3dc5d47672c0.png)

**Additional information:**

SupahNoob:
I've noticed this on multiple devices, and the only thing that appeared to fix the error was a force stop and reopen.

dshokouhi:
> I've noticed this on multiple devices, and the only thing that appeared to fix the error was a force stop and reopen.

I noticed that when I do the force stop sometimes itll last a while and sometimes itll be less than a day before it stops reporting.  I found restarting the phone at that point increases the timeframe of having to force stop the app.  My wifes phone dies on a near daily basis due to battery life so she never sees this lol.

SupahNoob:
I just reset one of my devices (Pixel 2XL) and exhausted literally every option (device admin, background data on, clean install) to get these stats to appear. The last one I tried was force stopping the app and reopening, and this was the only solution.

It's hard to tell how long it'll last for, but hopefully this helps debug the issue.

dshokouhi:
Yea next time this happens if I remember I will try to check the ADB logs, that may hold the key here.

Going to post the commands here too so maybe our logs can show the answer.  It almost sounds like the service crashes or gets in a weird state requiring the app to be killed and restarted.

[...]

PR #854, opened 2020-08-31
Use a foreground service for the sensor worker to make updates more reliable

dshokouhi:
WorkManager offers the ability to use a foreground service to make sure updates become more reliable.  I have been testing this out all day and I can see better updates compared to the play store beta and I do not see any additional battery drain!  This is a temporary notification that will appear in the status bar only during a sensor update. The priority is set to low as that is expected behavior.  This uses its own notification channel so users can make changes if they like.  This is very similar to recent changes done to the Google Phone app where we temporarily see the voice-mail syncing notification. 

The notification:
image

Production app:
image

With this change:
image

Should fix: #848 and hopefully #551 🤞

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorWorker.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorWorker.kt
@@ -1,14 +1,21 @@
 package io.homeassistant.companion.android.sensors
 
+import android.app.NotificationChannel
+import android.app.NotificationManager
 import android.content.Context
+import android.content.Context.NOTIFICATION_SERVICE
+import android.os.Build
 import android.util.Log
+import androidx.core.app.NotificationCompat
 import androidx.work.Constraints
 import androidx.work.CoroutineWorker
 import androidx.work.ExistingPeriodicWorkPolicy
+import androidx.work.ForegroundInfo
 import androidx.work.NetworkType
 import androidx.work.PeriodicWorkRequestBuilder
 import androidx.work.WorkManager
 import androidx.work.WorkerParameters
+import io.homeassistant.companion.android.R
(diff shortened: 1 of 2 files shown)


dshokouhi:
Some nice logs to go with everything

```
2020-09-01 15:02:27.189 12768-13546/io.homeassistant.companion.android.debug I/WM-Processor: Moving WorkSpec (6c1d4fc9-8475-4297-a6dc-426154554946) to the foreground
2020-09-01 15:02:27.200 12768-12768/io.homeassistant.companion.android.debug I/WM-SystemFgDispatcher: Started foreground service Intent { act=ACTION_START_FOREGROUND cmp=io.homeassistant.companion.android.debug/androidx.work.impl.foreground.SystemForegroundService (has extras) }
2020-09-01 15:02:27.510 12768-12854/io.homeassistant.companion.android.debug I/WM-WorkerWrapper: Worker result SUCCESS for Work [ id=6c1d4fc9-8475-4297-a6dc-426154554946, tags={ io.homeassistant.companion.android.sensors.SensorWorker } ]
2020-09-01 15:02:27.513 12768-12768/io.homeassistant.companion.android.debug I/WM-SystemFgDispatcher: Stopping foreground service
```

rs443:
Some quick questions!

1. Will the foreground service also update location? (If not, maybe it would be a good idea as the location updates sometimes is unreliable and stops updating)

2. If all non location sensors are disabled as in #879 and if the answer on question 1 is no, will the foreground service till be run?

dshokouhi:
@rs443 Location sensors are also updated with all the other sensors: https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt#L32

PR #858, opened 2020-08-31
Switch to string for translating sensor names

dshokouhi:
After adding the sensor description I noticed we were not using strings for the names that appear for the sensor groups.  I am not sure about translating the actual name that appears for the sensor itself I believe that gets sent over to HA?


dshokouhi:
Going to add the actual sensor name so marked as draft for now

Issue #867, opened 2020-09-02
New sensors should be opt-in


tribut:
**Is your feature request related to a problem? Please describe.**
When I opened my mobile device in HA, I was surprised to find new data there (step data, for example). I had disabled all sensors in the app at some point. Apparently, new sensors were added and automatically enabled.

**Describe the solution you'd like**
Make newly added sensors opt-in or add a flow that asks for permission to send data when the app is installed or new sensors are added.

**Describe alternatives you've considered, if any**

**Additional context**

Issue #856, opened 2020-08-31
Do not hold proximity sensor lock


pkern:
**Is your feature request related to a problem? Please describe.**
With version 2.3.0-306-full the proximity sensor is kept open indefinitely until the app is terminated.

**Describe the solution you'd like**
Proximity sensory usage should be opt in, not opt out. Otherwise a distracting dot will be shown on the display while the display is on.

**Describe alternatives you've considered, if any**
n/a

**Additional context**
See https://github.com/home-assistant/android/pull/801#issuecomment-684019752

dshokouhi:
What device are you experiencing this issue with? We unregister from the sensor the moment we get a `onSensorChanged` event from Android regardless of the reading.

https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/sensors/ProximitySensorManager.kt#L84

pkern:
Unfortunately I cannot answer you this question but I will forward the link. Thank you. (The fast response is really appreciated and I'm sorry that I can't go into more detail and am also unable to just step through the program to see where that hangs. I do know from the log that the sensor is held active for hours specifically by Home Assistant.)

skynetua:
I'm not sure, but probably we need to register/unregister receiver for application context and not like:
`mySensorManager.registerListener(this, proximitySensors, SENSOR_DELAY_NORMAL)`

For now we can re-create sensor manager object before receiver unregistered.
Correct me if i'm wrong.

dshokouhi:
@skynetua we actually are creating it based off the latest context we have: https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/sensors/ProximitySensorManager.kt#L47  

The issue here is that the `onSensorChanged` event is never being triggered for this device so more than likely this user never see's any data from this sensor in Home Assistant.  Maybe we can help alleviate this by using a boolean value to keep track of when we create the listener so we don't create multiple.  That however won't fix this issue if we never get the expected event.

pkern:
In this case there are many proximity sensor events.

[...]

Issue #582, opened 2020-05-07
Make Geolocation Opt-In


tinloaf:
**Home Assistant Android version:**
1.8.0-144

**Android version:**
10 (Recent LineageOS)

**Phone model:**
Xiaomi Mi 8

**Home Assistant version:**
0.109.3

**Last working Home Assistant release (if known):**
n/a

**Description of problem:**
The fact that the app continually (reverse-) geocodes the current GPS location means that my location is permanently being sent to Google Maps. This is something that (in my opinion) an app should not do without at least informing me about it, in fact I think this should be an opt-in option. At the very least, there should be a way of disabling it (opt-out).

I realize I can just revoke the location permission (which I have just done), but now my automations related to zones don't work anymore. 

I could imagine that a considerable part of HA's user base is quite privacy-conscious and would be as surprised as I was to learn that their location is permanently being sent to Google.

**Traceback (if applicable):**
n/a

**Screenshot of problem:**
n/a

**Additional information:**
n/a

andriej:
I'm also keen more on opt-in rather than default opt-out policy, even tho all localisation messages are going thru google servers anyway ;-)

tinloaf:
No, they are not. I've just had that very same discussion twice on Discord. You can turn off your "location history" and Google's Wifi-assisted GPS localization. That way, no GPS locations should be disclosed to Google (or anybody else, for that matter). It's kind of weird that even Google offers the option to not send GPS locations to themselves, but the HA app has a mandatory GPS-to-Google-sending thing.

Troon:
I'll vote for this one too. The `geocoded_location` isn't particularly accurate or useful in my experience, and I'd appreciate being able to turn it off.

nagyrobi:
+1 for this

skylarmt:
> It's kind of weird that even Google offers the option to not send GPS locations to Google, but the HA app has a mandatory GPS-to-Google-sending thing.

Not actually weird or surprising, considering HA is refusing to support phones without Google services, even when offered payment to do so. See also issues #42, #124, and #676.

Issue #518, opened 2020-03-16
Precision selection combo or Smart Precision feature


McGiverGim:
**Is your feature request related to a problem? Please describe.**
The precision value taken for the tracking is very strict, so sometimes you abandon a zone and any automation related with this is not executed until a lot of time later because the new positions are discarded.
For example, I have a zone in the train station of Madrid. When my wife takes the train (and abandons the station) I have an automation. In my case it is not triggered until one hour later, when my wife gets to the destination station, at 200 Km of the origin station.
Maybe this is an edge case, with a High Speed train that can get to more than 300 Km/h, tunnels, the metal of the train, etc. but Google Maps is able to track her with fair good precision for at least be able to throw an automation when abandoning the zones.

**Describe the solution you'd like**
I think that a simply selection box with precision `ALL/LOW/MED/HIGH` to let the user select what he can use, with `HIGH` by default and maybe a warning if changed will be enough. 

If for some reason the project don't want to give this option to the users, maybe an implementation of a Smart Precision can be done. I don't know how much data we have in the app, but if we have enough, we can do something like that:

1. If the accuracy of the new location is smaller than the minimum accuracy defined, send the location, like now. This is the actual behaviour and does not change nothing.
2. If the accuracy of the new location is bigger than the minimum accuracy, then:
2.1. If the old location was inside a zone, and the new location is detected like away, then send the new location if the distance > (np_accuracy + z_radius)
![image](https://user-images.githubusercontent.com/2673520/76734598-90facc00-6763-11ea-85ff-be061ae06d54.png)

2.2. If the new location is inside a zone, and the old location was away or in a different zone, then send the new location if (distance + np_accuracy) < z_radius:
![image](https://user-images.githubusercontent.com/2673520/76734656-b38ce500-6763-11ea-9a08-9fd296f5cc34.png)

In this way, we only send the location when the new position is completely inside the new zone/away, including the accuracy. This give us more margin in the accuracy, especially when going from a zone to away, to detect correctly the change of state and start automations that now are being blocked by an accuracy too exigent. This will only be used when changing zone, while staying in the zone only the point 1 is used, like now, so the GPS coordinates will be the same than now the major part of the time. And can be an option "Smart zone detection" or similar deactivated by default.

**Describe alternatives you've considered, if any**
Only the two commented: a selection box in the config or a Smart Precision detection.

**Additional context**
None.

balloob:
I don't like the idea of a new option, but I do like the idea of the advanced zone check.

[...]

PR #875, opened 2020-09-02
Allow sensors to be disabled by default until user enables them

dshokouhi:
Fixes: #867 & #856 

Now that we have a lot of sensors lets just allow sensible defaults so we don't overload new users. This really impacts onboarding more than current users.  For now battery sensors will be defaulted to true, location and network sensors will default to true only if given permission during onboarding.  Geocoded location will be false by default to allow it to be a opt-in sensor instead of opt-out (I know this goes against what #853 just did but we did have a specific request for Geocoded to be opt-in instead of opt-out #582 )

Are we ok with the defaults for new users?

One issue I saw which is purely cosmetic is that after enabling one sensor another sensor will briefly show the toggle enabled even though it is not enabled. Going back to the same screen will show the correct state so maybe a caching issue?

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
@@ -155,7 +155,7 @@ class SensorDetailFragment(
             sensorEntity.enabled = isEnabled
             sensorDao.update(sensorEntity)
         } else {
-            sensorEntity = Sensor(basicSensor.id, false, false, "")
+            sensorEntity = Sensor(basicSensor.id, isEnabled, false, "")
             sensorDao.add(sensorEntity)
         }
         refreshSensorData()
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
@@ -12,6 +12,7 @@ import androidx.preference.contains
 import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
 import io.homeassistant.companion.android.database.AppDatabase
+import io.homeassistant.companion.android.database.sensor.Sensor
 import io.homeassistant.companion.android.database.sensor.SensorDao
 
 class SensorDetailFragment(
@@ -51,9 +52,10 @@ class SensorDetailFragment(
         findPreference<SwitchPreference>("enabled")?.let {
             val dao = sensorDao.get(basicSensor.id)
             val perm = sensorManager.checkPermission(requireContext())
-            if (dao == null) {
+            if (dao == null && sensorManager.enabledByDefault) {
                 it.isChecked = perm
-            } else {
+            }
+            if (dao != null) {
                 it.isChecked = dao.enabled && perm
             }
             updateSensorEntity(it.isChecked)
@@ -148,10 +150,13 @@ class SensorDetailFragment(
     private fun updateSensorEntity(
         isEnabled: Boolean
     ) {
-        val sensorEntity = sensorDao.get(basicSensor.id)
+        var sensorEntity = sensorDao.get(basicSensor.id)
         if (sensorEntity != null) {
             sensorEntity.enabled = isEnabled
             sensorDao.update(sensorEntity)
+        } else {
+            sensorEntity = Sensor(basicSensor.id, false, false, "")
+            sensorDao.add(sensorEntity)
         }
         refreshSensorData()
     }
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorManager.kt
@@ -13,6 +13,7 @@ interface SensorManager {
 
     val name: Int
     val availableSensors: List<BasicSensor>
+    val enabledByDefault: Boolean
 
     data class BasicSensor(
         val id: String,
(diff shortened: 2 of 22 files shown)


skynetua:
I doubt that we should disable the whole manager rather the particular sensor if we are going to split non static attributes to separate sensors.

dshokouhi:
Well I know that most of these should be disabled by default. During onboarding for example if you grant phone permissions then only the phone sensor is enabled along with battery sensors.  The SIM slots are not enabled.  By default I think we should disable all sensors and during onboarding bring the user to the sensor screen so they can enable whichever they choose. Since we will be splitting things out we will probably need to restructure the managers better to fit this change along with their own changes during the split. We are going to have a lot of new sensors so best to leave it all up to the user.  This is really all about the new user experience and being hit with a lot of new entities any time we add stuff or they just signed up.  Existing users won't see a change until new sensors are added, then they just need to enable it.

anyuta1166:
> Geocoded location will be false by default to allow it to be a opt-in sensor instead of opt-out (I know this goes against what #853 just did but we did have a specific request for Geocoded to be opt-in instead of opt-out #582 )

Very strange decision. IMO, this is the basic set of sensors which should be enabled:

- battery (always)
- network, location & geocoded (if the user granted a premission).

The people are very strange. They use Google services, that means Google already have tons of information about them, and they worry about Google knowing their location.
BTW, cell-tower/wi-fi based location comes from Google too - device sends info about nearby cell-towers and Wi-Fi APs to Google and receives approximate location from Google....
So must we disable location sensor too? ;-0
If you are afraid of Google, just use minimal version instead...

anyuta1166:
> @anyuta1166 The problem is not always that the user don't want Google to know their location... the problem is that the sensor appears in the Home Assistant UI and all the people that has access to it (the family) knows your location. This is ok for me, but maybe others don't want to publish this kind of information to family and friends.
 @McGiverGim  1) You can opt-out by disabling the sensor
2) Location (not geocoded) it still sent to HA. What is the difference between the dot on the map and the textual represention? All the people can still see you as the dot on the map.

anyuta1166:
Geocoded location is just an addition to regular location - it translates the dot on the map to the textual representation (address).

So I thing they should be both either enabled or disabled by default.

McGiverGim:
Is not exactly the same... they use different location mechanisms. For this reason, sometimes Home Assistant detects me at some zone but the Geocoded location detects me at 200Km of distance, that is the right location. I opened some time ago a feature request to try to fix this but without luck.

The regular location is needed to enable/disable automations related with zones, the Geocoded location is only informative, so is not needed. I suppose this is the reason to no enable it by default.

I use both, to me is not a problem, but maybe for others yes...

anyuta1166:
Geocoded location is updated every 15 minutes (or when anupdate is triggered by another sensor). Regular location updates every couple of minutes (or even faster if you use GPS). That makes the difference. Geocoded location may show your location 15 minutes ago while regular location is up-to-date.

McGiverGim:
That not was my problem, it is the "precision" of the standard location. If it is "big" it is discarded. When on a high speed train, for example, it seems is always big, so it discards all of them, including locations at more than 100Km of the previous location... I opened a PR with the description here: https://github.com/home-assistant/android/issues/518

But this is out of the target of this PR. I only wanted to expose some reason why some user may prefer to have it deactivated. I use both, so to me is not a problem.

dshokouhi:
Lets please keep the discussion on accuracy elsewhere so we don't get off topic.  Lets also not assume every user wants to send their data over to a different API for processing.  Making sensors opt in instead of opt out makes for a friendlier first time experience.  If you really want geocoded after onboarding just enable it.  Before that the entity is not created in HA and no data is collected, which is what some users are after.  In fact with this code the sensor won't even exist in the on device database UNTIL the user enables it.

anyuta1166:
Wouldn't this conflict with #853 in case of Geocoded sensor? What happens if we try to enable sensor that doesn't exist.

dshokouhi:
@anyuta1166 I mentioned that I know it goes against #853 in the PR, that was why I called it out 😄 .  

If a user enables a sensor that does not exist then and only then do we create the sensor and send the data over to HA: https://github.com/home-assistant/android/pull/875/files#diff-4f2b16c8e5d824da9b572e6255d81f9fR159

app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
```diff
@@ -148,10 +150,13 @@ class SensorDetailFragment(
     private fun updateSensorEntity(
         isEnabled: Boolean
     ) {
-        val sensorEntity = sensorDao.get(basicSensor.id)
+        var sensorEntity = sensorDao.get(basicSensor.id)
         if (sensorEntity != null) {
             sensorEntity.enabled = isEnabled
             sensorDao.update(sensorEntity)
+        } else {
+            sensorEntity = Sensor(basicSensor.id, false, false, "")
```
JBassett, line 158: Shouldn't this take into account `isEnabled`?

app/src/main/java/io/homeassistant/companion/android/sensors/SensorManager.kt
```diff
@@ -36,19 +37,20 @@ interface SensorManager {
         var sensor = sensorDao.get(sensorId)
         val permission = checkPermission(context)
 
-        // If we haven't created the entity yet do so and default to enabled
-        if (sensor == null) {
+        // If we haven't created the entity yet do so and default to enabled if required
+        if (sensor == null && enabledByDefault) {
             sensor = Sensor(sensorId, permission, false, "")
```
JBassett, line 42:
```suggestion
        if (sensor == null) {
            sensor = Sensor(sensorId, permission && enabledByDefault, false, "")
```

JBassett: By making this change the bellow changes won't be needed.  It will ensure that sensor is never null going forward.

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
dshokouhi: When I was testing it out I tried that but it resulted in the entities being enabled this way they are always kept disabled and they also respect the onboarding permission screen as well.

dshokouhi: looks like it works as expected now :) thank you for that

app/src/main/java/io/homeassistant/companion/android/sensors/SensorManager.kt
dshokouhi: oh much better :) thank you

PR #877, opened 2020-09-03
Only show sensors that are available for the device

dshokouhi:
Now that sensors are disabled by default lets completely hide sensors that the device reports it does not have. One thing to note about this is that certain devices don't respect this API and still report they have a device when indeed they do not.  Our current null checks prevent that from being an issue anyways.

I was unable to add this check to the manager itself as we do not get context during this moment, I added this in the fragment to just completely hide it they are disabled anyways.  We can also use this for other sensors in the future that may exist in certain API levels too.  For now this change only impacts hardware sensors since not all devices have them.

Here we can see the emulator does not show the step sensor since it does not exist:
image

Here is a Fire HD8 that only has a light sensor, the other 3 are notably not available:
image

I do have one Fire HD7 that runs on lineage OS with google services, this device unfortunately reports it has the sensors and just returns a null value so we ignore those anyways.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
@@ -56,35 +56,33 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
 
         setPreferencesFromResource(R.xml.sensors, rootKey)
 
-        SensorReceiver.MANAGERS.sortedBy { it.name }.forEach { manager ->
+        SensorReceiver.MANAGERS.sortedBy { it.name }.filter { it.hasSensor(requireContext()) }.forEach { manager ->
             val prefCategory = PreferenceCategory(preferenceScreen.context)
             prefCategory.title = getString(manager.name)
 
-            if (manager.hasSensor(requireContext())) {
-                preferenceScreen.addPreference(prefCategory)
-                manager.availableSensors.sortedBy { it.name }.forEach { basicSensor ->
+            preferenceScreen.addPreference(prefCategory)
+            manager.availableSensors.sortedBy { it.name }.forEach { basicSensor ->
 
-                    val pref = Preference(preferenceScreen.context)
-                    pref.key = basicSensor.id
-                    pref.title = getString(basicSensor.name)
+                val pref = Preference(preferenceScreen.context)
+                pref.key = basicSensor.id
+                pref.title = getString(basicSensor.name)
 
-                    pref.setOnPreferenceClickListener {
-                        parentFragmentManager
-                            .beginTransaction()
-                            .replace(
-                                R.id.content,
-                                SensorDetailFragment.newInstance(
-                                    manager,
-                                    basicSensor
-                                )
+                pref.setOnPreferenceClickListener {
+                    parentFragmentManager
+                        .beginTransaction()
+                        .replace(
+                            R.id.content,
+                            SensorDetailFragment.newInstance(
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
```diff
@@ -59,29 +60,40 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
...
-            manager.availableSensors.sortedBy { it.name }.forEach { basicSensor ->
+            var hasSensor: Boolean
+            val packageManager: PackageManager? = context?.packageManager
+            hasSensor = when (manager.name) {
+                R.string.sensor_name_light -> packageManager!!.hasSystemFeature(PackageManager.FEATURE_SENSOR_LIGHT)
+                R.string.sensor_name_pressure -> packageManager!!.hasSystemFeature(PackageManager.FEATURE_SENSOR_BAROMETER)
+                R.string.sensor_name_proximity -> packageManager!!.hasSystemFeature(PackageManager.FEATURE_SENSOR_PROXIMITY)
+                R.string.sensor_name_steps -> packageManager!!.hasSystemFeature(PackageManager.FEATURE_SENSOR_STEP_COUNTER)
+                else -> true
+            }
```
JBassett, line 71: We should add a function to SensorManager (`hasSensor(context)`)that checks for this and defaults to true and override in the needed managers.

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
dshokouhi: ah so thats how we bring context in 😃 thank you!

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
```diff
@@ -59,29 +59,32 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
         SensorReceiver.MANAGERS.sortedBy { it.name }.forEach { manager ->
             val prefCategory = PreferenceCategory(preferenceScreen.context)
             prefCategory.title = getString(manager.name)
-            preferenceScreen.addPreference(prefCategory)
-            manager.availableSensors.sortedBy { it.name }.forEach { basicSensor ->
 
-                val pref = Preference(preferenceScreen.context)
-                pref.key = basicSensor.id
-                pref.title = getString(basicSensor.name)
+            if (manager.hasSensor(requireContext())) {
```
JBassett, line 63: If you really wanted to get fancy you could filter before the foreach and remove the whole if statement.

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
```diff
@@ -59,29 +59,32 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
         SensorReceiver.MANAGERS.sortedBy { it.name }.forEach { manager ->
```
JBassett, line 59:
```suggestion
        SensorReceiver.MANAGERS.sortedBy { it.name }.filter { it.hasSensor(requireContext()) }.forEach { manager ->
```

dshokouhi: Ah even cleaner, will add now :)

PR #885, opened 2020-09-05
Add power sensors: Interactive, Doze, Power Save

dshokouhi:
3 new binary sensors, all disabled by default:

* Interactive  - Reflects the interactive state of the device, most typically when the screen turns on and off.  I have also added a broadcast receiver so the state updates immediately. We should look into adding this for the widgets, having them update when the screen turns on would add to the experience 👍 

* Doze mode!! Only available on Android 6.0+, will update immediately upon a state change.  One attribute of the sensor is `ignoring_battery_optimizations` for our application. Tested the state change using: https://developer.android.com/training/monitoring-device-state/doze-standby#testing_doze

image

* Power Save mode - Updates immediately when the device enters and exits power saving mode.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -8,6 +8,7 @@ import android.content.IntentFilter
 import android.media.AudioManager
 import android.net.wifi.WifiManager
 import android.os.Build
+import android.os.PowerManager
 import android.telephony.TelephonyManager
 import io.homeassistant.companion.android.common.dagger.AppComponent
 import io.homeassistant.companion.android.common.dagger.Graph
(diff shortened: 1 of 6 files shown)


anyuta1166:
Is it possible to call registerReceiver only when a sensor is enabled and unregister when the sensors is disabled? We are going to trigger on too many events again.

dshokouhi:
Well these events in particular won't get triggered as often as next alarm can, they are triggered by the users action and only their action.  Currently in `HomeAssistantApplication.kt` we don't get application context so we can't check there but also I think this only gets called on application start up so it may not be the appropriate place to do this.  We also already had a request to update the widgets when the screen turns on, by updating all sensors when the screen comes on would add to the user experience of seeing things up to date in HA.  

If people really want things to be as up to date as possible then adding more receivers and ONLY sending updates when a state change has occurred would add to that experience.  There really is no drain when the system broadcasts these receivers.  I think the best approach would be stop the state updates if we have nothing new for any of them.

dshokouhi:
In my near 24 hour test it wasn't that bad for interactive updates honestly, may actually shed light to users on how often they use their device :)  Plus would offer some neat automations for tablets that are not wall mounted so you can tell when they are actively being used.

image

JBassett (approved):

Issue #859, opened 2020-08-31
states.sensor.XXX_geocoded_location.attributes.Location is now a string??


i00:
## The problem

The geocoded_location.location sensor used to have location expressed as an array of 2 numbers for example:
[-27.468556999999997, 153.0240753]
Thus calling: {{ states.sensor.XXX_geocoded_location.attributes.Location[0] }} would return -27.468556999999997
Now it is a string:
'[-27.468556999999997, 153.0240753]'
Thus calling: {{ states.sensor.XXX_geocoded_location.attributes.Location[0] }} returns '['

## Environment

- Home Assistant Core release with the issue: 0.114.4
- Last working Home Assistant Core release (if known):
- Operating environment (OS/Container/Supervised/Core): Container
- Integration causing this issue: Mobile App?
- Link to integration documentation on our website: https://www.home-assistant.io/integrations/mobile_app/ ?

## Problem-relevant `configuration.yaml`

```yaml
{{ states.sensor.XXX_geocoded_location.attributes.Location[0] }}
```

## Traceback/Error logs

```txt

```

dshokouhi:
@i00 thanks for confirming....was this working fine previously right? do you happen to know what app version it was working on and what version you have today?

skynetua:
It is changed after #809. We store all attribute values as string to db.
`@ColumnInfo(name = "value")
  val value: String`

dshokouhi:
thanks @skynetua I see the same issue on `boolean` attributes too, need to wrap my sensor templates in strings now.  Did not notice this until now :)

i00:
@skynetua ... so is this going to be fixed?..
Can't the sensor be aware of what the data is for each property (ie. .Location is a array of numbers) and cast the properties out to the correct data type?

skynetua:
@i00 it can, I'm not sure we need to fix it right now. We have got statement not to send non-static values as attributes but as separate sensors. So probably it will be changed soon.

dshokouhi:
I think for the meantime we can use some jinja magic to strip the string quotes right?

i00:
Thanks for the merge fixes my issue :D as i can now use the new lat / long attributes.

PR #889, opened 2020-09-06
Breaking Change: Allow attributes to send over their proper data format

dshokouhi:
Marking this as a breaking change since it changes the values that are being sent over for our attributes so users may need to adjust their templates.

Attributes that are an `Int`, `Float` or `Boolean` are now properly converted, all others default to `String` as that is the existing behavior.

Fixes: #859 however with the breaking change that we no longer send latitude and longitude in the same attribute, they are now split up to match the device tracker.  These will also probably later on be transferred to its own sensor as they are considered non-static however for now lets fix the issue before we actually break up the attributes from the sensor.


app/src/main/java/io/homeassistant/companion/android/sensors/SensorManager.kt
```diff
@@ -84,10 +83,26 @@ interface SensorManager {
...
+                if (item.key == entry.key)
+                    sensorDao.add(
+                        Attribute(
+                            basicSensor.id,
+                            entry.key,
+                            entry.value.toString(),
+                            valueType
+                        )
+                    )
+            }
```
JBassett, line 105:
```suggestion
            sensorDao.add(
                Attribute(
                    basicSensor.id,
                    item.key,
                    item.value.toString(),
                    valueType
                )
            )
```

app/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
```diff
@@ -156,5 +157,10 @@ abstract class AppDatabase : RoomDatabase() {
                 database.execSQL("CREATE TABLE IF NOT EXISTS `sensor_attributes` (`sensor_id` TEXT NOT NULL, `name` TEXT NOT NULL, `value` TEXT NOT NULL, PRIMARY KEY(`sensor_id`, `name`))")
             }
         }
+        private val MIGRATION_7_8 = object : Migration(7, 8) {
+            override fun migrate(database: SupportSQLiteDatabase) {
+                database.execSQL("ALTER TABLE `sensor_attributes` ADD `value_type` TEXT NOT NULL DEFAULT ''")
```
JBassett, line 162: Maybe default to string rather than empty?

app/src/main/java/io/homeassistant/companion/android/database/sensor/Attribute.kt
```diff
@@ -10,5 +10,7 @@ data class Attribute(
     @ColumnInfo(name = "name")
     val name: String,
     @ColumnInfo(name = "value")
-    val value: String
+    var value: String,
+    @ColumnInfo(name = "value_type")
+    var valueType: String = ""
```
JBassett, line 15: Don't default, we should always set.

app/src/main/java/io/homeassistant/companion/android/database/sensor/SensorWithAttributes.kt
```diff
@@ -14,7 +14,17 @@ data class SensorWithAttributes(
     val attributes: List<Attribute>
 ) {
     fun toSensorRegistration(): SensorRegistration<Any> {
-        val attributes = attributes.map { it.name to it.value }.toMap()
+        val attributes = attributes.map {
+            val attributeValue = when (it.valueType) {
+                "" -> ""
```
JBassett, line 19: If we don't default we should never get this, we have this for state because there are times that there is no valid state.

JBassett (changes requested):

JBassett (approved):

PR #891, opened 2020-09-07
Breaking Change: Only send updates when a state change has been detected

dshokouhi:
This PR is another breaking change....sorry about this but we need to cut down on the amount of updates we are sending when there is no new data.  I will need to submit a follow up PR splitting up the non-static attributes so until that point some of the attributes will be slower to update.  If you feel I should add them here to avoid that being a breaking change I can do so but this already feels like a big change in itself so I opted to keep them separate.  So far Audio Sensor (all attributes), Battery State (all attributes but they all update anyways due to plug/unplug events with the exception of `battery_health`), Bluetooth (only for `is_bt_on` the rest update with the state), Storage Sensor (external attributes should become their own sensor however its tough to create on the fly if user removes the SD card so it may need to remain here) and Wifi Connection (all attributes) will be the ones that need to be split up immediately as we have broadcast receivers to update some of those attributes more quickly.

We will be keeping track of state changes in the database and only send updates when we have them.  If none of the sensors have a state change to report then nothing will get sent.  The default for this value is set to `true` the first time we register and update a sensor.

The only sensors that do not respect this change are the hardware sensors: Light, Pressure, Proximity and Step Counter.  These sensors need to be reconsidered on how they send updates since when we request for updates we actually don't see them when they are sent.  The `onSensorChanged` event can be triggered by any of these sensors at any point in time so the state change detection does not work there.  This is something I did not realize until I made this change and hence why I left the debug log so we can figure out what works best for those sensors. I think maybe triggering another update inside the event may fix that? But thats for another PR 😃

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
@@ -30,7 +30,7 @@ import io.homeassistant.companion.android.database.widget.TemplateWidgetEntity
         StaticWidgetEntity::class,
         TemplateWidgetEntity::class
     ],
-    version = 8
+    version = 9
 )
 abstract class AppDatabase : RoomDatabase() {
     abstract fun authenticationDao(): AuthenticationDao
@@ -63,7 +63,8 @@ abstract class AppDatabase : RoomDatabase() {
                     MIGRATION_4_5,
                     MIGRATION_5_6,
                     MIGRATION_6_7,
-                    MIGRATION_7_8
+                    MIGRATION_7_8,
+                    MIGRATION_8_9
                 )
                 .build()
         }
@@ -160,6 +161,11 @@ abstract class AppDatabase : RoomDatabase() {
         private val MIGRATION_7_8 = object : Migration(7, 8) {
             override fun migrate(database: SupportSQLiteDatabase) {
                 database.execSQL("ALTER TABLE `sensor_attributes` ADD `value_type` TEXT NOT NULL DEFAULT 'string'")
+            }
+        }
+        private val MIGRATION_8_9 = object : Migration(8, 9) {
+            override fun migrate(database: SupportSQLiteDatabase) {
+                database.execSQL("ALTER TABLE `sensors` ADD `state_changed` INTEGER NOT NULL DEFAULT ''")
             }
         }
     }
--- a/app/src/main/java/io/homeassistant/companion/android/database/sensor/Sensor.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/database/sensor/Sensor.kt
@@ -13,6 +13,8 @@ data class Sensor(
     var enabled: Boolean,
     @ColumnInfo(name = "registered")
     var registered: Boolean,
+    @ColumnInfo(name = "state_changed")
+    var stateChanged: Boolean,
     @ColumnInfo(name = "state")
     var state: String,
     @ColumnInfo(name = "state_type")
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationFragment.kt
@@ -251,6 +251,7 @@ class MobileAppIntegrationFragment : Fragment(), MobileAppIntegrationView {
                 uniqueId,
                 isChecked,
                 false,
+                true,
                 ""
             )
             sensorDao.add(sensor)
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
@@ -155,7 +155,7 @@ class SensorDetailFragment(
             sensorEntity.enabled = isEnabled
             sensorDao.update(sensorEntity)
         } else {
-            sensorEntity = Sensor(basicSensor.id, isEnabled, false, "")
+            sensorEntity = Sensor(basicSensor.id, isEnabled, false, true, "")
             sensorDao.add(sensorEntity)
         }
         refreshSensorData()
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorManager.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.content.pm.PackageManager
 import android.os.Process.myPid
 import android.os.Process.myUid
+import android.util.Log
 import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.database.AppDatabase
 import io.homeassistant.companion.android.database.sensor.Attribute
@@ -39,7 +40,7 @@ interface SensorManager {
 
         // If we haven't created the entity yet do so and default to enabled if required
         if (sensor == null) {
-            sensor = Sensor(sensorId, permission && enabledByDefault, false, "")
+            sensor = Sensor(sensorId, permission && enabledByDefault, false, true, "")
             sensorDao.add(sensor)
         }
 
@@ -67,6 +68,8 @@ interface SensorManager {
     ) {
         val sensorDao = AppDatabase.getInstance(context).sensorDao()
         val sensor = sensorDao.get(basicSensor.id) ?: return
+        Log.d("SensorManager", "Old sensor state ${sensor.state} compared to new state $state for ${sensor.name}")
+        sensor.stateChanged = (sensor.state != state.toString())
         sensor.id = basicSensor.id
         sensor.state = state.toString()
         sensor.stateType = when (state) {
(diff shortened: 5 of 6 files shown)


anyuta1166:
Could we check for state changes only when an update is triggered by an intent (that is, from SensorReceiver,onReceive function), and still send data from periodic worker every 15 minutes even if it was not changed (like a ping to show that the device is still available)?
There were a thoughts to make device unavailable if HA doesn't receive periodic updates, and you are trying to break it again and again.

dshokouhi:
Well the `device_tracker` should be checking into HA regardless of this change, this is for sensors only.  If we restrict receiver requests then the sensors that don't update with a receiver will update slower.  If we are getting the event we should let all sensors continue to benefit from it.

For what you are asking to make the device `unavailable` is something that should be handled at the core level of things because how can the device say it is out of the network or shut down if its not online to send it.  Core should determine based on the last communication sent from the device IMO.

anyuta1166:
> Well the `device_tracker` should be checking into HA regardless of this change, this is for sensors only.

Well, sounds reasonable.

anyuta1166:
Hmm.. I like the idea of sending updates only when a state change has been detected. But I don't like the idea of splitting up the sensors.

That's how my UI looks like: 

Screenshot_20200908-144334

I have an icon or title with basic information, and I can click on it to see the additional information (attributes), and it is very compact. I hate even splitting the battery sensor into battery level and battery state, because it takes up more space.
If you split, for example, Wi-Fi sensor into 8 sensors, where do I put all of them?

anyuta1166:
More than that, I think the best approach will be to filter out unwanted intents, as I did with alarm and bluetooth sensor. If we filter out intents for ALL disabled sensors (not only alarm and bluetooth) , that way each received intent will mean an update to the enabled sensor which we should send out anyway. Sending all sensors info instead of the one that just updated is a little overhead which is neglectable.

dshokouhi:
@anyuta1166 we were asked by the HA core team to split up the non-static attributes that update outside of the state changes.  So not all attributes but most of them need to be split up into their own sensors. This is something the core team does during their reviews and they want the mobile apps to follow them.

dshokouhi:
We also send all sensor info so the ones that don't update with a broadcast receiver can get faster updates.  Not all sensors and attributes update as fast as the rest, so there is no point in getting them to update slower they should be updating faster.  There is no extra battery drain associated with it.  Do you really want slower updates for other stuff?

anyuta1166:
> we were asked by the HA core team to split up the non-static attributes that update outside of the state changes.
> This is something the core team does during their reviews and they want the mobile apps to follow them.

Too bad. This is a bad decision. Looks like I'll have to maintain a fork with all those changes reverted :-(

anyuta1166:
> so the ones that don't update with a broadcast receiver can get faster updates.

Which sensors you are talking about that don't update with a broadcast receiver?

I think it's the battery sensor only. And as you said somewhere in previous conversations, we could subscribe to battery updates (but don't do it now).

dshokouhi:
> Too bad. This is a bad decision.


Imagine a new user who never used jinja wanted to automate based off attributes.  Its not good for a first time user to have to learn how to create a template sensor to achieve such a simple task.  Lots of first time users struggle with it so we need to make sure they are taken care of.  This why we added the ability to disable sensors by default so we don't overload new users with tons of things.  Not all of them like the new sensors enabled by default.  We have to think of the new user experience.

> Which sensors you are talking about that don't update with a broadcast receiver?

The audio sensor has several attributes that don't update with a receiver like volume and audio mode and is music active, the storage sensor, all hardware sensors, last reboot and any others that don't have a receiver.  We still have more sensors to add :)


> And as you said somewhere in previous conversations, we could subscribe to battery updates (but don't do it now).

Yea so after we get all sensors to stop sending unnecessary updates we can subscribe to battery changed and see if it does not make extra drain since it contains a lot of information.  We have to see how well it works but it should benefit all sensors since battery changed gets triggered when you do things like turn on/off screen.  When I did my testing it was sending tons of update spam and it was not good.

anyuta1166:
Am I right that this commit takes only sensor state changes into account? That is, if the state has not changed, but an attribute has changed, the updated info will not be send, right?

anyuta1166:
> Imagine a new user who never used jinja wanted to automate based off attributes. Its not good for a first time user to have to learn how to create a template sensor to achieve such a simple task.

I'm happy that HA becomes more user friendly and some things that required manual .yaml editing now can be done from web UI. But that's an overshoot in my opinion.

You can easily create a template sensor based on an attribute of another sensor, but you can't combine 8 different sensors back into 1, as far as I know.

ludeeus:
> but you can't combine 8 different sensors back into 1 sensor with attributes, as far as I know.

Sure you can ;)
<https://www.home-assistant.io/integrations/template/#attribute_templates>

```yaml
sensor:
  - platform: template
    sensors:
      solar_angle:
        value_template: "{{ state_attr('sun.sun', 'elevation') }}"
        attribute_templates:
          attr1: "{{ state_attr('sun.sun', 'elevation') }}"
```

dshokouhi:
> Am I right that this commit takes only sensor state changes into account? That is, if the state has not changed, but an attribute has changed, the updated info will not be sent, right?

From the PR description:

>  I will need to submit a follow up PR splitting up the non-static attributes so until that point some of the attributes will be slower to update. If you feel I should add them here to avoid that being a breaking change I can do so but this already feels like a big change in itself so I opted to keep them separate. So far Audio Sensor (all attributes), Battery State (all attributes but they all update anyways due to plug/unplug events with the exception of battery_health), Bluetooth (only for is_bt_on the rest update with the state), Storage Sensor (external attributes should become their own sensor however its tough to create on the fly if user removes the SD card so it may need to remain here) and Wifi Connection (all attributes) will be the ones that need to be split up immediately as we have broadcast receivers to update some of those attributes more quickly.

My plan was depending on what happens from the review to start splitting them out.  I can prepare PR's ahead of time but I suspect itll take a few days to get all that ready and merged when ready.  I am even ok with holding off on merging this and preparing the other PR's so they can all be merged together.  I will not be removing the attributes in those PR's  What we will do is for for 1 public release we will keep the attributes and have the new sensors split and disabled.  This will give users plenty of time to switch from template sensors over to the actual provided sensors.  Will need to make a blog post highlighting this change too.  Then after that release we can remove them but we will be giving users time to prepare.

> You can easily create a template sensor based on an attribute of another sensor, but you can't combine 8 different sensors back into 1 sensor with attributes, as far as I know.

you can: https://www.home-assistant.io/integrations/template/#attribute_templates

anyuta1166:
> Sure you can ;)

> you can: https://www.home-assistant.io/integrations/template/#attribute_templates

Thanks, I didn't know.

But currently sensors such as Wi-Fi and battery have a dynamic icon that is based on current connectivity or battery level, for example.
While making a template sensors, I can access state and attributes of an original sensor. But I don't see a way to access it's icon and copy it into template sensor.
There is a icon_template, but as I see it allows to set an icon based on a another sensor's state or attribute, not based on another sensor's current icon.

ludeeus:
You do that with a device class @anyuta1166.
But these kind of things are more suited for discord or the forum :+1:

JBassett (approved):
So,  I'm going to merge this.  We will stay in beta until all the sensors are broken out so the majority of users won't be affected by the multiple breaking changes and they will only have to deal with it once.

anyuta1166:
> You do that with a device class

Device class doesn't behave the same as battery sensor:
```
       var batteryIcon = "mdi:battery"

        if (chargingStatus == "unknown") {
            batteryIcon += "-unknown"

            return batteryIcon
        }

        if (isCharging)
            batteryIcon += "-charging"

        if (chargerType == "wireless")
            batteryIcon += "-wireless"

        val batteryStep: Int = percentage / 10
        batteryIcon += when (batteryStep) {
            0 -> "-outline"
            10 -> ""
            else -> "-${batteryStep}0"
        }

        return batteryIcon
```

anyuta1166:
> You do that with a device class

Oh, I figured this out. I can use `icon_template: "{{ state_attr('sensor.my_device_battery_level', 'icon') }}"`

Anyway, there is going to be a struggle with a bunch of template sensors in favor of users who can't deal with them.

anyuta1166:
> 
> 
> We also send all sensor info so the ones that don't update with a broadcast receiver can get faster updates. Not all sensors and attributes update as fast as the rest, so there is no point in getting them to update slower they should be updating faster. There is no extra battery drain associated with it. Do you really want slower updates for other stuff?

@dshokouhi Sorry for posting here, I just wanted to answer you.

You say, no extra battery drain?

My parents recently started complaining about the battery drain. Previously the battery lived for 2-3 days, but some time ago it started discharging in just 1 day, causing inconvenience.

After investigation I've found that the cause is the recent Home Assistant version.

I've installed my custom build on their phones and they are happy again. The battery drain is gone.

The most significant related changes in my custom build are:

- an intent doesn't trigger sensors update if the corresponding sensor is enabled (applied to all sensors, not only alarm and bluetooth)
- sensors update doesn't trigger location sensor. Location sensor works independently and is only triggered by the periodic work and by passively received location updates that we are subscribed for (why do we ever need to explicitly request a location if we are already subscribed for updates?).

The battery discharge curve has changed drastically after these modifications.

PR #901, opened 2020-09-08
Split up bluetooth state into its own sensor, part 1

dshokouhi:
Following on from #891 we are splitting up one attribute from the bluetooth sensor.  Part 1 just creating the sensor and leaving the attribute where it is.

`is_bt_on` is now a binary sensor

This is the only attribute that updates outside of the state updates for the existing bluetooth connection sensor.  The other attributes only update when the state of bluetooth connection changes. Paired devices is a list of devices the user has paired to the android system, like wise those devices will only show up under connected paired devices.  Some devices may or may not pair with the system while also pairing with their stand alone app.  These devices are also caught and reported however they are either connected or not connected so they also update with the state regardless of where they are paired. The state always reflects the total count of connected devices.

I have also adjusted the receiver logic to account for this change.  Previously `ACTION_STATE_CHANGED` was added only for this attribute to update faster.  The attribute will still update with this change only a few seconds slower as bluetooth is waiting for the device to connect.  I made this split in case a user decides they only want one sensor and not both.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
@@ -98,7 +98,7 @@ class BluetoothSensorManager : SensorManager {
             mapOf(
                 "connected_paired_devices" to connectedPairedDevices,
                 "connected_not_paired_devices" to connectedNotPairedDevices,
-                "is_bt_on" to isBtOn,  // Remove after next release
+                "is_bt_on" to isBtOn, // Remove after next release
                 "paired_devices" to bondedString
             )
         )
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
```diff
@@ -18,14 +18,20 @@ class BluetoothSensorManager : SensorManager {
             R.string.sensor_description_bluetooth_connection,
             unitOfMeasurement = "connection(s)"
         )
+        val bluetoothState = SensorManager.BasicSensor(
+            "bluetooth_state",
+            "binary_sensor",
+            R.string.basic_sensor_name_bluetooth_state,
+            R.string.sensor_description_blueooth_state
```
anyuta1166, line 25: There is a typo - blueooth

app/src/main/res/values/strings.xml
```diff
@@ -157,6 +157,7 @@ like to connect to:</string>
   <string name="sensor_description_interactive">Whether or not the device is in an interactive state</string>
   <string name="sensor_description_doze">Whether or not the device is in Doze mode</string>
   <string name="sensor_description_power_save">Whether or not the device is in Power Save mode</string>
+  <string name="sensor_description_blueooth_state">Whether or not bluetooth is enabled on the device</string>
```
anyuta1166, line 160: There is a typo too - blueooth

JBassett (approved):

PR #906, opened 2020-09-09
Fix state change detection if the sensor is already registered

dshokouhi:
Noticed an issue that once a sensor registered a state change it kept sending updates, it also impacted enabling a sensor for the first time.  With this change we only send one update when it detects a change and if we fail to update it then we will consider sending another.  Also made the logging statement more meaningful


dshokouhi:
No longer necessary after #907

Issue #882, opened 2020-09-04
Feature Request - Public IP sensor whilst on 4G/LTE


pandanz:
**Is your feature request related to a problem? Please describe.**
This is not related to a problem but is a feature request that would enable other services to be accessed whilst away from home.

**Describe the solution you'd like**
It would be great for the Android App to provide a sensor to home assistant with the public IP address the phone is currently using whilst on 4G/LTE.

The use case is, so i can allow that single IP address through my firewall to access services via Nginx into my home network. This will minimise my potential exposure of opening up my firewall. I use Nabu Casa for HA but I would use this feature for other services. I can dynamically update my firewall via API/SSH so i would create an automation that is triggered based on the public IP address changing.

**Describe alternatives you've considered, if any**
I currently use a VPN when I need to access services however having a single IP address allowed which can dynamically change in my firewall makes access much simpler.

skynetua:
I've played around with this feature a bit and could get only internal ip in my cellular network. I'm not sure that any provider assign real public ip. Probably we need to use some external services, like `https://api.ipify.org?format=json` if we are going to add this.

PR #908, opened 2020-09-09
Initial public ip sensor commit

skynetua:
Fixes: #882 

PR introduces recently requested public ip sensor as a part of NetworkSensorManager. Feature uses https://www.ipify.org/ service api to determine ip address so you should agree using it when enable sensor.

SmartSelect_20200909-180824_Home Assistant

Sensor does not have any onChange event so it will be updated on regular basis and while wifi connection state changed so this should be enough.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
@@ -5,7 +5,16 @@ import android.content.Context
 import android.net.wifi.WifiInfo
 import android.net.wifi.WifiManager
 import android.os.Build
+import android.util.Log
 import io.homeassistant.companion.android.R
+import okhttp3.Call
+import okhttp3.Callback
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.Response
+import okio.IOException
+import org.json.JSONException
+import org.json.JSONObject
 
 class NetworkSensorManager : SensorManager {
     companion object {
@@ -55,6 +64,12 @@ class NetworkSensorManager : SensorManager {
             R.string.sensor_description_wifi_signal,
             unitOfMeasurement = "dBm"
         )
+        val publicIp = SensorManager.BasicSensor(
+            "public_ip_address",
+            "sensor",
+            R.string.basic_sensor_name_public_ip,
+            R.string.sensor_description_piblic_ip
+        )
     }
 
     override val enabledByDefault: Boolean
@@ -62,7 +77,16 @@ class NetworkSensorManager : SensorManager {
     override val name: Int
         get() = R.string.sensor_name_network
     override val availableSensors: List<SensorManager.BasicSensor>
-        get() = listOf(wifiConnection, bssidState, wifiIp, wifiLinkSpeed, wifiState, wifiFrequency, wifiSignalStrength)
+        get() = listOf(
+            wifiConnection,
+            bssidState,
+            wifiIp,
+            wifiLinkSpeed,
+            wifiState,
+            wifiFrequency,
+            wifiSignalStrength,
+            publicIp
+        )
 
     override fun requiredPermissions(): Array<String> {
         return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
@@ -85,6 +109,7 @@ class NetworkSensorManager : SensorManager {
         updateWifiSensor(context)
         updateWifiFrequencySensor(context)
         updateWifiSignalStrengthSensor(context)
+        updatePublicIpSensor(context)
     }
 
     private fun updateWifiConnectionSensor(context: Context) {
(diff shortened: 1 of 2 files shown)


dshokouhi:
So one issue with putting this in the NetworkSensorManager is that if the user had already granted location permissions then this sensor would be enabled by default.  Maybe the phone sensor is better since I know if the user has granted phone permissions new sensors are still disabled by default.

skynetua:
Hmm, I see it. So why other new sensors splitted from old `wifiConnection` are not enabled by defoult? Just checked on new installation.

skynetua:
It become enabled even if I set `enabledByDefault=false`. Need to dig in for investigation

dshokouhi:
@skynetua you know what I misspoke, you are right it would be disabled by default sorry about that :)

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
```diff
@@ -55,14 +64,29 @@ class NetworkSensorManager : SensorManager {
             R.string.sensor_description_wifi_signal,
             unitOfMeasurement = "dBm"
         )
+        val publicIp = SensorManager.BasicSensor(
+            "public_ip_address",
+            "sensor",
+            R.string.basic_sensor_name_public_ip,
+            R.string.sensor_description_piblic_ip
```
dshokouhi, line 71: typo in the name `piblic` should be `public` 😃

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
```diff
@@ -375,4 +407,34 @@ class NetworkSensorManager : SensorManager {
                 (ip shr 16 and 0xFF) + "." +
                 (ip shr 24 and 0xFF)
     }
+
+    private fun updatePublicIpSensor(context: Context) {
+        var ip = "unknown"
```
dshokouhi, line 412: You are missing the required permission call here I think so it may be forcing the sensor to update and become enabled.

See an example: https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt#L91

Also you may want to add a permission check like we are doing here: https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt#L99

app/src/main/res/values/strings.xml
```diff
@@ -241,6 +242,7 @@ like to connect to:</string>
   <string name="sensor_description_wifi_state">Whether or not WiFi is enabled on the device</string>
   <string name="sensor_description_wifi_frequency">The frequency band of the connected network</string>
   <string name="sensor_description_wifi_signal">The signal strength of the device to the WiFi network</string>
+  <string name="sensor_description_piblic_ip">The public IP address of the device. Enabling this you agree to use external service https://ipify.org to determine IP address.</string>
```
dshokouhi, line 245: Same as above 😃

dshokouhi (commented):
I believe these things are why this sensor and only this sensor becomes enabled on the settings screen.

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
dshokouhi: Not sure if the last permission check is required but I am pretty sure the first one is.

skynetua: Yep, first check is the reason why it become enabled. 
Permission check is not required.

dshokouhi (approved):
Tested the debug APK and did not see an issue.  Great to see a WTH being requested and worked on!

Once merged we should let these guys know :) https://community.home-assistant.io/t/wth-cant-i-have-a-sensor-based-on-the-public-ip-address-from-the-android-app/224258

pandanz:
@skynetua @dshokouhi, you guys work fast!!! Thanks heaps

JBassett (approved):

hanzch:
First of all, thank you all for your contributions. I encountered some small problems in use, you can refer to whether you need to optimize.

At present, it seems that the IP sensor can only be uploaded synchronously when the first APP is opened and initialized. Even when the APP is turned on and the WIFI/4G switch is switched, the instant update cannot be updated and the upload is updated again. I waited about 5 minutes.

May I ask whether it can be achieved. When the APP is running in the background, the update is uploaded with the WIFI change, or it is uploaded regularly (for example, update once in 60s).


APP version: 2021.10.0-full
HASS version: 2021.12.4
Android version: MIUI 12.5 (Android 11)
Phone model: Xiaomi 9

dshokouhi:
@hanzch please dont comment on old merged PRs, either create a new issue or ask for help in discord/forums

hanzch:
> @hanzch please dont comment on old merged PRs, either create a new issue or ask for help in discord/forums

Sorry, delete now

PR #919, opened 2020-09-11
Ensure only the master branch push creates a beta play store build.

JBassett:


.github/workflows/daily.yaml
```diff
@@ -52,7 +52,22 @@ jobs:
...
+
+      - name: Check SHA
+        id: check_sha
+        uses: actions/cache@v1
+        with:
+          path: check-SHA
+          key: check-SHA-${{ github.sha }}
+
       - name: Deploy to Playstore Beta
+        # Only run if a new commit is present.  This should prevent a new beta being created
```
dshokouhi, line 68: I noticed that when I update from firebase and there is no other new update for the day the play store still sends me the same build. Is that to be expected or is it technically a new build?

JBassett: They are built at different times but with the same version code.  They are essentially the same build, but android allows you to reinstall the APK so depending on who you install it from it might prompt to "update" but it won't do anything.

PR #922, opened 2020-09-12
Decouple location updates from other sensor updates.

anyuta1166:
I guess this will not be merged too, because it goes against recent sensor unification changes, but may be will give you an idea.

So the whole story.

My parents recently started complaining about the battery drain. Previously the battery lived for 2-3 days, but some time ago it started discharging in just 1 day, causing inconvenience.

After investigation I've found that the cause is the recent Home Assistant app version.

I've installed my custom build on their phones and they are happy again. The battery drain is gone.

The most significant related changes in my custom build are:

1. a received intent doesn't trigger sensors update if the corresponding sensor is disabled (applied to all sensors, not only alarm and bluetooth)
2. sensors update doesn't trigger location sensor. Location sensor works independently and is only triggered by the periodic work and by passively received location updates that we are subscribed for (why do we ever need to explicitly request a location on each and every sensors update call if we are already subscribed for updates?).

The battery discharge curve has changed drastically after these modifications.

The most significant change is the second one. And this PR is how I did it.

The issue is that every time we receive an intent we are subscribed for, we are triggering all sensors update. This in turn triggers a location request. 1 minute limitation didn't help too much (but broke sensors detail screen with every 10 seconds update).

As I said earlier, why do we ever need to explicitly request a location on each and every sensors update call if we are already subscribed for updates?

I've decoupled location updates from another sensors update, and also removed 1 minute limitation as it is not needed any more.

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
@@ -21,8 +21,6 @@ import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
 import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
 import io.homeassistant.companion.android.common.data.integration.UpdateLocation
-import io.homeassistant.companion.android.database.AppDatabase
-import io.homeassistant.companion.android.database.sensor.Attribute
 import javax.inject.Inject
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -289,17 +287,6 @@ class LocationSensorManager : BroadcastReceiver(), SensorManager {
             Log.w(TAG, "Not getting single accurate location because of permissions.")
             return
         }
-        val now = System.currentTimeMillis()
-        val sensorDao = AppDatabase.getInstance(latestContext).sensorDao()
-        val fullSensor = sensorDao.getFull(backgroundLocation.id)
-        val latestAccurateLocation = fullSensor?.attributes?.firstOrNull { it.name == "lastAccurateLocationRequest" }?.value?.toLongOrNull() ?: 0L
-        // Only update accurate location at most once a minute
-        if (now < latestAccurateLocation + 60000) {
-            Log.d(TAG, "Not requesting accurate location, last accurate location was too recent")
-            return
-        }
-        sensorDao.add(Attribute(backgroundLocation.id, "lastAccurateLocationRequest", now.toString(), "string"))
-
         val maxRetries = 5
         val request = createLocationRequest()
         request.priority = LocationRequest.PRIORITY_HIGH_ACCURACY
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
@@ -29,7 +29,6 @@ class SensorReceiver : BroadcastReceiver() {
             GeocodeSensorManager(),
             LastRebootSensorManager(),
             LightSensorManager(),
-            LocationSensorManager(),
             NetworkSensorManager(),
             NextAlarmManager(),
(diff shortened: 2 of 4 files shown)


rs443:
This sounds logic to me, that is how a regular user would expect it to work. Disabling all non location sensors to optimize battery usage while having a reliable and battery efficient location tracker.

How often do you see location updates with this build? Is there a risk that removing the location request limit would trigger too many requests draining the battery?

Frequent location updates when driving home to accurately trigger automations is very useful, but high frequent updates when stationary is not equally as favorable. Are there any logic in when and how often updates are being triggered? I was about to  suggest an enhancement request to slow down on updates when stationary while make them more frequent while moving and getting close to home zone. Like in https://github.com/gcobb321/icloud3 that uses a variable polling rate.

anyuta1166:
> How often do you see location updates with this build?

Interval of location updates is defined in the code as somewhere between 1 and 5 minutes. Actual interval varies depending on the system choice.
Here is an example from my Home Assistant logs:
```
2020-09-12 13:23:56 DEBUG (MainThread) [homeassistant.components.mobile_app.webhook] Received webhook payload from Anna for type update_location: {'location_name': '', 'gps': [55.6373486, 37.3570968], 'gps_accuracy': 12, 'speed': 0, 'altitude': 201, 'course': 87, 'vertical_accuracy': 2}
2020-09-12 13:24:56 DEBUG (MainThread) [homeassistant.components.mobile_app.webhook] Received webhook payload from Anna for type update_location: {'location_name': '', 'gps': [55.6373486, 37.3570968], 'gps_accuracy': 12, 'speed': 0, 'altitude': 201, 'course': 87, 'vertical_accuracy': 2}
2020-09-12 13:27:03 DEBUG (MainThread) [homeassistant.components.mobile_app.webhook] Received webhook payload from Anna for type update_location: {'location_name': '', 'gps': [55.6373486, 37.3570968], 'gps_accuracy': 12, 'speed': 0, 'altitude': 201, 'course': 87, 'vertical_accuracy': 2}
2020-09-12 13:28:49 DEBUG (MainThread) [homeassistant.components.mobile_app.webhook] Received webhook payload from Anna for type update_location: {'location_name': '', 'gps': [55.6373486, 37.3570968], 'gps_accuracy': 12, 'speed': 0, 'altitude': 201, 'course': 87, 'vertical_accuracy': 2}
2020-09-12 13:30:51 DEBUG (MainThread) [homeassistant.components.mobile_app.webhook] Received webhook payload from Anna for type update_location: {'location_name': '', 'gps': [55.6373486, 37.3570968], 'gps_accuracy': 12, 'speed': 0, 'altitude': 201, 'course': 87, 'vertical_accuracy': 2}
2020-09-12 13:32:56 DEBUG (MainThread) [homeassistant.components.mobile_app.webhook] Received webhook payload from Anna for type update_location: {'location_name': '', 'gps': [55.6373486, 37.3570968], 'gps_accuracy': 12, 'speed': 0, 'altitude': 201, 'course': 87, 'vertical_accuracy': 2}
2020-09-12 13:35:06 DEBUG (MainThread) [homeassistant.components.mobile_app.webhook] Received webhook payload from Anna for type update_location: {'location_name': '', 'gps': [55.6373486, 37.3570968], 'gps_accuracy': 12, 'speed': 0, 'altitude': 201, 'course': 87, 'vertical_accuracy': 2}
2020-09-12 13:37:10 DEBUG (MainThread) [homeassistant.components.mobile_app.webhook] Received webhook payload from Anna for type update_location: {'location_name': '', 'gps': [55.6373486, 37.3570968], 'gps_accuracy': 12, 'speed': 0, 'altitude': 201, 'course': 87, 'vertical_accuracy': 2}
```

> Is there a risk that removing the location request limit would trigger too many requests draining the battery?

No. That limit applied only to explicit requests, which are mostly removed. Since they are removed, the limit is no longer needed.

Note that these explicit requests was sent **in addition** to passively received location updates that we are already subscribed for, which caused significant battery drain. The limit was introduced as an attempt to fix this, but it was not successful.

JBassett:
We plan to include location in sensor updates for now.

PR #932, opened 2020-09-14
Change exoplayer/play_hls payload type

uvjustin:
The frontend may need to send additional options to the exoplayer/play_hls interface such as whether to be muted or not: https://github.com/home-assistant/frontend/issues/6857 . With that in mind it is probably better to use a JSONObject instead of a String for the payload.


uvjustin:
Since this is potentially breaking as the current app will not work with the new format of the call, we will temporarily remove ExoPlayer from the frontend side until this PR goes through and the app is updated.

JBassett (approved):

PR #938, opened 2020-09-15
Add the ability to share things (text) to Home Assistant.

JBassett:
share


dshokouhi (approved):
This is great! Can't wait to see the use cases for this feature!

dshokouhi:
Are we able to grab the package that the share came from? Might add to the experience with what is being shared :)

JBassett:
> Are we able to grab the package that the share came from? Might add to the experience with what is being shared :)

Added!

PR #890, opened 2020-09-07
Make minimum accuracy user configurable

anyuta1166:
Make minimum accuracy user configurable.
It's a long awaited feature and also fixes #868

PR #939, opened 2020-09-15
Allow sensors to have custom settings

dshokouhi:
This PR allows any sensor to have a custom setting to fit its need.  A table identical to attributes is created and most of the logic around the sensor detail fragment is the same too with the exception that we use a `EditTextPreference` to get the value for the manager to handle.  I chose to keep it separate from attributes since these are really more app specific but I think we can allow the managers to provide them as an attribute if we see the need.

With this PR each sensor manager is responsible for: 

* Updating its `BasicSensor.description` to describe each setting
* Creating the setting in the table and in proper order
* Handling null or unwanted values
* Setting `valueType: "number"` or `valueType: "string"` to handle a dynamic keyboard change
* Setting `valueType: "toggle"` if they require a toggle to enable a setting
* Setting a sensible default if required.

For this PR I have chosen `wifi_bssid` to have a custom setting. The user must first enable the toggle `get_next_bssid` then and only then will we create a new field to rename the current BSSID. The user will need to enable the toggle again if they want to rename another BSSID, the toggle will turn itself off after creating one field.  In my house I have 2 AP's so I have chosen to send `Den` or `Bedroom` depending on the connected mac address. This will negate the need to use a template sensor and without the need to use secret values to mask it.

Things to be worked in a future PR would be to allow a user to hide a setting and possibly reset them.  Later on we can use `valueType` to also dictate if instead of a `EditTextPreference` or `SwitchPreference` to use something like Mult-select List or Checkbox etc...  I think for now we should stick to some small basics for an MVP. In the current state it should allow a wide variety of things to be done including #890 

image

image

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
@@ -181,11 +181,18 @@ class NetworkSensorManager : SensorManager {
         val settingName = "replace_$bssid"
         val sensorDao = AppDatabase.getInstance(context).sensorDao()
         val fullSensor = sensorDao.getSettings(bssidState.id)
+        val getNextBSSID = fullSensor?.settings?.firstOrNull { it.name == "get_next_bssid" }?.value ?: "false"
         val currentSetting = fullSensor?.settings?.firstOrNull { it.name == settingName }?.value ?: ""
-        if (currentSetting != "") {
-            bssid = currentSetting
+        if (getNextBSSID == "true") {
+            if (currentSetting == "") {
+                sensorDao.add(Setting(bssidState.id, "get_next_bssid", "false", "toggle"))
+                sensorDao.add(Setting(bssidState.id, settingName, "", "string"))
+            }
         } else {
-            sensorDao.add(Setting(bssidState.id, settingName, "", "string"))
+            if (currentSetting != "")
+                bssid = currentSetting
+
+            sensorDao.add(Setting(bssidState.id, "get_next_bssid", "false", "toggle"))
         }
 
         val icon = if (bssid != "<not connected>") "mdi:wifi" else "mdi:wifi-off"
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
@@ -7,6 +7,8 @@ import android.net.wifi.WifiManager
 import android.os.Build
 import android.util.Log
 import io.homeassistant.companion.android.R
+import io.homeassistant.companion.android.database.AppDatabase
+import io.homeassistant.companion.android.database.sensor.Setting
 import okhttp3.Call
 import okhttp3.Callback
 import okhttp3.OkHttpClient
@@ -174,7 +176,18 @@ class NetworkSensorManager : SensorManager {
             conInfo = wifiManager.connectionInfo
         }
 
-        val bssid = if (conInfo!!.bssid == null) "<not connected>" else conInfo.bssid
+        var bssid = if (conInfo!!.bssid == null) "<not connected>" else conInfo.bssid
+
+        val settingName = "replace_$bssid"
+        val sensorDao = AppDatabase.getInstance(context).sensorDao()
+        val fullSensor = sensorDao.getSettings(bssidState.id)
+        val currentSetting = fullSensor?.settings?.firstOrNull { it.name == settingName }?.value ?: ""
+        if (currentSetting != "") {
+            bssid = currentSetting
+        } else {
+            sensorDao.add(Setting(bssidState.id, settingName, "", "string"))
+        }
+
         val icon = if (bssid != "<not connected>") "mdi:wifi" else "mdi:wifi-off"
         onSensorUpdated(
             context,
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
@@ -72,6 +72,7 @@ class NetworkSensorManager : SensorManager {
             R.string.basic_sensor_name_public_ip,
             R.string.sensor_description_public_ip
         )
+        private const val GET_CURRENT_BSSID = "get_current_bssid"
     }
 
     override val enabledByDefault: Boolean
@@ -180,12 +181,12 @@ class NetworkSensorManager : SensorManager {
 
         val settingName = "replace_$bssid"
         val sensorDao = AppDatabase.getInstance(context).sensorDao()
-        val fullSensor = sensorDao.getSettings(bssidState.id)
-        val getCurrentBSSID = fullSensor?.firstOrNull { it.name == "get_current_bssid" }?.value ?: "false"
-        val currentSetting = fullSensor?.firstOrNull { it.name == settingName }?.value ?: ""
+        val sensorSettings = sensorDao.getSettings(bssidState.id)
+        val getCurrentBSSID = sensorSettings?.firstOrNull { it.name == GET_CURRENT_BSSID }?.value ?: "false"
+        val currentSetting = sensorSettings?.firstOrNull { it.name == settingName }?.value ?: ""
         if (getCurrentBSSID == "true") {
             if (currentSetting == "") {
-                sensorDao.add(Setting(bssidState.id, "get_current_bssid", "false", "toggle"))
+                sensorDao.add(Setting(bssidState.id, GET_CURRENT_BSSID, "false", "toggle"))
                 sensorDao.add(Setting(bssidState.id, settingName, bssid, "string"))
             }
         } else {
@@ -194,7 +195,7 @@ class NetworkSensorManager : SensorManager {
             else
                 sensorDao.removeSetting(bssidState.id, settingName)
 
-            sensorDao.add(Setting(bssidState.id, "get_current_bssid", "false", "toggle"))
+            sensorDao.add(Setting(bssidState.id, GET_CURRENT_BSSID, "false", "toggle"))
         }
 
         val icon = if (bssid != "<not connected>") "mdi:wifi" else "mdi:wifi-off"
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
@@ -181,18 +181,20 @@ class NetworkSensorManager : SensorManager {
         val settingName = "replace_$bssid"
         val sensorDao = AppDatabase.getInstance(context).sensorDao()
         val fullSensor = sensorDao.getSettings(bssidState.id)
-        val getNextBSSID = fullSensor?.settings?.firstOrNull { it.name == "get_next_bssid" }?.value ?: "false"
-        val currentSetting = fullSensor?.settings?.firstOrNull { it.name == settingName }?.value ?: ""
-        if (getNextBSSID == "true") {
+        val getCurrentBSSID = fullSensor?.firstOrNull { it.name == "get_current_bssid" }?.value ?: "false"
+        val currentSetting = fullSensor?.firstOrNull { it.name == settingName }?.value ?: ""
+        if (getCurrentBSSID == "true") {
             if (currentSetting == "") {
-                sensorDao.add(Setting(bssidState.id, "get_next_bssid", "false", "toggle"))
-                sensorDao.add(Setting(bssidState.id, settingName, "", "string"))
+                sensorDao.add(Setting(bssidState.id, "get_current_bssid", "false", "toggle"))
+                sensorDao.add(Setting(bssidState.id, settingName, bssid, "string"))
             }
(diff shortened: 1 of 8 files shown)


anyuta1166:
> Any mac address found by this sensor will have a setting in place so the user can replace the mac address with something thats more meaningful to them

Is there an ability to delete a setting? If a user connects to a large number of BSSIDs, he will end up with dozens of settings.

For example, I often use public Wi-Fi network. SSID is always the same, but each bus, bus station itself, train and subway cars have a different BSSID.

anyuta1166:
> In the current state it should allow a wide variety of things to be done including #890

Great, but I prefer my own solution. Your solution means there will be 3 different settings (for background location, zone location and geocoded location), hidden deeper inside the sensor settings.

anyuta1166:
If there's no ability to delete collected BSSIDs, at least make a setting "Collect BSSIDs" or like that, which can be turned off (or better be off by default).

skynetua:
Why do we need to collect them? As for me, we should store only user-added things into the one single string, like pairs - BSID1:ALIAS1, BSID2:ALIAS2. Or even better like json array.

dshokouhi:
> Is there an ability to delete a setting?

I mentioned that we should look into doing this after this PR, for a MVP PR that will add to the complexity of things and we should keep it simple first.  We need to think of a way that works for all sensors and their needs.  Not all sensors will want a setting to be deleted, every sensor has a unique need.


> Great, but I prefer my own solution. Your solution means there will be 3 different settings (for background location, zone location and geocoded location), hidden deeper inside the sensor settings.

This was already addressed in #890 and it was mentioned that they should be separated.  Logically speaking inside sensor details, where you enable a sensor, is the best place to keep the settings as the description is there and we can properly translate it.  These settings are meant to have a default value that is sensible so in most cases a user won't need to touch them.

> (or better be off by default - because we already have SSID sensor and BSSID -> custom name translation should be for advanced users).

This sensor is already disabled by default and SSID is only enabled if the user grants location permissions, users have to explicitly enable this sensor.  Many users would still consider BSSID, by itself, an advanced sensor so I don't see many enabling it.

> Why do we need to collect them?

I mentioned the why in my post, its to help reduce the need for template sensors and secret values making the automation aspect more user friendly.  Anyone who wants to automate off this will need to use a secret value or a secret inside a template sensor just to get the proper names displayed.  I don't expect people to remember MAC addresses but I do expect them to know which ones they care about and to want to name them.   In a household with multiple AP's it can be very helpful to see where a person is.


> As for me, we should store only user-added things into the one single string, like pairs - BSID1:ALIAS1, BSID2:ALIAS2. Or even better like json array.

This greatly adds to the complexity for first release I was thinking only simple text fields so sensors can do what they need to do without getting too crazy initially.  There won't be too many BSSID usage here only really for people who have multiple AP's and want to track them.  Normal users only need to stick to the Wifi connection sensor as that contains the SSID name.  Also creating a JSON array in a text field would not be user friendly.

dshokouhi:
Ok I have now added some more logic here to address the concerns raised around too many entries, thank you for that.  I will update the original post but here is whats new:

Managers can now define `valueType: "toggle"` so they can create a toggle as needed.  Managers are responsible for creating the toggle, creating the settings in a proper order that makes sense so if the toggle should be the first option it should be created first. The default should be `false`.

The sensor detail fragment will handle creating the toggle based on the `valueType` and also will handle setting the value to `"true"` and storing it so the manager can continue with its logic.

So for `wifi_bssid` the flow will be user enables the toggle, next update we set the toggle to off and create one new field for the current BSSID.  User can then rename the BSSID as they see fit.  Then the next time they encounter a BSSID they want to rename they will need to enable the toggle again to grab the current value and store it.

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
```diff
@@ -174,7 +176,25 @@ class NetworkSensorManager : SensorManager {
...
+        var bssid = if (conInfo!!.bssid == null) "<not connected>" else conInfo.bssid
+
+        val settingName = "replace_$bssid"
+        val sensorDao = AppDatabase.getInstance(context).sensorDao()
+        val fullSensor = sensorDao.getSettings(bssidState.id)
+        val getNextBSSID = fullSensor?.settings?.firstOrNull { it.name == "get_next_bssid" }?.value ?: "false"
+        val currentSetting = fullSensor?.settings?.firstOrNull { it.name == settingName }?.value ?: ""
+        if (getNextBSSID == "true") {
+            if (currentSetting == "") {
+                sensorDao.add(Setting(bssidState.id, "get_next_bssid", "false", "toggle"))
```
JBassett, line 188: If we are still on the same BSSID on the next update should we turn it off?

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
```diff
@@ -174,7 +176,25 @@ class NetworkSensorManager : SensorManager {
...
+                sensorDao.add(Setting(bssidState.id, settingName, "", "string"))
```
JBassett, line 189: Default to the non original value?

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
```diff
@@ -174,7 +176,25 @@ class NetworkSensorManager : SensorManager {
...
+            }
+        } else {
+            if (currentSetting != "")
+                bssid = currentSetting
```
JBassett, line 193: Maybe if `currentSetting == ""` we should delete the setting?

app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
```diff
@@ -147,6 +152,61 @@ class SensorDetailFragment(
             } else
                 it.isVisible = false
         }
+
+        findPreference<PreferenceCategory>("sensor_settings")?.let {
+            if (sensorData.enabled && !settings.isNullOrEmpty()) {
+                settings.forEach { setting ->
```
JBassett, line 158: Maybe we should sort these?  Not sure if I have a good suggestion on what to sort on though...

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
```diff
@@ -147,6 +152,61 @@ class SensorDetailFragment(
...
+                        pref.isChecked = setting.value == "true"
+                        pref.isIconSpaceReserved = false
+                        pref.setOnPreferenceChangeListener { _, newState ->
+                            val isEnabled = newState as Boolean
+
+                            if (isEnabled) {
+                                sensorDao.add(Setting(basicSensor.id, setting.name, "true", "toggle"))
+                                return@setOnPreferenceChangeListener true
+                            }
+                            return@setOnPreferenceChangeListener false
```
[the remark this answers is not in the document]
dshokouhi: I think in that case we can use a checkbox preference then if the value should stick instead of flipping?

app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
```diff
@@ -147,6 +152,61 @@ class SensorDetailFragment(
...
+                        }
+                        if (!it.contains(pref))
+                            it.addPreference(pref)
+                    }
+                    if (setting.valueType == "string" || setting.valueType == "number") {
```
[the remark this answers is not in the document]
dshokouhi: Sure I can do that

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
dshokouhi: I think I should rename it to say `get_current_bssid` since thats actually what it is doing :) I thought of this the other night since while the user is on this screen we update it every 10 seconds so it would be pretty difficult for it to be the actual next BSSID in reality.

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
dshokouhi: Well if the user did not change the setting fast enough it would remove it so I don't think that would be friendly.  The page where they create the entries updates every 10 seconds so I wanted to be mindful of that :)

app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
dshokouhi: So that was why I was thinking the manager should be responsible for the order they create the settings. This way when we create the page they appear in an expected order.  If we do any sorting then the managers would need to make sure the names are in expected order. Since the setting names are displayed to the user in a `key` format I felt this to be a better approach.

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
dshokouhi: I kept it this way in case the user took too long on the settings page due to the refresh timing.  This way to them it all looks the same until they set the name they choose.

JBassett: I was suggesting this:
```suggestion
                sensorDao.add(Setting(bssidState.id, settingName, bssid, "string"))
```
This way the setting gets added but it will still report the same value to Home Assistant and not instantly delete ;-)

app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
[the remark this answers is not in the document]
JBassett: By returning false you are telling the preference that they can't change the state.  So they can enable the preference but if they instantly regret the decision they can't turn if off until the manager does it.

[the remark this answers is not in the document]
JBassett: Try to quickly toggle on then off, you won't be able it.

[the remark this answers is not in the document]
dshokouhi: Oh so we should handle setting it to false here and always return true?

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
JBassett: Above I suggest that we change the default to the string of the BSSID rather than `""`

app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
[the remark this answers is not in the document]
JBassett: That was my thought, yep

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
```diff
@@ -181,18 +181,20 @@ class NetworkSensorManager : SensorManager {
         val settingName = "replace_$bssid"
         val sensorDao = AppDatabase.getInstance(context).sensorDao()
         val fullSensor = sensorDao.getSettings(bssidState.id)
```
JBassett, line 183: Since this is the list of sensor settings we should change the name probably.

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
```diff
@@ -181,18 +181,20 @@ class NetworkSensorManager : SensorManager {
...
-        val getNextBSSID = fullSensor?.settings?.firstOrNull { it.name == "get_next_bssid" }?.value ?: "false"
-        val currentSetting = fullSensor?.settings?.firstOrNull { it.name == settingName }?.value ?: ""
-        if (getNextBSSID == "true") {
+        val getCurrentBSSID = fullSensor?.firstOrNull { it.name == "get_current_bssid" }?.value ?: "false"
```
JBassett, line 184: Maybe extract `get_current_bssid` as a constant?

app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
```diff
@@ -147,6 +151,61 @@ class SensorDetailFragment(
...
+                        pref.isChecked = setting.value == "true"
+                        pref.isIconSpaceReserved = false
+                        pref.setOnPreferenceChangeListener { _, newState ->
+                            val isEnabled = newState as Boolean
+
+                            if (isEnabled) {
+                                sensorDao.add(Setting(basicSensor.id, setting.name, "true", "toggle"))
+                            } else {
+                                sensorDao.add(Setting(basicSensor.id, setting.name, "false", "toggle"))
+                            }
```
JBassett, line 172:
```suggestion
                            sensorDao.add(Setting(basicSensor.id, setting.name, isEnabled.toString(), "toggle"))
```

JBassett (approved):

PR #941, opened 2020-09-16
Hide NFC if the device reports it has no hardware

dshokouhi:
Was helping out a user in Discord who could not see the NFC option and it ended up being his HA version was not on 0.114 or higher.  While inspecting the code I noticed we do not check for NFC hardware when we look to make the option visible.  Now we will do a check with package manager and if successful we will continue with current logic looking at the HA version, if unavailable the option will be hidden.  Tested on my pixel 4 xl that has NFC support as well as 2 amazon tablets that do not have NFC, NFC remains hidden on tablets that don't have the hardware.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
@@ -1,5 +1,6 @@
 package io.homeassistant.companion.android.settings
 
+import android.content.pm.PackageManager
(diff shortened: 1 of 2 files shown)


JBassett (changes requested):
We should also add this here so the UI doesn't think the device can write NFC tags: https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt#L362

JBassett (approved):

Issue #910, opened 2020-09-09
Last reboot sensor reports changing values


Hypfer:
**Home Assistant Android version:** beta-367-3a39428-full

**Android version:** 9

**Phone model:** Moto Z3 Play

**Home Assistant version:** 0.114.3

**Last working Home Assistant release (if known):** /

**Description of problem:**

The last reboot sensor changes its value by a few seconds, even though the device isn't rebooted.
It's probably not a rounding error, since there can be 5s+ differences.

**Traceback (if applicable):**

```

```

**Screenshot of problem:**
![image](https://user-images.githubusercontent.com/974410/92634219-3dada480-f2d4-11ea-8f48-788a356a36d8.png)
![image](https://user-images.githubusercontent.com/974410/92634371-7ea5b900-f2d4-11ea-9395-9faf20a9be90.png)
![image](https://user-images.githubusercontent.com/974410/92634381-85ccc700-f2d4-11ea-862e-cbd17b834138.png)

**Additional information:**

It doesn't happen with the exact same version of the app running on a Samsung SM-T510 tablet with Android 10 so might either be a quirk with this particular phone _or_ Android 9.
Of course, the issue could also be load-related.

Looking a the code, there's really not much that could go wrong so I'm very confused
https://github.com/home-assistant/android/blob/c221e85fa4e5fbf0157bc3ce21394c1723cd3d36/app/src/main/java/io/homeassistant/companion/android/sensors/LastRebootSensorManager.kt#L53-L61

dshokouhi:
You know I have yet to see this issue on my Pixel...this might be one of those device differences.  One thing we can consider doing is removing the seconds completely as I had already stripped out microseconds but it can still flip flop if its at that sweet spot.

dshokouhi:
Another thing we can do is to check the value we last stored compared to the new value and only change it if its more than a few minutes difference.

Hypfer:
As mentioned in the discord, the time used to calculate this value is by default getting synced with the mobile network.
So I've turned off the mobile network time sync option in the android settings and this is how the sensor looks like now:
![image](https://user-images.githubusercontent.com/974410/92700013-97eb4b80-f34e-11ea-8ab0-e32704836bb1.png)

E-Plus/O2/Telefonica Germany is known for recurring issues regarding the network time. This also explains why the Wifi-only Tablet doesn't show this behaviour, nor do other phones on non-broken carriers.

[...]

PR #951, opened 2020-09-19
Add deadband setting for last reboot sensor

dshokouhi:
Fixes: #910 

As we noticed that there may be a difference between cell phone and cell tower time lets add an `offset` for the last reboot sensor. The default value is set to `60000` or 1 minute, we can adjust it but I think this should fix the issue for most users.  Making it a setting in case they want to make the value larger or smaller.  The description has been updated to explain the setting and I have also moved some static strings to be constant values.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/LastRebootSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/LastRebootSensorManager.kt
@@ -17,7 +17,7 @@ class LastRebootSensorManager : SensorManager {
     companion object {
         private const val TAG = "LastReboot"
         private const val LOCAL_TIME = "Local Time"
-        private const val OFFSET = "offset"
+        private const val DEADBAND = "deadband"
         private const val TIME_MILLISECONDS = "Time in Milliseconds"
 
         private val lastRebootSensor = SensorManager.BasicSensor(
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/LastRebootSensorManager.kt
```diff
@@ -49,8 +55,19 @@ class LastRebootSensorManager : SensorManager {
         var local = ""
         var utc = "unavailable"
 
+        val sensorDao = AppDatabase.getInstance(context).sensorDao()
+        val fullSensor = sensorDao.getFull(lastRebootSensor.id)
+        val sensorSetting = sensorDao.getSettings(lastRebootSensor.id)
+        val lastTimeMillis = fullSensor?.attributes?.firstOrNull { it.name == TIME_MILLISECONDS }?.value?.toLongOrNull() ?: 0L
+        val settingOffset = sensorSetting?.firstOrNull { it.name == OFFSET }?.value?.toIntOrNull() ?: 60000
```
JBassett, line 62: I'm not sure if offset is the best term to describe this.  It's more of a deadband for the reboot sensor right?

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/sensors/LastRebootSensorManager.kt
dshokouhi: Yea that would be a better way to describe it 😄

JBassett (approved):

Issue #953, opened 2020-09-19
App crashes when you try to add a field for Service Button


Mincka:
**Home Assistant Android version:**
beta-383-debeadb-full
**Android version:**
10
**Phone model:**
Samsung S20+
**Home Assistant version:**
0.115
**Last working Home Assistant release (if known):**
N/A
**Description of problem:**
When I tap on "Add field" button in the Service Button setup view, the app crashes in certain conditions.

1) Set the Service (e.g. `script.turn_on`)
2) Choose an `entity_id`
3) Tap on "Add field"
4) Enter a name and validate -> the app crashes

If the field is added *before* choosing the `entity_id`, it does not crash. 🤯

dshokouhi:
Thanks for the report and detailed steps to reproduce 👍 

Here is the crash log:

```
2020-09-19 09:40:42.623 6401-6401/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: io.homeassistant.companion.android, PID: 6401
    java.lang.ClassCastException: java.util.Collections$SingletonList cannot be cast to java.lang.String
        at io.homeassistant.companion.android.widgets.WidgetDynamicFieldAdapter.onBindViewHolder(WidgetDynamicFieldAdapter.kt:130)
        at io.homeassistant.companion.android.widgets.WidgetDynamicFieldAdapter.onBindViewHolder(WidgetDynamicFieldAdapter.kt:16)
        at androidx.recyclerview.widget.RecyclerView$Adapter.onBindViewHolder(RecyclerView.java:7065)
        at androidx.recyclerview.widget.RecyclerView$Adapter.bindViewHolder(RecyclerView.java:7107)
        at androidx.recyclerview.widget.RecyclerView$Recycler.tryBindViewHolderByDeadline(RecyclerView.java:6012)
        at androidx.recyclerview.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(RecyclerView.java:6279)
        at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:6118)
        at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:6114)
        at androidx.recyclerview.widget.LinearLayoutManager$LayoutState.next(LinearLayoutManager.java:2303)
        at androidx.recyclerview.widget.LinearLayoutManager.layoutChunk(LinearLayoutManager.java:1627)
[...]

PR #954, opened 2020-09-19
Fix dynamic field crash for service call widget

dshokouhi:
Fixes: #953 

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/WidgetDynamicFieldAdapter.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/WidgetDynamicFieldAdapter.kt
@@ -2,6 +2,7 @@ package io.homeassistant.companion.android.widgets
 
 import android.text.Editable
 import android.text.TextWatcher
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/widgets/WidgetDynamicFieldAdapter.kt
```diff
@@ -127,7 +128,11 @@ class WidgetDynamicFieldAdapter(
         // Currently value can by Any? but will currently only be storing String?
         // This may have to be changed later if multi-select gets implemented
         if (serviceFieldList[position].value != null) {
-            serviceFieldList[position].value as String
+            try {
+                serviceFieldList[position].value as String
+            } catch (e: Exception) {
+                Log.d("WidgetField", "Unable to get service field list", e)
```
JBassett, line 134: We should extract the TAG to a variable.
```suggestion
                Log.d(TAG, "Unable to get service field list", e)
```

JBassett (approved):

PR #958, opened 2020-09-19
Catch the exception when we get null for attributes

dshokouhi:
Fixes the following sentry error:

```
kotlin.TypeCastException: null cannot be cast to non-null type kotlin.collections.Map<*, *>
    at io.homeassistant.companion.android.widgets.StaticWidget.resolveTextToShow(StaticWidget.kt:128)
    at io.homeassistant.companion.android.widgets.StaticWidget$resolveTextToShow$1.invokeSuspend
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
    at android.os.Handler.handleCallback(Handler.java:883)
    at android.os.Handler.dispatchMessage(Handler.java:100)
    at android.os.Looper.loop(Looper.java:214)
    at android.app.ActivityThread.main(ActivityThread.java:7697)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:516)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:950)
```

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidget.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidget.kt
@@ -128,7 +128,7 @@ class StaticWidget : AppWidgetProvider() {
         var fetchedAttributes: Map<*, *>
         var attributeValues: List<String?>
         try {
-            fetchedAttributes = entity?.attributes as Map<*, *>
+            fetchedAttributes = (entity?.attributes as? Map<*, *>)!!
             attributeValues = attributeIds.split(",").map { id -> fetchedAttributes.get(id)?.toString() }
             return entity.state.plus(if (attributeValues.isNotEmpty()) stateSeparator else "").plus(attributeValues.joinToString(attributeSeparator))
         } catch (e: Exception) {
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidget.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidget.kt
@@ -128,9 +128,9 @@ class StaticWidget : AppWidgetProvider() {
         var fetchedAttributes: Map<*, *>
         var attributeValues: List<String?>
         try {
-            fetchedAttributes = (entity?.attributes as? Map<*, *>)!!
+            fetchedAttributes = entity?.attributes as? Map<*, *> ?: mapOf<String, String>()
             attributeValues = attributeIds.split(",").map { id -> fetchedAttributes.get(id)?.toString() }
-            return entity.state.plus(if (attributeValues.isNotEmpty()) stateSeparator else "").plus(attributeValues.joinToString(attributeSeparator))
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidget.kt
```diff
@@ -125,9 +125,16 @@ class StaticWidget : AppWidgetProvider() {
 
         if (attributeIds == null) return entity?.state
 
-        val fetchedAttributes = entity?.attributes as Map<*, *>
-        val attributeValues = attributeIds.split(",").map { id -> fetchedAttributes.get(id)?.toString() }
-        return entity.state.plus(if (attributeValues.isNotEmpty()) stateSeparator else "").plus(attributeValues.joinToString(attributeSeparator))
+        var fetchedAttributes: Map<*, *>
+        var attributeValues: List<String?>
+        try {
+            fetchedAttributes = entity?.attributes as Map<*, *>
```
JBassett, line 131: By putting this in a try catch we stop the crash but don't fix the problem.  The problem is `entity?.attributes` is returning null and trying to be cast to a map.  If you add `as?` it will still return null
```suggestion
            fetchedAttributes = entity?.attributes as? Map<*, *>
```

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidget.kt
dshokouhi: Looks like Kotlin wanted extra null checks here :)

```
(entity?.attributes as? Map<*, *>)!!
```

app/src/main/java/io/homeassistant/companion/android/widgets/StaticWidget.kt
```diff
@@ -125,9 +125,16 @@ class StaticWidget : AppWidgetProvider() {
 
         if (attributeIds == null) return entity?.state
 
-        val fetchedAttributes = entity?.attributes as Map<*, *>
-        val attributeValues = attributeIds.split(",").map { id -> fetchedAttributes.get(id)?.toString() }
-        return entity.state.plus(if (attributeValues.isNotEmpty()) stateSeparator else "").plus(attributeValues.joinToString(attributeSeparator))
+        var fetchedAttributes: Map<*, *>
+        var attributeValues: List<String?>
+        try {
+            fetchedAttributes = (entity?.attributes as? Map<*, *>)!!
```
JBassett, line 131: So the `!!`  is short hand to ensure the object isn't null so that would have thrown the same/similar error.  You want to declare the map as follows:
```suggestion
            fetchedAttributes = entity?.attributes as? Map<*, *> ?: mapOf<String, String>()
```
That will set fetchedAttributes to an empty map if null.

dshokouhi: Oh ok that makes sense, thanks for your help :)

JBassett (approved):

Issue #950, opened 2020-09-18
Expose values from the `android.net.TrafficStats` or `NetworkStats` API as sensors.


kquinsland:
**Is your feature request related to a problem? Please describe.**
This is not related to a performance issue or other bug. This is purely a feature-request. It is not a high priority feature as there is a workaround that will work in some cases.

**Describe the solution you'd like**
for the past few releases, android has been making a *lot* of changes to some of the network related APIs. Fortunately, the [`android.net.trafficstats`](https://developer.android.com/reference/android/net/TrafficStats) API is still standing.

As is the (newer) [`NetworkStatsManager`](https://developer.android.com/reference/android/app/usage/NetworkStatsManager.html) API.

I am asking for the HomeAssistant app to expose the device traffic stats as sensors to home assistant. The APIs linked above allow for traffic stats per network interface which may be useful in some cases... but not mine. My intended use case is to monitor the amount of data sent/received via the cell modem so HA can alert me when I am close to the limits for the month.

This feature will be useful for people that rely on LTE for internet. HA could automatically configure a router or firewall or piHole to limit connectivity if the data usage crosses a threshold as well.

**Describe alternatives you've considered, if any**
Currently, I am using Tasker to send a SMS off to the carrier and parse the payload when it comes back. This works, but is very fragile and relies on having Tasker and the carrier offering a SMS number that can be used to check this.

**Additional context**

https://stackoverflow.com/questions/17674790/how-do-i-programmatically-show-data-usage-of-all-applications

**EDIT:** Markdown links...

kquinsland:
Absolutely mind blowing how fast you got that added.  Thanks for implementing this!

PR #961, opened 2020-09-20
Add traffic stat sensors mobile and total data usage for transmitting and receving

dshokouhi:
4 new sensors, all disabled by default. Fixes: #950 

* Mobile Rx GB
* Mobile Tx GB
* Total Rx GB
* Total Tx GB

The mobile sensors are only visible on devices that have cellular data, it is safe to assume that all devices will report total data as they are online. I opted not to check if the device has phone capabilities so we can include tablets that have cellular data.  This check is done in `hasSensor` which gets called before the available sensors so we are hiding them properly.  Tested this in the emulator, my pixel 4 xl and a Android 5.1 tablet.

Data is converted from bytes to GB as that is what the world measures in data usage.  Set to 3 decimal places so people don't see 0.00 like I did 😛

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/TrafficStatsManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/TrafficStatsManager.kt
@@ -91,6 +91,7 @@ class TrafficStatsManager : SensorManager {
             mobileRx = TrafficStats.getMobileRxBytes().toFloat()
         } catch (e: Exception) {
             Log.e(TAG, "Error getting the mobile rx bytes", e)
+            return
         }
 
         mobileRx /= GB
@@ -114,6 +115,7 @@ class TrafficStatsManager : SensorManager {
             mobileTx = TrafficStats.getMobileTxBytes().toFloat()
         } catch (e: Exception) {
             Log.e(TAG, "Error getting the mobile tx bytes", e)
+            return
         }
 
         mobileTx /= GB
@@ -136,6 +138,7 @@ class TrafficStatsManager : SensorManager {
             totalRx = TrafficStats.getTotalRxBytes().toFloat().absoluteValue
         } catch (e: Exception) {
             Log.e(TAG, "Error getting the total rx bytes", e)
+            return
         }
         totalRx /= GB
 
@@ -159,6 +162,7 @@ class TrafficStatsManager : SensorManager {
             totalTx = TrafficStats.getTotalTxBytes().toFloat().absoluteValue
         } catch (e: Exception) {
             Log.e(TAG, "Error getting the total tx bytes", e)
+            return
         }
 
         totalTx /= GB
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/TrafficStatsManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/TrafficStatsManager.kt
@@ -1,0 +1,173 @@
+package io.homeassistant.companion.android.sensors
+
+import android.content.Context
+import android.net.ConnectivityManager
+import android.net.NetworkCapabilities
+import android.net.TrafficStats
+import android.util.Log
+import io.homeassistant.companion.android.R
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/TrafficStatsManager.kt
```diff
@@ -0,0 +1,173 @@
...
+            return
+
+        var mobileRx = 0f
+        val icon = "mdi:radio-tower"
+
+        try {
+            mobileRx = TrafficStats.getMobileRxBytes().toFloat()
+        } catch (e: Exception) {
+            Log.e(TAG, "Error getting the mobile rx bytes", e)
+        }
```
JBassett, line 94: If we have an issue getting the value we shouldn't continue to update the sensor otherwise you'll get random `0 GB` sent.

app/src/main/java/io/homeassistant/companion/android/sensors/TrafficStatsManager.kt
```diff
@@ -0,0 +1,173 @@
...
+            return
+
+        var mobileTx = 0f
+        val icon = "mdi:radio-tower"
+
+        try {
+            mobileTx = TrafficStats.getMobileTxBytes().toFloat()
+        } catch (e: Exception) {
+            Log.e(TAG, "Error getting the mobile tx bytes", e)
+        }
```
JBassett, line 117: If we have an issue getting the value we shouldn't continue to update the sensor otherwise you'll get random `0 GB` sent.

app/src/main/java/io/homeassistant/companion/android/sensors/TrafficStatsManager.kt
```diff
@@ -0,0 +1,173 @@
...
+            return
+
+        var totalRx = 0f
+        val icon = "mdi:radio-tower"
+
+        try {
+            totalRx = TrafficStats.getTotalRxBytes().toFloat().absoluteValue
+        } catch (e: Exception) {
+            Log.e(TAG, "Error getting the total rx bytes", e)
+        }
```
JBassett, line 139: If we have an issue getting the value we shouldn't continue to update the sensor otherwise you'll get random `0 GB` sent.  Also do we need the `absoluteValue`?

app/src/main/java/io/homeassistant/companion/android/sensors/TrafficStatsManager.kt
```diff
@@ -0,0 +1,173 @@
...
+            return
+
+        var totalTx = 0f
+        val icon = "mdi:radio-tower"
+
+        try {
+            totalTx = TrafficStats.getTotalTxBytes().toFloat().absoluteValue
+        } catch (e: Exception) {
+            Log.e(TAG, "Error getting the total tx bytes", e)
+        }
```
JBassett, line 162: If we have an issue getting the value we shouldn't continue to update the sensor otherwise you'll get random `0 GB` sent.  Also do we need the `absoluteValue`?

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/sensors/TrafficStatsManager.kt
dshokouhi: for some reason the total bytes both rx and tx return a negative value on my pixel lol so I added `absoluteValue` to make sure its a positive number

app/src/main/java/io/homeassistant/companion/android/sensors/TrafficStatsManager.kt
dshokouhi: Same as above, my pixel was showing a negative value so just wanted to make it positive...it only occurs for total bytes lol

JBassett (approved):
Hmm, not sure why total would return negative.  Didn't see anything in the docs about it.  /shrug

Issue #868, opened 2020-09-02
Some issues with geo location sensor


sajikur:
**Home Assistant Android version: 2.3.3-full

**Android version: 10

**Phone model: OnePlus 5T

**Home Assistant version: HassOS 4.12 HA version 0.144.4

**Last working Home Assistant release (if known): all

**Description of problem: geo sensor is stuck in home location. Compass phone is blocked. No geo data send to HA

**Traceback (if applicable):**

```

**Screenshot of problem:**

**Additional information: I've switch off geolcation location, backgroud and zone location and compass is turned on. Sometime need to reboot phone to read this geo informations

anyuta1166:
And I've told many times to either increase MINIMUM_ACCURACY or make it user configurable...

MINIMUM_ACCURACY = 200 is too low to accept cell-tower based geolocation updates.

sajikur:
> And I've told many times to either increase MINIMUM_ACCURACY or make it user configurable...
> 
> MINIMUM_ACCURACY = 200 is too low to accept cell-tower based geolocation updates.
> 
> This way we mainly receive only GPS-based updates (only when a GPS navigation app is opened) and Wi-Fi-based updates (only when ( Wi-Fi is on **or** background Wi-Fi tracking is enabled ) **and** there are Wi-Fi APs known by Google nearby).
> 
> No GPS & no Wi-Fi APs nearby = FAIL.

perfect! Well, how many meters are the right value?

anyuta1166:
> perfect! Well, how many meters are the right value?

It is discussable. I've set it to 5000 for sure. My family often travel to a rural area and accuracy values of 2-3 km are common there.

Anyway, the accuracy threshold is not user configurable now. I've built my own build with manual code changes.

sajikur:
I take advantage of your availability. Can the google maps component on HA 
conflict with geo location on ANDROID APP? Thanx

In data 2 settembre 2020 11:07:33 Anna Tikhomirova 
<notifications@github.com> ha scritto:

dshokouhi:
So this bug report is for a HA integration? The logs suggest its not the app but the Google Maps integration which is not the same thing.  These 2 integrations are completely separate.

sajikur:
I've both. So I thought could be a conflict between two requests on Google api
In data 2 settembre 2020 14:43:37 Daniel Shokouhi 
<notifications@github.com> ha scritto:

anyuta1166:
There is no conflict.
[...]

PR #964, opened 2020-09-20
Settings for location accuracy!

JBassett:
Fixes: https://github.com/home-assistant/android/issues/868

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
@@ -65,7 +65,7 @@ class GeocodeSensorManager : SensorManager {
                 val sensorDao = AppDatabase.getInstance(context).sensorDao()
                 val sensorSettings = sensorDao.getSettings(geocodedLocation.id)
                 val minAccuracy = sensorSettings
-                    .firstOrNull { it.name == "Minimum Accuracy" }?.value?.toIntOrNull()
+                    .firstOrNull { it.name == SETTING_ACCURACY }?.value?.toIntOrNull()
                     ?: DEFAULT_MINIMUM_ACCURACY
                 sensorDao.add(Setting(geocodedLocation.id, SETTING_ACCURACY, minAccuracy.toString(), "number"))
 
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
@@ -8,11 +8,15 @@ import android.os.Build
 import android.util.Log
 import com.google.android.gms.location.LocationServices
 import io.homeassistant.companion.android.R
+import io.homeassistant.companion.android.database.AppDatabase
+import io.homeassistant.companion.android.database.sensor.Setting
 import java.lang.Exception
 
 class GeocodeSensorManager : SensorManager {
 
     companion object {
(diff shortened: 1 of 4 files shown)


app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
```diff
@@ -58,7 +62,14 @@ class GeocodeSensorManager : SensorManager {
                     return@addOnSuccessListener
                 }
 
-                if (location.accuracy <= LocationSensorManager.MINIMUM_ACCURACY)
+                val sensorDao = AppDatabase.getInstance(context).sensorDao()
+                val sensorSettings = sensorDao.getSettings(geocodedLocation.id)
+                val minAccuracy = sensorSettings
+                    .firstOrNull { it.name == "Minimum Accuracy" }?.value?.toIntOrNull()
```
dshokouhi, line 68: Should we replace the string with the constant here?

app/src/main/res/values/strings.xml
```diff
@@ -188,6 +189,7 @@ like to connect to:</string>
   <string name="sensor_description_internal_storage">Information about the total and available storage space internally</string>
   <string name="sensor_description_last_reboot">The date and time of the devices last reboot</string>
   <string name="sensor_description_light_sensor">The current level of illuminance</string>
+  <string name="sensor_description_location_accurate">Allow Home Assistant to send a notification to request an accurate location along with the application periodically requesting an accurate location.</string>
```
dshokouhi, line 192: Should we also describe the settings the user has control over?

app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
JBassett: Yep!

app/src/main/res/values/strings.xml
JBassett: Done

dshokouhi (approved):
Looks good to me.

One question if a user has location permission granted the new sensor `singleAccurateLocation` is automatically enabled?

JBassett:
> Looks good to me.
> 
> One question if a user has location permission granted the new sensor `singleAccurateLocation` is automatically enabled?

Yep!

anyuta1166:
> Yep!

I don't think so. Initially sensors are disabled because we have no permissions. After that we explicitly enable zone and background sensors after getting permissions during initial setup.
So for **new** installations this sensor will be disabled by default.

anyuta1166:
Why do you ever need a new sensor? Looks like another crutches. 

> Allow Home Assistant to send a notification to request an accurate location

This is used not only by Home Assistant notifications, but also triggered by any sensors update request, The description is misleading.

JBassett:
> > Yep!
> 
> I don't think so. Initially sensors are disabled because we have no permissions. After that we explicitly enable zone and background sensors after getting permissions during initial setup.
> So for **new** installations this sensor will be disabled by default.

Feel free to test the change as I did while developing, it does indeed auto enable.


> Why do you ever need a new sensor? Looks like another crutches.

The thought behind the new sensor is for the minimum accuracy.  This allows users to specify what they specifically define as minimum accuracy for a background, zone, or accurate location independently from one another.


> This is used not only by Home Assistant notifications, but also triggered by any sensors update request, The description is misleading.

Feel free to open a PR to clarify the description, however this change only affect the location being requested/sent.  It has absolutely no affect on the other non location sensors.

Issue #876, opened 2020-09-03
Scanning NFC tags always produces "Open Link?" confirmation pop-up (is it a bug?)


donkawechico:
**Home Assistant Android version:**
2.3.0-306-full
**Android version:**
10
**Phone model:**
Pixel 2
**Home Assistant version:**
0.114.4
**Last working Home Assistant release (if known):**
N/A
**Description of problem:**
Not sure if it's a bug or not, but any time I scan an NFC tag, I get a confirmation dialog titled "Open Link?".  I have the option to "Cancel" or "Open link". When I tap "Open link", an intent appears that's clearly part of Home Assistant app, and the event fires successfully.

My expectation (or rather, my "hope") is that if my phone is unlocked, I should be able to scan a tag, and have the event fire without any interaction from me.

I'm guessing this is a security measure on Android that can't be disabled, but I thought I would post the issue in the hopes that it's something that can be changed.

**Traceback (if applicable):**

```

```

**Screenshot of problem:**
![image](https://user-images.githubusercontent.com/1480827/92050126-23f2e700-ed41-11ea-8d1b-d949deae00ae.png)
**Additional information:**
I've verified that the "Open by default" settings show that opening links is set to open in the app, and that "www.home-assistant.io is listed as a supported link (see screenshot)

![image](https://user-images.githubusercontent.com/1480827/92051470-bd21fd80-ed41-11ea-9a82-db90668ae6c7.png)

JBassett:
That open link ui is not from this application.  I also have a Pixel 2 and I'm unable to reproduce.  Do you have any other NFC applications installed?

dshokouhi:
I believe this is an android thing. On my phone Tasker, HA and another NFC app get triggered however for these tags specifically I chose HA to be the default and I no longer get that prompt.  This prompt specifically I have not seen.  Maybe there is another NFC app triggering this pop up?

JayNewstrom:
I also notice this behavior, I'm running Android 10.0 on a Pixel 4a. I also think this is an Android level dialog, since when I click yes, it takes me to the home assistant app.

donkawechico:
> I also have a Pixel 2 and I'm unable to reproduce.

Are you Verizon by any chance? Maybe their flavor of Android adds this as some kind of security measure.

> Do you have any other NFC applications installed?

Not that I'm aware of. I went ahead and uninstalled a few potentials like tasker, AutoTools, AutoApps, AutoVoice, etc. Still get the popup.

[...]

PR #969, opened 2020-09-23
Write application record to NFC tags.

JayNewstrom:
Relates to #876

This fixes an issue where on certain devices a dialog was showing up when scanning the NFC tags.

After this change, scanned NFC tags go directly into the home assistant android companion app.

I also tested on an iOS device to ensure the written NFC tag is correctly read.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/nfc/NFCUtil.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/nfc/NFCUtil.kt
@@ -10,13 +10,15 @@ import android.nfc.NfcAdapter
 import android.nfc.Tag
 import android.nfc.tech.Ndef
 import android.nfc.tech.NdefFormatable
+import io.homeassistant.companion.android.BuildConfig
 import java.io.IOException
 
 object NFCUtil {
     @Throws(Exception::class)
     fun createNFCMessage(url: String, intent: Intent?): Boolean {
         val nfcRecord = NdefRecord.createUri(url)
-        val nfcMessage = NdefMessage(arrayOf(nfcRecord))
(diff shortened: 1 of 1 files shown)


dshokouhi:
I think we may want to default to writing the old format if the size of the card cannot fit the new data, this way we continue to support existing cards that may not be able to write as much data.  We had a few users discover in Discord that some cards can't even take the 70 bytes we write today.  I think with this we increase by 50 bytes correct?

JayNewstrom:
My newly written tag is 63 bytes.

My previously written tag is 26 bytes.

JayNewstrom:
I'm using the debug application id though, which should be `".debug".length` longer. than it will in production. Which will still likely put you over that 50 limit.

JayNewstrom:
I can probably make a PR to the iOS repo to add this same behavior, if you'd like.

JBassett (approved):

JayNewstrom:
@JBassett do you want me to also make a PR to do this on iOS? 

I'm not an iOS developer, so it'll take me a little while to figure it out, but I can try tomorrow if there's not someone else to do it.

JBassett:
I won't speak for @robbiet480 or @zacwest but I'm sure if you make a contribution they will review and give you feedback.  I think it's in everyone's interest to ensure tags work well on all platforms.

robbiet480:
Sounds good to me!

donkawechico:
> We had a few users discover in Discord that some cards can't even take the 70 bytes we write today

Perhaps dumb question, but is "https://www.home-assistant.io/" part of the byte load you're referring to? If so, is a shortened url a way to reduce the footprint?

Issue #959, opened 2020-09-19
Person image upload feature not showing files and media like chrome


dshokouhi:
**Home Assistant Android version:**
Latest master
**Android version:**
11
**Phone model:**
Pixel 4 XL
**Home Assistant version:**
0.115.1
**Last working Home Assistant release (if known):**
N/A new feature
**Description of problem:**

The new person image upload feature does not prompt the user for permissions nor does it show the file dialog to let them select an image.  In chrome this works as expected, in our app we don't get storage permissions yet so we may need to add permissions or also handle showing the prompt?

**Traceback (if applicable):**

```
N/A nothing showed up in logcat while reproducing the issue
```

**Screenshot of problem:**

![image](https://user-images.githubusercontent.com/1634145/93686836-f1016080-fa6d-11ea-9ccc-3e9e6854c718.png)

**Additional information:**

dshokouhi:
Simply adding the permissions to the app was not good enough it seems 🤔

dshokouhi:
Fixed in #986

PR #974, opened 2020-09-23
Support person profile image upload

dshokouhi:
Fixes: #959 

Upon navigating to Configuration > Persons and then selecting the image field we are now presented with 2 options: Camera or Files.

Selecting Camera will open up the default camera, once a user takes a picture and confirms it the image is then handed off to the frontend so the user can crop and save etc...

Selecting Files open up a file manager that will only display images, selecting an image will then hand it off to the frontend so the user can perform the rest of the actions.

If a user did not already grant camera permissions upon selecting the upload icon they will prompted to grant proper permissions.  I had also added reading/writing permissions as those are required for some versions of android.


dshokouhi:
Closing in favor of #986 will need to rethink how to change the file chooser on this page to present the camera and restrict to images only.

PR #975, opened 2020-09-24
Add a link to the change log in app configuration

dshokouhi:
Link to the change log in app configuration.  Link updates dynamically based on the current build.  Local builds will default to the release page while all others like beta and production will link to the tags for the release.

Later on we should consider a pop-up showing the change log after the app was updated and control the pop-up here but for now a fall back to github seems like a good approach :)  The changelog would be generated from the releases page anyways.


app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -125,6 +127,17 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
...
 
+        findPreference<Preference>("changelog")?.let {
+            val link = if (BuildConfig.VERSION_NAME.startsWith("LOCAL"))
+                "https://github.com/home-assistant/android/releases"
+            else "https://github.com/home-assistant/android/releases/tag/${BuildConfig.VERSION_NAME}"
+            it.summary = link
+            val intent = Intent()
+            intent.action = "android.intent.action.VIEW"
+            intent.data = Uri.parse(link)
+            preferenceScreen.findPreference<Preference>("changelog")!!.intent = intent
```
JBassett, line 138: Can't this just be this?
```suggestion
            it.intent = intent
```

dshokouhi: yup, not sure why I thought I had to do that :P

Issue #892, opened 2020-09-07
Feature Request: Caller number


SeraphimSerapis:
**Is your feature request related to a problem? Please describe.**
To implement something like a call log across multiple devices, I'd love to have access to phone numbers I call/that call me via a sensor.

**Describe the solution you'd like**
A sensor that exposes the phone number of incoming/outgoing calls.

**Describe alternatives you've considered, if any**
None

**Additional context**
A similar solution can be achieved for landlines using the [FritzBox Callmonitor](https://www.home-assistant.io/integrations/fritzbox_callmonitor/) integration.

daniele-athome:
Is this feature present on the Google Play version? I can't find it nor the required permission... (using version 2021.1.2-full on Android 9)

dshokouhi:
> Is this feature present on the Google Play version? I can't find it nor the required permission... (using version 2021.1.2-full on Android 9)

This was removed as soon as it was added because our app doesn't meet google criteria for the permission request. We had no choice but to remove it to continue publishing updates. It may make an appearance in the minimal version.

daniele-athome:
I see. Isn't there some approval process to request exceptions you could use? This is indeed a legitimate and reasonable use of the permission IMHO.

dshokouhi:
We did and they did not accept our app for them.

daniele-athome:
Ok thank you. A real shame :-(

PR #978, opened 2020-09-25
Incoming/Outgoing phone number tracking

skynetua:
Fixes: #892 

PR adds calling number track feature. It uses deprecated in api29 constant, but I don't find any other simple suitable solution. I hope google will add some before removed that. It tracks both incoming and outgoing calls.

There is small refactor of SensorReceiver class as well.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
@@ -2,6 +2,7 @@ package io.homeassistant.companion.android.sensors
 
 import android.Manifest
 import android.content.Context
+import android.content.Intent
 import android.os.Build
 import android.telephony.SubscriptionInfo
 import android.telephony.SubscriptionManager
@@ -17,6 +18,13 @@ class PhoneStateSensorManager : SensorManager {
             "sensor",
             R.string.basic_sensor_name_phone,
             R.string.sensor_description_phone_state
+        )
+
+        val callNumber = SensorManager.BasicSensor(
+            "call_number",
+            "sensor",
+            R.string.basic_sensor_name_call_number,
+            R.string.sensor_description_call_number
         )
 
         val sim_1 = SensorManager.BasicSensor(
@@ -41,45 +49,73 @@ class PhoneStateSensorManager : SensorManager {
 
     override val availableSensors: List<SensorManager.BasicSensor>
         get() = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP_MR1)
-            listOf(phoneState, sim_1, sim_2)
-        else listOf(phoneState)
+            listOf(phoneState, callNumber, sim_1, sim_2)
+        else listOf(phoneState, callNumber)
 
     override fun requiredPermissions(sensorId: String): Array<String> {
-        return arrayOf(Manifest.permission.READ_PHONE_STATE)
+        return if (sensorId == callNumber.id) {
+            arrayOf(Manifest.permission.READ_PHONE_STATE, Manifest.permission.READ_CALL_LOG)
+        } else arrayOf(Manifest.permission.READ_PHONE_STATE)
     }
 
     override fun requestSensorUpdate(
         context: Context
     ) {
-        updatePhoneStateSensor(context)
+        checkPhoneState(context)
         updateSimSensor(context, 0)
         updateSimSensor(context, 1)
     }
 
-    private fun updatePhoneStateSensor(context: Context) {
-        if (!isEnabled(context, phoneState.id))
-            return
-        var currentPhoneState = "unavailable"
-        if (checkPermission(context, phoneState.id)) {
-            val telephonyManager =
-                (context.applicationContext.getSystemService(Context.TELEPHONY_SERVICE) as TelephonyManager)
+    private fun checkPhoneState(context: Context) {
+        if (isEnabled(context, phoneState.id) || isEnabled(context, callNumber.id)) {
+            var currentPhoneState = "unknown"
 
-            currentPhoneState = when (telephonyManager.callState) {
-                0 -> "idle"
-                1 -> "ringing"
-                2 -> "offhook"
-                else -> "unknown"
+            if (checkPermission(context, phoneState.id)) {
+                val telephonyManager =
+                    (context.applicationContext.getSystemService(Context.TELEPHONY_SERVICE) as TelephonyManager)
+
+                currentPhoneState = when (telephonyManager.callState) {
+                    0 -> "idle"
+                    1 -> "ringing"
+                    2 -> "offhook"
+                    else -> "unknown"
+                }
             }
-        }
+            if (isEnabled(context, phoneState.id))
+                updatePhoneStateSensor(context, currentPhoneState)
 
+            if (isEnabled(context, callNumber.id) && currentPhoneState in arrayOf("idle", "unknown"))
+                updateCallNumberSensor(context, "none")
+        }
+    }
+
+    private fun updatePhoneStateSensor(context: Context, state: String) {
         var phoneIcon = "mdi:phone"
-        if (currentPhoneState == "ringing" || currentPhoneState == "offhook")
+        if (state == "ringing" || state == "offhook")
             phoneIcon += "-in-talk"
 
         onSensorUpdated(context,
             phoneState,
-            currentPhoneState,
+            state,
             phoneIcon,
+            mapOf()
+        )
+    }
+
+    fun updateCallNumber(context: Context, intent: Intent) {
+        if (checkPermission(context, callNumber.id)) {
+            if (intent.hasExtra(TelephonyManager.EXTRA_INCOMING_NUMBER)) {
+                val number = intent.getStringExtra(TelephonyManager.EXTRA_INCOMING_NUMBER)
+                updateCallNumberSensor(context, number)
+            }
+        }
+    }
+
+    private fun updateCallNumberSensor(context: Context, number: String) {
+        onSensorUpdated(context,
+            callNumber,
+            number,
+            "mdi:numeric",
             mapOf()
         )
     }
@@ -110,7 +146,7 @@ class PhoneStateSensorManager : SensorManager {
                         attrs["mcc"] = info.mccString.toString()
                         attrs["mnc"] = info.mncString.toString()
                         attrs["is opportunistic"] = info.isOpportunistic
-                    if (info.dataRoaming == SubscriptionManager.DATA_ROAMING_ENABLE) attrs["data roaming"] = "enable"
+                        if (info.dataRoaming == SubscriptionManager.DATA_ROAMING_ENABLE) attrs["data roaming"] = "enable"
                         else attrs["data roaming"] = "disable"
                     }
                 }
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
@@ -4,6 +4,7 @@ import android.bluetooth.BluetoothAdapter
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
+import android.telephony.TelephonyManager
 import android.util.Log
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
 import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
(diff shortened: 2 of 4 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
```diff
@@ -82,6 +92,15 @@ class PhoneStateSensorManager : SensorManager {
             phoneIcon,
             mapOf()
         )
+
+        if (isEnabled(context, callNumber.id) && currentPhoneState == "idle") {
+            onSensorUpdated(context,
```
dshokouhi, line 97: This does not trigger an update to get sent to HA, it only updates the DB here. If the state changes next time we send an update it will send to HA.

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
```diff
@@ -124,4 +143,20 @@ class PhoneStateSensorManager : SensorManager {
             )
         }
     }
+    fun updateCallNumber(context: Context, intent: Intent) {
+        if (!isEnabled(context, callNumber.id))
+            return
+        if (checkPermission(context)) {
+            if (intent.hasExtra(TelephonyManager.EXTRA_INCOMING_NUMBER)) {
+                val number = intent.getStringExtra(TelephonyManager.EXTRA_INCOMING_NUMBER)
```
dshokouhi, line 151: What is the default state if we can't retrieve the number? Or if there is no active call?

app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
```diff
@@ -57,41 +58,35 @@ class SensorReceiver : BroadcastReceiver() {
...
+            if (isSensorEnabled(context, sensor!!) != true) {
+                Log.d(TAG, String.format
+                    ("Sensor %s corresponding to received event %s is disabled, skipping sensors update", sensor, intent.action))
+                return
             }
         }
 
+        if (intent.action.equals(TelephonyManager.ACTION_PHONE_STATE_CHANGED) &&
+            isSensorEnabled(context, PhoneStateSensorManager.callNumber.id)!!) {
+            PhoneStateSensorManager().updateCallNumber(context, intent)
```
dshokouhi, line 87: This does not trigger an actual update to get sent to HA, that happens in `updateSensors` down below

app/src/main/res/values/strings.xml
```diff
@@ -182,6 +183,7 @@ like to connect to:</string>
   <string name="sensor_description_battery_state">The current charging state of the battery</string>
   <string name="sensor_description_bluetooth_connection">Information about currently connected bluetooth devices</string>
   <string name="sensor_description_bluetooth_state">Whether or not bluetooth is enabled on the device</string>
+  <string name="sensor_description_call_number">The cell number of the incoming or outgoing call. Phone State sensor has to be enabled for proper work</string>
```
dshokouhi, line 186:
```suggestion
  <string name="sensor_description_call_number">The cell number of the incoming or outgoing call. Phone State sensor has to be enabled for the state to update properly</string>
```

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
```diff
@@ -41,16 +49,18 @@ class PhoneStateSensorManager : SensorManager {
...
     override val availableSensors: List<SensorManager.BasicSensor>
         get() = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP_MR1)
-            listOf(phoneState, sim_1, sim_2)
-        else listOf(phoneState)
+            listOf(phoneState, callNumber, sim_1, sim_2)
+        else listOf(phoneState, callNumber)
 
     override fun requiredPermissions(): Array<String> {
-        return arrayOf(Manifest.permission.READ_PHONE_STATE)
+        return arrayOf(Manifest.permission.READ_PHONE_STATE, Manifest.permission.READ_CALL_LOG)
```
dshokouhi, line 56: Does this mean that we can't use the phone state sensor without granting the `READ_CALL_LOG` permission?

dshokouhi (commented):
Just some questions and comments on the functionality

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
skynetua: Yes for a while. I'm gonig to find simple solution to split required permissions per sensor entity. But it is not in the scope of current PR.

dshokouhi: Doesn't this mean it breaks existing functionality? I'm not sure thats a good thing for those that use it currently.

app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
skynetua: Yep, update will trigger right after that

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
skynetua: There is 2 ways for it:
1. Call is finished. In this case phoneState sensor changes to "idle" and function updatePhoneStateSensor(context) will update this sensor with "none" state.
2. We don't have READ_CALL_LOG permission, sosensor willbe disabled.

The reason of this strange logic is that when we have READ_CALL_LOG permission, system send ACTION_PHONE_STATE_CHANGED broadcast twice. But only one time with EXTRA_INCOMING_NUMBER added.

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
skynetua: Yes, like for any other sensors

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
skynetua: It will just disable sensor. I'll try to fix this in next pr on Monday, so most likely only beta users can be affected. Or we can just merge this pr after it.

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
dshokouhi: So based on this, I wonder if we should consider hiding this sensor until phone state sensor is enabled? It sounds like its more of a requirement right?  Like the user needs to first grant access to even detect if the phone state has changed, then they can get the next call number.

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
dshokouhi: Yea the line numbers were playing tricks on me 🙈 sorry about that

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
skynetua: Actually it will not do anything without PhoneState sensor enabled. We skip sensors update in sensor receiver before this function can get this intent.

dshokouhi: Right but its a better user experience to hide and disable a sensor they can't use if it depends on another sensor being enabled.  It will confuse a user if they can see the sensor and attempt to enable it just for it not to update.

skynetua: Ok, I believe your expirience. I wounder how do you get time and patient to troubleshout all of this complaints on community forum, reddit, github and so on. Nice work, mate! 👍

dshokouhi: Haha the flow is generally the same at my job and working from home makes it easy to be responsive :) Thank you

app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
```diff
@@ -103,6 +98,10 @@ class SensorReceiver : BroadcastReceiver() {
         }
     }
 
+    private fun isSensorEnabled(context: Context, id: String): Boolean? {
+        return AppDatabase.getInstance(context).sensorDao().get(id)?.enabled
+    }
```
JBassett, line 103: If you make this return a `Boolean` not a `Boolean?` you won't need those dangerous `!!` when calling the function.
```suggestion
    private fun isSensorEnabled(context: Context, id: String): Boolean {
        return AppDatabase.getInstance(context).sensorDao().get(id)?.enabled == true
    }
```

app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
```diff
@@ -57,41 +58,35 @@ class SensorReceiver : BroadcastReceiver() {
...
+            val sensor = skippableActions[intent.action]
+            if (isSensorEnabled(context, sensor!!) != true) {
+                Log.d(TAG, String.format
+                    ("Sensor %s corresponding to received event %s is disabled, skipping sensors update", sensor, intent.action))
+                return
             }
         }
 
+        if (intent.action.equals(TelephonyManager.ACTION_PHONE_STATE_CHANGED) &&
+            isSensorEnabled(context, PhoneStateSensorManager.callNumber.id)!!) {
```
JBassett, line 86:
```suggestion
        if (intent.action.equals(TelephonyManager.ACTION_PHONE_STATE_CHANGED) &&
            isSensorEnabled(context, PhoneStateSensorManager.callNumber.id)) {
```

app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
```diff
@@ -57,41 +58,35 @@ class SensorReceiver : BroadcastReceiver() {
...
-            BluetoothAdapter.ACTION_STATE_CHANGED -> {
-                val sensorDao = AppDatabase.getInstance(context).sensorDao()
-                val sensorBtState = sensorDao.get(BluetoothSensorManager.bluetoothState.id)
-                if (sensorBtState?.enabled != true) {
-                    Log.d(TAG, "Bluetooth State Sensor disabled, skipping sensors update")
-                    return
-                }
+        if (skippableActions.containsKey(intent.action)) {
+            val sensor = skippableActions[intent.action]
+            if (isSensorEnabled(context, sensor!!) != true) {
```
JBassett, line 78:
```suggestion
            if (!isSensorEnabled(context, sensor!!)) {
```

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
JBassett: Might need to move permissions to sensors rather than managers.

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
skynetua: Corrected as prposed

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
```diff
@@ -124,4 +162,8 @@ class PhoneStateSensorManager : SensorManager {
             )
         }
     }
+
+    override fun getDependentSensors(): Map<String, String> {
+        return mapOf(callNumber.id to phoneState.id)
+    }
```
JBassett, line 168: Not sure if I like the idea of having sensors depend on one another...

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
```diff
@@ -41,16 +49,20 @@ class PhoneStateSensorManager : SensorManager {
...
+        return if (sensorId == callNumber.id) {
+            arrayOf(Manifest.permission.READ_PHONE_STATE, Manifest.permission.READ_CALL_LOG)
+        } else arrayOf(Manifest.permission.READ_PHONE_STATE)
     }
 
     override fun requestSensorUpdate(
         context: Context
     ) {
+        /*We will update callNumber sensor together with phoneState sensor only when state is idle.
+        At active call callNumber sensor will be already updated by SensorReceiver.onReceive call intent*/
```
JBassett, line 65: Rather than do that why don't we get the phone state in this function then pass that down to `updatePhoneStateSensor` and a new function `resetCallNumber`?  That way we can decouple the two sensors.

app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
```diff
@@ -66,15 +68,39 @@ class SensorDetailFragment(
...
+                    if (parentSensor == null || !parentSensor.enabled) {
+                        Toast.makeText(
+                            context,
+                            format(getString(R.string.sensor_dependency_parent), sensors.getValue(basicSensor.id)),
+                            Toast.LENGTH_SHORT)
+                            .show()
+                        return@setOnPreferenceChangeListener false
+                    }
+                }
+
```
JBassett, line 83: If we decouple the sensors we don't need any of this logic.

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
skynetua: Done

JBassett (approved):

PR #986, opened 2020-09-29
Add file chooser to webview

dshokouhi:
This PR adds basic support to show the file chooser prompt when the webview calls for it (i.e. person image upload).

In this PR we are not restricting the file type nor do we show the camera, this is for basic file upload support.  Person image upload already looks for images and will reject any invalid selected file.

We will need to think about a better way to do things like #974 while still allowing the user to upload any file.  This consideration is important in case any add-on or future feature calls for more than just images.  Maybe we should consider looking at the URL to determine what type of file chooser to show?

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -99,7 +99,7 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
     private lateinit var authenticator: Authenticator
     private lateinit var exoPlayerView: PlayerView
 
-    private var mFilePathCallback: ValueCallback<Array<Uri?>?>? = null
+    private var mFilePathCallback: ValueCallback<Array<Uri>>? = null
     private var isConnected = false
     private var isShowingError = false
     private var alertDialog: AlertDialog? = null
@@ -300,17 +300,14 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
                 }
 
                 override fun onShowFileChooser(
-                    view: WebView?,
-                    uploadMsg: ValueCallback<Array<Uri?>?>,
-                    fileChooserParams: FileChooserParams?
+                    view: WebView,
+                    uploadMsg: ValueCallback<Array<Uri>>,
+                    fileChooserParams: FileChooserParams
                 ): Boolean {
-                    val i = Intent(Intent.ACTION_GET_CONTENT)
-                    i.addCategory(Intent.CATEGORY_OPENABLE)
-                    i.type = "*/*"
                     mFilePathCallback = uploadMsg
-
-                    val chooserIntent = Intent.createChooser(i, resources.getString(R.string.select_file))
-                    startActivityForResult(chooserIntent, FILE_CHOOSER_RESULT_CODE)
+                    val i = fileChooserParams.createIntent()
+                    i.type = "*/*"
+                    startActivityForResult(i, FILE_CHOOSER_RESULT_CODE)
                     return true
                 }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -296,6 +299,21 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
                     }
                 }
 
+                override fun onShowFileChooser(
+                    view: WebView?,
+                    uploadMsg: ValueCallback<Array<Uri?>?>,
+                    fileChooserParams: FileChooserParams?
```
JBassett, line 305:
```suggestion
                    view: WebView,
                    uploadMsg: ValueCallback<Array<Uri>>,
                    fileChooserParams: FileChooserParams
```

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -296,6 +299,21 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
...
+                ): Boolean {
+                    val i = Intent(Intent.ACTION_GET_CONTENT)
+                    i.addCategory(Intent.CATEGORY_OPENABLE)
+                    i.type = "*/*"
+                    mFilePathCallback = uploadMsg
+
+                    val chooserIntent = Intent.createChooser(i, resources.getString(R.string.select_file))
+                    startActivityForResult(chooserIntent, FILE_CHOOSER_RESULT_CODE)
+                    return true
```
JBassett, line 314:
```suggestion
                    mFilePathCallback = uploadMsg
                    startActivityForResult(fileChooserParams.createIntent(), FILE_CHOOSER_RESULT_CODE)
                    return true
```

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -97,6 +99,7 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
     private lateinit var authenticator: Authenticator
     private lateinit var exoPlayerView: PlayerView
 
+    private var mFilePathCallback: ValueCallback<Array<Uri?>?>? = null
```
JBassett, line 102:
```suggestion
    private var mFilePathCallback: ValueCallback<Array<Uri>>? = null
```

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -426,6 +444,13 @@ class WebViewActivity : AppCompatActivity(), io.homeassistant.companion.android.
             webView.evaluateJavascript("externalBus(${JSONObject(message)})") {
                 Log.d(TAG, "NFC Write Complete $it")
             }
+        } else if (requestCode == FILE_CHOOSER_RESULT_CODE) {
+            val results: Uri? = data?.data
+            if (results != null)
+                mFilePathCallback!!.onReceiveValue(arrayOf(results))
+            else
+                mFilePathCallback?.onReceiveValue(arrayOf())
+            mFilePathCallback = null
```
JBassett, line 453:
```suggestion
            mFilePathCallback?.onReceiveValue(WebChromeClient.FileChooserParams.parseResult(resultCode, intent))
            mFilePathCallback = null
```

JBassett (changes requested):
I think we can make it even simpler.  Also while reading the docs I saw this: https://developer.android.com/reference/kotlin/android/webkit/WebChromeClient.FileChooserParams#isCaptureEnabled()  
If we do this does that give us camera for free?

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
dshokouhi: So if we do not provide an override for the intent type then the user is unable to select any image locally asides from a screenshot.  They are able to go to other apps like Photos but on device images are blocked.

image

I think it is ok to override the type as the front end can handle it if we send an incorrect file type.

image

If we use the suggested intent without modification the user only sees screenshots and not recent photos

image

Notice the date/time above compared to showing all files

image

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
dshokouhi: Had to replace `intent` with `data` as data is what we get in this method, otherwise we never send the image.

Issue #940, opened 2020-09-16
Read NFC Tag actually writes


pixeye33:
Home Assistant Android version:
2.4.1-full

Description of problem:
Inside the "NFC Tags" screen, when i click "Read NFC tag" and present a non-hass tag : the app writes the tag !
I would expect the app to write tag only when using the "write nfc tag" button.

aaamoeder:
I also see this behaviour.. I'm on 114.4 and the latest beta build of the android companion app

BoneheadFraggle:
Same here. Pixel 2/Android 11/2.4.1-full

PR #998, opened 2020-10-01
Ensure that we don't write to tags unless explicitly asked to do so.

JBassett:
Fixes: https://github.com/home-assistant/android/issues/940


dshokouhi (approved):
Looks good to me

PR #999, opened 2020-10-01
Remove old migrations since we use room now.

JBassett:


dshokouhi:
Was this from the old breaking change we had where they had to be recreated?

JBassett:
Yep, since we are several versions beyond needing this now is the time to get rid of it so we don't try to use it again.

dshokouhi (approved):
Looks good to me

PR #1000, opened 2020-10-01
Refactor widgets into packages and add previews to interface.

JBassett:
Now we have a preview image to the widget you are selecting!
widgets


dshokouhi:
That preview looks great!

JBassett:
> That preview looks great!

I'm not proud of how long it took me in GIMP to make those look good, haha

dshokouhi (approved):
Looks good to me, tested debug APK and was able to create the widgets

PR #1006, opened 2020-10-04
Allow all sensors to be enabled or disabled at once

dshokouhi:
* Adds a new preference category for managing all sensors
* Tells the user who many total available sensors they have on that device
* Add a toggle to enable or disable all sensors (will remain `off` if all sensors are not enabled)
* Upon enabling requests all required permissions (any that are not granted results in those sensors not being enabled)
* Once enabled a list of all sensors with custom settings are listed to help raise visibility

image

image

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
@@ -5,10 +5,12 @@ import android.os.Handler
 import androidx.preference.Preference
 import androidx.preference.PreferenceCategory
 import androidx.preference.PreferenceFragmentCompat
+import androidx.preference.SwitchPreference
 import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
 import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
 import io.homeassistant.companion.android.database.AppDatabase
+import io.homeassistant.companion.android.database.sensor.Sensor
 import javax.inject.Inject
 
 class SensorsSettingsFragment : PreferenceFragmentCompat() {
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
```diff
@@ -20,28 +22,73 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
...
+                it.summary = getString(R.string.manage_all_sensors_summary, (totalDisabledSensors + totalEnabledSensors))
+            }
+
+            findPreference<SwitchPreference>("enable_disable_sensors")?.let {
+                if (totalDisabledSensors == 0) {
+                    it.title = getString(R.string.disable_all_sensors, totalEnabledSensors)
+                    it.summary = ""
+                    it.isChecked = true
+                } else {
+                    if ((totalDisabledSensors + totalEnabledSensors) == totalDisabledSensors)
```
JBassett, line 67: Isn't this just the following?
```suggestion
                    if (totalEnabledSensors == 0)
```

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
```diff
@@ -20,28 +22,73 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
...
                     findPreference<Preference>(basicSensor.id)?.let {
                         val sensorEntity = sensorDao.get(basicSensor.id)
+                        val sensorSettings = sensorDao.getSettings(basicSensor.id)
+                        if (!sensorSettings.isNullOrEmpty()) {
+                            sensorSettings.forEach { setting ->
+                                if (!setting.name.isNullOrBlank())
+                                    if (getString(basicSensor.name) !in sensorsWithSettings) {
+                                        sensorsWithSettings += getString(basicSensor.name)
+                                        sensorsWithSettings.sort()
+                                    }
```
JBassett, line 39:
```suggestion
                                if (!setting.name.isNullOrBlank() && getString(basicSensor.name) !in sensorsWithSettings){
                                    sensorsWithSettings += getString(basicSensor.name)
                                    sensorsWithSettings.sort()
                                }
```

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
```diff
@@ -20,28 +22,73 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
...
             val sensorDao = AppDatabase.getInstance(requireContext()).sensorDao()
-            SensorReceiver.MANAGERS.forEach { managers ->
-                managers.availableSensors.forEach { basicSensor ->
+            SensorReceiver.MANAGERS.sortedBy { it.name }.filter { it.hasSensor(requireContext()) }.forEach { managers ->
+                managers.availableSensors.sortedBy { it.name }.forEach { basicSensor ->
                     findPreference<Preference>(basicSensor.id)?.let {
                         val sensorEntity = sensorDao.get(basicSensor.id)
+                        val sensorSettings = sensorDao.getSettings(basicSensor.id)
+                        if (!sensorSettings.isNullOrEmpty()) {
+                            sensorSettings.forEach { setting ->
```
JBassett, line 34: It would probably be better to add a DAO query to just get all the sensor settings, or better still one that just returns the list of names of settings since that's all you actually need rather than loop through and request individually.  That way we make one query when the screen loads rather than one query per sensor per update  But I'm still on the fence with how useful it is to show off all the sensors with settings.

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
```diff
@@ -56,6 +103,37 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
...
+                            sensorEntity.lastSentState = ""
+                            sensorDao.update(sensorEntity)
+                        } else {
+                            sensorEntity = Sensor(basicSensor.id, enabledAll, false, "")
+                            sensorDao.add(sensorEntity)
+                        }
+                    }
+                }
+                if (!permArray.isNullOrEmpty())
+                    requestPermissions(permArray, 0)
```
JBassett, line 131: This should probably have have something similar to the detail screen that listens for the permission response and toggles off the preference if not all permissions were granted rather than waiting for the update method to be called again..

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
dshokouhi: Yea I am not certain about it either, I can remove it for now so we only focus on enabling and disabling them all.  Maybe as the settings grow more we can decide a better way to manage them as a whole.

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
dshokouhi: So originally I had taken this from the details screen but due to the for loop I had to take the permission request out and make them after we are done gathering all the required permissions.  I can look at the `onRequestPermissionsResult` to see if it makes the update faster there.

JBassett (approved):

Issue #956, opened 2020-09-19
Open alternate app on notification action


TheEagle13:
**Is your feature request related to a problem? Please describe.**
I have notifications setup to alert me when there is motion detected outside, however the clickActions or actions seem to be limited to just Home assistant actions and I have to manually open my Surveillance Camera App to view the feed from my camera. 

**Describe the solution you'd like**
I'd like for there to be a ClickAction or something in the notification that allows me to tap it and it will automatically launch another App that is already installed on my phone. 

**Describe alternatives you've considered, if any**
Not aware of any alternatives

**Additional context**

drinfernoo:
I was just looking into this as well... Would be a great feature to have, as I'd like to be able to open directly to the Wyze app to see my Wyze Cam's live feed from a notification, rather than opening into HA for the RTSP stream.

Wyze's own app has much better integration, and allows to send audio through the camera, which HA does not, as far as I can tell. This may be possible in the future if someone can create an integration that supports Wyze *Cams*, but until then, this would be my main use case.

PR #1007, opened 2020-10-05
Allow opening of apps via notifications.

JBassett:
Fixes: https://github.com/home-assistant/android/issues/956


JBassett:
This allows you to open an application via notification by setting the URI to `app://{packageName}`.  For example:
```
message: test
data:
  clickAction: app://com.twitter.android
  actions:
    - action: URI
      uri: app://com.google.android.apps.photos
      title: Photos
```

dshokouhi (approved):
Thanks for updating the docs too!!

burzka:
Opening apps from notifications doesn't work as expected on iOS. If I click on something in the actions menu, it works. When I click the notification action itself, it opens the Home Assistant app and not the url app.

PR #1010, opened 2020-10-05
Update readme push_url link after refactor

dshokouhi:
We needed to update the URL as we were getting a 404 after the refactor.


JBassett (approved):
We really should extract that to a properties file.

Issue #912, opened 2020-09-10
Add TTS for Notifications


dshokouhi:
**Is your feature request related to a problem? Please describe.**
Not really a problem, certain notifications I like my phone to speak it out.

**Describe the solution you'd like**
Ideally I would love to send a notification with maybe a attribute like `tts: true` I thought maybe a channel but maybe a setting to benefit all users would be beneficial.

**Describe alternatives you've considered, if any**
Currently using tasker to read my messages to do this.

Might be able to use: https://developer.android.com/reference/android/speech/tts/TextToSpeech

PR #1011, opened 2020-10-06
Allow notifications to do TTS

dshokouhi:
This PR will speak a notification instead of posting it to the device.  For the first release we are not setting the locale so it will default to what the device has as a default.  As we already have commands that use the `message` we are going to use that here and the `title` is where the actual spoken text will reside.  In case of a blank title we have a default, in case of an error we will send a toast.

Example automation:
```
message: TTS
title: I am speaking
```

Fixes: #912 

Docs: https://github.com/home-assistant/companion.home-assistant/pull/351

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -12,6 +12,7 @@ import android.os.Build
 import android.speech.tts.TextToSpeech
 import android.text.Spanned
 import android.util.Log
+import android.widget.Toast
 import androidx.annotation.RequiresApi
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
@@ -60,6 +61,7 @@ class MessagingService : FirebaseMessagingService() {
         const val REQUEST_LOCATION_UPDATE = "request_location_update"
         const val CLEAR_NOTIFICATION = "clear_notification"
         const val REMOVE_CHANNEL = "remove_channel"
+        const val TTS = "TTS"
     }
 
     @Inject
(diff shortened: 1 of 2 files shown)


app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -134,6 +141,29 @@ class MessagingService : FirebaseMessagingService() {
...
+            if (it == TextToSpeech.SUCCESS) {
+                textToSpeech?.speak(tts, TextToSpeech.QUEUE_ADD, null, "")
+                Log.d(TAG, "speaking notification")
+
+                // We need to use some sort of delay to properly shut down the TTS engine
+                // We will be in the background and we don't call onDestroy here
+                // Without this delay nothing will be said and its probably not a good idea to keep it open
+                Thread.sleep(10000)
+                textToSpeech?.stop()
+                textToSpeech?.shutdown()
```
JBassett, line 160: Rather than doing this we should use [a listener](https://developer.android.com/reference/android/speech/tts/TextToSpeech#setOnUtteranceProgressListener(android.speech.tts.UtteranceProgressListener)) to determine when we are done so we can shutdown then.

app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -134,6 +141,29 @@ class MessagingService : FirebaseMessagingService() {
         }
     }
 
+    private fun speakNotification(title: String?) {
+        var textToSpeech: TextToSpeech? = null
+        var tts = title
+        if (tts.isNullOrEmpty())
+            tts = getString(R.string.tts_no_title)
+        textToSpeech = TextToSpeech(applicationContext
```
JBassett, line 149: We can probably make this a val here and remove the var above.
```suggestion
        val textToSpeech = TextToSpeech(applicationContext
```

JBassett (changes requested):

app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
dshokouhi: When I test this change locally the variable below where we call `textToSpeech?` has an error, I assume because its in the same method here?

JBassett (approved):

Issue #418, opened 2020-02-13
Send an event when notification dismissed


wbilger:
As title says, just a feature request.
Would it be possible to send an event when a notificatin is swiped away (dismissed).

robbiet480:
[This would be taken care of by my proposal to unify and upgrade push notifications in Home Assistant Company for Android and iOS](https://github.com/home-assistant/companion.home-assistant/issues/176).

wbilger:
Sounds great, awesome work!

mario-tux:
I understand that the cited proposal (https://github.com/home-assistant/companion.home-assistant/issues/176) would solve this problem too but, until now, it is not yet implemented.

Getting a simple event (like any action already supported) when the notification is dismissed would allow to send the same notification on multiple personal devices (smartphone, tablet, ....) and to dismiss the same notification from all of them when it is swiped from one.

Please consider to include this further event: waiting for a better notification management, it would be very helpful.

PR #1012, opened 2020-10-06
Send an event when a notification is dismissed

dshokouhi:
Fixes: #418 

When a notification is dismissed we will send `mobile_app_notification_dismissed` event along with the `message` as part of the event data.

Example data:

```
{
    "event_type": "mobile_app_notification_dismissed",
    "data": {
        "message": "test",
        "device_id": "DEVICE_ID"
    },
    "origin": "REMOTE",
    "time_fired": "2020-10-06T05:36:12.864583+00:00",
    "context": {
        "id": "ID",
        "parent_id": null,
        "user_id": "USER_ID"
    }
}
```

Docs: https://github.com/home-assistant/companion.home-assistant/pull/352

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/notifications/NotificationDeleteReceiver.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/notifications/NotificationDeleteReceiver.kt
@@ -31,7 +31,8 @@ class NotificationDeleteReceiver : BroadcastReceiver() {
             .build()
             .inject(this)
 
-        val dataList = intent.getStringExtra(EXTRA_DATA)
+        @SuppressWarnings("unchecked")
+        val hashData = intent.getSerializableExtra(EXTRA_DATA) as HashMap<String, *>
         val group = intent.getStringExtra(EXTRA_NOTIFICATION_GROUP)
         val groupId = intent.getIntExtra(EXTRA_NOTIFICATION_GROUP_ID, -1)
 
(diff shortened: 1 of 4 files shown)


dshokouhi:
Will need to create a page in the docs for this, I'll get it done tomorrow :)

app/src/full/java/io/homeassistant/companion/android/notifications/NotificationDeleteReceiver.kt
```diff
@@ -23,5 +41,23 @@ class NotificationDeleteReceiver : BroadcastReceiver() {
         // This maybe the case if timeoutAfter has deleted the notification
         // Then only the empty group is left and needs to be cancelled
         notificationManagerCompat.cancelGroupIfNeeded(group, groupId)
+
+        val data = mutableMapOf(
+            "message" to message
+        )
```
JBassett, line 47: We should probably be consistent with what we send into the events.  We normally send the whole data object we receive with the notification not just the message.

JBassett (changes requested):

app/src/full/java/io/homeassistant/companion/android/notifications/NotificationDeleteReceiver.kt
dshokouhi: I had to do some conversion in order to pass `data` properly, but now we get everything :)

JBassett (approved):

Issue #531, opened 2020-03-27
Display notification rate limit in settings


dshokouhi:
**Is your feature request related to a problem? Please describe.**
Currently it is difficult for users of the Android app to tell how many notifications they have sent and if they are close to hitting the rate limit.  The only way to tell on Android is to enable info level logging for the integration.

**Describe the solution you'd like**
It would be good for the app to try and mimic the iOS app and display the current number of notifications that have been sent.

https://companion.home-assistant.io/docs/notifications/notification-details#rate-limiting

**Describe alternatives you've considered, if any**
User can only tell how many they set by adjusting the logging integration.  See above link for instructions on how they do that currently.

**Additional context**

dshokouhi:
iOS app reference: https://github.com/home-assistant/iOS/blob/master/HomeAssistant/Views/Settings/NotificationRateLimitsAPI.swift

PR #1015, opened 2020-10-07
Add notification rate limit info to settings

dshokouhi:
Fixes: #531 

This PR adds a preference that contains the notification rate limit data.  This preference is only visible if the user is on the `full` flavor and if the call was successful.  One thing to note for some reason the call does not return any data when we first visit settings, its only subsequent visits does the call save.

image

Docs: https://github.com/home-assistant/companion.home-assistant/pull/353

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsPresenterImpl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsPresenterImpl.kt
@@ -189,7 +189,7 @@ class SettingsPresenterImpl @Inject constructor(
         }
     }
 
-    override fun getNotificationRateLimits(): String {
+    override fun getNotificationRateLimits(): HashMap<String, *>? {
         return runBlocking {
             integrationUseCase.getNotificationRateLimits()!!
         }
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsPresenterImpl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsPresenterImpl.kt
@@ -189,9 +189,13 @@ class SettingsPresenterImpl @Inject constructor(
         }
     }
 
-    override fun getNotificationRateLimits(): HashMap<String, *>? {
+    override fun getNotificationRateLimits(): Any {
         return runBlocking {
-            integrationUseCase.getNotificationRateLimits()!!
+            try {
+                integrationUseCase.getNotificationRateLimits()!!
+            } catch (e: Exception) {
+                Log.d(TAG, "Unable to get rate limits")
+            }
         }
     }
 }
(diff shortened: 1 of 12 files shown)


dshokouhi:
Marked as draft while I move logic to `integrationService`

common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationService.kt
```diff
@@ -71,6 +71,12 @@ interface IntegrationService {
         @Body request: IntegrationRequest
     ): Array<Panel>
 
+    @POST
+    suspend fun getRateLimit(
+        @Url url: String,
+        @Body request: Map<String, String>
+    ): HashMap<String, *>
```
JBassett, line 78: Rather than returning a Hash Map here we should return our own data class.

common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
```diff
@@ -301,6 +305,19 @@ class IntegrationRepositoryImpl @Inject constructor(
         return localStorage.getLong(PREF_SESSION_EXPIRE) ?: 0
     }
 
+    override suspend fun getNotificationRateLimits(): HashMap<String, *>? {
```
JBassett, line 308: This should return an actual data class rather than a hashmap.

common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
```diff
@@ -301,6 +305,19 @@ class IntegrationRepositoryImpl @Inject constructor(
...
+        val pushToken = localStorage.getString(PREF_PUSH_TOKEN) ?: ""
+        val requestBody = mapOf("push_token" to pushToken)
```
JBassett, line 310: This request should also be a data class.

app/src/main/java/io/homeassistant/companion/android/settings/SettingsPresenterImpl.kt
```diff
@@ -188,4 +188,10 @@ class SettingsPresenterImpl @Inject constructor(
                     (Integer.parseInt(splitVersion[0]) > 0 || Integer.parseInt(splitVersion[1]) >= 114)
         }
     }
+
+    override fun getNotificationRateLimits(): HashMap<String, *>? {
+        return runBlocking {
+            integrationUseCase.getNotificationRateLimits()!!
```
JBassett, line 194: If this service call fails we will crash the app.  We should add some error handling.

JBassett (changes requested):

JBassett (approved):

Issue #554, opened 2020-04-15
Send notification of others apps as events


MiguelAngelLV:
Some sensors can't be integrate with Home Assistant, so, if Home Assistant companion can read the  notification of android app service and send to Home Assistant, we could read the event.

Example:

Google Nest Hello, a doorbell, can't be integrate with HA.
If HA Companion can read the notification of Nest from my phone and send to HA, I could make the speaker sound when someone press the chime.

Other example: Notify lost calls when I forget the phone in other room.

dshokouhi:
Lets try not to be vague in the feature request, lets be concrete about what you want.  If you want multiple sensors it would be best if you create one request for each sensor.  Not sure if you considered it but a lot of what you want to do can also be done using Tasker.

MiguelAngelLV:
Using notifications you can create "any event" for other apps: Get a message, low battery, alarm, google now alert...

The idea is add a filter (regex, substring, contain text, all from a app...) and send a event.

Example:

from telephone app, when the notification contain mom, send event "mom_call".

xkjq:
This request is exactly what I have been looking for. I'm currently using NotifyMe (https://play.google.com/store/apps/details?id=com.danielpuiu.notifyme&hl=en_GB) as a middle man (phone notification -> NotifyMe emails account -> imap_email_content).

It would allow integration of any phone app (that produces notifications) with home assistant.

Presumable, as suggested above, a generic sensor (much like imap_email_content actually) that can be configured via home assistant would be the best route as then it would be up to the user to integrate individual apps.

MiguelAngelLV:
Instead of sensor, I think that a event is more simple and elegant. The app id is the event_id and the notification text the playload.

After, in Home Assistant, you could catch the event and update a sensor if you want.

gary-reyes:
The android app Ariela provides a notification grabber sensor. It basically reports the latest notification from the device. It requires an additional permission from android 

I use the notification grabber sensor to create my own template sensors. I suppose sending events rather than a sensor would accomplish similar things

Hypfer:
This definitely needs a whitelist to combat the impact on the recorder component as well as overall phone battery life.

Also privacy 👀

JBassett:
> This definitely needs a whitelist to combat the impact on the recorder component as well as overall phone battery life.
> 
> Also privacy 👀

It has a whitelist, scroll to the bottom of the detail screen and you can set one.  When no wait list is specified all are included.

Hypfer:
[...]

PR #1018, opened 2020-10-07
Last Notification Sensor

JBassett:
Fixes: https://github.com/home-assistant/android/issues/554

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
@@ -40,6 +40,9 @@ class NotificationSensorManager : NotificationListenerService(), SensorManager {
     override fun onNotificationPosted(sbn: StatusBarNotification) {
         super.onNotificationPosted(sbn)
 
+        if (!isEnabled(applicationContext, lastNotification.id))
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
```diff
@@ -0,0 +1,75 @@
...
+            .contains(context.packageName)
+    }
+
+    override fun requestSensorUpdate(context: Context) {
+        // Noop
+    }
+
+    override fun onNotificationPosted(sbn: StatusBarNotification) {
+        super.onNotificationPosted(sbn)
+
```
dshokouhi, line 42: So normally we do a return here if the sensor is disabled, is that not needed with this type of sensor since its not part of the usual update list?

dshokouhi (approved):
Looks awesome, just a small comment for clarity :)

app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
JBassett: Done!

dshokouhi (approved):

Issue #561, opened 2020-04-19
Play ringtone on notification


markoheric:
I think it would be nice, if we could have critical notification play ringtone, like when you're receiving a call. If its critical, it should be ringing until confirmed by the phone user, maybe limited by the time.

The great example is also "find my phone" on google, where you can make your phone ring. Well something like that.

I think little ding sound is easily missed for critical notifications.

Regards.

dshokouhi:
This might be solvable when notification channels get added, but having it ring until confirmed may not be possible but you will be able to set a longer ringtone on just those.

beardedgeek:
I was about to log a bug to say high priority notifications don't overide default Android silent, and came across this so it looks like it's not a bug, it just doesn't exist yet.

SImilar functionlity can be seen when using the pushover service. A "1" is high priority, and a "2" is critical. You set a retry timeout, and a total expiry time. 

I was hoping to move away from pushover and use the native app but until I can make sure I'll always recieve high priority notifications, regardless of my phone being on silent or not, I'll have to stick with the laggy Pushover.

dshokouhi:
So it looks like in order to ignore the phone being on silent/vibrate we need to use the alarm stream so it can ring no matter what.

https://stackoverflow.com/questions/15578812/troubles-play-sound-in-silent-mode-on-android

This would be pretty useful actually especially for really important things like the alarm being triggered or smoke in the house etc...

Elyx0:
Doesn't look fresh, 2014 ish

dshokouhi:
So do we want to introduce a new setting to use the alarm stream or do we attempt to piggy back off the critical notification example to match iOS?  Personally I was thinking if we can just use a specific channel (ex: `channel: alarm-stream`) that would be great but then that means anyone on Android 7 and lower won't be able to use it.

@Elyx0 its ok if its old....a lot of stuff on stackoverflow is still relevant as the underlying API's have not changed much.  In this case the ringtone and stream are still there.

membersound:
I would prefer picking a solution that works most reliable.
Eg there might be restrictions on non-stock android phones (like EMUI huawei), which *might* restrict one or the other solution.
Think this is way more important than supporting legacy (android 7) devices.

dshokouhi:
@membersound this would use stock android API's if a non-stock phone decides to go against whats in the android docs there won't be much we can do for it.  More than likely if other apps support it on your device then it should work here as the code would be similar.  We wont really know until this is implemented and out in the beta how well it works on other devices.

PR #1019, opened 2020-10-07
Allow notifications to use the alarm stream

dshokouhi:
Fixes: #561 

The best way to make sure a notification stays as expected while also respecting Android 8+ channels lets use a specific channel to bypass the devices ringer mode.  The notification will make a sound (regardless if the device is on vibrate/silent) and behave much like an alarm, the stream will stop once the user interacts with their device. The default sound will be the default alarm which may be enough for most users.  There is also code to support legacy devices too, we are just reading the channel name and changing the stream accordingly.

Example:

```
message: Time to wake up!
data:
  channel: alarm_stream
```

Docs: https://github.com/home-assistant/companion.home-assistant/pull/354

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -210,12 +210,8 @@ class MessagingService : FirebaseMessagingService() {
         val notificationBuilder = NotificationCompat.Builder(this, channelId)
             .setSmallIcon(R.drawable.ic_stat_ic_notification)
 
-        if (data["channel"] == "alarm_stream") {
-            notificationBuilder.setCategory(Notification.CATEGORY_ALARM)
-            notificationBuilder.setSound(RingtoneManager.getDefaultUri(RingtoneManager.TYPE_ALARM), AudioManager.STREAM_ALARM)
-        } else {
-            notificationBuilder.setSound(RingtoneManager.getDefaultUri(RingtoneManager.TYPE_NOTIFICATION))
-        }
+        handleSound(notificationBuilder, data)
+
         handlePersistent(notificationBuilder, tag, data)
 
         handleLargeIcon(notificationBuilder, data)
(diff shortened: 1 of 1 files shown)


dshokouhi:
Thought of another interesting use case for this.....users can use HA to create their own alarms on the device entirely controlled from HA :)  Don't even need to touch an alarm app, just set it in HA and have the house wake you up as needed.

app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -206,8 +209,13 @@ class MessagingService : FirebaseMessagingService() {
...
         val notificationBuilder = NotificationCompat.Builder(this, channelId)
             .setSmallIcon(R.drawable.ic_stat_ic_notification)
-            .setSound(RingtoneManager.getDefaultUri(RingtoneManager.TYPE_NOTIFICATION))
 
+        if (data["channel"] == "alarm_stream") {
+            notificationBuilder.setCategory(Notification.CATEGORY_ALARM)
+            notificationBuilder.setSound(RingtoneManager.getDefaultUri(RingtoneManager.TYPE_ALARM), AudioManager.STREAM_ALARM)
+        } else {
+            notificationBuilder.setSound(RingtoneManager.getDefaultUri(RingtoneManager.TYPE_NOTIFICATION))
+        }
```
JBassett, line 218: We should break the sound handling into a method like the rest of the items.  I think the category can go in there for now as well.

JBassett (changes requested):

JBassett (approved):

cwalger:
Hello, this looked promising but I am using a Samsung S10e and the latest version of the app from the play store and the notifications do not play a sound when the phone is set to vibrate or to mute.  When my phone is set to sound, the notifications do use the alarm ringtone, so it is partially working.  Thank you

dshokouhi:
@cwalger please do not use merged PRs to discuss issues, there is already an open issue for this that you can read through and provide feedback and logs as requested: https://github.com/home-assistant/android/issues/1186

cwalger:
Sorry, noted, and thank you for the link.  I read through a few threads elsewhere that pointed to this as the solution.

Issue #436, opened 2020-02-16
Add control of device settings/peripherals


robert-alfaro:
Being able to control mobile device's settings and peripherals such as screen brightness/timeout, Wifi, Bluetooth, etc. would be amazing.

Automations act upon the sensor data from mobile to send a control message back to the device to execute some action based on e.g. gps, battery, connected ble device name/mac, etc.

Currently I use an custom Android app named (see [Hass Phone Control](https://community.home-assistant.io/t/hass-phone-control/97182) on community forum) which uses Firebase Cloud Messaging to perform this behavior -- would be awesome to integrate such functionality and flexibility into the official companion app. The below example is just the beginning of what all could be possible.

### Example
If mobile exits Home zone, then an automation triggered on this state change sends message with the following control requests:
  - turn wifi off
  - turn bluetooth on
  - launch app, e.g. Google Play Music

Perhaps that automation (or another) has a condition to check battery level or charge state and could also send:
  - set brightness to 100%
  - set screen timeout to 5 minutes

Upon mobile entering Home or Work zone, another automation could reverse/set things appropriately.

dshokouhi:
Is this app Hass Phone Control open source? Don't see it anywhere. but the forum  I wonder if the author would be willing to contribute here? :)

robert-alfaro:
> Is this app Hass Phone Control open source? Don't see it anywhere but the forum I wonder if the author would be willing to contribute here? :)

He recently opened the source upon my request (personally I'm trying to get included with some Android development.. I'm a firmware/backend guy). The source is Java but obviously can be made kotlin.

https://github.com/Gerben321/homeassistant-phone-control-app

apop880:
I'd love this idea as well. I currently use an app called Automate (similar to Tasker) for some of this (for example, for putting my phone in Vibrate mode when I get to work), but it would be great to unify everything in one place.

thops:
I would love to be able to configure "do not disturb" settings.

dshokouhi:
So this request is a bit of a catch-all and doesn't ask for a specific setting to control. I think to start with we will close this issue by enabling control of turning on/off one setting like DND.  Then we should send in requests for the rest of the commands this way we are not stuck with keeping one issue open while we handle all the possible cases.

I can definitely see a use case for DND as it can be tied to something like your `alarm_control_panel`.  BT, WiFi and screen brightness feel very trivial to me personally.  Launching an app is now possible since #1007 has been merged.

PR #1024, opened 2020-10-08
Allow control of DND via notification

dshokouhi:
Fixes: #436 

This PR introduces a new device command `message: command_dnd` which takes a mandatory parameter set by the user as the `title`.  The only accepted titles are that which is of the DND sensor state: `off` `alarms_only` `priority_only` `total_silence` anything else that is provided results in the notification posting as normal. As DND control is only supported in Android 6+ if a command is sent to an unsupported device that device will also receive the notification as normal.

This type of command requires a special permission which we can only launch an activity.  This activity is launched when the user sends the first command if the permission has not already been granted.  Upon the next command received everything will process as usual.

Examples:
```
message: command_dnd
title: priority_only
```

```
message: command_dnd
title: off
```

There is also some organization done here for future PR's that may introduce new commands so everything can be neatly organized and added.

Docs: Pending (Looking into creating a notification command page for better organization, and also may need to adjust based on review comments)
FCM: Pending (will wait to submit a PR once we finalize this one)

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -66,8 +66,13 @@ class MessagingService : FirebaseMessagingService() {
         const val CLEAR_NOTIFICATION = "clear_notification"
         const val REMOVE_CHANNEL = "remove_channel"
         const val TTS = "TTS"
-        const val DND_ON = "dnd_on"
-        const val DND_OFF = "dnd_off"
+        const val COMMAND_DND = "command_dnd"
+        const val DND_PRIORITY_ONLY = "priority_only"
+        const val DND_ALARMS_ONLY = "alarms_only"
+        const val DND_ALL = "off"
+        const val DND_NONE = "total_silence"
+        val DEVICE_COMMANDS = listOf(COMMAND_DND)
+        val DND_COMMANDS = listOf(DND_ALARMS_ONLY, DND_ALL, DND_NONE, DND_PRIORITY_ONLY)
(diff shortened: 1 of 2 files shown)


app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -111,6 +118,29 @@ class MessagingService : FirebaseMessagingService() {
...
+                                    }
+                                }
+                            } else {
+                                mainScope.launch {
+                                    Log.d(TAG, "Invalid DND command received, posting notification to device")
+                                    sendNotification(it)
+                                }
+                            }
+                        }
+                        else -> Log.d(TAG, "No command received")
```
JBassett, line 141: Shouldn't we send the notification to be consistent here as well?

JBassett (approved):

app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
dshokouhi: It technically shouldn't be possible to hit it due to line 121 because we are specifying all device commands there.  My goal was that each command would be added with a new condition and variable per the list above.  I thought about it but opted not to because we should be specific here, I can add it if you think we should.

JBassett: Ok, I see what you mean.  We can keep it the way it is.

JBassett (approved):

robert-alfaro:
Very cool, thanks!!

Issue #398, opened 2020-02-12
Notification history view


robbiet480:
Would be good to have a view of past notifications, although this might be better accomplished by storing notifications on backend and offering a frontend view so that its cross platform.

iOS has needed this for a long time.

shbatm:
It would be nice, but maybe not practical or feasible, to show the notifications in the persistent_notification tray on the webview in the app.

CodeFinder2:
+1 for this feature from me as well. Regardless of the solution for #372, very (I mean VERY) long messages are still not completely shown (which probably makes sense since it's just a notification 😁). To read them completely, one would probably need such a history view to be able to also view (arbitrarily?) long messages.

Also, IF there is a length limitation, a service call should properly fail for notifications exceeding this limit (to indicate this).

shbatm:
Some Google-fu shows that the hard limit is 1024 characters, but what each device will display looks like it varies -- [example here](https://support.airship.com/hc/en-us/articles/213491643-What-are-the-Maximum-Characters-for-Push-Notifications-).

PR #1054, opened 2020-10-15
Add notification history page to settings

dshokouhi:
Fixes: #398 

This PR adds a notification history view.  We will store the notifications as they come in to a new table and display them to a user.  As this can become a lot for a first PR we will default to showing the last 25 notifications received so the page won't perform poorly.  In future PR's we can add deleting a single entry and clearing the whole table.

As `message` is the only required data field we will default to showing the time the notification was received, the `message` and the full data payload.  Later on we can look at making the rest look better.

image

image

image


dshokouhi:
ok not really sure how to resolve these 😅 i will revert back to what i had originally and wait to see what i am doing wrong 😝

JBassett (approved):

PR #1068, opened 2020-10-16
Only check redirect uri if it starts with AUTH_CALLBACK

balloob:
We should only hijack the navigation to a url if it starts with our auth callback. It is legit to redirect if the auth callback is in the url.

This _should_ (but I have not tested it yet) fix the Android app going back to the server selection screen if it hits Home Assistant while it's still serving the landing page.


balloob:
Tested by Ludeeus. Did not fix the bug but still works, so is a better implementation.

JBassett (approved):

PR #1070, opened 2020-10-16
Breaking Change: Switch call number permission for android 11 changes and google play policy

dshokouhi:
We had a rejection notice from Google so I _think_ this is what they are asking for.

There is a google play policy about the usage of `READ_CALL_LOG` and that our app has to meet certain criteria to support it as a primary function.  This permission is only used for the `call_number` sensor and the old permission supports the minimum API that we support.

https://support.google.com/googleplay/android-developer/answer/9047303#intended
https://support.google.com/googleplay/android-developer/answer/9888170/

Unfortunately in Android 11 to get the phone number has changed too so we need to use a completely different method and permission.  The permission only works in API 26 and above so this will be a breaking change to some users.

https://developer.android.com/about/versions/11/privacy/permissions#phone-numbers

Unless I am misunderstanding something this is what we need to do? Tested this on my Pixel 4 XL and the phone number still gets pulled in.

As we have to consider users in an unsupported API may have this sensor previously enabled I have opted to just hide the sensor from the manage sensor screen and it will skip updates completely so we don't hit any errors.  The good news is that this sensor is not in production yet so the impacted users will not be as much.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -15,7 +15,7 @@
     <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
     <uses-permission android:name="android.permission.NFC" />
     <uses-permission android:name="android.permission.READ_PHONE_STATE" />
-    <uses-permission android:name="android.permission.READ_CALL_LOG" />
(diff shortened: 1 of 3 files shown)


dshokouhi:
Closing as this is just not correct, wrong data 🤦

PR #1072, opened 2020-10-18
Remove call number sensor due to policy changes at Google.

JBassett:


dshokouhi:
should we also remove the strings for the sensor?

sep-don-barrett:
Not everyone installs through the Google Play Store and its increasingly arbitrary policy requirements.  Might I suggest leaving the code in place and not building it for Play Store releases? Some of us like this capability.

dshokouhi:
@sep-don-barrett you can use the new Last Notification sensor to do the exact same thing.  This call number sensor was a beta feature and not released yet.   Last Notification is also a beta feature and if you use the phone app in its allow list you can get the same data.

JBassett:
It could be moved to the minimal build of the application if desired.

PR #1075, opened 2020-10-18
Add privacy policy

JBassett:
More Play Store policy changes.
image
image


balloob:
Maybe also mention in the enable location tracking description that the data is only sent directly to your instance without involvement of any third parties.

JBassett:
@balloob Added the info to the description.

balloob:
Maybe an idea to make the privacy policy not a button, but a link in the description? Now it is a little confusing what the button to go next is.

JBassett:
@balloob good idea, this looks much better:
image

Issue #971, opened 2020-09-23
App Configuration panel ignores user defined language


spacegaier:
**Home Assistant Android version:** 2.4.1.-full

**Android version:** 8.0.0

**Phone model:** SM-G930F

**Home Assistant version:** Latest dev (0.116.0)

**Last working Home Assistant release (if known):**

**Description of problem:**
The user logged into the app has its language set to English, which is generally adhered to in the app, with the exception of the "App Configuration" panel where everything is still in German (language of the mobile phone itself).

Possibly related: The exposed sensors from the phone also use a German name in HA.

**Traceback (if applicable):**

```

```

**Screenshot of problem:**

**Additional information:**

dshokouhi:
You also need to check your user profile in HA, you can say a different language in there as well

dshokouhi:
You need to make sure the phone and HA matches the language you expect to see. HA frontend only respects the user profile, the app respects your phone language

spacegaier:
@dshokouhi So I understand that correctly: You are saying that my use case is currently not supported (having a German phone but wanting to run all of HA in English)?

Is there no way for the app to respect the HA user profile to be consistent and not display a language mixture?

dshokouhi:
To my understanding apps on your phone take on the system provided language.  Our app is a unique case though because we have a webview that has its own language independent from your phone.  We provide translation strings in our app and your phone will decide when to show what based on its selected language.  I don't believe the app can override that logic while also pulling in the current selected user profile language.  We don't get notified when that value changes from your profile so even if we did it, it would be out of sync any time it gets changed. If we are able to offer the app to override the system language then this would be really confusing to users.  They would then need to differentiate between their device language, the app language and also the HA frontend language.

spacegaier:
So on a technical level, is this understanding then correct: 

The "App Configuration" panel icon that is shown in the HA sidebar in the app is not actually part of the regular HA frontend (so not injected by the app into it) but instead is basically a link/trigger that tells the actual Android app to show an Android settings page with the app/sensor settings and that page then has to follow the device language since not actually part of the regular HA frontend?

That at least seems to align with my observations now that I paid a closer look at the app.

dshokouhi:
That is correct, the Android app controls what appears in the settings. The front end detects that an external app is being used and then shows the App Configuration menu that we can use.

https://developers.home-assistant.io/docs/frontend/external-bus/#show-config-screen-config_screenshow
[...]

PR #1083, opened 2020-10-20
Add feature to override device language

skynetua:
Fixes: #971 

PR adds setting to override device default language.
Also:
New icon for NFC tags
Update appcompat to 1.2.0

SmartSelect_20201020-163945_Home Assistant
SmartSelect_20201020-164013_Home Assistant


app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesProvider.kt
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Locale
+
+class LanguagesProvider {
+
+    fun getSupportedLanguages(context: Context): Map<String, String> {
+        val listAppLocales = mutableMapOf<String, String>()
+
+        val resources = context.resources
+        val configuration = resources.configuration
+        val listLocales = resources.assets.locales
```
skynetua, line 20: Have no ideas why AssetManager ignores regional locales there, like zh-rCN, zh-rTW, fy-rNL. Any suggestion?

dshokouhi:
Hmm did the actions get stuck here? 

Also I think we probably need to add this to the onboarding otherwise users will still have the same issue as reported in #971 since they will have 5 entities to rename before they can change the language.

I am not certain on the asset manager issue, maybe it has to do with the languages we support or the testing device?

app/src/main/res/xml/preferences.xml
```diff
@@ -31,12 +31,17 @@
...
     <PreferenceCategory
-        android:title="@string/sensors">
+        android:title="@string/entities">
         <Preference
             android:key="sensors"
             android:icon="@drawable/leak"
             android:title="@string/sensor_title"
             android:summary="@string/sensor_summary"/>
+        <Preference
+            android:key="nfc_tags"
```
dshokouhi, line 41: NFC tags are not an entity, they only fire an event to HA.

dshokouhi:
I see the screenshot shows "Not Set" for the language, I think by default it should be the system language and if not provided set to English but never Not Set.  It should always have some value.  Also is the list sorted by display name or language code :)

skynetua:
> I see the screenshot shows "Not Set" for the language, I think by default it should be the system language and if not provided set to English but never Not Set. It should always have some value. Also is the list sorted by display name or language code :)

It is. Just a bit outdated screenshot)

app/src/main/res/xml/preferences.xml
skynetua: What pref category name you'd recommend?

dshokouhi: Umm I think it made sense to keep it under Other?

skynetua:
> Hmm did the actions get stuck here?
> 
> Also I think we probably need to add this to the onboarding otherwise users will still have the same issue as reported in #971 since they will have 5 entities to rename before they can change the language.
> 
> I am not certain on the asset manager issue, maybe it has to do with the languages we support or the testing device?

It's waiting for merge conflict resolved.

As for AssetManager - we support 25 languages so far, and it ignores 4 of them.

app/src/main/res/xml/preferences.xml
skynetua: Ok, I'll revert this

JBassett (changes requested):
You are also going to need to ensure all languages are included in the bundled apks that get installed:
```
android {

    ...
    bundle {
        language {
            // Specifies that the app bundle should not support
            // configuration APKs for language resources. These
            // resources are instead packaged with each base and
            // dynamic feature APK.
            enableSplit = false
        }
    }
}
```

app/src/main/res/xml/preferences.xml
```diff
@@ -59,11 +64,12 @@
             android:icon="@drawable/ic_fullscreen"
             android:title="@string/fullscreen"
             android:summary="@string/fullscreen_def"/>
-        <Preference
-            android:key="nfc_tags"
-            android:icon="@drawable/ic_plus"
-            android:title="@string/nfc_title_settings"
-            android:summary="@string/nfc_summary" />
+        <ListPreference
```
dshokouhi, line 67: I wonder if we should provide a description about what actually gets changed when we modify the app language.  Users probably want to know that it not only impacts the UI but also the entity names that we send out.

app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesProvider.kt
skynetua: My bad, it doesn't ignore. Moreover it return region locales we don't support dirrectly, like fr-CA(because we support fr) or es-US and I see no way to exclude them with current solution.

app/src/main/res/xml/preferences.xml
skynetua: Yep, it is probably more useful than current setting value.

app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesProvider.kt
```diff
@@ -0,0 +1,57 @@
...
+
+        listAppLocales[resources.getString(R.string.lang_option_label_default)] = resources.getString(R.string.lang_option_value_default)
+        val defString = getStringResource(context, "")
+
+        listLocales.forEach {
+            if (getStringResource(context, it) != defString || it == Locale.ENGLISH.language) {
+                listAppLocales[StringUtils.capitalize(getDefaultLocale(configuration).displayLanguage).toString()] = it
```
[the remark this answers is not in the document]
skynetua: It is, corrected this.

common/src/main/java/io/homeassistant/companion/android/common/dagger/Graph.kt
```diff
@@ -44,7 +44,7 @@ class Graph(
                     ),
                     LocalStorageImpl(
                         application.getSharedPreferences(
-                            "themes",
+                            "prefs",
```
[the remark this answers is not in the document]
skynetua: Yes, should I change it back? I don't think it's a big deal :)

skynetua:
> You are also going to need to ensure all languages are included in the bundled apks that get installed:
> 
> ```
> android {
> 
>     ...
>     bundle {
>         language {
>             // Specifies that the app bundle should not support
>             // configuration APKs for language resources. These
>             // resources are instead packaged with each base and
>             // dynamic feature APK.
>             enableSplit = false
>         }
>     }
> }
> ```

I'm not sure what to do with that. I beleeve AssetsManager will handle this sucessfully.

app/src/main/res/values/strings.xml
```diff
@@ -122,6 +122,10 @@ Home Assistant instance</string>
   <string name="label_icon">Icon:</string>
   <string name="label_label">Label:</string>
   <string name="label_service">Service:</string>
+  <string name="lang_option_label_default">Device Language</string>
+  <string name="lang_option_value_default">default</string>
+  <string name="lang_summary_settings">Changing default language may have influence on device entities naming</string>
```
dshokouhi, line 127: Should we also mention that this setting is independent of the User Profile language option?

app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+            val listLocales = langManager.getLocales()!!.split(',')
+            listLocales.forEach {
+                if (it.isNotEmpty()) {
+                    val name = makeLocale(it).displayLanguage.capitalize()
+                    listAppLocales["$name ($it)"] = it
+                }
+            }
+        }
+
+        listAppLocales[resources.getString(R.string.lang_option_label_default)] = resources.getString(R.string.lang_option_value_default)
```
dshokouhi, line 43: We should probably make the default option be the first option the user see's currently it is the last option.

app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+        val listAppLocales = mutableMapOf<String, String>()
+        val langManager = LanguagesManagerProvider().getManager(context)
+        val resources = context.resources
+
+        if (langManager.getAppVersion() != BuildConfig.VERSION_NAME || langManager.getLocales().isNullOrEmpty()) {
+            val listLocales = resources.assets.locales
+            val defString = getStringResource(context, "")
+            var supportedLocales = ""
+
+            listLocales.forEach {
```
dshokouhi, line 24: Would it make more sense to sort this by the display name? Not sure how it works in other languages but its odd to have it sorted by the language code to me.

app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesProvider.kt
skynetua: Agree, correct it

app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesProvider.kt
skynetua: Probably you are right

app/src/main/res/values/strings.xml
skynetua: I'm not sure. After this changes we will have the same lang for all activities, including frontend language. So basicaly this setting override profile language as well.

dshokouhi: When I say frontend I refer to the HA frontend that is visible in our Webview activity.  The webview is unaware of the devices language and is also unaware of the apps language.  We don't send this data to the HA frontend so it will never be aware of this change.  Load up the APK and test it for yourself.  You will see the HA frontend when you back out of settings is not updated.

skynetua: Yep, I meant it as well. I'm not sure how, but it follows the selected app language. I just don't know a good way to recreate webview activity when we back out of settings. Just reopen the application and you will see the new langg applied to frontend.

dshokouhi: Ok I see that now, having to restart the app isn't immediately apparent since that is what needs to be happen.  I wonder if we should add an alert telling the user we need to restart once they change the language and perform the restart if they accept.  Users may get surprised if they back out and find HA in one language and then suddenly it changes when the app got closed for some reason 😃

skynetua: We do not need to restart the whole app but only webview activity. Like the settings activity is recreates.

@JBassett what is the prefered way to notify webview activity about setting changed? I'm not sure about StartActivityForResult callback or listening for shared preference cahange?

dshokouhi: I think changing the theme in settings does something similar right?

skynetua: Not exactly. AppCompatDelegate is doing all work for themes applying.

common/src/main/java/io/homeassistant/companion/android/common/dagger/Graph.kt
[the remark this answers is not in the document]
JBassett: People get very annoyed when we have breaking changes.  You can refactor the name and stuff, just keep the name of shared preference to keep backwards compatibility.

app/src/main/res/values/strings.xml
JBassett: In the `onResume` maybe store the current locale then check on every `onResume` to see if it changed, if it did call `recreate()`?

app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesManagerProvider.kt
```diff
@@ -0,0 +1,21 @@
...
+    lateinit var lm: LanguagesManager
+
+    fun getManager(context: Context): LanguagesManager {
+        DaggerSettingsComponent.builder()
+            .appComponent((context.applicationContext as GraphComponentAccessor).appComponent)
+            .build()
+            .inject(this)
+
+        return lm
+    }
```
JBassett, line 20: Why does this exist?

app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+import android.content.Context
+import android.content.res.Resources
+import android.os.Build
+import android.os.LocaleList
+import android.util.DisplayMetrics
+import io.homeassistant.companion.android.BuildConfig
+import io.homeassistant.companion.android.R
+import java.util.Locale
+
+class LanguagesProvider {
```
JBassett, line 12: Why note have this be injectable?

app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    fun getSupportedLanguages(context: Context): Map<String, String> {
+        val listAppLocales = mutableMapOf<String, String>()
+        val langManager = LanguagesManagerProvider().getManager(context)
```
JBassett, line 16: Then have this injected?

app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -169,6 +170,12 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
             it.summary = BuildConfig.VERSION_NAME
         }
 
+        findPreference<ListPreference>("languages")?.let {
+            val languages = LanguagesProvider().getSupportedLanguages(requireContext())
```
JBassett, line 174: This provider should be injected I think.

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesProvider.kt
skynetua: Idk) Made it injectable

app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesManagerProvider.kt
skynetua: I don't remember what forced me to use it on early stage. Removed

app/src/main/res/values/strings.xml
skynetua: Yep, works ok.

JBassett (approved):

Issue #987, opened 2020-09-29
Editable widgets


uphillbattle:
**Is your feature request related to a problem? Please describe.**
Widgets are not editable.

**Describe the solution you'd like**
It would be helpful if widgets were editable. Currently, if a widget is set up and you would like to tweak something, you have to delete the widget and create it again.

**Describe alternatives you've considered, if any**
N/A

**Additional context**
N/A

PR #1084, opened 2020-10-21
Allow editing of the entity state widget in settings

dshokouhi:
Fixes: #987 

This PR allows for the entity state widget to be edited.  As we store all widget data in the database we will display all the entity state widgets in a list to the user.  Selecting a widget will take the user to the configure activity with all fields pre-populated and entity information already loaded (given HA is up and running).  Updating the widget on this screen will immediately reflect on the home screen.

One thing I am not sure of is as we add more widget types do we keep them separate in their own fragment or should we just keep them in a category.  I don't know how crazy people get with creating widgets so leaving it up for discussion :) Personally I don't have a lot of widgets so keeping them all in one page separated by category is good enough.

image

image

image

### DIFF
--- a/app/src/main/res/xml/preferences.xml
+++ b/app/src/main/res/xml/preferences.xml
@@ -107,7 +107,7 @@
         android:key="widgets"
         android:title="Widgets">
         <Preference
-            android:key="manage_entity_state_widgets"
+            android:key="manage_widgets"
             android:icon="@drawable/ic_widget"
             android:title="@string/manage_widgets"
             android:summary="@string/manage_widgets_summary" />
--- a/app/src/main/res/xml/preferences.xml
+++ b/app/src/main/res/xml/preferences.xml
@@ -104,6 +104,15 @@
             android:summary="@string/rate_limit_summary"/>
     </PreferenceCategory>
     <PreferenceCategory
+        android:key="widgets"
+        android:title="Widgets">
+        <Preference
+            android:key="manage_entity_state_widgets"
+            android:icon="@drawable/ic_widget"
+            android:title="@string/manage_entity_state_widgets"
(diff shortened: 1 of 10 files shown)


app/src/main/res/xml/preferences.xml
```diff
@@ -103,6 +103,15 @@
...
             android:summary="@string/rate_limit_summary"/>
     </PreferenceCategory>
+    <PreferenceCategory
+        android:key="widgets"
+        android:title="Widgets">
+        <Preference
+            android:key="manage_entity_state_widgets"
+            android:icon="@drawable/ic_widget"
+            android:title="@string/manage_entity_state_widgets"
+            android:summary="@string/manage_entity_state_widgets_summary" />
```
JBassett, line 113: Do we really need one for each?  Maybe just separate the widgets by type with a PreferenceCategory in the next screen rather than having a bunch on the main settings screen?

dshokouhi: ok will update :) I wasn't sure which route to go

JBassett (changes requested):

app/src/main/res/xml/preferences.xml
```diff
@@ -103,6 +103,15 @@
             android:icon="@drawable/ic_notifications"
             android:summary="@string/rate_limit_summary"/>
     </PreferenceCategory>
+    <PreferenceCategory
+        android:key="widgets"
+        android:title="Widgets">
+        <Preference
+            android:key="manage_entity_state_widgets"
```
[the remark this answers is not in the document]
dshokouhi: yea totally missed this one :)

JBassett (approved):

PR #1088, opened 2020-10-21
Ensure that we default to expected values to meet requirements for android power menu

dshokouhi:
Should fix the following sentry error:

```
java.lang.IllegalArgumentException: stepValue=0,000000 <= 0
    at android.service.controls.templates.RangeTemplate.validate(RangeTemplate.java:184)
    at android.service.controls.templates.RangeTemplate.<init>(RangeTemplate.java:83)
    at io.homeassistant.companion.android.controls.DefaultSliderControl$Companion.createControl(DefaultSliderControl.kt:37)
    at io.homeassistant.companion.android.controls.HaControlsProviderService$createPublisherForAllAvailable$1$1.invokeSuspend(HaControlsProviderService.kt:85)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:738)
```

According to the `input_number` code we don't need to adjust much as they already take care of most requirements:

https://developer.android.com/reference/android/service/controls/templates/RangeTemplate#RangeTemplate(java.lang.String,%20float,%20float,%20float,%20float,%20java.lang.CharSequence)

https://github.com/home-assistant/core/blob/dev/homeassistant/components/input_number/__init__.py#L53

https://github.com/home-assistant/core/blob/dev/homeassistant/components/input_number/__init__.py#L68

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/controls/DefaultSliderControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/DefaultSliderControl.kt
@@ -37,9 +37,13 @@ class DefaultSliderControl {
                 RangeTemplate(
                     entity.entityId,
                     (entity.attributes["min"] as? Number)?.toFloat() ?: 0f,
-                    (entity.attributes["max"] as? Number)?.toFloat() ?: 0f,
+                    (entity.attributes["max"] as? Number)?.toFloat() ?: 1f,
                     entity.state.toFloatOrNull() ?: 0f,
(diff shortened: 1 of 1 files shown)


dshokouhi:
It seems the issue here is that when an entity was restored to `unavailable` all attributes come in as `null` since they do not exist and that is why the defaults matter for input number as we set them based on current attributes

JBassett (approved):

Issue #1094, opened 2020-10-23
Battery Icon when 100% and not charging


SeLLeRoNe:
**Home Assistant Android version:**
beta-467-1fdbe71-full
**Android version:**
10
**Phone model:**
OnePlus Nord and Mi 9T Pro
**Home Assistant version:**
0.117.0b2
**Last working Home Assistant release (if known):**

**Description of problem:**
When a phone is charged to 100% the icon in use is: icon: mdi:battery-charging
When the phone is unplugged but still at 100% the icon is not changed on the battery_level sensor even while the battery_state sensor report "dischargin" with the "battery-minus" icon
**Traceback (if applicable):**

```

```

**Screenshot of problem:**
![image](https://user-images.githubusercontent.com/9599441/96984290-b5502f80-1520-11eb-9236-b8405d7bc2b5.png)
![image](https://user-images.githubusercontent.com/9599441/96984229-b1bca880-1520-11eb-9944-8ed5b3efbfb2.png)

This is just as reference to show that when Charging in my screenshot I have a message on the secondary info:
![image](https://user-images.githubusercontent.com/9599441/96984643-ce58e080-1520-11eb-81ee-0f33bb718d93.png)

**Additional information:**

dshokouhi:
We no longer send updates to HA when the state hasn't changed so this is to be expected. I guess we can remove the battery charging part completely for this sensor. But we won't send an update just to change the icon.

PR #1095, opened 2020-10-23
Battery Icon when 100% and not charging

mik629:
* fix `IsCharging` state determination

I assume this at least should fix one-time battery status determination, not sure about changing statuses after plugging/unplugging. Sorry, I couldn't find any sandbox/test environment where I could reproduce this bug and test its fix. So, this PR is more of a hypothesis.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/BatterySensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/BatterySensorManager.kt
@@ -205,8 +205,7 @@ class BatterySensorManager : SensorManager {
     private fun getIsCharging(intent: Intent): Boolean {
         val status: Int = intent.getIntExtra(BatteryManager.EXTRA_STATUS, -1)
 
-        return status == BatteryManager.BATTERY_STATUS_CHARGING ||
-                status == BatteryManager.BATTERY_STATUS_FULL
+        return status == BatteryManager.BATTERY_STATUS_CHARGING
     }
 
     private fun getChargerType(intent: Intent): String {


homeassistant:
Hi @mik629, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;android#1095).

Thanks!

dshokouhi:
If you are trying to fix #1094  then you need to fix it in `updateBatteryLevel` and not for `getIsCharging`

app/src/main/java/io/homeassistant/companion/android/sensors/BatterySensorManager.kt
```diff
@@ -205,8 +205,7 @@ class BatterySensorManager : SensorManager {
     private fun getIsCharging(intent: Intent): Boolean {
         val status: Int = intent.getIntExtra(BatteryManager.EXTRA_STATUS, -1)
 
-        return status == BatteryManager.BATTERY_STATUS_CHARGING ||
-                status == BatteryManager.BATTERY_STATUS_FULL
+        return status == BatteryManager.BATTERY_STATUS_CHARGING
```
dshokouhi, line 208: Without the `FULL` check then when the battery is full `isCharging` will return `false` but that is not desired here as the device is still plugged in

dshokouhi:
> Sorry, I couldn't find any sandbox/test environment where I could reproduce this bug and test its fix. So, this PR is more of a hypothesis.

You can use the android emulator to simulate the different states the battery can have

app/src/main/java/io/homeassistant/companion/android/sensors/BatterySensorManager.kt
JBassett: Agreed.

JBassett (changes requested):

PR #1098, opened 2020-10-23
Add automations and scripts to power menu controls

vigonotion:
With this it's possible to start scripts via the new power menu controls and enable/disable automations.

I'm using the SceneControl for scripts because it's exactly what i need, but it might be an idea to rename that class to something like "RoutineControl"

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
@@ -53,6 +53,7 @@ class HaControlsProviderService : ControlsProviderService() {
     private var updateSubscriber: Flow.Subscriber<in Control>? = null
 
     private val domainToHaControl = mapOf(
+        "automation" to DefaultSwitchControl,
         "camera" to null,
         "climate" to ClimateControl,
         "cover" to CoverControl,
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
```diff
@@ -61,6 +62,7 @@ class HaControlsProviderService : ControlsProviderService() {
         "media_player" to null,
         "remote" to null,
         "scene" to SceneControl,
+        "script" to SceneControl,
```
dshokouhi, line 65: If a script has a `delay` then technically I think it can be stopped by turning it off.  Unless that has changed recently?

vigonotion: Oh I didn't know that! Although I think it would be kinda unintuitive in most cases if the scripts state changes from Off to On and instantly back to Off

dshokouhi: Looks like the frontend ditched showing a toggle when it needs to and just shows `execute` now so I think it should be ok :)

JBassett (approved):

PR #1100, opened 2020-10-24
Locks and better lights in power menu

vigonotion:
- Add locks to power menu

- Only show range slider for lights when lights actually support brightness (via supported_features)

photo_2020-10-24_16-32-54

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/controls/LightControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/LightControl.kt
@@ -9,8 +9,10 @@ import android.service.controls.DeviceTypes
 import android.service.controls.actions.BooleanAction
 import android.service.controls.actions.ControlAction
 import android.service.controls.actions.FloatAction
+import android.service.controls.templates.ControlButton
 import android.service.controls.templates.RangeTemplate
 import android.service.controls.templates.ToggleRangeTemplate
+import android.service.controls.templates.ToggleTemplate
 import androidx.annotation.RequiresApi
 import io.homeassistant.companion.android.R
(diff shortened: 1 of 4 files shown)


dshokouhi (approved):
Looks good to me

app/src/main/java/io/homeassistant/companion/android/controls/LightControl.kt
```diff
@@ -40,22 +42,31 @@ class LightControl {
...
-                        100f,
-                        (entity.attributes["brightness"] as? Number)
-                            ?.toFloat()
-                            ?.div(255f)
-                            ?.times(100) ?: 0f,
-                        1f,
-                        "%.0f%%"
-                    )
-                )
+                    if ((entity.attributes["supported_features"] as Int) and 1 == 1)
```
JBassett, line 45: This some debugging logic I'm guessing?

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/controls/LightControl.kt
vigonotion: That's a bitwise operation. (Supported features AND 1) should be equal to one, it just looks a bit strange

JBassett: Can we add a constant so we know 1 the first bit is the brightness support?

vigonotion: yes good idea, I added it now

JBassett (approved):

Issue #97, opened 2019-12-04
Support encrypted communication with HA


robbiet480:
Just for that extra bit of security, we should implement [encrypted communication with HA](https://developers.home-assistant.io/docs/en/app_integration_sending_data.html#implementing-encryption) over the webhook like iOS supports.

robbiet480:
Some notes on this:

- We need to use an updated and peer reviewed library to do this. @JBassett says he looked at https://github.com/terl/lazysodium-android, one option I suggested. The other option is https://github.com/joshjdevl/libsodium-jni. Sadly, both are Java. The only Kotlin option hasn't been updated in 2+ years.
- Initial support has been added to HA Core to enable encryption on existing registrations (at https://github.com/home-assistant/home-assistant/pull/31743). Call webhook action `enable_encryption` on HA Core 0.106+, receive a JSON body with `secret` as the response. HA version can be derived via `get_config` webhook action. The logic in app should look like this (psuedocode):
    ```
    if app_has_no_encryption_key and ha_version > 0.106:
        attempt_to_enable_encryption()
    ```
- If enabling encryption fails due to older HA version, then we should either keep doing that check every so often (whenever app starts?) or have a button in settings to allow users to manually enable encryption.

balloob:
Do we know the HA version that we talk to ? Seems like something we should know in the app to be able to enable/disable functionality.

robbiet480:
@balloob We can get that info from the `get_config` webhook action.

JBassett:
Initial work https://github.com/home-assistant/home-assistant-android/tree/feature/webhookEncryption

I am running into issues with HA decrypting my requests...  If anyone wants to take a crack at it please do!

DirkBaumeister:
So, I just discovered that the android app in 2026 still does not support encryption of (cloud) webhooks.
Is this feature somewhere on the roadmap?

I think it would be a great addition to have the same security as the iOS App does here.

TimoPtr:
> So, I just discovered that the android app in 2026 still does not support encryption of (cloud) webhooks. Is this feature somewhere on the roadmap?
> 
> I think it would be a great addition to have the same security as the iOS App does here.

The encryption through Cloud is happening using HTTPS, the encryption you are mentioning from iOS is relevant when using HTTP.

DirkBaumeister:
You are mixing two things up here :)

[...]

PR #1126, opened 2020-10-30
Feature/webhook encryption - Test

ParamonPlay2205:
This seemly works fine on my device used for testing.


common/build.gradle
```diff
@@ -22,4 +22,7 @@ dependencies {
     kapt "com.google.dagger:dagger-compiler:${daggerVersion}"
 
     implementation "com.squareup.retrofit2:retrofit:$retrofit2Version"
+
+    implementation 'com.goterl.lazycode:lazysodium-android:4.1.0@aar'
```
JBassett, line 26: Looks like `4.2.0` is the latest?

JBassett (commented):
Nice start!

dshokouhi:
Is this for #97? We should probably check that every feature that makes a call to HA still works as expected. Not sure if we need to test against different network setups.

dshokouhi:
I just noticed that Travis CI is added to the checks here.  Was this PR based off an older branch?

balloob:
This PR seems to have gone stale. Closing it.

Issue #1020, opened 2020-10-07
Allow sending Intent via notifications


scorpio862:
**Is your feature request related to a problem? Please describe.**
This is not related to a problem. Recently a new feature has been added which allowes to open android apps by notifications and this is great. I just think it would be even better if this app was able to send an Intent to control other apps, not just open them. 
For example if my phone disconnects from my home wifi, Home Assistant server would send a notification to HA companion app and this would send an intent to WireGuard app to connect to my vpn server. I think this feature would have a lot of use cases.

**Describe the solution you'd like**
It would be great if Home Assistant companion app was able to control other apps on the phone by sending Intents after it receives a notification from Home Assistant server.

**Describe alternatives you've considered, if any**
None

**Additional context**

fortepc:
Just to add a bit here since dshokouhi directed me here after I incorrectly assumed the new uri support would work with a tel:1234567890 link, it just want to make sure this request would cover supporting some of the basic uri links I'm use to using in HTML like what I've put in this js bin as an example. https://jsbin.com/zexodovaho/edit?html,output 

```html
  <p><a href="tel:1112223333">tel:1112223333</a> should open up in your dialer to call someone, regardless of which phone or dialer app you use</p>
  <p><a href="sms:1112223333">sms:1112223333</a> should open up in your default SMS app to text someone, regardless of which phone or sms app you use</p>
  <p><a href="geo:40.044167,-86.140940">geo:40.044167,-86.140940</a> should open up in your Map app regardless of what phone or map app you use, and if you have multiple map apps the phone will prompt you to select which one you want to open it with.</p>
```

The reason I bring this up is because my understanding of intents is that you are sending specific information into a specific app, and if the user I'm sending the actionable notification is using `com.textra` for their texting app instead of `com.google.android.apps.messaging` I'd have to know specify that in the intent. If I'm right, may I suggest adding a fallback option and to just process the uri the same way the browser does it if it's in an href? If I'm wrong, well I'm sorry for my ignorance 😅

PR #1129, opened 2020-10-31
Allow sending broadcast intents to other apps via notification

dshokouhi:
Fixes: #1020 

We must set `message: command_broadcast_intent` and the `title` must contain the Action and `channel` must contain the Package name.

Format:

```
message: command_broadcast_intent
title: ACTION
data:
  channel: PACKAGE_NAME
```

Tested with sleep as android API: https://docs.sleep.urbandroid.org/devs/intent_api.html#action-intents-to-control-sleep

Example:

```
message: command_broadcast_intent
title: com.urbandroid.sleep.alarmclock.STOP_SLEEP_TRACK
data:
  channel: com.urbandroid.sleep
```

If `title` or `channel` is not provided we post the notification to the device as normal, otherwise if there is another failure we show a toast.

Docs: https://github.com/home-assistant/companion.home-assistant/pull/376
FCM: pending


JBassett (changes requested):
Rather than overloading title to have both package and action lets break it up into 2 fields.

JBassett (approved):

PR #1130, opened 2020-10-31
Revert material bump as it introduced a dark theme issue with the widgets

dshokouhi:
The icon for the service button widget was coming in as light theme when the device theme was set to dark, reverting the library bump fixes the issue.  Seems the new library version introduced an issue where the night theme color is not being read properly.

Ref: https://discord.com/channels/330944238910963714/562408603345092636/772138149395824661


JBassett (approved):

SeLLeRoNe:
Since this update the bulb icon on the widget turned black.

dshokouhi:
@SeLLeRoNe please do not comment on merged PRs, always create an issue if you find one.  For this the issue actually existed before this PR, I thought it fixed the issue but it didn't.  We may have a solution to check and see if it actually resolves it.

PR #1131, opened 2020-10-31
Add last update sensor

dshokouhi:
Fixes: #529 

The state will either be `SensorWorker` or the intent that caused the update to be sent.

Docs: https://github.com/home-assistant/companion.home-assistant/pull/377

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/LastUpdateManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/LastUpdateManager.kt
@@ -31,7 +31,7 @@ class LastUpdateManager : SensorManager {
     override fun requestSensorUpdate(
         context: Context
     ) {
-        sendLastUpdate(context, "")
+        // No op
     }
 
     fun sendLastUpdate(context: Context, intentAction: String?) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/LastUpdateManager.kt
```diff
@@ -0,0 +1,56 @@
...
+        get() = listOf(lastUpdate)
+
+    override fun requiredPermissions(sensorId: String): Array<String> {
+        return emptyArray()
+    }
+
+    override fun requestSensorUpdate(
+        context: Context
+    ) {
+        sendLastUpdate(context, "")
```
JBassett, line 34: Why even call this if all it ever does is return without actually updating?

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/sensors/LastUpdateManager.kt
dshokouhi: Force of habit lol

PR #1138, opened 2020-11-02
Sentry Filtering

JBassett:
## Summary
This should remove exceptions about being unable to connect home assistant instances.  This will greatly reduce the number of events in Sentry and help with diagnosing issues.

## Screenshots

N/A

N/A

## Any other notes

Sentry: ANDROID-1EB, ANDROID-1E6, ANDROID-1EK, ANDROID-1F3, etc

### DIFF
--- a/app/src/full/AndroidManifest.xml
+++ b/app/src/full/AndroidManifest.xml
@@ -5,11 +5,7 @@
 
     <application
         android:name=".HomeAssistantApplication" >
-        <meta-data android:name="io.sentry.auto-init" android:value="${sentryEnabled}" />
-        <meta-data android:name="io.sentry.release" android:value="${sentryRelease}" />
-        <meta-data android:name="io.sentry.session-tracking.enable" android:value="true" />
-        <meta-data android:name="io.sentry.ndk.enable" android:value="false" />
-        <meta-data android:name="io.sentry.dsn" android:value="https://2d646f40f9574e0b9579e301a69bb030@o427061.ingest.sentry.io/5372876" />
+        <meta-data android:name="io.sentry.auto-init" android:value="false" />
 
         <receiver
             android:name=".notifications.NotificationActionReceiver"
--- a/app/src/full/AndroidManifest.xml
+++ b/app/src/full/AndroidManifest.xml
@@ -6,6 +6,7 @@
     <application
         android:name=".HomeAssistantApplication" >
         <meta-data android:name="io.sentry.auto-init" android:value="false" />
+        <meta-data android:name="io.sentry.release" android:value="${sentryRelease}" />
 
         <receiver
             android:name=".notifications.NotificationActionReceiver"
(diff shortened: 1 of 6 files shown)


app/src/full/AndroidManifest.xml
```diff
@@ -5,11 +5,8 @@
 
     <application
         android:name=".HomeAssistantApplication" >
-        <meta-data android:name="io.sentry.auto-init" android:value="${sentryEnabled}" />
+        <meta-data android:name="io.sentry.auto-init" android:value="false" />
         <meta-data android:name="io.sentry.release" android:value="${sentryRelease}" />
-        <meta-data android:name="io.sentry.session-tracking.enable" android:value="true" />
-        <meta-data android:name="io.sentry.ndk.enable" android:value="false" />
```
dshokouhi, line 11: So this was added due to a specific crash in the Pixel 4 XL, I may need to test it locally and see if the crash continues or not with debug turned off.

app/src/full/java/io/homeassistant/companion/android/CrashHandling.kt
```diff
@@ -0,0 +1,33 @@
...
+import javax.net.ssl.SSLProtocolException
+
+fun initCrashReporting(context: Context) {
+    // Don't init on debug builds
+    if (BuildConfig.DEBUG)
+        return
+
+    SentryAndroid.init(context) { options ->
+        options.isEnableSessionTracking = true
+        options.isEnableNdk = false
```
dshokouhi, line 19: I see now, disregard my deleted comment lol

app/src/full/java/io/homeassistant/companion/android/CrashHandling.kt
```diff
@@ -0,0 +1,33 @@
...
+        options.dsn = "https://2d646f40f9574e0b9579e301a69bb030@o427061.ingest.sentry.io/5372876"
+        options.setBeforeSend { event, hint ->
+            return@setBeforeSend when (hint) {
+                is ConnectException,
+                is SocketTimeoutException,
+                is SSLException,
+                is SSLHandshakeException,
+                is SSLProtocolException,
+                is UnknownHostException -> null
+                else -> event
```
dshokouhi, line 29: Should we include `HTTPException`? Like: https://sentry.io/organizations/home-assistant-io/issues/1823402301/?project=5372876&query=is%3Aunresolved

Sentry also has their ANR issues that I am not sure if we need them? https://sentry.io/organizations/home-assistant-io/issues/1863659368/?project=5372876&query=is%3Aunresolved

JBassett: > Should we include `HTTPException`?

Because of how the dependencies are setup we can't right now, but we can explore in the future.

> Sentry also has their ANR issues that I am not sure if we need them?

ANR (App not responding) are events we should investigate further.  Most likely means we are doing something on the main thread we shouldn't be.

dshokouhi (approved):
Looks good to me, we don't need to bump the library do we? I noticed they changed the dependency name lately 😬

JBassett:
> Looks good to me, we don't need to bump the library do we? I noticed they changed the dependency name lately

Good catch, just checked https://docs.sentry.io/platforms/android/migration/ and validated our upgrade with no issues.

PR #1154, opened 2020-11-06
Add support for Chronometer notifications

gadgetchnnel:
## Summary

Android notifications have a **Chronometer** option which shows a timer at the top which counts up or down to the given time. This PR enables the sending of notifications with this option. This takes two new options under **data**:

* chronometer - true to enable chronometer mode
* when - timestamp in milliseconds of the time to count up or down to 

For example:

```yaml
  - alias: Notify of Next Alarm Time
    trigger:
      ...
    action:
      service: notify.mobile_app_<your_device_id_here>
      data:
        title: Next Alarm
        message: >-
          Next Alarm At {{ states('sensor.<your_device_id_here>_next_alarm') }}
        data:
          chronometer: true
          when: >-
            {{ state_attr('sensor.<your_device_id_here>_next_alarm', 'Time in Milliseconds') / 1000 }}
```

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#386

## Any other notes

This requires additional keys in the FCM repo - link to PR: https://github.com/home-assistant/mobile-apps-fcm-push/pull/28

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -411,7 +411,7 @@ class MessagingService : FirebaseMessagingService() {
         builder: NotificationCompat.Builder,
         data: Map<String, String>
     ) {
-        val notificationWhen = data[WHEN]?.toLong() ?: 0
+        val notificationWhen = data[WHEN]?.toLong()?.times(1000) ?: 0
         val usesChronometer = data[CHRONOMETER]?.toBoolean() ?: false
 
         if (notificationWhen != 0L) {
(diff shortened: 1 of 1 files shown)


app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -402,6 +407,25 @@ class MessagingService : FirebaseMessagingService() {
         }
     }
 
+    private fun handleChronometer(
+        builder: NotificationCompat.Builder,
+        data: Map<String, String>
+    ) {
+        val notificationWhen = data[WHEN]?.toLong() ?: 0
```
dshokouhi, line 414: So one of the existing features we offer is `timeout` which currently takes time in seconds and later on we do the conversion to milliseconds. Do we want to keep things consistent here?

gadgetchnnel: Updated to treat this as seconds

dshokouhi: So I apologize, when I commented on this I didn't realize the time in milliseconds was based on a date rather than a duration.  So in this case it is indeed different and you were correct from the beginning.  Lets wait for the official review before changing it.  Sorry for any inconvenience :)

JBassett (approved):

Issue #942, opened 2020-09-16
Unable to read/write NFC tag


aaamoeder:
**Home Assistant Android version:** beta-382-532ae9d-full

**Android version:** 10

**Phone model:** Oneplus7

**Home Assistant version:** 114.4

**Last working Home Assistant release (if known):** n/a

**Description of problem:** I cannot read or write an NFC tag which I can otherwise read or write fine using the "NFC tools" app.

**Traceback (if applicable):** n/a

```

```

**Screenshot of problem:**
![image](https://user-images.githubusercontent.com/20947118/93369365-938fc880-f84f-11ea-8f1e-1e19463f3ba5.png)
![image](https://user-images.githubusercontent.com/20947118/93369398-9be80380-f84f-11ea-8a9e-c8931885fafb.png)
![image](https://user-images.githubusercontent.com/20947118/93369410-a0acb780-f84f-11ea-84b8-eb045a2babd9.png)

**Additional information:** As you can see in the screenshot I have written a small text "by nfc tools" to the tag, so nothing should be wrong with the tag imo

urfin78:
I have the same problem with a different kind of tag and I think is related to the size of the tag - mine is 46byte. 
If i try to write an uuid like: 554ee751-3adf-4f92-87f5-7517dcb60ec2 (which was created when i add a nfc tag manually in the homeassistant v0.115.0b11 web interface ) with the nfc tools app, it fails with an error stating not enough capacity.

aaamoeder:
That might indeed be the case, 
I'm also unable to write any longer texts in the data field.
Perhaps they should give some minimum requirements for NFC tags before I order some that I can't use ..

dshokouhi:
NTAG215 is what i have personally used...the only requirements I know of are mentioned here: https://companion.home-assistant.io/docs/integrations/universal-links

I am unsure of what size to use but I think if you stick to a re-writable tag you should be ok

JBassett:
I think the minimum byte count is 70?  This is actually an example of the tag we write.
`https://www.home-assistant.io/tag/123e4567-e89b-12d3-a456-426614174000`

hirschy81:
same problem running HA 115.2

Armadill0:
Same problem here with 0.115.4 reading a NTAG216. Working fine with NFC Tools:
![photo_2020-09-28_21-26-33](https://user-images.githubusercontent.com/6536990/94476842-502f4580-01d1-11eb-9620-32652edcc6cd.jpg)

[...]

PR #1178, opened 2020-11-09
Add some debug logs for scanning a tag

dshokouhi:
## Summary

Lets add some debug logs so users can help investigate #942 

## Screenshots

n/a

## Any other notes

Lets see if we can get some users to test the debug apk before merging?

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
@@ -224,6 +224,7 @@ class IntegrationRepositoryImpl @Inject constructor(
                     ).isSuccessful
             } catch (e: Exception) {
                 // Ignore failure until we are out of URLS to try!
+                Log.d(TAG, "Unable to send tag data for URL: $it with data: $data", e)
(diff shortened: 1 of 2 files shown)


common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
```diff
@@ -224,6 +224,7 @@ class IntegrationRepositoryImpl @Inject constructor(
                     ).isSuccessful
             } catch (e: Exception) {
                 // Ignore failure until we are out of URLS to try!
+                Log.d(TAG, "Unable to send tag data for URL: $it with data: $data", e)
```
JBassett, line 227: Not sure if we want to add this, it's going to spam a lot of logs for some peoples setups.

dshokouhi: ok I was thinking maybe the call here failed.  I will instead add a logger to `handleTag` function to determine if the call was successful or not.  I suspect based on the logging, the issue either happens when we are parsing the tag for the ID or when we send the data back home.

JBassett (approved):

Issue #1174, opened 2020-11-09
Add current time/timezone sensor


FuzzyMistborn:
**Is your feature request related to a problem? Please describe.**

Not a "problem" but something I'd like added to better allow some automating.

**Describe the solution you'd like**

A sensor that has the phone's current time and/or timezone.  Having one for current time as well as a separate sensor for timezone would be handy.

**Describe alternatives you've considered, if any**

**Additional context**

I'd like this to combine with the last_alarm sensor.  I control my phone alarm via HASS, and sometimes set an earlier alarm than my "normal" alarm, so I don't want my "normal" alarm reset until after that time has passed.  IE i set an alarm for 3am but my normal alarm is 6, I don't want to wake up at 3 and have my 6am alarm turned back on until after 6am has past).  I worked around this by having a reset at a later time of the day, but that's tied to my current timezone.  If i'm traveling there's the issue of timezones potentially screwing things up, so having the phone's current timezone/time would be useful.

FuzzyMistborn:
#1182 adds the timezone, I was also looking for a sensor with the actual local time as well if that's possible.

JBassett:
Due to how we we have to update the time sensor we won't be adding a current local time.  You should be able to derive the devices local time in Home Assistant by using the time zone information.

FuzzyMistborn:
Understood. Thank you.

PR #1182, opened 2020-11-10
Add a time zone sensor

dshokouhi:
## Summary

Fixes: #1174 by providing a sensor that represents the current time zone. There are also a few attributes to help describe the time zone such as if its currently observing day light savings time, the ID of the zone (similar to that of HA), the short name and also if the time zone will ever observe day light savings.

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#392

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/TimeZoneManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/TimeZoneManager.kt
@@ -1,6 +1,7 @@
 package io.homeassistant.companion.android.sensors
 
 import android.content.Context
+import android.os.SystemClock
 import io.homeassistant.companion.android.R
 import java.util.Date
 import java.util.Locale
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/TimeZoneManager.kt
```diff
@@ -0,0 +1,62 @@
...
+        onSensorUpdated(context,
+            currentTimeZone,
+            currentZone,
+            icon,
+            mapOf(
+                "in_daylight_time" to timeZone.inDaylightTime(date),
+                "time_zone_id" to timeZone.id,
+                "time_zone_short" to timeZone.getDisplayName(timeZone.inDaylightTime(date), TimeZone.SHORT, Locale.ENGLISH),
+                "uses_daylight_time" to timeZone.useDaylightTime()
+            )
```
JBassett, line 59: I think we should add the offset from UTC as well.

JBassett (changes requested):

JBassett (approved):

Issue #448, opened 2020-02-18
Home network does not work


pippo73:
**Home Assistant Android version:**
1.7.2-130

**Android version:**
Android 10 arrow rom
**Phone model:**
Xiaomi mipad4
**Home Assistant version:**
Home Assistant 0.105.5
**Last working Home Assistant release (if known):**
Don't know 
**Description of problem:**
I've configured both the internal and the external addresses and the ssdi
But when I turn back to the app it says that in not possible to connect to the server

**Traceback (if applicable):**

```

```

**Screenshot of problem:**

![Screenshot_Home_Assistant_20200218-211219](https://user-images.githubusercontent.com/757121/74780853-bac5ed80-52a0-11ea-9fd2-01907f27610d.png)

![Screenshot_Home_Assistant_20200218-225241](https://user-images.githubusercontent.com/757121/74781423-be0da900-52a1-11ea-972e-c82593dda3b9.png)

**Additional information:**

dshokouhi:
If you need to accept the SSL certificate in a browser when you use your local IP like you have defined then this will not work.  Your best bet would be to enable NAT loopback at the router and just use the external URL.  This may also be of help: https://companion.home-assistant.io/docs/troubleshooting/networking

pippo73:
I think that there is no way to have an ssl certificate on the FQDN and avoid it in the internal address. If not, please, could you explain me how? Unfortunately my router doesn't allow me to set the NAT loopback. 
I think mine is quite a common situation. 
I think I will just use just the external address since it will work always 😉
Thank you @dshokouhi anyway

GreenhouseAU:
I ran in to the same problem but the fix that worked for me is I have to enable google location. If I have it disabled, only loads remote. If it's enabled, it behaves as expected.

I think this is a bug and makes the app not useful for me. I keep location disabled expect when I'm driving (and it's too inconvenient to keep changing location 'modes' in android).

Is this the only option for the app figuring out it's connected to a certain ssid?

dshokouhi:
@GreenhouseAU as of right now the only way to determine if you are on the home network is for location services to be enabled.  Please file a new issue for this as its not related to this issue.

@pippo73 to my understanding this option is really more for users where the external URL does not resolve locally for them, if the external URL works locally then you don't need to use the feature.

SteveDinn:
> I think that there is no way to have an ssl certificate on the FQDN and avoid it in the internal address. If not, please, could you explain me how? 

[...]

Issue #924, opened 2020-09-13
Wifi not working correctly


TECHNOFAB11:
**Home Assistant Android version:**
2.4.0-full

**Android version:**
OxygenOS 10.3.4.GM57AA (Android 10)

**Phone model:**
OnePlus 7

**Home Assistant version:**
| | |
|---|---|
|arch|x86_64|
|dev|false|
|docker|true|
|hassio|false|
|installation_type|Home Assistant Container|
|os_name|Linux|
|os_version|5.4.0-0.bpo.4-amd64|
|python_version|3.8.3|
|timezone|Europe/Berlin|
|version|0.114.4|
|virtualenv|false|

**Last working Home Assistant release (if known):**
None

**Description of problem:**
I had a lot of problems over the last months with getting ip banned via the app. I was wondering why, as it shouldn't ban local addresses. Thanks to the authenticated sensor I saw that the app connected via the public url all the time (weirdly enough the widgets didn't as far as I can tell). Now I looked into the sensor settings in HA Android and there seems to be a WiFi issue. It knows that WiFi is on, it knows my internal IP, but it still says "not connected". HA has all permissions, but it still doesn't seem to be able to get the wifi information (BSSID is always 02:00:00:00...) like seen on the screenshots. 

**Traceback (if applicable):**
```
```

**Screenshot of problem:**
![Screenshot 1](https://user-images.githubusercontent.com/34860318/93016198-a2604c00-f5bf-11ea-8dec-9283a42c64a3.jpg)
![Screenshot 2](https://user-images.githubusercontent.com/34860318/93016199-a2f8e280-f5bf-11ea-8bed-072b9b783a44.jpg)

**Additional information:**

dshokouhi:
Your SSID has `is_hidden` set to true so more than likely the API is not able to retrieve the SSID (at least that is my guess).  All the app is doing is grabbing the API data and sending the data to HA.  We don't change any of the data unless we get nothing in response from the API.  If you set your network not to be a hidden network do you see the sensor functioning as expected?

smartviperbg:
Same here with Huawei P30 Pro and P30 lite with last updatet androids. Before update the home assistant app it's working great and shows all of my networks.

Maltavius:
Have the same issue on my Wife's Samsung S8
Sometimes it works and sometimes it's just Unkown SSID
![image](https://user-images.githubusercontent.com/10133945/94363432-5b478000-00c2-11eb-81f3-76a3b979e82a.png)

[...]

PR #1187, opened 2020-11-11
Improve permission check/Add location disabled check

chriss158:
## Summary

#448 introduces a missing location permission check for the wifi option.

This PR fixes missing location permission check when using the wifi option. If the user tries to maintain wifi ssids, a location permission dialog popups. If the user decides to grant the permission, the ssid dialog will popup. If the user denies the permission no ssid dialog will be displayed.

Next step would be to check if the user has disabled location in the android settings and show a warn dialog. 

## Any other notes

I'm unsure if we need to check for ACCESS_BACKGROUND_LOCATION as the SSID is also determined in background. Method `isInternal()` of class  `UrlRepositoryImpl`. Any ideas?


dshokouhi:
There is one other check required. It was discovered #924 that the SSID will not be pulled if GPS is disabled for the device, despite having proper location permissions.  So we also need to check for that as well.  In Android 11 there does not seem to be a way to disable GPS but on certain versions of android they are and we should probably account for this too.

chriss158:
Yes, that's what I meant by
`Next step would be to check if the user has disabled location in the android settings and show a warn dialog.`

So a warn dialog should be shown on the wifi option and all wifi sensors if they are enabled?

chriss158:
This dialog will be shown if location service is disabled an on click on the wifi option. Please check for correct grammar/spelling ^^

dshokouhi:
@chriss158 may want to remove that screenshot ;)  

> So a warn dialog should be shown on the wifi option and all wifi sensors if they are enabled?

For the WiFi sensors we need to add the check in a few places actually.  When the user enables an individual sensor, when the user enables all sensors and also we need to account for when the sensor gets an update and GPS was disabled.  We do check permissions but GPS being disabled is a different check we need to include.

I like the dialog but I wonder if we should hide the WiFi and internal URL bits while permissions and GPS are disabled? This way existing users know whats required of them?

> This dialog will be shown if location service is disabled and on click on the wifi option. Please check for correct grammar/spelling 😅
> If the user clicks on "settings" he will be redirected to the location setting view.

I think maybe just drop `service` from `location service`?

chriss158:
Thanks. I didn't noticed the dark url in the background :cry:

You mean that we should disabled these two option while location is disabled? How we get the user informed that he need to enable the location service?

dshokouhi:
If the settings won't function without the permissions and GPS being enabled then yes. No point in letting the user interact with something they can't use until they perform the action.  We may just need to inform the user as they come to the settings screen or post a notification on the device if they are using internal URL and meet the criteria?

@JBassett what do you think would be a proper approach here?

chriss158:
This way users which are already using this option but have disabled location service, gets warned on the next time they open the settings. And not on the next time they click on the wifi option. This is definitely an advantage .

JBassett:
I think fully disabling might cause more confusion.  Have the dialog seems like a decent approach.  I would go as far to say that we should surface that dialog on the Web View if someone disables location permissions.

chriss158:
Implemented the location check on

- Resume webview 
- Onboarding location tracker setting
- Wifi SSID setting
- All individual sensors which are using ACCESS_FINE_LOCATION or ACCESS_COARSE_LOCATION as permission
- Enable all sensors

**Resume webview**

If the user clicks yes, he will redirected to the location service setting of android.
If the user clicks no, everything remains as it is. Options will be left enabled, but in a not working state duo to disabled location service.
If the user decides to disable the corresponding sensors, the he can just go to the corresponding sensor settings and disable them. 
If the user decides to remove the Wifi SSID, then following dialog will be shown.

Here the user can either disable the option. Then all wifi ssids will be cleard, or again go to the location setting of android.


**Onboarding location tracker setting**

If the user clicks no, the setting will remain disabled, no sensor will be activated.
If the user clicks yes, he will redirected to the location service setting of android.

**Wifi SSID setting**

If the user clicks no, no dialog for wifi ssid pop ups.
If the user clicks yes, he will redirected to the location service setting of android.



**All individual sensors which are using ACCESS_FINE_LOCATION or ACCESS_COARSE_LOCATION as permission**

If the user clicks no, the setting will remain disabled, no sensor will be activated.
If the user clicks yes, he will redirected to the location service setting of android.

**Enable all sensors**

If the user clicks no, the setting will remain disabled, no sensor at all will be activated.
If the user clicks yes, he will redirected to the location service setting of android.

app/src/main/java/io/homeassistant/companion/android/util/DisabledLocationHandler.kt
```diff
@@ -0,0 +1,60 @@
...
+import androidx.appcompat.app.AlertDialog
+import io.homeassistant.companion.android.R
+
+object DisabledLocationHandler {
+    fun isLocationEnabled(context: Context): Boolean {
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            // This is new method provided in API 28
+            val lm: LocationManager =
+                context.getSystemService(Context.LOCATION_SERVICE) as LocationManager
+            lm.isLocationEnabled
```
dshokouhi, line 19: So it was to my understanding that certain versions of Android allow the user to keep location enabled however the user can optionally turn off GPS.  I thought what we needed to do here was to also check:

`lm.isProviderEnabled(LocationManager.GPS_PROVIDER)`

Which tells us if GPS itself was disabled.  According to the users in https://github.com/home-assistant/android/issues/924#issuecomment-700430278 they are able to leave location on but GPS off.  Did you find similar results too?

Judging by the API it seems both calls can return different results unless I am misunderstanding this?

https://developer.android.com/reference/android/location/LocationManager#isProviderEnabled(java.lang.String)
https://developer.android.com/reference/android/location/LocationManager#isLocationEnabled()

chriss158: Yes is i think you are right. With Android P onwards it is enough to call `LocationManger.isLocationEnabled`. For everthing else we should call `LocationManger.isProviderEnabled(LocationManager.NETWORK_PROVIDER)` or `LocationManger.isProviderEnabled(LocationManager.GPS_PROVIDER)` depending on which permission is needed (COARSE or FINE).

I've updated the code to this end.

chriss158:
Some points which came to my mind while testing:

- The request for fine/coarse permission and for background permission on the wifi setting is for the user not reasonable.  Maybe a a explanation popup (with a link to the wiki) should be shown before requesting those permissions. 

- Revoked permessions are not rechecked on startup of the app. With this PR this would be easy to check on startup.

app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -248,15 +249,22 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
...
-                    permissionsToCheck = arrayOf(Manifest.permission.ACCESS_COARSE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION)
-                }
+            val fineLocation = DisabledLocationHandler.containsFineLocationPermission(permissionsToCheck)
 
-                if (permissionsToCheck.isNullOrEmpty() || checkPermission(permissionsToCheck, LOCATION_REQUEST_CODE)) {
+            if (DisabledLocationHandler.isLocationEnabled(requireContext(), fineLocation)) {
+                var permissionsToRequest: Array<String>? = null
+                if (!permissionsToCheck.isNullOrEmpty() && Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                    // For Android 11 we need to remove the Background Location Permission Check
+                    // as for Android 11 the Background Location Request needs to be done separately
```
dshokouhi, line 264: I don't think this should be limited to android 11 in the condition, per androids docs this enforcement is for apps that *target* android 11 so I think it enforces it on everyone right? 

https://developer.android.com/about/versions/11/privacy/location#request-background-location-separately

`If your app targets Android 11 or higher, the system enforces this best practice.`

This sounds like we have to keep it this way for all users right?  Unless I misunderstood this part?

chriss158: Ups this comment confusing. The background permission check is not removed. Only the request for background permission is removed. The code should be fine. Only the comment is wrong. I will fix that comment.

dshokouhi: So beforehand we never added this check specifically for android 11: https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt#L122

I wonder if we should just use the best practice on all versions that require background access instead of an additional condition check?

chriss158: Personally i wouldn't do that for all versions, because that would be again another popup for the user. I think on Android 11 these two popups (which are required) are also confusing. But that's just the way it is.

Btw shouldn't this line you mentioned be only for Android >=R instead of >= Q?

dshokouhi: So background permissions is required for android Q so we do the split there.  To the end user nothing changes.  Foreground and background permissions have always been 2 separate pop-ups.  Now one needs to be granted before the other and we handle that in the code so to the user the only change is the order of granting permissions.  The additional pop-up you see on the onboarding screen is new and requested by Google as a prominent disclosure, this is part of the new policy changes they had.

app/src/main/java/io/homeassistant/companion/android/util/DisabledLocationHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    fun containsFineLocationPermission(permissions: Array<String>): Boolean {
+        return permissions.contains(Manifest.permission.ACCESS_FINE_LOCATION)
+    }
+
+    fun containsCoarseLocationPermission(permissions: Array<String>): Boolean {
+        return permissions.contains(Manifest.permission.ACCESS_COARSE_LOCATION)
+    }
+
+    fun showLocationDisabledWarnDialog(activity: Activity, context: Context, message: String, withDisableOption: Boolean = false, callback: (() -> Unit)? = null) {
```
JBassett, line 33: Rather than having multiple options for the dialog lets just use the `location_disabled_general_message` one and list all sensors that will be disabled?

JBassett (changes requested):
I think we should just stick with a single alert type with the same verbiage no matter what.  Might need to move the `checkAndWarnForDisabledLocation` function into the `DisabledLocationHandler`.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -246,6 +249,10 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                     }
                     return false
                 }
+
+                override fun onPageFinished(view: WebView?, url: String?) {
+                    super.onPageFinished(view, url)
+                }
```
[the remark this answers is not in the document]
chriss158: Ups still left from my test. Removed.

app/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationFragment.kt
```diff
@@ -71,23 +72,33 @@ class MobileAppIntegrationFragment : Fragment(), MobileAppIntegrationView {
...
-                            .setPositiveButton(android.R.string.ok) { _, _ ->
-                                requestPermissions(
-                                    sensorId
-                                )
+                    var checked = false
+                    if (isChecked) {
+                        if (!DisabledLocationHandler.isLocationEnabled(context, true)) {
+                            DisabledLocationHandler.showLocationDisabledWarnDialog(requireActivity(), context, context.getString(R.string.location_disabled_option_message))
+                        } else {
+                            if (!LocationSensorManager().checkPermission(requireContext(), sensorId)) {
```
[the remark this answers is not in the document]
chriss158: Done. Simplified it.

app/src/main/java/io/homeassistant/companion/android/util/DisabledLocationHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+            .setTitle(R.string.location_disabled_title)
+            .setMessage(message)
+            .setPositiveButton(positionTextId) { _, _ ->
+                val intent = Intent(
+                    Settings.ACTION_LOCATION_SOURCE_SETTINGS
+                )
+                intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
+                intent.addFlags(Intent.FLAG_ACTIVITY_NO_HISTORY)
+                intent.addFlags(Intent.FLAG_ACTIVITY_EXCLUDE_FROM_RECENTS)
+                context.startActivity(intent)
```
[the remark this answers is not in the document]
chriss158: Done.

app/src/main/java/io/homeassistant/companion/android/util/DisabledLocationHandler.kt
chriss158: You are right. Doesn't hurt to have this information on the popup.

chriss158:
Following changes done:
* Improved enable all sensor settings. The enable all does now enable all other sensors
   when location is disabled, instead of not allow to turn on the other sensors too
* The enable all sensor does now deactivate the sensor if a permission denied
* Request location background permission in two steps for R upwards instead of Q upwards (The two step permission request is first needed on R).
* Location permission info dialog added on all wifi options. This is to inform the user why this permission is needed

JBassett (approved):

RobertFey:
Hi, verry annoiing this feature.
Whatever set i do. Like disable all sensors or agree or disagree with using location this popup comes always up.

dshokouhi:
@RobertFey it is not just sensors but also the internal URL feature.  If you are using any feature that requires location to be enabled and you don't have it enabled then you can expect to see this pop-up all the time.  Do not use features that require location to prevent this pop-up.  The instructions in the pop-up are outlined there.  If you continue to have issues after following all of the instructions please create a new issue or find an existing one if it exists.  Merged PR's are not a place for issue discussion.

RobertFey:
> @RobertFey it is not just sensors but also the internal URL feature. If you are using any feature that requires location to be enabled and you don't have it enabled then you can expect to see this pop-up all the time. Do not use features that require location to prevent this pop-up. The instructions in the pop-up are outlined there. If you continue to have issues after following all of the instructions please create a new issue or find an existing one if it exists. Merged PR's are not a place for issue discussion.

Thank you for the reply @dshokouhi . You are right. I will check and try later and will create an issue if applicable.
Have a nice evening.

cyclops1982:
Hi,

First of all - thanks for a great working android app. I'm one of those users that uses it every day, with great pleasure.

I however am not very happy with the current update that introduced this check. The reason is that i use the app primarily at home, when i don't care about my location being correctly reported (i connect to the WiFi which does my home/away-status detection).

When i go out of the house (geez, it doesn't happen very often with this lockdown), i turn on my location most of the times because of navigation, or to make sure my HA knows where i am.

I happen to be good with DNS, so my internal and external URL are the same...

So, at the moment this confirmation happens *every time* i open the app when i'm at home and have my location turned off. I know it's only one click to say 'no', but it's very annoying for sure....

Would it be an option that we get a configuration setting for this check to be performed every time? That would solve things for me..

soxhi8:
The pull request makes for a rather bad user experience. Getting people to know why a certain feature won't work without location services is important, but throwing a pop-up on every startup is annoying. There are situations in which a user actively decides to turn off location services.

From my personal experience, informing the user at a feature level is plenty, giving a one-time popup is more than enough. A checkbox "Don't remind me again, for this feature" might be a usable compromise.

dshokouhi:
Please do not comment on merged PR if you guys want to turn off location that's fine but you need to stop using the features. A lot of users have been confused as to why certain features are no longer working so we need to inform all users. Sorry if it's an inconvenience but the solution is to either leave location on with the features you want or disable those features to prevent the pop-up.

cyclops1982:
@dshokouhi 
Sorry to comment again - should i open a new issue?

The issue is not that we don't like the feature, or don't want the feature turned of. We just don't want the question/check *every time* we open the app.
Like. I turned on the light, went to another app to do something, switched back because the light was too bright and *boom* the question comes up again.

How it worked before was, that the sensors were just not updated correctly. That's fine of us. We dont' care about location when we have GPS turned of. That's a choice right?

The issue is that the current dialog has the option for 'yes' or 'no'.
It should IMHO have 3rd open that says "ok, don't ask again"

Let me know if i should create a new issue for that or not.

chriss158:
Maybe just open a new issue to continue discussing 👍

PR #1194, opened 2020-11-14
Pause and resume the webview as needed

### LINKED ISSUE #979: Camera traffic, when app in background.
2.4.1-full

**Android version:**
10
**Phone model:**
Xiaomi mi 9
**Home Assistant version:**
0.115.3
no
**Description of problem:**

When app in background, video streams from cameras does not stop stream. Almost all apps stops video stream at onPause() Android lifecycle, but HA does not. Now the problem is even more serious, because from the activities of the sensors, the system almost never kills the application. I have a lot of cameras and I wasted 70 gigabytes in a month because of this bug. Plus the battery runs out quickly.

**Screenshot of problem:**
Screenshot_2020-09-25-18-25-18-432_com miui securitycenter

**Additional information:**
Solution: You should stop stream, when device goes onStop(), maybe communicate with HA server, and show video placeholder.

dshokouhi:
## Summary

This is an attempt to fix #979 from the app side.  I was looking at the WebView docs and it seems that webview itself can be paused and resumed as needed.  I think this might help out.  This is what I did for testing.

1. Ran `master` before these changes.  Loaded up a live view picture entity card, verified the camera was streaming.  Swiped to home screen.  Waited 10 minutes, came back to webview and it seemed like the video continued.  Performed a similar test with more info panel and saw the same results.
2. Ran these changes and did the same test.  After waiting 10 minutes and coming back to webview I noticed the whole page reloaded, I think this was caused by the connection restoring?  When I did the test on the more info popup I noticed that when I came back the video was paused at first and then resumed a second later.

Its possible that `webView.onPause` is what we need but I am not sure. Based on the docs it sounds like it can't hurt?

https://developer.android.com/reference/android/webkit/WebView#onPause()
https://developer.android.com/reference/android/webkit/WebView#onResume()

I am not sure if these are needed but since the app is in the background we probably don't need the timers running? (if we have them or if a custom card can?)
https://developer.android.com/reference/android/webkit/WebView#pauseTimers()
https://developer.android.com/reference/android/webkit/WebView#resumeTimers()

## Screenshots

N/A

Documentation: home-assistant/companion.home-assistant#N/A

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -433,6 +433,8 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
 
     override fun onResume() {
         super.onResume()
+        webView.onResume()
+        webView.resumeTimers()
         if (currentLang != languagesManager.getCurrentLang())
             recreate()
         if (!unlocked && !presenter.isLockEnabled())
@@ -884,5 +886,11 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                 showError()
             }
         }, CONNECTION_DELAY)
+    }
+
+    override fun onPause() {
+        super.onPause()
+        webView.onPause()
+        webView.pauseTimers()
     }
 }


chriss158:
I can't see any point to let the webview (inc. timers) run in background. The only thing i can imagine is to get the webview not to be reloaded on startup. 

Maybe following setting will help to not reload the webview
https://developer.android.com/guide/topics/manifest/activity-element#always

JBassett (approved):

JBassett:
Lets beta it and see how it does!

dshokouhi:
@chriss158 the reloading is actually closed by the restoring of the closed connection by the frontend if I am not mistaken.  I don't think we have control over that here.  The goal here is that after 5 minutes not only should the connection be closed by all webview processing as well.

PR #1203, opened 2020-11-17
Revert "Pause and resume the webview as needed (#1194)"

dshokouhi:
This reverts commit d99daecae511b1d3bf7eea3ac5644ef0cf266236.

## Summary

Lets revert as its causing reconnection delays.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -433,8 +433,6 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
 
     override fun onResume() {
         super.onResume()
-        webView.onResume()
-        webView.resumeTimers()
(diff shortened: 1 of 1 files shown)


JBassett (approved):

chriss158:
Why exactly did you revert this?

Currently i'm fighting with my samsung tablet which doesn't disconnect after the 5 minutes timeout after the app has been hidden (tablet locked). Same procedure does work on the emulated pixel without problems.  

Then i added only `webView.onPause` and `webView.OnResume` in the corresponding methods. Now also my samsung tablet does disconnect after the 5 minute timeout. 

I did not  see any side effects or connection delays. Maybe the `pauseTimers` and `resumeTimers` should be avoided.

dshokouhi:
After the disconnect the reconnect would take closer to 10 seconds as opposed to the 1 second that we see today. Also the issue that was intended to get fixed was not fixed by the code change so we need to revisit things. We should not see delays when things reconnect.

Issue #1217, opened 2020-11-29
power menu: improper value range for heaters/temperature


flyser:
**Home Assistant Android version:** 3.0.2-full

**Android version:** 11

**Phone model:** Google Pixel 3 XL

**Home Assistant version:**  0.118.4

**Last working Home Assistant release (if known):** none

**Description of problem:**
This is an usability issue and not a real bug. I am using the homeassistant app to control my heaters (via Homematic Integration) and would love to use the awesome new power menu for that as well. However the settable value range of 0 to 100°C does not really make sense for heaters, which are usually controlled between 5 and 30°C (or switched between: off, 5-30°, max).
Is that something that's in the control of the homematic app or is this actually a Homematic integration bug?

dshokouhi:
@flyser does your thermostat have `min_temp` and `max_temp` attributes in the states panel?  This might allow us to provide a better range per entity.

Also 0-100 applies to fahrenheit too which in turn makes some sense but not fully ;)  The current range provides the best for both use cases lol

flyser:
Hey @dshokouhi, thanks for the quick reply :)
Yes they have the min and max_temp attributes. Here is a list of all attributes:

     {'attributes': {'battery': 2.5,
                    'current_temperature': 18.2,
                    'friendly_name': ~,
                    'hvac_modes': ['auto', 'heat', 'off'],
                    'id': ~,
                    'interface': 'wireless',
                    'max_temp': 30.5,
                    'min_temp': 4.5,
                    'target_temp_step': 0.5,
                    'mode': 'Manual',
                    'preset_mode': None,
                    'preset_modes': ['boost', 'comfort', 'eco'], # I wonder what these are. Homematic doesn't have a comfort or eco setting
                    'rssi_peer': -68,
                    'supported_features': 17,
                    'temperature': 5.0,
                    'valve': 0},

T-bond:
I would like to share information about my climate entities as well, maybe it can with the issues in the PR.
One is a car climate control (from a third party integration for [Volkswagen We Connect](https://github.com/robinostlund/homeassistant-volkswagencarnet))
```
hvac_modes: heat, off
min_temp: 7
max_temp: 35
current_temperature: null
temperature: 24
friendly_name: Car
icon: mdi:radiator
assumed_state: true
supported_features: 1
```
And the another one is the HA's generic termostat I set in the configuration for my homemade thermostat:
```
hvac_modes: heat, off
min_temp: 18
max_temp: 27
[...]

PR #1222, opened 2020-12-01
Use min and max temp for climate range in power menu

dshokouhi:
## Summary

Fixes: #1217 

We now set the climate range based on `min_temp` and `max_temp`.  For the current value we first try `temperature` then try `current_temperature` and then default to 0 if both are not provided.  My zwave thermostat does not have a `temperature` attribute when it is `off` but `current_temperature` is always there.  We also do a sanity check to ensure current value is always in the range to prevent loading errors.

## Screenshots

N/A

Documentation: home-assistant/companion.home-assistant# N/A

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
@@ -49,12 +49,21 @@ class ClimateControl {
                     else -> entity.state
                 }
             )
+            val minValue = (entity.attributes["min_temp"] as? Number)?.toFloat() ?: 0f
+            val maxValue = (entity.attributes["max_temp"] as? Number)?.toFloat() ?: 100f
+            var currentValue = (entity.attributes["temperature"] as? Number)?.toFloat() ?: (
+                    entity.attributes["current_temperature"] as? Number)?.toFloat() ?: 0f
+            // Ensure the current value is never lower than the minimum or higher than the maximum
+            if (currentValue < minValue)
+                currentValue = minValue
+            if (currentValue > maxValue)
+                currentValue = maxValue
             control.setControlTemplate(
                 RangeTemplate(
                     entity.entityId,
-                    0f,
-                    100f,
-                    (entity.attributes["temperature"] as? Number)?.toFloat() ?: 0f,
+                    minValue,
+                    maxValue,
+                    currentValue,
                     .5f,
                     "%.0f"
                 )


app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
```diff
@@ -49,12 +49,21 @@ class ClimateControl {
                     else -> entity.state
                 }
             )
+            val minValue = (entity.attributes["min_temp"] as? Number)?.toFloat() ?: 0f
+            val maxValue = (entity.attributes["max_temp"] as? Number)?.toFloat() ?: 100f
```
JBassett, line 53: `min_temp` and `max_temp` on my nest thermostat are used differently.  My target temperature is 70 so min and max are both 70.  This would break the integration for.

dshokouhi: oh, I was afraid of that....hmm so not sure how we can proceed with fixing #1217 its more a usability thing than an actual bug.

JBassett: Perhaps we can get what units are set in HA and then use a better range?

dshokouhi: Yea I will do 2 things in the next update.

1. We will use `unit_system` as defined in https://github.com/home-assistant/core/blob/dev/homeassistant/components/mobile_app/webhook.py#L553
2. I will reference the thermostat card and see if we can use the same ranges defined there to make it match the front end more
3. Probably use some sane defaults if step 2 doesn't work like I think it will, so maybe best to stick to 5-40 for Celsius and 40-90 for Fahrenheit

dshokouhi: Looks like the `climate` integration itself uses a Celsius range....not sure if these defaults make sense either when compared to the users expectation.

https://github.com/home-assistant/core/blob/dev/homeassistant/components/climate/__init__.py#L75

the thermostat card itself goes by `target_temp_low` which also not all thermostats support (like mine) LOL so we may need to stick to sane defaults in this case.

https://github.com/home-assistant/frontend/blob/dev/src/panels/lovelace/cards/hui-thermostat-card.ts#L125

T-bond: Maybe I missed a point, but wouldn't be a good solution to use the PR's current logic, and fall back to some default range (with the new logic of using the `unit_system` for defaults), when the current temperature is equal to the min and max temperature, like in case of JBassett?

dshokouhi: Am I misunderstanding something around the official HA dev docs around climate or is it suggesting that `nest` is not using these attributes appropriately?

https://developers.home-assistant.io/docs/core/entity/climate#properties

The docs do not mention that these attributes must be tied to the current set point of the thermostat.  It sounds like `min_temp` and `max_temp` are more device specific ranges and are not meant to be a set point specific range like the target temps suggest.

JBassett: After looking over the documentation I think you are correct.  They should be using `target_temperature_high` and `target_temperature_low`.  If we get a defect in we might be able to get it fixed before the next release of core then we can merge you change.

JBassett: Scratch that, it's because I have my thermostat locked (kids find it fun to spin the dial, but my bill doesn't find it funny).  Because I have it locked at a single temperature that's what's returned.

JBassett (approved):

PR #1232, opened 2020-12-10
Fix uninitialized property access error

dshokouhi:
## Summary

Was looking at the Sentry logs and came across the following error, we probably want to make sure all supported versions have some location permission to check here:

```
kotlin.UninitializedPropertyAccessException: lateinit property permissionsToCheck has not been initialized
    at io.homeassistant.companion.android.settings.SettingsFragment.onDisplayPreferenceDialog(SettingsFragment.kt:259)
    at androidx.preference.PreferenceManager.showDialog(PreferenceManager.java:537)
    at androidx.preference.DialogPreference.onClick(DialogPreference.java:257)
    at androidx.preference.Preference.performClick(Preference.java:1182)
    at androidx.preference.Preference.performClick(Preference.java:1166)
    at androidx.preference.Preference$1.onClick(Preference.java:181)
    at android.view.View.performClick(View.java:5204)
    at android.view.View$PerformClick.run(View.java:21153)
    at android.os.Handler.handleCallback(Handler.java:739)
    at android.os.Handler.dispatchMessage(Handler.java:95)
    at android.os.Looper.loop(Looper.java:148)
    at android.app.ActivityThread.main(ActivityThread.java:5417)
    at java.lang.reflect.Method.invoke(Method.java)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```
## Screenshots

n/a
## Link to pull request in Documentation repository

Documentation: home-assistant/companion.home-assistant#n/a

PR #1240, opened 2020-12-16
Fix crash when run on Android 7.1 or earlier

zmarties:
## Summary

There was a missing else to set the value of a variable, which meant that the code crashed when run on Android 7.1 or earlier.  Rather than just add that in to the if statement, switch to a when statement which better allows Kotlin to police when variables are not being set.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
@@ -249,13 +249,17 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
 
     override fun onDisplayPreferenceDialog(preference: Preference) {
         if (preference is SsidPreference) {
-            lateinit var permissionsToCheck: Array<String>
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
-                permissionsToCheck = arrayOf(Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION)
-            } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                permissionsToCheck = arrayOf(Manifest.permission.ACCESS_COARSE_LOCATION)
+            val permissionsToCheck: Array<String> = when {
+                Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q -> {
+                    arrayOf(Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION)
+                }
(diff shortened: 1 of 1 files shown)


dshokouhi:
duplicate of #1232

app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -249,13 +249,17 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
...
-                permissionsToCheck = arrayOf(Manifest.permission.ACCESS_COARSE_LOCATION)
+            val permissionsToCheck: Array<String> = when {
+                Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q -> {
+                    arrayOf(Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION)
+                }
+                Build.VERSION.SDK_INT >= Build.VERSION_CODES.O -> {
+                    arrayOf(Manifest.permission.ACCESS_COARSE_LOCATION)
+                }
+                else -> {
+                    arrayOf()
```
dshokouhi, line 260: This should not be blank because we still require permissions for older versions of Android.  The permissions should match the other sensors that rely on it.  This is required for us to get SSID info.

https://github.com/home-assistant/android/blob/master/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt#L504-L513

zmarties: I'm not sure that that is correct.  The app works all the way back to Android 5, but the need to request runtime permissions was not introduced till Android 6.

There may well be a second bug here, where the course location check should be against M not O, but I think there are still Android versions for which an empty array is the correct answer.  It's interesting that there is a second place in the code making the same decisions.  Clearly that code should not be duplicated.

dshokouhi: > It's interesting that there is a second place in the code making the same decisions. Clearly that code should not be duplicated.

This is because sensors as I linked to have separate permission checks that are part of sensor architecture.  The code that you are touching was actually for App Configuration and we also decided to inform the user upon app launch that they are missing permissions or dont have GPS enabled.  The permission and location check is required for us to get certain data.  So the permission check for sensors is still required.

The code that I linked to has been there for a very long time and my own devices on Android 5 do not complain about the run time permissions.  They are simply ignored I think?

zmarties: If it all works fine on Android 5, then that sounds great.  It seems therefore you have this all in hand in the duplicate you linked - should I just close this PR?

PR #1242, opened 2020-12-17
Add a call timeout and some debug logs

### LINKED ISSUE #653: Widget sometimes spins forever
**Home Assistant Android version:** 1.10.0-183

**Android version:** 10

**Phone model:** Pixel 2

**Home Assistant version:** 0.110.4

**Last working Home Assistant release (if known):** N/A, this has always happened for me (several months since I set it up)

**Description of problem:**

I have several widgets on my home screen which I use to trigger scripts. They mostly work fine (displaying a green check mark when the action is complete), but sometimes, after the script is finished executing, the widget continues to spin indefinitely (>24 hours). Restarting the phone clears the condition.

**Screenshot of problem:**
Screenshot_20200705-125654

**Additional information:**
I wonder if it might be related to screen lock, since I have a habit of explicitly locking my screen with the power button when I'm finished with an action. But I have not investigated this.

dshokouhi:
## Summary

This is an attempt to fix #653 by adding a `callTimeout`

By default OKHTTP sets this timeout to `0` which means the call will forever try to complete.  I am not sure if 30 seconds is too long or if it should be 10 seconds however, I opted for 30 seconds to match the `readTimeout`.  There are other timeouts that we do not set like `writeTimeout` and `connectTimeout` that both default to 10 seconds.

https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/
https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/call-timeout/

I have also added some debug logs in case it does not fix the issue so we can at least look at the logs to get a better idea of why it keeps spinning.

Testing involved making a few different calls to HA server like loading App Configuration, using a button widget, using a static widget.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/button/ButtonWidget.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/button/ButtonWidget.kt
@@ -211,7 +211,9 @@ class ButtonWidget : AppWidgetProvider() {
                         }
                     }
 
+                    Log.d(TAG, "Sending service call to Home Assistant")
                     integrationUseCase.callService(domain, service, serviceDataMap)
+                    Log.d(TAG, "Service call sent successfully")
 
                     // If service call does not throw an exception, send positive feedback
                     feedbackColor = R.drawable.widget_button_background_green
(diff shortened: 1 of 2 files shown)


dshokouhi:
Interesting looks like lint failed because it could not load the dependencies going to close and reopen to retrigger the build

dshokouhi:
I dont believe this is caused by my PR so will wait for a review to see what went wrong.  I may try again later on today to see if it resolves itself

JBassett (approved):

Issue #1033, opened 2020-10-11
Notification when Home Assistant is unreachable


ellaizee:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like.**
If the app can't connect to the Home Assistant instance, send a notification. Smartthings did this (covered power and internet outages) and I just realized Home Assistant doesn't. It can be optional in the settings since some people wouldn't care, but I like to know if there's an issue at home and start a timer on my refrigerated items.

**Describe alternatives you've considered, if any**
None.

**Additional context**
None.

dshokouhi:
So just to clarify you want a notification posted on the device when it is unable to reach home assistant?  Like for example if you are restarting and it attempts to send an update?

ellaizee:
Preferred use case would be when there is a failure of some sort (equipment, power, internet) rather than restarts. Could build a timer into it and allow the user to select if they want to be notified after 5, 10, 15, etc minutes of downtime. I'm not sure how often the app reaches out for updates, but I did get a blank notification for a bit when my power was out yesterday, which I assumed was the app trying and failing to connect to my instance. Obviously, people with a UPS would have more of a delay, but those don't last forever.

dshokouhi:
the blank notification that shows up briefly and disappears will be fixed in the next production release, its currently fixed in the latest beta. The app reaches out every 15 minutes or when a sensor gets triggered. In the meantime look into uptime robot or down detector....those do the same thing and can work on any site.

poscreoqueera:
+1. I would like to receive a notification from the app when Home Assistant is not available. It's very useful for instance when there is a power outage, so that you can take whatever actions you need on that situation. Keep in mind that we may have notifications when motion is detected, doors are opened, etc. If we are not notified when there is a power outage, we may think none of these is happening... which could be a security risk.
Thanks!!

iulko:
+1 I opened duplicate witchout realizing it. The need for this option around the internet is pretty huge.

iulko:
@dshokouhi  is #1265 now possible to implement? Seems like the best option

dshokouhi:
> @dshokouhi is #1265 now possible to implement? Seems like the best option

the app architecture has completely changed since that attempt, a new approach would need to be considered with all the newly added features.

In the meantime I recommend using down detector or something similar to achieve the same result

iulko:
> > @dshokouhi is #1265 now possible to implement? Seems like the best option
> 
[...]

PR #1265, opened 2020-12-28
Initial commit of failed connection attempt notification

dshokouhi:
## Summary

Fixes: #1033 

This PR will keep track of failed connection attempts.  As of now I have chosen `5` as the maximum before pushing the notification however we can change this to a higher number.

For just about every call we make we either reset the count or add to the count for failed connection attempts.

The notification is persistent and has a chronometer to help the user understand how long its been since the app was able to reach HA.

Once a call is successful the notification will be removed without any intervention by the user.

I wanted to submit this now to get feedback from beta testers to find out how the functionality is and what improvements we need to make.

We will also be using a different notification channel so users can customize it to their liking.

I tried to test as much as I could by updating widgets, triggering sensor updates, launching HA all while my dev server was offline.  Testing was good on all fronts and have made adjustments to when the calls actually need to be made.  Sensor Receiver for example only makes the assumption if we are unable to update the sensor because we first register then update so only makes sense to make 1 call there.

## Screenshots

image


JBassett (changes requested):
If we are going to do something like this we need a better cross cutting location to do it.  I don't think it's a good idea to have the create / remove scattered across all the service calls like that.  It should really be a filter on all those calls instead.

dshokouhi:
How do you recommend we go about getting `context` in the `IntegrationRepositoryImpl.kt` file? Maybe its my misunderstanding of dagger but I didn't see how we can inject context like we can in other locations.  That would allow us to keep all the logic in the same file so we can post a notification.  I'd love for it to be as part of the retrofit but I didn't see a way to do that there either.

dshokouhi:
I think the only way to do this properly is to pass in `context` from all the calls that are made to HA so we can keep it all in one place `IntegrationRepositoryImpl.kt`.  Would that make sense for this or is there a cleaner way to go about it?

dshokouhi:
Made some updates to make everything more centralized.  WebView and the entity state widget are the only places I was unable to change the logic.  Webview needs its own logic for this as we rely on frontend messages as well as the cannot connect pop-up.  For the entity state widget its the only widget that doesn't have a service call and checking for null values won't really work here compared to the rest.  I can remove it if we think it should be removed but majority of logic is now kept in the integration repository.  Only passing in context so we can post the notification correctly.

dshokouhi:
Did some thorough testing and discovered sometimes the connection count gets reset and fails again causing a new notification that doesn't get cleared later on when connection is restored.  Will need to find out where that bug is but will keep this open to see if this is a desired path anyways.

dshokouhi:
Looks like I can reproduce by clearing app from recent apps screen so marking as draft again

dshokouhi:
Closing until the app has proper websocket support, as of now the current implementation is just not ideal and touches on too many files.

Issue #678, opened 2020-07-29
Try other URL if connection fails


gregg098:
**Is your feature request related to a problem? Please describe.**
This would be a general user experience enhancement for app connection, but the situation that got me here is described below:

In the app, I set the Home Assistant URL to my Nabu Casa address. Then I set the Home Network WiFi SSID to my home wifi name, and then the internal connection URL is http://[local ip]:8123/

This works fine, unless I use a VPN at home (Wireguard through a personal VPS and also Private Internet Access). Then it still tries to use the local address, even though all traffic is routed over a VPN, which makes it unable to connect.

**Describe the solution you'd like**
The error message when the app cannot connect has the options of Wait, Refresh, and Settings. It would be nice to have a "try other URL" or something like this. Idea being that if the remote URL doesn't work, try the local. If the local doesn't work, try the remote.

An alternative is it just tries the alternate URL and lets you know that the preferred connection failed (based on Wifi SSID?).

**Describe alternatives you've considered, if any**
Can update each VPN app to add routes, but this is cumbersome and not user friendly.

**Additional context**
None

PR #1287, opened 2021-01-09
Try the other URL when refreshing webview on a connection error

dshokouhi:
## Summary

Fixes: #678 by switching between internal and external URL when the user clicks on refresh.  If the user does not have an internal connection URL then we will default to the external one always.  Also updated the error message to let users to know that refreshing more than one does something 😄 

## Screenshots

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -917,8 +917,11 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
             alert.setPositiveButton(R.string.settings) { _, _ ->
                 startActivity(SettingsActivity.newInstance(this))
             }
+            val isInternal = runBlocking {
+                urlRepository.isInternal()
+            }
             alert.setNegativeButton(
-                if (failedConnection == "external")
+                if (failedConnection == "external" && isInternal)
                     R.string.refresh_internal
                 else
                     R.string.refresh_external
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -59,6 +59,7 @@ import io.homeassistant.companion.android.PresenterModule
 import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.authenticator.Authenticator
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
+import io.homeassistant.companion.android.common.data.url.UrlRepository
 import io.homeassistant.companion.android.database.AppDatabase
 import io.homeassistant.companion.android.database.authentication.Authentication
 import io.homeassistant.companion.android.nfc.NfcSetupActivity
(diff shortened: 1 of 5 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -796,7 +802,15 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
-                webView.reload()
+                runBlocking {
+                    failedConnection = if (failedConnection == "external") {
+                        webView.loadUrl(urlRepository.getUrl(true).toString())
+                        "internal"
+                    } else {
+                        webView.loadUrl(urlRepository.getUrl(false).toString())
+                        "external"
+                    }
+                }
```
JBassett, line 813: Rather than trying both urls we should make sure we are just trying to load the correct url instead.  You should just have to call `webView.loadUrl(urlRepository.getUrl().toString())` I believe.

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
JBassett: But the way this is currently implemented there is no way for a user to know which URL is going to be used.

dshokouhi: We can post a toast on the device saying which URL is going to be tried or I can update the text in the pop-up to let them know which one failed.  I think offering another button may confuse/complicate things since not everyone uses internal connection URL.

dshokouhi: Looks like the alert dialog does not allow to use variables in `setMessage` I think the current message might suffice since it tells users hitting refresh cycles through the URLs.  If we think a toast message would be good I can add that too, we already know the user is looking at the device :)

dshokouhi: ok we will now update the refresh button to reflect which URL will be attempted next :) the pop-up message is now updated to reflect this too.

image

image

dshokouhi: fixed the label logic so users who don't use internal URL will not see "refresh internal url" :)

JBassett (approved):

jmarfer:
Buenas noches,

En la versión de app mvl android de HA 2021.2.2-full no puedo entrar en configurar "ssid de la red Wifi de casa", de tal forma que me pone por defecto en URL interna ( en gris) la misma que la externa. Por este motivo no puedo acceder a HA desde mi mvl por conexión local.

Comprobada versión app mvl android de HA 3.0.2-full de mi hijo, que la tiene sin actualizar, si que deja configurar ssid.

Cómo podría solucionar este problema?

Muchas gracias de antemano

Un saludo!

dshokouhi:
> Buenas noches,
> 
> En la versión de app mvl android de HA 2021.2.2-full no puedo entrar en configurar "ssid de la red Wifi de casa", de tal forma que me pone por defecto en URL interna ( en gris) la misma que la externa. Por este motivo no puedo acceder a HA desde mi mvl por conexión local.
> 
> Comprobada versión app mvl android de HA 3.0.2-full de mi hijo, que la tiene sin actualizar, si que deja configurar ssid.
> 
> Cómo podría solucionar este problema?
> 
> Muchas gracias de antemano
> 
> Un saludo!

Please do not comment on merged PR also you need to type on english for us to help. Please open a new issue.

Issue #1278, opened 2021-01-05
Be able to send broadcast intent from other app to home assistant app


AlexKnowsIt:
**Is your feature request related to a problem? Please describe.**

I would like to integrate my smartwatch into home assistant. Both gadgetbridge as well as home assistant are now able to send now broadcast messages but neither of them is accepting one from another. 

**Describe the solution you'd like**

A way to wait for broadcast intents from other services. This could be made into a sensor where the properties of the broadcast intents are saved in. Like this it would be possible for example to send an broadcast to home assistant when I fall asleep and this could trigger turning of the lights

**Describe alternatives you've considered, if any**

Right now I thought about using tasker and let with their app wait for the signal from gadgetbridge and then create a notification that home assistant can grab but this is not as reliable as it could be

**Additional context**

Thanks already for all the cool features of the app, I find everyday something new to automate with it because of all the cool sensors that are now available!

dshokouhi:
So the android app can't just wait for any broadcast intent sent from any app.  What we can probably do is setup a broadcast intent that we listen to and other apps can send data to, but this means the other app has to update to support it.  We discussed having some type of intent API for other apps to integrate like Sleep as Android.  Sleep as Android has actually since supported webhooks and mqtt so all of those intents can be sent to directly to HA without needing to send data to our app.  This is something you can also ask the developers of gadgetbridge while we try to come up with a way to integrate other apps.  However honestly speaking I think if the app were to support sending data to HA natively then you take out one step in the process.

AlexKnowsIt:
Ah OK. Didn't know that. But gadgetbridge is sending to the nodomain domain, I think that's the reason I can use for example easer or tasker to receive the broadcast.
A broadcast would look like this
`nodomain.freeyourgadget.gadgetbridge.WokeUp`
Would this be possible for the home assistant companion app?

dshokouhi:
> A broadcast would look like this
> `nodomain.freeyourgadget.gadgetbridge.WokeUp`
> Would this be possible for the home assistant companion app?

We will not hard code another apps intents into our app.  More than likely whatever we build will require other apps to update to work with us.  Much like a Tasker plugin requires changes from the app.

AlexKnowsIt:
[...]

PR #1292, opened 2021-01-13
Allow users to register for intents and receive them as events

dshokouhi:
## Summary

Fixes: #1278 

For this request I have opted to tie the intents to "Last Update Trigger" as it makes the most sense for it to be a managed setting and all registered intents in our app get sent to the Sensor Receiver.  This allows the app to not have other intents hard coded while also allowing users to receive these intents as an event in Home Assistant.  Events make the most sense because they are not subject to the state changing logic that we have.

The settings follow the same behavior as the BSSID sensor where the user has to turn on a toggle to create a new `string` setting.  Once created the user will enter the broadcast intent string and save it.  Once the user has saved all the intents they wish to register for the application will need to be restarted.  I haven't found an easy way to trigger a restart when values have changed so I wanted to leave it up to the user for now.  The user can also clear out the setting to have it removed in the next update.

As of now the event will only contain the action string as most of the other data is usually null.  We can take on `extras` at a later point in time.

Example event data from Sleep as Android intents:

```
Event 3 fired 3:27 PM:
{
    "event_type": "android.intent_received",
    "data": {
        "intent": "com.urbandroid.sleep.alarmclock.SLEEP_TRACKING_STOPPED_AUTO",
        "device_id": "DEVICE_ID"
    },
    "origin": "REMOTE",
    "time_fired": "2021-01-12T23:27:25.757377+00:00",
    "context": {
        "id": "ID",
        "parent_id": null,
        "user_id": "USER_ID"
    }
}
Event 2 fired 3:27 PM:
{
    "event_type": "android.intent_received",
    "data": {
        "intent": "com.urbandroid.sleep.alarmclock.SLEEP_TRACKING_STARTED_AUTO",
        "device_id": "DEVICE_ID"
    },
    "origin": "REMOTE",
    "time_fired": "2021-01-12T23:27:18.458957+00:00",
    "context": {
        "id": "ID",
        "parent_id": null,
        "user_id": "USER_ID"
    }
}
```
## Screenshots

image

Documentation: home-assistant/companion.home-assistant#429


jawilson:
Not only is this my dream, this is the exact use case I have for it (Sleep as Android). Using Tasker to call the service API manually is just a pain, this would be such an amazing addition.

AlexKnowsIt:
This makes it possible to connect all the devices that gadgetbridge supports to home assistant and send triggers for falling asleep and waking up. This would be an amazing addition for automations!

JBassett (approved):

jawilson:
Excellent work! I'm so excited for this!

vincegre:
Hi

Sorry for intrusion but did you find a way to use it to get steps count from Gadgetbridge ? As far as I can see in all intents avalaible in GB, there is non that allows to get access at steps counted by the gadget linked at GB :(

Thanks

Vincèn

dreimer1986:
> Hi
> 
> Sorry for intrusion but did you find a way to use it to get steps count from Gadgetbridge ? As far as I can see in all intents avalaible in GB, there is non that allows to get access at steps counted by the gadget linked at GB :(
> 
> Thanks
> 
> Vincèn

Hey @vincegre 
I stumbled over the same flaw you found already. My way to get the data is... crude and evil, but it works.
https://github.com/dreimer1986/Mixed-HomeAssistant-stuff/tree/main Here you can find my two bigger projects that are not openepaperlink based. One of them is the ... special way to get the data you are looking for. My case is for my not even yet supported in nightly SR08 ring, but it can be modified to anything the app writes into it's database.
In theory it's just this: Send a intent to the GadgetBridge App to save the database to a defined in app path. In my case a SMB/Synology Drive share on my NAS and from there get the data off the database directly. Ugly, but works fine here.

PR #1316, opened 2021-01-24
Location disabled notification

### LINKED ISSUE #1311: Option to ignore "Location is disabled" question
The feature is related to https://github.com/home-assistant/android/pull/1187 (I commented there as well)

Since the update of the android app, the app keeps asking me to turn on location. It's annoying as it even happens when you quickly switch to another app.

I do not have my location on within android settings all the time. It's very specific choice i make. Before this pull request was merged, it meant that the sensors that were dependent on that didn't work - that was fine for me.

I use these sensors primarily to see the location of our family members. it's really practical as (when you want to) turn on GPS and then you can see on a map where the rest of the family is. I go for a run, which isn't that dangerous in itself, but it's a comforting feeling for my wife to know where i am. I can see when my wife is on here way back from wherever she was and start making food in time. 
Of course, i understand that tracking the location /requires/ GPS to be on, but i don't have GPS on all the time when i'm at home... or when i don't want to be tracked. If i'm away /with/ the family, then google doesn't have to know where i am.

The current dialog shows yes, or no.
I'd like to add "no, don't ask me again, i know that the sensors won't work when i have location turned off" (or of course, a little checkbox that says "don't ask again"
This preference would ideally be stored in a setting that you can later undo again.

Make a setting that would allow me to not perform the check.

**Additional context**
Overall, i'm a very happy HA user :-)

chriss158:
Fixes #1311 

## Summary

The message popup for disabled location can be quite annoying. With this PR the popup is replaced by a permanent notification message. 
Casual users will be still notified about the disabled location.
But power users can disabled the notification channel to ignore the warning.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
@@ -275,10 +275,16 @@ class SettingsFragment : PreferenceFragmentCompat(), SettingsView {
                 } else openSsidDialog()
             } else {
                 if (presenter.isSsidUsed()) {
-                    DisabledLocationHandler.showLocationDisabledWarnDialog(requireActivity(), arrayOf(getString(R.string.pref_connection_wifi)), true) {
-                        presenter.clearSsids()
-                        preference.setSsids(emptySet())
-                    }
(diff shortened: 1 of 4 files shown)


dshokouhi (approved):
Looks good to me.

JBassett (approved):

PR #1321, opened 2021-01-27
High accuracy mode for background location sensor

### LINKED ISSUE #973: High precision mode for location updates
I would like to obtain more accurate location updates for a short period of time, in order to trigger time- and location-critical events. For example, I would like to open my garage as soon I approach by bike.

I would like to have a high-precision location update mode that I could enable or disable through a special message from HA. In this mode, the app would send GPS-only location updates every 15 seconds. So, I would expect the following:

1. Send a notification to the app to start high precision mode

    ```yaml
    automations:
      - ...
        action:
          - service: notify.mobile_app
            data:
              message: start_location_tracking
    ```

2. The app would show a persistent notification "Home Assistant is tracking your location". A background service (?) continuously acquires GPS fixes every 15 seconds, and the existing location tracking mechanism sends these fixes to HA. Similar to fitness trackers.

3. Send a notification to the app to stop high precision mode

    ```yaml
    automations:
      - ...
        action:
          - service: notify.mobile_app
            data:
              message: stop_location_tracking
    ```
   Alternatively, it times out after e.g. 15 minutes.

* *Background location*: Location provided by background location is not exact enough, as it is mostly based on Wifi or cell towers in my case. Additionally, it updates at most once a minute, so it triggers too late for my use case, if at all.
* *Zone background / location*: Same as above, it does not update fast enough for my case.
* *`request_location_update`*: Is not exact enough, as it uses the same location mechanism as background location (i.e. Wifi or cell tower).
* *Other triggers like Wifi*: Triggering too late. And it also works for locations where there is Wifi reception. This would not work for locations outside home.

**Additional context**
In my use case, I would add a broader zone around my garage (e.g. 300m in diameter) that would start high precision tracking. When I enter a narrower zone (e.g. 50m in diameter), it would open my garage and stop high precision tracking.

When I have a fitness tracker running in parallel, HA also receives the location update intents. So in order to implement this, we would only need a background service and a notification, and the location update mechanism could stay as it is. This would also benefit other use cases like triggering geofence zones.

chriss158:
Fixes #973

PR for mobile-apps-fcm-push https://github.com/home-assistant/mobile-apps-fcm-push/pull/35

## Summary

To get accurate device tracking location results in home assistant you needed to use 3rd party apps like GPSLogger or Owntracks. To automate these apps (ex. Start GPSLogger when connected to car bluetooth) you needed also a 3rd party app like MacroDroid or Tasker. To top it off. Those apps are hard to configure for casual users.

So, to get accurate location results you need to install and configure two 3rd party apps for every smartphone in the household.

With this PR i tried to integrate those features into the companion app.

### High accuracy mode settings
Everything is controlled in the background location sensor settings. 
If you enable _High accuracy mode_ the high accuracy location service will be started with the interval settings defined in _High accuracy mode update interval_.
To inform the user that the app is using gps a ongoing notification will be shown.
To use the high accuracy mode, only when connected to a Bluetooth device, the user can choose devices with the _High accuracy mode only when connected to BT devices_. Then the high accuracy mode will only be used if connected to one of those defined devices. The _High accuracy mode_ needs to be enabled to use the other high accuracy options. So the _High accuracy mode_ switch acts like an master switch for the high accuracy mode.

### High accuracy mode notification
If the high accuracy mode is used, a notification will be shown. The notification shows the current address, if the geocoded sensor ist enabled. If the geocoded sensor is not enabled, the notification shows the current coordinates. Also the notification shows the current aprox. accuracy. 
The user can also disable the high accuracy mode right from the notification with the `Disable` button.

### Notification command for high accuracy mode
Turn off:
```
title: turn_off
message: command_high_accuracy_mode
```

Turn on:
```
title: turn_on
message: command_high_accuracy_mode
```

### Other changes

- On location update (high accuracy mode enabled or disabled) the geocoded sensor will be notified about the new location (if enabled)
- Refactored the `BluetoothSensorManager` to use the `BluetoothUtils` class. This is done because the `LocationSensorManager` also needs to know which bluetooth devices are connected/paired. So i unified this into one class.
- Every sensor setting is created with `isSingleLineTitle` to use longer title names in the settings which are not truncated.
- New value type for sensor settings added. `list-bluetooth` will list all bluetooth devices as a popup.

Documentation: home-assistant/companion.home-assistant#439

### DIFF
--- a/app/src/full/AndroidManifest.xml
+++ b/app/src/full/AndroidManifest.xml
@@ -31,6 +31,11 @@
             </intent-filter>
         </service>
 
+        <service
+            android:name=".location.HighAccuracyLocationService"
+            android:enabled="true"
+            android:exported="true"/>
+
         <meta-data
             android:name="com.google.firebase.messaging.default_notification_icon"
             android:resource="@drawable/ic_stat_ic_notification" />
--- a/app/src/full/AndroidManifest.xml
+++ b/app/src/full/AndroidManifest.xml
@@ -23,6 +23,10 @@
             android:enabled="true"
             android:exported="true" />
 
+        <receiver android:name=".location.HighAccuracyLocationReceiver"
+            android:enabled="true"
+            android:exported="true" />
+
         <service
             android:name=".notifications.MessagingService"
             android:exported="false">
--- a/app/src/full/java/io/homeassistant/companion/android/location/HighAccuracyLocationReceiver.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/location/HighAccuracyLocationReceiver.kt
@@ -3,8 +3,6 @@ package io.homeassistant.companion.android.location
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
-import io.homeassistant.companion.android.database.AppDatabase
-import io.homeassistant.companion.android.database.sensor.Setting
 import io.homeassistant.companion.android.sensors.LocationSensorManager
 
 class HighAccuracyLocationReceiver : BroadcastReceiver() {
@@ -18,8 +16,7 @@ class HighAccuracyLocationReceiver : BroadcastReceiver() {
             HIGH_ACCURACY_LOCATION_DISABLE ->
             {
                 HighAccuracyLocationService.stopService(context)
-                val sensorDao = AppDatabase.getInstance(context).sensorDao()
-                sensorDao.add(Setting(LocationSensorManager.backgroundLocation.id, LocationSensorManager.SETTING_HIGH_ACCURACY_MODE, "false", "toggle"))
+                LocationSensorManager.setHighAccuracyModeSetting(context, false)
             }
         }
     }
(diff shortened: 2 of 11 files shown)


dshokouhi:
This is pretty awesome :) Great job on the Bluetooth device names, love that idea!  Checked out the debug APK and works well from what I can see.  If need be to fully fix that linked bug if the users need more options we can always add a notification command with a sensor to represent the state so they can automate it from HA directly

chriss158:
You are right. It is pretty easy to add a device command to control the high accuracy mode from HA directly. My last commit added this feature. Sadly i couldn't test this one, because my push notifications aren't working anymore. I cannot get the mobile-apps-fcm-push to deploy. I need to investigate this. 
Maybe you can give it a try?

chriss158:
Needed to change the firebase to the Blaze plan. Now it is working and the device command to control the high accuracy mode are working too 👍 

PR for mobile-apps-fcm-push https://github.com/home-assistant/mobile-apps-fcm-push/pull/35

dshokouhi:
We need to make sure to add docs for this feature as well :) looks like we dont have a `docs-missing` label yet lol

chriss158:
Done 👍 
https://github.com/home-assistant/companion.home-assistant/pull/439

JBassett (approved):

Issue #1306, opened 2021-01-19
Add option to start Wireguard automatically when the home wifi SSID is not in use


antlarr:
**Is your feature request related to a problem? Please describe.**
When I leave home I need to start the Wireguard VPN on my phone to be able to connect to Home Assistant. The problem is sometimes I forget to start it so sensors/phone location are not updated in HA (so automations based on those sensors don't work) and push notifications do not reach my phone until I remember to turn on the Wireguard VPN.

**Describe the solution you'd like**
It would be nice if HA Companion started the Wireguard tunnel automatically when the internal url can't be reached, the home wifi SSID is not being used/detected, the location is outside the home zone, the external url is being used or any combination of those options you think it's appropriate (probably checking the SSID in use is the easiest case, might work good enough and I guess HA Companion already checks that so it wouldn't increase the power usage of the app).

To be more precise, some configuration options would need to be added:
* A "Start wireguard tunnel when outside home wifi" checkbox that enables all this.
* A "Wireguard tunnel name to start" option where the user can enter the name of the Wireguard tunnel to turn on.

Also, it might be worth adding a "Stop wireguard tunnel when home wifi detected" that disables the wireguard tunnel when the home wifi SSID is detected.

Just in case it helps to implement this feature, [this reddit comment](https://www.reddit.com/r/tasker/comments/fn6wou/wireguard_now_supports_control_by_remote_apps/fl93kw8/?utm_source=reddit&utm_medium=web2x&context=3)  explains how to turn on/off a wireguard tunnel from an external application using intents.

**Describe alternatives you've considered, if any**

A more generic alternative might be to provide configuration options so the user can define any intent to be sent to any application when the SSID is detected/not detected or when the phone goes in or out of a zone (an automation happening on the phone as there might be no connection to the HA core). In any case, just providing an option to start/stop a Wireguard tunnel might be enough for most users.

dshokouhi:
Sorry we will not be hard coding and adding in the permissions for wireguard like that link.  If they accept broadcast intents you can use [command broadcast](https://companion.home-assistant.io/docs/notifications/notification-commands#broadcast-intent)

antlarr:
> Sorry we will not be hard coding and adding in the permissions for wireguard like that link. If they accept broadcast intents you can use [command broadcast](https://companion.home-assistant.io/docs/notifications/notification-commands#broadcast-intent)

[...]

PR #1325, opened 2021-01-28
Allow extras with broadcast intent command

dshokouhi:
## Summary

Adds support for `extras` by allowing users to enter them under the `group` parameter.  We will split the extras by a comma `,` and then split the key/values by the colon `:`.  This will allow users to add/remove fields dynamically in an automation using templates.

Example to turn on a sleep as android alarm labeled as `work`:

```
message: command_broadcast_intent
title: "com.urbandroid.sleep.alarmclock.ALARM_STATE_CHANGE"
data:
  channel: "com.urbandroid.sleep"
  group: "alarm_label:work,alarm_enabled:true"
```

There is a potential for users to discover bugs with other applications here as well lol.

Documentation: home-assistant/companion.home-assistant#440

## Any other notes

One important thing to note is that all value are sent as a string from firebase so this PR does work around that.

https://github.com/home-assistant/mobile-apps-fcm-push/blob/master/functions/android.js#L53

Special thanks to @antlarr https://github.com/home-assistant/android/issues/1306#issuecomment-763004212 for the suggestion :)

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -355,6 +355,14 @@ class MessagingService : FirebaseMessagingService() {
                 try {
                     val packageName = data["channel"]
                     val intent = Intent(title)
+                    val extras = data["group"]
+                    if (!extras.isNullOrEmpty()) {
+                        val items = extras.split(',')
+                        for (item in items) {
+                            val pair = item.split(":")
+                            intent.putExtra(pair[0], pair[1])
+                        }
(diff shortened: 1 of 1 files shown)


JBassett (approved):

mdegat01:
Since this hasn't come out of beta yet, is it too late to change the field? Seems unnecessarily confusing that you use `tag` to send extras with activity intents and `group` to send extras with broadcast intents. Can it be adjusted so they both use the same field for those @dshokouhi ?

dshokouhi:
@mdegat01 good catch, I'll change that in the PR for the permission request since we need to do that anyways :)

Issue #1326, opened 2021-01-28
Notification command for google maps intent


mdegat01:
**Is your feature request related to a problem? Please describe.**

I have a number of shortcuts on my home screen right now built with Tasker that say things like this:
- Navigate to Car
- Navigate to Kylee by Car
- Navigate to Kylee by Bike
- ...
- 
They all are basically identical. What they do is this:
1. Make an API call to HA to get the state of a particular device tracker
2. Fire the [intent](https://developers.google.com/maps/documentation/urls/android-intents#launch-turn-by-turn-navigation) to launch google maps in navigation mode to that latitude and longitude with the specified mode

While this works, I think this would make a good notification command. I think a lot of other people would find it useful to be able to include a button in their UI or service call in their automations which sends directions to a person, place or device right to their phone. 

It could be specifically `command_launch_navigation" or could be more generic like `command_google_maps` with the ability for users to specify whatever intent string they wish. If you take a look at the [doc](https://developers.google.com/maps/documentation/urls/android-intents) you'll see there's a whole bunch of useful options that are all very similar:
- Open a map at a specific point
- Open streetview at a specific point
- Search for places around a specific point
- Etc.

For all of these you just send an Activity-type broadcast of type `android.intent.action.VIEW` to `com.google.android.apps.maps` with the data set to what you want to do.

**Describe the solution you'd like**

Would like the addition of a notification command like this:
```yaml
message: command_google_maps
title: google.navigation:q={{ location }}&mode={{ mode }}
```
Which then creates the activity intent and fires it with the data I provided.

**Describe alternatives you've considered, if any**

Could also make it much more generic like this
```yaml
message: command_activity_intent
title: google.navigation:q={{ location }}&mode={{ mode }}
data:
  channel: android.intent.action.VIEW
  group: com.google.android.apps.maps
```
This would say to fire an activity intent of type `VIEW` with the data specified in `title` to the package specified in `group`. That could then be used for other things besides just google maps.

**Additional context**

If it helps, this is the configuration I've been using to fire the intent from tasker
![Intent_1](https://user-images.githubusercontent.com/2037026/106093171-c8fdca00-60fd-11eb-9390-b574d4d6fbea.png) ![Intent_2](https://user-images.githubusercontent.com/2037026/106093179-cc915100-60fd-11eb-8250-582bc7ba4288.png)
I actually just realized I haven't been specifying the package as `com.google.android.apps.maps` even though it says to do so in the documentation, interesting it works without it.

Issue #1148, opened 2020-11-06
Action button to open a video in the media folder


engrbm87:
**Is your feature request related to a problem? Please describe.**
With the release of the Media browser feature, is there was to directly open a video file from the notification using an action button? This would be useful in case recordings from cameras are stored in the media folder.

**Describe the solution you'd like**
The option to configure an action button that opens a video file stored in the media folder directly on the mobile.

**Describe alternatives you've considered, if any**

**Additional context**

dshokouhi:
So I don't know if this is exactly what you are looking for but this should now be doable using existing features and the new `command_activity` notification command

https://companion.home-assistant.io/docs/notifications/notification-commands#activity

Try to use this command following this guide from google to see if it will launch a video, may need some trial and error and research to get it right.

https://developer.android.com/guide/components/intents-common#Music

PR #1327, opened 2021-01-28
Add a notification command to launch an activity

dshokouhi:
## Summary

Fixes: #1326 

A new command `command_activity` is added to handle this specific intent as we need to do several things to properly launch an activity.  The `title` must contain the URI to send to the app.  The `channel` must be used to set the package name of where the intent needs to be sent to. The `group` must be set the intending action string. We handle generating the URI and launching the activity with the intent.

Example:

```
message: command_activity
title: "geo:0,0?q=1600+Amphitheatre+Parkway%2C+CA"
data:
  channel: "com.google.android.apps.maps"
  group: "android.intent.action.VIEW"
```

```
message: command_activity
title: "google.navigation:q=arbys"
data:
  channel: "com.google.android.apps.maps"
  group: "android.intent.action.VIEW"
```

Documentation: home-assistant/companion.home-assistant#442

## Any other notes

FCM PR: https://github.com/home-assistant/mobile-apps-fcm-push/pull/36

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -237,7 +237,7 @@ class MessagingService : FirebaseMessagingService() {
                             }
                         }
                         COMMAND_ACTIVITY -> {
-                            if (!it[TITLE].isNullOrEmpty() && !it["channel"].isNullOrEmpty())
+                            if (!it[TITLE].isNullOrEmpty() && !it["channel"].isNullOrEmpty() && !it["group"].isNullOrEmpty())
                                 handleDeviceCommands(it)
                             else {
                                 mainScope.launch {
@@ -446,8 +446,9 @@ class MessagingService : FirebaseMessagingService() {
             COMMAND_ACTIVITY -> {
                 try {
                     val packageName = data["channel"]
+                    val action = data["group"]
                     val mapsIntentUri = Uri.parse(title)
-                    val mapIntent = Intent(Intent.ACTION_VIEW, mapsIntentUri)
+                    val mapIntent = Intent(action, mapsIntentUri)
                     mapIntent.setPackage(packageName)
                     mapIntent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
                     startActivity(mapIntent)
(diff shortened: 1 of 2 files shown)


mdegat01:
Nice! Yea that is clearly a lot easier to use then my suggestions haha.

dshokouhi:
Looks like the checks got stuck, going to close and re-open

app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -393,6 +411,29 @@ class MessagingService : FirebaseMessagingService() {
...
+                    var mode = data["channel"]
+                    val validModes = listOf("d", "b", "l", "w")
+                    if (mode !in validModes)
+                        mode = "d"
+                    val navigate = "google.navigation:q=$title&mode=$mode"
+                    val mapsIntentUri = Uri.parse(navigate)
+                    val mapIntent = Intent(Intent.ACTION_VIEW, mapsIntentUri)
+                    mapIntent.setPackage("com.google.android.apps.maps")
+                    mapIntent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
+                    startActivity(mapIntent)
```
JBassett, line 425: I'm not sure we want to hard link to google navigation.  Maybe use the common intent?  I know it might not be identical but this way if someone doesn't use Google Maps this will still work for them.
https://developer.android.com/guide/components/intents-common#ViewMap

JBassett (changes requested):

app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
dshokouhi: Hmm ok so a special command that starts an activity with `Intent.ACTION_VIEW` preset?

dshokouhi:
Per: https://developer.android.com/guide/components/intents-common#Music this might also fix: #1148 as a temporary workaround until we set something up similar for actionable notifications.

app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
dshokouhi: ok think i got it generic enough now lol

JBassett (approved):

Issue #960, opened 2020-09-19
Elite ibeacon compatible wij esphome


rdehuyss:
Please add the feature to emit ibeacons.

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I'm always frustrated when presence detection is not working and triggering the alarm.

**Describe the solution you'd like**
Emit ibeacons from the app so that can be used as presence detection by esphome. Automatic room_presence detection as esphome devices are located in areas

**Describe alternatives you've considered, if any**
Other apps like beacon simulator

**Additional context**

Alfiegerner:
There are a couple of Android apps that do this, e.g. Beacon Simulator and Beaconscope, but neither persist with Bluetooth or phone restarts. 

Beacon simulator source: https://github.com/vhiribarren/beacon-simulator-android

Beaconscope transmitter source: 
https://github.com/davidgyoung/ningo-android/blob/master/app/src/main/java/com/davidgyoungtech/beaconscanner/BeaconTransmitter.java

dshokouhi:
> I'm thinking this could sit as another Bluetooth sensor

Agreed, this way it will persist between restarts as it will be part of our foreground service.

> and an MVP initial build I was thinking that most details could be defaulted:
> 
> * beacon type is defualted to iBeacon
> * UUID, major and minor all defaulted
> * Power and frequency defaulted

MVP is always a good idea, once things get reviewed and merged with the bare minimum its easier to add features later on :)


> * Force Bluetooth on.  when sensors are refreshed, check if bluetooth, if off start bluetooth and restart transmitting.  For those using beacon functionality for home and room presence this would be extremely useful

Not sure how much this makes sense given that sensors update every 15 minutes and anytime any of the enabled sensors triggers an update.  Some users may have very frequent updates getting triggered so behavior could be sporadic there when you have all sensors enabled.  One thing that the full version of the app offers is a command to control bluetooth on the device remotely so could be useful to use that :)



> Is that worth tackling?

I say definitely and we are here to help, you can find the team members on discord #devs_mobile_apps

We can always use more contributers!!!

> Or would feature parity be a consideration?

[...]

PR #1343, opened 2021-02-02
Add BLE ibeacon transmitter as sensor to allow room presence.

Alfiegerner:
## Summary

Added BLE emitting sensor.  This sends out iBeacon messages over bluetooth every second, so that room presence tracking can be accomplished using roomassistant or esp32-mqtt-room.  See issue #960

## Screenshots

Screenshot_20210203-004725_Home Assistant
Screenshot_20210203-004918_Home Assistant

Documentation: home-assistant/companion.home-assistant#449
Documentation: home-assistant/companion.home-assistant#451 - for updated docs for new Command Notification, command_ble_transmitter

## Any other notes

For the corresponding iphone changes, a good reference point would be app produced over at roomassistant: https://github.com/mKeRix/room-assistant/issues/309


app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
```diff
@@ -60,35 +76,67 @@ class BluetoothSensorManager : SensorManager {
...
+            bleTransmitterDevice.id1 = id1
+            bleTransmitterDevice.id2 = id2
+            bleTransmitterDevice.id3 = id3
         }
-        val icon = if (isBtOn) "mdi:bluetooth" else "mdi:bluetooth-off"
+        return result
+    }
 
+    private fun updateBLEtrasnmitter(context: Context) {
+        if (!isEnabled(context, bleTransmitter.id) && bleTransmitterDevice.transmitting) // sensor has been turned off, stop transmitting
```
dshokouhi, line 127: After we stop what we need to stop we also need to `return` so none of the other logic gets executed since the sensor is disabled.

app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
```diff
@@ -60,35 +76,67 @@ class BluetoothSensorManager : SensorManager {
...
+        val icon = if (bleTransmitterDevice.transmitting) "mdi:bluetooth" else "mdi:bluetooth-off"
         onSensorUpdated(
-            context,
-            bluetoothState,
-            isBtOn,
-            icon,
-            mapOf()
+                context,
+                bleTransmitter,
+                bleTransmitterDevice.state,
```
dshokouhi, line 135: There appears to be 3 different states that can be sent? how does that work for this `binary_sensor`?  or is this supposed to be a normal `sensor` :)

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/TransmitterManager.kt
```diff
@@ -0,0 +1,81 @@
...
+    private lateinit var physicalTransmitter: BeaconTransmitter
+    private lateinit var beacon: Beacon
+
+    private fun buildBeacon(haTransmitter: BLETransmitter): Beacon {
+        val builder = Beacon.Builder()
+        builder.setTxPower(-59)
+        builder.setId1(haTransmitter.id1)
+        builder.setId2(haTransmitter.id2)
+        builder.setId3(haTransmitter.id3)
+        builder.setManufacturer(0x004c) //  apple ibeacon
```
dshokouhi, line 20: Does this mean we are hard coded only to ibeacon?

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/TransmitterManager.kt
```diff
@@ -0,0 +1,81 @@
...
+    @Synchronized
+    fun startTransmitting(context: Context, haTransmitter: BLETransmitter) {
+        if (!shouldStartTransmitting(haTransmitter)) {
+            return
+        }
+        if (idsHaveChanged(haTransmitter) && this::physicalTransmitter.isInitialized) {
+            stopTransmitting(haTransmitter)
+        }
+        if (!this::physicalTransmitter.isInitialized) {
+            val parser = BeaconParser().setBeaconLayout("m:2-3=0215,i:4-19,i:20-21,i:22-23,p:24-24") // ibeacon only
```
dshokouhi, line 42: Should we only do ibeacon? What about the others?

dshokouhi (changes requested):
Some of these requests may just be a question as I am not overly familiar with how this works.  There are some things we need to address as well.

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/TransmitterManager.kt
Alfiegerner: Yes, for now. We can expand to other formats if required later, but i think for most use cases one beacon type is fine, and iBeacon appears to be the most popular.

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/TransmitterManager.kt
Alfiegerner: See my other comment on this, iBeacon only for now, can revisit in later pull requests if needed.

app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
```diff
@@ -60,35 +76,67 @@ class BluetoothSensorManager : SensorManager {
...
-        if (checkPermission(context, bluetoothState.id)) {
-            isBtOn = BluetoothUtils.isOn(context)
+    private fun updatedBLEDevice(context: Context): Boolean {
+        var result = false
+        var id1 = getSetting(context, bleTransmitter, BLE_ID1, "string", DEFAULT_BLE_ID1)
+        var id2 = getSetting(context, bleTransmitter, BLE_ID2, "string", DEFAULT_BLE_ID2)
+        var id3 = getSetting(context, bleTransmitter, BLE_ID3, "string", DEFAULT_BLE_ID3)
+        if (bleTransmitterDevice.id1 != id1 || bleTransmitterDevice.id2 != id2 || bleTransmitterDevice.id3 != id3) {
+            result = true
+            bleTransmitterDevice.id1 = id1
```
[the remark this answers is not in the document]
Alfiegerner: The ids form the unique identifier for the iBeacon. Used for client apps like room assistant for tracking, whitelisting by entire id or partial masks, e.g. Major.  I've used the property name id* to follow the abstract beacon convention.  For iBeacon only use we could rename them to Uuid, major, minor.

app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
Alfiegerner: Thanks, will fix this up.

app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
Alfiegerner: At the moment there are 3 states, Transmitting, ErrorTransmitting and Stopped. Stoppped should only happened if BT is disabled wheb we try and transmit.

I'll change stopped state words to indicate BT not enabled.

In future merge I'd like to add a listener to BT adapter turning on and off, to maintain transmitter state and start when re-enabled.

app/src/main/res/values/strings.xml
```diff
@@ -298,6 +299,7 @@ like to connect to:</string>
   <string name="sensor_description_battery_level">The current battery level of the device</string>
   <string name="sensor_description_battery_state">The current charging state of the battery</string>
   <string name="sensor_description_bluetooth_connection">Information about currently connected bluetooth devices</string>
+  <string name="sensor_description_bluetooth_ble_emitter">Send BLE iBeacon message every second, used to track presence  around house, e.g. together with roomassistant or esp32-mqtt-home projects.</string>
```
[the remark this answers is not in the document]
Alfiegerner: Will add, thanks.

app/src/main/res/values/strings.xml
```diff
@@ -420,6 +422,9 @@ like to connect to:</string>
   <string name="state_locked">Locked</string>
   <string name="state_off">Off</string>
   <string name="state_on">On</string>
+  <string name="state_transmitting">Transmitting</string>
```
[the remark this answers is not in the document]
Alfiegerner: I think this is a stray line, will remove. I tried adding states using R.strings but then saw the other sensors using hardcoded English, so changed tack, but missed deleting this.

app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
dshokouhi: Ok if we wish to remain as a `binary_sensor` then we should report `true` for the state if we are `transmitting` only, otherwise if you want the state to say `transmitting` and the others we will need to change the sensor type to `sensor`

https://github.com/home-assistant/android/blob/4300a9bc5ec3a2c86fc9ca2520e74d34a27bee29/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt#L35

Alfiegerner: Gah, sorry, missed that, will change to sensor.

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
Alfiegerner: Removed

app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
```diff
@@ -433,6 +446,12 @@ class MessagingService : FirebaseMessagingService() {
                 if (title == TURN_ON)
                     bluetoothAdapter.enable()
             }
+            COMMAND_BLE_TRANSMITTER -> {
+                if (title == TURN_OFF)
+                    BluetoothSensorManager.enableDisableBLETransmitter(applicationContext, false)
+                if (title == TURN_ON)
+                    BluetoothSensorManager.enableDisableBLETransmitter(applicationContext, true)
+            }
```
Alfiegerner, line 454: Hi @dshokouhi Apologies I couldn't resist the urge to add something else :) adding a Command Notification to allow people to turn of Ble Transmitter as required from HA.  I think this might away with future requirements I was considering to add a listener to re-enable transmitter on bluetooth (re) start, as we can orchestrate that from HA. It also adds other possibilities as well, like turning off Transmitter to save battery when not required.

The code looks pretty straight forward and I've added a separate commit to MessagingService.kt, in case we want to remove it from this PR. Also a separate PR to docs project under home-assistant/companion.home-assistant#451 .

The only snag that I've got is that I haven't been able to test it - I've been looking at  getting firebase set up with https://github.com/home-assistant/mobile-apps-fcm-push but haven't succeeded so far (running into script running permissions issue that I can't solve as don't have full access).  I was wondering if we'd consider adding to a beta as is, or if you could suggest any  other way of working round the problem. 

Thanks,

dshokouhi:
testing out the latest debug APK on my pixel 4 and I get the following error when I turn on the sensor:

```
2021-02-03 21:49:57.269 8695-8830/io.homeassistant.companion.android.debug E/BeaconTransmitter: Cannot start advertising due to exception
    java.lang.NullPointerException: Attempt to invoke virtual method 'void android.bluetooth.le.BluetoothLeAdvertiser.startAdvertising(android.bluetooth.le.AdvertiseSettings, android.bluetooth.le.AdvertiseData, android.bluetooth.le.AdvertiseCallback)' on a null object reference
        at org.altbeacon.beacon.BeaconTransmitter.startAdvertising(BeaconTransmitter.java:213)
        at org.altbeacon.beacon.BeaconTransmitter.startAdvertising(BeaconTransmitter.java:159)
        at io.homeassistant.companion.android.bluetooth.ble.TransmitterManager.startTransmitting(TransmitterManager.kt:51)
        at io.homeassistant.companion.android.sensors.BluetoothSensorManager.updateBLEtransmitter(BluetoothSensorManager.kt:151)
        at io.homeassistant.companion.android.sensors.BluetoothSensorManager.requestSensorUpdate(BluetoothSensorManager.kt:75)
        at io.homeassistant.companion.android.sensors.SensorReceiver.updateSensors(SensorReceiver.kt:150)
        at io.homeassistant.companion.android.sensors.SensorWorker$doWork$2.invokeSuspend(SensorWorker.kt:80)
```

this is what the sensor screen stays at as well.

image

Alfiegerner:
> testing out the latest debug APK on my pixel 4 and I get the following error when I turn on the sensor:
> 
> ```
> 2021-02-03 21:49:57.269 8695-8830/io.homeassistant.companion.android.debug E/BeaconTransmitter: Cannot start advertising due to exception
>     java.lang.NullPointerException: Attempt to invoke virtual method 'void android.bluetooth.le.BluetoothLeAdvertiser.startAdvertising(android.bluetooth.le.AdvertiseSettings, android.bluetooth.le.AdvertiseData, android.bluetooth.le.AdvertiseCallback)' on a null object reference
>         at org.altbeacon.beacon.BeaconTransmitter.startAdvertising(BeaconTransmitter.java:213)
>         at org.altbeacon.beacon.BeaconTransmitter.startAdvertising(BeaconTransmitter.java:159)
>         at io.homeassistant.companion.android.bluetooth.ble.TransmitterManager.startTransmitting(TransmitterManager.kt:51)
>         at io.homeassistant.companion.android.sensors.BluetoothSensorManager.updateBLEtransmitter(BluetoothSensorManager.kt:151)
>         at io.homeassistant.companion.android.sensors.BluetoothSensorManager.requestSensorUpdate(BluetoothSensorManager.kt:75)
>         at io.homeassistant.companion.android.sensors.SensorReceiver.updateSensors(SensorReceiver.kt:150)
>         at io.homeassistant.companion.android.sensors.SensorWorker$doWork$2.invokeSuspend(SensorWorker.kt:80)
>         at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
>         at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
>         at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
>         at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
>         at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
>         at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
> ```
> 
> this is what the sensor screen stays at as well.
> 
> image

Looking now, just tried reinstalling latest apk from scratch and enabling / disabling and worked fine on my end.

dshokouhi:
Looks like I just needed to restart my phone lol, I got an ID now so I'll check out room assistant to test it out for the first time :)

Alfiegerner:
I was able to repro that issue with BT turned off, have just committed a fix to make sure we don't transmit if BT is off and add that info as a state message.

dshokouhi:
So I was able to play around with feature last night.  One thing that occurred to me was that it seemed to give a decent hit on the battery life for keeping the service up for 1 hour.  In the one hour test I did it drained as much battery as the production version that was there all day.  So makes me wonder if we should do a few things:

*  Tie the service to a sensor setting so users do not "Enable all sensors" and then wonder why they start to have excessive battery drain.  Once tied to a setting we don't have to worry about this and the command can continue to work as normal.  It kinda defeats the purpose of enable all here and I can see a lot of users hitting this issue in that case. 
* Add a warning in the docs about high battery drain and also in sensor description
* Make the transmitter a persistent foreground service so users know its active and to prevent it from ever getting killed by the system.

These 3 features came to mind since we use the same precautions for the new High Accuracy Mode setting that we have.

Any thoughts on this?

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/TransmitterManager.kt
```diff
@@ -0,0 +1,84 @@
+package io.homeassistant.companion.android.bluetooth.ble
+
+import android.bluetooth.BluetoothAdapter
+import android.bluetooth.le.AdvertiseCallback
+import android.bluetooth.le.AdvertiseSettings
+import android.content.Context
+import org.altbeacon.beacon.Beacon
+import org.altbeacon.beacon.BeaconParser
+import org.altbeacon.beacon.BeaconTransmitter
+object TransmitterManager {
```
JBassett, line 10: Should this be injected like the `ThemeManager`?

Alfiegerner:
Happy to look at these. Also could implement a transmission strength setting (Low, Medium, 
High) which could save battery too.

I'm not 100% sure on the UX of having two multiple foregrounded services, would that mean having 2 persistent notifications?

JBassett (changes requested):
@dshokouhi I agree we should make the description obvious that this is going to drain the battery.  I like the idea of a foreground service but I think this is ok for a first pass.  I'm not sure about the enable all sensors issue, part of me wonders how many people are actually using it at this point.  If you enable them all there are a lot of sensors!

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconTransmitter.kt
```diff
@@ -0,0 +1,11 @@
+package io.homeassistant.companion.android.bluetooth.ble
+
+data class IBeaconTransmitter(
```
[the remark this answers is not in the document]
Alfiegerner: Ah - this is specifically an iBeaconTransmitter class rather than an interface.  Can change anyway though as agree is confusing (some premature optimisation going anticipating a move to multiple beacon types).

dshokouhi:
> @dshokouhi I agree we should make the description obvious that this is going to drain the battery. I like the idea of a foreground service but I think this is ok for a first pass. I'm not sure about the enable all sensors issue, part of me wonders how many people are actually using it at this point. If you enable them all there are a lot of sensors!

LOL I know I'm not the only crazy one 😛 

My thoughts here were that currently when you enable all sensors (like I currently do) the battery drain is very minimal.  I only see 1-3% usage in a full day for all sensors enabled.  If a user enables them all on accident they will have no way of knowing without disabling each sensor one by one or realizing its a new feature where the culprit is.   By keeping it in a setting users can keep all sensors enabled (if they wish) and turn on the setting via notification when they want.  Some may have some OCD due to the enable all toggle lol.  I am ok either way but I have a feeling we will see a few bug reports on it.

Alfiegerner:
@dshokouhi @JBassett , How about:
 - i add battery warning to description / docs
 - I add the setting as per High accuracy mode, to not be part of 'enable all'.
 - I don't implement the additional foreground service, but leave as is in sensor service.

And we can revisit the service if it's getting killed. Sound okay?

JBassett:
@Alfiegerner sounds good.

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconTransmitter.kt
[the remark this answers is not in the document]
Alfiegerner: Okay, I've decided to leave this as IBeacon for now, to make it explicit that's what it is (and also as the library I'm using already has a BeaconTrasnmitter class)- can change if easily enough if the implication of an interface is too strong.

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/TransmitterManager.kt
Alfiegerner: Not sure if this is required - let me know.

Alfiegerner:
Hi @dshokouhi @JBassett,

I've added the following:
  - Added a setting that allows users to specify power output of BLE transmitter, from Ultra Low, Low, Medium and High
  - Added a setting that by default prevents the sensor being turned of when the Enable all sensors is toggled.  User can turn this off if they want.
  - Added warnings on battery drain to docs and description

I haven't done the refactor to inject TransmitterManager, I had a quick stab at this but couldn't get it working - but I can revisit that if you'd like.

Thanks!

p.s. there's a Changes Requested fag open but I can't see what it relates too, can you let me know if I've missed anything?  Thanks.

Alfiegerner:
Hi @JBassett @dshokouhi  - just checking anything else  you need from me on this?

Thanks!

JBassett (approved):

Issue #569, opened 2020-04-26
Implement ability to launch intents from app


klejejs:
**Home Assistant Android version:**
1.8.0-159
**Android version:**
10.0
**Phone model:**
Pixel 4
**Home Assistant version:**
0.107
**Last working Home Assistant release (if known):**

**Description of problem:**
Currently clicking on a URL that has intent inside it crashes the app. It would be nice if it opened another app in the intent.
**Traceback (if applicable):**

```

```

**Screenshot of problem:**

**Additional information:**
The intent URL I am using: [intent://qremote/#Intent;component=com.qnap.qremote/com.qnap.qremote.SplashActivity;end](intent://qremote/#Intent;component=com.qnap.qremote/com.qnap.qremote.SplashActivity;end)

klejejs:
I am sorry, but I think you you misunderstood the issue. I don't think it is related to notifications. The problem is related to "weblink" entities in Lovelace dashboard.

dshokouhi:
@klejejs that would then be a frontend request and not for the android app.  You are referring to a lovelace card.  The app just displays the card.  Please open a request in the frontend repo.

klejejs:
Frontend supports it except it opens Google Play Store, not the app itself. For opening another app, intents should be supported in the android app, not frontend. Also, the app straight up crashes, when you click on that link, it doesn't even open Play Store, so I think this is a valid issue and should be reopened.

klejejs:
I have added the intent URL to the original issue post and can confirm, that this is working in [Kiosk Browser](https://play.google.com/store/apps/details?id=de.ozerov.fully)

klejejs:
I am not very familiar with Kotlin, so I am not sure how to fix this, but here are the crash logs.
As far as I understand, the app should support the handling of "intent:" url's.

```
ActivityTaskManager: START u0 {act=android.intent.action.VIEW dat=intent://qremote/} from uid 10157
Shutting down VM
FATAL EXCEPTION: main
Process: io.homeassistant.companion.android, PID: 16658
android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW dat=intent://qremote/ }
       at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:2058)
       at android.app.Instrumentation.execStartActivity(Instrumentation.java:1716)
       at android.app.Activity.startActivityForResult(Activity.java:5210)
[...]

PR #1344, opened 2021-02-03
Override certain URLs to launch an app or intent scheme

dshokouhi:
## Summary

Fixes: #569 

We now accept `app://` and `intent://` prefix.  When testing in a desktop a blank tab opens so nothing harmful appears to happen elsewhere.

When we detect that a link starts with `app://` we will grab the android package and attempt to launch it, if the package does not exist we take the user to the play store (app or website)

When we detect the link starts with `intent://` we will parse the Uri with the intent scheme and attempt to launch the activity, if the app is not found we will also take the user to the play store (app or website)

Examples using lovelace weblink entity card type:

This example will launch Twitter if it is installed on the device, otherwise it will open the Google Play store app or website.
```yaml
- type: weblink
  name: Twitter
  url: "app://com.twitter.android"
```

This example will launch the barcode scanning app ready to scan via the Intent scheme, if the app is not installed the user will be directed to installing it.
```yaml
- type: weblink
  name: Scan
  url: "intent://scan/#Intent;scheme=zxing;package=com.google.zxing.client.android;end"
```

## Screenshots

video: https://photos.app.goo.gl/Q93ppHcNB7u8djY97

Documentation: home-assistant/companion.home-assistant#450

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -80,6 +80,9 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
         private const val AUDIO_REQUEST_CODE = 42
         private const val NFC_COMPLETE = 1
         private const val FILE_CHOOSER_RESULT_CODE = 15
+        private const val APP_PREFIX = "app://"
+        private const val INTENT_PREFIX = "intent://"
+        private const val MARKET_PREFIX = "https://play.google.com/store/apps/details?id="
 
         fun newInstance(context: Context, path: String? = null): Intent {
             return Intent(context, WebViewActivity::class.java).apply {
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -240,10 +243,51 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
+                                        Uri.parse(MARKET_PREFIX + intent.`package`.toString())
+                                    startActivity(marketIntent)
+                                }
+                                return true
+                            } else if (!webView.url.toString().contains(it.toString())) {
+                                val browserIntent = Intent(Intent.ACTION_VIEW, it)
+                                startActivity(browserIntent)
+                                return true
+                            } else
+                                Log.w(TAG, "No unique cases found to override")
```
JBassett, line 288: Not sure this justifies a warning, should probably just be debug.

```suggestion
                                Log.d(TAG, "No unique cases found to override")
```

JBassett (approved):

PR #1347, opened 2021-02-04
Fix cancel notification if last notification in group

### LINKED ISSUE #1346: 4 notifications can cause clearing sticky/persistant to fail
2021.1.1-full

**Android version:**
11

**Phone model:**
Pixel 5

**Home Assistant version:**
2021.1.5

None that I know of

**Description of problem:**
If there are more than three sticky and persistant notifications, and the last three sent are cleared before the first one sent. The last one cleared will fail to clear.

**Steps to reproduce:**

1. Send a stick/persistant notification to the phone, I did this through Developer Tools, could be done through a script pretty easily though.
  ```
  service: notify.mobile_app_alex_phone
  data:
      message: test this will not clear
      data:
        persistent: true
        tag: test0
        sticky: true
  ```
2. Send 3 more notifications to the phone without clearing any of them.
  ```
  service: notify.mobile_app_alex_phone
  data:
      message: test this will clear
      data:
        tag: test1
  ---
  service: notify.mobile_app_alex_phone
  data:
      message: test this will also clear
      data:
        tag: test2
  ---
  service: notify.mobile_app_alex_phone
  data:
      message: test this too shall clear
      data:
        tag: test3
  ```
3. Swipe clear the last 3 notifications you sent
4. Clear the persistent notification.
  ```
  service: notify.mobile_app_alex_phone
  data:
      message: clear_notification
      data:
        tag: test0
  ```

_expected behavior_: The first notification clears

_actual behavior_: The notification does not clear.

**Workaround:** Force closing the app from the android settings page for it clears the notification.

If you feel I need to get logs, I should be able to get that sorted.

**Screenshots:**
Screenshot_20210204-002851

**Additional information:**

I did some testing that suggests when the persistent message shows up doesn't matter so much as long as there are 4 messages.

One thought on this, I noticed that 4 is the break point at which notifications are rolled up into channels, so it might be related to that?

Might be related to #1223 or #1238 ? But If I understand correctly those fixes were released in 2021.1.1

chriss158:
Fixes #1346

## Summary

If a notification is the last one in a group, which is generated by the system and not the user, the notification cannot be canceled.
This is because the group which should be canceled has no groupId.
To fix this issue, we just cancel the last notification in the group. That way, the system removes the group by itself.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/util/NotificationManagerExtensions.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/util/NotificationManagerExtensions.kt
@@ -53,11 +53,15 @@ fun NotificationManagerCompat.cancelGroupIfNeeded(tag: String?, id: Int): Boolea
             if (isGroupSummary && groupNotifications.size == 1 ||
                 !isGroupSummary && groupNotifications.size == 2) {
                 val group = groupNotifications[0].notification.group
-                if (group != null) {
-                    val groupId = group.hashCode()
+
+                // If group is null, the group is a group which is generate by the system.
+                // This group can't be canceled, but it will be canceled by canceling the last notification inside of the group
+                // If the group isn't null, cancel the group
+                return if (group != null) {
+                    var groupId = group.hashCode()
                     this.cancel(group, groupId)
-                }
-                return true
+                    true
+                } else false
             }
         }
     }


dshokouhi (approved):
Thanks for the quick fix!!

JBassett (approved):

PR #1350, opened 2021-02-05
Allow users to add reply button to actionable notification

dshokouhi:
## Summary

When an `action` is set to `REPLY` we will create a reply button in the action.  The user can then enter any text and receive the reply back in the same event that we send for `mobile_app_notification_action`.  It made the most sense to attach this to our actions because android only allows [3 actions](https://developer.android.com/training/notify-user/build-notification#Actions) and reply can be one of them.  Given that a reply is an actionable button and we have to get the data back to HA might as well send it back in the event.

Example service call data:

```
message: reply test
data:
  actions:
  - action: REPLY
    title: reply
```

Example event data:

```
{
    "event_type": "mobile_app_notification_action",
    "data": {
        "message": "reply test",
        "action_1_title": "reply",
        "action_1_key": "REPLY",
        "action": "REPLY",
        "reply_text": "This is a reply",
        "device_id": "DEV_ID"
    },
    "origin": "REMOTE",
    "time_fired": "2021-02-05T05:08:08.883377+00:00",
    "context": {
        "id": "ID",
        "parent_id": null,
        "user_id": "USER_ID"
    }
}
```

## Screenshots

image

image

image

Documentation: home-assistant/companion.home-assistant#453

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
@@ -7,6 +7,7 @@ import android.os.Handler
 import android.util.Log
 import android.widget.Toast
 import androidx.core.app.NotificationManagerCompat
+import androidx.core.app.RemoteInput
 import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
 import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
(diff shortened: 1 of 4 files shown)


app/src/full/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
```diff
@@ -63,8 +64,13 @@ class NotificationActionReceiver : BroadcastReceiver() {
                 Toast.makeText(context, R.string.event_error, Toast.LENGTH_LONG).show()
             }
         }
+        val replyText = if (notificationAction.key == "REPLY")
+            RemoteInput.getResultsFromIntent(intent).getCharSequence(MessagingService.KEY_TEXT_REPLY).toString()
+        else
+            "no reply"
```
JBassett, line 70: Rather than having `reply_text` in every action event we should probably only add the `reply_text` if there is an actual reply.  You can probably just add it to the `notificationAction.data` here rather than adding to the signature of `fireEvent`.

JBassett (changes requested):

app/src/full/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt
dshokouhi: much better thank you :)

JBassett (approved):

mdegat01:
Love this! I didn't even think about this, that's brilliant. Have to start thinking about how to leverage this in my automations 🤔

Issue #1091, opened 2020-10-22
Add support for on-device haptics


agneevX:
**Home Assistant Android version:** `2.5.0-full`
**Android version:** 11
**Phone model:** Google Pixel 4a
**Home Assistant version:** 0.115.3

**Description of ~problem~ feature request:**

~My phone supports haptics natively, but I get vibrations when tapping on buttons in the front-end.~

Google Pixel devices have haptics, similar to iPhone devices, and is measured in `ms`.
Instead of a single form of vibration, haptics provide subtle but differential feedback during interactions with the frontend such as tapping buttons or opening menus.

Per Apple's Human Interface Guidelines, the following options are available for haptics and is currently implemented by devs, so they're maximum compatible.

https://companion.home-assistant.io/docs/integrations/haptics#developers-integrating-haptics-into-custom-cards

<img width="500" alt="image" src="https://user-images.githubusercontent.com/19761269/96914012-0846ca00-14c2-11eb-817a-9e5dea25ac00.png">

**Additional information:**
N/A

dshokouhi:
The frontend is what handles this as they control the Vibrate API on a website.  We don't have control here as this is a user profile setting which needs to be respected.  Also it may even be proper haptic just a longer and more powerful vibration.  You should see the same behavior in Chrome right?

agneevX:
> We don't have control here as this is a user profile setting which needs to be respected. 

In case of iOS, IIRC, different levels of haptic were added to the app, per Apple's HIG.

Surely the app can pickup the frontend code e.g. `haptic: light` and apply that instead of the standard vibration. 

> Also it may even be proper haptic just a longer and more powerful vibration. 

It's shorter and much less powerful and varies in intensity (measured in `ms`).

At the moment, it seems there's no support for haptics from the Android app.

dshokouhi:
I tried taking a look at this.  Android does not really support the constants defined by Apple.  Only a handful exist and thats only available to users on the latest version API 30.

https://developer.android.com/reference/android/view/HapticFeedbackConstants

I don't see apple defining how many `ms` each of the above is mentioned.  I think this feature was designed specifically for iOS so we may need to map our `ms` response for users who cant use teh constants up above.  Additionally it will not be as easy to map them to iOS and as of now the external bus sends back everything we see in the link so we can't really change that.

Some reference to use for older devices: https://stackoverflow.com/questions/25748230/confused-about-haptic-feedback-in-android

[...]

PR #1352, opened 2021-02-05
Add haptic/vibration support from the external bus in webview

dshokouhi:
## Summary

Fixes: #1091 by making a best effort to match the existing constants, vibration effects to what it sounds like for iOS.  For a first release these may be ok? We can always adjust as we see fit since we now handle each `hapticType` sent.  It was a bit difficult to map some of these 🙈 lol

https://developer.android.com/reference/android/view/HapticFeedbackConstants#constants_1
https://developer.android.com/reference/android/os/VibrationEffect.html#constants_1
https://developers.home-assistant.io/docs/frontend/external-bus/#trigger-haptic-haptic

Documentation: home-assistant/companion.home-assistant#456

## Any other notes

Tested with: https://github.com/custom-cards/button-card

Example lovelace config
```
      - type: "custom:button-card"
        color_type: card
        color: rgb(223, 255, 97)
        icon: mdi:volume-minus
        tap_action:
          action: call-service
          service: light.toggle
          service_data:
            entity_id: light.corner_lamp
          haptic: "success"
```

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -14,10 +14,13 @@ import android.os.Build
 import android.os.Bundle
 import android.os.Handler
 import android.os.Looper
+import android.os.VibrationEffect
+import android.os.Vibrator
 import android.text.method.HideReturnsTransformationMethod
 import android.text.method.PasswordTransformationMethod
 import android.util.Log
 import android.util.Rational
+import android.view.HapticFeedbackConstants
 import android.view.View
 import android.webkit.CookieManager
 import android.webkit.HttpAuthHandler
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -457,6 +460,47 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
+                                    "heavy" -> {
+                                        webView.performHapticFeedback(HapticFeedbackConstants.LONG_PRESS)
+                                    }
+                                    "selection" -> {
+                                        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R)
+                                            webView.performHapticFeedback(HapticFeedbackConstants.GESTURE_START)
+                                        else
+                                            vm.vibrate(50)
+                                    }
+                                }
```
JBassett, line 502: This should be extracted to a function rather than doing it inline.

JBassett (changes requested):

JBassett (approved):

PR #1356, opened 2021-02-09
Webview gesture quickbar

dshokouhi:
## Summary

Detects 3 finger swipe down gesture only on the webview activity launch the quickbar by sending the keycode for `e`.  When the quickbar is open the swipes will flip between the 2 types of quickbar views.

## Screenshots

https://photos.app.goo.gl/39KVvKNP6upAcPP47

Documentation: home-assistant/companion.home-assistant#459

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -184,7 +184,6 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
         webView.apply {
             setOnTouchListener { _, motionEvent ->
                 if (motionEvent.pointerCount == 3 && motionEvent.action == MotionEvent.ACTION_POINTER_3_DOWN) {
-                    Log.d(TAG, "3 fingers down")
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -180,7 +182,10 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
 
         webView = findViewById(R.id.webview)
         webView.apply {
-            setOnTouchListener { _, _ ->
+            setOnTouchListener { _, motionEvent ->
+                if (motionEvent.pointerCount == 3 && motionEvent.action == MotionEvent.ACTION_POINTER_3_DOWN) {
+                    dispatchKeyEvent(KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_E))
```
JBassett, line 187: Rather than faking the key presses we should probably send the javascript event on the bus.

dshokouhi: Is there an actual event for this?I didn't see it when I looked

JBassett:
Since I don't think there is an event we can fire to cause the entity bar to open for now this will work.

PR #1366, opened 2021-02-17
Fix left notification when high accuracy mode was disabled

chriss158:
## Summary

Again i had the high accuracy notification  still visible, even though the high accuracy mode was disabled due to not connected BT devices. This happens not every time and i can't reproduce this.

This PR hopefully fixes the problem.

Some points which may fix this issue
* Make notification objects static
* Use ServiceCompat for stopForeground
* Additionally call cancel to remove notification

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/location/HighAccuracyLocationService.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/location/HighAccuracyLocationService.kt
@@ -16,6 +16,7 @@ import android.os.IBinder
 import android.util.Log
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
+import androidx.core.app.ServiceCompat
 import androidx.core.content.ContextCompat
 import com.google.android.gms.location.FusedLocationProviderClient
 import com.google.android.gms.location.LocationRequest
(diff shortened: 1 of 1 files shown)


app/src/full/java/io/homeassistant/companion/android/location/HighAccuracyLocationService.kt
```diff
@@ -159,26 +161,41 @@ class HighAccuracyLocationService : Service() {
...
 
     override fun onDestroy() {
         fusedLocationProviderClient.removeLocationUpdates(getLocationUpdateIntent())
-        stopForeground(true)
+        ServiceCompat.stopForeground(this, ServiceCompat.STOP_FOREGROUND_REMOVE)
         isRunning = false
         Log.d(TAG, "High accuracy location service stopped")
 
+        // Remove notification again. Sometimes stopForeground(true) is not enough. Just to be sure
+        notificationManagerCompat.cancel(notificationId)
```
dshokouhi, line 197: I noticed I had to add similar logic in #1265 we just always have to cancel.  I think its due to the app being silently killed in the background.  I found a good way to test it is once you have the notification posted to remove the app from the recent apps list which isn't like a force stop but causes the variables to reset.  I think thats the case and reason for just always removing at the end.

JBassett (approved):

app/src/full/java/io/homeassistant/companion/android/location/HighAccuracyLocationService.kt
chriss158: Thanks for help 👍  but also with force close app i couldn't reproduce the issue. It is weird. I can only  see this issue when i'm using this in my car. And also not every time :/ Most of the time it does work.

PR #1397, opened 2021-03-05
Pull translations at build time

JBassett:
## Summary

I'm tired of remembering to sync up the translations when we build a release.  This will make the translation files be pulled at build time so they are always up to date.

N/A


.github/actions/download-translations/action.yml
```diff
@@ -0,0 +1,37 @@
+name: 'Create Release Numbers'
+description: 'Creates the current release number based on checked out code'
```
dshokouhi, line 2: Only thing that stood out to me, copy and paste? :)

JBassett: I would never.... ;-)

PR #1406, opened 2021-03-06
Fix BLE Transmitter Power setting - bug was causing settings to be ignored and default of High be used 

Alfiegerner:
## Summary

Fix a silly edit bug, Power level for BLE transmitter was never being used, and defaulted to high.
## Screenshots

n/a

n/a


Alfiegerner:
Actually -  just noticed this is being set correctly a few lines above, so this looks like a benign bug.  I'll close this PR and add it to other fixes.

PR #1415, opened 2021-03-08
Enable high accuracy mode when entering zone

chriss158:
## Summary

This PR adds the possibility to enable the high accuracy mode, when entering a specific zone (Requires an enabled location zone sensor).
In addition, the user can specify a radius in meters that extends the selected zone by the specified radius. This way the zone is entered X meters before the real zone.

To accomplish that, an internal expanded zone from the original zone is created (ex. `zone.home_expanded`) which adds the defined radius to the original radius. If the radius (Default 0) is 0, then the original zone is used.

If the user decides to use an expanded zone, ONLY the expanded zone, minus the original zone will be used (The blue zone in the image below).
**Example:**
Arriving zone.home
User enters `zone.home_expanded` (1000m radius around `zone.home`) -> High accuracy mode enabled
User enters `zone.home` -> High accuracy mode disabled

Leaving zone.home
User exits `zone.home`, therefore enters `zone.home_expanded` -> High accuracy mode enabled
User exits `zone.home_expanded` -> High accuracy mode disabled

Zone

Documentation: https://github.com/home-assistant/companion.home-assistant/pull/472

## Any other notes

Additional changes:

- A sensor setting can now be enabled/disabled which is reflected in the settings screen as a greyed out preference
- Refactored "creating a list setting" into a method `createListPreference`. Now `list-bluetooth`, `list-zones` and `list-apps` uses this method.
- Fire HA Event for every Zone Entered/Exited and not only for the first one

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
@@ -19,25 +19,29 @@ import androidx.preference.contains
 import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.bluetooth.BluetoothUtils
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
+import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
 import io.homeassistant.companion.android.database.AppDatabase
 import io.homeassistant.companion.android.database.sensor.Sensor
 import io.homeassistant.companion.android.database.sensor.SensorDao
 import io.homeassistant.companion.android.database.sensor.Setting
 import io.homeassistant.companion.android.util.DisabledLocationHandler
 import io.homeassistant.companion.android.util.LocationPermissionInfoHandler
+import kotlinx.coroutines.runBlocking
 
 class SensorDetailFragment(
     private val sensorManager: SensorManager,
-    private val basicSensor: SensorManager.BasicSensor
+    private val basicSensor: SensorManager.BasicSensor,
+    private val integrationUseCase: IntegrationRepository
 ) :
         PreferenceFragmentCompat() {
 
     companion object {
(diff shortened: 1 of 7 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
```diff
@@ -267,61 +275,23 @@ class SensorDetailFragment(
...
-                                            basicSensor.id,
-                                            setting.name,
-                                            newValue.toString().replace("[", "").replace("]", ""),
-                                            "list-bluetooth"
-                                    )
-                            )
-                            sensorManager.requestSensorUpdate(requireContext())
-                            return@setOnPreferenceChangeListener true
+                        val pref = createListPreference(key, setting, sensorDao, btDevices)
+                        if (!it.contains(pref))
```
nick2525, line 286: "If the condition of an if or when statement is multiline, always use curly braces around the body of the statement." https://kotlinlang.org/docs/coding-conventions.html#control-flow-statements  Kotlin coding-conventions

chriss158: Thanks for the tip. But do I miss something here? The if isn't multiline. Personally I would prefer the if as an one liner or at least with braces around. I would change this one to a one liner. 
But if the Kotlin conventions recommend a different way, I would adapt my preferences to that.

JBassett (approved):
By no fault of your own the location manager class has gotten unwieldy.  It's in desperate need of cleanup / refactor to more manageable pieces.  What PRs we have open for location will probably be the last approved until someone has some time to clean it up.

dshokouhi:
@JBassett I will create a feature request so we can keep an eye on it to improve, I agree its gotten hard to follow over time with all the fixes and enhancement :)

Issue #1417, opened 2021-03-09
App title has disappeared 


Sebclem:
**Home Assistant Android version:** beta-606-46b0b887-full

**Android version:** 9

**Phone model:** Galaxy S8 (SM-G950F - DreamLTE)

**Home Assistant version:** 2021.3.2

**Last working Home Assistant release (if known):**

**Description of problem:** 
The app title have disappeared, i don't know for how many times, i just have noticed this morning that i don't have title anymore.
I have checked in add detail menu, same issue, no title.
I have tried to reinstall the app, but same issue. (Uninstall via adb with -k  option to keep app data)

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```

**Screenshot of problem:**
![image](https://user-images.githubusercontent.com/23737709/110451766-d3c85900-80c4-11eb-8e41-27ac41363a0a.png)
![image](https://user-images.githubusercontent.com/23737709/110451781-d925a380-80c4-11eb-8505-cb6b224bb629.png)

**Additional information:**

chrzr:
Same issue here

Home Assistant Android version: beta-606-46b0b87-full

Android version: 11

Phone model: Google Pixel 4a 5G

dshokouhi:
What language do you guys have set on your device and in app configuration?  This issue is strange, I am on a later beta version and do not see it.  Both my device and app are set to English.

chrzr:
I have my device set to German. I have just tried the latest beta release from 28 minutes ago, but the title still does not show up.
I'll try setting the device to English. Maybe this will fix it. 🤔

dshokouhi:
ok thanks for reporting, can you confirm if the rest of the translations are present in the app when German is selected?

chrzr:
Seems like the german translations inside the app are fine. Only the title is affected.

![Screenshot_20210309-234924](https://user-images.githubusercontent.com/17109454/110549313-2be97480-8132-11eb-941b-ca816ce250bd.png)

dshokouhi:
Thanks for confirming @chrzr 

I found the issue, the app name on Lokalize was removed 

![image](https://user-images.githubusercontent.com/1634145/110549398-d8056e00-80e6-11eb-97cb-aab4d753efd0.png)

Seems to be the case for all languages which is odd as I thought we had a guard up for this.

@JBassett do you know how to easily restore the string `app_name` from Lokalize?

[...]

PR #1419, opened 2021-03-10
Don't use empty translations.

JBassett:
## Summary

Fixes: #1417 

## Screenshots

N/A

N/A


dshokouhi (approved):
looks to be a nice setting to use lol

side note: you would expect android to have certain protections in place so the actual app name is never blank lol

PR #1426, opened 2021-03-11
Improve location tracking on non GPS devices

chriss158:
## Summary

Based on following exception, i discovered some inconsistencies with location tracking and devices with no GPS.
```java
android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.settings.LOCATION_SOURCE_SETTINGS flg=0x50800000 }
    at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:2007)
    at android.app.Instrumentation.execStartActivity(Instrumentation.java:1673)
    at android.app.ContextImpl.startActivity(ContextImpl.java:953)
    at android.app.ContextImpl.startActivity(ContextImpl.java:924)
    at android.content.ContextWrapper.startActivity(ContextWrapper.java:379)
    at io.homeassistant.companion.android.util.DisabledLocationHandler$showLocationDisabledWarnDialog$1.onClick(DisabledLocationHandler.kt:101)
    at androidx.appcompat.app.AlertController$ButtonHandler.handleMessage(AlertController.java:167)
    at android.os.Handler.dispatchMessage(Handler.java:106)
    at android.os.Looper.loop(Looper.java:193)
    at android.app.ActivityThread.main(ActivityThread.java:6740)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```
I can't exactly say, what this causes, but as a solution, we just open the settings activity instead of the location settings activity, if the location settings activiy is not found.
First assumption was, that this was an amazon device. But i tried this on my amazon device without problems. Anyway this fix should work for that exception.

While testing on my amazon tablet i discovered the location tracking inconsistencies on devices with no GPS. This PR should fix these problems too.

Solved bugs:
* Fixed crash enabling location from DisabledLocationHandler on devices with no LOCATION_SOURCE_SETTINGS activity
* Disable single accurate location sensor on non GPS devices
* Fixed enabled location tracking after onboarding even if location tracking was disabled
* Open settings, if location settings can't be opend (ex. Custom-ROM without GAPPS, or Devices with no LOCATION_SOURCE_SETTINGS activity)

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
@@ -624,7 +624,7 @@ class LocationSensorManager : BroadcastReceiver(), SensorManager {
 
     override val availableSensors: List<SensorManager.BasicSensor>
         get() {
-            return if(DisabledLocationHandler.hasGPS(latestContext)) {
+            return if (DisabledLocationHandler.hasGPS(latestContext)) {
                 listOf(singleAccurateLocation, backgroundLocation, zoneLocation)
             } else {
                 listOf(backgroundLocation, zoneLocation)
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
@@ -299,12 +299,6 @@ class LocationSensorManager : BroadcastReceiver(), SensorManager {
         )
     }
 
-    override fun hasSensor(context: Context, sensorId: String): Boolean {
-        return if (sensorId == singleAccurateLocation.id && !DisabledLocationHandler.hasGPS(context)) {
-            false
-        } else super.hasSensor(context, sensorId)
-    }
-
     private suspend fun requestZoneUpdates() {
         if (!checkPermission(latestContext, zoneLocation.id)) {
             Log.w(TAG, "Not registering for zone based updates because of permissions.")
@@ -629,7 +623,13 @@ class LocationSensorManager : BroadcastReceiver(), SensorManager {
         get() = R.string.sensor_name_location
 
     override val availableSensors: List<SensorManager.BasicSensor>
-        get() = listOf(singleAccurateLocation, backgroundLocation, zoneLocation)
+        get() {
+            return if(DisabledLocationHandler.hasGPS(latestContext)) {
+                listOf(singleAccurateLocation, backgroundLocation, zoneLocation)
+            } else {
+                listOf(backgroundLocation, zoneLocation)
+            }
+        }
 
     override fun requiredPermissions(sensorId: String): Array<String> {
         return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
@@ -11,7 +11,9 @@ import android.os.PowerManager
 import android.util.Log
 import android.widget.Toast
 import androidx.core.content.ContextCompat.getSystemService
+import com.google.android.gms.location.FusedLocationProviderClient
 import com.google.android.gms.location.Geofence
+import com.google.android.gms.location.GeofencingClient
 import com.google.android.gms.location.GeofencingEvent
 import com.google.android.gms.location.GeofencingRequest
 import com.google.android.gms.location.LocationCallback
@@ -21,8 +23,10 @@ import com.google.android.gms.location.LocationServices
 import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.bluetooth.BluetoothUtils
 import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
+import io.homeassistant.companion.android.common.data.integration.Entity
 import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
 import io.homeassistant.companion.android.common.data.integration.UpdateLocation
+import io.homeassistant.companion.android.common.data.integration.ZoneAttributes
 import io.homeassistant.companion.android.database.AppDatabase
 import io.homeassistant.companion.android.database.sensor.Attribute
 import io.homeassistant.companion.android.database.sensor.Setting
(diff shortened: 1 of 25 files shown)


app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -614,7 +623,7 @@ class LocationSensorManager : BroadcastReceiver(), SensorManager {
     }
 
     override val enabledByDefault: Boolean
-        get() = true
+        get() = false
```
dshokouhi, line 626: What was the reason for this change again? I think we kept this as `true` to ensure onboarding works.  Although onboarding has changed since the introduction of `hasSensor`.

Anyways if we need it to be `false` then we can remove the method as it defaults to `false`

chriss158: Good point. I think i need to tweak that one.

Right now, the location tracking switch defaults to "Location enabled" in the android settings.
https://github.com/home-assistant/android/blob/f38ec600e3cba15ebadec3e474f73410c78203c8/app/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationFragment.kt#L75

If the location tracking switch is disabled and the user doesn't change the switch. location tracking will be disabled, but the sensor will be left enabled.
If the location tracking switch is enabled and the user doesn't change the switch, location tracking will be enabled.

With this change, i only switched this logic :dizzy_face:

We need to make sure, that the sensors will set accordingly to the default of location tracking option

dshokouhi: just remember a few things about this.  the `device_tracker` entity will always show up, the sensors themselves arent really sensors just house the logic for updating.  if location permission is not granted there won't be any location updates and the sensors should be disabled by default then.  I think?

chriss158: Yes, i would leave them disabled by default, if there is no additional reason behind this.

if i call `setLocationTracking(it.isChecked)` after
https://github.com/home-assistant/android/blob/f38ec600e3cba15ebadec3e474f73410c78203c8/app/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationFragment.kt#L75

Then this should be ok, i think.
Location tracking then should be disabled/enabled accordingly to the android location setting.

app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -296,6 +299,12 @@ class LocationSensorManager : BroadcastReceiver(), SensorManager {
         )
     }
 
+    override fun hasSensor(context: Context, sensorId: String): Boolean {
+        return if (sensorId == singleAccurateLocation.id && !DisabledLocationHandler.hasGPS(context)) {
```
dshokouhi, line 303: So is the goal to only hide one sensor here? Because to do that you just need to modify <https://github.com/home-assistant/android/blob/5a2e0476951ddb4b330661e7aeb5265f0ccf4215/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt#L632> to return the sensors to be shown.

chriss158: Yes i want to remove the sensor. 

Other sensors does also use the `hasSensor` method to check if a sensor is available. The only problem was, that the `LocationSensoManager` has multiple sensors. That's why, i included the `sensorId` parameter to `hasSensor`.

If i change the logic to the `availableSensors` property, i would do it differently on the `LocationSensorManager` as on the other sensors. I like booth solutions. What solution do you prefer?

dshokouhi: `availableSensors` is designed to show which sensors are available meaning there will always be one so the manager is shown all the time.  `hasSensor` was created to hide the entire manager completely so it ignores `availableSensors` like in the case of LightSensorManager as it only has 1 sensor we want to hide the entire block.  So one hides the parent the other only the children.

I personally like to keep it separate because sometimes a device will support some sensors but not all like if they are on API24 and the manager has sensors in API23 and API 26 we can hide 26 with just the list and keep the manager shown with the others.

chriss158: Thanks for clarification. Now i understood the concept behind `hasSensor`. Then i agree fully with you 💯

dshokouhi (changes requested):
Just tested the debug APK and from a fresh install, when I select location tracking during onboarding the location sensors are not enabled when they all should be enabled.  So if a user checks that toggle when they change the device name before pushing finish, all location sensors that can be enabled should be enabled upon selecting finish.  If you change `enabledByDefault` back to `true` it should take care of this.

image

chriss158:
Thanks for testing. The if condition was wrong.
Did also a complete test on the simulated Pixel 4 and on the Amazon Tablet (Without GPS)
- Test onboarding with default location tracking -> Disabled on Pixel 4 (Because missing Permissions)/Enabled on Amazon (Because old Android. No need for permission handling) 👍 
- Test onboarding with enabling/disabling locating tracking -> All 3 sensors are disabled/enabled  correctly 👍

app/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationFragment.kt
```diff
@@ -171,32 +170,39 @@ class MobileAppIntegrationFragment : Fragment(), MobileAppIntegrationView {
...
         arrayOf(
             LocationSensorManager.backgroundLocation,
             LocationSensorManager.zoneLocation,
             LocationSensorManager.singleAccurateLocation
         ).forEach { basicSensor ->
             var sensorEntity = sensorDao.get(basicSensor.id)
+            var sensorEnabled = enabled
+            if (basicSensor.id == LocationSensorManager.singleAccurateLocation.id && !DisabledLocationHandler.hasGPS(requireContext())) {
+                sensorEnabled = false
+            }
```
JBassett, line 198: Do we need this here if we aren't even showing the sensor when they don't have GPS?

JBassett:
Also need to fix merge conflicts.

app/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationFragment.kt
chriss158: Thanks for the review 👍  No, not really. I just wanted to keep the database consistent with the display. I will remove it.

chriss158:
Merged master and and resolved code review

JBassett (approved):

PR #1430, opened 2021-03-11
Add search/filter feature to the "Manage Sensor" screen

chriss158:
## Summary

This PR adds the possibility to hide disabled sensors.

## Any other notes

Additional changes:
* Made "Disabled" sensor summary translatable
* Show "Enabled" as sensor summary if sensor is enabled but has no state

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
@@ -7,7 +7,11 @@ import android.os.Bundle
 import android.os.Handler
 import android.os.Looper
 import android.provider.Settings
+import android.view.Menu
+import android.view.MenuItem
+import android.widget.SearchView
 import androidx.core.app.NotificationManagerCompat
+import androidx.core.view.MenuItemCompat
 import androidx.preference.Preference
 import androidx.preference.PreferenceCategory
 import androidx.preference.PreferenceFragmentCompat
@@ -87,6 +91,7 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
         private var permissionsAllGranted = true
         private var settingsWithLocation = mutableListOf<String>()
         private var enableAllSensors = false
+        private var showOnlyEnabledSensors = false
 
         fun newInstance(): SensorsSettingsFragment {
             return SensorsSettingsFragment()
@@ -102,22 +107,7 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
 
         setPreferencesFromResource(R.xml.sensors, rootKey)
 
-        findPreference<SwitchPreference>("show_only_enabled_sensors")?.let {
-
-            it.summary = getShowOnlyEnabledSensorsSummary(getShowOnlyEnabledSensors())
-
-            it.setOnPreferenceChangeListener { _, newState ->
-                val newShowOnlyEnabledSensors = newState as Boolean
-
-                getShowOnlyEnabledSensors(newShowOnlyEnabledSensors)
-
-                it.summary = getShowOnlyEnabledSensorsSummary(newShowOnlyEnabledSensors)
-
-                showHideSensorsIfNeeded(newShowOnlyEnabledSensors)
-
-                return@setOnPreferenceChangeListener true
-            }
-        }
+        showOnlyEnabledSensors = getShowOnlyEnabledSensors()
 
         findPreference<SwitchPreference>("enable_disable_sensors")?.let {
 
@@ -178,10 +168,9 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
             }
         }
 
-        val showOnlyEnabledSensors = getShowOnlyEnabledSensors()
         SensorReceiver.MANAGERS.sortedBy { getString(it.name) }.filter { it.hasSensor(requireContext()) }.forEach { manager ->
             val prefCategory = PreferenceCategory(preferenceScreen.context)
-            prefCategory!!.title = getString(manager.name)
+            prefCategory.title = getString(manager.name)
             preferenceScreen.addPreference(prefCategory)
 
             manager.availableSensors.sortedBy { getString(it.name) }.forEach { basicSensor ->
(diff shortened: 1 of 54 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -153,6 +155,7 @@ to your home internet.</string>
   <string name="enable_remaining_sensors">Enable %1$d Sensors</string>
   <string name="enabled_summary">When enabled values will be sent to Home Assistant</string>
   <string name="enabled_title">Enabled</string>
+  <string name="enabled">Enabled</string>
```
dshokouhi, line 158: should we just stick to using `enabled_title` since the string is the same?

dshokouhi (approved):
Looks good to me, tested debug apk with a few sensors.  Very minor comment from me since the string seems to be the same.

app/src/main/res/values/strings.xml
chriss158: When i think about that now, i would use only one. If the strings need to be different, than we can still change this.

dshokouhi (approved):

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
```diff
@@ -96,6 +102,23 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
 
         setPreferencesFromResource(R.xml.sensors, rootKey)
 
+        findPreference<SwitchPreference>("show_only_enabled_sensors")?.let {
+
+            it.summary = getShowOnlyEnabledSensorsSummary(getShowOnlyEnabledSensors())
```
JBassett, line 107: This can be done in the XML via this: https://developer.android.com/reference/androidx/preference/SwitchPreference#setSwitchTextOff(int)

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
```diff
@@ -184,10 +207,75 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
...
+            integrationUseCase.setShowOnlyEnabledSensors(showOnlyEnabledSensors)
+        }
+    }
+    private fun getShowOnlyEnabledSensors(): Boolean {
+        var showOnlyEnabledSensors = false
+        runBlocking {
+            showOnlyEnabledSensors = integrationUseCase.getShowOnlyEnabledSensors()
+        }
+        return showOnlyEnabledSensors
+    }
```
JBassett, line 225: I don't think we should store this as a setting.  This is just really a filter and should default to off but not be stored.

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
chriss158: Thanks i used:  `android:summaryOff` /  `android:summaryOn`

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
chriss158: I thought of this more like a setting rater than a temp. filter. My use case is, that i configure my sensors only one time (after app installation) and then work (check sensors states/reconfigure enabled sensors) from time to time with the sensors. In this use case it is really convenient to just see the enabled sensors, every time i enter the manage sensor screen.

Nevertheless, I like the temp. filter idea. But then i would not use a `SwitchPreference` to filter things. 
I imagine that in the upper right corner there is a filter symbol and a search symbol to find certain sensors.
But to realize this, i need to rework this PR. I think this is the better approach.

dshokouhi: I was under the impression that preferences are always stored and remember what the user selected/typed.  When I check notification history and change the filter the next time I come back it remembers it.  Even after I force stop the app and reopen.

chriss158: That's interesting. I tried it, without storing a setting, and the setting was reset every time i entered the manage sensor screen. I need to have a look at this again.

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
```diff
@@ -184,10 +207,75 @@ class SensorsSettingsFragment : PreferenceFragmentCompat() {
...
+            getString(R.string.only_enabled_sensors_shown)
+        } else {
+            getString(R.string.all_sensors_shown)
+        }
+    }
+
+    private fun showHideGroupsIfNeeded() {
+        for (pref in preferenceScreen) {
+            val prefGroup = pref as PreferenceGroup
+            if (prefGroup != null) {
```
nick2525, line 263: As you cast as "pref as PreferenceGroup" it can't be null, It will be exception if pref is not PreferenceGroup, if you want nullable, cast as "pref as? PreferenceGroup"

chriss158: Thanks 👍 Changed that

chriss158:
I reworked the PR. Mainly replaced the preference "Show enabled sensors only" by an actionbar item and added a search functionality in the actionbar.

app/src/main/java/io/homeassistant/companion/android/sensors/SensorsSettingsFragment.kt
chriss158: Ok i reworked the PR a bit. (See post below) But i still think, its better to save the filters. I think some users would complain about the reset filter. For me it feels not right, when the filter gets reset every time.

@dshokouhi You were right. The preference save it's value automatically. But, now that i use the action bar menu items, i need to use the storage.

dshokouhi:
I think the filter and search bar make it look clean :) great idea there!

JBassett (changes requested):
Still not sold on storing the value but we can beta it.
Need to fix merge conflicts then good to merge.

chriss158:
No problem. Let's remove it and see if there are popping up issues. If yes, then we can still implement a setting 👍

chriss158:
I think i misunderstood you. 
Storing a setting is only needed when the setting needs to be restored after the app was force closed. Right? That is not what i want.
I wanted that the setting is hold as long as the app runs. If the app was force closed or killed by the system, it's okay that the setting will be reset.

The `showOnlyEnabledSensors` is now a static class variable. It's value is hold as long the app is in the memory. Am i right here? Till now i thought the value is only hold while the `Manage Sensors` screen is open and will be reset after leaving the screen. But that is  not the case 👍

JBassett (approved):
This was exactly what I was thinking!

PR #1431, opened 2021-03-12
Fix statusbar color

### LINKED ISSUE #272: Status Bar doesn't always match theme
Currently the status bar theme is pulled from the current globally set theme.  In order to get the theme set for the device an enhancement has been requested of the front end to provide that information.

Depends on: https://github.com/home-assistant/home-assistant-polymer/issues/4220

### LINKED ISSUE #1365: Android status bar is always light in dark theme
**Home Assistant Android version:** 2021.1.1-full

**Android version:** 11

**Phone model:** Pixel 3XL

**Home Assistant version:** 2021.2.3

**Last working Home Assistant release (if known):** unknown

**Description of problem:** After setting Android theme and Home Assistant Android app theme to dark and even after reboot the status bar and the bottom navigation bar are still light.

```

```

**Screenshot of problem:**
Screenshot_20210216-220005

**Additional information:** Not sure if it is connected with #272

chriss158:
Fixes #272
Fixes #1365

## Summary

This PR tries to get the statusbar color from the webview HTML directly. If the color can be parsed, then this color will be used as statusbar color. If not, then the HA backend theme color will be used like before.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -310,6 +310,14 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
             }
 
             webChromeClient = object : WebChromeClient() {
+                override fun onProgressChanged(view: WebView?, newProgress: Int) {
+                    if (newProgress == 100) {
+                        view?.loadUrl(
+                            "javascript:window.externalApp.getHeaderColor(" +
+                                    "document.getElementsByTagName('html')[0].computedStyleMap().get('--app-header-background-color')[0]);"
+                        )
+                    }
+                }
                 override fun onJsConfirm(
                     view: WebView,
                     url: String,
@@ -401,6 +409,10 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
             }
 
             addJavascriptInterface(object : Any() {
+                @JavascriptInterface
+                fun getHeaderColor(color: String) {
+                    presenter.setStatusbarColor(color)
+                }
                 @JavascriptInterface
                 fun getExternalAuth(payload: String) {
                     JSONObject(payload).let {
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -861,15 +861,18 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
     }
 
     override fun setStatusBarAndNavigationBarColor(color: Int) {
-        var flags = window.decorView.systemUiVisibility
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -179,6 +161,36 @@ class WebViewPresenterImpl @Inject constructor(
...
     }
 
+    override fun setStatusbarColor(colorString: String) {
+        mainScope.launch {
+            try {
+                val statusbarNavBarColor = parseColorWithRgb(colorString)
+                view.setStatusBarAndNavigationBarColor(statusbarNavBarColor)
+                return@launch
+            } catch (e: Exception) {
+                Log.e(TAG, "Issue getting/setting theme from webview. Get/Set theme from HA", e)
```
nick2525, line 171: It will be nice to see timber logs https://github.com/JakeWharton/timber as it is very popular

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -310,6 +310,14 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
 
             webChromeClient = object : WebChromeClient() {
+                override fun onProgressChanged(view: WebView?, newProgress: Int) {
+                    if (newProgress == 100) {
+                        view?.loadUrl(
+                            "javascript:window.externalApp.getHeaderColor(" +
+                                    "document.getElementsByTagName('html')[0].computedStyleMap().get('--app-header-background-color')[0]);"
+                        )
+                    }
+                }
```
JBassett, line 320: Do we have to hook into the `onProgressChanged`?  Can we use the events on the external bus instead here?: https://github.com/home-assistant/android/pull/1431/files#diff-76bbedb2918a5db79c7070e9bb2e01d9241388745f60b0719c0d1f74123d89bdR439-R446

app/src/main/java/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
chriss158: I think that would be a extra PR to change the logging framework

dshokouhi: Yea out of scope we should get a feature request opened so we can start a discussion there on what the requirements are.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
chriss158: Tried calling from the `connection-status` event on the external bus, but that one is too early. The `config/get` event was ok. But i don't know if this is the correct place to set the statusbar color? Maybe there are side effects? What do you think?

JBassett (approved):

PR #1455, opened 2021-03-25
Rework high accuracy location service

chriss158:
## Summary

This PR tries to fix following error in the high accuracy location service:
`Context.startForegroundService() did not then call Service.startForeground()`

This is a very common error when using Android services. Some say, this is fault of the android framework and some say it's the fault of the app. 
Please read here: https://stackoverflow.com/questions/44425584/context-startforegroundservice-did-not-then-call-service-startforeground

The error occurred, for me,  this morning randomly, resulting in a non working high accuracy mode.

I tried to fix this issue by calling the `startForeground` method as soon as possible. And that is in the `onCreate` method of the service. Previously it was called in the `onStartCommand` method. Maybe this was too late in some circumstances.

Additionally i tried to refactor everything by using the a extended version of the class posted here: 
https://proandroiddev.com/pitfalls-of-a-foreground-service-lifecycle-59f014c6a125
Also this guide explains the problem of the error.

If there is a problem again, i added much more log outputs, to analyse the problem.

In my local test, i tried everything to reproduce this.
- Activate and deactivate the high accuracy mode as fast as possible
- Disabled the high accuracy mode when start/restarting

I couldn't  reproduce this with this PR.

I also removed this additional `notifcationManager.cancel` call, because this should not be needed... In my tests, the notification was never be NOT removed.

@dshokouhi I ping you, because be spoke some weeks before about the error, which you had also discovered while updating the sensors. Maybe the links will help here also.

## Any other notes

Translated the "Unkown" text in the GeoCoded sensor 👍

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
@@ -92,7 +92,7 @@ class GeocodeSensorManager : SensorManager {
                 )
             }.orEmpty()
 
-            val prettyAddress = address?.getAddressLine(0) ?: "Unknown"
+            val prettyAddress = address?.getAddressLine(0) ?: context.getString(R.string.state_unknown)
 
             HighAccuracyLocationService.updateNotificationAddress(context, location, prettyAddress)
 
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
@@ -92,19 +92,14 @@ class GeocodeSensorManager : SensorManager {
                 )
             }.orEmpty()
 
-            var prettyAddressFound = true
             var prettyAddress = address?.getAddressLine(0)
-            if (prettyAddress == null) {
-                prettyAddressFound = false
-                prettyAddress = context.getString(R.string.unknown_address)
-            }
 
-            HighAccuracyLocationService.updateNotificationAddress(context, location, prettyAddress)
+            HighAccuracyLocationService.updateNotificationAddress(context, location, if (!prettyAddress.isNullOrEmpty()) prettyAddress else context.getString(R.string.unknown_address))
 
             onSensorUpdated(
                 context,
                 geocodedLocation,
-                if (prettyAddressFound) prettyAddress else "Unknown",
+                if (!prettyAddress.isNullOrEmpty()) prettyAddress else "Unknown",
                 "mdi:map",
                 attributes
             )
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
@@ -92,14 +92,19 @@ class GeocodeSensorManager : SensorManager {
                 )
             }.orEmpty()
 
-            val prettyAddress = address?.getAddressLine(0) ?: context.getString(R.string.unknown_address)
+            var prettyAddressFound = true
+            var prettyAddress = address?.getAddressLine(0)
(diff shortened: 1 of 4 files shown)


app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
```diff
@@ -92,7 +92,7 @@ class GeocodeSensorManager : SensorManager {
                 )
             }.orEmpty()
 
-            val prettyAddress = address?.getAddressLine(0) ?: "Unknown"
+            val prettyAddress = address?.getAddressLine(0) ?: context.getString(R.string.state_unknown)
```
dshokouhi, line 95: This variable is used for the state of the sensor, we shouldn't translate the state string that we send back. Those need to be in English

chriss158: Thanks, i didn't know that. Fixed that 👍

dshokouhi:
> @dshokouhi I ping you, because be spoke some weeks before about the error, which you had also discovered while updating the sensors. Maybe the links will help here also.

So for sensor updates we use WorkManager that has its own `setForeground` call which we use: https://developer.android.com/reference/kotlin/androidx/work/CoroutineWorker#setforeground I believe this has its own logic for it because when I see the errors in Sentry they do not reference our app but I imagine in your case the error referenced our code?

The deprecated notice is from android S I believe.

Here is what they (Google) said needs to be worked on: https://issuetracker.google.com/issues/176085964

app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
```diff
@@ -92,7 +92,7 @@ class GeocodeSensorManager : SensorManager {
                 )
             }.orEmpty()
 
-            val prettyAddress = address?.getAddressLine(0) ?: context.getString(R.string.state_unknown)
+            val prettyAddress = address?.getAddressLine(0) ?: context.getString(R.string.unknown_address)
```
dshokouhi, line 95: Are we planning on skipping this string `context.getString(R.string.unknown_address)` in translations? Do we need to mark the string a special way?  Because right now this will send the translated string of `Unknown` to HA which is undesired when the address returns as null

chriss158:
Yes the error reference the home assistant app. 
Yeah internally the call the `startForeground`. But somehow the timing is there incorrect. Sad that this won't be addressed, as so many users have problems with this logic :/ But thats another topic. Just wanted to give you a quick update on what i found.

app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
chriss158: Ah sorry. Now i get it what you mean. Now it's ok i think?

dshokouhi: Yup looks like its ok now

app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
```diff
@@ -92,14 +92,19 @@ class GeocodeSensorManager : SensorManager {
                 )
             }.orEmpty()
 
-            val prettyAddress = address?.getAddressLine(0) ?: "Unknown"
+            var prettyAddressFound = true
```
JBassett, line 95: Do we need this variable since you check for null and set the unknown value already?

JBassett (changes requested):

app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
chriss158: No 👍  Resolved

JBassett (approved):

Issue #1456, opened 2021-03-26
Actionable notification action "REPLY" displays "reply" instead of service call title


NobleKangaroo:
**Home Assistant Android version:**
beta-620-eba8f15-full

**Android version:**
11.0 "R"

**Phone model:**
Samsung Galaxy Note 10+

**Home Assistant version:**
2021.3.4

**Last working Home Assistant release (if known):**
n/a - this is a new feature added 5 Feb 2021 - #1350

**Description of problem:**
Creating an automation with an actionable notification and sending it to the phone should display the configured titles on buttons. Action and URI do this, however reply does not.

Service call used to send notification:
```yaml
service: notify.mobile_app_sm_n975u
data:
  title: "Mailbox {{ 'opened' if trigger.to_state.state == 'on' else 'closed' }}"
  message: "Mailbox {{ 'opened' if trigger.to_state.state == 'on' else 'closed' }}"
  data:
    channel: Mailbox
    image: https://my-hass-url/local/img/camera.front_yard_camera_cropped.jpg
    actions:
      - title: "1-AAA"
        action: "action"
      - title: "2-BBB"
        action: "REPLY"
      - title: "3-CCC"
        action: "URI"
        uri: "https://google.com/"
```

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**
```
n/a
```

**Screenshot of problem:**
<img>https://i.imgur.com/aAOoH7X.png</img>

**Additional information:**
As shown in the image above, the actions "action" and "URI" display their corresponding titles on the companion app buttons, however the action "REPLY" does not.

I've found the relevant code ([NotificationActionReceiver.kt#L67-L73](https://github.com/home-assistant/android/blob/3639d42937e19ccffa2a662b2589f37b235a27c1/app/src/full/java/io/homeassistant/companion/android/notifications/NotificationActionReceiver.kt#L67-L73) and [MessagingService.kt#L950-L959](https://github.com/home-assistant/android/blob/3639d42937e19ccffa2a662b2589f37b235a27c1/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt#L950-L959)) which at first glance by untrained eyes looks like it's doing what it's supposed to be doing.

At time of writing, I've not yet found the relevant code in Home Assistant to check if the issue lies there.

dshokouhi:
[...]

PR #1457, opened 2021-03-26
Use service call provided action title for reply actions

dshokouhi:
## Summary

Fixes: #1456 

Looks like we had this hard coded instead of using what was provided in the notification service call


JBassett (approved):

NobleKangaroo:
Awesome work. Thanks for a super fast turnaround!

PR #1463, opened 2021-03-29
Add show and share logs feature

chriss158:
## Summary

This PR adds a `Show and Share logs` button to the configuration. 
The button shows the log with possibility to share the log. In order to do that, it's necessary to request the `WRITE_EXTERNAL_STORAGE` permission.
The file will be shared by the android share feature to any application. 
For example: 
- A file manager to save the log file on the external storage. 
- Email to send a email with the log
- (Github to open a issue with the log). Sadly this doesn't work, as the github app only supports images to share. Also they do not support issue templates when sharing :/ Thats why github is excluded from the share list.

Some points to consider:
- ~~The complete log file will be shared. Maybe warn the user about sensitive data which will be shared?~~ **Now we share only application logs and show a alert to notice the user that he may share sensitive data.**
- ~~Replace the current log framework with timber (https://github.com/JakeWharton/timber) to only share Home Assistant logs. However this required a major replacement of all log commands. Maybe with another PR?~~ **Right now, i see no point of using timber. If we want to reduce the debug logs in release apk, then we may migrate to timber. But i think this should be another PR.**

## Any other notes

- Reordered the "Need help" block above the "App version" block in the configuration


app/src/main/res/values/strings.xml
```diff
@@ -520,4 +520,8 @@ like to connect to:</string>
   <string name="sensor_description_battery_temperature">The current battery temperature</string>
   <string name="search_sensors">Search sensors</string>
   <string name="filter_show_only_enabled_sensors">Only enabled sensors</string>
+  <string name="share_logs">Share logs</string>
+  <string name="share_logs_summary">Collect and share logs with the Home Assistant team will help to solve issues. Please share the logs only if you have been asked to do so by a Home Assistant developer</string>
```
dshokouhi, line 524: Should we mention that sensitive information may be included?

dshokouhi:
This is amazing! Lets make sure to add this to the docs troubleshooting section as well. Not sure if we should take down everything just yet in the docs but this will be super helpful to a lot of users! Great work!

dshokouhi:
Debug APK is great, this gets about what 10 minutes worth of logs or specifically 250k lines? Couldn't tell by the code 😛

app/src/main/res/values/strings.xml
chriss158: Yes you are right. Added a extra alert dialog for that 👍

chriss158:
Now we can show logs in the app and share the log (Only HA App logs) by the share button. Additionally a alert dialog will be shown about sensitive data which may be shared.

See first post for screenshots.

dshokouhi (commented):
Looks good to me.  I tried to share the logs with github but it seems to be an error in the github app due to this log line

```
2021-03-31 16:10:01.212 1583-2616/? I/ActivityTaskManager: START u0 {cmp=com.github.android/.activities.CreateIssueComposeActivity (has extras)} from uid 10239
2021-03-31 16:10:01.256 1583-2491/? W/ActivityManager: Permission Denial: opening provider androidx.core.content.FileProvider from ProcessRecord{6735c06 1406:com.github.android/u0a239} (pid=1406, uid=10239) that is not exported from UID 10524
```

not sure if we can fix that? Also users won't be presented the issue template with this.  Can we remove github from the share list until we fix these things? otherwise every issue created will get a follow up asking for more info 😝

chriss158:
I fixed that error, but github doesn't work as they only support image sharing. Also that the app does not use the issue template is a bummer :/ I excluded the github app from the sharing list.

JBassett (approved):

PR #1467, opened 2021-03-29
Implement action bar items on notification screens and add help icon to all sensors and other areas

dshokouhi:
## Summary

To build on from #1430 after talking to @chriss158  we thought it was a good idea to link to the docs directly.

In this PR all sensors and/or sensor managers can provide a link directly to the docs. First we will try to get the docs link from `basicSensor`, if not provided we get it from `sensorManager` which has a default to the main sensor page. I opted to add the link optionally to both instances as sometimes a manager has its own link and other times the individual sensor has a link. 

Also the show only enabled sensors filter will be hidden if a user has all sensors enabled as they will not be filtering on anything.

Notifications now have filtering and deleting actions moved to the action bar.  Search has been added so a user can search against the `message` for all notifications.

Notifications and widgets also link to their respective docs location.

The title of the activity is now set for notifications, sensors and widgets screens, should allow for more items added.
## Screenshots

image

image

image

image

image

image

image


dshokouhi:
Need to make some changes for light theme icon and also since I am already in notifications fragment will make some further improvements there like implementing search and moving the management buttons to the action bar.

app/src/main/java/io/homeassistant/companion/android/settings/notification/NotificationHistoryFragment.kt
```diff
@@ -17,69 +22,99 @@ import java.util.GregorianCalendar
...
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        val prefCategory = findPreference<PreferenceCategory>("list_notifications")
+        val notificationDao = AppDatabase.getInstance(requireContext()).notificationDao()
+        if (item.itemId in listOf(R.id.last25, R.id.last50, R.id.last100)) {
+            filterValue = when (item.itemId) {
+                R.id.last25 -> "last25"
+                R.id.last50 -> "last50"
+                R.id.last100 -> "last100"
+                else -> "last25"
+            }
```
JBassett, line 94: Rather than return a string here, then in `filterNotifications` turn it into a number why not just pass the number of messages to filter?

app/src/main/java/io/homeassistant/companion/android/settings/notification/NotificationHistoryFragment.kt
```diff
@@ -143,4 +178,25 @@ class NotificationHistoryFragment : PreferenceFragmentCompat() {
...
+            "last100" -> {
+                val newList = notificationDao.getLastItems(100)
+                reloadNotifications(newList, prefCategory)
+            }
+            else -> {
+                val notificationList = notificationDao.getLastItems(25)
+                reloadNotifications(notificationList, prefCategory)
+            }
+        }
+    }
```
JBassett, line 201:
```suggestion
    private fun filterNotifications(filterValue: int, notificationDao: NotificationDao, prefCategory: PreferenceCategory?) {
            val notificationList = notificationDao.getLastItems(filterValue)
            reloadNotifications(notificationList, prefCategory)
    }
```

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/settings/notification/NotificationHistoryFragment.kt
dshokouhi: as usual a much cleaner method 😃  thank you!

JBassett (approved):
Just need to fix the merge conflicts and you are good to go.

JBassett (approved):

Issue #1472, opened 2021-04-01
Shortcut not loading correct view after closing app using home button


gadgetchnnel:
**Home Assistant Android version:** beta-644-6e9ac72-full

**Android version:** 10 (EMUI 11.0.0)

**Phone model:** Huawei Mate 20 Pro

**Home Assistant version:** 2021.4.0b0

**Last working Home Assistant release (if known):**

**Description of problem:**

When opening a pinned or dynamic shortcut and then backing out of the app using the home button, opening a second shortcut will sometimes just open the app showing the last open view.

Have been able to reliably reproduce this by:
* Open Home Assistant app
* Close app with back button
* Opening first shortcut
* Back out of the app using the home button
* Open second shortcut
* Back out of the app using the home button
* Open first shortcut again

If the app is closed using the back button, the shortcuts work as expected.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```

**Screenshot of problem:**

Screen recording of a minimal reproduction using two test shortcuts and views:

https://user-images.githubusercontent.com/2099542/113353169-306e0b00-9335-11eb-81aa-7a975b9daedb.mp4

(Updated screen recording using debug build of app and showing the closing of the HA app)

**Additional information:**

Using Nova Launcher, but also been able to reproduce using Huawei Home launcher.

dshokouhi:
This is what I see when I perform the same steps on my Pixel 4 XL

https://photos.app.goo.gl/yecZMhPPYAStLTGP6

Do you see the same issue with actionable notifications? Have you tried a debug build that has a clean app state to see if its just bad data that needs to be cleaned up?

gadgetchnnel:
So after further tests, this is only happening if I have exited the HA app using the back button before using the first shortcut. If I use the Home button to exit the app the shortcuts take me to the right place.

The same think is happening with **clickAction** in notifications, but again only if the app was closed using the back button before receiving the notification.

dshokouhi:
hmm can you update the video with that finding so I can see if I can reproduce it here?  I just tried to exit the app using back button and then launch shortcuts but they still work as expected lol.  Also lets look at the raw logs to see what is happening here we should see something I think.  I would also still check the debug build too to ensure its not data related.

gadgetchnnel:
I've updated the video sowing the closing of the HA app first, and also using the debug build to eliminate data issues.
[...]

PR #1473, opened 2021-04-01
Allow existing service call widgets be used as shortcuts

dshokouhi:
## Summary

Following on from the other shortcut PRs we will now allow users to create a shortcut to execute an existing service call widget.  As of now widgets are not being imported as I would like to do that in a separate PR where we will do an import and also allow users to pin a shortcut directly from the configure widget screen.  The plan is to reuse the same data and send an extra intent to know where to save the widget data to. Will probably require some more refactoring so a bit out of scope here.  This should allow users to move existing widgets to another screen and create shortcuts in their desired location while we work on importing and allowing the creation without needing a widget.

For now all button widgets can be selected and made into a shortcut. Shortcuts require an activity so the service call will execute in `onPageStarted` so users will see the activity briefly open then close. Toast messages are present for when the widget no longer exists and when the command was executed.

Video: https://photos.app.goo.gl/UVZzZNpLixw6ddwt9

Also fixes : #1472 by setting a new flag, same fix applied to actionable notifications

## Screenshots

image

image

image

Documentation: home-assistant/companion.home-assistant#487


dshokouhi:
Converted to draft, may cancel the PR if #1474 works out better

dshokouhi:
Closing as shortcuts should only launch activities

PR #1474, opened 2021-04-02
Allow shortcuts to execute scripts

dshokouhi:
## Summary

So I am not too sure how #1473 will feel to users as its a bit odd to create a widget just for a shortcut and also if you add a widget to a shortcut you can't remove it. To me the button widget is very static in data, once you set them up thats the data being used for the one service call every time its used.  I think scripts are probably a good and quick use case as most of the time they can be ran without any additional data and can include more than one service call :)

So in this PR we will show a list of scripts if a user has at least one script added.  As we need to link to an activity it will also use webview to quickly launch and execute the service call.  A toast message will be shown whether or not the call was successful.

In the interest of review time I will mark #1473 as draft and have also imported the fix for #1472
## Screenshots

image

image

Documentation: home-assistant/companion.home-assistant#488


dshokouhi:
Closing as shortcuts are meant to launch activities

PR #1490, opened 2021-04-06
Add translatable sensor settings

chriss158:
## Summary

This PR makes the sensor settings translatable and i also made every existing sensor setting translatable.

Every sensor setting can be translated in `strings.xml` with following name:
`
sensor_setting_<setting_name>_title
`
whereas <setting_name> is replace by the name of the setting.

To translate sensor settings which have dynamic names:
`replace_mac_<some_mac_address>`
you need to use following `var<number>:<some_var>:`
E.g.: `replace_mac_var1:<some_mac_address>:`
Then the var1 variable will be used as an argument for the `getString` method.
The suffix `replace_mac` will be used as the <setting_name>.
E.g: Translation:
 `<string name="sensor_setting_network_replace_mac_title">Replace MAC address %1$s with ...</string>`
`%1$s` will then be replace with variable of var1.

Also list values can be translated in `strings.xml`
`
sensor_setting_<setting_name>_<setting_value>_label
`
whereas <setting_name> is replace by the name of the setting
and <setting_value> is the internal value of the setting.

To achieve this, I've added a new field `entries` to the sensor_settings table. So every new sensor settings which uses a list, should use the `entries` field.

Since the sensor settings names have been changed, i have added a migration that maps all the old sensor settings names to the new sensor settings names. I have tested the migration process from the master branch to this branch several times without any problems.


app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
```diff
@@ -177,11 +177,11 @@ class NetworkSensorManager : SensorManager {
         val settingName = "replace_$bssid"
```
dshokouhi, line 177: This is also a sensor setting for the sensor if your intention is to translate all of settings

app/src/main/java/io/homeassistant/companion/android/sensors/LastUpdateManager.kt
```diff
@@ -61,15 +61,15 @@ class LastUpdateManager : SensorManager {
         )
         val sensorDao = AppDatabase.getInstance(context).sensorDao()
         val allSettings = sensorDao.getSettings(lastUpdate.id)
-        val addNewIntent = allSettings.firstOrNull { it.name == ADD_NEW_INTENT }?.value ?: "false"
+        val addNewIntent = allSettings.firstOrNull { it.name == SETTING_ADD_NEW_INTENT }?.value ?: "false"
         val intentSetting = allSettings.firstOrNull { it.name == "intent" + allSettings.size }?.value ?: ""
```
dshokouhi, line 65: `"intent"` is also the name of the setting we may want to translate :)

chriss158: I've been thinking about that, too. But if you translate it into e.g. German, it doesn't make any sense anymore because it's more of a technical term. In German at least, the English term is much better in my opinion.

dshokouhi: yea I wasnt sure if it would translate well since its more of a Android specific term here right?

chriss158: Yes, when i talk/write in german about an android specific "intent" , i really say/write "Intent"

app/src/main/java/io/homeassistant/companion/android/sensors/NetworkSensorManager.kt
chriss158: Yes you are right. Should be also translated.

dshokouhi:
Took another look I think we are missing the Notification sensors as Last Notification and Last Removed Notification both have an Allow List as well. Probably just needs the same string as the Alarm?  I still need to test the APK but haven't found a chance to, will try to later on today

chriss158:
Thanks! Should be fixed and also merged master.

chriss158:
Tested again the migration of the sensor settings. From master -> translated-sensor-settings. Looking good for me. 

Before:
image

After:
image

JBassett (approved):

Issue #1475, opened 2021-04-02
Media sensors


stipe76:
**Is your feature request related to a problem? Please describe.**

No

**Describe the solution you'd like**

Is it possible to create a new set of sensors that report the states of current media being played on the device?

I'd like to be able to use a media player card to control media instead of switching apps, particularly while driving. Artist, title, previous, next, play, pause functionality would be great. 

**Describe alternatives you've considered, if any**

**Additional context**

dshokouhi:
We only know the media sessions that we create. We can't control other sessions.

dshokouhi:
Meant to close this, unless there is an API that will do this I don't think its possible to control other apps.  The closest we have to detection is the Music Active audio sensor along with Speakerphone sensor.

stipe76:
Is there a way to give us at least the data, like artist, title, etc?

Currently, I can see them via the last notification attributes, until a new notification comes in. It would be nice to have them there as their own sensors. 

Thanks!

dshokouhi:
Looks like we actually can get some media data but not from all apps.  A quick test shows spotify works but youtube will not.  Going to reopen

retrieves all media sessions given the user has granted the notification permission:
https://developer.android.com/reference/android/media/session/MediaSessionManager#getActiveSessions(android.content.ComponentName)

the data:
https://developer.android.com/reference/android/media/session/MediaController

additional data if provided:
https://developer.android.com/reference/android/media/MediaMetadata

playback state:
https://developer.android.com/reference/android/media/session/PlaybackState

Some brief output that shows some of the data I can pull from spotify

```
2021-04-09 21:22:45.554 24282-24380/io.homeassistant.companion.android.debug D/AudioSensor: Media list size is 1
2021-04-09 21:22:45.555 24282-24380/io.homeassistant.companion.android.debug D/AudioSensor: Media session from: com.spotify.music and Muse and state PlaybackState {state=2, position=47800, buffered position=0, speed=0.0, updated=379861430, actions=142132, custom actions=[Action:mName='Add to collection, mIcon=2131231678, mExtras=Bundle[mParcelledData.dataSize=144]], active item id=-1, error=null}
```

Here I am just pulling artist name, package and playback state as well as the list of media sessions available...again not sure how many apps will work with this since youtube did not but spotify did.

one thing to keep in mind is that there is no media change intent so this sensor will not update immediately

stipe76:
I'm using YouTube Music, not sure if that would be any different. 

[...]

PR #1499, opened 2021-04-10
Add a media session sensor

dshokouhi:
## Summary

Fixes: #1475 

So this is an interesting sensor because the media sessions require either a system only permission, which we won't qualify for, or the notification listener (aka notification manager).  This listener is required for getting the data so this sensor is like a notification sensor but only for media sessions.

For the state I had a few considerations. First keeping the state as the total count of sessions won't work because when the user has 1 session (primary) none of the attributes will update.  Keeping the state as just the playback state of the primary controller wont update the attributes often either especially when playing for a long period of time. Things like artist and title won't change so almost whats the point in including them.  We will first attempt to set the state as the title of the media when it is actively playing. Otherwise, the state will be the playback state or unavailable if we have no active sessions.  I figure the more updates we get in the better for users.

Attributes will include data from all active media sessions including a total count of active media sessions. The attributes will have the package name appended.

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#492

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
@@ -1,7 +1,11 @@
 package io.homeassistant.companion.android.sensors
 
 import android.Manifest
+import android.content.ComponentName
 import android.content.Context
+import android.media.MediaMetadata
+import android.media.session.MediaSessionManager
+import android.media.session.PlaybackState
 import android.service.notification.NotificationListenerService
 import android.service.notification.StatusBarNotification
 import android.util.Log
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
```diff
@@ -184,4 +195,58 @@ class NotificationSensorManager : NotificationListenerService(), SensorManager {
...
+                    "playback_position_" + item.packageName to item.playbackState?.position,
+                    "playback_state_" + item.packageName to getPlaybackState(item.playbackState?.state)
+                )
+            }
+        }
+        attr += mapOf("total_media_session_count" to sessionCount)
+        onSensorUpdated(
+            context,
+            mediaSession,
+            primaryTitle,
```
JBassett, line 229: I think state should be the playback state, not potentially the title.  That way it stays consistent.

JBassett (changes requested):

JBassett (approved):

JBassett:
Just need to fix merge conflicts.

JBassett (approved):

PR #1501, opened 2021-04-10
Improve Themes

### LINKED ISSUE #1496: Option to manually change the "app background" color

Since 2021.4.1 the top and bottom bars are always HA blue. I personally don't ever fancy that in apps with a app header.
But mainly but with a dark theme, I think the ambition of a dark mode is kind of put aside with blue top and bottom bars.

I would recommend an option to change the bar color to the users preference.
Maybe 3 options:
1. Default: HA blue as today.
2. Follow system theme. I don't know if there's a standard system setting to utilise, but other wise the HA day/night background. Just not the header background, since it looks like a GIANT header when the header and top bar uses the same color.
3. User preference with a hex code or even better, a color picker with option to enter hex code.

Use Nacho Notch, but I am trying to use as few additional apps as possible since my main mobile is company enrolled and hence minimal possibilities to adjust.

**Additional context**

Example 1, I have chosen a really dark theme, but the perception is not very dark due to the HA blue top and bottom

image

Example 2, I have chosen a really dark theme, and using Nacho Notch to make it look like pre-2021.4.1. Easier on the eyes at night.

image

Example 3, It doesn't make sense to have the HA blue when the user have chosen to use all dark mode.

image

Example 4, An example that I think works really well is Gmail where the top and bottom bar have the same color as the app background and follows the day/night and the search bar is a discreet deviation.

image

chriss158:
Fixes #1496

## Summary

This PR improves the themes and the coloring of the status bar/navigation bar.

* **Add "Follow Home Assistant" theme setting**
  Default theme option is now "Follow Home Assistant". This one will color the status bar/navigation bar like the `app-header-background-color` variable of the set HA Theme. This is exactly the same behavior as we have now, except that the option is now called "Follow Home Assistant" instead of "Follow System-Settings".
I decided to do this, because with the option "Follow system-settings" you can choose if the statusbar ALWAYS colors according to the system, no matter which color the HA theme has (See #1496). Also at the moment it is so, that the theme setting only colors the settings of the app. That is kind of pointless. So, now you can also color the status bar/nav bar in the WebView.

* **Add instant status bar/nav bar color change, when change HA theme**
  Now the status bar/nav bar color is changed immediately with the change of the HA theme
* **Fix status bar coloring when using a light theme**
   I noticed a bug which colors a light status bar background with a light status bar text. This was caused by the method `showSystemUI`, because it removes the flag `SYSTEM_UI_FLAG_LIGHT_STATUS_BAR`
* **Use new WindowInsets API on >= Android 11**
  Since setting `window.decorView.systemUiVisibility` is marked as deprecated as of Android 11, I implemented the WindowInset API for coloring the status bar/nav bar. However, `systemUiVisibility` is also used for toggling fullscreen mode. This should be adapted to the WindowInset API with another PR in the future.
* **Color status bar/nav bar a bit darker, than HA Theme**
  To follow the design guidelines here: https://material.io/design/color/the-color-system.html , the status bar/nav bar is now colored a little bit darker than the HA theme color.

* **Enhance logging**

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/themes/ThemesManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/themes/ThemesManager.kt
@@ -63,7 +63,7 @@ class ThemesManager @Inject constructor(
                         WebSettingsCompat.FORCE_DARK_ON
                     )
                 }
-                "system" -> {
+                "android", "system" -> {
                     val nightModeFlags =
                         context.resources.configuration.uiMode and Configuration.UI_MODE_NIGHT_MASK
                     if (nightModeFlags == Configuration.UI_MODE_NIGHT_YES) {
@@ -93,7 +93,7 @@ class ThemesManager @Inject constructor(
             "dark" -> {
                 AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_YES)
             }
-            "system" -> {
+            "android", "system" -> {
                 AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM)
             }
             else -> {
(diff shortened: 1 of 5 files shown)


dshokouhi (approved):
Tested debug APK and didn't see anything out of the ordinary with default theme.

JBassett (approved):

Issue #1505, opened 2021-04-11
Reading NFC tags with any favor


T-bond:
**Is your feature request related to a problem? Please describe.**

I wrote out an NFC tag with the minimal flavor of application.
Tried to read it on another phone with only the full flavor installed.
Instead of a successful scan, it tried to open the phone store application to install the minimal flavor.
I thought that maybe it is a bug with the phone, or I missed something but the [documentation](https://www.home-assistant.io/blog/2020/09/15/home-assistant-tags/#the-apps) states:
> Tags are not bound to the phone that wrote them, any phone can scan them.

So after that I checked what got written to the tag and added a new AAR record with the another package name, which solved the issue.

**Describe the solution you'd like**

If any of the Home Assistant Android Companion app is installed on the phone it should read the NFC tag correctly.

**Describe alternatives you've considered, if any**

I don't know about any alternative without reintroducing #876.

**Additional context**

I think this should make the current workaround #296 (using multiple versions of the app) more livable.

I also made a small PR for adding all the package names to the tag when writing out new NFC tags: #1506

T-bond:
Merged.

PR #1506, opened 2021-04-11
Write all published application IDs to NFC tags

T-bond:
Signed-off-by: Nagy Tamás (T-bond) <tbondvagyok@gmail.com>

## Summary

See #1505

## Screenshots

No frontend change.

Documentation not changed.

## Any other notes

If multiple favor is installed on the user's phone, then the OS will choose which one to open. I could not find documentation about the order they are read. I tested on 3 Samsung phone and on one One Plus, and it worked the order the record were on NFC:

1.  The full favored application
2. The minimal favored application

I don't know if it will work on every phone, as there was not so much documentation about it on the internet. (I also haven't checked the Android source code to see how it works)
The ones I tested it worked correctly, and only one application got triggered.

It should automatically add new favors and follow any application ID changes if they are introduced or modified in the future.

This PR does **not** add ARR records to the NFC tag for the debug variants, as I think keeping the data size to minimal is still better for the end user.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -82,21 +82,8 @@ android {
         }
 
         // Generate a list of application ids into BuildConfig
-        var values = ""
-        productFlavors.all { flavor ->
-            if (values.isNotEmpty())
-                values += ","
-
-            values += "\""
-            values += if (flavor.applicationId == null)
-                defaultConfig.applicationId
-            else
-                flavor.applicationId
-
-            flavor.applicationIdSuffix?.let { values += it }
-            values += "\""
-
-            true
+        val values = productFlavors.joinToString {
+            "\"${it.applicationId?: defaultConfig.applicationId}${it.applicationIdSuffix}\""
         }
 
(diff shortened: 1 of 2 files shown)


homeassistant:
Hi @T-bond, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;android#1506).

Thanks!

app/build.gradle.kts
```diff
@@ -80,6 +80,26 @@ android {
...
+            values += if (flavor.applicationId == null)
+                defaultConfig.applicationId
+            else
+                flavor.applicationId
+
+            flavor.applicationIdSuffix?.let { values += it }
+            values += "\""
+
+            true
+        }
```
JBassett, line 100:
```suggestion
        // Generate a list of application ids into BuildConfig
        val values = productFlavors.joinToString {
            "\"${it.applicationId?: defaultConfig.applicationId}${it.applicationIdSuffix}\""
        }
```

JBassett (changes requested):

Issue #1515, opened 2021-04-15
Clear stale device from list of BT devices triggering High Accuracy Mode


crowedavid:
**Is your feature request related to a problem? Please describe.**

No, mostly cosmetic.

**Describe the solution you'd like**

Provide a way to clear a stale device (i.e. one that no longer appears in the list of paired BT devices) from the list of BT devices that trigger High Accuracy mode for location sensor. An automatic method to ask to remove the device if it no longer appears in the list of paired BT devices might be the best way but I'll be happy to have a button to clear all devices from the list and then reselect the desired devices.

**Describe alternatives you've considered, if any**

Reconnecting the device and clearing it from HA app would work but I no longer have the BT device I want to clear.

**Additional context**

I recently changed BT devices I use in my car and trigger the HA High Accuracy mode for location updates.  Unfortunately, I deleted the device from the Android BT configuration and noticed it still appears in the list with no way I see to remove it from the HA app.

dshokouhi:
If you go back to the settings page and edit the list and click on it should update it. This is the only page we update the settings. Can you try that and see if it works?

crowedavid:
i've gone into the settings to update the list and tried unselecting all (3) devices (4 devices total, 3 are currently paired) available from the list and it obediently removes the three devices unselected and leaves the one that no longer appears in the list of paired devices.

stipe76:
> i've gone into the settings to update the list and tried unselecting all (3) devices (4 devices total, 3 are currently paired) available from the list and it obediently removes the three devices unselected and leaves the one that no longer appears in the list of paired devices.

I can confirm exact behaviour.

dshokouhi:
can you guys share a screenshot of the issue please?  Does this stale device appear in the bluetooth connection sensor too?  Looking at the code the list preference should be resetting its values with the same values you get from bluetooth connection attributes.

crowedavid:
below is a sequence of screenshots:

1. this shows the list of BT trigger devices - the BMW65774 device is the one that is no longer paired with my phone:
![Screenshot_20210415-120837](https://user-images.githubusercontent.com/4466952/114932827-cd787b80-9dec-11eb-8ba0-55eaa1cf58bd.png)

2. this shows the list of devices to select from - BMW65774 does not appear in the list:
![Screenshot_20210415-120908](https://user-images.githubusercontent.com/4466952/114933029-13cdda80-9ded-11eb-996e-f06e74b03a4b.png)

[...]

PR #1518, opened 2021-04-15
Clear out preference value if the value does not match the provided entry list

dshokouhi:
## Summary

Fixes: #1515 by clearing out the preference value if we detect a mismatch between the entries and the selected value.

Tested this by creating a zone, adding it to high accuracy, removing the zone and coming back to sensor setting screen.


JBassett:
Need to fix merge conflicts.

JBassett (approved):

Issue #1546, opened 2021-05-17
Android 11 Power Menu, Light percent / slider no longer working


rwarner:
**Home Assistant Android version:**
2021.5.1-full

**Android version:**
Android 11

**Phone model:**
Pixel 3

**Home Assistant version:**
2021.5.4

**Last working Home Assistant release (if known):**
2021.4.6

**Description of problem:**

Originally, I used to be able to do the folllowing:

Unlock phone -> Hold Power Button -> Access quick controls for HA buttons -> Slide my finger left/right to control the light percentage on each Light Button.

It now seems as though all of the light buttons are ONLY buttons for off/on functionality and that, that former feature is gone?

Unsure why this would just disappear

https://companion.home-assistant.io/docs/integrations/android-power-menu
https://developer.android.com/guide/topics/ui/device-control

Seems to default now to a "ToggleTemplate" rather than a "ToggleRangeTemplate":

https://github.com/home-assistant/android/blob/cf27aa994fba0938aaaf6707a505421ef74d129b/app/src/main/java/io/homeassistant/companion/android/controls/LightControl.kt#L61-L82

JBassett:
@rwarner I just validated on my zigbee lights and the slider is working as expected.  Are you sure that your light's show supporting brightness in Home Assistant?

rwarner:
Interesting, yeah the lights show brightness in the entities areas. No issues using any of the sliders within the app or directly on HA. Just not supporting the feature in my power menu

dshokouhi:
I also cannot reproduce this.  Maybe something is corrupted and needs to be reset? Have you tried any basic troubleshooting steps?

Also what @JBassett was mentioning is that your light needs to have `SUPPORT_BRIGHTNESS` feature supported in HA for the slider to show up.  If it shows in teh HA frontend more than likely the integration added support for that.

We will need more to go off on this issue but first try some basic troubleshooting steps outlined in the companion docs.

rwarner:
```your light needs to have SUPPORT_BRIGHTNESS feature supported in HA ```

Where exactly can I check this? I'm in Developer Tools -> States -> I go to said light object/entity and in the attributes it has

supported_color_modes: brightness
color_mode:brightness
brightness: 254

just as a few attributes that include "brightness"

These are just hue lights connected through zigbee on HA, nothing special.

```Have you tried any basic troubleshooting steps?```

Yeah, but I will keep poking around and list everything I tried next.

dshokouhi:
[...]

PR #1564, opened 2021-05-24
Account for color mode when determining brightness slider

dshokouhi:
## Summary

Fixes: #1546 by checking the supported color modes on a device in addition to supported features. As we do not store the current HA version and users might be on a version earlier than 2021.5 we will account for both methods to make it easier.

For more details on the change: https://github.com/home-assistant/core/issues/50563#issuecomment-847251138

Also dev docs that cover which color modes support dimming/brightness: https://developers.home-assistant.io/docs/core/entity/light/#color-modes

I tested this with both a light that has brightness and one that doesn't.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/controls/LightControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/LightControl.kt
@@ -42,14 +42,8 @@ class LightControl {
 
             // On HA Core 2021.5 and later brightness detection has changed
             // to simplify things in the app lets use both methods for now
-            var supportsBrightness = false
             val supportedColorModes = entity.attributes["supported_color_modes"] as? List<String>
-            if (supportedColorModes != null) {
-                for (item in supportedColorModes) {
-                    if (item !in NO_BRIGHTNESS_SUPPORT)
-                        supportsBrightness = true
-                }
-            }
+            val supportsBrightness = if (supportedColorModes == null) false else !(supportedColorModes - NO_BRIGHTNESS_SUPPORT).isEmpty()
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/controls/LightControl.kt
```diff
@@ -38,6 +39,17 @@ class LightControl {
...
+            // On HA Core 2021.5 and later brightness detection has changed
+            // to simplify things in the app lets use both methods for now
+            var supportsBrightness = false
+            val supportedColorModes = entity.attributes["supported_color_modes"] as? List<String>
+            if (supportedColorModes != null) {
+                for (item in supportedColorModes) {
+                    if (item !in NO_BRIGHTNESS_SUPPORT)
+                        supportsBrightness = true
+                }
+            }
```
JBassett, line 52:
```suggestion
            val supportedColorModes = entity.attributes["supported_color_modes"] as? List<String>
            val supportsBrightness = if (supportedColorModes == null) false else !(supportedColorModes - NO_BRIGHTNESS_SUPPORT).isEmpty()
```

JBassett: I think the syntax is correct?

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/controls/LightControl.kt
dshokouhi: yup just tested it locally and it still works :) even when the condition below is modified thank you!

JBassett (approved):

PR #1576, opened 2021-06-02
Finish null check for tiles

dshokouhi:
## Summary

This should properly send the expected `null` values (that #1548 expected) to hopefully fix:

```
java.lang.NullPointerException: qsTile must not be null
    at io.homeassistant.companion.android.qs.Tile2Service.getTile(Tile2Service.kt:11)
    at io.homeassistant.companion.android.qs.TileExtensions.onTileAdded(TileExtensions.kt:46)
    at android.service.quicksettings.TileService$H.handleMessage(TileService.java:404)
    at android.os.Handler.dispatchMessage(Handler.java:107)
    at android.os.Looper.loop(Looper.java:241)
    at android.app.ActivityThread.main(ActivityThread.java:7582)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:941)
```

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/qs/Tile1Service.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/qs/Tile1Service.kt
@@ -7,8 +7,11 @@ import androidx.annotation.RequiresApi
 @RequiresApi(Build.VERSION_CODES.N)
 class Tile1Service : TileExtensions() {
 
-    override fun getTile(): Tile {
-        return qsTile
+    override fun getTile(): Tile? {
+        return if (qsTile != null)
+            qsTile
+        else
+            null
     }
 
     override fun getTileId(): String {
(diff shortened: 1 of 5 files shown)


app/src/main/java/io/homeassistant/companion/android/qs/Tile1Service.kt
```diff
@@ -7,8 +7,11 @@ import androidx.annotation.RequiresApi
 @RequiresApi(Build.VERSION_CODES.N)
 class Tile1Service : TileExtensions() {
 
-    override fun getTile(): Tile {
-        return qsTile
+    override fun getTile(): Tile? {
+        return if (qsTile != null)
+            qsTile
+        else
+            null
```
geekygecko, line 14: Just changing the return type to optional Tile? should fix this crash. I don't think there is any need to do the null check here, it is the same as writing `return qsTile`

dshokouhi: Honestly the tile should never be `null` to begin with lol I haven't been able to reproduce it locally so wanted to just ensure the value we are getting instead of a surprise in the next release

geekygecko: It is a good change to make as this is implementing a method from a Java class and it could be null.

JBassett (approved):

Issue #1728, opened 2021-09-28
Add a sensor for whether Work Profile is enabled


nielsrowinbik:
**Is your feature request related to a problem? Please describe.**
I have created a so called "work mode" (see [code](https://github.com/nielsrowinbik/home/blob/master/packages/work_mode.yaml)) in my Home Assistant instance that turns on the lights at my desk, sends a push notification through my phone, etc. I currently have no way of syncing this work mode (and, in turn, the lights at my desk) with the [work profile](https://developer.android.com/work/managed-profiles) on my android phone, which can be toggled on and off.

**Describe the solution you'd like**
I would like a sensor that reflects the state of the [work profile](https://developer.android.com/work/managed-profiles) on my phone in Home Assistant (could be a `binary_sensor` since it is either `on` or `off`), so that I can sync the state of my phone's work profile with my Home Assistant's work mode.

**Describe alternatives you've considered, if any**
I do not see any alternatives. Suggestions are welcome!

**Additional context**
I suppose the sensor could work much like the [Do Not Disturb sensor](https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/sensors/DNDSensorManager.kt), assuming there is a similar variable.

dshokouhi:
Looking online this will be a binary sensor as we can only check if any app is opened in work profile or not. If any app is opened in the work profile then we know the profile is enabled.

Here is the stackoverflow link to get the answer: https://stackoverflow.com/a/50869390

nielsrowinbik:
Thanks for the comment @dshokouhi. If that's the only way to know whether the profile is turned on or not then it kind of defeats the purpose for me. I'd also have to get my employer to enable the Home Assistant app to be downloaded from the work profile's Play Store which I doubt I'll be able to do ;).

The ideal solution would allow for the sensor to indicate whether the work profile is turned on or off from within the "personal" profile. I'll have a look online later to see if this can be done at all on Android.

dshokouhi:
@nielsrowinbik to my understanding all the app does is query each app to see if its running in the work profile and if it is then that means the work profile is enabled. There is no direct API to check, each app has to be checked individually.  Its not checking the HA app by itself, its checking every app on the device so it should work as intended I would imagine.  Heres a link to the actual API that would be used.

https://developer.android.com/reference/android/app/admin/DevicePolicyManager#isDeviceOwnerApp(java.lang.String)

[...]

PR #1743, opened 2021-10-01
Add a work profile sensor

dshokouhi:
## Summary

Fixes: #1728 by checking the status of each installed app to see if the work policy is enabled

Documentation: home-assistant/companion.home-assistant#581

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/DevicePolicyManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/DevicePolicyManager.kt
@@ -1,0 +1,71 @@
+package io.homeassistant.companion.android.sensors
+
+import android.app.admin.DevicePolicyManager
+import android.content.Context
+import android.util.Log
+import io.homeassistant.companion.android.R
+
+class DevicePolicyManager : SensorManager {
+    companion object {
+        private const val TAG = "DevicePolicyMgr"
+
+        val isWorkProfile = SensorManager.BasicSensor(
+            "is_work_profile",
+            "binary_sensor",
+            R.string.sensor_name_work_profile,
+            R.string.sensor_description_work_profile
+        )
+    }
+
+    override fun docsLink(): String {
+        return "https://companion.home-assistant.io/docs/core/sensors#work-profile-sensor"
+    }
+    override val enabledByDefault: Boolean
+        get() = false
+    override val name: Int
+        get() = R.string.sensor_name_device_policy
+
+    override fun getAvailableSensors(context: Context): List<SensorManager.BasicSensor> {
+        return listOf(isWorkProfile)
+    }
+
+    override fun requiredPermissions(sensorId: String): Array<String> {
+        return emptyArray()
+    }
+
+    override fun requestSensorUpdate(context: Context) {
+        updateWorkProfile(context)
+    }
+
+    private fun updateWorkProfile(context: Context) {
+
+        if (!isEnabled(context, isWorkProfile.id))
+            return
+
+        var isWorkProfileEnabled = false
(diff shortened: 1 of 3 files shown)


dshokouhi:
converted to draft until we hear from the user that this actually works

app/src/main/java/io/homeassistant/companion/android/sensors/DevicePolicyManager.kt
```diff
@@ -0,0 +1,71 @@
...
+        var isWorkProfileEnabled = false
+        try {
+            val devicePolicyManager = context.getSystemService(Context.DEVICE_POLICY_SERVICE) as DevicePolicyManager
+
+            val activeAdmins = devicePolicyManager.activeAdmins
+
+            if (activeAdmins != null) {
+                for (admin in activeAdmins) {
+                    if (devicePolicyManager.isProfileOwnerApp(admin.packageName))
+                        isWorkProfileEnabled = true
```
zmarties, line 54: You need to break after this match, otherwise you are really just checking the last activeAdmin.

app/src/main/res/values/strings.xml
```diff
@@ -596,6 +596,9 @@ like to connect to:</string>
   <string name="tile_missing_entity_title">Missing Script or Scene</string>
   <string name="tile_missing_entity_summary">Please add a script or scene to Home Assistant and revisit this screen to enable the below field</string>
   <string name="template_widget">Template Widget</string>
-    <string name="basic_sensor_name_high_accuracy_mode">High Accuracy Mode</string>
+  <string name="basic_sensor_name_high_accuracy_mode">High Accuracy Mode</string>
   <string name="sensor_description_high_accuracy_mode">Whether or not high accuracy mode is active on the device</string>
+  <string name="sensor_name_device_policy">Device Policy</string>
+  <string name="sensor_name_work_profile">Work Profile</string>
+  <string name="sensor_description_work_profile">Whether or not the work profile is currently active on the device</string>
```
zmarties, line 603: I know that you have just copied the format of existing description strings, but this form is arguably incorrect.  It should not include "or not", since that changes the meaning to "Regardless of whether the work profile is currently active on the device".  The correct description should simply be "Whether the work profile is currently active on the device".

One such discussion of the difference is at https://www.grammarphobia.com/blog/2013/03/whether-or-not.html

app/src/main/java/io/homeassistant/companion/android/sensors/DevicePolicyManager.kt
```diff
@@ -0,0 +1,71 @@
...
+
+            onSensorUpdated(
+                context,
+                isWorkProfile,
+                isWorkProfileEnabled,
+                icon,
+                mapOf()
+            )
+        } catch (e: Exception) {
+            Log.e(TAG, "Error getting the devices work profile", e)
```
zmarties, line 68: Simple typo "device's" rather than "devices".

app/src/main/java/io/homeassistant/companion/android/sensors/DevicePolicyManager.kt
dshokouhi: @zmarties this code doesnt work either way due to the nature of work profiles because the service won't run. However to my understanding only 1 package can be the current profile owner so this will only evaluate to `true` one time in the loop

https://developer.android.com/reference/android/app/admin/DevicePolicyManager#isProfileOwnerApp(java.lang.String)

The goal was to check if the phone was in work profile or not so any app will do so it wouldnt be wise to skip the others.

dshokouhi:
Closing as this doesn't seem feasible

Issue #1748, opened 2021-10-03
Class parameter required for successful intent


titiviking:
**Home Assistant Android version:**
beta-734-6de530a-full

**Android version:**
Android 10

**Phone model:**
OnePlus 8 Pro

**Home Assistant version:**
2021.9.7

**Last working Home Assistant release (if known):**
/

**Description of problem:**
Android does not always identify proper context to launch intent. The _class_ parameter is therefore required.
Currently HA does not allow to transmit the corresponding _Class_ parameter.

The implementation should provide a way to pass _Class_ parameter to enable proper usage of intents for any application.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```

**Screenshot of problem:**

**Additional information:**

I encountered the issue with OpenVPN Connect.
Using Tasker or Automate, I'm able to launch the desirect AutoConnect intent. Tasker & Automate specify the _class_ parameter along with the intent.
Manually removing the _class_ parameter fails to launch the intent with error:
     `android.content.ActivityNotFoundException: No Activity found to handle Intent`

The current code to parse _command_activity_ and _command_broadcast_intent_ ([here](https://github.com/home-assistant/android/blob/master/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt)) currently don't parse any provided class parameter.

Blog post discussing this topic: https://community.home-assistant.io/t/android-intent-openvpn-autoconnect-not-working/343119/8

dshokouhi:
not really a bug, more of an enhancement for additional applications to work :) thanks for submitting the request.

PR #1749, opened 2021-10-03
Allow intent commands to provide the class name

dshokouhi:
## Summary

Fixes: #1748  by allowing the class name to be sent via the `sticky` parameter.


titiviking:
Bug report/enhancement request documented in https://github.com/home-assistant/android/issues/1748

dshokouhi:
> Why reuse sticky? Seems like something that should have it's own data field.

Sure I can rename this one to its own value, then next PR I will rename the others to have its own value so we now longer hijack a value :)

PR #1763, opened 2021-10-10
Wear os scenes and scripts support

leroyboerefijn:
## Summary
Updated HomeActivity to contain a recylerview that displays:

- Scenes
- Scripts
- Logout

All can be activated by clicking.

Note: currently there is no support yet for custom icons, the IconDialog cannot be used to get an icon by it's mdi name string as far as I can tell. The only alternative I could find () requires a Github Access Token. Any ideas to solve this are welcome :)

## Screenshots
Screenshot_20211010_165932_debug

## Any other notes
Next step is to add tile support. This is a bit larger change, since it requires to save a list of (favorite) entities in a db

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
@@ -42,7 +42,7 @@ class HomeActivity : AppCompatActivity(), HomeView {
             .build()
             .inject(this)
 
-        adapter = HomeListAdapter(ArrayList())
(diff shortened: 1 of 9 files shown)


leroyboerefijn:
I looked at the source of the library. They do have tags, but I'm not sure if they match MDI. In any case, I couldn't find an interface to really go from the tag to the icon, but maybe I'm just missing something?

Getting beta testers involved would be awesome! I think it would make sense to have both PRs merged before doing that. Maybe even an additional small one with some minor usability things?

Issue #1769, opened 2021-10-11
Discovery not working on Android 6 and below due to missing attributes in NsdServiceInfo


dshokouhi:
**Home Assistant Android version:**

Latest master

**Android version:**

Android 6 and below (My Nexus 5 is on Android 6.0.1)

**Phone model:**

Nexus 5 but issue impacts any device on Android 6 and below

**Home Assistant version:**

2021.09.x

**Last working Home Assistant release (if known):**

Probably was never working based on HA version

**Description of problem:**

Discovery is not working on devices Android 6 and below and per Google it was never planned to be fixed. The direction is to use a 3rd party library.  The question is do we use a 3rd party lib for these 2 versions or just remove discovery for these versions? 

https://issuetracker.google.com/issues/37020436#comment16

All attributes are missing and we expect these: https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/onboarding/discovery/HomeAssistantSearcher.kt#L66-L67

The instances are found but without these attributes we can't discern the base URL. The version can probably be removed I assume?

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```

**Screenshot of problem:**

**Additional information:**

PR #1771, opened 2021-10-12
Improve onboarding for new users

dshokouhi:
## Summary

Fixes: #1769 

This PR does several things some of which are inspired by https://github.com/home-assistant/iOS/pull/1878:

- Adds a welcome screen to introduce users to the app and a short HA explanation
- Upon continuing from the welcome screen the user will be brought to the discovery fragment given that they are on a supported version of android for discovery to work.
- If a user is on android 6 and below they will be taken to the manual setup fragment
- Manual setup fragment has a few text changes and additions
- The OK button was renamed to Connect and now only enables if a user has entered a valid URL. Valid URL must have proper scheme and is formed well.
- Authentication fragment has been enhanced to show the actual SSL or HTTP errors a user may encounter before being logged in, previously these errors were only shown in the webview activity. 

## Screenshots

image

image

image

image

image

image

## Any other notes

I tried my best to design this with all screen sizes in mind 😄

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -613,7 +613,7 @@ like to connect to:</string>
   <string name="learn_more">Learn More</string>
   <string name="continue_connect">Continue</string>
   <string name="manual_title">What is your Home Assistant URL?</string>
-  <string name="manual_desc">Enter the URL of your Home Assistant server. Make sure the URL includes the protocol and port. For example: http://homeassistant.local:8123 or https://example.duckdns.org.</string>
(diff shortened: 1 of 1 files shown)


zacwest:
We decided ad-hoc: `s/instance/server`.
I'll be updating the iOS PR probably this evening with updated screenshots/flow but you can see a video version of it in #devs_mobile_apps from last night.

TomBrien:
Looks nice! It might be worth putting a line break before each of the example URLs to avoid the wrapping.

Issue #1762, opened 2021-10-10
Consolidated User Agent for more strict firewall block rules


arthurgeek:
**Is your feature request related to a problem? Please describe.**

I understand (and agree with) that 2FA flows is not something you want to add (https://github.com/home-assistant/android/issues/45). This means, to increase security, we have to resort to generic firewall block settings. That's exactly what I'm doing in my Cloudflare domain, but the solution is not the most secure one.

**Describe the solution you'd like**

I'd like to be able to change the User Agent in the app settings to something that's unique to my app. This way I can have a firewall setting that checks for this User Agent and only enables HTTP calls  from my specific app. This would be the most secure firewall config I can think of, since it can be traced back to my specific app install.

**Describe alternatives you've considered, if any**

Like I mentioned above, there are some other firewall rules you can use, but they're not as restricted as having a custom user agent.

**Additional context**

dshokouhi:
The app already provides a unique custom user agent string is this not enough to go by?  This differentiates it from any other app as it contains the app name and version number

https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/data/HomeAssistantRetrofit.kt#L48

arthurgeek:
@dshokouhi thank you for looking into this. I should mention the above calls are only for the auth part of the app. I have not logged in or tried to send sensor data to check if there were UAs.

> Yes we can add the device model to the call to help differentiate it.

To be clear, I was thinking of the device name you set up when first open the app. That one has a higher chance of being unique than device model or app name.

arthurgeek:
I wasn't able to get the calls to register or updating sensors, because I can only see blocked calls. If I'm able to update my firewall settings to take into consideration the new UA (as soon as it's sent from the WebView), I can give it another try to use the app and see if there will be any blocked calls.

dshokouhi:
heres the latest update :) https://github.com/home-assistant/android/actions/runs/1339486249

arthurgeek:
I believe the `/service_worker.js` call still uses the default WebView UA.

dshokouhi:
> There's one small thing: the user agent is still not 100% consistent, WebView calls UA is `HomeAssistant/Android HA GitHub LOCAL-full` while webhook calls is `HomeAssistant/Android HA GitHub LOCAL-1`.

in production that last part will be the version name so it won't be an issue, only in debug builds will it read a bit odd. Otherwise it will always list the app version.



[...]

PR #1776, opened 2021-10-13
Add device name to user agent and ensure user agent is used in all communication

dshokouhi:
## Summary

Fixes: #1762 

- Device Name has been added to the user agent string
- User agent header added to every call where the header was being replaced
- Override webview load URL to include the custom header in both authentication and webview itself

Note: I had tried to keep everything in one place but ran into looping errors related to dagger/dependency so I had opted to add `getDeviceName` call in a few places to avoid this.


dshokouhi:
@bramkragten and @balloob  just wanted to check if this PR will conflict with the HA frontend. We will be overriding all calls with the app with our own custom header. Currently it is there but apparently not all communication the app uses utilizes the defined header so we needed to add it just about everywhere lol.

Issue #1565, opened 2021-05-26
Option to prioritize the internal URL


agexdev:
**Is your feature request related to a problem? Please describe.**

Some time ago I noticed that apps can't get the SSID of the wifi when GPS if off on Android 11. I opened an issue but there was no real fix because that's how Android works. Now im still getting ip banned nearly daily unfortunately (maybe because of the wifi, unifi is weird) and the issue is that every device in the home network is banned aswell (the public IP is banned, thus all devices). This is very annoying.

**Describe the solution you'd like**

An option that lets me force the app to prefer the internal URL. I don't wanna have GPS on 24/7. If the app can't reach the internal URL it may use the external one. When wifi is disabled completely it could directly use the external one (it can detect if wifi is connected without gps, that suffices). But as soon as wifi is on it should (with the option turned on) prioritize the internal URL and use the external only as fallback 

**Describe alternatives you've considered, if any**

Turning GPS on, but that drains a lot of battery

**Additional context**

First issue where the GPS issue was discussed: #924

dshokouhi:
This will not fix your issue with being IP banned.  You may continue to get IP banned.  I htink the real issue is to fix the IP banning but I believe that may be a core issue and not an app issue.

> Turning GPS on, but that drains a lot of battery

this is not really true anymore, android has had a ton of improvements and is no longer a concern or battery drainer.  You might want to look into your apps that are causing this excess drain if that is what you are seeing.

Currently internal URL is already prioritized if we know we can use it. We also do indeed try all URL's if the internal one fails first.

TECHNOFAB11:
Interesting. Regardless of the IP banning stuff it never uses the internal URL except for when I have GPS turned on (and it thus can read the SSID). If available, internal is obviously faster and more secure. 

What do you mean by "if we can use it"? If the SSID is in the whitelist? If so, the override option would still be helpful. 
If the app should already prioritize internal URL without checking the SSID it's not working on my devices. Like said, none of my devices running Android 10+ use the internal URL if GPS is off. 

Also, if the app would have used the internal URL it would only ban my phone. I have a whitelist for internal URLs tho, so it would ignore it anyways. 

So some way of telling the app to always use internal URL first if wifi is connected would be helpful imo

dshokouhi:
> What do you mean by "if we can use it"? If the SSID is in the whitelist? If so, the override option would still be helpful.
[...]

PR #1788, opened 2021-10-15
Add option to always try the internal URL first

dshokouhi:
## Summary

Fixes: #1565 by adding a new option that enables once a Internal URL has been entered. Once selected we will always add the internal URL irregardless of the connected wifi SSID. This will be helpful for the users who prefer to keep location off. 

The webview portion of the app already offers the ability to switch from internal to external.

## Screenshots

image


JBassett:
So now with that change did we basically eliminate the SSID checking?  If so, we should probably remove/disable the SSID preference.

dshokouhi:
> So now with that change did we basically eliminate the SSID checking? If so, we should probably remove/disable the SSID preference.

so I made it an `or` condition for either `isInternal()` or `isPrioritizeInternal()` so if the user did not check `Always Prioritize` it will continue to function as it does today :)

PR #1796, opened 2021-10-17
Add Wear OS Settings section

dshokouhi:
## Summary

Adds a new settings section that only shows up if a user has the Wear OS app installed as its then safe to assume we can check for active wear devices and check if the wear app is installed.  This largely follows Googles example app along with our own logic. I left it as is for now so we can come back and update this section as we see fit. For now I thought it would be nice to let users know that they can install the app from here and that the initial communication is ready to go for adding future features. If the user does not have our wear app installed they can click a button to launch the play store so they can install the app.

https://github.com/android/wear-os-samples/tree/main/WearVerifyRemoteApp

## Screenshots

image

image

image

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -133,7 +133,6 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-reflect:1.5.31")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.2")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.2")
-    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-play-services:1.5.2")
 
     implementation("com.google.dagger:dagger:2.39.1")
     kapt("com.google.dagger:dagger-compiler:2.39.1")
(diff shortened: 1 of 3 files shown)


dshokouhi:
> And what if the user hasn't yet signed up for the beta program? Will the link to the play store still work?

Hmm it probably won't work and in the app we don't have a reliable way that I am aware of to detect beta users as those users can be on the production version depending on when they got the update and if a new beta was cut and pushed to the play store.  Maybe we should change the messaging until we are ready for a production release?

dshokouhi:
> Yeah, I don't see a straightforward way to detect it either. But maybe you can just add it in the description, to notify the user that a beta version is required?

So in our play store listing we were required to mention it so we have this one liner: `We are also working on supporting Wear OS. If interested be sure to check out the beta!`

https://play.google.com/store/apps/details?id=io.homeassistant.companion.android

What I can do also is to just add a note with a link to the beta so the user can sign up there too in case the app doesn't install or the link isn't working.

leroyboerefijn:
Awesome! Seems very clear to me :)

SkechyWolf:
Was just wondering what about the galaxy watch as with the galaxy watch 4 it has you install the Samsung wearable app

dshokouhi:
> Was just wondering what about the galaxy watch as with the galaxy watch 4 it has you install the Samsung wearable app

Had no idea that watch needed a different app, according to Google its the only one that uses a different app. Will need to update the condition to look for that `package`.  There isnt any real functionality here asides from the initial watch to phone communication but if you want to check it out youll need to install the wear OS app until we get the code updated.

https://support.google.com/wearos/answer/6056630?hl=en&co=GENIE.Platform%3DAndroid

Also in the future please don't comment on merged PR's as this might get missed an issue would be more appropriate :)

PR #1842, opened 2021-10-27
Wear OS shortcut Tile

leroyboerefijn:
## Summary
This is a first Tile implementation for Wear OS. It contains up to 7 shortcuts, that are simple click to activate. 

At this time, there are no settings to select which entities to show. For now, it just shows the first 7 (or fewer) scene entities in alphabetical order.

## Screenshots
favorite_entities_tile_example

### Settings
Screenshot_20211110_220020_debug

Shortcuts settings (top)
Screenshot_20211110_220023_debug

Shortcut settings (bottom)
Screenshot_20211110_220027_debug

Choose entity
Screenshot_20211110_220030_debug

Documentation: home-assistant/companion.home-assistant#612


leroyboerefijn:
Before continuing with this item, I will wait for #1844 to be merged. Then I'll create a new PR to add a settings activity (using compose).
This tiles PR can then be rebased, to add settings for selecting entities for the tile

leroyboerefijn:
Added settings, where you can choose up to 7 entities to display on the tile.

Two remaining issues I see:

- I don't know how to get a resource string into the Tile, so the text when no entities are selected is currently hardcoded
- To perform an action, the tile needs to call an activity. I set the theme of this activity to @android:style/Theme.Translucent.NoTitleBar to make sure that no layout is shown. This worked at first, but now clicking the buttons will open the last openend view in the app

wear/src/main/java/io/homeassistant/companion/android/settings/SettingsComposable.kt
```diff
@@ -166,3 +217,160 @@ fun ScreenSetFavorites(
...
+
+@Composable
+fun ScreenSetTileShortcuts(
+    shortcutEntities: MutableList<String>,
+    context: Context,
+    swipeDismissableNavController: NavHostController,
+    onShortcutEntitySelectionChange: (Int) -> Unit
+) {
+
+    val scalingLazyListState: ScalingLazyListState = rememberScalingLazyListState()
```
dshokouhi, line 229: If you add `RotaryEventState(scrollState = scalingLazyListState)` in between 229 and 230 then the bezel will work with rotary input :)

Same thing if you apply it to the other lazy list :)

wear/src/main/AndroidManifest.xml
```diff
@@ -43,6 +43,24 @@
 
         <!-- To show confirmations and failures -->
         <activity android:name="androidx.wear.activity.ConfirmationActivity" />
+
+        <!-- Tiles -->
+        <service
+            android:name=".tiles.FavoriteEntitiesTile"
+            android:label="@string/favorite_entities"
+            android:description="@string/favorite_entities"
```
dshokouhi, line 51: Maybe add some kind of different description like "Select up to 7 entities" because in the Wear OS app I see the name repeating :)

image

dshokouhi:
Took a look at the debug APK, the flow is nice in how we add entities. I dig it, the tile is great as it provides very quick access.  Are we able to make any updates to the tile while it is active? 

Not for this PR but a general question, if a user had a need for more tiles do we just need to duplicate everything? Is there a limit on how many tiles an app can have? I see we need to define them in the manifest so its similar to quick setting tiles where the app has to define the finite set of tiles it can set.

leroyboerefijn:
> Took a look at the debug APK, the flow is nice in how we add entities. I dig it, the tile is great as it provides very quick access. Are we able to make any updates to the tile while it is active?

Yes, the Tile should be updated whenever the settings are changed. Does this work for you?

> Not for this PR but a general question, if a user had a need for more tiles do we just need to duplicate everything? Is there a limit on how many tiles an app can have? I see we need to define them in the manifest so its similar to quick setting tiles where the app has to define the finite set of tiles it can set.

Yes, they are defined in the manifest indeed, so I suppose we need to duplicate everything. Or maybe we can trick the manifest to add it multiple times? In any case we cannot dynamically create new tiles. Personally I think having multiple shorcut tiles would be less preferable to just opening the list in the app, but we can consider the option in a future PR.

wear/src/main/AndroidManifest.xml
leroyboerefijn: Good suggestion!

I also changed the name of the tile to Shortcuts. Also because it is really separate from the favorites we have in the app.

wear/src/main/java/io/homeassistant/companion/android/settings/SettingsComposable.kt
leroyboerefijn: Awesome, works like a charm!

dshokouhi:
> Yes, the Tile should be updated whenever the settings are changed. Does this work for you?

Sorry this was more of a general question, I was just wondering if for example we could show the state of an entity which only makes sense if we could update the state while the tile was active (i.e. in view of the user)



> Yes, they are defined in the manifest indeed, so I suppose we need to duplicate everything. Or maybe we can trick the manifest to add it multiple times?

I did not see a way to trick the manifest when I added quick setting tiles, each one had to be defined individually which is why they are all numbered like that lol

leroyboerefijn:
> Sorry this was more of a general question, I was just wondering if for example we could show the state of an entity which only makes sense if we could update the state while the tile was active (i.e. in view of the user)

Ah I see! Yes, this should be possible. I think there are generally two ways: set a refresh interval to always update (need to consider battery life vs usefulness). Or request a refresh from an application/service (maybe interesting with the websocket?).
I don't think you can update the color of the icons, but you can update the color of the circles.

> I did not see a way to trick the manifest when I added quick setting tiles, each one had to be defined individually which is why they are all numbered like that lol

Ahh too bad. I understand that the typical use case is only one instance per app, but a bit of flexibility would have been nice haha

common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
```diff
@@ -352,6 +354,17 @@ class IntegrationRepositoryImpl @Inject constructor(
...
+        val jsonArray = JSONArray(localStorage.getString(PREF_TILE_SHORTCUTS) ?: "[]")
+        return List(jsonArray.length()) {
+            jsonArray.getString(it)
+        }
+    }
+
+    override suspend fun setTileShortcuts(entities: List<String>) {
+        localStorage.putString(PREF_TILE_SHORTCUTS, JSONArray(entities).toString())
+    }
+
```
JBassett, line 367: Why are we storing as a `String` if it's a list?

wear/src/main/java/io/homeassistant/companion/android/home/HomePresenterImpl.kt
```diff
@@ -106,4 +106,13 @@ class HomePresenterImpl @Inject constructor(
...
     }
+
+    override suspend fun getTileShortcuts(): List<String> {
+        return integrationUseCase.getTileShortcuts()
+    }
+
+    override suspend fun setTileShortcuts(entities: List<String>) {
+        integrationUseCase.setTileShortcuts(entities)
+        view.refreshShortcutsTile()
+    }
```
JBassett, line 117: Rather than passing a string we should create a new class to pass with the needed fields.  This will help when the Room DB is ready for wear.

wear/src/main/java/io/homeassistant/companion/android/home/views/ChooseEntityView.kt
```diff
@@ -0,0 +1,99 @@
...
+        item {
+            SetTitle(id = R.string.shortcuts)
+        }
+        item {
+            Chip(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(bottom = 16.dp),
+                icon = { Image(asset = CommunityMaterial.Icon.cmd_delete) },
+                label = { Text(text = "None") },
```
JBassett, line 60: Should be a string resource.

wear/src/main/java/io/homeassistant/companion/android/home/views/ChooseEntityView.kt
```diff
@@ -0,0 +1,99 @@
...
+                    Text(
+                        text = attributes["friendly_name"].toString(),
+                        maxLines = 2,
+                        overflow = TextOverflow.Ellipsis
+                    )
+                },
+                enabled = validEntities[index].state != "unavailable",
+                onClick = {
+                    val elementString = "${validEntities[index].entityId},${attributes["friendly_name"]},${attributes["icon"]}"
+                    onEntitySelected(elementString)
```
JBassett, line 93: This should pass that new object that we are using rather than a concatenated string.

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
```diff
@@ -19,14 +22,15 @@ import com.mikepenz.iconics.typeface.library.community.material.CommunityMateria
...
     Text(
         text = stringResource(id = id),
         textAlign = TextAlign.Center,
         fontSize = 15.sp,
         fontWeight = FontWeight.Bold,
         modifier = Modifier
             .fillMaxWidth()
+            .padding(top = topPadding)
     )
 }
```
JBassett, line 35: Latest develop doesn't have this anymore.

wear/src/main/java/io/homeassistant/companion/android/tiles/ShortcutsTile.kt
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    private suspend fun getEntities(): List<String> {
+        DaggerTilesComponent.builder()
+            .appComponent((applicationContext as GraphComponentAccessor).appComponent)
+            .build()
+            .inject(this@ShortcutsTile)
+
+        return integrationUseCase.getTileShortcuts()
+    }
```
JBassett, line 139: As mentioned previously, this should be an object with properties rather than just a string.

JBassett (changes requested):

common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
leroyboerefijn: For localStorage, the only other option is to store it as a map, but then we loose the order of the list

wear/src/main/java/io/homeassistant/companion/android/home/HomePresenterImpl.kt
leroyboerefijn: I like that idea

leroyboerefijn: Implemented it in SimplifiedEntity, let me know what you think.

Also, shall I go ahead to do the same for favorite entities? They currently use the same method

wear/src/main/java/io/homeassistant/companion/android/home/views/ListHeader.kt
```diff
@@ -29,12 +32,13 @@ fun ListHeader(
...
 @Composable
-fun ListHeader(id: Int) {
+fun ListHeader(id: Int, topPadding: Dp = 0.dp) {
     ListHeader {
         Row {
             Text(
                 text = stringResource(id = id),
-                color = Color.White
+                color = Color.White,
+                modifier = Modifier.padding(top = topPadding)
```
JBassett, line 41: https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier
Rather than passing in a padding we should pass in the whole modifier.

wear/src/main/java/io/homeassistant/companion/android/home/HomePresenterImpl.kt
```diff
@@ -106,4 +107,13 @@ class HomePresenterImpl @Inject constructor(
...
         integrationUseCase.setWearHomeFavorites(favorites.toSet())
     }
+
+    override suspend fun getTileShortcuts(): List<SimplifiedEntity> {
+        return integrationUseCase.getTileShortcuts().map { SimplifiedEntity(it) }
+    }
+
+    override suspend fun setTileShortcuts(entities: List<SimplifiedEntity>) {
+        integrationUseCase.setTileShortcuts(entities.map { it.entityString })
+        view.refreshShortcutsTile()
```
JBassett, line 117: Rather than pass this around (View -> onClick -> ViewModel -> Presenter -> View) we should just put the call right in the onClick handlers where needed I think?

JBassett (changes requested):

wear/src/main/java/io/homeassistant/companion/android/home/views/ListHeader.kt
leroyboerefijn: Makes more sense indeed :)

wear/src/main/java/io/homeassistant/companion/android/home/HomePresenterImpl.kt
leroyboerefijn: Passing on the view now, to call it immediately in the onClick handler

common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
JBassett: Ok.... but only because we will be removing this in favor of the DB calls long term.

leroyboerefijn: If you see another option, I'm happy to change it now as well :)
This was the best I could come up with

JBassett (approved):

wear/src/main/java/io/homeassistant/companion/android/data/SimplifiedEntity.kt
```diff
@@ -1,6 +1,10 @@
 package io.homeassistant.companion.android.data
 
-data class SimplifiedEntity(var entityId: String, var friendlyName: String, var icon: String) {
+data class SimplifiedEntity(
+    var entityId: String,
+    var friendlyName: String = entityId,
+    var icon: String = ""
```
leroyboerefijn, line 6: Will the new defaults not overwrite the second constructor below?

hkusulja:
I hope that i will be able to choose in app settings which (concrete) lights and switches to be in my tile, that are my favorites / most commonly used to be accessible as much as possible.

leroyboerefijn:
Yes you can. You can see the settings in the description at the top, or try the latest version of the app.

hkusulja:
I am using now latest beta app from store, I have added tile, and configured in settings > shortcuts for tile.
However, tile is static with some default icons and no functionality. I have tried to change shortcuts, but Tile is the same. Maybe it is not dynamically updated at all. Is this a bug to report? I have Wear OS 3.0 (Samsung galaxy watch 4)

leroyboerefijn:
If the tile remains static with the default shortcuts, An error has occurred during loading.

If you have selected entities without custom icons, that was an issue that will be fixed in the next version. 

If you have another error, please create a bug report and add recent logs!

hkusulja:
@leroyboerefijn i do not see any error message. My entities have default icons (which are different for each entity). Ok, waiting for next week update and then i will open new issue if it will not work..., thank you for fast response

Issue #1803, opened 2021-10-19
Wear OS - Lights list is empty


hkusulja:
**Home Assistant Android version:** beta-1534-9f3f552-full

**Android version:** 11

**Phone model:** Samsung Galaxy 

**Home Assistant version:** core-2021.10.6

**Last working Home Assistant release (if known):**

**Description of problem:**
Opening Home Assistant (beta) app from Google Play store (latest) on my Samsung Galaxy Watch Active 4 (Wear OS 3.0) , the scenes list is working, however lights list is empty / showing lot of (empty) rows (there is scroll) with no text. So it is not possible to click on light and turn it on / off

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```

**Screenshot of problem:**

**Additional information:**

Issue #1838, opened 2021-10-26
Wear OS - switches control


hkusulja:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

As you have scenes, scripts and lights, please implement option to control switches on Wear OS.

**Describe alternatives you've considered, if any**

**Additional context**

PR #1844, opened 2021-10-27
Convert wear home activity to compose

dshokouhi:
## Summary

Fixes: #1803 and #1838 

Converts the home activity to use Compose. Added a few more entity domains with support for more.  This requires a minimum SDK bump to API 25 as that is required for compose on Wear OS.  There is also a loading screen while we wait for the list of entities to load

## Screenshots

image

image

Documentation: home-assistant/companion.home-assistant#605


wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
```diff
@@ -91,4 +89,159 @@ class HomeActivity : AppCompatActivity(), HomeView {
         startActivity(intent)
         finish()
     }
+
+    @Composable
+    private fun LoadHomePage() {
+
+        val entities = runBlocking {
+            presenter.getEntities()
```
JBassett, line 97: This should be passed into the composable not called within. 
 https://developer.android.com/jetpack/compose/state#state-in-composables

By doing that you should be able to detach the loading of the entities from the initial creation of the view.   That should help with the slowness in loading the view.

JBassett (changes requested):

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
dshokouhi: Not sure I notice it loading any faster, subsequent loads are faster though so maybe its that initial load will be slow?

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
```diff
@@ -33,53 +60,28 @@ class HomeActivity : AppCompatActivity(), HomeView {
...
 
-        adapter = HomeListAdapter()
-        adapter.onSceneClicked = { entity -> presenter.onEntityClicked(entity) }
-        adapter.onButtonClicked = { id -> presenter.onButtonClicked(id) }
-
-        findViewById<WearableRecyclerView>(R.id.home_list)?.apply {
-            layoutManager = LinearLayoutManager(this@HomeActivity)
-            adapter = this@HomeActivity.adapter
+        if (presenter.onViewReady()) {
+            val entities = runBlocking {
```
JBassett, line 71: This should be a class variable, not just a function variable.  It should also use the mutable state:
`val entities by mutableStateListOf<Entity<Any>>()`

JBassett: Rather than running this blocking we should allow the UI to draw and load this on an IO thread then update the class variable of `entities`.  This will make it feel snappy on the first load.

JBassett (changes requested):

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
dshokouhi: this is proving to be very difficult 🤔 would appreciate any hints :)

dshokouhi: ok looks like we had to switch to a view model for this, now recomposition triggers and its a lot faster, especially on subsequent reloads :)

dshokouhi: I ended up reverting changes to `onViewReady` as it was no longer necessary for recomposition

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
dshokouhi: This is now taken care of with the new view model

dshokouhi:
There is now a loading screen while we wait for the viewmodel to update

image

JBassett (approved):

hkusulja:
Thank you so much for contribution for my github feature list here.
I am awaiting Sunday so I can get beta from Google Play Store and try it.
Question is, is it possible to have multiple tiles, and where / how to defined what is on tile list and what is not?
For example, I have lot of lights and switches, and just some of them want that are available on first / main tile for me. And some others on another tile, anyhow, not all should be visible on tiles. Is this managed somewhere in HA web or mobile app or somewhere alse?

leroyboerefijn:
@hkusulja Tile support will be added in a separate PR and will have to wait a tiny bit longer. Right now it will only support one tile with 7 shortcuts. It will be merged once I finish the settings in the wear os app, to select which 7. 

The same settings will be added to the android app as well, somewhere in the future. Multiple tile support is not considered yet, but could always be a feature request of course

hkusulja:
Ok, no problem about waiting, this team is already doing a great job.
It would be easier to choose / manage tile settings on the phone app, instead on the watch directly, but ok.
One fully working tile is fine as a priority and later for multiple tiles, i will check and create github feature request.

PR #1854, opened 2021-10-29
Fix sensor update for BT and Next alarm sensor not working in background

### LINKED ISSUE #1813: Android 12 brings back High Accuracy mode notification lingers issue
2021.10.0-full

**Android version:**
12

**Phone model:**
Pixel 5

**Home Assistant version:**
core-2021.9.7

**Description of problem:**
After upgrading to Android 12 the symptoms of issue #1509 have returned.

```

```

**Screenshot of problem:**

**Additional information:**
I didn't change any other settings in the app or Home Assistant itself - just upgraded to Android 12 yesterday.  I use BT devices to trigger High Accuracy mode and the notification will remain after the first connection and then go away after the second. This pattern has been consistently repeatable so far but only 24+ hours since I upgraded.

chriss158:
Fixes #1813

## Summary

This PR fixes that sensor updates do not work, when the app is in background. Currently all sensor will be only updated if the app was opened. 

Because now exported=false is used in the Receiver SensorReceiver, the intents are only allowed from the own app. If exported=true is set, intents from outside of the app will be allowed.

Before the Android 12 commits, this value was not set. Then the default value of true was used.
> https://developer.android.com/guide/topics/manifest/receiver-element.html#exported
 If unspecified, the default value depends on whether the broadcast receiver contains intent filters. If the receiver contains at least one intent filter, then the default value is "true". Otherwise, the default value is "false".

Now that the exported attribute is mandatory with Android 12, we need to set this flag to true.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -59,7 +59,7 @@
 
         <!-- Start things like SensorWorker on device boot -->
         <receiver android:name=".sensors.SensorReceiver"
-            android:exported="false">
+            android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.BOOT_COMPLETED" />
                 <action android:name="android.intent.action.MY_PACKAGE_REPLACED" />
@@ -130,7 +130,7 @@
         </receiver>
 
         <receiver android:name=".widgets.template.TemplateWidget" android:label="@string/template_widget"
-            android:exported="false">
+            android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.BOOT_COMPLETED" />
                 <action android:name="android.intent.action.QUICKBOOT_POWERON" />


dshokouhi:
Interesting dont think I noticed this earlier :) but glad you were able to resolve it! Do we really need it for template widget? Mine is updating as expected or is it related to the intents as well?

chriss158:
I dont use the widgets myself, but as the intent filter of the TemplatWidget also use some system intents, i think we need to set the flag also there to true.

chriss158:
Maybe we even have to set the exported attribute to true everywhere where we didn't use the exported attribute before. Because at that stage it was also true by not using it.

chriss158:
Ok I have looked at all the modified exported attributes. In my opinion all of them are ok. But for the receivers that use `android.appwidget.action.APPWIDGET_UPDATE` or `android.appwidget.action.APPWIDGET_CONFIGURE` I'm not sure. Since the widgets work for @dshokouhi, I would leave the attribute exported here on false

dshokouhi:
> Currently all sensor will be only updated if the app was opened.

This comment forced me to look at all my sensor states. I do not notice this issue all registered intents are still received here things like screen on, wifi, BT etc... they all update as expected. I assume this flag is probably related to the foreground service being launched for location maybe?

As long as you were able to reproduce and confirm the fix I think its good though :)

chriss158:
sorry wrote without thinking :D Indeed not all. But i think the BT sensor and Next Alarm sensor. Can you confirm the next alarm sensor?

dshokouhi:
> Can you confirm the next alarm sensor?

image

At 645 am is when I hit the snooze button :P

dshokouhi:
If you notice in the comments for #1813 the user also mentions that android itself showed the BT connection lingering for them, I think that might be the actual cause of the issue there but its more curious since you can reproduce it. With that said if the export flag fixed it for you then it should definitely be changed :)

chriss158:
Well, it was for me so that it does not work regularly. Especially after I got out of the car. I.e. the app was a bit in the background. The bluetooth sensor has still shown the connection with my car. As soon as i opend up the app, the sensor will be updated.

I will do a short drive and will test again ^^

chriss158:
Worked flawlessly 👍

dshokouhi (approved):
Thanks for testing!

dshokouhi:
Was helping out a user and look what I see in logcat...

```
2021-10-30 08:38:12.220 1744-7389/? W/BroadcastQueue: Permission Denial: broadcasting Intent { act=android.bluetooth.device.action.ACL_CONNECTED flg=0x5000010 (has extras) } from com.android.bluetooth (pid=2543, uid=1002) to io.homeassistant.companion.android/.sensors.SensorReceiver is not exported from uid 10275
2021-10-30 08:38:12.221 1744-2159/? W/BroadcastQueue: Permission Denial: broadcasting Intent { act=android.bluetooth.device.action.ACL_CONNECTED flg=0x5000010 (has extras) } from com.android.bluetooth (pid=2543, uid=1002) to io.homeassistant.companion.android/.widgets.template.TemplateWidget is not exported from uid 10275
```

JBassett (approved):

PR #1868, opened 2021-11-03
Move entities call back to presenter, use ToggleChip for toggle domains

dshokouhi:
## Summary

Since we switched to `ViewModel` the actual function gets called continuously as an effort to update the UI faster. This resulted in so many calls logcat itself would not like it due to the debug `OkHTTP` logs. So now we update the view model variable from the presenter so we can control the updates. 

This PR also uses `ToggleChip` for any entity that uses `domain.toggle` service call.

## Screenshots

image

## Any other notes

For now I opted to update the UI when an entity is selected to update the UI faster. In the future may look into https://google.github.io/accompanist/swiperefresh/ or something similar so updates wont be as impactful.


wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
```diff
@@ -235,26 +244,72 @@ class HomeActivity : ComponentActivity(), HomeView {
...
+            Chip(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(top = if (index == 0) 30.dp else 10.dp),
+                icon = {
+                    if (iconBitmap != null) {
+                        Image(asset = iconBitmap)
+                    } else
+                        Image(asset = CommunityMaterial.Icon.cmd_cellphone)
+                },
```
JBassett, line 295:
```suggestion
                icon = {
                    Image(asset = iconBitmap ?: CommunityMaterial.Icon.cmd_cellphone)
                },
```

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
```diff
@@ -235,26 +244,72 @@ class HomeActivity : ComponentActivity(), HomeView {
...
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(top = if (index == 0) 30.dp else 10.dp),
+                appIcon = {
+                    if (iconBitmap != null) {
+                        Image(asset = iconBitmap)
+                    } else
+                        Image(asset = CommunityMaterial.Icon.cmd_cellphone)
+                },
```
JBassett, line 264:
```suggestion
                appIcon = {
                    Image(asset = iconBitmap ?: CommunityMaterial.Icon.cmd_cellphone)
                },
```

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
```diff
@@ -72,9 +79,11 @@ class HomeActivity : ComponentActivity(), HomeView {
             .inject(this)
 
         presenter.onViewReady()
+        mainScope.launch {
+            entityViewModel.entitiesResponse = presenter.getEntities()
+        }
```
JBassett, line 84: This code is reused in many places, it should be a function that can be called rather than repeated.

JBassett (changes requested):

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
dshokouhi: Thank you for the suggestion! As I had other changes it was easier to incorporate in the same commit :)

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
dshokouhi: Thank you for the suggestion! As I had other changes it was easier to incorporate in the same commit :)

JBassett (approved):

PR #1875, opened 2021-11-05
Add settings and favorites for wear so users can quickly execute actions they do regularly faster

dshokouhi:
## Summary

- Adds a new settings section so users can set and clear favorites (clear is important in case the entity list is old like the user deleted an entity)
- When favorites are set they load as a chip on the home screen, then as soon as entities are loaded they convert to toggles if necessary. This allows them to be executed before we finish loading

Fixes: #1881 
## Screenshots

image

image

image

image

image

Documentation: home-assistant/companion.home-assistant#610

## Any other notes

Theres a lot going on in this PR in a future PR lets do the phone side for grabbing and setting favorites :)

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
@@ -14,11 +14,6 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.CompositionLocalProvider
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.saveable.rememberSaveable
-import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
@@ -52,12 +47,12 @@ import io.homeassistant.companion.android.common.dagger.GraphComponentAccessor
 import io.homeassistant.companion.android.common.data.integration.Entity
 import io.homeassistant.companion.android.onboarding.OnboardingActivity
 import io.homeassistant.companion.android.onboarding.integration.MobileAppIntegrationActivity
-import io.homeassistant.companion.android.util.LocalRotaryEventDispatcher
-import io.homeassistant.companion.android.util.RotaryEventDispatcher
-import io.homeassistant.companion.android.util.RotaryEventHandlerSetup
-import io.homeassistant.companion.android.util.RotaryEventState
-import io.homeassistant.companion.android.settings.SettingsComposable
-import io.homeassistant.companion.android.util.CommonFunctions
+import io.homeassistant.companion.android.settings.ScreenSetFavorites
+import io.homeassistant.companion.android.settings.ScreenSettings
+import io.homeassistant.companion.android.util.SetTitle
+import io.homeassistant.companion.android.util.getIcon
+import io.homeassistant.companion.android.util.setChipDefaults
+import io.homeassistant.companion.android.util.updateFavorites
 import io.homeassistant.companion.android.viewModels.EntityViewModel
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -98,7 +93,7 @@ class HomeActivity : ComponentActivity(), HomeView {
 
         presenter.onViewReady()
         updateEntities()
-        CommonFunctions().updateFavorites(entityViewModel, presenter)
+        updateFavorites(entityViewModel, presenter, mainScope)
         setContent {
             LoadHomePage(entities = entityViewModel.entitiesResponse, entityViewModel.favoriteEntities)
         }
@@ -125,8 +120,6 @@ class HomeActivity : ComponentActivity(), HomeView {
     @Composable
     private fun LoadHomePage(entities: Array<Entity<Any>>, favorites: MutableSet<String>) {
 
-        val rotaryEventDispatcher = RotaryEventDispatcher()
-
         if (entities.isNullOrEmpty() && favorites.isNullOrEmpty()) {
             Column {
                 Spacer(
@@ -134,7 +127,7 @@ class HomeActivity : ComponentActivity(), HomeView {
                         .fillMaxWidth()
                         .padding(top = 10.dp)
                 )
-                CommonFunctions().SetTitle(id = R.string.loading)
+                SetTitle(id = R.string.loading)
                 Chip(
                     modifier = Modifier
                         .padding(top = 50.dp, start = 10.dp, end = 10.dp),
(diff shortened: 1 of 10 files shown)


wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
```diff
@@ -103,10 +119,11 @@ class HomeActivity : ComponentActivity(), HomeView {
         finish()
     }
 
+    @ExperimentalWearMaterialApi
     @Composable
-    private fun LoadHomePage(entities: Array<Entity<Any>>) {
+    private fun LoadHomePage(entities: Array<Entity<Any>>, favorites: MutableSet<String>) {
 
-        if (entities.isNullOrEmpty()) {
+        if (entities.isNullOrEmpty() && favorites.isNullOrEmpty()) {
```
leroyboerefijn, line 126: Is there any possibility where entities is null or empty, but favorites is not?

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
```diff
@@ -317,11 +384,190 @@ class HomeActivity : ComponentActivity(), HomeView {
...
+                    .padding(top = 10.dp),
+                icon = {
+                    Image(asset = CommunityMaterial.Icon.cmd_cog)
+                },
+                label = {
+                    Text(
+                        text = stringResource(id = R.string.settings)
+                    )
+                },
+                onClick = { swipeDismissableNavController.navigate("Settings") },
```
leroyboerefijn, line 403: Shouldn't you use SCREEN_SETTINGS here?

leroyboerefijn (commented):
Awesome work once again! I'm impressed with your speed 😋

Haven't tested it out yet, but looks good!

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
dshokouhi: Yes there is, so down below in the `else` we do individual checks for favorites and entities again. So basically if the user has not had entities loaded yet and hasnt set favorites they see old loading. If they dont have entities yet but do have favorites they will see favorites and the loading chip, then loading screen/chip will disappear once we get the list of entities.

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
dshokouhi: lol yup totally missed it haha, fixed now

leroyboerefijn:
I'm not sure what the best practices are for navigation in compose. But I was wondering if the HomeActivity is not becoming too large when it will contain all of the settings pages and subpages as well?
Would it make sense to have all settings related stuff in a separate Activity?

dshokouhi:
Well we could move the composable to a separate file but navigating to composables instead of activities seems to be what the docs say.

https://developer.android.com/jetpack/compose/navigation

Also using the current method gives us the cool animation and handles the backstack etc..

leroyboerefijn:
> Well we could move the composable to a separate file but navigating to composables instead of activities seems to be what the docs say.
> 
> https://developer.android.com/jetpack/compose/navigation
> 
> Also using the current method gives us the cool animation and handles the backstack etc..

Okay good then! I'll try to test it soon, am curious about the animations and functionality!

dshokouhi:
ok did some refactoring so HomeActivity has a bit less going on, hopefully that helps make things easier to follow

wear/src/main/java/io/homeassistant/companion/android/settings/SettingsComposable.kt
```diff
@@ -0,0 +1,161 @@
...
+                },
+                enabled = entityViewModel.favoriteEntities.isNotEmpty()
+            )
+        }
+    }
+
+    @Composable
+    fun ScreenSetFavorites(
+        validEntities: List<Entity<Any>>,
+        scalingLazyListState: ScalingLazyListState,
```
leroyboerefijn, line 100: I believe this should be a separate object. Now the state of the list is shared between lists. This also causes the set favorites list to not start at the top for me.

dshokouhi: hmm I think that its probably best to always start at the top here, maybe later on we can sort based on the checked status?

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
```diff
@@ -0,0 +1,73 @@
...
+import com.mikepenz.iconics.typeface.library.community.material.CommunityMaterial
+import io.homeassistant.companion.android.R
+import io.homeassistant.companion.android.home.HomePresenter
+import io.homeassistant.companion.android.viewModels.EntityViewModel
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+
+class CommonFunctions {
```
JBassett, line 27: I don't think this really needs to be a class with a bunch of functions, rather just a bunch of top level functions.

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
```diff
@@ -0,0 +1,73 @@
...
+        )
+    }
+
+    fun updateFavorites(entityViewModel: EntityViewModel, presenter: HomePresenter) {
+        mainScope.launch { entityViewModel.favoriteEntities = presenter.getWearHomeFavorites().toMutableSet() }
+    }
+
+    fun saveFavorites(favorites: Set<String>, presenter: HomePresenter) {
+        mainScope.launch { presenter.setWearHomeFavorites(favorites.toSet()) }
+    }
```
JBassett, line 57: Pass in the scope so you can make these top level.

wear/src/main/java/io/homeassistant/companion/android/settings/SettingsComposable.kt
```diff
@@ -0,0 +1,158 @@
...
+import com.mikepenz.iconics.compose.Image
+import com.mikepenz.iconics.typeface.library.community.material.CommunityMaterial
+import io.homeassistant.companion.android.R
+import io.homeassistant.companion.android.common.data.integration.Entity
+import io.homeassistant.companion.android.home.HomeActivity
+import io.homeassistant.companion.android.home.HomePresenter
+import io.homeassistant.companion.android.util.CommonFunctions
+import io.homeassistant.companion.android.viewModels.EntityViewModel
+
+class SettingsComposable {
```
JBassett, line 39: Again if these are passing everything via function parameters should this be a class?

JBassett (changes requested):

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
dshokouhi: yea good point :)

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
dshokouhi: sure, thank you :)

wear/src/main/java/io/homeassistant/companion/android/settings/SettingsComposable.kt
dshokouhi: almost missed this one 😛  thank you

JBassett (approved):

Issue #1855, opened 2021-10-30
BLE Transmitter stopped advertising


schildgenb:
**Home Assistant Android version:**
beta-1594-e881a8a-full
**Android version:**
12
**Phone model:**
Pixel 3XL
**Home Assistant version:**
2021.10.6
**Last working Home Assistant release (if known):**
202110.6
**Description of problem:**
BLE Transmitter has stopped advertising. This stopped with 2021.10.0 version of Android. My wife's phone stopped as well. I looked through the logs and found this 

10-29 23:23:03.281 16400 16400 D BluetoothAdapter: isLeEnabled(): ON

10-29 23:23:03.288 16400 16400 E BeaconTransmitter: Advertisement start failed, code: 3

10-29 23:23:04.368 16400 16588 D

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```

**Screenshot of problem:**

**Additional information:**

dshokouhi:
are you using esphome? I think there is an important update. This bug is very similar to https://github.com/home-assistant/android/issues/1764 which has the same error message. The error code is actually saying the service has already started. There were some changes done to the output of the transmitter so you may need to adjust your settings on the receiving end.

schildgenb:
I am using ESPresense for room tracking. I tried updating the settings
there but still nothing. I have even downloaded scanners and none of them
would pick up my wife's phone as they had before the last update.

dshokouhi:
> I am using ESPresense for room tracking. I tried updating the settings there but still nothing. I have even downloaded scanners and none of them would pick up my wife's phone as they had before the last update.


Did you also make sure to adjust the sensor settings as the power output had changed?

@Alfiegerner FYI in case you didn't see this 🙂

schildgenb:
I adjusted all of the settings to try different various things on both the
app and ESPresense. I tried the lastest beta as well and nothing shows up.
I download beacon simulator just to make sure it was not my phone but that
worked straight away. I downloaded scanner apps on my wife's phone but
nothing shows up besides the beacon simulator.

mloebl:
I'm having the same issue, BLE/ESPresense works great with BeaconScope, but no longer working for me at all with the HA Android app starting this week. Using beta-1575-e1cc6c8-full

kmalinich:
Same issue here. I'm on beta-1600-c6a5231-full

Alfiegerner:
[...]

PR #1876, opened 2021-11-05
Request proper android 12 permissions for BLE transmitter

dshokouhi:
## Summary

Fixes: #1855 by requesting a new permission for users on Android 12 as it is now a requirement.


Alfiegerner:
Thank you for looking at this @dshokouhi , was on my list for the weekend, appreciate you covering it off.

dshokouhi:
No problem when I saw the permission was part of Android 12 I realized it was something that should've been done as part of my earlier PR.

dshokouhi:
@Alfiegerner looking at the feedback in the bug we may actually still need your help :( I was hopeful this was all we needed but not sure if we need more changes or if these are not correct? Its very strange but admittedly I don't have a way of testing this as I do not use the feature.  If you end up needing to add these permissions to your fix I will happily close this PR :)

PR #1884, opened 2021-11-07
Initial WebSocket Implementation

JBassett:
## Summary

Adding initial support for the Web Socket API.  For now the only service calls that have been moved over are getting the config, get services, and getting all states.  A future PR will add support for subscribing to events via the websocket.

## Screenshots

N/A

N/A


common/src/main/java/io/homeassistant/companion/android/common/data/HomeAssistantApis.kt
```diff
@@ -0,0 +1,86 @@
...
+        var cookieManager: CookieManager? = null
+        try {
+            cookieManager = CookieManager.getInstance()
+        } catch (e: Exception) {
+            // Noop
+        }
+        if (cookieManager != null) {
+            builder.cookieJar(CookieJarCookieManagerShim())
+        }
+        builder.callTimeout(30L, TimeUnit.SECONDS)
```
nick2525, line 65: Better to put in constants in this file

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
```diff
@@ -0,0 +1,195 @@
...
+import kotlinx.coroutines.withTimeout
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.Response
+import okhttp3.WebSocket
+import okhttp3.WebSocketListener
+import okio.ByteString
+import javax.inject.Inject
+
+class WebSocketRepositoryImpl @Inject constructor(
```
nick2525, line 30: maybe put WebSocket handle in separate gradle module?

JBassett: Once we get the library totally flushed out internally I can see us extracting it into it's own repo but for now I'm just as content leaving it tightly coupled.  As we add more we are going to run into issues that are just easier to tackle while it's internal.

common/src/main/java/io/homeassistant/companion/android/common/data/HomeAssistantApis.kt
JBassett: Done

dshokouhi (approved):
Ran a quick test on my phone and didn't seem to see an issue here in any place that we use `getServices` and `getEntities` all seems to be working well :) Great job!!!

Issue #1880, opened 2021-11-07
Wear OS - expand/collapse category lights, scripts, swithes, etc.


HunterX86:
**Is your feature request related to a problem? Please describe.**
I have to scroll through all the entities to turn on for example a light.

**Describe the solution you'd like**
The option to collapse or expand category lights, scripts, swithes, etc.

**Describe alternatives you've considered, if any**

**Additional context**

Issue #1885, opened 2021-11-07
Wear OS - Ability to re-order categories


brewston:
Is your feature request related to a problem? Please describe.
Ease of use.

Describe the solution you'd like :

In settings, the ability to re-order categories. eg be able to place Scenes first, Switches second etc

Describe alternatives you've considered, if any : currently the categories are alphabetical only.

Additional context

dshokouhi:
Re-order and maybe even allow to hide certain categories too?

brewston:
Yes, I think being able to hide catergories would be good too.

dshokouhi:
So we had some discussion here and we think a better user experience would be move to the loading of all entities to another screen. What do you think about applying this request to the favorites section? That means we would split the favorites by the category which you can then move around with this request. This would allow for us to load those entities faster so the initial launch experience is better on the home screen.  To check out the new favorites section you would need to be on the latest beta which is not in the play store yet.

brewston:
So will favorites have sub-sections too ? I was expecting just a short list of entities. I think I'll wait and see the next beta this weekend, if that's OK ? Thanks.

leroyboerefijn:
@dshokouhi Any ideas on the best flow for this in settings? On the phone I would propose a drag-drop list, but on wear I'm not sure if that works well. We could also do something similar to the flow for the shortcuts Tile, but it might be a bit limited for reordering categories...

dshokouhi:
> So will favorites have sub-sections too ? I was expecting just a short list of entities. I think I'll wait and see the next beta this weekend, if that's OK ? Thanks.

for now its one section but yes we can split it up to be sub-sections.



> Any ideas on the best flow for this in settings? On the phone I would propose a drag-drop list, but on wear I'm not sure if that works well. We could also do something similar to the flow for the shortcuts Tile, but it might be a bit limited for reordering categories...

Since we are defining the categories shown I was thinking something like: first category, second category etc... lol

dshokouhi:
The loading of all categories has been moved to another screen now, re-ordering favorites will be possible in #1963 closing this request as the home screen has changed.

PR #1886, opened 2021-11-07
Allow categories in Wear to be collapsed and expanded for easier navigation

dshokouhi:
## Summary

Fixes: #1880  For now I opted not to save this to local storage because its probably something that should be in Settings and completed alongside #1885 if a user really wants input booleans collapsed on each time we load its probably better for them to keep it at the bottom or maybe even hidden?

I also switched to a `ListHeader` for the entities as it accounts for proper title spacing and arrangement.

## Screenshots

image

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
@@ -228,7 +228,7 @@ class HomeActivity : ComponentActivity(), HomeView {
                                                     Chip(
                                                         modifier = Modifier
                                                             .fillMaxWidth()
-                                                            .padding(top = if (index == 0) 30.dp else 10.dp),
+                                                            .padding(top = if (index == 0) 0.dp else 10.dp),
                                                         icon = {
                                                             Image(
                                                                 asset = getIcon(
--- a/wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
@@ -6,15 +6,19 @@ import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.viewModels
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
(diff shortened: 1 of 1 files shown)


wear/src/main/java/io/homeassistant/companion/android/viewModels/EntityViewModel.kt
```diff
@@ -9,4 +9,9 @@ import io.homeassistant.companion.android.common.data.integration.Entity
 class EntityViewModel : ViewModel() {
 
     var entitiesResponse: Array<Entity<Any>> by mutableStateOf(arrayOf())
+    var expandedInputBooleans: Boolean by mutableStateOf(true)
+    var expandedLights: Boolean by mutableStateOf(true)
+    var expandedScenes: Boolean by mutableStateOf(true)
+    var expandedScripts: Boolean by mutableStateOf(true)
+    var expandedSwitches: Boolean by mutableStateOf(true)
```
JBassett, line 16: This is bad, these shouldn't be in the view model.  View model is only for data not the view state.

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
```diff
@@ -161,39 +164,68 @@ class HomeActivity : ComponentActivity(), HomeView {
...
+                                    id = R.string.input_booleans,
+                                    expanded = entityViewModel.expandedInputBooleans
+                                )
+                            }
+                            if (entityViewModel.expandedInputBooleans) {
+                                items(inputBooleans.size) { index ->
+                                    SetEntityUI(inputBooleans[index], index)
+                                }
                             }
                         }
```
JBassett, line 178: These should use state hoisting: https://developer.android.com/jetpack/compose/state#state-hoisting

JBassett (changes requested):

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
dshokouhi: ok i think its already doing that now with the proper values? it does work :)

dshokouhi: one other thing to consider here is that `item {}` needs to be used within the lazy list itself. I can't move it out to another composable to handle it.

PR #1756, opened 2021-10-06
Add wear sensors

Kisty:
## Summary

Adding wearOS device sensors. Starting with battery. This is a work in progress.

## Link to pull request in Documentation repository

Documentation: home-assistant/companion.home-assistant#

## Any other notes

#1613

Issue #1613, opened 2021-07-17
Add sensors to Wear OS


Kisty:
**Is your feature request related to a problem? Please describe.**

I would like to be able to be notified when the Wear OS smartwatch is low on battery.

**Describe the solution you'd like**

Add the battery sensor to the wear app and perhaps others? Unsure what sensors but if I'm adding battery it would be cool to add more sensors like the phone has.

**Describe alternatives you've considered, if any**

I've heard that Tasker and AutoWear achieves something similar but would be great to have a first-party solution.

**Additional context**

I'm an Android developer by trade so would be happy to look in to this.

dshokouhi:
As of now the wear OS app just has a placeholder image. We want the app to be a standalone so it will need work manager and the likes added to the app for sensor functionality.  With that said if you have the time and want to contribute please feel free to.  If you have any questions you can reach out in the #devs_mobile_apps discord.  I agree with you, the more sensors the better :)

Kisty:
Thanks. OK, will use Discord. Am getting there with a working example but there's a bit of plumbing to do.

Started by copying the `BatterySensorManager` and appended all the sensors with `_wear` and using Play services Wearable API to query for info. It won't be perfect but will be working.

As phones can have multiple wearables, I think for this iteration I'll just track the first one.

I hear what you say about using `WorkManager`. I think `WearableListenerService` will run in it's own process but may have to do it on another thread instead. `WearableListenerService` from `play-services-wearable` is the way you communicate to wearables (since they don't have to be connected to the phone all the time.

Am thinking about either getting the wearable to do a periodic sync to mobile app then hook in to the `SensorManager` API and piggy back off that. Would I be able to create the sensor though as a different prefix instead of the mobile device prefix for the `entity_id`?

dshokouhi:
> Am thinking about either getting the wearable to do a periodic sync to mobile app

So the decision was actually for the wearable app to be a standalone meaning it should not rely on sending data to the app itself rather to the users HA server directly.  It should do auth and register the wearable device just like the android phone does.  It should not rely on the phone app to do anything honestly, the phone app could be made to help make something easier like managing sensor and tiles but it should still function on its own without the need for an app to enable sensors and send sensor data back home.

Chaphasilor:
@Kisty #1691 just got merged, so you should be able to integrate the auth into your PR / existing work now :D  

[...]

PR #1890, opened 2021-11-08
Add Wear OS battery sensors

Kisty:
## Summary

Adding Wear OS device sensors. Starting with battery. This is a work in progress.

- [x] Documentation: home-assistant/companion.home-assistant#619

## Any other notes

#1613 #1756

- [x] Hoist database to `common` module
- [x] Hoist base sensors API to `common` module
- [x]  Add battery sensor
- [x] Add notes to Companion documentation repo


app/src/full/java/io/homeassistant/companion/android/location/HighAccuracyLocationService.kt
```diff
@@ -125,7 +125,7 @@ class HighAccuracyLocationService : Service() {
             val disablePendingIntent = PendingIntent.getBroadcast(context, 0, disableIntent, PendingIntent.FLAG_MUTABLE)
 
             notificationBuilder = NotificationCompat.Builder(context, channelID)
-                .setSmallIcon(R.drawable.ic_stat_ic_notification)
+                .setSmallIcon(io.homeassistant.companion.android.common.R.drawable.ic_stat_ic_notification)
```
Kisty, line 128: I'm not too keen on this, but the build works. Had to move some resources to the `common` module but you have to reference the resources from the `common` module's `R` file. Any idea how to refererence better? Some options:
a) add `common/main/res` to `app` and `wear` sourcesets
b) maybe use `ids.xml` to reference the library resources as if it's in their own `res` directory
c) use import alias (should be an implementation detail and something that maintainers should need to worry about)

Kisty: I opted for b) but didn't need the package name in the end. Hope it helps in the future

Kisty:
OK, I think I've got the battery sensor working with an emulator and dev HA instance. Can someone give it a try? Going to see what other sensors I can bring in

dshokouhi:
> OK, I think I've got the battery sensor working with an emulator and dev HA instance. Can someone give it a try? Going to see what other sensors I can bring in

As soon as you resolve the conflicts I believe we can get the actions started so we can test the APK :)  Apologize for all the conflicts, I know how annoying they can become :(

I would be happy to test this :)

Kisty:
> > OK, I think I've got the battery sensor working with an emulator and dev HA instance. Can someone give it a try? Going to see what other sensors I can bring in
> 
> As soon as you resolve the conflicts I believe we can get the actions started so we can test the APK :) Apologize for all the conflicts, I know how annoying they can become :(
> 
> I would be happy to test this :)

Been trying to keep it with upstream but it's a little tricky. I see Dagger's been replaced by Hilt, which is cool in theory. Did you have any issues with injection for a class that had a base class? Reason I ask is that getting Dagger to inject from the base class was hard work, but got there (didn't generate the Daggerised class of `SensorReceiver` when the Sensor component expected it). 

Will it just work?

dshokouhi:
> Been trying to keep it with upstream but it's a little tricky. I see Dagger's been replaced by Hilt, which is cool in theory. Did you have any issues with injection for a class that had a base class? Reason I ask is that getting Dagger to inject from the base class was hard work, but got there (didn't generate the Daggerised class of `SensorReceiver` when the Sensor component expected it).
> 
> Will it just work?

There have been some odd quirks I noticed like inside the NFC fragments they were all crashing because the parent activity was missing `@AndroidEntryPoint`. I think @JBassett will be able to help as I am still trying to get my footing around with Hilt. I do see its easier to use in some cases like adapting it in #1908 took no time.  What I did there was look at some of the changes in #1933 and try to mimic what was done there. Maybe that will be of help :)

Kisty:
> > Been trying to keep it with upstream but it's a little tricky. I see Dagger's been replaced by Hilt, which is cool in theory. Did you have any issues with injection for a class that had a base class? Reason I ask is that getting Dagger to inject from the base class was hard work, but got there (didn't generate the Daggerised class of `SensorReceiver` when the Sensor component expected it).
> > Will it just work?
> 
> There have been some odd quirks I noticed like inside the NFC fragments they were all crashing because the parent activity was missing `@AndroidEntryPoint`. I think @JBassett will be able to help as I am still trying to get my footing around with Hilt. I do see its easier to use in some cases like adapting it in #1908 took no time. What I did there was look at some of the changes in #1933 and try to mimic what was done there. Maybe that will be of help :)

Thanks man. That's cool. OK, so the original battery state is reported OK, but changing the emulator battery level doesn't get reported back to HA. Any ideas? How does the sensor worker get fired exactly?

dshokouhi:
> Thanks man. That's cool. OK, so the original battery state is reported OK, but changing the emulator battery level doesn't get reported back to HA. Any ideas? How does the sensor worker get fired exactly?

So getting each and every battery % requires us to register for `ACTION_BATTERY_CHANGED` which in my experience is the most chatty intent ever. I once tried to only update the state if the percentage changed but because it triggers so often I was still seeing multiple calls registering every second it had an update so it was not viable.  Try to test with the charger intents, those should trigger immediately and when they trigger it should also update the battery level at the same time.  The worker itself uses a 15 minute update interval which is the fastest that it accepts.

dshokouhi:
Sorry just realized I didn't completely answer your question about Sensor Worker itself.  So the sensor worker runs on the 15 minute interval where it checks in with each of the `managers` that are added to it. 

In addition to this we also register intent receivers to get updates like when the user changes ringer mode, changes DND, wifi change, the phone etc...  For this some are registered in the manifest as they are required to:

https://github.com/home-assistant/android/blob/master/app/src/main/AndroidManifest.xml#L63-L76

And others are registered in our Application class:

https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt#L46-L128

For instant updates we need the corresponding intent for the sensor and unfortunately not all have an intent, otherwise the normal update interval will kick in and all sensors are checked for an update

In cases where you need to force an update in the code you can use `SensorWorker.Start(context)` and it will trigger an update to HA immediately.  We do this in a few places like the sleep activity sensor:

https://github.com/home-assistant/android/blob/master/app/src/full/java/io/homeassistant/companion/android/sensors/ActivitySensorManager.kt#L133

leroyboerefijn:
Great work here @Kisty ! I'd be happy to test as well 🙂

Just wanted to chime in on the sensor updates. I think we should be careful, since Wear OS is a bit more battery limited. If I remember correctly, it also doesn't have a network connection constantly for that reason. Therefore, doing network actions in the background might even consume a bit more power since it needs to establish the connection first.

Just wanted to bring this up. To me 15 minute intervals sound fine. And of course intents for sensors that require immediate updating still make sense if they don't update too frequently. I can imagine that listening to a charger connected/disconnected intent would be nice for example!

dshokouhi:
> Just wanted to chime in on the sensor updates. I think we should be careful, since Wear OS is a bit more battery limited. If I remember correctly, it also doesn't have a network connection constantly for that reason. Therefore, doing network actions in the background might even consume a bit more power since it needs to establish the connection first.

So the SensorWorker itself also has constraints in place for network connection in particular, it won't attempt to do anything if there is no active connection :)

https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/sensors/SensorWorker.kt#L40

> And of course intents for sensors that require immediate updating still make sense if they don't update too frequently. I can imagine that listening to a charger connected/disconnected intent would be nice for example!

This why its nice to give users an option for what they want to enable so they can decide if its too much. The battery intents are not going to cause a lot of concern but I can see things like Network state causing issues because it triggers when switching from wifi <-> cell so I can see that happening when a watch goes from bt <-> wifi. By default we only enable the battery sensors which won't cause much of a drain :)

Also the worker won't send an update if the state has not changed since the last update so we have somethings in place to alleviate some concerns :)

Kisty:
For the documentation. What would be expected? I see sensors are already mentioned in core. Does something need to be mentioned that you can add sensors?

Ah, that means I need to add UI to control what sensors are enabled...

So I wasn't waiting 15 mins to test the monitor, hence why I wasn't seeing updates.

Kisty:
Hmmm, unsure whether to bundle the other Wear OS sensors in this PR. Thoughts? I can see it could be relatively trivial to get the other sensors working, though will need to test them. I could concentrate on just getting the battery sensor working first...

dshokouhi:
> For the documentation. What would be expected? I see sensors are already mentioned in core. Does something need to be mentioned that you can add sensors?

For now we can just add a short section in the existing Wear OS page. I think for the first release we are just doing battery sensors right? If so then we can link to the sensors anchor page since all the information is still the same as the API's are the same.



> Ah, that means I need to add UI to control what sensors are enabled...

Lets get a settings section for enabling the sensors in the watch, the phone side can be done in another PR. Probably good to get a good base for the initial sensor first so adding others in the future is more streamlined :)



> Hmmm, unsure whether to bundle the other Wear OS sensors in this PR. Thoughts?

Lets do a separate PR for the others as this one is already pretty big and will take some time to review :)

wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
```diff
@@ -86,6 +87,16 @@ class HomeActivity : ComponentActivity(), HomeView {
...
     }
 
+    override fun onResume() {
+        super.onResume()
+        SensorWorker.start(this)
+    }
+
+    override fun onPause() {
+        super.onPause()
+        SensorWorker.start(this)
```
leroyboerefijn, line 97: Shouldn't you stop it here?

dshokouhi: No you dont want to stop it as its a background service, otherwise the sensors wont continue to update in the background.  The reason why the WebviewActivity has this call in `onPause` as well as `onResume` is because it allows the App Importance sensor to update letting the user know when the app is open or not.

leroyboerefijn: Thanks for the explanation, never mind then!

Kisty: Incidentally, are there any other places where it would be appropriate to put the `SensorWorker.start()` calls? I see in the mobile app the location and other Google Play Services sensors use it. Thought it was a good place to put it. 

Without the call would account for why the sensors didn't get updated! 🤦‍♂️

dshokouhi: So we want it once when the app starts and then after that anytime you need to force an update like in the case of app importance we needed to send an update when the app is paused. It depends on the situation really. The manifest intents will trigger an update to the receiver directly but won't start the worker on its schedule which is what we need to do in this activity

Kisty: Ah, so I might need to update the `AndroidManifest.xml` to get `SensorReceiver` to receive specific intents.

dshokouhi: Only certain intents can be registered in the manifest others will need to be a runtime receiver. What we have in the manifest today is what can be there.

Kisty:
> > Ah, that means I need to add UI to control what sensors are enabled...
> 
> Lets get a settings section for enabling the sensors in the watch, the phone side can be done in another PR. Probably good to get a good base for the initial sensor first so adding others in the future is more streamlined :)

So would you expect a settings page similar to the mobile app in this PR?

JBassett:
> > > Ah, that means I need to add UI to control what sensors are enabled...
> > 
> > 
> > Lets get a settings section for enabling the sensors in the watch, the phone side can be done in another PR. Probably good to get a good base for the initial sensor first so adding others in the future is more streamlined :)
> 
> So would you expect a settings page similar to the mobile app in this PR?

At this point I would stop adding features, let's get what you have mergeable. Then we can go on to add a UI to it.

dshokouhi:
> At this point I would stop adding features, let's get what you have mergeable. Then we can go on to add a UI to it.

I agree, when sensors were first introduced to the phone app there was no UI for it :) Sorry for the potential scope creep @Kisty

Kisty:
Didn't realise there were merge conflicts. Will sort them out

Kisty:
https://github.com/Kisty/home-assistant-android/blob/a8076b7d89d5ff77507f829fe650bf6ae2b12da5/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt#L56

I see this action to update sensors isn't being referenced anymore. Have I lost something in the PR or is that true for `master`?

common/src/main/java/io/homeassistant/companion/android/common/sensors/BatterySensorManager.kt
```diff
@@ -16,38 +16,32 @@ class BatterySensorManager : SensorManager {
             R.string.basic_sensor_name_battery_level,
             R.string.sensor_description_battery_level,
             "battery",
-            "%",
-            stateClass = SensorManager.STATE_CLASS_MEASUREMENT,
-            entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC
```
dshokouhi, line 21: we should keep these instead of removing :)

Kisty: Oops! Sorry

dshokouhi:
So it looks like the debug APK is ready. For some reason I don't see any sensor entities being added to HA but I do see the worker saying it was successful in the logs.  One thing that I noticed is that we are not registering for the battery intents like we do in the phone app so the instant updates for battery connected/disconnected are not triggering.

https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt#L50-L58

The phone APK seems to be working though but I am not sure if the APK is correct because of the ktlint failure?

Kisty:
> So it looks like the debug APK is ready. For some reason I don't see any sensor entities being added to HA but I do see the worker saying it was successful in the logs. 

Strange. Will double check here.

> One thing that I noticed is that we are not registering for the battery intents like we do in the phone app so the instant updates for battery connected/disconnected are not triggering.
> 
> https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt#L50-L58
Done

> The phone APK seems to be working though but I am not sure if the APK is correct because of the ktlint failure?

That's good to hear. Where is the debug wear APK from?

dshokouhi:
> Edit2: Still nothing. How does the mobile app register for sensors initially? Must be something I'm missing.

That logic is in the sensor receiver where we check the app version, core version and registration. If you enable debug logging for the `mobile_app` integration you can see all the webhook data coming in like when we register and send updates. So after we login and hit the homescreen we should have the initial sensor state registered. If you look at onboarding for the phone app you can see how it should behave.

Also when you call SensorWorker.start() it will trigger an immediate update and replace the schedule from that point on.

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
```diff
@@ -0,0 +1,177 @@
...
+
+abstract class SensorReceiverBase : BroadcastReceiver() {
+    private val ioScope: CoroutineScope = CoroutineScope(Dispatchers.IO + Job())
+
+    protected abstract val tag: String
+    protected abstract val currentAppVersion: String
+    protected abstract val managers: List<SensorManager>
+
+    @Inject
+    lateinit var integrationUseCase: IntegrationRepository
```
dshokouhi, line 23: So I tested the latest APK and I did not see any sensors get created. Then I took my watch off the charger and got a crash that led to this line. I believe we may need to add `@AndroidEntryPoint` above line 15 because we are injecting here.  Not 100% on that though.

```
2021-11-26 00:50:49.670 18559-18665/io.homeassistant.companion.android.debug E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-1
    Process: io.homeassistant.companion.android.debug, PID: 18559
    kotlin.UninitializedPropertyAccessException: lateinit property integrationUseCase has not been initialized
        at io.homeassistant.companion.android.common.sensors.SensorReceiverBase.getIntegrationUseCase(SensorReceiverBase.kt:23)
        at io.homeassistant.companion.android.common.sensors.SensorReceiverBase$onReceive$2.invokeSuspend(SensorReceiverBase.kt:80)
```

Strange that this error occurs when our registered intent is fired but the worker itself says its successful without registering any sensors.

```
2021-11-26 00:52:22.968 19057-19087/io.homeassistant.companion.android.debug I/WM-WorkerWrapper: Worker result SUCCESS for Work [ id=03ca93f0-c6bd-4bbb-b678-0410d67ebc77, tags={ io.homeassistant.companion.android.sensors.SensorWorker } ]
```

Kisty: Hmm, so it looks that somethings changed that I haven't incorporated in the wear app in the last 2 weeks. I think I've now got a working imlementation which was taken from the mobile app as part of the check for disabled location. Calling `SensorManager#isEnabled()` seems to be the only way to add the sensors to the DB.

https://github.com/home-assistant/android/pull/1890/commits/0d6660952311433973f2aa3f9a443c47a6c291b3#diff-e4be6176c8e8184dc31ba7addb801b46028bbaa439cdd48b80183296dc96d6f3R59

dshokouhi: The biggest impactful update I can see is either the Entity Category PR or the Hilt upgrade PR as both touched sensors and how we register them.  I am not sure about why calling `SensorManager.isEnabled()` is the only way to get them registered because we have `enabledByDefault` to keep track of what should be enabled upon onboarding.

Kisty: What I mean is that to acutally add the sensors to the database, it seems that the only way is through the `SensorManager#isEnabled()` call. Although they are enabled by default, the sensors aren't added to the DB so there's nothing to check! The worker conks out becuase there are no sensors in the database, thus no enabled sensors so it's all like, there's nothing to do here. Do you get me?

https://github.com/Kisty/home-assistant-android/blob/0d6660952311433973f2aa3f9a443c47a6c291b3/common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorManager.kt#L60-L64

https://github.com/Kisty/home-assistant-android/blob/0d6660952311433973f2aa3f9a443c47a6c291b3/common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorWorkerBase.kt#L37

dshokouhi: Ah yes you are right about that. I thought we used to do this in SensorReceiver but I think we moved it and I had forgotten about that, oops LOL. The only places we use `sensorDao.add(sensorId)` is in the location handler, onboarding for location sensors and  in the sensor fragments as a sensor gets enabled.  Good catch!  Looking forward to testing once the actions run :)

Kisty: OK, cool. Are you able to approve the actions?

Would it be worth (maybe in another PR) to add the sensors somewhere else explicitly or leave it as-is?

dshokouhi: >  Are you able to approve the actions?

Im not able to :( otherwise I would've approved them :P

> Would it be worth (maybe in another PR) to add the sensors somewhere else explicitly or leave it as-is?

umm lets see what comes of the review to see if it should be moved. For now i think its ok since users will come to the home screen after logging in.  Later on when we add any sensor that requires location permissions we probably want to start using the location handler anyways

dshokouhi: ok the actions ran and after logging in I am able to see the sensors in HA and the intents are working!

image

image

wear/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
```diff
@@ -1,7 +1,39 @@
...
+        // This will cause interactive and power save to update upon a state change
+        registerReceiver(
+            sensorReceiver,
+            IntentFilter().apply {
+                addAction(Intent.ACTION_SCREEN_OFF)
+                addAction(Intent.ACTION_SCREEN_ON)
+                addAction(PowerManager.ACTION_POWER_SAVE_MODE_CHANGED)
+            }
+        )
+    }
```
dshokouhi, line 38: We probably don't need these until we add the interactive and power save sensors, otherwise we will trigger a lot of unwanted updates :)

wear/src/main/AndroidManifest.xml
```diff
@@ -18,6 +18,21 @@
...
+            <intent-filter>
+                <action android:name="android.intent.action.BOOT_COMPLETED" />
+                <action android:name="android.intent.action.MY_PACKAGE_REPLACED" />
+                <action android:name="android.intent.action.MY_PACKAGE_SUSPENDED" />
+                <action android:name="android.intent.action.MY_PACKAGE_UNSUSPENDED" />
+                <action android:name="android.intent.action.QUICKBOOT_POWERON" />
+                <action android:name="com.htc.intent.action.QUICKBOOT_POWERON" />
+                <action android:name="android.app.action.NEXT_ALARM_CLOCK_CHANGED" />
+                <action android:name="android.bluetooth.device.action.ACL_CONNECTED" />
+                <action android:name="android.bluetooth.device.action.ACL_DISCONNECTED" />
```
dshokouhi, line 33: We should probably register for these once we add the corresponding sensors

dshokouhi (commented):
Just a couple of small comments on specific sensor intents :)

Also not sure if we want to leave all the commented out code in the Wear SensorReceiver although I understand we are probably going to add most if not all to the wear app in future PRs. Well whichever ones will work :P

wear/src/main/AndroidManifest.xml
Kisty: Good catch. Shall I remove it for now?

Kisty:
> Also not sure if we want to leave all the commented out code in the Wear SensorReceiver although I understand we are probably going to add most if not all to the wear app in future PRs. Well whichever ones will work :P

Thanks, I think to keep things tidy, I'll remove instead of commenting code out.

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
Kisty: That's fantastic! It's a long awaited feature. For it to work for someone else is great! Going to be a big help for a bunch of users!

wear/src/main/AndroidManifest.xml
JBassett: Yes please!

wear/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    override val currentAppVersion: String
+        get() = BuildConfig.VERSION_NAME
+
+    override val managers: List<SensorManager>
+        get() = MANAGERS
+
+    companion object {
+        const val TAG = "SensorReceiver"
+        val MANAGERS = listOf(
```
JBassett, line 24: For now lets remove these rather than comment out.

wear/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
```diff
@@ -0,0 +1,71 @@
...
+        )
+
+        const val ACTION_REQUEST_SENSORS_UPDATE =
+            "io.homeassistant.companion.android.background.REQUEST_SENSORS_UPDATE"
+    }
+
+    // Suppress Lint because we only register for the receiver if the android version matches the intent
+    @SuppressLint("InlinedApi")
+    override val skippableActions = mapOf<String, String>()/*
+        "android.app.action.NEXT_ALARM_CLOCK_CHANGED" to NextAlarmManager.nextAlarm.id,
```
JBassett, line 58: For now lets remove these rather than comment out.

JBassett (changes requested):

JBassett:
Nice work @Kisty!  Only those small things and I think we can get this merged!!

common/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
Kisty: What we could do is move all the strings to `common` which would mean that you wouldn't need to worry about what strings go in what modules.

If you see in some of my earlier commits, I managed to bring strings to `common` so that they can be used as if they're in the same module by aliasing the strings in `aliases_common.xml`. Just thought it would be nice when new code is written and you don't need to dig around finding the `R.string.sensors` which is in `common.R` file rather than `app.R` or whatever.

The downside with aliasing is that all strings that are moved to `common` would need to be aliased one-by-one.

On the other hand, if all strings were moved to `common`, then no aliasing would be needed.

wear/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
Kisty: Done

wear/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
Kisty: Done

wear/src/main/AndroidManifest.xml
Kisty: Done!

dshokouhi (approved):
Looks good to me!

JBassett (approved):
@Kisty once we fix the ktlint error I'll merge!

Kisty:
Ah, poop. I forgot to run the ktlintFormat!

Issue #1878, opened 2021-11-05
Wear OS - Show a toast notification and vibrate when pressing an item in wear os app


HunterX86:
**Is your feature request related to a problem? Please describe.**
When pressing an item, i don't know if for example a light is turned on or i run a script.

**Describe the solution you'd like**
Show a toast message and vibrate (haptic feedback) when pressing an item.

**Describe alternatives you've considered, if any**
[Android Toast](https://developer.android.com/guide/topics/ui/notifiers/toasts)
[Haptic Feedback](https://developer.android.com/reference/android/view/HapticFeedbackConstants)

**Additional context**

dshokouhi:
Nice request! It looks like some of these are also mentioned in the material design guidelines. We should also show the time at the top.  I wonder if for haptic feedback if we should make it a setting users can enable/disable?

Also there is a new update coming where the chips are converted to toggles so that should help reflect the state in a better way.

https://developer.android.com/training/wearables/design/overlays#time

leroyboerefijn:
I especially like the haptic feedback. Would it make sense to follow the system settings for haptics? If at all possible.

For toasts, I'm not personally a fan. But as a setting, why not!

Finally for the clock text, there should be a compose item for it ;)

dshokouhi:
> Would it make sense to follow the system settings for haptics? If at all possible.

I think it should, is there a system setting for this? I didn't see it on my watch but we would need the vibrate permission

leroyboerefijn:
On my Galaxy watch 4 I have a system vibrations toggle. No idea if we could even use that. But on second thought, maybe it makes more sense as an app setting?

dshokouhi:
yea i think an app setting makes more sense, especially since I dont see a global setting on my Fossil watch lol.  Probably a good idea to leave it off by default so it matches current behavior?

HunterX86:
Here is an example that can be implemented to show a toast for the chips
`chip_message.setOnClickListener {Toast.makeText(context, "Message", Toast.LENGTH_SHORT).show()}`

dshokouhi:
@HunterX86 we have actually switched to Compose so the function call is a bit different. Right now we are trying to figure out what a good default value is. I think we may leave it all off and let user turn on in settings so the functionality is the same. If we were to leave any on maybe the haptic feedback because when I tested the toast message it takes up the whole screen so not sure how desirable that is but the haptic feedback makes sense.

PR #1893, opened 2021-11-09
Add settings for wear haptic feedback and toast on entity selection

dshokouhi:
## Summary

Fixes: #1878 by providing 2 new settings that default to `false` to preserve existing behavior. Once enabled the user will see a toast message and/or haptic feedback based on their choosing.

## Screenshots

image

image

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
@@ -1,6 +1,7 @@
 package io.homeassistant.companion.android.home
 
 import androidx.compose.runtime.mutableStateListOf
+import androidx.compose.runtime.mutableStateOf
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import io.homeassistant.companion.android.common.data.integration.Entity
@@ -20,11 +21,17 @@ class MainViewModel : ViewModel() {
         private set
     var favoriteEntityIds = mutableStateListOf<String>()
         private set
+    var isHapticEnabled = mutableStateOf(false)
+        private set
+    var isToastEnabled = mutableStateOf(false)
+        private set
 
     fun loadEntities() {
         viewModelScope.launch {
             favoriteEntityIds.addAll(homePresenter.getWearHomeFavorites())
             entities.addAll(homePresenter.getEntities())
+            isHapticEnabled()
+            isToastEnabled()
         }
     }
 
@@ -59,6 +66,28 @@ class MainViewModel : ViewModel() {
         viewModelScope.launch {
             favoriteEntityIds.clear()
             homePresenter.setWearHomeFavorites(favoriteEntityIds)
+        }
+    }
+
+    fun isHapticEnabled() = viewModelScope.launch {
+        isHapticEnabled = mutableStateOf(homePresenter.getWearHapticFeedback())
+    }
+
+    fun isToastEnabled() = viewModelScope.launch {
+        isToastEnabled = mutableStateOf(homePresenter.getWearToastConfirmation())
+    }
+
+    fun setHapticEnabled(enabled: Boolean) {
+        viewModelScope.launch {
+            homePresenter.setWearHapticFeedback(enabled)
+            isHapticEnabled = mutableStateOf(enabled)
+        }
+    }
+
+    fun setToastEnabled(enabled: Boolean) {
+        viewModelScope.launch {
+            homePresenter.setWearToastConfirmation(enabled)
+            isToastEnabled = mutableStateOf(enabled)
         }
     }
 
--- a/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
@@ -26,12 +26,12 @@ class MainViewModel : ViewModel() {
     var isToastEnabled = mutableStateOf(false)
         private set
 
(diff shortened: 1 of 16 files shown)


wear/src/main/java/io/homeassistant/companion/android/home/views/EntityUi.kt
```diff
@@ -24,15 +27,17 @@ import io.homeassistant.companion.android.util.setChipDefaults
...
+    val haptic = LocalHapticFeedback.current
+    val context = LocalContext.current
     val attributes = entity.attributes as Map<*, *>
     val iconBitmap = getIcon(attributes["icon"] as String?, entity.entityId.split(".")[0], LocalContext.current)
 
     if (entity.entityId.split(".")[0] in HomePresenterImpl.toggleDomains) {
         ToggleChip(
             checked = entity.state == "on",
-            onCheckedChange = { onEntityClicked(entity.entityId) },
+            onCheckedChange = { onEntityClicked(entity.entityId, attributes["friendly_name"].toString(), haptic, context) },
```
JBassett, line 40: Haptic and toast should be done here rather than in the callback.

wear/src/main/java/io/homeassistant/companion/android/home/views/EntityUi.kt
```diff
@@ -71,7 +76,7 @@ fun EntityUi(
                 )
             },
             enabled = entity.state != "unavailable",
-            onClick = { onEntityClicked(entity.entityId) },
+            onClick = { onEntityClicked(entity.entityId, attributes["friendly_name"].toString(), haptic, context) },
```
JBassett, line 79: Haptic and toast should be done here rather than in the callback.

wear/src/main/java/io/homeassistant/companion/android/home/views/SettingsView.kt
```diff
@@ -1,66 +1,175 @@
...
+                        text = stringResource(id = R.string.irreverisble)
+                    )
+                },
+                enabled = favorites.isNotEmpty()
+            )
+        }
+
+        item {
+            val haptic = LocalHapticFeedback.current
+            var checked by rememberSaveable { mutableStateOf(isHapticEnabled) }
```
JBassett, line 108: We don't want to remember between recompositions, we should always use `isHapticEnabled`

wear/src/main/java/io/homeassistant/companion/android/home/views/SettingsView.kt
```diff
@@ -1,66 +1,175 @@
...
-            colors = ChipDefaults.primaryChipColors(
-                contentColor = Color.Black
-            ),
-            secondaryLabel = {
-                Text(
-                    text = stringResource(id = R.string.irreverisble)
+            )
+        }
+        item {
+            var checked by rememberSaveable { mutableStateOf(isToastEnabled) }
```
JBassett, line 142: We don't want to remember between recompositions, we should always use `isToastEnabled`

JBassett (changes requested):

wear/src/main/java/io/homeassistant/companion/android/home/views/SettingsView.kt
dshokouhi: so without this and even with using the viewmodel variable the checked status does not update until the page is recomposed which only happens when I navigate to another screen and back

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
```diff
@@ -61,6 +71,21 @@ class MainViewModel : ViewModel() {
...
+        return runBlocking { homePresenter.getWearToastConfirmation() }
+    }
+
+    fun setHapticEnabled(enabled: Boolean) {
+        runBlocking { homePresenter.setWearHapticFeedback(enabled) }
+    }
+
+    fun setToastEnabled(enabled: Boolean) {
+        runBlocking { homePresenter.setWearToastConfirmation(enabled) }
+    }
```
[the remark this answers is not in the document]
dshokouhi: So when I try to switch to `viewModelScope` I end up getting issues with `homePresenter` not being initialized. I think its expecting a value too quickly here?

wear/src/main/java/io/homeassistant/companion/android/home/views/SettingsView.kt
dshokouhi: so without this and even with using the viewmodel variable the checked status does not update until the page is recomposed which only happens when I navigate to another screen and back

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
```diff
@@ -20,11 +21,19 @@ class MainViewModel : ViewModel() {
...
 
-    fun loadEntities() {
+    private fun loadEntities() {
         viewModelScope.launch {
             favoriteEntityIds.addAll(homePresenter.getWearHomeFavorites())
+            isHapticEnabled.value = homePresenter.getWearHapticFeedback()
+            isToastEnabled.value = homePresenter.getWearToastConfirmation()
             entities.addAll(homePresenter.getEntities())
+            isHapticEnabled.value = homePresenter.getWearHapticFeedback()
+            isToastEnabled.value = homePresenter.getWearToastConfirmation()
```
JBassett, line 36: Don't need to call twice!

JBassett (changes requested):

JBassett (approved):

PR #1895, opened 2021-11-10
Major Wear Cleanup

JBassett:
## Summary

This PR moves a bunch of stuff around.  I tried to follow the Jetpack Compose guidelines and best practices.

## Screenshots

No changes to actual look and feel

N/A


dshokouhi (approved):
Took a look at the debug APK, everything loads fine with the exception that favorites no longer loads during the loading screen. A better way to handle this is to store things in a DB rather than local storage. In a future PR maybe we can temporarily add that call back until we get Room setup once #1890 is merged?

JBassett:
> Took a look at the debug APK, everything loads fine with the exception that favorites no longer loads during the loading screen. A better way to handle this is to store things in a DB rather than local storage. In a future PR maybe we can temporarily add that call back until we get Room setup once #1890 is merged?

I do think we can create a temporary Entity for the favorites that we pass into the EntityUi for now.

PR #1896, opened 2021-11-10
Fixing splash screen Wear OS

HunterX86:
## Summary

- Changed splash logo
- Changed background color to @color/colorAccent

## Screenshots

Added Logo with transparent background
splash_logo

N/A


SkechyWolf:
Hey instead of adding a whole new image what if we just pulled in ic_launcher_foreground? and set the width and height to around 325px 
Just an idea since we already have one with a transparent background
Looks great by the way

HunterX86:
we can do that, as long as the background is the same blue color as the logo.

dshokouhi:
> we can do that, as long as the background is the same blue color as the logo instead of black.

Any reason why this needs to be like that? The old screen was black and when I look at apps made by google that also have a splash screen the background is black behind the logo. The blue is also not the apps typical background color either.  Can you share a screenshot of how it all looks together too please?

HunterX86:
This is what i meant.
I use the accentcolor that is used in the app.
splashscreen example

dshokouhi:
@HunterX86 any particular reason why the background color needs to match the logo instead of remaining black like the rest of the background color of the app is?  Is there a material guideline I am missing here?

HunterX86:
There is not a paticulary material guideline for it, but don't you think it looks nice and clean.
If you don't agree with me, can you make it like this:
splashscreen example2

dshokouhi:
> There is not a paticulary material guideline for it, but don't you think it looks nice and clean. If you don't agree with me, can you make it like this: The background color is black and the logo color is the accentcolor.

So I am not a designer but when I look at other apps like Google Play Store they dont fill the screen entirely. I am ok with with this if everyone else is but you were quite particular about the background color which is why I enquired about it.

HunterX86:
> So I am not a designer but when I look at other apps like Google Play Store they dont fill the screen entirely. I am ok with with this if everyone else is but you were quite particular about the background color which is why I enquired about it.

As a matter of fact i am a graphic designer :smiley:

Here are 2 examples of splashscreens that fills the whole screen with a background color
[example 1](https://imagekit.androidphoria.com/wp-content/uploads/mejores-smartwatch-con-telegram.jpg)
[example 2](https://www.wareable.com/media/images/2017/02/lifesum-1486695969-rIyX-column-width-inline.jpg)

dshokouhi:
> As a matter of fact i am a graphic designer 😃

Oh excellent we are very happy to have you here!!! Thank you for the provided examples as well :)

dshokouhi (approved):

JBassett (approved):

PR #1902, opened 2021-11-11
Clean up welcome and notification details compose views

dshokouhi:
## Summary

Start to use better compose methods to follow along with #1895 

Also adds preview for notification details

Also adding a vertical scroll modifier to the columns in case the screen is small

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/notification/views/NotificationDetailView.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/notification/views/NotificationDetailView.kt
@@ -1,6 +1,5 @@
 package io.homeassistant.companion.android.settings.notification.views
 
-import android.os.Parcelable
 import android.text.Html
 import android.widget.TextView
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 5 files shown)


app/src/main/java/io/homeassistant/companion/android/settings/notification/views/NotificationDetailView.kt
```diff
@@ -0,0 +1,97 @@
...
+import java.util.GregorianCalendar
+
+@Parcelize
+data class NotificationData(val received: Long, val message: String, val data: String) : Parcelable
+
+@Composable
+fun LoadNotification(notification: NotificationItem) {
+    val notificationItem = rememberSaveable {
+        NotificationData(notification.received, notification.message, notification.data)
+    }
```
JBassett, line 36: So I'm not sure we need the `rememberSaveable` here?

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/settings/notification/views/NotificationDetailView.kt
dshokouhi: `rememberSaveable` will retain when the UI is redrawn like when the orientation changes.

https://developer.android.com/jetpack/compose/state#restore-ui-state

I think I added this in the first PR we made to pass in the state?  I can remove it if you like

JBassett: You only need the `remember` and `rememberSavable` if the state only exists inside the compose function.  Since the data is coming from outside the compose function we don't need to save it there.

JBassett (approved):

PR #1908, opened 2021-11-11
Editing wear favorites from the phone

dshokouhi:
## Summary

There is a lot going on here but I wanted to get the working code submitted early in case it can help anyone else looking to do something similar.  Also a good time to think about the layout of phone settings overall. 

- [x] Get list of entities
- [x] Get favorites from wearable
- [x] Load entities and favorites along with selected favorites via checkbox
- [x] Update favorites on wearable
- [x] Recreate app bar in compose
- [x] Re-organize code

## Screenshots

image

image

Documentation: home-assistant/companion.home-assistant#613

## Any other notes

Got a lot to do and we need to think about the design. Such as when a user has more than one wearable. How to access each of the settings etc...


leroyboerefijn:
Great work! Awesome that we're starting to be able to change the wear app settings from the phone :)
Below a few feature requests. But please let me know if you think it's better to do this in a separate PR.

- Could you display friendly names in the settings? Or would you say that is for a follow up PR?
- Now when I make changes in the settings on the phone, it doesn't automatically update. It also doesn't update when I go home and re-open the app. Would it make sense to send a broadcast in PhoneSettingsListener, which is then picked up somewhere to update the list in the MainViewModel?

wear/src/main/java/io/homeassistant/companion/android/phone/PhoneSettingsListener.kt
```diff
@@ -0,0 +1,96 @@
...
+        super.onCreate()
+        Log.d(TAG, "We have favorite message listener")
+    }
+
+    override fun onMessageReceived(event: MessageEvent) {
+        Log.d(TAG, "Message received: $event")
+        if (event.path == "/send_home_favorites") {
+            val nodeId = event.sourceNodeId
+            sendHomeFavorites(nodeId)
+        } else if (event.path == "/save_home_favorites")
```
leroyboerefijn, line 40: Do we need the "/save_home_favorites" message? If the data is put in the DataClient, we will be notified in the onDataChanged listener, right? Or is there a scenario where we do get this message but not the data changed notification?

dshokouhi:
> Great work! Awesome that we're starting to be able to change the wear app settings from the phone :)

Thank you I appreciate that! This one was a bit of a challenge to get all the pieces working together but luckily a lot of the foundation was already there from your previous PR for sending the HA instance over :)

> * Could you display friendly names in the settings? Or would you say that is for a follow up PR?

I could do that but then it might make sense to show a category saying these are Input Booleans or what not. I opted for entity ID so users know which entity it is.  Honestly I figured we could redo the `Rows` to show the icons etc... in a more friendly way 😁  

Whatever we do here though should probably be reusable so maybe making the Row a better compose element makes the most sense. We have a few places to add these to like QS tiles and Shortcuts for example. But neither of those were updated to compose yet.



> * Now when I make changes in the settings on the phone, it doesn't automatically update. It also doesn't update when I go home and re-open the app.

So updating Wear when the app is open is something I couldn't get to work when I submitted this. However when the app was re-opened it always pulled in the latest favorites which were updated from the phone. 

>Would it make sense to send a broadcast in PhoneSettingsListener, which is then picked up somewhere to update the list in the MainViewModel?

 I think a broadcast might be helpful here I'll have to look into that. 🤔, honestly didn't think about it since I felt the most important parts were already done 😂

wear/src/main/java/io/homeassistant/companion/android/phone/PhoneSettingsListener.kt
dshokouhi: Honestly I thought we needed both but I very well could be mistaken lol

leroyboerefijn:
> I could do that but then it might make sense to show a category saying these are Input Booleans or what not. I opted for entity ID so users know which entity it is. Honestly I figured we could redo the `Rows` to show the icons etc... in a more friendly way 😁
> 
> Whatever we do here though should probably be reusable so maybe making the Row a better compose element makes the most sense. We have a few places to add these to like QS tiles and Shortcuts for example. But neither of those were updated to compose yet.

That makes a lot of sense indeed. I suppose we can reuse that row composable as you say. Maybe we can add a main label with the friendly name (if it has one) and a secondary label with the entity ID? Then the next question is do you want the icon on the left and the checkbox on the right? Or the other way around? Or both on one side? :P


> So updating Wear when the app is open is something I couldn't get to work when I submitted this. However when the app was re-opened it always pulled in the latest favorites which were updated from the phone.

Interesting! Somehow that doesn't work for me. I really need to fully close the app with the back gesture, instead of just the home button...

> I think a broadcast might be helpful here I'll have to look into that. 🤔, honestly didn't think about it since I felt the most important parts were already done 😂

Oh that is definitely true! This is more a finishing touch on the user experience :)

wear/src/main/java/io/homeassistant/companion/android/phone/PhoneSettingsListener.kt
dshokouhi: yup you are right we are not using `sendMessage` with `/save_home_favorites`  so this should not be needed, I was just trying to very closely follow what was laid out before :)

dshokouhi:
> Interesting! Somehow that doesn't work for me. I really need to fully close the app with the back gesture, instead of just the home button...

ok that makes sense then because currently favorites only update in `onCreate` so I think a broadcast receiver to update the view model will be necessary then to trigger the update. I always went to the home screen by using the gesture during my testing :)

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearHomeView.kt
```diff
@@ -0,0 +1,42 @@
...
+        val navController = rememberNavController()
+        NavHost(navController = navController, startDestination = SettingsWearMainView.LANDING) {
+            composable(SettingsWearMainView.FAVORITES) {
+                LoadWearFavoritesSettings(
+                    settingsWearViewModel.entities,
+                    settingsWearViewModel.favoriteEntityIds.toList(),
+                    { b: Boolean, s: String, a: Activity -> settingsWearViewModel.onEntitySelected(b, s, a) },
+                    {
+                        settingsWearViewModel.favoriteEntityIds.toList().contains(
+                            settingsWearViewModel.favoriteEntityIds.toList()[it]
```
leroyboerefijn, line 29: I noticed a bug where trying to deselect an entity in some cases results in duplicating the entity (both still selected). I think it is due to lines 28 and 29. This should return whether the entity is selected.

leroyboerefijn:
> Yea lots of considerations here, we could also try to make our own chip to help match the wear UI 😂

That would be a fun one haha 😁 But would indeed be nice to have a common design element

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearHomeView.kt
dshokouhi: So when an entity is selected it actually moves to the favorite list, then once deselected from the favorite it goes back down to the entity list. Does it actually pass that to the device?

wear/src/main/java/io/homeassistant/companion/android/phone/PhoneSettingsListener.kt
leroyboerefijn: Yeah with the onboarding I was struggling a bit as well. In the end I think I might have overdone it a bit there as well. I wanted to get anything already in the DataClient, but then also send a message to the phone to send it again. The idea was that it would show what is available ASAP, but still update with the latest info. I could have a look to see if the onboarding one can be simplified as well

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearHomeView.kt
leroyboerefijn: So I'm not sure if I understand your question. But in the log I see that the flow for sending the changes to the wear device is followed correctly.

In any case, I'm also not sure if I'm pointing to the correct source of the bug. But I don't understand how these two lines are supposed to work. Loosely translated it states: `list.contains(list[index])`. I'm not sure how that resolves to whether the  entity is selected as a favorite.

Let me also better describe the bug. I'm trying to deselect one of the favorite entities, which are displayed on the top with selected items. If I press it, instead of deselecting it duplicates the item. So the same entity is shown twice at the top of the list, both selected. If I press again, it duplicates again.
When I leave the screen and come back again, the list is normal again.
This does not always happen. Sometimes I can deselect an entity without problems.

dshokouhi: ok let me try and play with it while I look into getting a receiver setup to update the favorite view model.  The purpose of the call is to determine if the favorite item should be checked or not, its similar to what we do set favorites. I think because it doesn't always happen might suggest a timing issue where the list is updated slower than when the user selects the entity maybe?

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearFavoritesView.kt
```diff
@@ -0,0 +1,140 @@
...
+                    fontWeight = FontWeight.Bold
+                )
+            }
+            items(favoritesList.size) { index ->
+                Row(
+                    modifier = Modifier
+                        .padding(15.dp)
+                        .clickable {
+                            onEntitySelected(
+                                isFavoriteSelected(index),
```
leroyboerefijn, line 81: So to add to the discussion in https://github.com/home-assistant/android/pull/1908/files/5f58c8f00b5e07da265da5239ac8fdeb0abd60bf#r750641450

This line is where that function is called, as well as in line 88.

You only reach this line for items in `favoritesList`, since that's what you loop over in line 75. So this `Row` belongs to an entity in the `favoritesList`, which in turn is `settingsWearViewModel.favoriteEntityIds.toList()`. Now when a user clicks the `Row`, a check is done on whether the clicked item is a favorite. This check takes the item from `settingsWearViewModel.favoriteEntityIds.toList()` and checks if it is part of `settingsWearViewModel.favoriteEntityIds.toList()`. 

So what is the goal of `isFavoriteSelected(index)`? Is it to verify if the item is still in the list? Shouldn't it always return true for any Row created from the `favoritesList`? At least I'm not really following what is going on, or why this check is happening haha

dshokouhi: ok i see the issue now where it duplicates....its due to the `Row` itself and actually it should always return `false` because its already selected as its part of favorites. I was able to reproduce and correct this. The `isFavoriteSelected` is actually important to keep track of the checked state itself but it was causing a weird loop in that one area. Good catch!!

dshokouhi:
Ok fixed the issue regarding favorites duplicating. I also added some code to update the favorite entities upon resuming the app. It was a bit difficult to try to get a receiver setup and its actually not correct. What we should do is save the entries to the DB and use a `Flow` of those values so its always updating just like we do today for entity updates.

wear/src/main/java/io/homeassistant/companion/android/phone/PhoneSettingsListener.kt
dshokouhi: yup this line was not needed, but we still need the reference in the manifest

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    fun onEntitySelected(checked: Boolean, entityId: String, activity: Activity) {
+        if (checked)
+            favoriteEntityIds.add(entityId)
+        else
+            favoriteEntityIds.remove(entityId)
+        sendHomeFavorites(favoriteEntityIds.toList(), activity)
+    }
+
+    private fun sendHomeFavorites(favoritesList: List<String>, activity: Activity) = viewModelScope.launch {
```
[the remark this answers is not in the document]
JBassett: I think we need to change this to an `AndroidViewModel` rather than passing the activity:
> If the ViewModel needs the Application context, for example to find a system service, it can extend the AndroidViewModel class and have a constructor that receives the Application in the constructor, since Application class extends Context.

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
```diff
@@ -0,0 +1,117 @@
...
+
+        val putDataRequest = PutDataMapRequest.create("/save_home_favorites").run {
+            dataMap.putString("favorites", favoritesList.toString())
+            setUrgent()
+            asPutDataRequest()
+        }
+
+        Wearable.getDataClient(activity).putDataItem(putDataRequest).apply {
+            addOnSuccessListener { Log.d(TAG, "Successfully sent favorites to wear") }
+            addOnFailureListener { Log.d(TAG, "Failed to send favorites to wear") }
```
JBassett, line 72: Should probably add a toast?

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
```diff
@@ -0,0 +1,117 @@
...
+            Log.d(TAG, "Found existing favorites: ${this.count}")
+            this.forEach {
+                val data = getFavorites(DataMapItem.fromDataItem(this.first()).dataMap)
+                Log.d(TAG, "Favorites: $data")
+                favoriteEntityIds.clear()
+                favoriteEntityIds.addAll(
+                    data.removeSurrounding("[", "]").split(", ").map { it }
+                )
+            }
+        }
```
JBassett, line 88: Using await is bad, it's going to cause IO on the main thread.  We should use the view model scope and launch rather than await.

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    fun getFavorites(map: DataMap): String {
+        map.apply {
+            return getString("favorites", "")
+        }
```
JBassett, line 94: This feels strange as is.
```suggestion
        return map.getString("favorites", "")
```

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
```diff
@@ -0,0 +1,117 @@
...
+            Log.d(TAG, "Requesting favorite data")
+            Wearable.getMessageClient(activity).sendMessage(
+                node.id,
+                "/send_home_favorites",
+                ByteArray(0)
+            ).apply {
+                addOnSuccessListener { Log.d(TAG, "Request to favorites sent successfully") }
+                addOnFailureListener { Log.d(TAG, "Failed to get favorites") }
+            }
+        }
```
JBassett, line 115: Await bad, use view model scope and launch a coroutine.

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearFavoritesView.kt
```diff
@@ -0,0 +1,140 @@
...
+    "input_boolean", "light", "switch", "script", "scene"
+)
+
+@Composable
+fun LoadWearFavoritesSettings(
+    entities: Map<String, Entity<*>>,
+    favoritesList: List<String>,
+    onEntitySelected: (Boolean, String, Activity) -> Unit,
+    isFavoriteSelected: (Int) -> Boolean,
+    activity: Activity
```
JBassett, line 42: Shouldn't pass this.

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearHomeView.kt
```diff
@@ -0,0 +1,42 @@
...
+    MdcTheme {
+        val navController = rememberNavController()
+        NavHost(navController = navController, startDestination = SettingsWearMainView.LANDING) {
+            composable(SettingsWearMainView.FAVORITES) {
+                LoadWearFavoritesSettings(
+                    settingsWearViewModel.entities,
+                    settingsWearViewModel.favoriteEntityIds.toList(),
+                    { b: Boolean, s: String, a: Activity -> settingsWearViewModel.onEntitySelected(b, s, a) },
+                    {
+                        settingsWearViewModel.favoriteEntityIds.toList().contains(
```
JBassett, line 28: Isn't this already a list?

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearFavoritesView.kt
```diff
@@ -0,0 +1,140 @@
...
+val supportedDomains = listOf(
+    "input_boolean", "light", "switch", "script", "scene"
+)
+
+@Composable
+fun LoadWearFavoritesSettings(
+    entities: Map<String, Entity<*>>,
+    favoritesList: List<String>,
+    onEntitySelected: (Boolean, String, Activity) -> Unit,
+    isFavoriteSelected: (Int) -> Boolean,
```
JBassett, line 41: Why do we need to pass this function?  Can't we determine inline?

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearLandingView.kt
```diff
@@ -0,0 +1,79 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import io.homeassistant.companion.android.R
+import io.homeassistant.companion.android.util.wearDeviceName
+
+@Composable
+fun SettingWearLandingView(
+    context: Context,
```
JBassett, line 29: Don't pass context

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    private val TAG = "SettingsWearViewModel"
+    private val CAPABILITY_WEAR_FAVORITES = "send_home_favorites"
+
+    var entities = mutableStateMapOf<String, Entity<*>>()
+        private set
+    var favoriteEntityIds = mutableStateListOf<String>()
+
+    fun init(integrationUseCase: IntegrationRepository) {
+        this.integrationUseCase = integrationUseCase
```
JBassett, line 33: I think you should be able to inject this?

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearMainView.kt
```diff
@@ -0,0 +1,77 @@
...
+        Thread { settingsWearViewModel.findExistingFavorites(this) }.start()
+        Thread { settingsWearViewModel.requestFavorites(this) }.start()
+    }
+
+    override fun onPause() {
+        super.onPause()
+        Wearable.getDataClient(this).removeListener(this)
+    }
+
+    override fun onDataChanged(dataEvents: DataEventBuffer) {
```
JBassett, line 63: This belongs in the View Model I think.

JBassett (changes requested):

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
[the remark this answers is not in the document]
dshokouhi: done

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
dshokouhi: done

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
dshokouhi: thank you :)

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
dshokouhi: ok switched over to view model scope

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearFavoritesView.kt
dshokouhi: I cant recall, I have switched to inline :)

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
dshokouhi: So I wasnt able to I thought we had to do something similar to the comment in the Wear view model so I mimiced what we did for now.

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearMainView.kt
dshokouhi: I moved over the logic but we still need `onPause` and `onResume` from the activity to properly set the listener

dshokouhi:
Hope I addressed them all appropriately :)

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
JBassett: https://developer.android.com/training/dependency-injection/hilt-jetpack#viewmodels

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
```diff
@@ -0,0 +1,152 @@
...
+
+class SettingsWearViewModel(application: Application) :
+    AndroidViewModel(application),
+    DataClient.OnDataChangedListener {
+
+    val app = application
+    private lateinit var integrationUseCase: IntegrationRepository
+
+    private val TAG = "SettingsWearViewModel"
+    private val CAPABILITY_WEAR_FAVORITES = "send_home_favorites"
```
JBassett, line 33: Should be in the companion object.

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
```diff
@@ -0,0 +1,152 @@
...
+import io.homeassistant.companion.android.R
+import io.homeassistant.companion.android.common.data.integration.Entity
+import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
+import kotlinx.coroutines.launch
+
+class SettingsWearViewModel(application: Application) :
+    AndroidViewModel(application),
+    DataClient.OnDataChangedListener {
+
+    val app = application
```
JBassett, line 29: Shouldn't need to do this, AndroidViewModel has a getter for getting the application.

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearHomeView.kt
```diff
@@ -0,0 +1,36 @@
...
+            composable(SettingsWearMainView.FAVORITES) {
+                LoadWearFavoritesSettings(
+                    settingsWearViewModel.entities,
+                    settingsWearViewModel.favoriteEntityIds.toList()
+                ) { b: Boolean, s: String ->
+                    settingsWearViewModel.onEntitySelected(
+                        b,
+                        s
+                    )
+                }
```
JBassett, line 27: Let's just pass the whole view model down, since we use this many items off it.

JBassett (changes requested):

app/src/full/java/io/homeassistant/companion/android/settings/SettingsWearViewModel.kt
dshokouhi: So I tried to use `getApplication` but the call failed. I can try it again. Maybe I need to assign it first before using in the functions?

dshokouhi: Ok this was my bad, I had to set a top level variable in order to properly attain it

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearMainView.kt
dshokouhi: moved all logic to the view model and just passing in the the proper calls in `onResume` and `onPause` now

app/src/full/java/io/homeassistant/companion/android/settings/views/SettingsWearMainView.kt
```diff
@@ -0,0 +1,55 @@
...
+            LoadSettingsHomeView(settingsWearViewModel, currentNodes.first().displayName)
+        }
+        settingsWearViewModel.init()
+    }
+
+    override fun onResume() {
+        super.onResume()
+        settingsWearViewModel.startWearListening()
+        Thread { settingsWearViewModel.findExistingFavorites() }.start()
+        Thread { settingsWearViewModel.requestFavorites() }.start()
```
JBassett, line 48: Since these launch coroutines I don't think you need the thread stuff anymore.

dshokouhi: yup, not needed thank you!

JBassett (approved):

PR #1909, opened 2021-11-12
Change splash screen icon

SkechyWolf:
## Summary

Changes the splash screen image to ic_launcher_foreground instead of having a whole new transparent HA logo. Changing it does lose a little quality but it's not really noticeable.

## Screenshots

image


SkechyWolf:
One thing regarding ic_launcher_foreground is that the left side of the roofs pixel don't step down the same as the right side

LasseRosenow:
What about using a vector drawable? It can be easily created from the svg from the homeassistant.io homepage.

JBassett (approved):

Issue #1916, opened 2021-11-14
Wear OS app crashes if you are not signed in as the WebSocket cannot find a URL


SkechyWolf:
**Home Assistant Android version:**  home-assistant-wear-beta-1666-1e5534d (Newest Release as of 11/13/2021)

**Android version:** Wear OS 3

**Phone model:** Galaxy Watch 4

**Home Assistant version:** 2021.11.3

**Last working Home Assistant release (if known):** N/A

**Description of problem:** If you are not already signed in the app crashes as the WebSocket cannot find a URL

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```
2021-11-13 16:41:21.298 26212-26212/io.homeassistant.companion.android.debug E/AndroidRuntime: FATAL EXCEPTION: main
    Process: io.homeassistant.companion.android.debug, PID: 26212
    java.lang.Exception: Unable to get URL for WebSocket
        at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.connect(WebSocketRepositoryImpl.kt:164)
        at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.sendMessage(WebSocketRepositoryImpl.kt:189)
        at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.getStateChanges(WebSocketRepositoryImpl.kt:127)
        at io.homeassistant.companion.android.common.data.integration.impl.IntegrationRepositoryImpl.getEntityUpdates(IntegrationRepositoryImpl.kt:468)
        at io.homeassistant.companion.android.home.HomePresenterImpl.getEntityUpdates(HomePresenterImpl.kt:61)
        at io.homeassistant.companion.android.home.MainViewModel$loadEntities$1.invokeSuspend(MainViewModel.kt:43)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.internal.DispatchedContinuationKt.resumeCancellableWith(DispatchedContinuation.kt:367)
        at kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable(Cancellable.kt:30)
        at kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable$default(Cancellable.kt:25)
[...]

PR #1917, opened 2021-11-14
Catch exception when we fail to get entity updates

dshokouhi:
## Summary

Fixes: #1916 by catching the exception the same as we do for `getEntities`


JBassett:
While this will stop the crash it won't ever try to reconnect unless the app re-registers.

dshokouhi:
Closing as this is not the correct approach

PR #1933, opened 2021-11-16
Hilt Upgrade

JBassett:
## Summary

Upgrading from raw dagger to using Hilt.

## Screenshots

No visual changes.

N/A


dshokouhi (approved):
I did not see any issues come up during my time testing this. With the phone app I looked at all settings screens, the widget configuration activity, registering sensors and their updates. I had also created some widgets and saw the data pulled in as expected like entity state widget. I was unable to test notifications here as the debug build doesn't support it from actions. With wear I was able to login and confirm the issue was fixed and the entities load up fine and they execute as expected. At this point everything seems to be working.

PR #1940, opened 2021-11-18
Allow the websocket to reconnect

JBassett:
## Summary

Currently if we loose our connection with the websocket we also loose our state changed registration.  This should force the websocket to reconnect.

## Screenshots

N/A

N/A


dshokouhi (approved):
Looks good to me, tested briefly with wear and phone app

dshokouhi:
This PR also corrects an issue where the wear OS app is crashing when the watch has no data connection and the app is launched

PR #1941, opened 2021-11-18
Remove startup activity animation

LasseRosenow:
## Summary
Currently when starting the home assistant app, first the launch screen is shown and then a transition to the webview activity is triggered.
The issue is, that this transition is animated, which should not be the case as a launch screen is supposed to just appear and reveal the app behind itself.

This PR disables the transition animation on startup.

## Screenshots

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>

https://user-images.githubusercontent.com/10547444/142456015-5363b5ba-b37d-4e0a-b346-5634809a07f2.mp4

</td>
<td>

https://user-images.githubusercontent.com/10547444/142456724-4c1c53f2-0736-4fc0-b383-a1a8cb49062c.mp4

</td>
</tr>
</table>

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt
@@ -23,6 +23,7 @@ class LaunchActivity : BaseActivity(), LaunchView {
 
         startActivity(WebViewActivity.newInstance(this, intent.data?.path))
         finish()
+        overridePendingTransition(0, 0) // Disable activity start/stop animation
     }
 
     override fun displayOnBoarding(sessionConnected: Boolean) {
@@ -31,6 +32,7 @@ class LaunchActivity : BaseActivity(), LaunchView {
 
         startActivity(intent)
         finish()
+        overridePendingTransition(0, 0) // Disable activity start/stop animation
     }
 
     override fun onDestroy() {


dshokouhi (approved):
Awesome thank you! Feel free to join us in #devs_mobile_apps on discord if you ever wanted to discuss anything. We do have some UI/UX discussions every now and then. I think your expertise can be helpful :)

leroyboerefijn:
Would it be possible to animate the launch icon into the icon on the welcome screen?

LasseRosenow:
That would be difficult, as for example on android 12 the whole launch screen looks different for example because android has taken control of its styling.

So we could only do it for < 12 which is not worth the effort in my opinion

dshokouhi:
For android 12 there is a new Splash Screen API that we might be able to use: https://developer.android.com/guide/topics/ui/splash-screen

Right now the app is using whatever the default is

LasseRosenow:
This new API is pretty limited though. I'm not sure if that would allow us to make the necessary changes that would be needed to achieve what @leroyboerefijn wants. 

Maybe I can take a look on that in the future, but I think its a bit overkill for this PR, which was only meant as a "bug fix" for now.

Also the animation between the welcome screen and the launch screen is not so important for now, as its really only seen maybe once or twice per user, until they are logged in. Also before thinking about changing the animation there I would propose to start with redesigning the welcome screen itself first, as it looks quite unpolished itself

dshokouhi:
> Maybe I can take a look on that in the future, but I think its a bit overkill for this PR, which was really only meant as a "bug fix" to remove those weird animations for now.

I agree, out of scope for this PR



> Also before thinking about changing the animation there I would propose to start with redesigning the welcome screen itself first, as it looks quite unpolished itself (not meant to be offensive, I know there are much more important things than the welcome screen to look good :) )

Admittedly I am not the best person for design so no offense taken 😃 . The text you see on screen was approved and also used in the iOS app so please feel free to update the design itself.

I hope that with Compose things like that are easier because I certainly find UI building easier.  One thing we have started is rebuilding the UI in Compose. The only other screen currently "redone" on the phone side was the Notification Detail Page. All my current and future PRs will also be in Compose include #1908

JBassett (approved):

PR #1943, opened 2021-11-19
WIP Make initial webview background use launch screen background color

LasseRosenow:
## Summary
This builds on top of home-assistant/android/pull/1938 which introduced a proper launch screen using the `#111111` as background color in dark mode, which is the same as the home assistent HTML frontend uses for its launch screen.

Now the issue is, that the webView initially draws a black background, which is later painted over by the css of the home assistant frontend. This results in a black color in between the launch screen and the frontend which both use `#111111`.

With this PR I set the default webView background color to the background color of the launch screen, so that there is no black color in between anymore.

## Screenshots
<table>
<tr>
<th>Before</th><th>After</th>
</tr>
<tr>
<td>

https://user-images.githubusercontent.com/10547444/142516041-f2e674ec-d1dd-436a-b91f-bf5196c3bd53.mp4

</td>
<td>

https://user-images.githubusercontent.com/10547444/142516034-6669b4d8-1d6d-4ef3-9334-6eed12517569.mp4

</td>
</tr>
</table>


LasseRosenow:
As said in the first post, this is based on home-assistant/android/pull/1938 so it can only be merged for itself when home-assistant/android/pull/1938 is merged. So I mark it as WIP until home-assistant/android/pull/1938 is merged. :)

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -529,6 +530,10 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
             )
         }
 
+        // Set webView background color to transparent to override default black and default white backgrounds
+        // and let css draw its own background when its ready
+        webView.setBackgroundColor(ResourcesCompat.getColor(resources, R.color.colorLaunchScreenBackground, theme))
```
LasseRosenow, line 535: This is the only new change in this pr, that is not part of home-assistant/android/pull/1938

LasseRosenow:
I figured that having the black background replaced by a background that matches the splash screen is better, but also still not perfect, as it would result in

Android splash screen => webview with same background, but without the icon => webapps splash screen with icon again..

So it would be a better solution if we draw the splash screen behind the webView as well and make the webView Itself transparent. This would skip the background color without an icon part and provide a butter smooth transition in between the android and the web app's splash screen.

Of course in android 12 there would still be the part without the splash screen in between, as we cannot draw it behind the webView as we don't exactly know how it looks like, maybe I can find some solution for that when looking a bit deeper into the android 12 apis. But nonetheless it would still be better than having a black color that is even more out of place in between so I think that would be a good start.

I close this PR for know and will open a new one with my newer approach later ;)

PR #1944, opened 2021-11-19
Instant Update Widgets

JBassett:
## Summary

This PR gets widgets to update near instantly by using the websocket!  Clicking the template will update the value as well.

N/A


dshokouhi:
This is amazing, just tested it out. Only thing I noticed is that after I created the first widget I had to turn the screen off and on in order for the flow to start. I did a simple test with a binary sensor immediately after creating the first widget without letting the screen turn off lol. Then with the screen remaining on any widget I create thereafter also receives the update. Love that my template widget will update this fast!

JBassett:
> This is amazing, just tested it out. Only thing I noticed is that after I created the first widget I had to turn the screen off and on in order for the flow to start. I did a simple test with a binary sensor immediately after creating the first widget without letting the screen turn off lol. Then with the screen remaining on any widget I create thereafter also receives the update. Love that my template widget will update this fast!

Latest commit should fix that use case.

dshokouhi (approved):
Looks good to me! Love how fast it is!

Issue #1935, opened 2021-11-16
Add lock domain to wear os app please


blair287:
**Is your feature request related to a problem? Please describe.**

No

**Describe the solution you'd like**

Add lock domain to wear os app just like lights and switches.

**Describe alternatives you've considered, if any**

**Additional context**

dshokouhi:
@leroyboerefijn in order to add the `lock` domain to the tiles we need to read the state to send the proper service call. I don't think tiles currently reads the state right?

leroyboerefijn:
At the moment we don't indeed. But we can check the state in the broadcast receiver if it is a lock type

blair287:
I don't know how to write apps but willing to learn, anything I can do to help let me know please

leroyboerefijn:
@blair287 I'm going to implement the support for lock. Unfortunately I don't have any locks myself, so could you help me?

What is the entity state when the lock is locked and unlocked?

And after implementation, it would be great if you could help me test!

leroyboerefijn:
@blair287 please give PR #1959 a try!

blair287:
> @blair287 please give PR #1959 a try!

Anyway to build the app? I see can install the action apk but I don't want 2 of the app on phone or watch.

blair287:
> @blair287 I'm going to implement the support for lock. Unfortunately I don't have any locks myself, so could you help me?
> 
> What is the entity state when the lock is locked and unlocked?
> 
> And after implementation, it would be great if you could help me test!

The states are locked and unlocked.

dshokouhi:
> > @blair287 please give PR #1959 a try!
> 
> Anyway to build the app? I see can install the action apk but I don't want 2 of the app on phone or watch.

the debug apk installs side by side the production app and its a good way for us to tell when a feature is working. You can install the app and uninstall after the test. It won't impact anything and it has a red icon so easy to distinguish.

dshokouhi:
Closed via #1959

blair287:
> Closed via #1959

Could it be reopened the lock and unlock works but the switch in the app doesn't show any feedback to if its locked or unlocked.

Or shall I open a new issue?

leroyboerefijn:
Ah you're right, forgot to change this line:
https://github.com/home-assistant/android/blob/9fd3164583ad99983510404ae672207ba893db3c/wear/src/main/java/io/homeassistant/companion/android/home/views/EntityUi.kt#L45

PR #1959, opened 2021-11-22
Wear os add lock domain

leroyboerefijn:
## Summary
Adding support for the lock domain. This solves #1935.

## Any other notes
Note, I don't have any lock devices. I tested that the implementation doesn't break existing functionality, but I couldn't test the working of a lock.


wear/src/main/java/io/homeassistant/companion/android/home/HomePresenterImpl.kt
```diff
@@ -27,7 +27,7 @@ class HomePresenterImpl @Inject constructor(
             "media_player", "remote", "siren", "switch"
         )
         val supportedDomains = listOf(
-            "input_boolean", "light", "switch", "script", "scene"
+            "input_boolean", "light", "switch", "script", "scene", "lock"
```
dshokouhi, line 30: I wonder if we should add this to the `toggleDomain` list so it shows up as  a `ToggleChip` like it does in lovelace?

wear/src/main/java/io/homeassistant/companion/android/home/views/MainView.kt
```diff
@@ -128,7 +130,7 @@ fun MainView(
                                     )
                                 },
                                 onClick = {
-                                    onEntityClicked(favoriteEntityID)
+                                    onEntityClicked(favoriteEntityID, "unknown")
```
dshokouhi, line 133: Hmm so how is this going to work when a user adds a `lock` as a favorite when we don't have the proper state?

wear/src/main/java/io/homeassistant/companion/android/home/HomePresenterImpl.kt
leroyboerefijn: You're right, will add it and change the order in the when statement

wear/src/main/java/io/homeassistant/companion/android/home/views/MainView.kt
leroyboerefijn: Very good question. So if we don't have a state, there are three options:

1. Do nothing
2. Always lock
3. Always unlock

I think option 3 should be ruled out, since you don't want to accidentally unlock your front door. For the other two it is debatable. Right now, the default action is to lock (so also if the state is "unknown"). If you want we can update it to do nothing if the state does not have one of the two predefined values.

dshokouhi: Hmm this is a tough one, I think if we default to locking the lock that would be acceptable as we don't want to cause an accidental unlock :) I am also ok with doing nothing too.

leroyboerefijn: Hmm yeah either is fine for me :)

Is there any behavioral guidelines from home assistant? Or do you want to make the call?

dshokouhi: Hmm don't see anything mentioned in the docs: https://developers.home-assistant.io/docs/core/entity/lock/

We can wait for code review and see what @JBassett thinks we should do :)

JBassett: Locking is fine for now.  Once we cache the state in the DB we can use that  to determine state and what the next action should be.

JBassett (approved):

JBassett (approved):

PR #1989, opened 2021-12-03
Add support for urlencoded intent extras

### LINKED ISSUE #1949: Intent Extras cannot contain `:` and `,`
2021.10.0-full

**Android version:**
11 (LineageOS 18.1 unofficial)

**Phone model:**
Samsung Galaxy S9

**Home Assistant version:**
2021.10.6

N/A

**Description of problem:**
When using `command_broadcast_intent` with extras in the `group` parameter, there is no support for extras containing `:`, `,` or both.
This becomes a problem when receiving applications expect e.g. a json string in the extras. One example of this is Gadgetbridge, where some features in the latest development version require such extras.

```
N/A
```

**Screenshot of problem:**
N/A

**Additional information:**
The lack of support for `:` in extras can easily be fixed by changing https://github.com/home-assistant/android/blob/40e7907dbd742809c043479a081ed8e19dd9efa8/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt#L517 to something like `val pair = item.split(":", ignoreCase = false, limit = 2)`.
The case of the `,` is another matter entirely and will probably require a more thorough rework of the extras feature.

mvn23:
## Summary

When specifying intent extras for notification command `command_broadcast_intent`, the format used to specify these extras prohibits the use of certain special characters (`,` and `:`). With this PR, we allow extras to be urlencoded to overcome this limitation.
To use this functionality, the extras format is extended by allowing the user to append `:urlencoded` to any extra. The format then becomes a comma-separated list of the following:
`<INTENT_NAME>:<INTENT_VALUE>[:urlencoded]`
If the `:urlencoded` suffix is present, `<INTENT_VALUE>` will be urldecoded before adding it to the intent.
Since the old processing code did not handle `:` in `<INTENT_VALUE>` correctly, existing functionality is not affected.

## Screenshots

N/A

Documentation: home-assistant/companion.home-assistant#626

## Any other notes

fixes #1949

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt
@@ -67,6 +67,7 @@ import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
 import org.json.JSONObject
 import java.net.URL
+import java.net.URLDecoder
 import java.util.Locale
 import javax.inject.Inject
 import kotlin.collections.HashMap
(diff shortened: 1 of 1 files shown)


dshokouhi:
We should also update `command_activity` as that also takes Intent Extras

https://github.com/home-assistant/android/blob/master/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt#L1419-L1434

mvn23:
Thank you for the feedback, `command_activity` is now supported as well.
I was not aware that this should be fixed in more than 1 place, so I am marking this PR as a WIP for now. I am planning to break the extras value processing out into a separate function to avoid having to maintain the same code in multiple places. I will have time to do this over the next few days.

mvn23:
Found some time this night during my break, Extras processing now takes place in a separate function.
This PR is ready for review.

dshokouhi:
Lets make sure to get a docs update here so we can provide some clear examples to our users :)

mvn23:
Yes, that's the next item on my to-do list. Should be ready by Sunday.

mvn23:
Docs PR is there, but it may still need some adjustment.

JBassett (approved):

PR #1999, opened 2021-12-09
New Media Player Widget Design

jannis3005:
## Summary

Updated the Media Player Widget Design. It's now showing Artist, Title and album name if available. it also now updates more frequently when something is currently playing.

## Screenshots

Screenshot_1639070705
Screenshot_1639070738

Looking at the Docs for the Media Player Widget I don't think this needs a documentation update, as it essentially works the same as it did previously regarding user interaction.


homeassistant:
Hi @jannis3005, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;android#1999).

Thanks!

app/src/main/java/io/homeassistant/companion/android/widgets/media_player_controls/MediaPlayerControlsWidget.kt
```diff
@@ -278,10 +338,38 @@ class MediaPlayerControlsWidget : AppWidgetProvider() {
...
+            mediaFinishedIn =
+                (mediaPlayerEntity.attributes["media_duration"] as Double).minus(mediaPlayerEntity.attributes["media_position"] as Double)
+                    .toInt()
+        }
+        if (mediaFinishedIn < 30) {
+            nextUpdate = mediaFinishedIn * 1000
+        } else {
+            nextUpdate = 30000
+        }
+        am.set(AlarmManager.RTC_WAKEUP, System.currentTimeMillis().plus(nextUpdate), pendingUpdateMediaIntent)
```
dshokouhi, line 365: Instead of using alarm manager to schedule an update (which will interfere with the next alarm sensor), we should update the widget with the websocket updates as we subscribe to entity updates and can update as soon as the media player changes its state. Take a look at the entity widget and template widget to see how those update as soon as a state change is detected. This should help simplify the code here a bit and its a new feature that the app just gained that we should continue to use because we are already subscribed to these update :)  It should also remove the need for all the additional `getEntity` calls too

jannis3005: Okay, ill take a look

dshokouhi:
Thanks for your PR 🙏 I dig the design updates! I do agree docs updates are not needed here :)

jannis3005:
I also thought about adding a little icon in the top right indicating the source (possibly the icon of the media player entity) but I did not yet figure out how to convert the string (mdi:whatever) into something I can put into an image view.

dshokouhi:
> I also thought about adding a little icon in the top right indicating the source (possibly the icon of the media player entity) but I did not yet figure out how to convert the string (mdi:whatever) into something I can put into an image view.

I did this recently for the notification icon actually :) Hopefully this will help

https://github.com/home-assistant/android/blob/master/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt#L776-L778

Also if you need any help or have questions you can always feel free to join us in discord `#devs_mobile_apps` :)

jannis3005:
> > I also thought about adding a little icon in the top right indicating the source (possibly the icon of the media player entity) but I did not yet figure out how to convert the string (mdi:whatever) into something I can put into an image view.
> 
> I did this recently for the notification icon actually :) Hopefully this will help
> 
> https://github.com/home-assistant/android/blob/master/app/src/full/java/io/homeassistant/companion/android/notifications/MessagingService.kt#L776-L778
> 
> Also if you need any help or have questions you can always feel free to join us in discord `#devs_mobile_apps` :)

image
image


What do you think @dshokouhi? Still need to sort the state updating though. Haven't figured it out yet.

jannis3005:
I've cleaned up the code. hopefully, it is in a good state now. I think the design is also final and it is working much better than previously.

dshokouhi:
> image image
> 
> What do you think @dshokouhi?

I think this design looks awesome, good job!

dshokouhi (approved):
Tested the debug APK and it all seems to work. The only consideration we have here is that I noticed there are a lot of image updates going on so it does have the potential to rack up data with instant updates.  I wonder if the image should only be updated when a state has changed to reduce the amount of updates there?  We can wait for another review before we make a decision on the amount of image updates. The updated design is nice, do we think we should update the preview image to match the updated design?

JBassett (approved):
I think Picasso has some caching by default so loading the same URL should result in a pull from cache not a downloading every time but we should double check.

PR #2000, opened 2021-12-10
Migrate to theme-update event

### LINKED ISSUE #1997: Status- and Navigation bars are not being colored anymore on Core 2021.12.0b3
**Home Assistant Android version:** 2021.10.0-full

**Android version:** 11

**Phone model:** Xiaomi Mi 9T Pro

**Home Assistant version:** 2021.12.0b5

**Last working Home Assistant release (if known):** 2021.11.5

**Description of problem:**
I just updated to Home Assistant 2021.12.0b5 and realised, that the status and navigation bars are not getting colored anymore. This is especially visible in light mode, where they just stay black instead of getting colored blue.

I did some debugging and figured, that the android app is listening on a json object post with the type `config/get` which used to be send after the web app loaded. But it is not being send anymore. But when you open the Frontend Settings page, it is loaded, and the status and navigation bar are getting their colors.
But of course this is not how it is supposed to be. 

I also opened an Issue on the frontend repo, though I'm not really sure, where this post request comes from etc., so I'm not even sure if thats the right place and also nobody is answering issues on the frontend recently it seems like.
See: home-assistant/frontend/issues/10823

I figured that the android app team would be interested in this issue and might help reaching out to the people who are able to fix it :)

LasseRosenow:
## Summary
This fixes #1997. It uses the new theme-update external bus message (introduced in home-assistant/frontend/pull/10819), to detect theme changes within the frontend and update the status and navigation bars accordingly.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -438,6 +438,16 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
 
             addJavascriptInterface(
                 object : Any() {
+                    // TODO This feature is deprecated and should be removed after 2022.6
+                    @JavascriptInterface
+                    fun onHomeAssistantSetTheme() {
+                        // We need to launch the getAndSetStatusBarNavigationBarColors in another thread, because otherwise the evaluateJavascript inside the method
+                        // will not trigger it's callback method :/
+                        GlobalScope.launch(Dispatchers.Main) {
+                            getAndSetStatusBarNavigationBarColors()
+                        }
+                    }
+
                     @JavascriptInterface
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -501,17 +492,6 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                                     webView.evaluateJavascript(script) {
                                         Log.d(TAG, "Callback $it")
                                     }
-
-                                    getAndSetStatusBarNavigationBarColors()
```
JBassett, line 505: Since we support back ~ 6 months worth of core versions I say we leave this.  Just add a comment that after 2022.6 releases we can remove it.

JBassett (changes requested):

JBassett (approved):

PR #2026, opened 2021-12-19
Add a work profile sensor

zmarties:
The information on whether the work profile is available only seems to
be given by tracking the state of the last received broadcast intent
indication - MANAGED_PROFILE_[UN]AVAILABLE

Thus add a receiver for these intents, and slightly change the architecture
so that the intent can be passed down when the sensor needs to be updated.

I think there's a ineffciency in the architecture - currently all
sensors are asked to update whenever any broadcast the app tracks is
received, regardless of whether than broadcast is relevant to the the
sensor or now - for example all sensors update when battery state changes.

## Screenshots

image
image

First screenshot shows work profile turned off, and so sensor is false.
Second screenshot shows work profile turned on (see the key in the status bar), and hence sensor is true.


dshokouhi:
> The information on whether the work profile is available only seems to be given by tracking the state of the last received broadcast intent indication - MANAGED_PROFILE_[UN]AVAILABLE
> 
> Thus add a receiver for these intents, and slightly change the architecture so that the intent can be passed down when the sensor needs to be updated.

Please share some screenshots showing the sensor updates in action for this sensor. When I did testing on my own I did indeed use these intents but the application was unable to receive them and yes they were registered as inline receivers.



> I think there's a ineffciency in the architecture - currently all
> sensors are asked to update whenever any broadcast the app tracks is
> received, regardless of whether than broadcast is relevant to the the
> sensor or now - for example all sensors update when battery state changes.

This is as designed because not all sensors update immediately and benefit more and update faster by updaing with other sensors. We should not change this.

app/src/main/java/io/homeassistant/companion/android/sensors/DevicePolicyManager.kt
```diff
@@ -0,0 +1,70 @@
...
+        val isWorkProfile = SensorManager.BasicSensor(
+            "is_work_profile",
+            "binary_sensor",
+            R.string.sensor_name_work_profile,
+            R.string.sensor_description_work_profile
+        )
+    }
+
+    override fun docsLink(): String {
+        return "https://companion.home-assistant.io/docs/core/sensors#work-profile-sensor"
```
dshokouhi, line 21: We need to make sure to add back the docs PR so this link will work.

app/src/main/java/io/homeassistant/companion/android/sensors/PowerSensorManager.kt
```diff
@@ -99,7 +99,7 @@ class PowerSensorManager : SensorManager {
             dozeState,
             icon,
             mapOf(
-                "ignoring_battery_optizimations" to powerManager.isIgnoringBatteryOptimizations(packageName)
+                "ignoring_battery_optimizations" to powerManager.isIgnoringBatteryOptimizations(packageName)
```
dshokouhi, line 102: This is a breaking change and unrelated to this PR, PRs should only contain changes that are part of it.

app/src/main/java/io/homeassistant/companion/android/sensors/DevicePolicyManager.kt
```diff
@@ -0,0 +1,70 @@
...
+    override fun getAvailableSensors(context: Context): List<SensorManager.BasicSensor> {
+        return listOf(isWorkProfile)
+    }
+
+    override fun requiredPermissions(sensorId: String): Array<String> {
+        return emptyArray()
+    }
+
+    override fun requestSensorUpdate(context: Context) {
+        updateWorkProfile(context)
```
dshokouhi, line 40: I don't think the code will reach this anymore right?

dshokouhi (changes requested):
Lets also make sure the sensor is indeed working, we should also get the user who requested the sensor to perform a test once the actions have run.

app/src/main/java/io/homeassistant/companion/android/sensors/PowerSensorManager.kt
zmarties: I sort of regarded it as part of it - I had to take a crash course in sensor writing, and as such used the PowerManager as an example, so got to read the code for that in quite some depth, thus spotting this problem.

I also assumed this was just an internal name - I couldn't find any use of that string elsewhere in the app, so didn't think it broke anything.  Happy to drop it from the PR if that's how things are done here.

dshokouhi: This is an attribute of the sensor and changing it has the potential to break any template depending on it, we like to give a heads up on breaking changes to users and not just make a undocumented change. So a separate PR highlighting the breaking change can minimize its impact.

app/src/main/java/io/homeassistant/companion/android/sensors/DevicePolicyManager.kt
zmarties: In the normal flow of things that is true, but to minimize changes I did not change the interface, so I think this function still needs to be provided.

zmarties:
> 



> Lets also make sure the sensor is indeed working, we should also get the user who requested the sensor to perform a test once the actions have run.

I work for a company that has installed a work profile on my phone, so I have been able to test that it works for me.

app/src/main/java/io/homeassistant/companion/android/sensors/DevicePolicyManager.kt
dshokouhi: The function does need to be there but the update call here will always return `false` so its code doesnt have any operation to provide a valid state if its hit.  In cases like this you can just have a blank function with a comment.

app/src/main/java/io/homeassistant/companion/android/sensors/PowerSensorManager.kt
zmarties: @dshokouhi Looking for advice as to how this project expects me to do this - should I make a further commit that reverts this line, or do I rewrite history, and simply force push the existing commit without this line changed?

dshokouhi: @zmarties its ok to submit another commit reverting the change :)

dshokouhi:
> I work for a company that has installed a work profile on my phone, so I have been able to test that it works for me.


Awesome thank you for confirming that, this sensor really needed someone with the proper setup to get it working :) I was shooting in the dark with my other PR 😂

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorManager.kt
```diff
@@ -66,6 +67,15 @@ interface SensorManager {
...
+    // Request to update a sensor, including any broadcast intent which may have triggered the request
+    // The intent will be null if the update is being done on a timer, rather than as a result
+    // of a broadcast being received.
+    fun requestSensorUpdate(context: Context, intent: Intent?) {
+        // Few sensors care about the intent, so allow them to just implement the interface that
+        // does not get passed that parameter.
+        requestSensorUpdate(context)
+    }
+
     fun requestSensorUpdate(context: Context)
```
JBassett, line 79: Rather than create a whole new function, why not just give an optional parameter?
```suggestion
    fun requestSensorUpdate(context: Context, intent: Intent? = null)
```

JBassett (changes requested):

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorManager.kt
zmarties: This is the abstract definition of a funtion in an interface, so I don't think that adding an optional parameter acts as you are implying.  If I changed the abstract function, I would then have to add the new parameter to each of the existing concrete classes that implement the interface (about 25 files), and as a breaking change, this would also break every private fork of Home Assistant that has implemented its own Sensor classes.

By instead adding a new function to the interface, with a concrete body, no changes are needed to existing implementation classes, but new classes can chose to implement the new method if they need to.

JBassett: For now we can leave it, however I would argue we shouldn't worry about forks who have no interest in contributing upstream.  If we need a refactor, we will do it.

Issue #1372, opened 2021-02-23
Quick Access Device Controls - area in the subtitle


RouNNdeL:
**Is your feature request related to a problem? Please describe.**

I'd like to see the entity's area (if assigned) to be displayed in the Quick Access Device Controls from Android 11. 
Currently only the entity's `friendly_name` or it's `id` is shown in the title.

**Describe the solution you'd like**

I'd like to see the entity's area displayed under its name/id.

**Describe alternatives you've considered, if any**

I wanted to implement this myself and submit a PR, but I could not find any references in the code to the entity's area.

**Additional context**

The official documentation shows the use of a subtitle to display the room the device is in.

![image](https://user-images.githubusercontent.com/22538735/108836066-aeb9ed80-75d0-11eb-8aff-51b6536e30c2.png)

![image](https://user-images.githubusercontent.com/22538735/108837143-1fadd500-75d2-11eb-8852-89136304d854.png)

dshokouhi:
> I wanted to implement this myself and submit a PR, but I could not find any references in the code to the entity's area.

This is because areas are not provided over the REST API and the app does not support websockets yet.

RouNNdeL:
Oh, that's a bit disappointing. Are there some developer guidelines for this project I could follow to help out?

dshokouhi:
We just use Android docs and also HA dev docs: https://developers.home-assistant.io/

dshokouhi:
@RouNNdeL we have recently had websockets support added, if you wanted to take a crack at this you are more than welcome to :)  One thing to keep in mind is not all websocket calls are documented in HA dev docs. You'll have to inspect the HA core code base. For areas you can look at area registry and probably compare to the entity registry?

dshokouhi:
Fixed in #2027

PR #2027, opened 2021-12-19
Add area/device/entity registry in order to show area on device controls

jpelgrom:
## Summary

This PR implements issue/request #1372.
In order to show the area, we need to load the area, device and entity registries. Just the entity registry is not enough: an entity might not have an area set, in which case we need to use the area of the related device. There doesn't appear to be a way to avoid this, and as an example both the [frontend](https://github.com/home-assistant/frontend/blob/dev/src/panels/config/entities/ha-config-entities.ts#L367-L368) and [template helper](https://github.com/home-assistant/core/blob/dev/homeassistant/helpers/template.py#L1061-L1072) load both registries to get information.

## Screenshots

The UI is provided by the system, this is on a Google Pixel 4a running Android 12.
|Before|After (the fan does not have an area set)|
|-----|-----|
|ha-controls-before|ha-controls-after|

I don't think this is necessary, the current page about device controls doesn't include a specific description of what is shown on the tile and/or any screenshots.

## Any other notes

Because the registries are only accessible via the websocket API, some additional changes have been made to fix issues/prevent code duplication when using the websocket... Based on my testing, everything works as before but it'd be great if someone else can check this to be sure. Specifically:
 - In HaControlsProviderService: collecting flows with `takeWhile { condition }` doesn't immediately end the subscription once the condition becomes false, as the condition [appears to be only checked when a new value is emitted](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/flow/operators/Limit.kt#L85-L92). This was noticed because the events for registry updates aren't sent as frequently as state updates, and the websocket connection would remain open even after closing the controls.<br>To fix this, cancelling the CoroutineScope also immediately cancels any subscription. This created a race condition for me once in sending the unsubscribe messages, where the messages were sent out of order and one failed as a result of non-sequential IDs, but I haven't been able to reproduce this later.
 - Subscribing to events has been generalised based on the `state_changed` event subscription that was already present. This requires some casting which seems to work fine but a thumbs up from someone else would be good to have.


homeassistant:
Hi @jpelgrom, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;android#2027).

Thanks!

app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
```diff
@@ -35,6 +36,7 @@ class ClimateControl {
                 )
             )
             control.setTitle((entity.attributes["friendly_name"] ?: entity.entityId) as CharSequence)
+            control.setSubtitle(area?.name ?: "")
```
zmarties, line 39: You have added this identical line to every single control class. That very much suggests to me that it should be done in the superclass that all controls inherit from.

Actually, the existing code is giving a poor example to follow, since it also sets the title using the same code repeated in every child class, which again should be handled by the superclass.

app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
```diff
@@ -77,41 +89,97 @@ class HaControlsProviderService : ControlsProviderService() {
...
+                                        getAreaForEntity(it.entityId, areaRegistry, deviceRegistry, entityRegistry)
+                                    )
+                                    subscriber.onNext(control)
+                                }
+                            }
+                        }
+                        webSocketScope.launch {
+                            areaRegistryFlow.collect {
+                                areaRegistry = webSocketRepository.getAreaRegistry()
+                                entities.forEach {
```
zmarties, line 136: There are 3 identical copies of this forEach loop.  I'd make that a small private function, and simply pass the different areaRegistry in to it each time, to make the code easier to follow.

app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
```diff
@@ -140,4 +208,26 @@ class HaControlsProviderService : ControlsProviderService() {
...
+        areaRegistry: List<AreaRegistryResponse>,
+        deviceRegistry: List<DeviceRegistryResponse>,
+        entityRegistry: List<EntityRegistryResponse>
+    ): AreaRegistryResponse? {
+        val rEntity = entityRegistry.firstOrNull { it.entityId == entityId }
+        if (rEntity != null) {
+            // By default, an entity should be considered to be in the same area as the associated device (if any)
+            // This can be overridden for an individual entity, so check the entity registry first
+            if (rEntity.areaId != null) {
+                return areaRegistry.firstOrNull { it.areaId == rEntity.entityId }
```
zmarties, line 223: I'll admit to not following exactly what is going on here, but should this really be comparing area and entity ids?  At first glance those don't seem to be comparable.

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
```diff
@@ -128,21 +173,36 @@ class WebSocketRepositoryImpl @Inject constructor(
     }
 
     @ExperimentalCoroutinesApi
-    override suspend fun getStateChanges(): Flow<StateChangedEvent> {
-        stateChangedMutex.withLock {
-            if (stateChangedFlow == null) {
+    override suspend fun getStateChanges(): Flow<StateChangedEvent> = subscribeToEventsForType("state_changed")
```
zmarties, line 176: I'd have made all the type strings that you subscribe to into constants - that way there is no chance of a typo when you come to handle them in handleEvent().

app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
jpelgrom: That should indeed be comparing the area id, fixed!

app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
jpelgrom: Good idea, I've also included the loop for loading the initial state as it was already almost identical.

app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
jpelgrom: I agree with you but decided to follow the existing code style to keep changes in this PR to a minimum (it is already suffering from feature creep in my opinion in WebSocketRepositoryImpl). If merged I can submit another PR for this?

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
jpelgrom: There aren't any typos right now :) But it is nice to have when used in more than one place and it looks like the iOS websocket implementation also uses constants for this situation - changed.

app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
dshokouhi: Lets make sure to keep our PRs down to a minimum as it helps make the review process easier. It can be very easy to keep finding things to fix but it doesn't mean it should all be done in one PR. The smaller the PR the easier it is to get reviewed and merged.

zmarties: My point was that doing it in the superclass probably would have made the PR smaller - just do it once, not have to add it separately to 11 classes.  (Fixing the existing repeated code to set the title I agree is going beyond the scope of this PR).  More than happy that these can be addressed in a follow on PR.

jpelgrom: Perhaps it would've saved a line or two of code in the end, but we're talking about 9 new lines of code which are easy to review whereas @zmarties' suggestion will require changing the interface to a class, updating all controls accordingly, and checking everything to see whether it is unique to the control or not and should go the the HaControl class.
Marking this as resolved, improvements to go in a separate PR.

JBassett (approved):

dshokouhi:
@jpelgrom just a FYI as we dont get notified via email, this branch has some merge conflicts  to resolve :)

jpelgrom:
> @jpelgrom just a FYI as we dont get notified via email, this branch has some merge conflicts to resolve :)

Noticed it about an hour ago and will solve the conflict + update the added functions to also use return try/catch later today.

jpelgrom:
Conflicts should be resolved, and based on a quick test everything still works as it did before merging.

(force pushed because I forgot to run ktlint before committing)

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
```diff
@@ -64,11 +74,11 @@ class WebSocketRepositoryImpl @Inject constructor(
     private var connection: WebSocket? = null
     private val connectedMutex = Mutex()
     private var connected = Job()
-    private val stateChangedMutex = Mutex()
-    private var stateChangedFlow: SharedFlow<StateChangedEvent>? = null
+    private val eventSubscriptionMutex = mutableMapOf<String, Mutex>()
```
JBassett, line 77: The more I'm thinking about this, we should just use a single Mutex rather than one per event type.  Technically this could cause a race condition the way it's written.

JBassett (changes requested):

JBassett:
Sorry for the last minute change request @jpelgrom

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
jpelgrom: I didn't really think much about it, but I guess it could if you try to subscribe to the same event again within a few ms. Should be a fairly quick message so any delay because you're now waiting for other subscribe_event calls to complete is negligible.

jpelgrom:
No problem, it was a good request :)

Speaking of which, I managed to reproduce the race condition I mentioned in the first bullet point of 'Any other notes'. Looks like cancelling the CoroutineScope which fires multiple requests at the same time does create a race condition with IDs and then time used to serialize the message (not much else happening inbetween getting ID and sending the message as we're already connected?).
I can reproduce this when closing the device controls, it happens in about 1 out of 4/5 times. We could 'solve' this for now by re-using the Mutex in the `awaitClose` action, but technically this could happen any time messages are sent very quickly after another using `sendMessage` (and that function's code doesn't change in this PR) (but in the current stable version messages so quickly after another is very unlikely to happen)...

```
2021-12-24 21:23:26.399 8166-8189/io.homeassistant.companion.android.debug D/HaConProService: cancel
2021-12-24 21:23:26.403 8166-8405/io.homeassistant.companion.android.debug D/WebSocketRepository: Unsubscribing from area_registry_updated
2021-12-24 21:23:26.405 8166-8799/io.homeassistant.companion.android.debug D/WebSocketRepository: Unsubscribing from device_registry_updated
2021-12-24 21:23:26.406 8166-8801/io.homeassistant.companion.android.debug D/WebSocketRepository: Sending message number 30: {type=unsubscribe_events, subscription=24, id=30}
2021-12-24 21:23:26.406 8166-8360/io.homeassistant.companion.android.debug D/WebSocketRepository: Unsubscribing from entity_registry_updated
2021-12-24 21:23:26.407 8166-8799/io.homeassistant.companion.android.debug D/WebSocketRepository: Sending message number 32: {type=unsubscribe_events, subscription=26, id=32}
2021-12-24 21:23:26.407 8166-8402/io.homeassistant.companion.android.debug D/WebSocketRepository: Sending message number 31: {type=unsubscribe_events, subscription=25, id=31}
2021-12-24 21:23:26.408 8166-8799/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number 32 sent
2021-12-24 21:23:26.409 8166-8402/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number 31 sent
2021-12-24 21:23:26.409 8166-8801/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number 30 sent
2021-12-24 21:23:26.431 8166-8349/io.homeassistant.companion.android.debug D/WebSocketRepository: Websocket: onMessage (text)
(...)
2021-12-24 21:23:26.438 8166-8349/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number 32 received: {"id": 32, "type": "result", "success": true, "result": null}
2021-12-24 21:23:26.438 8166-8349/io.homeassistant.companion.android.debug D/WebSocketRepository: Websocket: onMessage (text)
2021-12-24 21:23:26.440 8166-8349/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number 30 received: {"id": 30, "type": "result", "success": false, "error": {"code": "id_reuse", "message": "Identifier values have to increase."}}
2021-12-24 21:23:26.441 8166-8349/io.homeassistant.companion.android.debug D/WebSocketRepository: Websocket: onMessage (text)
2021-12-24 21:23:26.442 8166-8349/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number 31 received: {"id": 31, "type": "result", "success": false, "error": {"code": "id_reuse", "message": "Identifier values have to increase."}}
(...)
```

JBassett (approved):

PR #2028, opened 2021-12-19
Replace huge camera widget example image with a sensible sized one

zmarties:
Looking at the built apk, about 12% of the whole download size was being
accounted for by a huge image, that is only used to give an example image
in the camera widget.

Two methods have reduced this to an insignificant size now:
- use a lower resolution image, no need for this to be many megapixels
- store it as a jpg, which is great for photos, rather than a png


dshokouhi (approved):
Looks good to me, good catch! I dont think android cares about it being a JPG but have you made sure the image still loads in the widget picker?

zmarties:
Yes, I verified this loads in the widget picker as expected.  (Android is smart - the resource is actually just specified giving the filename part, and the extension is immaterial, as long as Andoid can cope with the file format).

JBassett (approved):

PR #2033, opened 2021-12-20
Wear OS sensors UI

Kisty:
## Summary

In order to add the other sensors from mobile, the wear app needs to be able to control the sensors. Currently, battery snesors are added.

## Screenshots

Decided with team to port to Compose, rather than the existing settings method for the mobile app. 
**Draft:** The below is just a quick PoC using same method as the mobile app.
Screenshot_20211220_145843
Screenshot_20211220_145848

- [ ] Documentation: home-assistant/companion.home-assistant#

## Any other notes

#1756


Kisty:
Will leave enabling the other sensors for now to keep PR as small as possible

wear/src/main/res/xml/sensors.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dshokouhi, line 1: we dont need this file anymore as we are now using compose

wear/src/main/res/xml/sensor_detail.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dshokouhi, line 1: I dont think we need this anymore as we are using compose now

wear/src/main/res/layout/activity_settings.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dshokouhi, line 1: I dont think we need this anymore as we are using compose now

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorsView.kt
```diff
@@ -0,0 +1,99 @@
...
+            positionIndicator = {
+                if (scalingLazyListState.isScrollInProgress)
+                    PositionIndicator(scalingLazyListState = scalingLazyListState)
+            },
+            timeText = { TimeText(!scalingLazyListState.isScrollInProgress) }
+        ) {
+            val sensorManagers = getSensorManagers()
+            ScalingLazyColumn(
+                modifier = Modifier
+                    .fillMaxSize(),
```
dshokouhi, line 50: I think we need to add the rotary input handler here so we can scroll with the bezel like we can in other screens. Eventually we wont need to add this to every screen but I am not sure when Google will add the native functionality :)

As of now it would need to be added on each new page we introduce, just to keep things consistent.

wear/build.gradle.kts
```diff
@@ -91,6 +91,7 @@ dependencies {
 
     implementation("com.google.android.material:material:1.5.0")
 
+    implementation("androidx.preference:preference-ktx:1.1.1")
```
dshokouhi, line 94: Probably dont need this as we are using compose now?

dshokouhi (commented):
Just some minor comments for now based on a very quick look at the code. Also I think this file needs to be cleaned up as well since it was moved out in a previous release. 

`full/java/io/homeassistant/companion/android/notifications/NotificationContentReceiver.kt`

I did test the debug APK and the UI and sensor updates do seem to be working as expected with respect to being disabled and what not. Great job!!

Kisty:
Rebased and I think it's ready for merge

JBassett (approved):

PR #2066, opened 2021-12-23
Change sentry exception filtering and catch some potential unhandled erros

dshokouhi:
## Summary

Updates sentry error filtering to new format and replaces a deprecated sentry call.

Also added a few try/catch blocks to ensure all failures are handled appropriately. Ran some tests against device controls and a few widgets to ensure nothing is broken. 

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
@@ -88,10 +88,12 @@ class HaControlsProviderService : ControlsProviderService() {
                         controlIds.forEach {
                             val entity = integrationRepository.getEntity(it)
                             val domain = it.split(".")[0]
-                            val control = domainToHaControl[domain]?.createControl(
-                                applicationContext,
-                                entity
-                            )
+                            val control = entity?.let { it1 ->
+                                domainToHaControl[domain]?.createControl(
+                                    applicationContext,
+                                    it1
+                                )
+                            }
                             subscriber.onNext(control)
                         }
 
--- a/app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
@@ -88,12 +88,10 @@ class HaControlsProviderService : ControlsProviderService() {
                         controlIds.forEach {
                             val entity = integrationRepository.getEntity(it)
                             val domain = it.split(".")[0]
-                            val control = entity?.let { it1 ->
-                                domainToHaControl[domain]?.createControl(
-                                    applicationContext,
-                                    it1
-                                )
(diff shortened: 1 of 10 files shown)


common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
```diff
@@ -129,36 +144,39 @@ class WebSocketRepositoryImpl @Inject constructor(
...
-                        producerScope = null
-                        stateChangedFlow = null
-                    }
-                }.shareIn(ioScope, SharingStarted.WhileSubscribed())
+                    }.shareIn(ioScope, SharingStarted.WhileSubscribed())
+                }
             }
-
-            return stateChangedFlow!!
+        } catch (e: Exception) {
```
balloob, line 176: Ideally you would catch the exceptions that we expect, not all of them. Or else we'll miss actual bugs.

dshokouhi: Oh I thought I was doing that by placing the entire block in a try/catch :)

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
```diff
@@ -129,36 +144,39 @@ class WebSocketRepositoryImpl @Inject constructor(
...
+            Log.e(TAG, "Unable to get flow of entities", e)
         }
+        return stateChangedFlow!!
```
JBassett, line 179: This will more than likely throw another exception, we should return an `emptyFlow()`

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
```diff
@@ -88,10 +88,12 @@ class HaControlsProviderService : ControlsProviderService() {
                         controlIds.forEach {
                             val entity = integrationRepository.getEntity(it)
                             val domain = it.split(".")[0]
-                            val control = domainToHaControl[domain]?.createControl(
-                                applicationContext,
-                                entity
-                            )
+                            val control = entity?.let { it1 ->
```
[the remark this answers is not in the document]
dshokouhi: I had actually reverted this change when I made the switch from integration repository to websocket repository :)

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
dshokouhi: I was worried about that, done

JBassett (approved):

PR #2073, opened 2021-12-24
Fix getStateChanges call always returning empty flow

jpelgrom:
## Summary

The commit made for [this comment in PR #2066](https://github.com/home-assistant/android/pull/2066#discussion_r774740738) introduced a new bug: a call to `getStateChanges()` now always returns an empty flow, meaning subscribing to the websocket does nothing (and widgets / device controls no longer update until you force a reload).

## Screenshots

n/a

Not needed (bugfix)

## Any other notes

This code is also in the currently released stable version of the app, 2012.12.2, should I (also) have made an issue?


jpelgrom:
Fixing to the review comment doesn't seem to work, so here's [a link to the commit I'm referring to](https://github.com/home-assistant/android/pull/2066/commits/033d5e518db66765234850cbaab641a388bc522f) as well

dshokouhi:
🤦‍♂️ good catch!!

JBassett:
Saw the resolved but didn't double check the change, it's on me.  Looks like we are getting another release sooner rather than later.

JBassett (approved):

PR #2080, opened 2021-12-26
Add last used app sensor

dshokouhi:
## Summary

Adds a new sensor that reports the last used app since the last update. Currently this sensor does not update immediately so will need to look into methods for faster updates. This sensor requires a special permission unlike the others we have so had to add additional checks for it.  We will try to determine the application label and if unable to determine a label we will send the package name.

## Screenshots

image

image

Documentation: home-assistant/companion.home-assistant#647

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/LastAppSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/LastAppSensorManager.kt
@@ -1,7 +1,6 @@
 package io.homeassistant.companion.android.sensors
 
 import android.Manifest
-import android.app.usage.UsageStats
 import android.app.usage.UsageStatsManager
 import android.content.Context
 import android.content.pm.PackageManager
@@ -9,7 +8,6 @@ import android.os.Build
 import android.util.Log
 import androidx.annotation.RequiresApi
 import io.homeassistant.companion.android.common.sensors.SensorManager
-import java.util.TreeMap
 import io.homeassistant.companion.android.common.R as commonR
 
 class LastAppSensorManager : SensorManager {
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/LastAppSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/LastAppSensorManager.kt
@@ -1,0 +1,77 @@
+package io.homeassistant.companion.android.sensors
+
+import android.Manifest
+import android.app.usage.UsageStats
+import android.app.usage.UsageStatsManager
+import android.content.Context
+import android.os.Build
+import androidx.annotation.RequiresApi
+import io.homeassistant.companion.android.common.sensors.SensorManager
+import java.util.TreeMap
+import io.homeassistant.companion.android.common.R as commonR
+
+class LastAppSensorManager : SensorManager {
+    companion object {
+        private const val TAG = "LastApp"
+
+        val last_used = SensorManager.BasicSensor(
+            "last_used_app",
+            "sensor",
+            commonR.string.basic_sensor_name_last_used_app,
+            commonR.string.sensor_description_last_used_app
+        )
+    }
+
+    override fun docsLink(): String {
+        return "https://companion.home-assistant.io/docs/core/sensors#last-used-app-sensor"
+    }
+    override val enabledByDefault: Boolean
+        get() = false
+    override val name: Int
+        get() = commonR.string.sensor_name_last_app
+
+    override fun getAvailableSensors(context: Context): List<SensorManager.BasicSensor> {
+        return listOf(last_used)
+    }
+
+    override fun hasSensor(context: Context): Boolean {
+        return Build.VERSION.SDK_INT >= Build.VERSION_CODES.M
+    }
+    @RequiresApi(Build.VERSION_CODES.M)
+    override fun requiredPermissions(sensorId: String): Array<String> {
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/LastAppSensorManager.kt
```diff
@@ -0,0 +1,88 @@
...
+        val current = System.currentTimeMillis()
+        val appList = usageStats.queryUsageStats(UsageStatsManager.INTERVAL_DAILY, current - 1000 * 1000, current)
+        val sortedMap = TreeMap<Long, UsageStats>()
+        if (!appList.isNullOrEmpty() && appList.size > 0) {
+            for (item in appList) {
+                sortedMap[item.lastTimeUsed] = item
+            }
+        }
+
+        var lastApp = sortedMap[sortedMap.lastKey()]?.packageName ?: "none"
```
JBassett, line 69: You should be able to do this using `maxByOrNull`, it'll make things much simpler.

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorManager.kt
```diff
@@ -42,7 +43,15 @@ interface SensorManager {
...
-            context.checkPermission(it, myPid(), myUid()) == PackageManager.PERMISSION_GRANTED
+            if (sensorId != "last_used_app")
+                context.checkPermission(it, myPid(), myUid()) == PackageManager.PERMISSION_GRANTED
+            else {
+                val pm = context.packageManager
+                val appInfo = pm.getApplicationInfo(context.packageName, 0)
+                val appOpsManager = context.getSystemService(Context.APP_OPS_SERVICE) as AppOpsManager
+                val mode = appOpsManager.checkOpNoThrow(AppOpsManager.OPSTR_GET_USAGE_STATS, appInfo.uid, appInfo.packageName)
+                mode == AppOpsManager.MODE_ALLOWED
+            }
```
JBassett, line 54: When happens when we try to check for the permission like we check for the other?

app/src/main/java/io/homeassistant/companion/android/sensors/SensorDetailFragment.kt
```diff
@@ -127,6 +129,13 @@ class SensorDetailFragment(
                                         requestPermissions(permissions)
                                     }
                                 )
+                            } else if (sensorManager is LastAppSensorManager) {
+                                val pm = context.packageManager
+                                val appInfo = pm.getApplicationInfo(context.packageName, 0)
+                                val appOpsManager = context.getSystemService(Context.APP_OPS_SERVICE) as AppOpsManager
+                                val mode = appOpsManager.checkOpNoThrow(AppOpsManager.OPSTR_GET_USAGE_STATS, appInfo.uid, appInfo.packageName)
+                                if (mode != AppOpsManager.MODE_ALLOWED)
+                                    requestPermissions(permissions)
```
JBassett, line 138: This is duplicate code that should be turned into a common function if we really need it.

JBassett (changes requested):

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorManager.kt
dshokouhi: it continues to return false when we check for the permission in the manifest, its strange honestly but I think because its a protected permission? I was fighting with our permission check until I added t his in :)

app/src/main/java/io/homeassistant/companion/android/sensors/LastAppSensorManager.kt
dshokouhi: This change continues to show an incorrect app. I tried to use it like, did I not do it properly?

`var lastApp = sortedMap.maxByOrNull { it.value.packageName } ?: "none"`

dshokouhi: actually it was my bad, I got it now :)

app/src/main/java/io/homeassistant/companion/android/sensors/LastAppSensorManager.kt
```diff
@@ -0,0 +1,88 @@
...
+        val current = System.currentTimeMillis()
+        val appList = usageStats.queryUsageStats(UsageStatsManager.INTERVAL_DAILY, current - 1000 * 1000, current)
+        val sortedMap = TreeMap<Long, UsageStats>()
+        if (!appList.isNullOrEmpty() && appList.size > 0) {
+            for (item in appList) {
+                sortedMap[item.lastTimeUsed] = item
+            }
+        }
+
+        var lastApp = sortedMap.maxByOrNull { it.key }?.value?.packageName ?: "none"
```
JBassett, line 69: This is what I was suggesting:
```suggestion
        val lastApp = usageStats.queryUsageStats(UsageStatsManager.INTERVAL_DAILY, current - 1000 * 1000, current).maxByOrNull { it.lastTimeUsed }?.packageName ?: "none"
```

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/sensors/LastAppSensorManager.kt
dshokouhi: thank you for that :) that is much better!

JBassett (approved):

Issue #2056, opened 2021-12-22
Websockets unable to send message


dshokouhi:
**Home Assistant Android version:**

2021.12.0

**Android version:**

Any

**Phone model:**

Any

**Home Assistant version:**

Hard to determine from sentry error

**Last working Home Assistant release (if known):**

n/a as websockets is new

**Description of problem:**

Looking at sentry this error is impacting a good chunk of users, over 40k to be precise

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```
java.lang.Exception: Unable to send message: {type=unsubscribe_events, subscription=287}
    at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.sendMessage(WebSocketRepositoryImpl.kt:212)
    at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.access$sendMessage(WebSocketRepositoryImpl.kt:47)
    at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl$sendMessage$1.invokeSuspend
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
    at kotlinx.coroutines.internal.ScopeCoroutine.afterResume(Scopes.kt:33)
    at kotlinx.coroutines.AbstractCoroutine.resumeWith(AbstractCoroutine.kt:102)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:46)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
    at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

**Screenshot of problem:**

**Additional information:**

JBassett:
I'm not sure this is a crashing error but the fix is larger than I want to do right now before a point release.

PR #2091, opened 2021-12-30
Work on error handling from Websocket.

JBassett:
## Summary

This PR attempts to better handle exceptional cases better that throwing exceptions and allowing them to crash the app.  This should hopefully gracefully stop working rather than hard crashing.

Fixes: https://github.com/home-assistant/android/issues/2056


dshokouhi (approved):
tested both phone and wear APK all seem to be ok

PR #2094, opened 2021-12-30
Group device controls by area first, if set, when adding them

jpelgrom:
## Summary

(Submitted as a result of a discussion in the Discord mobile apps dev channel)
When adding device controls order them by area first, if set, and only then by domain. This matches the autogenerated Lovelace dashboard and probably offers the best user experience.

## Screenshots

The UI is provided by the system, this is on a Google Pixel 4a running Android 12. This screenshot is not complete because it is a very long list for me but the result should be obvious: 'Beneden' and 'Boven' are areas I have set and after that it shows devices not in any area by domain (such as Automation and Fan in the screenshot).
ha-controls-areas-grouped

Probably not needed?


dshokouhi (approved):
You are quick! Looks good to me!

JBassett (approved):

PR #2106, opened 2022-01-04
Initial commit of the Quest build

dshokouhi:
## Summary

Introduces 2 new build variants for the Quest as a release version and a debug version. Hopefully I got everything right. The idea here was to make it match the `minimal` flavor while also complying with the pre-requisites from Oculus. Also went ahead and hid some features from settings that are just not available on this flavor. 

https://developer.oculus.com/resources/publish-uploading-mobile/

As of now notifications are only allowed for apps that are published to the Oculus Quest store so until we get to that point we cannot test that feature. In terms of sensors we have 47 of them and all seem to be functional. Some good sensors for this device asides from battery and charging status would be the new Last Used app sensor and also the Interactive sensor to determine when the device is in use. Last notification and Last Removed Notification don't appear to be functional here however Active Notifications does work. Intents work as expected so the Interactive sensor updates immediately, same with charging the device. Did not notice any issues with crashing and sensor worker starts right up with the device when it powers on.

Documentation: home-assistant/companion.home-assistant#pending probably will come later once we pass the review process

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt
@@ -100,7 +100,7 @@ class LaunchActivity : AppCompatActivity(), LaunchView {
                 val deviceName = intent.getStringExtra("DeviceName").toString()
                 val deviceTrackingEnabled = intent.getBooleanExtra("LocationTracking", false)
                 val messagingToken = getMessagingToken()
-                if (messagingToken.isNullOrBlank()) {
+                if (messagingToken.isNullOrBlank() && BuildConfig.FLAVOR == "full") {
                     AlertDialog.Builder(this@LaunchActivity)
                         .setTitle(commonR.string.firebase_error_title)
                         .setMessage(commonR.string.firebase_error_message)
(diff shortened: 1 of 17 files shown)


app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt
```diff
@@ -100,7 +100,7 @@ class LaunchActivity : AppCompatActivity(), LaunchView {
                 val deviceName = intent.getStringExtra("DeviceName").toString()
                 val deviceTrackingEnabled = intent.getBooleanExtra("LocationTracking", false)
                 val messagingToken = getMessagingToken()
-                if (messagingToken.isNullOrBlank()) {
+                if (messagingToken.isNullOrBlank() && BuildConfig.FLAVOR == "full") {
```
jpelgrom, line 103: While this makes sense for the Quest where no notifications when sideloading is the default, shouldn't minimal users still get the warning about no push notifications? (at least until the local push PR is merged)

dshokouhi: Well minimal users never had notifications so the pop-up doesn't really make sense for them, I also think in the past we skipped this message but I cant recall. Its really intended for the full version in case we fail to register for notifications which typically happens in debug builds. For the minimal version its mentioned on F-Droid as well as our documentation that notifications are not supported. The notification section is also hidden in settings for these flavors.

https://companion.home-assistant.io/docs/core/android-flavors

JBassett (approved):

PR #1970, opened 2021-11-27
Wear redesign

leroyboerefijn:
## Summary
The `MainView` has been redesigned based on the discussion in #1795:

- Kept favorites list at the top
- Moved other entities behind a menu, sorted on domains

Note that I see this redesign as a pre-requisite for Area support

In addition, I made some general design changes to better align with Material Design guidelines. Finally, I implemented a `MaterialTheme` to better manage colors used in compose. Sorry if the PR has become a bit large!

## Screenshots
Main view starts with favorites:
Screenshot_20211127_134215_debug

Then comes a menu to select entities:
Screenshot_20211127_134224_debug

For example, the Scenes menu:
Screenshot_20211127_134238_debug

And on the bottom, the logout `Chip` has been moved to the settings menu:
Screenshot_20211127_134232_debug

## Link to pull request in Documentation repository

Documentation: home-assistant/companion.home-assistant#

PR #2118, opened 2022-01-05
Add areas to entity list on Wear

### LINKED ISSUE #1795: Area support for lights in WearOS app

Navigating the lights is slow due to them all being presented in a list.

Please could you add support for Areas (eg Kitchen, Bedroom etc) and have this be the default view ? It would be much quicker to find the correct light to toggle this way.

N/A

**Additional context**

None.

jpelgrom:
## Summary

Based on the work in #1970, this PR uses area information for entities to improve the main list shown on Wear OS. Should fix #1795.
I haven't worked on Wear OS or with Compose before, so happy to get any feedback from more frequent Wear OS contributors 🙂

Still to do:

- [x] Fix 'All entities' chip
- [x] Fix setting favorites + tiles from the watch settings
- [x] Sort area and domain chips/headers alphabetically, the current implementation with a map has no stable order
- [x] Listen for registry update events and refresh the list
- [x] Further tests for when all entities are in areas / no entities are in areas

Further possible improvements: 

- [x] ~Find a nicer way to create the lists while still allowing Compose to detect the changes (not considering favorites and shortcuts, we need to be able to make combinations of areas with entities in them, entities by area, and entities by domain)~ Edit: I've now implemented this by creating main entities and areas lists, and nested lists grouping entities by area and domain. Any further filtering is done in Compose which should be quick and allow it to stay live.
- [x] ~The code for determining the area of an entity is now used in both the Wear entity list and (phone) device controls, this should probably be moved somewhere else so it can be shared~ Created a utility object in the common code of the app for this. The function would typically be placed in a ViewModel or repository but there is no shared VM and the WebSocketRepository doesn't follow a repository pattern and do caching/transforming.

## Screenshots

Favorites is still at the top, directly below that the user will see their areas with entities:
ha-wear-1

Tapping on an area shows a list of all entities in the area:
ha-wear-2
(the first item is an input boolean, the second a light, ...)

Back on the main screen, below the areas the app shows chips for domains with entities not in any area (so 'More entities' refers to more than in the areas):
ha-wear-3

And finally at the bottom of the list is the chip for 'All entities', which shows all entities regardless of area status, and the settings:
ha-wear-4

If the user has no areas with supported entities in them, nothing is shown and below favorites the app shows domains like it currently does:
ha-wear-5

If the user has placed all the supported entities in an area, the app won't show any domains and just end with the all entities and settings chips.
ha-wear-6

Documentation: home-assistant/companion.home-assistant#653

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
@@ -13,6 +13,7 @@ import io.homeassistant.companion.android.common.data.websocket.WebSocketReposit
 import io.homeassistant.companion.android.common.data.websocket.impl.entities.AreaRegistryResponse
 import io.homeassistant.companion.android.common.data.websocket.impl.entities.DeviceRegistryResponse
 import io.homeassistant.companion.android.common.data.websocket.impl.entities.EntityRegistryResponse
+import io.homeassistant.companion.android.util.RegistriesDataHandler
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.cancel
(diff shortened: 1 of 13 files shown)


jpelgrom:
OK, I think this is ready for testing and review now! Screenshots have also been updated.

Based on the emulator there isn't any noticeable performance impact aside from slightly longer initial loading, hopefully that is also true for real watches.

While I tried not to make any large architectural changes, working with 0-n areas and (not) displaying entities depending on whether you view them from an area list or somewhere else, you can't copy the approach of 6 fixed lists for domains. There are now two main lists, areas with entities and domains with entities, and depending on what is needed the app filters or loops their contents. I think this is the best approach when you also need to give Compose something to observe.
(Also pretty amazing how much lines of code this removes)

dshokouhi (approved):
Tested the debug APK and it works well, thank you for taking this on!

JBassett (approved):

PR #2128, opened 2022-01-08
Add Mount State sensor to the Quest

SkechyWolf:
## Summary

Adds a mount state sensor to the quest
Shows as a binary sensor on the frontend 

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#659

## Any other notes

N/A

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
@@ -50,6 +50,7 @@ class SensorReceiver : SensorReceiverBase() {
             PowerSensorManager(),
             PressureSensorManager(),
             ProximitySensorManager(),
+            QuestSensorManager(),
             StepsSensorManager(),
             StorageSensorManager(),
             TimeZoneManager(),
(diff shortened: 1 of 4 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
```diff
@@ -50,6 +50,7 @@ class SensorReceiver : SensorReceiverBase() {
             PowerSensorManager(),
             PressureSensorManager(),
             ProximitySensorManager(),
+            QuestSensorManager(),
```
dshokouhi, line 53: We need to also add the intent action string and sensor ID to the `skippableActions` below

dshokouhi (changes requested):
Great work the sensor is reporting as expected when I tested the debug APK! Just 1 minor comment so we can skip unwanted updates if the user did not want it enabled.

app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
SkechyWolf: Should be fixed

dshokouhi (approved):

dshokouhi:
Just noticed the docs link goes to an already merged PR, lets update the docs to mention this new sensor on the Quest specific page please.

JBassett (approved):

PR #2137, opened 2022-01-12
Assign input types for some onboarding fields

ndarilek:
- Use "text" input type so first letter of username isn't automatically capitalized.
- Use URI type for manual URL so app stops autocorrecting my instance URI.

### DIFF
--- a/app/src/main/res/layout/dialog_authentication.xml
+++ b/app/src/main/res/layout/dialog_authentication.xml
@@ -17,6 +17,7 @@
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:hint="@string/username"
+            android:inputType="text"
             android:singleLine="true" />
     </com.google.android.material.textfield.TextInputLayout>
 
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupView.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupView.kt
@@ -54,7 +54,7 @@ fun ManualSetupView(
             modifier = Modifier.align(Alignment.CenterHorizontally),
             label = { Text(stringResource(id = commonR.string.input_url)) },
             singleLine = true,
-            keyboardOptions = KeyboardOptions(imeAction = ImeAction.Done),
+            keyboardOptions = KeyboardOptions(keyboardType = KeyboardType.Uri, imeAction = ImeAction.Done),
             keyboardActions = KeyboardActions(
                 onDone = {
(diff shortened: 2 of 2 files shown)


app/src/main/res/layout/dialog_authentication.xml
```diff
@@ -17,6 +17,7 @@
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:hint="@string/username"
+            android:inputType="text"
```
dshokouhi, line 20: I dont think this is related to the Manual Setup View, this screen is used in the webview when a user hits an authentication dialog which is after being logged in.

ndarilek:
Got it. Any idea how I'd prevent the app from capitalizing my username 
when I enter it? Or are they not case-sensitive? This doesn't happen on 
my desktop so I didn't think it was something in the WebView itself.

Whenever signing in, the first thing I have to do is uncapitalize my 
username. The second is fixing all Android's attempts to autocomplete my 
URL to "http://home assistant. local" Hoping to get these fixed.

Thanks.

ndarilek:
Is there a way to do that without having to uninstall the app from my 
device and reassociate everything? I'm a screen reader user, emulators 
aren't easy at all to set up with TalkBack enabled, and I really don't 
want to undo my on-device HA setup. I believe this should explicitly set 
at least the URI input correctly, which will disable autocompleting my 
manual URL.

Also, if I'm reading the CI failure correctly, these lints don't appear 
to be in anything changed on my end. Am I misreading? The only way to 
accessibly view CI failures is to pull up the raw logs, and those are 
pretty noisy.

Thanks.

dshokouhi:
> Is there a way to do that without having to uninstall the app from my device and reassociate everything? I'm a screen reader user, emulators aren't easy at all to set up with TalkBack enabled, and I really don't want to undo my on-device HA setup. I believe this should explicitly set at least the URI input correctly, which will disable autocompleting my manual URL. 

when you use android studio it creates a debug version of the app which is treated like a separate app and has a different icon so it is easy to distinguish and can be installed side by side the production version without harm.

> Also, if I'm reading the CI failure correctly, these lints don't appear to be in anything changed on my end. Am I misreading? The only way to accessibly view CI failures is to pull up the raw logs, and those are pretty noisy. Thanks.

The failure is actually 

```
e: /home/runner/work/android/android/app/src/main/java/io/homeassistant/companion/android/onboarding/manual/ManualSetupView.kt: (57, 62): Unresolved reference: KeyboardType
```

Which means it is missing the import :)

JBassett:
Closing due to no response.

PR #2142, opened 2022-01-13
Show names on shortcuts tile

### LINKED ISSUE #1976: Wear OS - Add custom text to tile icon
Using tile to change multiple light groups (living room, kitchen, bedroom etc.) you cannot separate icons anyhow.

One solution is to make icon little smaller and add small custom Text under icon.
OR add one letter to top of light bulb so I could manually change that to something useful. 
OR add big letter behind light bulb and make that icon transparent so I can see the letter.

I would be happy to have this configure in home-assistant server config, home assistant android app file OR even somehow in wear os app.

Ability to rearrange icons might even work for me (so I could make those show almost as in my floor plan). 
But rearrange icons + some extra custom info would be best combo.

**Additional context**
What multiple light groups shows now:
image

leroyboerefijn:
## Summary
Adds the option to show names on shortcuts tile. Fixes #1976

## Screenshots
Screenshot_20220113_220629_sysui

Not needed

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/tiles/ShortcutsTile.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/tiles/ShortcutsTile.kt
@@ -8,6 +8,7 @@ import androidx.core.graphics.drawable.toBitmap
 import androidx.wear.tiles.ActionBuilders
 import androidx.wear.tiles.ColorBuilders.argb
 import androidx.wear.tiles.DimensionBuilders.dp
+import androidx.wear.tiles.DimensionBuilders.sp
 import androidx.wear.tiles.LayoutElementBuilders
 import androidx.wear.tiles.LayoutElementBuilders.Box
 import androidx.wear.tiles.LayoutElementBuilders.Column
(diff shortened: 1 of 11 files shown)


jpelgrom:
For consistency and to avoid the main settings list from getting to long, can I suggest moving the 'show text on tile' setting to the Shortcuts tile's composable, just like the template tile refresh interval setting? So something like:
device-2022-01-13-235901

leroyboerefijn:
Good suggestion! I'll copy your implementation!

wear/src/main/java/io/homeassistant/companion/android/tiles/ShortcutsTile.kt
```diff
@@ -211,10 +219,34 @@ class ShortcutsTile : TileService() {
...
+                .setWidth(dp(iconSize))
+                .setHeight(dp(iconSize))
                 .build()
         )
+        if (showLabels) {
+            addContent(
+                LayoutElementBuilders.Arc.Builder()
+                    .addContent(
+                        LayoutElementBuilders.ArcText.Builder()
+                            .setText(entity.friendlyName)
```
dshokouhi, line 231: I know this will probably vary from device to device to device but should we consider some type of character limit here? What happens if a user has a large friendly name that wraps around the icon?

leroyboerefijn: Good question, I'll run a test

leroyboerefijn: Screenshot_20220115_104340_sysui

The one on the bottom left is now clipping. So it's not fully overlapping but a tiny bit weird

leroyboerefijn:
Settings were updated as per request of @jpelgrom

wear/src/main/java/io/homeassistant/companion/android/tiles/ShortcutsTile.kt
dshokouhi: fair amount of characters are shown probably ok to let the user decide if they want to change their name or deal with it

JBassett (approved):

PR #2148, opened 2022-01-14
Fix handling of line feed control char

chriss158:
## Summary

When sending a message to an Iphone and an Android phone, it is currently necessary to distinguish between a newline control character. 

The Iphone uses `\n` and Android uses the html code `<br>`.

This is a problem when a message is sent via a Notify group consisting of an Iphone and Android. 

```
notify:
  - name: all_devices
    platform: group
    services:
      - service: mobile_app_android
      - service: mobile_app_iphone
```
```
service: notify.all_devices
data:
  message: test1\n\ntest2\n\ntest3
```

As a simple solution we just replace all `\n` with `<br>`. Then we can also use the control character `\n` for Android phones.

## Screenshots

image


dshokouhi (approved):
Always a good thing to keep notifications lined up like this :)

JBassett (approved):

Issue #2055, opened 2021-12-22
WearOS - Multi-factor Authentication


ch4d1:
**Is your feature request related to a problem? Please describe.**

The WearOS companion app currently does not seem to have implemented multi-factor authentication. Logging in with an account that has this activated is therefore not possible.
**Describe the solution you'd like**

Implementation of multi-factor authentication
**Describe alternatives you've considered, if any**

Authentication via the smartphone app. Entering login information via the WearOS keyboard is a pain.
**Additional context**

dshokouhi:
You can actually login using your phone already and that works with MFA

mikkelzzz3:
@dshokouhi You can login on the Wear OS app using your phone? How does that work. I only have option to input username and password, and then it says afterwards it couldn't connect. (Because I have MFA)

dshokouhi:
> @dshokouhi You can login on the Wear OS app using your phone? How does that work. I only have option to input username and password, and then it says afterwards it couldn't connect. (Because I have MFA)

Open the phone app, go to App configuraiton or configuration > companion app. Then go to Wear OS Settings and once it finds your device login from there.  MFA doesnt work directly from the watch yet but using the phone app we can perform MFA as the phone app has support for it so it can forward that to the watch

ch4d1:
> > @dshokouhi You can login on the Wear OS app using your phone? How does that work. I only have option to input username and password, and then it says afterwards it couldn't connect. (Because I have MFA)
> 
> Open the phone app, go to App configuraiton or configuration > companion app. Then go to Wear OS Settings and once it finds your device login from there. MFA doesnt work directly from the watch yet but using the phone app we can perform MFA as the phone app has support for it so it can forward that to the watch

It works but should be mentioned somehow when setting the app up.

dshokouhi:
> It works but should be mentioned somehow when setting the app up.

We shouldn't mention a workaround in the UI, we should actually fix the issue which is getting the MFA prompt on a wearable. Workarounds for issue trackers to help users get by until things are fixed.

ch4d1:
> > It works but should be mentioned somehow when setting the app up.
> 
> We shouldn't mention a workaround in the UI, we should actually fix the issue which is getting the MFA prompt on a wearable. Workarounds for issue trackers to help users get by until things are fixed.

Sure thing, but it would be best to offer the workaround as an official method, as the default wear os keyboard is missing symbols. Also typing in longer passwords is still a pain.

PR #2154, opened 2022-01-18
Add 'Sign in on phone' button to Wear OS login

jpelgrom:
## Summary

The phone app allows you to sign in to your connected Wear OS device, let's show the user this is possible on the watch.

While it is great that you can sign in directly on the watch, entering all the required information on such a small screen can be annoying or sometimes not possible (due to available characters on the IME). This PR will add a check to see if the companion app is available and supports signing in, and if so adds a button to open the app setting from the watch.  After signing in on the phone, the Wear OS app automatically reloads (this was already present).

See also: comments on #2055 (users didn't know they could sign in on the phone), [sign-in best practices on developer.android.com](https://developer.android.com/training/wearables/design/sign-in).

## Screenshots

wear-sign-in-on-phone
wear-continue-on-your-phone

There is no documentation on authentication at all, so I don't think this is needed

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/onboarding/OnboardingActivity.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/onboarding/OnboardingActivity.kt
@@ -9,6 +9,7 @@ import android.view.View
 import androidx.appcompat.app.AppCompatActivity
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.wear.activity.ConfirmationActivity
+import androidx.wear.remote.interactions.RemoteActivityHelper
 import androidx.wear.widget.WearableRecyclerView
 import com.google.android.gms.tasks.Tasks
 import com.google.android.gms.wearable.CapabilityClient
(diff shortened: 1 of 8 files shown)


wear/src/main/java/io/homeassistant/companion/android/onboarding/OnboardingActivity.kt
```diff
@@ -87,6 +98,28 @@ class OnboardingActivity : AppCompatActivity(), OnboardingView {
...
+                null // a Wear device only has one companion device so this is not needed
+            )
+            val confirmation = Intent(this, ConfirmationActivity::class.java).apply {
+                putExtra(ConfirmationActivity.EXTRA_ANIMATION_TYPE, ConfirmationActivity.OPEN_ON_PHONE_ANIMATION)
+                putExtra(ConfirmationActivity.EXTRA_ANIMATION_DURATION_MILLIS, 2500)
+                putExtra(ConfirmationActivity.EXTRA_MESSAGE, getString(commonR.string.continue_on_phone))
+            }
+            startActivity(confirmation)
+        } catch (e: Exception) {
+            Log.e(TAG, "Unable to open sign in activity on phone")
```
dshokouhi, line 118: We should make sure to print the exception in the log

JBassett (approved):

PR #2165, opened 2022-01-19
Move Wear OS dependency to full implementation

### LINKED ISSUE #2153: Minimal flavor contains gms
2022.1.1

**Description of problem:**
The minimal flavor contains google mobile services (gms), in recent versions, it probably comes from somewhere else because I did not find it in the code, you can see more [here](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/10283)

dshokouhi:
## Summary

Another attempt to fix #2153 this time I am a bit more confident. After running a gradle check on the dependencies I came to notice that our remote interactions dependency does indeed use GMS. Which makes sense because the watch communicates with the phone using googles servers for relaying data back and forth. More details can be found in https://github.com/home-assistant/android/issues/2153#issuecomment-1016813714

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -163,8 +163,6 @@ dependencies {
     implementation("androidx.navigation:navigation-ui-ktx:2.3.5")
     implementation("com.google.android.material:material:1.4.0")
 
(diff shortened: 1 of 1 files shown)


dshokouhi:
does not fix the actual issue so closing

PR #2166, opened 2022-01-19
Websocket Notification Bug Fixes

JBassett:
## Summary

Fix a bunch of small issues that were found with Websocket notifications.

* Race condition in Websocket send
* Add source of notification (FCM vs Websocket)
* Fix missing pings when we should be ensuring a Websocket connection
* Move Websocket setting into other since it's not technically only about notifications.


dshokouhi (approved):
Tested debug APK, I do indeed see the ping pong messages now

jpelgrom:
>Move Websocket setting into other since it's not technically only about notifications.

While technically it might be true right now it is only about notifications? Only other websocket usage is in services managed by the system so those shouldn't care whether or not this service is running.

JBassett:
> While technically it might be true right now for users it is only about notifications? Only other websocket usage is in services managed by the system so those shouldn't care whether or not this service is running.

Right now the system could technically kill the system we use to update widgets if not for this setting.

PR #2170, opened 2022-01-20
Fix minimal including Play Services dependencies

### LINKED ISSUE #2153: Minimal flavor contains gms

jpelgrom:
## Summary

Yet another attempt to fix #2153.
According to Gradle, two dependencies were still adding Google libraries to the app (see [this comment](https://github.com/home-assistant/android/issues/2153#issuecomment-1016813714)). This PR addresses that by:
- embedding ExoPlayer Cronet for minimal builds + excluding the Play Services dependency from the default (see [this comment](https://github.com/home-assistant/android/issues/2153#issuecomment-1016933008) for details)*
- moving the Wear OS remote interactions to full, already only used in the full app

Checked the merged manifest and classes in the generated APK and no `com.google.android.gms` to be found!

\* Note that this will increase the app download size for minimal and quest by ~8MB. Because this doesn't require code changes and should ensure the same behavior for all versions I think it is worth the increased size.

## Screenshots

n/a

n/a

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -163,8 +163,6 @@ dependencies {
     implementation("androidx.navigation:navigation-ui-ktx:2.3.5")
     implementation("com.google.android.material:material:1.4.0")
 
-    implementation("androidx.wear:wear-remote-interactions:1.0.0")
-
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin:2.13.1")
     implementation("com.squareup.okhttp3:okhttp:4.9.3")
     implementation("com.squareup.picasso:picasso:2.8")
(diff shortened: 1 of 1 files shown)


dshokouhi:
Interesting the file size jump for the `quest` build is much higher than the `minimal` build. I wonder why, maybe because of the API levels required for `quest`?

before:
image

after:
image

dshokouhi (approved):
Not sure what's up with the file size for `quest` but we are using the recommended path and the scan results show no play service dependencies :) Thanks for looking into this!

jpelgrom:
I actually didn't build the Quest APK locally becuase I assumed it would be the same but the native libraries are bigger somehow? Quest has a higher minimum APK so if anything I would expect it to get smaller. It does show the same download size, could it be duplicating files inside the .so libs somehow?

Minimal:
image

Quest:
image

Everything else is almost the same size according to Android Studio's analyzer.

JBassett (approved):

PR #2175, opened 2022-01-21
Add helpers for duplicate Wear Compose code

NotWoods:
## Summary

Adds helpers for duplicate Wear Compose code. Lots of layout-related code is duped in every Wear OS screen, and the goal here is to make it easier to keep things consistent.

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/home/views/ExpandableListHeader.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/views/ExpandableListHeader.kt
@@ -1,0 +1,72 @@
+package io.homeassistant.companion.android.home.views
+
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Row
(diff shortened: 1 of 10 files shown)


wear/src/main/java/io/homeassistant/companion/android/home/views/ExpandableListHeader.kt
```diff
@@ -0,0 +1,72 @@
...
+    string: String,
+    expanded: Boolean,
+    onExpandChanged: (Boolean) -> Unit
+) {
+    androidx.wear.compose.material.ListHeader(
+        modifier = Modifier
+            .clickable { onExpandChanged(!expanded) }
+    ) {
+        Row {
+            val plusMinus = if (expanded) "+" else "-"
```
dshokouhi, line 43: This should be `!expanded` in the condition because the `+` sign is meant to indicate the header can be expanded.

JBassett (approved):

PR #2189, opened 2022-01-24
Migrate manage widget settings page to compose

dshokouhi:
## Summary

Just migrating the page to compose so we are no longer on preferences.  Kept the design mostly the same except there is no summary anymore. I couldnt think of a good design for that so decided to stick with `TextButton` for now.

## Screenshots

image


app/src/main/java/io/homeassistant/companion/android/settings/widgets/views/ManageWidgetsView.kt
```diff
@@ -0,0 +1,102 @@
...
+        modifier = Modifier.padding(10.dp)
+    ) {
+        if (viewModel.buttonWidgetList.isNullOrEmpty() && viewModel.staticWidgetList.isNullOrEmpty() &&
+            viewModel.mediaWidgetList.isNullOrEmpty() && viewModel.templateWidgetList.isNullOrEmpty()
+        ) {
+            item { Text(stringResource(id = R.string.no_widgets)) }
+        }
+        if (!viewModel.buttonWidgetList.isNullOrEmpty()) {
+            item {
+                Text(stringResource(id = R.string.list_button_widgets))
```
jpelgrom, line 36: I know this hasn't changed, but the "List of" part of the string feels unnecessary and using "Button widgets", "Entity state widgets", etc as headings also work.

jpelgrom (commented):
I am not sure using regular text for headings and `TextButton`s for each widget is the right way forward, especially when you lots of widgets this de-emphasizes the headings and the buttons have a minimum width + padding which can break up the list*. Using the previous style works well when removing the subtitle, and is a pattern you frequently see in for example the settings app.
image

\* For example in your screenshot, the distance between the entity state widgets buttons is larger than the distance between a button and a heading. The text for the "test" widget is indented because of the minimum width. "alarm test" is an entity state widget but visually it is closer to the media player widgets.

dshokouhi:
@jpelgrom  thank you for calling out the padding that was bugging me and I couldnt figure it out and thought it was just an anomaly lol.  As of now I opted for a button to fix the alignment issue and to keep some consistency with other settings screen that take the user to another screen.

JBassett (changes requested):
Just a tiny thing about the `init`.  You should just use the built in one, not create your own.

JBassett (approved):

PR #2190, opened 2022-01-24
Migrate manage tiles screen to compose

dshokouhi:
## Summary

Migrates the manage tiles setting page to compose.

## Screenshots

image


app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
```diff
@@ -0,0 +1,144 @@
...
+                    }
+                }) {
+                    Text(tileName)
+                }
+            }
+        }
+        Text(
+            text = stringResource(id = R.string.tile_label),
+            modifier = Modifier.padding(start = 10.dp, bottom = 10.dp)
+        )
```
jpelgrom, line 85: Suggestion: move this (and the one for the subtitle) to a label for the `TextField`
image

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
```diff
@@ -0,0 +1,144 @@
...
+        TextField(
+            value = viewModel.tileLabel.value.toString(),
```
jpelgrom, line 87: This is a `String?` and converting a null value with `toString()` will result in a string "null", I think this should be changed to just return "" if the value is null (also for subtitle)

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
```diff
@@ -0,0 +1,144 @@
...
+            for (item in sortedEntities) {
+                DropdownMenuItem(onClick = {
+                    viewModel.selectedEntityId.value = item.entityId
+                    expandedEntity = false
+                }) {
+                    Text(text = item.entityId, fontSize = 15.sp)
+                }
+            }
+        }
+        TextButton(onClick = {
```
jpelgrom, line 129: Is this button really necessary or could the app also update the database (and push it to the system) when any of the input fields change?
If it is necessary: this button should be emphasized to make sure the user actually taps on it and doesn't quit without saving. Suggestion: use a regular, full-width `Button`.

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
```diff
@@ -0,0 +1,144 @@
...
+                text = stringResource(id = R.string.tile_icon),
+                fontSize = 15.sp,
+                modifier = Modifier.padding(end = 10.dp)
+            )
+            viewModel.icon.value?.toBitmap()?.asImageBitmap()
+                ?.let { Image(it, contentDescription = stringResource(id = R.string.tile_icon)) }
+        }
+        TextButton(onClick = { expandedEntity = true }) {
+            Text(
+                text = stringResource(id = R.string.tile_entity) + ": ${viewModel.selectedEntityId.value}",
```
jpelgrom, line 114: If no entity has been selected this will show up as "Select a entity to toggle or call: null`, it should probably be changed to show "not set" or something similar if the value is null

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
```diff
@@ -0,0 +1,144 @@
...
+fun ManageTilesView(
+    viewModel: ManageTilesViewModel,
+    iconDialog: IconDialog,
+    childFragment: FragmentManager
+) {
+    val context = LocalContext.current
+    var expandedTile by remember { mutableStateOf(false) }
+    var expandedEntity by remember { mutableStateOf(false) }
+    Column(modifier = Modifier.padding(20.dp)) {
+        TextButton(onClick = { expandedTile = true }) {
```
jpelgrom, line 49: I don't think making everything a large button is a good idea because 60% (or more, depending on how you calculate it) of the screen is now a button. Suggestion: use text with an outlined button like for the Wear OS template tile refresh interval setting on a phone (also for icon/entity selection):
image

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
```diff
@@ -0,0 +1,144 @@
...
+
+        TextButton(onClick = {
+            iconDialog.show(childFragment, viewModel.selectedTile.value)
+        }) {
+            Text(
+                text = stringResource(id = R.string.tile_icon),
+                fontSize = 15.sp,
+                modifier = Modifier.padding(end = 10.dp)
+            )
+            viewModel.icon.value?.toBitmap()?.asImageBitmap()
```
jpelgrom, line 109: A color filter should be added because in light mode the default icon is white

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
```diff
@@ -0,0 +1,144 @@
...
+                            icon.setColorFilter(context.resources.getColor(io.homeassistant.companion.android.R.color.colorAccent), PorterDuff.Mode.SRC_IN)
+                            viewModel.icon.value = icon
+                        }
+                    }
+                }) {
+                    Text(tileName)
+                }
+            }
+        }
+        Text(
```
jpelgrom, line 82: It might be nice to add a `Divider` here to visually separate the tile selection (general) from the individual settings (specific)

jpelgrom (commented):
A few suggestions, some are related to the design (feel free to ignore them if you don't agree) but others are about displaying raw data (in most cases, a regression from the old UI)

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
dshokouhi: So this button was designed to ensure the user is actually changing the tile they meant to change. If we make it autoupdate when any field is changed it has the potential of overwriting data a user did not expect. I will make this a button to make it more emphasized :)

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
dshokouhi: actually a user is required to supply a value so I will opt to show a blank string, we can't add a tile without a label and entity ID at a minimum and clearly I missed this during the migration 😬 I will also update the save button to reflect this.

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
dshokouhi: Just submitted a change and I thought it would apply here but for some reason its not applying to the default icon. Not sure what I did wrong here :)

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
jpelgrom: On the row, [add `verticalAlignment = Alignment.CenterVertically`](https://github.com/home-assistant/android/blob/b7f6a7322cfc156f2854513bb5fa32d7fdbe89af/app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTile.kt#L80)

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
jpelgrom: It looks like you only apply the tint to the `selectedIcon` but the default icon is the default value for `drawableIcon` (which is never updated because `selectedIcon` is not set).  You'd have to move the filter to always tint the drawable icon in the view model, or do it in Compose on the `Image`: `colorFilter = ColorFilter.tint(colorResource(R.color.colorIcon))`.

dshokouhi: much cleaner to do it in compose :) thanks for the help!!!

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
```diff
@@ -0,0 +1,161 @@
...
+                modifier = Modifier.padding(end = 10.dp)
+            )
+            OutlinedButton(onClick = { expandedTile = true }) {
+                Text(
+                    viewModel.selectedTileName.value
+                )
+            }
+        }
+
+        DropdownMenu(expanded = expandedTile, onDismissRequest = { expandedTile = false }) {
```
jpelgrom, line 62: Suggestion: wrap the `OutlinedButton` for selecting the tile with a `Box`, and move the dropdown menu inside there, because now the dropdown doesn't align with the button. See [Wear OS template tile settings as an example](https://github.com/home-assistant/android/blob/b7f6a7322cfc156f2854513bb5fa32d7fdbe89af/app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTile.kt#L92-L113).
image

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        Divider()
+        TextField(
+            value = viewModel.tileLabel.value ?: "",
+            onValueChange = { viewModel.tileLabel.value = it },
+            label = {
+                Text(
+                    text = stringResource(id = R.string.tile_label),
+                    modifier = Modifier.padding(start = 10.dp, bottom = 10.dp)
```
jpelgrom, line 85: While the padding at the start looks fine if there is no text, when text has been added it is indented and looks out of place. The bottom padding is also useless/ignored by Compose.
image
(also applies to subtitle input)

app/src/main/java/io/homeassistant/companion/android/settings/qs/ManageTilesFragment.kt
```diff
@@ -66,143 +81,17 @@ class ManageTilesFragment constructor(
...
-            )
-            tileDao.add(tileData)
-            tileList = tileDao.getAll()
-            Toast.makeText(requireContext(), commonR.string.tile_updated, Toast.LENGTH_SHORT).show()
-            return@setOnPreferenceClickListener true
+    private fun reloadIcon() {
+        val iconDrawable = viewModel.selectedIcon.value?.let { iconPack.getIcon(it)?.drawable }
+        if (iconDrawable != null) {
+            val icon = DrawableCompat.wrap(iconDrawable)
+            icon.setColorFilter(resources.getColor(R.color.colorAccent), PorterDuff.Mode.SRC_IN)
```
jpelgrom, line 93: No longer required because you're doing it in Compose now (2a982aca97dd4a7bdfe5f541ab4f9dd05f87874d)

app/src/main/java/io/homeassistant/companion/android/settings/qs/ManageTilesViewModel.kt
```diff
@@ -0,0 +1,91 @@
...
+                }
+            }
+        }
+    }
+
+    fun updateExistingTileFields() {
+        tileLabel.value = currentTile()!!.label
+        tileSubtitle.value = currentTile()!!.subtitle
+        selectedEntityId.value = currentTile()!!.entityId
+        selectedIcon.value = currentTile()!!.iconId
```
JBassett, line 80: We should only pull the value once out of the DB.
If the current tile could be null we should default the values to something.

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/settings/qs/ManageTilesViewModel.kt
dshokouhi: We actually only call this once when the user selects a tile and only if the current tile is not null :)

https://github.com/home-assistant/android/blob/1868790b53404d419b525c024d1136eb79a4dfcb/app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt#L69-L74

JBassett: `currentTile()` is going to pull it from the database 4 times in a row.  I'm saying create a variable for it rather than extract it from the database 4 times in a row.

dshokouhi: 🤦‍♂️ got it lol

JBassett (approved):

Issue #1162, opened 2020-11-07
Show a breaking changes prompt or notification upon updating, if needed


dshokouhi:
**Is your feature request related to a problem? Please describe.**
We should try to better inform users about breaking changes so they know what is expected of them.  If possible, show the whole change log for the current version but breaking changes should be highlighted at the top.  Although we have a blog post detailing this, not all users see it so its a better experience to inform them some how.  We can even consider a notification as not all users will open the app after updating.

**Describe the solution you'd like**
As PR's come in we should mark them as a breaking change (if required) so we can identify them easily.  We should find a good way to prompt users about this.  I think a notification might make the most sense, but an in app prompt can also work to ensure they have seen it.  We should also include the blog post link in the play store release notes.

**Describe alternatives you've considered, if any**
The alternative is to expect all users will read either github change log or the blog post.

**Additional context**
I see users having issues in the forums, reddit and probably eventually on discord.  The better we can handle the situation the less upset people can be.

dshokouhi:
This looks like it might be a viable library to use: https://github.com/AppDevNext/ChangeLog might be able to use a github action to pull in the changelog on each commit :)

jrial:
Still an issue.

PR #2195, opened 2022-01-25
Add changelog dialog to show when the app is updated

dshokouhi:
## Summary

Fixes: #1162 

Uses a library to help show a changelog prompt that is themed for day and night. The dialog will show up after the app is updated and can be accessed in settings as well.

As the github changelog won't be helpful to all users its probably in our best interest to decide on what to add in the changelog. There is some room for HTML like for emphasizing a breaking change. Probably a good time to discuss the format but I wanted to get this out so users don't have surprises like we saw with the new persistent notification.

During beta releases we should keep the version as `Beta` and only update it to the release version when we are ready to cut a release. This will allow us to keep the changelog for the release growing while we remain in beta. Again some of this we can decide to change later if we feel we need to based on user feedback and anything else we find during discussions.

## Screenshots

image

image


JBassett (approved):

jrial:
Does not fully resolve the problem. It's a step in the good direction, but I've already had to rescue systems that were no longer accessible over the network, and therefore in no position to be showing me any popups informing me of what I did wrong. Breaking changes, as the name implies, break things. You want to be forewarned of this, not (maybe) learn about it after the fact. Would be better if this popup would be triggered when you click the "update" button, and not just be shown at the end of the process, when things are already in a broken state.

dshokouhi:
> Does not fully resolve the problem. It's a step in the good direction, but I've already had to rescue systems that were no longer accessible over the network, and therefore in no position to be showing me any popups informing me of what I did wrong. Breaking changes, as the name implies, break things. You want to be forewarned of this, not (maybe) learn about it after the fact. Would be better if this popup would be triggered when you click the "update" button, and not just be shown at the end of the process, when things are already in a broken state.

I think you might be confusing the Android app for an actual HA core update. There is nothing in the Google play store to cause such a prompt for an update. The breaking change in this case was to notify users of a new persistent notification. This PR directly solved the linked issue. Also in the future please don't comment on merged PRs instead create a new issue.

PR #2201, opened 2022-01-26
Migrate Manage Shortcuts settings screen to compose

dshokouhi:
## Summary

Migrates the existing screen to a compose format changing the design slightly to match the rest of the compose rewrites we have done. This also fixes a minor issue in the current design where the icon was not being restored. Going forward any shortcut updated will also retain its existing icon as we are storing the icon ID as an intent extra. We are unable to do so for any existing shortcut.

## Screenshots

image

image


app/src/main/java/io/homeassistant/companion/android/settings/shortcuts/views/ManageShortcutsView.kt
```diff
@@ -0,0 +1,447 @@
...
+                        4 -> viewModel.deleteShortcut4.value = true
+                        5 -> viewModel.deleteShortcut5.value = true
+                    }
+                }
+            }
+            Button(
+                onClick = {
+                    when (index) {
+                        1 -> {
+                            if (viewModel.deleteShortcut1.value)
```
jpelgrom, line 327: Why is this only shown for shortcut 1, and not for all shortcuts?

app/src/main/java/io/homeassistant/companion/android/settings/shortcuts/views/ManageShortcutsView.kt
```diff
@@ -0,0 +1,447 @@
...
+fun ManageShortcutsView(
+    viewModel: ManageShortcutsViewModel,
+    iconDialog: IconDialog,
+    childFragment: FragmentManager
+) {
+    val context = LocalContext.current
+    var expandedEntity by remember { mutableStateOf(false) }
+    var expandedPinnedShortcuts by remember { mutableStateOf(false) }
+
+    LazyColumn(modifier = Modifier.padding(20.dp)) {
```
jpelgrom, line 52: Because this is a long screen, I actually noticed that the padding is applied to the `LazyColumn` and as a margin around the content, result: the content doesn't go all the way to the top or bottom of the screen when scrolling. Suggestion: use `contentPadding` instead.

jpelgrom (commented):
You say that the design has been changed slightly to match other Compose rewrites, maybe it is also a good idea to switch to a "Select a shortcut to edit: [x]" design which is used for tiles? That would give us the option to show configured pinned shortcuts in the future (dropdown can be any length).

While using 6 variables and multiple if/when statements works, I wonder if there is a better way to do this now that the app is no longer limited by a PreferenceScreen. Did you consider a list with a data class?

app/src/main/java/io/homeassistant/companion/android/settings/shortcuts/views/ManageShortcutsView.kt
dshokouhi: totally missed this one 😅

dshokouhi:
> You say that the design has been changed slightly to match other Compose rewrites, maybe it is also a good idea to switch to a "Select a shortcut to edit: [x]" design which is used for tiles? That would give us the option to show configured pinned shortcuts in the future (dropdown can be any length).

This is actually the design for the actual pinned shortcuts down below. Dynamic shortcuts are a bit odd because some launchers simply won't show them all so its up to the user to decide which ones to use and maybe confusing if we group them together as the behavior is not the same.



> While using 6 variables and multiple if/when statements works, I wonder if there is a better way to do this now that the app is no longer limited by a PreferenceScreen. Did you consider a list with a data class?

I was stuck thinking about this for a while. Just looked into using a data class and maybe I'm misunderstanding something but it feels like it would overcomplicate things.  I think that if we used the same form for dynamic and pinned it would make sense however one has a hard limit of 5 with a soft limit imposed by the launcher while the other has no limits as long as the launcher supports it.  Personally speaking I only use pinned shortcuts, at the time when we added shortcuts we opted to use Dynamic instead of Static because static didn't make sense. Some people just like having the shortcuts under the long press action on the icon.

app/src/main/java/io/homeassistant/companion/android/settings/shortcuts/views/ManageShortcutsView.kt
```diff
@@ -0,0 +1,456 @@
...
+                            stringResource(id = R.string.shortcut_pinned_desc)
+                    )
+                },
+                modifier = Modifier.padding(top = 16.dp, bottom = 16.dp)
+            )
+
+            Text(stringResource(id = R.string.shortcut_type))
+
+            Row {
+                ShortcutRadioButtonRow(viewModel = viewModel, type = "lovelace", index = index)
```
jpelgrom, line 237: These buttons don't work for the pinned shortcuts (because the index is 6)

dshokouhi: 🤦‍♂️  thank you 🙏

jpelgrom:
> This is actually the design for the actual pinned shortcuts down below.

When I tested it earlier it didn't actually give me the option to pin the shortcuts to the home screen but instead added them as a dynamic shortcut (where you have to long press on the icon). It does work correctly now, so maybe I just used the wrong field each time.

After adding the shortcut the dropdown doesn't immediately show up or update with the new shortcut, we should probably reload the list when the shortcut is added / fragment is resumed to make sure it is up-to-date.

> Dynamic shortcuts are a bit odd because some launchers simply won't show them all so its up to the user to decide which ones to use and maybe confusing if we group them together as the behavior is not the same.

Hmm yes the behavior is different but in the end they have the same options (the pinned shortcut ID could be removed and automatically generated, and for selecting the shortcut show the description which is also shown in the launcher instead of the ID). Adding a new option that can only be edited when creating a new item, "Show on long press on app icon", could merge both in to one list but that would require additional logic to handle the limit of 5 and a notice that the launcher may not support it. Out of scope for this PR :)

Irregardless of the idea in the previous sentence, I still think we could group the dynamic shortcuts together and default to showing shortcut 1. That will make the list shorter so you don't have to scroll so far when you want to add/edit a pinned shortcut.

> I was stuck thinking about this for a while. Just looked into using a data class and maybe I'm misunderstanding something but it feels like it would overcomplicate things.

After thinking about it I agree that with this setup it would overcomplicate things. If you decided to switch to a dropdown to select a shortcut to edit, it would make a lot of sense though: add a new variable in the viewmodel to store which shortcut is being edited and a list of possible options, and when it changes update the shortcutIcon/Label/Desc/Path/Type variables with the new values.

dshokouhi:
> It does work correctly now, so maybe I just used the wrong field each time.

Yea thats also the reason for the note in the pinned shortcut area. When I first was testing it I kept forgetting to the change the ID :)



> After adding the shortcut the dropdown doesn't immediately show up or update with the new shortcut, we should probably reload the list when the shortcut is added / fragment is resumed to make sure it is up-to-date.

Good call will add an update in `onResume` as adding a pinned shortcut always takes the user to their home screen. I have similar logic for the dynamic shortcuts as they get added/deleted but was missing this one


> Irregardless of the idea above, I still think we should group the dynamic shortcuts together and default to showing shortcut 1. That will make the list shorter and pinned shortcuts easier to find (users might not scroll to the bottom if they see shortcut 2, shortcut 3, shortcut 4, ..., and if you do know about pinned shortcuts and simply want to edit/create one it is still unnecessary scrolling).

Yea there was a lot going on already (obviously from the stuff you already caught 😁 ) so when I was writing it I thought to keep the design the same (for now) but to also fix the one quirk regarding the icon not loading. Wanted to get some discussion going on what a proper layout would look like.  Personally I would opt to move the pinned shortcuts to the top but because they don't have as wide support as dynamic I decided to keep dynamic there.  I like the drop down idea to remove the amount of items. We could also play around with having tabs at the bottom to switch between dynamic and pinned too. As we keep discussing more and more ideas start to come. The nice thing about switching to compose is that it'll be a bit easier to migrate things to a different layout :)



> After thinking about it I agree that with this setup it would overcomplicate things. If you decided to switch to a dropdown to select a shortcut to edit, it would make a lot of sense though: add a new variable in the viewmodel to store which shortcut is being edited and a list of possible options, and when it changes update the shortcutIcon/Label/Desc/Path/Type variables with the new values. No data classes required if dynamic/pinned is kept separated.

Yea there are definitely improvements we could add here. I think another big improvement would be to get a list of panels from the websocket API and use that for a drop down to select a panel. However now that I think about it I am not sure if it includes a `view`.  My original idea to keep the shortcuts numbered like they are today is because its easier for a user to add them one by one instead of them starting at shortcut 5 and working up and wondering why the first one listed doesn't show, assuming they use the drop down out of order.  The hard limit on them does make it weird which is why I favor pinned over dynamic.

jpelgrom:
> Personally I would opt to move the pinned shortcuts to the top but because they don't have as wide support as dynamic I decided to keep dynamic there.

Well if pinned shortcuts aren't supported they aren't shown at all, right? In that case moving them to the top when available is still possible.

> My original idea to keep the shortcuts numbered like they are today is because its easier for a user to add them one by one instead of them starting at shortcut 5 and working up and wondering why the first one listed doesn't show, assuming they use the drop down out of order. 

Right so if we wanted to create an abstraction for the user where both types are merged into one list, the app would also have to manage the dynamic shortcut order to move items to position [n] matching what is shown in the app.
If the app keeps dynamic and pinned separated, but uses a dropdown for selecting and editing both, it can still number dynamic shortcuts and show names for pinned shortcuts.

app/src/main/java/io/homeassistant/companion/android/settings/shortcuts/views/ManageShortcutsView.kt
```diff
@@ -0,0 +1,458 @@
...
+            )
+            Divider()
+        }
+
+        val shortcutCount = if (viewModel.canPinShortcuts)
+            ManageShortcutsSettingsFragment.MAX_SHORTCUTS + 1
+        else
+            ManageShortcutsSettingsFragment.MAX_SHORTCUTS
+
+        items(shortcutCount) { i ->
```
JBassett, line 68: There should be a function for creating the shortcut view itself rather than one giant function.

app/src/main/java/io/homeassistant/companion/android/settings/shortcuts/ManageShortcutsViewModel.kt
```diff
@@ -0,0 +1,314 @@
...
+    var shortcutDesc1 = mutableStateOf("")
+        private set
+    var shortcutPath1 = mutableStateOf("")
+        private set
+    var shortcutType1 = mutableStateOf("lovelace")
+        private set
+    var drawableIcon1 = mutableStateOf(AppCompatResources.getDrawable(application, R.drawable.ic_stat_ic_notification_blue))
+        private set
+    var deleteShortcut1 = mutableStateOf(false)
+        private set
```
JBassett, line 64: This should be abstracted to a Data class and use a mutable list of those data objects rather than manually iterative over them.

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/settings/shortcuts/ManageShortcutsViewModel.kt
dshokouhi: @jpelgrom you were right about this and it was my own misunderstanding :) 

Thank you both for the help 🙏

JBassett (approved):

PR #2216, opened 2022-01-30
Breaking change: hide non-primary entities on Wear home screen

jpelgrom:
## Summary

Right now the Wear OS app is usable, but the current approach can still create long lists on a small screen if you have a lot of entities and/or when we start supporting more domains in the future. Thinking about what you'd use the app for and how we matched the autogenerated Lovelace dashboard in showing areas first and then domains for usability, I believe we should hide configuration and diagnostic (non-primary) entities on the home screen from the main areas and domains lists, which [the autogenerated dashboard also started doing in 2021.11](https://www.home-assistant.io/blog/2021/11/03/release-202111/#entity-categorization).

This PR filters the entities shown in 'Areas' and 'More entities' on the home screen to only show primary entities. All types of entities, primary and non-primary, will remain available using the 'All entities' button and anywhere else in the app, so you can still use it on the shortcut tile and even add it as a favorite to the app's home screen.

Because this might hide entities users are currently seeing on the app's home screen, this is a **breaking change**.

**Something I would like input on**: should there be an option in the app to show non-primary entities on the home screen (current behavior)? Lovelace doesn't offer such an option but you can create your own dashboard with everything you want, whereas you can't really change the app's home screen aside from favorites. 

## Screenshots

This will depend on the instance, but as an example, in the current app I have two configuration switches exposed by a motion sensor:
image image

As seen in HA's frontend:
image

After this change, the switches domain will be removed for me because I don't have any primary area-less switches:
image

The switches are still available in 'All entities':
image

Documentation: home-assistant/companion.home-assistant#673

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/entities/EntityRegistryResponse.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/entities/EntityRegistryResponse.kt
@@ -3,5 +3,6 @@ package io.homeassistant.companion.android.common.data.websocket.impl.entities
 data class EntityRegistryResponse(
     val areaId: String?,
     val deviceId: String?,
+    val entityCategory: String?,
     val entityId: String
 )
--- a/common/src/main/java/io/homeassistant/companion/android/util/RegistriesDataHandler.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/util/RegistriesDataHandler.kt
@@ -26,4 +26,11 @@ object RegistriesDataHandler {
         }
         return null
     }
+
+    fun getCategoryForEntity(
+        entityId: String,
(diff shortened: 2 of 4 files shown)


tripplehay:
Sounds like a great idea to me

tripplehay (approved):
Seems fine to me

dshokouhi:
> **Something I would like input on**: should there be an option in the app to show non-primary entities on the home screen (current behavior)? Lovelace doesn't offer such an option but you can create your own dashboard with everything you want, whereas you can't really change the app's home screen aside from favorites.

Personally I don't think we really need an option because if a user has a need to get to it quickly they can just add it as a favorite. Although we cant fully change the home screen being able to add favorites and rearrange them should suffice for most use cases.

JBassett (approved):

PR #2240, opened 2022-02-04
Breaking change: Switch to android API to determine state of DND

dshokouhi:
## Summary

This PR is a breaking change as we are going to switch from the undocumented variable and use the actual android API so we can use the standard practice. This android API is only available on Android 6+ devices.

Documentation: home-assistant/companion.home-assistant#678


JBassett (approved):

dshokouhi:
> For users on Android 5 who have previously enabled the sensor this will permanently throw an exception when the sensor worker runs, because the sensor is still turned on, the API isn't available on their device, and the sensor settings list will hide the sensor because `hasSensor` returns false.

Good call I will add a check to `requestSensorUpdates` to prevent it from checking. I don't think its a good practice to update the DB and disable it in these cases.

PR #2252, opened 2022-02-09
Add PinchToZoom option

micronen:
Allow enabling PinchToZoom in WebView via option in preferences

## Summary

Allow enabling PinchToZoom in WebView via option in preferences.
It allows, for e.g., using a picture ui with a lot of entities on a small screen phone. 
This was discused is several threads: 
https://community.home-assistant.io/t/ha-web-interface-enable-pinch-to-zoom-on-mobile-devices/30389
https://community.home-assistant.io/t/pinch-zoom-on-mobile-idea/104153
https://community.home-assistant.io/t/zoom-inn-android-app/353758
and maybe some more...

## Screenshots

Screenshot_20220209-194500_Home Assistant
Screenshot_20220209-194515_Home Assistant
Screenshot_20220209-194530_Home Assistant
Screenshot_20220209-194102_Home Assistant
Screenshot_20220209-193945_Home Assistant
Screenshot_20220209-194022_Home Assistant

Documentation: home-assistant/companion.home-assistant#683

## Any other notes

This is my 1st try at kotlin and/or android app. Please give me feedback if I made some rooky mistakes.


app/src/main/res/xml/preferences.xml
```diff
@@ -80,6 +80,11 @@
             android:icon="@drawable/ic_phone_check"
             android:title="@string/keep_screen_on"
             android:summary="@string/keep_screen_on_def"/>
+        <SwitchPreference
+            android:key="pinch_to_zoom"
+            android:icon="@drawable/ic_phone_check"
```
dshokouhi, line 85: Minor comment but maybe a more appropriate icon so it can look different from the option directly above it?

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -657,6 +666,15 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
             binding.blurView.setBlurEnabled(false)
         }
 
+        // Enable pinch to zoom
+        if (presenter.isPinchToZoomEnabled()) {
+            webView.getSettings().setBuiltInZoomControls(true)
+            webView.getSettings().setDisplayZoomControls(false)
```
jpelgrom, line 672: Because this is `false` for both settings, perhaps it would be better to move this to the setup block (inside `webView.apply {}` with the other settings)

micronen: OK. removed redundant code

dshokouhi:
So I am looking at the debug APK but I don't see where this setting is working. Does this only work on certain cards? I have a zwave map but thats a custom card and pinch to zoom on it works in the production version of the app so I don't think it has any impact there.  Maybe I am missing something here?

micronen:
On my phone it works for the whole app, menus and everything. You turn it on and can pinch anything.
I have pinch enabled for browsers. Maybe it is a requirement for it to work in the app?
Can you please, try the following from https://community.home-assistant.io/t/ha-web-interface-enable-pinch-to-zoom-on-mobile-devices/30389/17`

Put this in a file called zoom-enable.js in your config/www folder

document.querySelector("meta[name=viewport]").setAttribute("content", "width=device-width, viewport-fit=cover");
Add this to your configuration.yaml:

frontend:
  extra_module_url:
    - /local/zoom-enable.js

dshokouhi:
> > If this is required for this feature then we shouldn't move forward with it in the app
> 
> I agree, but: iOS has an option to enable pinch to zoom and will adjust the viewport meta tag on load to make it work when turned on, see home-assistant/iOS#1472. (That PR also links to frontend issues about enabling/disabling zoom by default.) It is a bit of a hack but personally I think that if it was implemented in the same way in this PR, it should be acceptable.

I agree having the app perform it is more desirable in this situation. I think it's ok to match iOS behavior here.

micronen:
I've added a JS query to override the content attribute in onPageFinished

```
webView.evaluateJavascript(
    "document.querySelector(\"meta[name=viewport]\").setAttribute(\"content\", \"width=device-width,user-scalable=yes,viewport-fit=cover,initial-scale=1\");"
) {}
```
Please try it again.

BTW, in the chrome browser you can use Settings -> Accessibility -> Force enable zoom, to make it disregard user-scalable=no

BTW2, why HASS even have this set to no in the first place?

SkechyWolf:
> Please try it again.
 Works now for me

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -265,6 +266,25 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                 }
 
                 override fun onPageFinished(view: WebView?, url: String?) {
+                    // Use idea from https://github.com/home-assistant/iOS/pull/1472 to filter viewport
```
dshokouhi, line 269: Should we also reset the values like iOS does when the feature is turned off?

dshokouhi:
> Please try it again.

latest commit works for me too, thank you :)



> BTW2, why HASS even have this set to no in the first place?

I think this PR will help answer this question :) https://github.com/home-assistant/frontend/pull/8353

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
micronen: Yes I think we should.
I'll work on it tonight

micronen: Please review & check latest version

micronen:
Please review & check latest version

dshokouhi (approved):
Tested debug APK and it works for me

JBassett (approved):

PR #2257, opened 2022-02-10
Fix websocket invalid auth race condition

jpelgrom:
## Summary

If the authentication that is provided to the websocket is invalid, Home Assistant will immediately close the websocket after sending the `auth_invalid` message. Because the `onClosing` function will reset the Job used for monitoring the authentication result, this can result in a race condition where the job is changed before the result is processed. As a result, the job won't complete and checking authentication will have to wait for the timeout to complete. Example log:

```
2022-02-10 21:55:44.561 10649-10682/io.homeassistant.companion.android.debug D/WebSocketRepository: Websocket: onOpen
2022-02-10 21:55:44.564 10649-10682/io.homeassistant.companion.android.debug D/WebSocketRepository: Websocket: onMessage (text)
2022-02-10 21:55:44.870 10649-10682/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number null received: {"type": "auth_required", "ha_version": "2022.2.5"}
2022-02-10 21:55:44.886 10649-10682/io.homeassistant.companion.android.debug D/WebSocketRepository: Websocket: onMessage (text)
2022-02-10 21:55:44.887 10649-10689/io.homeassistant.companion.android.debug D/WebSocketRepository: Auth Requested
2022-02-10 21:55:44.888 10649-10682/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number null received: {"type": "auth_invalid", "message": "Invalid access token or password"}
2022-02-10 21:55:44.893 10649-10682/io.homeassistant.companion.android.debug D/WebSocketRepository: Websocket: onClosing code: 1000, reason: 
2022-02-10 21:55:44.894 10649-10682/io.homeassistant.companion.android.debug D/WebSocketRepository: handleClosingSocket
2022-02-10 21:55:44.896 10649-10689/io.homeassistant.companion.android.debug D/WebSocketRepository: handleAuthComplete: false
```

This PR will use the same mutex that is used for initial auth when closing the connection. This will ensure that any (auth) result is delivered before the connection is closed (if, somehow, no auth result is delivered, the mutex is still unlocked by a timeout).

Additionally, if the auth is invalid that should return false as expected. An `auth_invalid` message calls `completeExceptionally` which cancels the job but that is still `isCompleted` (see ["Job states"](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/index.html)), so a check has been added to make sure it was completed without a cancellation.

## Screenshots

n/a

n/a

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
@@ -9,6 +9,7 @@ import com.fasterxml.jackson.module.kotlin.convertValue
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import io.homeassistant.companion.android.common.data.authentication.AuthenticationRepository
+import io.homeassistant.companion.android.common.data.authentication.AuthorizationException
 import io.homeassistant.companion.android.common.data.integration.ServiceData
 import io.homeassistant.companion.android.common.data.integration.impl.entities.EntityResponse
 import io.homeassistant.companion.android.common.data.url.UrlRepository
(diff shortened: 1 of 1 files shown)


jpelgrom:
After some additional thinking, I've decided to update the code to also exceptionally complete the auth job if we receive `onFailure`. The job won't receive a message to complete the job in that case, so the mutex would only be released by the timeout. It doesn't make sense to wait for the timeout then.

JBassett (approved):

PR #2261, opened 2022-02-11
Improve Wear OS error handling on home, check websocket connection state

jpelgrom:
## Summary

This PR improves the Wear OS' app error handling on the home screen, by showing errors and/or redirecting to onboarding instead of crashing the app on errors. More specifically:
- Handle network/server errors gracefully and show an error + retry button
- The websocket state is monitored and used to require logging in again if authentication permanently fails
- Fixes the app getting stuck on a loading screen if none of the entities on the server are in a supported domain

For more on specific changes, view the comments in individual commits.

## Screenshots

General error:
App showing 'Error loading entities' with a 'Retry' button

If the user has favorites, the error will be shown below the favorites (in place of the loading text/chip) and favorites are still accessible:
App showing a favorite at the top of the screen, and below that 'Error loading entities' with a 'Retry' button

No supported domains message if loading completed but the lists are still emtpy:
App showing an error 'No Supported Entities, Check the documentation for more information'

Settings remains accessible in all cases to allow changing tile settings, logging out, etc.

If you need to login again because authentication failed, the app will immediately send you to onboarding like the phone app.

I don't think this is needed, "Check the documentation" in the no supported domains error refers to the general documentation for Wear OS which lists supported domains.

## Any other notes

If you're going to test this, be aware that because the websocket doesn't immediately throw an error if something fails once the connection is established, the app might send multiple requests before stopping which can trigger maximum number of failed login -> IP ban protections.


JBassett (commented):
Just needs to fix a merge conflict.

JBassett (approved):

PR #2262, opened 2022-02-11
Remove toast when updating state/camera/template widget fails

jpelgrom:
## Summary

Based on [discussion on Discord](https://discord.com/channels/330944238910963714/377667559303938050/941728030685012019), this PR updates the camera, entity state and template widgets to remove the toast when updating the data fails, because these toasts could be very frequent<sup>1</sup> and/or hard to attribute to a specific widget<sup>2</sup>, and show an icon on the specific widget that failed instead.

1: if the user tapped repeatedly to update toasts will 'queue', bad connections can cause problems for template widget updates because they update on any state change
2: entity state/template widgets cache data so they don't have an obvious 'not working' state

## Screenshots

Showing failed updates, when the widget is able to update (again) the icon is not shown.
|Light|Dark|
|----|----|
|Three widgets, each with a red icon of two arrows in a circle for syncing with an exclamation mark in the middle, at the bottom of the widget, indicating that they are not updated, light mode|Three widgets, each with a red icon of two arrows in a circle for syncing with an exclamation mark in the middle, at the bottom of the widget, indicating that they are not updated, dark mode|

Not needed

## Any other notes

CC @dshokouhi, thanks for the input


dshokouhi (approved):
Thanks for taking this on!

jpelgrom:
Decided to change the icon because it is shown on _any_ failure, not just network related ones, and the previous cloud icon suggested otherwise. Screenshots have been updated.

app/src/main/java/io/homeassistant/companion/android/widgets/entity/EntityWidget.kt
```diff
@@ -30,6 +29,8 @@ class EntityWidget : BaseWidgetProvider() {
         internal const val EXTRA_TEXT_SIZE = "EXTRA_TEXT_SIZE"
         internal const val EXTRA_STATE_SEPARATOR = "EXTRA_STATE_SEPARATOR"
         internal const val EXTRA_ATTRIBUTE_SEPARATOR = "EXTRA_ATTRIBUTE_SEPARATOR"
+
+        private var lastResolvedTextSuccess = true
```
JBassett, line 33: This is like making the variable a static variable by putting it in the companion object.  By putting it here all Entity Widgets could show failed even if just one fails.

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/widgets/entity/EntityWidget.kt
jpelgrom: In testing I couldn't get one failure to change any other widgets, but I see how it can theoretically happen. Changed to return the information from the function.

JBassett (approved):

PR #2265, opened 2022-02-12
Clean up old quest flavor code and hide some features from unsupported devices

dshokouhi:
## Summary

Remove the checks for quest build, only use model checks where appropriate. Also hide some features from TV devices as they just won't work and will crash. I know we dont support the TV devices but just hiding things so we dont crash.

* Last app sensor was hidden from quest build but works in the minimal version as we still get the proper permission
* Network sensors do indeed works as we have the permission
* Next alarm will show up but dont think there are any alarms apps yet 😛 
* Notification sensors do work as we have proper permission however I saw some crashes related to the permission on TV devices so hiding this from them because those devices do not have standard notifications and use different media features. This will never work.
* There is no mobile data or sim cards on the quest so hiding that, wont crash on TV
* Settings - Hide from Quest devices as some features will never work
* Do not show notification channels to TV devices as they do not have the proper activity
* Hide websocket info from TV devices as they cannot control the channel


JBassett (changes requested):
I don't think we should have any any explicit `Build.MODEL` checks unless it's causing crashes.  There should be other ways to determine if a device supports an API.

dshokouhi:
> I don't think we should have any any explicit `Build.MODEL` checks unless it's causing crashes. There should be other ways to determine if a device supports an API.

ok I can replace this in a few places but what do you suggest we do for things like Widgets, Tiles and Shortcuts? I hid them in settings based on the model because the devices API level supports it even though the device itself won't.

JBassett (commented):
Needs a merge.

JBassett (approved):

Issue #1892, opened 2021-11-08
Add details pane with entity controls in Wear OS


leroyboerefijn:
**Is your feature request related to a problem? Please describe.**
You cannot adjust brightness, color, fan speed, etc. from the Wear OS app

**Describe the solution you'd like**
When tapping the Chip (not the toggle), a details pane should open that allows for customization of the entity. For lights this could be brightness and color, for fans this could be speed, for covers this could be percentage opened, etc.

**Describe alternatives you've considered, if any**
Can't think of alternatives

**Additional context**

dshokouhi:
So this will work great for existing `ToggleChip` as we can convert it to a `SplitToggleChip` however we may need to do specific domain overrides when it comes to things like `climate` entities. `script` and `scene` may not require this feature since there is nothing additional to set? But then it makes me wonder if `switch` needs it as well? 

hmm maybe a good place to start is what other details should we show on the details screen? Entity history maybe?  We should probably think about what the base details screen will look like so we can add domain specific functionality separately?  Lots of questions great request!

leroyboerefijn:
Indeed, this only applies to entities that do more than just toggle. The sliders/settings on the details pane also need to depend on the abilities of the entity. Next to entity type, not all lights support color for example. Some only support dimming, or color temperature, or whatever other property. So it might not be easy to get this to support all possibilities nicely.

For other details, that is a good question indeed. On one hand, Wear OS is intended for quick actions and glanceable information. On the other hand we want to provide some flexibility to our users. I think it will be a fine balance.

dshokouhi:
> Indeed, this only applies to entities that do more than just toggle. The sliders/settings on the details pane also need to depend on the abilities of the entity. Next to entity type, not all lights support color for example. Some only support dimming, or color temperature, or whatever other property. So it might not be easy to get this to support all possibilities nicely.

So we can actually perform a check in the entity attributes here. If the entity supports it then it will have a `supports_feature` flag in its attributes. We do a similar check in device controls: https://github.com/home-assistant/android/blob/1c748663e4e896e74f9ef4fddb9c45d26e8bea93/app/src/main/java/io/homeassistant/companion/android/controls/LightControl.kt#L46

https://developers.home-assistant.io/docs/core/entity/light#support-feature

dshokouhi:
[...]

PR #2288, opened 2022-02-15
Wear detail panel

leroyboerefijn:
## Summary
Long press on an entity (label of the chip) will open a detail panel with more information about the entity. If the entity is a light the following options are also present (but only if the entity supports these options):

- Brightness control
- Color temperature control

Other options are not yet supported, since I don't own RGB lights or covers/fans/etc.

Implements #1892 

## Screenshots

Screenshot_20220215_195225_debug
Screenshot_20220215_195236_debug
Screenshot_20220215_195242_debug

Documentation: home-assistant/companion.home-assistant#690


wear/src/main/java/io/homeassistant/companion/android/home/views/EntityUi.kt
```diff
@@ -57,7 +60,24 @@ fun EntityUi(
                 Text(
                     text = friendlyName,
                     maxLines = 2,
-                    overflow = TextOverflow.Ellipsis
+                    overflow = TextOverflow.Ellipsis,
+                    modifier = Modifier.pointerInput(Unit) {
```
jpelgrom, line 64: Long press now only works on the entity's label, I think this should be on the chip in case you have a short label. Is that possible? Otherwise the label should at least fill the width of the chip instead.

wear/src/main/java/io/homeassistant/companion/android/home/views/DetailsPanelView.kt
```diff
@@ -0,0 +1,163 @@
...
+            }
+
+            item {
+                ListHeader(R.string.details)
+            }
+            item {
+                Row(modifier = Modifier.fillMaxWidth().padding(horizontal = 8.dp)) {
+                    Text(stringResource(R.string.entity_id) + ": ${entity.entityId}")
+                }
+            }
```
jpelgrom, line 143: Is the entity ID useful information to have on your watch? I don't think so, but otherwise it should be at the bottom of the list, not at the top.

wear/src/main/java/io/homeassistant/companion/android/home/views/DetailsPanelView.kt
```diff
@@ -0,0 +1,163 @@
...
+                            },
+                            increaseIcon = {
+                                Image(
+                                    asset = CommunityMaterial.Icon.cmd_brightness_7,
+                                    colorFilter = ColorFilter.tint(Color.White)
+                                )
+                            },
+                            modifier = Modifier.padding(bottom = 8.dp)
+                        )
+                    }
```
jpelgrom, line 89: For readability this should probably be changed in to a Composable function (also for color temperature)

jpelgrom (commented):
This is looking quite nice!

A few thoughts, none dealbreakers:

- Did you experiment with setting the color for the `InlineSlider` to something else? For lights the default yellow could be perceived as 'the color temperature is set to this yellow tint'.
- I feel like there should be an on/off switch or a 'toggle/press' button on the details page, similar to how the normal details popup includes this. For lights you can turn them on/off by changing the brightness but that won't be possible for every entity.
- Adding all the variables for detecting features to the `DetailsPanelView` works for now, but does it scale (imagine all domains of device controls on the phone in a single class)? Not sure what the best solution is for this, but probably something with more/layers of Composables.

wear/src/main/java/io/homeassistant/companion/android/home/views/EntityUi.kt
leroyboerefijn: Unfortunately we cannot add a long press on the chip itself, since it already has a gesture handler and we can apparently not override it. So on the label seems the only possibility. You could check the linked issue for more info 🙂

I will make the label fill the width though, good suggestion!

wear/src/main/java/io/homeassistant/companion/android/home/views/DetailsPanelView.kt
leroyboerefijn: Yeah not sure indeed. Figured I just show the default Entity parameters. Will happily move it to the bottom though

wear/src/main/java/io/homeassistant/companion/android/home/views/DetailsPanelView.kt
leroyboerefijn: I tried to do this, but couldn't get that to work due to the `item`s. But now that I look at it again, I could put the `InlindleSlider` into it's own function. And maybe also separate the functionality checks into functions?

leroyboerefijn:
Thanks a lot for the review!

> * Did you experiment with setting the color for the `InlineSlider` to something else? For lights the default yellow could be perceived as 'the color temperature is set to this yellow tint'.

Good point! I'll have a look to see what works better. Maybe plane white?

> * I feel like there should be an on/off switch or a 'toggle/press' button on the details page, similar to how the normal details popup includes this. For lights you can turn them on/off by changing the brightness but that won't be possible for every entity.

Good suggestion, will add it

> * Adding all the variables for detecting features to the `DetailsPanelView` works for now, but does it scale (imagine all domains of device controls on the phone in a single class)? Not sure what the best solution is for this, but probably something with more/layers of Composables.

I agree, will look into it based on the other comment!

wear/src/main/java/io/homeassistant/companion/android/home/views/EntityUi.kt
jpelgrom: I didn't check all comments in the issue before commenting 😳 and I've actually looked at the source code for Chip before and was aware of the limitation, so I really should have noticed this before commenting. 
The comment was because I couldn't reliably trigger the details screen, so changing the label width will help!

wear/src/main/java/io/homeassistant/companion/android/home/views/DetailsPanelView.kt
jpelgrom: `item` requires a `*LazyListScope` to work, so you might need to change the code a little to make it work.

> separate the functionality checks into functions

That is actually a really good idea, maybe you can use extension functions on the `Entity` class so they can also be re-used in other places?

jpelgrom:
> Good point! I'll have a look to see what works better. Maybe plane white?

White is also on the color temperature scale so not sure if that's better. Although based on that logic every color will be a bad choice once we add support for RGB lights.
Maybe we can keep the default yellow for brightness, but change the color of the line (the part that is currently always yellow) on the temperature slider to match the color temperature that is set (so a blue-ish hue when set to a value close to minimum, white in the middle, yellow/orange when near the maximum)?

dshokouhi:
Looking at the debug APK and things look good so far. One thing that came to mind was that I wonder if we should also list the value of current brightness % and also for color temperature so users know what value to go with maybe? Probably not so much an issue with brightness as thats quite obvious but not certain about color temperature here.

leroyboerefijn:
> Maybe we can keep the default yellow for brightness, but change the color of the line (the part that is currently always yellow) on the temperature slider to match the color temperature that is set (so a blue-ish hue when set to a value close to minimum, white in the middle, yellow/orange when near the maximum)?

I looked into this a bit, but couldn't figure out how to convert mired to RGB. I did find this source: https://stackoverflow.com/questions/60777515/how-can-i-convert-mireds-to-rgb-or-display-mireds-in-css
But this is not as straightforward in Kotlin/Java it seems... Let me know if you have any ideas :)

leroyboerefijn:
@jpelgrom @dshokouhi I think I processed all of your review comments. I couldn't really figure out how to do the color yet, so any input is welcome!

I also added a toggle on the top, a `ToggleButton` was the best composable I could find for the job. I'm not too happy about the resulting UI though, so any input here is welcome as well!
image

dshokouhi:
>  I think I processed all of your review comments. I couldn't really figure out how to do the color yet, so any input is welcome!
> 
> I also added a toggle on the top, a `ToggleButton` was the best composable I could find for the job. I'm not too happy about the resulting UI though, so any input here is welcome as well!


Thanks for the screenshot :) 

I think we can probably drop the percentage point for brightness and round it :)

Maybe we can take some inspiration from HA here for some of the design if possible, not sure if the toggle will fit to the side but if we think about how it looks today with the chips it may just make more sense?

image

Do you think if we just put an image as the background for color temperature it would work?

jpelgrom:
Yes, thanks for the screenshot.

I don't think most people really know what mireds are, Kelvin seems more standard, so do we need to show the value for color temperature?

> I looked into this a bit, but couldn't figure out how to convert mired to RGB. (...) Let me know if you have any ideas :)

Actually it looks like the frontend doesn't care about min/max mireds and always uses the same gradient:
https://github.com/home-assistant/frontend/blob/cbd0ef6b651432f6f654222db3c0785b0d48b2aa/src/dialogs/more-info/controls/more-info-light.ts#L581-L586
I'd calculate the values for all 20 steps in advance and use them to set the color on the slider.

> Do you think if we just put an image as the background for color temperature it would work?

I don't think the `InlineSlider` supports setting an image, only simple colors, which is why I suggested changing the color based on the value. Happy to be proven wrong :)

wear/src/main/java/io/homeassistant/companion/android/home/views/DetailsPanelView.kt
```diff
@@ -0,0 +1,207 @@
...
+                }
+            }
+
+            item {
+                ListHeader(R.string.details)
+            }
+            item {
+                Text(
+                    stringResource(R.string.state) + ": ${entity.state}",
```
jpelgrom, line 91: These `stringResource(R.string.id) + ": $value"` type of headings should probably be changed to use a string with a placeholder so the app can support languages where a different symbol is used or there has to be an empty space between the label and colon.

For example:  `<string name="last_changed">Last changed: %1$s</string>` and `stringResource(R.string.last_changed, lastChanged)`

leroyboerefijn: Should have thought of that :P

jpelgrom: Don't forget the brightness and color brightness sliders ;)

leroyboerefijn:
Updated location of toggle looks a bit better indeed! I do worry what happens for long entity names though... But I don't know how to keep that within the bounds of the circular screen anyways...

image

For the colors, good to know that the frontend also doesn't calculate the real color range. I also don't think that the `InlineSlider` support a background image. So the suggestion from @jpelgrom seems the easiest way forward. If only I knew how to precalculate the colorscale...

> I'd calculate the values for all 20 steps in advance and use them to set the color on the slider.

jpelgrom:
> If only I knew how to precalculate the colorscale...

Here you go:
<details><summary>Table</summary>

|Stop|Hex|
|-----|-----|
|0%|#a6d1ff|
|5%|#afd6ff|
|10%|#b8dbff|
|15%|#c1e0ff|
|20%|#cae4ff|
|25%|#d3e9ff|
|30%|#dcedff|
|35%|#e5f2ff|
|40%|#eef6ff|
|45%|#f7fbff|
|50%|#ffffff|
|55%|#fff6e6|
|60%|#ffeccc|
|65%|#ffe3b3|
|70%|#ffd999|
|75%|#ffd080|
|80%|#ffc666|
|85%|#ffbd4d|
|90%|#ffb333|
|95%|#ffaa1a|
|100%|#ffa000|

</details>

(I used an online tool for making a CSS gradient and set it to the values from the frontend, and then added stops at every 5%. Compared side by side and I think it matches.).

leroyboerefijn:
Awesome @jpelgrom ! I implemented the color bar change with your table :D

PR #2302, opened 2022-02-19
Allow enabling chrome remote debugging in settings

### LINKED ISSUE #2162: Provide a debug flag setting to enable UI debugging

Enable the possibility to debug the UI through the Chrome DevTools in the official app build.

Offer a new setting inside the app to toggle debug state on and off.

**Additional context**

Idea was discussed in #2159 and evaluated as feasible.

dshokouhi:
## Summary

Fixes #2162 by adding a remote debugging switch (default off) to settings. On a debug build we will still have debugging enabled no matter what but on other builds we will let the user decide. This will allow users to troubleshoot issues without needing to install a debug build.

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#692

### DIFF
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -769,6 +769,6 @@
     <string name="tile_icon">Select a icon for the tile</string>
     <string name="tile_select">Select a tile to edit</string>
     <string name="shortcut_pinned">Pinned Shortcuts</string>
-    <string name="remote_debugging">Remote Debugging</string>
-    <string name="remote_debugging_summary">Allow remote debugging to troubleshoot Home Assistant frontend issues</string>
+    <string name="remote_debugging">WebView Remote Debugging</string>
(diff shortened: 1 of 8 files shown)


common/src/main/res/values/strings.xml
```diff
@@ -769,4 +769,6 @@
     <string name="tile_icon">Select a icon for the tile</string>
     <string name="tile_select">Select a tile to edit</string>
     <string name="shortcut_pinned">Pinned Shortcuts</string>
+    <string name="remote_debugging">Remote Debugging</string>
+    <string name="remote_debugging_summary">Allow remote debugging to troubleshoot Home Assistant frontend issues</string>
```
JBassett, line 773: I would mention that it's just for the Webview rather than just `Remote Debugging` because that might cause people to think they can debug more than just the webview.

dshokouhi: Ok I'll add that, I tried to avoid that terminology to not confuse users but in this case it's specifically for the webview so that makes sense

JBassett (approved):

PR #2317, opened 2022-02-25
Add new settings screen to change Sensor Update Frequency

dshokouhi:
## Summary

Adds a new setting screen for Sensor Update Frequency that allows users to select between 1 of 3 options. Normal, Fast While Charging and Fast Always.  When set to any option but Normal we will use the `ACTION_TIME_TICK` intent. Fast while charging is only available to users who have the Is Charging sensor enabled otherwise it will default to Fast Always.

Enabling this option will require a device restart, much like adding any intents to the Last Update sensor.

## Screenshots

image

image

image

Documentation: home-assistant/companion.home-assistant#696


JBassett:
I feel like this should be a setting more like the websocket setting rather than 2 toggles with the following options:
* Normal - 15 minute update time
* Fast while charging - 1 minute while charging
* Fast always - 1 minute all the time

dshokouhi:
> I feel like this should be a setting more like the websocket setting rather than 2 toggles with the following options:
> * Normal - 15 minute update time
> * Fast while charging - 1 minute while charging
> * Fast always - 1 minute all the time

I like that, I wasn't sure about the UI either here but will update it later when I'm home. Good suggestion!

dshokouhi:
Ok implemented the changes to match the other screen. Also took the opportunity to add some text about sensor updates to make things more clear. I ended up adding this to the setting table as well since the options are very similar.

dshokouhi:
I tested this by:
- A clean install of the branch
- Clean install with `master` did not enter settings or change websocket setting -> deployed branch
- Clean install with `master` changed websocket settings -> deployed branch

I believe that covers all DB related issues with adding this field, had some fun setting the default value 😛

app/src/main/java/io/homeassistant/companion/android/settings/sensor/views/SensorUpdateFrequencyView.kt
```diff
@@ -0,0 +1,76 @@
...
+            Text(stringResource(R.string.sensor_update_frequency_description))
+        }
+        Divider()
+        RadioButtonRow(
+            text = stringResource(R.string.sensor_update_frequency_normal),
+            selected = sensorUpdateFrequency == SensorUpdateFrequencySetting.NORMAL,
+            onClick = { onSettingChanged(SensorUpdateFrequencySetting.NORMAL) }
+        )
+        val isCharging = AppDatabase.getInstance(context).sensorDao().get(BatterySensorManager.isChargingState.id)
+        if (isCharging != null && isCharging.enabled)
```
jpelgrom, line 50: I don't think hiding the setting if the sensor has been turned off is a good idea, because
- it will make it harder to discover that this is an option, which might be a reason for users to turn on the required sensor
- if the user previously selected fast while charging, and then disabled the sensor, the UI will show the setting as fast always while it hasn't actually changed and will change back once the sensor is re-enabled

Personally, I would show a warning (different text?) or disable the option (greyed out) instead of hiding it when the required sensor is not activated. Falling back to normal update frequency instead of always fast updating also makes more sense to me, as it is the 'safer'/default option and users would have selected always if they wanted that.

dshokouhi: > I don't think hiding the setting if the sensor has been turned off is a good idea, because
> 
> * it will make it harder to discover that this is an option, which might be a reason for users to turn on the required sensor
> * if the user previously selected fast while charging, and then disabled the sensor, the UI will show the setting as fast always while it hasn't actually changed and will change back once the sensor is re-enabled
> 
> Personally, I would show a warning (different text?) or disable the option (greyed out) instead of hiding it when the required sensor is not activated. 

I like disabling the option, I think that makes more sense instead of hiding it.

> Falling back to normal update frequency instead of always fast updating also makes more sense to me, as it is the 'safer'/default option and users would have selected always if they wanted that.

So the reason why it falls back to Fast Always is because once a user selects the fast option we automatically register the intent for `TIME_TICK` on the next app start. I didn't want to change the setting when that happens because that means we need to add logic to the sensor detail screen to change the DB value but also to warn them that they need to restart the application to unregister the intent.  When a user enables this option (Fast While Charging) they already see the text saying the sensor is required to be enabled so I thought that was enough of a warning there and at this point they already had it enabled because they see the option. So disabling the sensor becomes a conscious choice. I would like to avoid having to overcomplicate the logic on the sensor detail screen especially given that the current logic needs to be switched to Compose eventually.

Do you think stronger wording and disabling instead of hiding will make it more clear?

I suppose we could also change the text for Normal to say "most battery friendly" as well?

app/src/main/res/drawable/ic_phone_charging.xml
```diff
@@ -0,0 +1,7 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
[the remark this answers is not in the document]
dshokouhi: 🙈

app/src/main/java/io/homeassistant/companion/android/settings/sensor/views/SensorUpdateFrequencyView.kt
jpelgrom: > So the reason why it falls back to Fast Always (...)

I understand why it's happening from a technical point of view, but from the user's point of view unknown charging status != charging (otherwise the option should be called fast except when unplugged).

> When a user enables this option (Fast While Charging) they already see the text (...) So disabling the sensor becomes a conscious choice.

Maybe when disabling it right now, but if they disable it a few months later? This is the only non-sensor setting in the app to use a sensor, and the update frequency could also have [read the API directly](https://developer.android.com/reference/android/os/BatteryManager#isCharging()) on most devices so this might be unexpected if they don't remember all the text for this setting (see: the websocket service doesn't require you to enable the interactive sensor).

> Do you think stronger wording and disabling instead of hiding will make it more clear?

I think the confusion can be solved if the 'while charging' setting text is changed when the sensor is disabled, by removing the last sentence and placing more info on a new line which will make it stand out:
- If not selected: turn on the sensor to use this option
- If selected: the required sensor is disabled, currently always updating fast

> I suppose we could also change the text for Normal to say "most battery friendly" as well?

Updating less / only while charging / ... is even more battery friendly, I think 'Normal' is fine.

dshokouhi: > Maybe when disabling it right now, but if they disable it a few months later? This is the only non-sensor setting in the app to use a sensor, and the update frequency could also have [read the API directly](https://developer.android.com/reference/android/os/BatteryManager#isCharging()) on most devices so this might be unexpected if they don't remember all the text for this setting (see: the websocket service doesn't require you to enable the interactive sensor).

if only that API was available in API 21



> I think the confusion can be solved if the 'while charging' setting text is changed when the sensor is disabled, by removing the last sentence and placing more info on a new line which will make it stand out:
> 
> * If not selected: turn on the sensor to use this option
> * If selected: the required sensor is disabled, currently always updating fast

Good call, will update the option text to better reflect this. Personally I don't think too many people will be disabling this sensor as its enabled by default and a pretty common use case.

jpelgrom: > if only that API was available in API 21

I know and that is why I used "on most devices". However it turns out the intent is sticky so there is actually an option for this that should work on API 21, I updated my comment a few minutes later but apparently too late :)

dshokouhi: > > if only that API was available in API 21
> 
> I know and that is why I used "on most devices". However it turns out the intent is sticky so there is actually an option for this that should work on API 21, I updated my comment a few minutes later but apparently too late :)

funny thats exactly what our BatterySensorManager does today, as its a sticky intent thats much better to go by I think. Let me run some tests...this will simplify things even more. Good call :)

dshokouhi: ok removed that requirement and updated the text, thanks for the suggestion :)

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
```diff
@@ -83,11 +84,10 @@ abstract class SensorReceiverBase : BroadcastReceiver() {
...
 
         ioScope.launch {
-            val sensorDao = AppDatabase.getInstance(context).sensorDao()
-            val isChargingSensor = sensorDao.get(BatterySensorManager.isChargingState.id)
             val settingDao = AppDatabase.getInstance(context).settingsDao().get(0)
-            if (isChargingSensor != null && isChargingSensor.state == "false" &&
-                isChargingSensor.enabled && settingDao != null &&
+            val batteryStatusIntent = context.registerReceiver(null, IntentFilter(Intent.ACTION_BATTERY_CHANGED))
+            val isCharging = batteryStatusIntent?.let { BatterySensorManager.getIsCharging(it) }
+            if (isCharging == false && settingDao != null &&
```
jpelgrom, line 90: Suggestion: use `isCharging != true` to also skip updating if the intent returned null (unknown status)

dshokouhi: again a great suggestion, thank you :)

common/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
```diff
@@ -483,6 +487,12 @@ abstract class AppDatabase : RoomDatabase() {
             }
         }
 
+        private val MIGRATION_23_24 = object : Migration(23, 24) {
+            override fun migrate(database: SupportSQLiteDatabase) {
+                database.execSQL("ALTER TABLE `settings` ADD `sensorUpdateFrequency` TEXT NOT NULL DEFAULT ${SensorUpdateFrequencySetting.NORMAL}")
```
JBassett, line 492: They sugest migrations not depend on code at all in case our next build changes that constant.  Also need `'`.
```suggestion
                database.execSQL("ALTER TABLE `settings` ADD `sensorUpdateFrequency` TEXT NOT NULL DEFAULT 'NORMAL'")
```

JBassett (changes requested):

JBassett (approved):

PR #2332, opened 2022-03-02
Add fan speed to Wear OS detail panel

jpelgrom:
## Summary

Now that there is a details screen for Wear OS, this PR adds support for controlling a `fan` entity's speed.

On the `InlineSlider`, `steps` has been set to 9 to allow controlling in steps of 10%, I think that is a good balance between quickly changing the value and specific control for fans.

(Also updates the label for the light brightness slider because that is also in %.)

## Screenshots

Wear OS app showing the details screen for a fan. At the top of the screen, the entity name "itho_fan" is shown with a toggle. Below that, there is a slider with "Speed: 20%" and two buttons to lower and raise the speed.

Documentation: home-assistant/companion.home-assistant#702


dshokouhi (approved):
Looks good to me

JBassett (approved):

Issue #2285, opened 2022-02-14
Option to clean up Notification Channels from within the App


RoboMagus:
**Is your feature request related to a problem? Please describe.**

From time to time new notification channels are added, and others fall out of use. This tends to accumulate a rather big list of channels. This is more a problem in how Android itself handles this, as from the notification settings users are unable to remove unused channels. However, as this can be done from within an app this is a feature that would improve the companion app.

**Describe the solution you'd like**

The option to delete unused notification channels from within the companion app settings.

**Describe alternatives you've considered, if any**

Manually going through the list of notification channels on my phone, and typing the notification command to clear it in HomeAssistant. This does work, however is quite tedious!

**Additional context**

This might not be a big _game-changing_ feature, but one that would definately improve the user experience of the app for those of us that often tinker with their notifications.

dshokouhi:
So instead of using this: https://companion.home-assistant.io/docs/notifications/notifications-basic#removing-a-channel

You want to have a UI in the app settings to delete it?

RoboMagus:
@dshokouhi, indeed.
From within the android app it should be fairly easy to populate a list of notification channels with a button next to it to delete an item.

That would make it a lot easier than sending the remove channel commands through HA.

dshokouhi:
Sure we can do that, in this case we will need to create a list removing any app created channel and only show user created channels.

PR #2340, opened 2022-03-04
Add manage notification channels settings screen so users can easily delete channels created by the app

dshokouhi:
## Summary

Fixes: #2285 by creating a page that lists all app created notification channels, replacing the existing page that simply takes the user to the screen listing all channels. User can then move on to edit the settings or delete the channel.  I did not do anything special about app created notification channels because users can already delete those using the command that exists today. Instead I opted to briefly explain what deleting means and what happens if they delete an app created channel.  There is also no delete confirmation because deleting isn't really deleting and we don't have a confirmation when using a command to delete it.

## Screenshots

image

n/a we dont talk about these screens in the docs


app/src/main/java/io/homeassistant/companion/android/settings/notification/NotificationViewModel.kt
```diff
@@ -0,0 +1,44 @@
...
+@RequiresApi(Build.VERSION_CODES.O)
+class NotificationViewModel @Inject constructor(
+    application: Application
+) : AndroidViewModel(application) {
+
+    val app = application
+
+    private var notificationManager = application.applicationContext.getSystemService<NotificationManager>()!!
+
+    var channelList = mutableStateListOf(notificationManager.notificationChannels)
```
jpelgrom, line 25: Right now this is a SnapshotStateList of the list of notification channels, is there any specific reason for that? Removing the inner list makes more sense to me, and it will also allow you to remove all the `[0]`s in the composable.

common/src/main/res/values/strings.xml
```diff
@@ -791,4 +791,5 @@
...
     <string name="sensor_worker_notification_channel">Manage Sensor Worker Notification</string>
     <string name="sensor_update_frequency_summary">Change the frequency of updates for sensors that do not update instantly.</string>
+    <string name="notification_channels_description">Notification channels allow you to control the look and feel of a certain type of notification. You can create notification channels on the fly by creating a notification with the channel parameter, refer to the help icon in the upper right hand corner to learn more. Below is a list of all notification channels created by your server admin or the app. Select the edit icon to adjust channel settings like ringtone and Do Not Disturb override. Select the delete icon to remove the channel from this list.\n\nDeleting a channel does not reset channel settings, if a notification is created with the same channel name it will retain its previous settings.\n\nWhile you may delete an app created notification channel, it will reappear once it is recreated.</string>
```
jpelgrom, line 794: Some text suggestions. Feel free to ignore them if you don't agree, I'm not a native English speaker.

> You can create notification channels on the fly by creating a notification with the channel parameter

"on the fly" sounds unnecessary to me, and you actually need to send a notification, not just write one.

> Below is a list of all notification channels created by your server admin or the app

"by your server admin" hasn't been used before, usually it's "your server" or "your instance" which sounds better to me. It's also a list of channels created by the server *and* app most of the time.

> if a notification is created with the same channel name it will retain its previous settings

The app never creates notifications, I'd use "received", and I think re-used sounds better than retained.

> While you may delete an app created notification channel, it will reappear once it is recreated.

This doesn't explain when it is recreated, up until now it has been suggested that the only way to recreate a channel is when sending a notification with the same `channel` parameter.

jpelgrom (commented):
In addition to the specific notes:
- I don't think the delete option should be available for the channels created by the app (`SensorWorker` and `Persistent Connection`) because the app 'requires' them and they will always get recreated. Even if you can do it via notification commands that doesn't mean we should expose it in the interface 🙂
- While deleting isn't really a hard delete and the settings re-appear when you send another notification for that channel, I still feel like there should be a confirmation required when you press on the delete button, or even better a Snackbar with an 'Undo' action (which creates a new channel, and the system will re-use the settings)

app/src/main/java/io/homeassistant/companion/android/settings/notification/views/NotificationChannelView.kt
```diff
@@ -0,0 +1,80 @@
...
+                    IconicsDrawable(context, "cmd-pencil").toBitmap().asImageBitmap()
+                val deleteIcon =
+                    IconicsDrawable(context, "cmd-delete").toBitmap().asImageBitmap()
+                Row {
+                    Icon(
+                        editIcon,
+                        "",
+                        modifier = Modifier
+                            .clickable { notificationViewModel.editChannelDetails(channelId) }
+                            .padding(end = 10.dp)
```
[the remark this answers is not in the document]
dshokouhi: I was trying to do better spacing here :) maybe I should just even it out lol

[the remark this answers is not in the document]
jpelgrom: If you follow the suggestion for 48dp start sizes, and set the padding for the entire row to `start = 12.dp`, I think it looks quite nice and the text/icons are indented similar to how it is now (2dp more)

app/src/main/java/io/homeassistant/companion/android/settings/notification/NotificationViewModel.kt
dshokouhi: Without this compose will not recompose when I call `updateChannelList`, even after I update it to clear all values.

common/src/main/res/values/strings.xml
dshokouhi: Made some adjustments based on these recommendations :)

dshokouhi:
> * I don't think the delete option should be available for the channels created by the app (`SensorWorker` and `Persistent Connection`) because the app 'requires' them and they will always get recreated. Even if you can do it via notification commands that doesn't mean we should expose it in the interface 🙂

Done now hiding the delete icon for these channels and also explaining in the note above that they cannot be deleted.



> * While deleting isn't really a hard delete and the settings re-appear when you send another notification for that channel, I still feel like there should be a confirmation required when you press on the delete button, or even better a Snackbar with an 'Undo' action (which creates a new channel, and the system will re-use the settings)

I went with the Snackbar approach here as it feels better fit, its not a permanent delete which should call for a confirmation I don't think. Nice with the undo action!

app/src/main/java/io/homeassistant/companion/android/settings/notification/NotificationViewModel.kt
jpelgrom: Any specific scenario you're testing where it doesn't work? When deleting a channel, and then recreating it using the 'Undo' option, I see it disappear and re-appear as expected. Code changed here to `notificationManager.notificationChannels.toMutableStateList()`, in the `updateChannelList` function to use `addAll` instead of `add`, and removed the `[0]` from the composable.

app/src/main/java/io/homeassistant/companion/android/settings/notification/views/NotificationChannelView.kt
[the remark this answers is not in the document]
jpelgrom: I actually meant the entire row with channel name + buttons (currently line 60), not the inner row for the buttons, now there is double padding but it also looks good 👍

app/src/main/java/io/homeassistant/companion/android/settings/notification/views/NotificationChannelView.kt
```diff
@@ -0,0 +1,115 @@
...
+                                                    context.getString(R.string.undo)
+                                                )
+                                                .let {
+                                                    if (it == SnackbarResult.ActionPerformed) {
+                                                        notificationViewModel.createChannel(
+                                                            NotificationChannel(
+                                                                channel.id,
+                                                                channel.name,
+                                                                channel.importance
+                                                            )
```
jpelgrom, line 99: Why don't you reuse the original `channel` instead of using the properties to create a new one?

app/src/main/java/io/homeassistant/companion/android/settings/notification/views/NotificationChannelView.kt
```diff
@@ -0,0 +1,115 @@
...
+fun NotificationChannelView(
+    notificationViewModel: NotificationViewModel
+) {
+    val scaffoldState = rememberScaffoldState()
+    val scope = rememberCoroutineScope()
+    val context = LocalContext.current
+    Scaffold(
+        scaffoldState = scaffoldState
+    ) {
+        val sortedChannelList = notificationViewModel.channelList[0].toList().sortedBy { it.name.toString() }
```
jpelgrom, line 46: - `toList()` is unnecessary because it is already a list
- Because you're always sorting the list, this should be done in the viewmodel instead

app/src/main/java/io/homeassistant/companion/android/settings/notification/views/NotificationChannelView.kt
dshokouhi: I thought I am using the original channel that provides the data for the item?

jpelgrom: Yes but you're using it to create anew `NotifictionChannel`. As I understand it, you can also replace these 5 lines with `channel` and it still works.

dshokouhi: ah I see what you mean, thank you :) for some reason I was under the impression I needed to set the properties separtely.

app/src/main/java/io/homeassistant/companion/android/settings/notification/NotificationViewModel.kt
dshokouhi: ah I was missing `.toMutableStateList()`

JBassett (approved):

Issue #2339, opened 2022-03-04
location won't appear on map card / case sensitive


finnzz:
Android app: 2202.2.1-full
home assistant: 2202.2.6

The geocoded location entity doesn't appear on the Lovelace map card because it is case sensative to the attributes Longitude / Latitude.

Most entities use all lowercase attribute labels.  Perhaps the HA companion app should be using lowercase attributes as well?

Or at least change the Latitude / Longitude attribute labels to latitude / longitude?  Right now the only user solution seems to be creating a new sensor to relabel the attribute names for those two.

dshokouhi:
> Right now the only user solution seems to be creating a new sensor to relabel the attribute names for those two.

Did you already test this theory to confirm it works? I didn't know attributes had to be lowercase. Don't see the HA dev docs saying this but we can make this change for these attributes so it plays nicely if you confirm the casing was the issue.

finnzz:
Yes, there is a thread about the geocoded_coded entity from the app not showing on the map card because it's looking for latitude/longitude, rather than Latitude/Longitude. And after creating a mirrored sensor with lowercase longitude/latitude attribute names it shows up on the map.

https://community.home-assistant.io/t/lovelace-map-card-not-displaying-geocoded-entities/210947

I can also confirm that if go into developer tools / states and set the geocoded location entity attribute to lowercase longitude/ latitude that it appears on the map card.

I don't actually know if lower case attribute names is official HA etiquette, but after seeing that thread I looked over a bunch of my entities, and I couldn't find any that use uppercase letters.

The case sensitivity seems a little unnecessary, but maybe there is a good reason for it?

dshokouhi:
Ok we will fix it for these attributes

I was pointed to some more correct HA core code:
https://github.com/home-assistant/core/blob/777ae584d49e45a074f74cdff47f0706597907c8/homeassistant/const.py#L434-L435

thanks for raising the issue :)

finnzz:
Thank you!

dshokouhi:
lets leave this open until the PR is merged :)

PR #2341, opened 2022-03-04
Breaking Change: Make coordinate attributes all lower case for HA compatibility

dshokouhi:
## Summary

This PR updates all attributes  to be all lower case to match the HA expectation in order to fix: #2339 

https://github.com/home-assistant/core/blob/777ae584d49e45a074f74cdff47f0706597907c8/homeassistant/const.py#L434-L435

While I am ok with going through all sensors updating the attributes what do we do about the notification sensors that do not follow that format? 

https://github.com/home-assistant/iOS/blob/master/Sources/Shared/API/Webhook/Sensors/GeocoderSensor.swift#L95-L110

Breaking change because users who relied on these attributes will need to update their configuration

Documentation: home-assistant/companion.home-assistant#706


JBassett:
I would at least update all the attributes on this sensor.  We can leave the rest of the sensors for another time, but it's strange to only fix a couple.

JBassett (approved):

hmmbob:
This has unexpected side effects - the device now shows twice on the map (1x `device_tracker`, 1x `geocoded location sensor`). See above issue

Issue #1270, opened 2021-01-01
Notifications: GIFs only rendering the first frame


SeraphimSerapis:
**Home Assistant Android version:**
3.0.2-full

**Android version:**
11

**Phone model:**
Pixel 5

**Home Assistant version:**
2020.12.2

**Last working Home Assistant release (if known):**
-

**Description of problem:**
When using the mobile notification feature with a GIF instead of a PNG or JPEG, it seems like only a single frame of the received GIF is rendered.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```

**Screenshot of problem:**
![signal-2021-01-01-212035](https://user-images.githubusercontent.com/160543/103445646-5a049100-4c77-11eb-8d9f-2f5ae7b86cce.png)

**Additional information:**
Example code to trigger the push notificiation:

```yaml
message: "Test GIF"
data:
  image: "https://media1.tenor.com/images/aba522811b497086c5d35757d6cfb269/tenor.gif"
```

dshokouhi:
android notifications do not support GIF https://stackoverflow.com/a/58482901 so this would not be a bug but maybe an enhancement request (if at all possible)

SeraphimSerapis:
Got it. Thanks!

dshokouhi:
they only take bitmap which is what the app provides

https://developer.android.com/reference/android/app/Notification.BigPictureStyle

SeraphimSerapis:
Reopening for a potential feature request.

I know the Nest app shows a brief animation for camera notifications. Is that a custom implementation?

dshokouhi:
yes those are all custom implementations and override the default android notifications, they need to be built from the ground up, even then I am not sure if a URL gif will work I believe the app has to create it and store it locally in order to display it instead of real time rendering like we do today

dshokouhi:
Found something that might be usable however it has serious limitations such as the user will need to supply the list of image URLs to flip through so a GIF wont work here.  This might be how the other camera apps do it.

https://nishi-emedstore.medium.com/notification-with-image-slider-example-in-android-2758b6d43d38

dshokouhi:
if you have updated to Android 12 it seems they now natively supported animated images. Curious to see if this just "works" now?

https://developer.android.com/about/versions/12/features#enriched_image_support_for_notifications

SeraphimSerapis:
[...]

PR #2342, opened 2022-03-05
FEAT: Ability to show video in notification as a gif

NickM-27:
## Summary

I have added the ability for a user to define `video` as part of an Android notification. This video will be downloaded and home assistant will attempt to parse frames from this video to show in a notification. The goal is to find 5 frames but it will work for any number of frames < 5 as well. 

## Screenshots

https://user-images.githubusercontent.com/14866235/156859680-5c76f67d-4e27-4ec5-a6ee-c6c96937ce91.mp4

Documentation: home-assistant/companion.home-assistant#705

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -1201,6 +1201,14 @@ class MessagingManager @Inject constructor(
                             )
                         }
 
+                        data[TITLE]?.let { rawTitle ->
+                            remoteViewFlipper.setTextViewText(R.id.title, rawTitle)
+                        }
+
+                        data[MESSAGE]?.let { rawMessage ->
+                            remoteViewFlipper.setTextViewText(R.id.info, rawMessage)
+                        }
+
                         builder.setCustomBigContentView(remoteViewFlipper)
                         builder.setStyle(NotificationCompat.DecoratedCustomViewStyle())
                     }
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -97,7 +97,7 @@ class MessagingManager @Inject constructor(
         const val TIMEOUT = "timeout"
         const val IMAGE_URL = "image"
         const val ICON_URL = "icon_url"
-        const val VIDEO_URL = "video_url"
+        const val VIDEO_URL = "video"
         const val LED_COLOR = "ledColor"
         const val VIBRATION_PATTERN = "vibrationPattern"
         const val PERSISTENT = "persistent"
@@ -204,6 +204,10 @@ class MessagingManager @Inject constructor(
             listOf(BLE_TRANSMIT_HIGH, BLE_TRANSMIT_LOW, BLE_TRANSMIT_MEDIUM, BLE_TRANSMIT_ULTRA_LOW)
         val BLE_ADVERTISE_COMMANDS =
             listOf(BLE_ADVERTISE_BALANCED, BLE_ADVERTISE_LOW_LATENCY, BLE_ADVERTISE_LOW_POWER)
+
+        // Video Values
+
+        const val VIDEO_MAX_FRAMES = 5
     }
 
     private val mainScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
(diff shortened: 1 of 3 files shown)


dshokouhi:
This PR needs 1 additional PR to add the parameter to firebase as well, just need to add `video` here since its a new field

https://github.com/home-assistant/mobile-apps-fcm-push/blob/master/functions/android.js#L43-L52

dshokouhi:
This PR made me think about https://github.com/home-assistant/android/issues/1270 I wonder if later on we can leverage the code here to fix it?

NickM-27:
> This PR made me think about #1270 I wonder if later on we can leverage the code here to fix it?

yeah, sounds like there is a first party fix that works for API 31 but perhaps the `MediaMetaDataRetriever` can parse frames from gifs as well. I wouldn't mind investigating as part of this PR but perhaps that would belong on its own PR.

dshokouhi:
> I wouldn't mind investigating as part of this PR but perhaps that would belong on its own PR.

lets do a separate PR for that to keep this one as small as possible :)

NickM-27:
@dshokouhi Let me know what you think of the docs changes. One thing still in my head that I think may be worth adding in the notes is that videos less than 10 seconds long probably won't work very well since it may not find enough frames to show.

Also fcm PR: home-assistant/mobile-apps-fcm-push#705

dshokouhi:
I just left a comment on the docs PR :) 



> One thing still in my head that I think may be worth adding in the notes is that videos less than 10 seconds long probably won't work very well since it may not find enough frames to show

I agree I think thats worth mentioning as a note specific to android as I dont think iOS has to worry about that. The only thing here that we want to make sure is that the service call can be the same for iOS and android :) Functionality doesn't need to match just that a user can put the 2 devices in a group and send the notification there.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -1168,6 +1180,98 @@ class MessagingManager @Inject constructor(
...
+                            )
+                        }
+
+                        data[TITLE]?.let { rawTitle ->
+                            remoteViewFlipper.setTextViewText(R.id.title, rawTitle)
+                        }
+
+                        data[MESSAGE]?.let { rawMessage ->
+                            remoteViewFlipper.setTextViewText(R.id.info, rawMessage)
+                        }
```
JBassett, line 1210: Doesn't the `handleText` function handle this?

NickM-27: No, when you set the style as `DecoreatedCustomViewStyle` you actually control the entire notification viewport so the title and text part from the builder is overwritten / not included and the custom view needs to include it itself.

NickM-27: To clarify on that last point, the docs here: https://developer.android.com/reference/android/app/Notification.DecoratedCustomViewStyle show how you can override specific parts of the notification. In this case I am overriding the bigContentView so the heads up view and the standard content view will use the `handleText` part but when the notification is expanded it needs to be handled by the custom view here.

NickM-27:
@dshokouhi Is there anything else needed for this PR? Just curious what the timeline may be 😄

dshokouhi:
Right now this PR is just waiting for a more official code review, depends on how much time the reviewer has.  Usually 1-2 weeks for a PR to be reviewed but could be longer, maybe faster if its fixing a critical bug :) There is a low PR count so its on the radar

NickM-27:
> Right now this PR is just waiting for a more official code review, depends on how much time the reviewer has.  Usually 1-2 weeks for a PR to be reviewed but could be longer, maybe faster if its fixing a critical bug :) There is a low PR count so its on the radar

Thanks! Sounds good. By the way I looked into use the MediaMetaDataRetriever to parse gifs but unfortunately I don't believe that format is supported :/

I'll have to look and see if there are any other approaches that are lightweight. (I know Glide can do it but that's a big library for that problem alone)

JBassett (approved):

conorlap:
I can't seem to get this working. I've installed [beta-2150-261ae31f](https://github.com/home-assistant/android/releases/tag/beta-2150-261ae31f) but no image/gif is appearing. Here's how I'm calling the notification:

```
service: notify.mobile_app_s20
data:
  message: Test
  data:
    video: "https://download.samplelib.com/mp4/sample-20s.mp4"

```
Am I doing something wrong?

Issue #2343, opened 2022-03-05
Intent action.CALL does not work. 


Hellis81:
**Is your feature request related to a problem? Please describe.**

Using the intent `android.intent.action.CALL` does not work, using `android.intent.action.DIAL` does work. 
It would be nice to be able to call directly.
HA app says it's a malformed intent.

**Describe the solution you'd like**

Unsure, but possibly a permission is needed according to Daniel.

**Describe alternatives you've considered, if any**

Using Tasker works, but it would be nice to have it in HA app.

**Additional context**

https://community.home-assistant.io/t/android-intents-sending-receiving-list-starting-activities-too/276192/81
https://developer.android.com/guide/components/intents-common#Phone

PR #2344, opened 2022-03-05
Check for phone permissions when using call intent

dshokouhi:
## Summary

Fixes: #2343 by adding the proper permission to the manifest and then directing users who have not granted the permission to the app info screen to grant the permission, just like we do for bluetooth permissions.

Documentation: home-assistant/companion.home-assistant#707


dshokouhi:
I have a feeling we may not be able to move forward with this PR because this might be one of those protected permissions we don't qualify for.  Just realized it but will leave it open for discussion for now.

https://developer.android.com/reference/android/content/Intent#ACTION_CALL

`Note: there will be restrictions on which applications can initiate a call;`

JBassett (approved):

PR #2359, opened 2022-03-09
Migrate "Manage Sensors" to Compose and ViewModels

jpelgrom:
## Summary

This PR migrates the "Manage Sensors" screen in the settings (and the detail/configuration screen for individual sensors) to use Compose and ViewModels with flows for showing realtime data instead of large Fragments and XML preferences.

Scrolling performance for Compose with long lists isn't great with debug builds, but is as smooth as the previous `PreferenceFragmentCompat`-based screens when using a release build. Not sure why this happens but I built a release build to verify and it really helps.

There is also quite a lot of complexity in these screens, so I haven't made any significant changes to the design for now to keep this PR easier to review.

## Screenshots

As mentioned above, the screens haven't changed significantly and unless you compare side-by-side most users probably won't notice the difference.
|       |Light|Dark|
|-----|-----|-----|
|'Manage sensors' list|image|image|
|Details for a sensor that is off|image|image|
|Details for a sensor that is on with attributes/settings|image|image|
|Details for a sensor that is on without attributes/settings|image|image|
|Sensor setting: `toggle` (first setting), and disabled settings (all except first and last setting)|image|image|
|Sensor setting: `string`/`number`|image|image|
|Sensor setting: `list` (single choice)|image|image|
|Sensor setting: `list-*` (multiple choices)|image|image|

Not needed


dshokouhi:
Very nice speed improvements! I noticed the toggles in some settings are now so fast you don't even see them on for a few seconds. This was not easy to accomplish since there is so much logic but hats off to you!

 Haven't had a chance to dive in to the code yet but the brief tests with the debug build were great.

jpelgrom:
Made a few small design tweaks, one of which includes the `Switch` 'thumb color' (circle) when off. The switch in onboarding already had a different color set so clearly someone agrees that the default is bad even if it is [by design according to Google](https://issuetracker.google.com/issues/185418990). I've used an eyedropper/pipette tool to get the color from non-Compose switches, and updated the switch in onboarding to match as well.

JBassett (approved):

PR #2372, opened 2022-03-14
Update Kotlin to 1.6 and Compose to 1.1

jpelgrom:
## Summary

Update Kotlin to 1.6.10 and Compose to 1.1.1. Changes, aside from the version numbers:
 - Remove `@ExperimantalComposeUiApi` and `@ExperimentalWearMaterialApi` annotations from the Home Assistant codebase, because the functions where they are used aren't experimental, we are just using an experimental API in those functions. After updating to 1.6 we now have to opt in to functions that use these annotations at a higher level for the build to succeed. The `@OptIn(ExperimantalXApi::class)` annotations are placed at the 'lowest level' usage (so where we directly use a function from Compose that has the experimental API annotation).
 - Websocket: `ProducerScope` is no longer experimental so the annotation can be completely removed.
 - Wear: the animation APIs that are used for the time text are no longer experimental so the annotation can be completely removed.

Compose 1.1 also includes a [breaking change for touch target sizing](https://android-developers.googleblog.com/2022/02/jetpack-compose-11-now-stable.html); I've checked all Compose based screens and nothing appears to be broken as a result of these changes.

Closes #2010

## Screenshots

n/a

Not needed


JBassett (changes requested):
I just checked and because of https://github.com/home-assistant/android/pull/2359 you'll need to rebase your changes and make a couple of changes to get the build working because if I merge this we'll fail the build.  Other than that it looks good.

jpelgrom:
Not sure what I did wrong but after rebasing there were a lot of additional changes, so I've just done a regular merge which seems to work as well

JBassett (approved):

Issue #2386, opened 2022-03-21
Enabling Second SIM sensor crashes the app on S22 Ultra


atulalvenkar:
**Home Assistant Android version:**
2022.3.0-full
**Android version:**
Android 12 One UI 4.1
**Phone model:**
Samsung Galaxy S22 Ultra
**Home Assistant version:**
2022.3.5
**Last working Home Assistant release (if known):**

**Description of problem:**
I converted my secondary SIM from a physical SIM to an eSIM couple days back. Today I saw the the second sim sensor was reporting unavailable in HA. I tried to toggle the sensor and the app crashed when turning it back on. I don't remember this happening when I had the physical SIM.

dshokouhi:
Thanks for reporting this issue, looking at sentry I believe this is the error. Seems like we are unable to get the display name which is used for the state of the sensor.

```
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.lang.Object.toString()' on a null object reference
    at io.homeassistant.companion.android.sensors.PhoneStateSensorManager.updateSimSensor(PhoneStateSensorManager.kt:124)
    at io.homeassistant.companion.android.sensors.PhoneStateSensorManager.requestSensorUpdate(PhoneStateSensorManager.kt:68)
    at io.homeassistant.companion.android.sensors.SensorDetailFragment.onCreatePreferences$lambda-2$lambda-1(SensorDetailFragment.kt:143)
    at io.homeassistant.companion.android.sensors.SensorDetailFragment.$r8$lambda$XMUSI_7loU7HvSZWwY4Z9mlnm0Y
    at io.homeassistant.companion.android.sensors.SensorDetailFragment$$ExternalSyntheticLambda5.onPreferenceChange
    at androidx.preference.Preference.callChangeListener(Preference.java:1118)
    at androidx.preference.TwoStatePreference.onClick(TwoStatePreference.java:67)
    at androidx.preference.Preference.performClick(Preference.java:1182)
    at androidx.preference.Preference.performClick(Preference.java:1166)
    at androidx.preference.SwitchPreference.performClick(SwitchPreference.java:195)
    at androidx.preference.Preference$1.onClick(Preference.java:181)
    at android.view.View.performClick(View.java:7792)
    at android.view.View.performClickInternal(View.java:7769)
    at android.view.View.access$3800(View.java:910)
    at android.view.View$PerformClick.run(View.java:30213)
    at android.os.Handler.handleCallback(Handler.java:938)
[...]

PR #2389, opened 2022-03-21
Fix SIM sensor crash

dshokouhi:
## Summary

Fixes: #2386 by checking that display name is not null and also catching any other potential failures

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
@@ -131,7 +131,7 @@ class PhoneStateSensorManager : SensorManager {
                             attrs["mcc"] = info.mccString.toString()
                             attrs["mnc"] = info.mncString.toString()
                             attrs["is opportunistic"] = info.isOpportunistic
-                             attrs["data roaming"] = if (info.dataRoaming == SubscriptionManager.DATA_ROAMING_ENABLE) "enable" else "disable"
+                            attrs["data roaming"] = if (info.dataRoaming == SubscriptionManager.DATA_ROAMING_ENABLE) "enable" else "disable"
                         }
                     } catch (e: Exception) {
                         Log.e(TAG, "Unable to get SIM data", e)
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
```diff
@@ -121,16 +122,22 @@ class PhoneStateSensorManager : SensorManager {
...
+                        attrs["carrier name"] = info.carrierName
+                        attrs["iso country code"] = info.countryIso
+                        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+                            attrs["carrier id"] = info.carrierId
+                            attrs["mcc"] = info.mccString.toString()
+                            attrs["mnc"] = info.mncString.toString()
+                            attrs["is opportunistic"] = info.isOpportunistic
+                            if (info.dataRoaming == SubscriptionManager.DATA_ROAMING_ENABLE) attrs["data roaming"] =
+                                "enable"
+                            else attrs["data roaming"] = "disable"
```
JBassett, line 137:
```suggestion
                             attrs["data roaming"] = if (info.dataRoaming == SubscriptionManager.DATA_ROAMING_ENABLE) "enable" else "disable"
```

app/src/main/java/io/homeassistant/companion/android/sensors/PhoneStateSensorManager.kt
```diff
@@ -121,16 +122,22 @@ class PhoneStateSensorManager : SensorManager {
...
-                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
-                        attrs["carrier id"] = info.carrierId
-                        attrs["mcc"] = info.mccString.toString()
-                        attrs["mnc"] = info.mncString.toString()
-                        attrs["is opportunistic"] = info.isOpportunistic
-                        if (info.dataRoaming == SubscriptionManager.DATA_ROAMING_ENABLE) attrs["data roaming"] = "enable"
-                        else attrs["data roaming"] = "disable"
+                    try {
+                        if (info.displayName != null)
+                            displayName = info.displayName.toString()
```
JBassett, line 127:
```suggestion
                        displayName = info.displayName?.toString() ?: displayName
```

JBassett (changes requested):

jpelgrom:
FYI, I've opened another PR #2391 to prevent crashing the UI as well. Still a good idea to catch and handle errors in the manager itself though :)

dshokouhi:
> FYI, I've opened another PR #2391 to prevent crashing the UI as well. Still a good idea to catch and handle errors in the manager itself though :)

I agree we need it in both places, good idea!

JBassett (approved):

PR #2396, opened 2022-03-23
Fix gesture detection on WebViewActivity

### LINKED ISSUE #2392: Webview gesture quickbar is triggered on 3 finger touch, not 3 finger swipe down
beta-2188-e24712dc-full

**Android version:**
12

**Phone model:**
Galaxy S20+

**Home Assistant version:**
2022.3.5

Don't know

**Description of problem:**
As soon as I touch the screen with 3 fingers, the quickbar appears, no need to swipe down. I cannot tell when it started to act like that. I tried with and without pinch to zoom enabled, same result. Also, the blue circle of one the finger stays displayed on all screens, not all the time.

https://imgur.com/9PNEctC

I tried to find relevant things in the logs but didn't find any. The original logging was removed in the PR. If you tell me what I need to search I can post an extract.

Original PR: https://github.com/home-assistant/android/pull/1356/commits
Code is still the same in master.

Side note: I add a look at the implementation and this [comment](https://github.com/home-assistant/android/pull/1356/commits/fc1362c4bfef37f891c02475096ca35316f09e26#r573195504). And indeed, I can add `e` every time there is an input field by using 3 finger touch. 😅

jpelgrom:
## Summary

Fixes #2392 

Unfortunately the default `GestureDetector.OnGestureListener` doesn't provide information about the number of pointers and the direction of a swipe, so a custom class has been created based on the default to fix these issues. At least adding more swipe gestures will be easier in the future :)

I've also checked the frontend for an event to fire to open the quick bar and this still doesn't appear to exist.

Swiping while opened doesn't switch between quick bar modes as mentioned in the documentation and there doesn't seem to be a reliable way to make this work, so I've also opened a PR to update the documentation to reflect what the gesture actually does.

## Screenshots

https://user-images.githubusercontent.com/8148535/159681345-0efb95d7-de16-4477-820d-1d14eb3723ce.mp4

Documentation: home-assistant/companion.home-assistant#719

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/util/OnSwipeListener.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/util/OnSwipeListener.kt
@@ -19,8 +19,8 @@ abstract class OnSwipeListener : View.OnTouchListener {
     var downEvent: MotionEvent? = null
     var numberOfPointers = 0
 
-    var minimumFlingVelocity: Int? = null
-    var maximumFlingVelocity: Int? = null
+    var minimumFlingVelocity = ViewConfiguration.getMinimumFlingVelocity()
+    var maximumFlingVelocity = ViewConfiguration.getMaximumFlingVelocity()
 
     override fun onTouch(v: View?, event: MotionEvent?): Boolean {
         event?.let { motionEvent ->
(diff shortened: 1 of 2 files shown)


dshokouhi (approved):
Tested the debug APK and it does work. Thank you for picking this up 🙏 

One minor comment is that I noticed we do need to drag our fingers I guess a bit longer than what some users may be already used to so not sure if we should mention that or make the drag distance shorter? This feature has been out for quite some time and even if it wasn't working 100% as described we may have users who think its broken as they now need to swipe longer.

dshokouhi:
ok I see what you mean, so we can't control it too much. Then I think its best we don't change this too much. Currently what we have implemented is really a workaround.  Once https://github.com/home-assistant/frontend/discussions/11511 is implemented we can probably begin to remove this workaround as it wouldn't be a hidden a feature.

jpelgrom:
With enough work we can control it however we want, but I think this works well enough. If we trigger the quick bar while still dragging, you still have to remove you fingers from the screen to start using the keyboard (probably, unless you're using an external keyboard but then you can press the keyboard shortcut directly), so there's no real benefit in triggering it slightly earlier. 

I actually didn't know about the upcoming frontend redesign, that looks really nice!

Mincka:
Thanks @jpelgrom for fixing this issue for this feature that "worked like designed".
I tested it and I can confirm a swipe down is now required.

Personally, I think that the minimum swipe distance is fine but the delay to do it felt a little bit short.
I needed a few tries to understand that the swipe needed to be done quite "fast" to trigger the action when the swipe distance is short. 
Is it something that could be improved by decreasing the required velocity a little bit?

Anyway, at least it's not triggering anymore on 3 finger touch, that's the main point, thanks again.

jpelgrom:
> I needed a few tries to understand that the swipe needed to be done quite "fast" to trigger the action when the swipe distance is short.
Is it something that could be improved by decreasing the required velocity a little bit?

Yes it probably can be improved by doing that. I've lowered the minimum velocity to 150, as quick three finger touches sometimes show up for me as swipes with a velocity of ~90-110 px/s. Could you try the latest build and see if a short swipe feels better?

The other cause might be that the app isn't detecting the swipe direction correctly, because technically it's only based on the first finger on the screen and not all three fingers, and when swiping a small distance it is easier to go slightly too far to the left/right. Fixing that is more difficult as it'd probably involve tracking all fingers individually and calculating the average angle, I'd rather not go there unless absolutely necessary.

jpelgrom:
I've changed the implementation for detecting swipe direction because previously it used the first/last pointer to determine the direction which wasn't always the same pointer (see above). It is now based on the average x and y velocity of _all_ pointers.

Direction detection now works a lot better for small swipes, and I haven't noticed any major difference for larger swipes.

Mincka:
The result was already perfect for me in 957a121a2f75393cd550c4d76f2a315012fc9065 with 100% success on swipe down attempts and 0% false positive on touches or swipes in other directions.

JBassett (approved):

PR #2397, opened 2022-03-23
Add missing volume streams to sensor and command

dshokouhi:
## Summary

Noticed that we were missing some of the volume streams, not all devices show to the user all the streams so better to cover all our bases. Samsung devices for example show System volume while others do not. All are still controllable and report data.

Documentation: home-assistant/companion.home-assistant#720


jpelgrom:
If the point is to support all possible audio streams, shouldn't there also be support for `STREAM_ACCESSIBILITY` and `STREAM_DTMF`?

dshokouhi:
Looks like we are not allowed to change the volume of accessibility streams 🙃 

image

dshokouhi:
Ok added in Accessibility and DTMF volume sensors, also added 1 missing audio mode (the other 2 we can add once we target the next version of android). Also added volume level control for DTMF.

JBassett (approved):

PR #2410, opened 2022-03-26
Use suspend calls for widget daos

NotWoods:
## Summary
Updating the various widget Daos to remove blocking operations, and instead use coroutines. Also tried to make updates so the code is easier to follow.

- Flow and List aren't nullable, so I removed the `?` annotations.
- Updated Compose views so one function handles displaying widget rows for each list

## Any other notes
- https://developer.android.com/reference/androidx/room/Query


dshokouhi:
Were you able to test the upgrade path from `master` to this branch to ensure the app is still functional post upgrade?

NotWoods:
I'll take a look. Since the DB queries weren't modified I don't expect upgrade issues.

NotWoods:
As far as I can tell the app is functional after upgrades. Created widgets using master then switched to this branch.

dshokouhi:
thank you for checking 🙏

JBassett (approved):

PR #2411, opened 2022-03-28
Add checks to verify DB migrations are actually done.

JBassett:
## Summary

With any luck this will fail builds if migrations are missing or incorrectly updating existing migrations.

## Screenshots

N/A

N/A

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -96,3 +96,9 @@ jobs:
       uses: kittaakos/upload-artifact-as-is@v0
       with:
         path: ./**/*.apk
+
+    - name: Check for modified DB migrations
+      run: if git diff --name-only --diff-filter=DMR origin/master... | grep "^common/schemas"; then exit 1; fi
+
+    - name: Check for missing DB migrations
+      run: test -n "$(git status --porcelain)"
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -101,4 +101,4 @@ jobs:
       run: if git diff --name-only --diff-filter=DMR origin/master... | grep "^common/schemas"; then exit 1; fi
 
     - name: Check for missing DB migrations
-      run: test -n "$(git status --porcelain)"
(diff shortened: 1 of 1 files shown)


jpelgrom:
Wouldn't it be better to run these checks as soon as possible (modified almost immediately, missing after building)? I can't think of a reason to build or archive the app if the action is going to fail because of changed or missing DB schema files.

JBassett:
So we at least need the second one after the build to check to see if a change was made to code that would affect the schema.  It would manifest as an uncommitted change.  And since that one needs to be done after a  build I just put both after the build.

.github/workflows/pr.yml
```diff
@@ -96,3 +96,9 @@ jobs:
       uses: kittaakos/upload-artifact-as-is@v0
       with:
         path: ./**/*.apk
+
+    - name: Check for modified DB migrations
+      run: if git diff --name-only --diff-filter=DMR origin/master... | grep "^common/schemas"; then exit 1; fi
+
+    - name: Check for missing DB migrations
+      run: test -n "$(git status --porcelain)"
```
jpelgrom, line 104: This check isn't working as intended, see this test PR for example output: https://github.com/jpelgrom/home-assistant-android/pull/2. I think you need to use the output to exit with a code other than 0 to fail the step/workflow.

JBassett: Changed around the command, looks to be working now.

jpelgrom: Looks like I commented on the wrong line, changed files is now working but in the step named 'Check for _missing_ DB migrations', the original check for changed files still doesn't seem to work correctly

jpelgrom:
Another scenario which isn't working: bumped the database version but didn't add the schema file to git. Example PR: https://github.com/jpelgrom/home-assistant-android/pull/4

The existing checks only seem to consider changes that are staged, but after Gradle is done building changes / new files aren't staged

jpelgrom:
@JBassett Is it OK if I have a go at fixing/implementing this? There haven't been a lot of database changes recently but this is still nice to have

PR #2412, opened 2022-03-28
We should throw less exceptions when possible.

JBassett:
## Summary
Noticed some crashes being reported for "crashes" when we should handle them more gracefully.  Go full Kotlin on things and avoid throwing in most cases.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/template/TemplateWidget.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/template/TemplateWidget.kt
@@ -65,7 +65,7 @@ class TemplateWidget : BaseWidgetProvider() {
             if (widget != null) {
                 var renderedTemplate = templateWidgetDao.get(appWidgetId)?.lastUpdate ?: "Loading"
                 try {
-                    renderedTemplate = integrationUseCase.renderTemplate(widget.template, mapOf())
+                    renderedTemplate = integrationUseCase.renderTemplate(widget.template, mapOf()) ?: "Error"
                     templateWidgetDao.updateTemplateWidgetLastUpdate(appWidgetId, renderedTemplate)
                     setViewVisibility(R.id.widgetTemplateError, View.GONE)
                 } catch (e: Exception) {
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
@@ -9,7 +9,6 @@ import com.fasterxml.jackson.module.kotlin.convertValue
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import io.homeassistant.companion.android.common.data.authentication.AuthenticationRepository
-import io.homeassistant.companion.android.common.data.authentication.AuthorizationException
 import io.homeassistant.companion.android.common.data.integration.ServiceData
 import io.homeassistant.companion.android.common.data.integration.impl.entities.EntityResponse
 import io.homeassistant.companion.android.common.data.url.UrlRepository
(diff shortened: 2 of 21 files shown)


app/src/main/java/io/homeassistant/companion/android/widgets/template/TemplateWidget.kt
```diff
@@ -65,7 +65,7 @@ class TemplateWidget : BaseWidgetProvider() {
             if (widget != null) {
                 var renderedTemplate = templateWidgetDao.get(appWidgetId)?.lastUpdate ?: "Loading"
                 try {
-                    renderedTemplate = integrationUseCase.renderTemplate(widget.template, mapOf())
+                    renderedTemplate = integrationUseCase.renderTemplate(widget.template, mapOf()) ?: "Error"
```
jpelgrom, line 68: Hardcoded "Error" string, you should use a string resource (or even better, in case of `null` use the error view that is used for exceptions, to allow the latest successfully rendered template to remain visible).

wear/src/main/java/io/homeassistant/companion/android/tiles/TemplateTile.kt
```diff
@@ -72,7 +72,7 @@ class TemplateTile : TileService() {
                     Timeline.Builder().addTimelineEntry(
                         TimelineEntry.Builder().setLayout(
                             Layout.Builder().setRoot(
-                                layout(renderedText)
+                                layout(renderedText.toString())
```
jpelgrom, line 75: This will result in the text "null" on the tile if rendering the template failed (because `integrationUseCase.renderTemplate` will return `null` on failure and `toString()` on that will give the text "null").
Lines 60-64 should probably be changed to also use the error string when the result is `null`.

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
```diff
@@ -334,7 +328,7 @@ class WebSocketRepositoryImpl @Inject constructor(
             connected.complete(true)
         } else {
             connectionState = WebSocketState.CLOSED_AUTH
-            connected.completeExceptionally(AuthorizationException())
+            connected.complete(false)
```
jpelgrom, line 331: This will cause problems for future attempts to send a message to the websocket because of the check at the start of the `connect()` function: the completable is now completed succesfully, and the app will assume there is a connection it can use.
`completeExceptionally` shouldn't _throw_ an exception, it just cancels the job, so I don't think this change is necessary. If you'd still like to change it, the check at the start of `connect()` should be updated to only return true if the job is completed, not cancelled, and the completion result was `true`.

JBassett: Fixed

wear/src/main/java/io/homeassistant/companion/android/tiles/TemplateTile.kt
JBassett: Done

app/src/main/java/io/homeassistant/companion/android/widgets/template/TemplateWidget.kt
JBassett: If we don't get a response then we should show the error screen.  Fixed.

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
```diff
@@ -258,7 +257,7 @@ class WebSocketRepositoryImpl @Inject constructor(
 
     private suspend fun connect(): Boolean {
         connectedMutex.withLock {
-            if (connection != null && connected.isCompleted) {
+            if (connection != null && connected.isCompleted && connected.await()) {
```
jpelgrom, line 260: Just making sure this is intentional: because `connected.await()` returns the result, if it is `false` this block will be skipped. With this change the app will immediately continue trying to set up a new connection, instead of the current behavior where it returns and the connection has to be set to null first (in `handleClosingSocket`). Is that the intended effect?

app/src/main/java/io/homeassistant/companion/android/widgets/template/TemplateWidget.kt
JBassett: Resolved upstream already

jpelgrom:
Considering how old this PR is and the number of merge conflicts should this be closed? Anything still surfacing recently and frequently as a crash is probably better handled in an up-to-date, smaller PR.

PR #2442, opened 2022-04-09
Breaking change: update user agent and append to webview UA

### LINKED ISSUE #2438: Home assistant web card shows "Graphics is not supported" when using Google charts
Home Assistant Android version: 2022.3.0-full
Android version: 12, SP2A.220405.004
Phone model: Pixel 6 Pro (also an issue on Samsung Z Fold 3, Pixel 4 XL)
Home Assistant version: 2022.4.0
Last working Home Assistant release (if known): Been broken for a few months, not sure which version it last worked
Description of problem:

Web card with a webpage using google charts does not work correctly in Android app (regression) but used to work in past releases (sometime in 2021). Web card works correctly when viewing Home Assistant from Chrome browser (desktop and android).

Screenshot of problem on Home Assistant app
Screenshot_20220407-160239_Home Assistant

Screenshot of it working from Chrome Browser using web version
Screenshot_20220407-160303_Chrome

jpelgrom:
## Summary

This PR updates the user agent used by the app to improve compatibility with sites that use user agent sniffing and to be more consistent, both for the different modules and with the iOS app. Because some users expect a certain user agent string from the app, **this is a breaking change**. Fixes #2438.

The app's main user agent is updated to `Home Assistant/{app version} (Android {Android version}; {device model})`: the same string for okhttp calls and the webview instead of different app versions, and a format that better matches 'normal' browsers and [the iOS app](https://github.com/home-assistant/iOS/blob/cd6534b12ce3e1cd30889cd2fcf74cf635ce80b5/Sources/Shared/API/HAAPI.swift#L58) (but not exactly the same as iOS to limit what information is exposed). Example output:
```
Home Assistant/LOCAL-1 (Android 12; Pixel 4a)
```

(I'm assuming `LOCAL-1` is replaced by something like `2022.4.0-2300` when a release is built?)

When using the webview, the app's user agent is added to the end of the original webview user agent to improve compatibility with sites that use user agent sniffing to enable access to features, but also still allow identifying the app. Example output:
```
Mozilla/5.0 (Linux; Android 12; Pixel 4a Build/SP2A.220405.003; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/100.0.4896.79 Mobile Safari/537.36 Home Assistant/LOCAL-1 (Android 12; Pixel 4a)
```

## Screenshots

n/a

n/a

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
@@ -22,8 +22,8 @@ import androidx.fragment.app.Fragment
 import androidx.fragment.app.activityViewModels
 import com.google.android.material.composethemeadapter.MdcTheme
 import dagger.hilt.android.AndroidEntryPoint
-import io.homeassistant.companion.android.BuildConfig
 import io.homeassistant.companion.android.R
+import io.homeassistant.companion.android.common.data.HomeAssistantApis
 import io.homeassistant.companion.android.common.data.authentication.impl.AuthenticationService
 import io.homeassistant.companion.android.onboarding.OnboardingViewModel
 import io.homeassistant.companion.android.onboarding.integration.MobileAppIntegrationFragment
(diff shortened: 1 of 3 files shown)


dshokouhi:
> (I'm assuming `LOCAL-1` is replaced by something like `2022.4.0-2300` when a release is built?)

yup thats correct :)

JBassett (approved):

Issue #2323, opened 2022-02-26
Turn on/off the persistent websocket connection via a notification command.


eyager1:
**Is your feature request related to a problem? Please describe.**

I like the speed of the persistent websocket connection, but the battery drain is a bit excessive when it comes to standby battery drain, especially when I'm at home and I don't need instant notifications.  But if I'm away from home I would like the instant notifications and am willing to accept the higher battery drain.   Right now the only solution is to manually change it in the app configuration.  It would be nice to be able to automate this.

**Describe the solution you'd like**

The simplest way to allow the persistent websocket connection to be automated to use a notification command and let home assistant handle the automation conditions.   Create something like `command_persistent_websocket`.   

**Describe alternatives you've considered, if any**

A tick-able list of conditions in the app configuration could be helpful as well.  A persistent websocket connection based on conditions like "When connected to charger", "When not in Home zone" or "When connected to wifi" could handle many of the most common use cases.

**Additional context**

dshokouhi:
> A tick-able list of conditions in the app configuration could be helpful as well. A persistent websocket connection based on conditions like \"When connected to charger\", \"When not in Home zone\" or \"When connected to wifi\" could handle many of the most common use cases.

So we did just add when on home wifi as a new option. I think if we continue to add more options this page will become more confusing. It might be best to just add a notification command to account for all the other cases.

PR #2445, opened 2022-04-10
Add persistent websocket connection notification command

NickM-27:
## Summary

Ability to toggle persistent connection mode using notification commands
Fix for #2323 

Documentation: home-assistant/companion.home-assistant#732

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -56,12 +56,14 @@ import io.homeassistant.companion.android.common.util.generalChannel
 import io.homeassistant.companion.android.common.util.getActiveNotification
 import io.homeassistant.companion.android.database.AppDatabase
 import io.homeassistant.companion.android.database.notification.NotificationItem
+import io.homeassistant.companion.android.database.settings.WebsocketSetting
 import io.homeassistant.companion.android.sensors.BluetoothSensorManager
 import io.homeassistant.companion.android.sensors.LocationSensorManager
 import io.homeassistant.companion.android.sensors.NotificationSensorManager
 import io.homeassistant.companion.android.sensors.SensorWorker
 import io.homeassistant.companion.android.settings.SettingsActivity
 import io.homeassistant.companion.android.util.UrlHandler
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -432,6 +435,19 @@ class MessagingManager @Inject constructor(
                             }
                         }
                     }
+                    COMMAND_PERSISTENT_CONNECTION -> {
+                        if (!jsonData[PERSISTENT].isNullOrEmpty()) {
```
dshokouhi, line 439: Lets add 1 more validation step to ensure the user is only entering accepted values, otherwise we should also post the notification to the device as the data is still not valid

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -1745,6 +1764,37 @@ class MessagingManager @Inject constructor(
...
+                    settingsDao.update(it)
+                }
+            }
+            WebsocketSetting.SCREEN_ON.name -> {
+                settingsDao.get(0)?.let {
+                    it.websocketSetting = WebsocketSetting.SCREEN_ON
+                    settingsDao.update(it)
+                }
+            }
+        }
```
dshokouhi, line 1795: We should also call `WebsocketManager.start(context)` so the service can stop/start immediately, just like we do on the settings page

dshokouhi (changes requested):
Just a couple of comments to make sure the service is updated immediately and also that we validate the values like we do with the other commands.

dshokouhi (approved):

jpelgrom:
Because this is a notification command, shouldn't home-assistant/mobile-apps-fcm-push also be updated to exclude notifications with this command from the rate limits?

dshokouhi:
> Because this is a notification command, shouldn't home-assistant/mobile-apps-fcm-push also be updated to exclude notifications with this command from the rate limits?

yup good catch, totally forgot about that :(

@NickM-27 can you create a PR there and link it here please?

NickM-27:
Created: https://github.com/home-assistant/mobile-apps-fcm-push/pull/71

JBassett (approved):

PR #2450, opened 2022-04-12
Don't block UI thread when opening tile + shortcuts settings

jpelgrom:
## Summary

Did you ever notice that opening the tile and shortcuts settings takes a little longer than other settings fragments? Turns out it is because the app was loading all the icons for the icon picker on the main thread, every time the fragment resumed. This PR fixes that by moving it to a background thread and only doing it once.

(also cleaning up some annotations for the shortcuts fragment)

## Screenshots

n/a

n/a

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragmentFactory.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragmentFactory.kt
@@ -1,5 +1,6 @@
 package io.homeassistant.companion.android.settings
 
+import android.annotation.SuppressLint
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentFactory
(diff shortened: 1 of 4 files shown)


dshokouhi (approved):

JBassett (approved):
This brings up the question, are we doing this on widgets and notifications?

jpelgrom:
> This brings up the question, are we doing this on widgets and notifications?

Notifications icons with `mdi:` don't use the icon picker resources because no picker is needed, instead it uses the iconics library which seems to perform much better (probably because it doesn't need to load everything, only look up by ID).

The button widget does use the icon picker and loads icons on the main thread (at least it is only loaded once). It should probably be improved but I didn't think about the actual widget itself, and the configuration activity includes more blocking calls in `onCreate` so I think it is better to make the icon dialog in the activity part of a larger refactoring PR and/or conversion to Compose.

PR #2469, opened 2022-04-22
Use enum for sensor setting type

NotWoods:
## Summary
Moving off a string type to a strict enum, since there are only certain valid values. Added a type converter to Room so this should be backwards compatible.


JBassett (approved):
Just need to solve the merge conflicts then it looks good.

dshokouhi:
I think this PR may have changed how sensor settings behave in the app, now every setting is a text field instead of being things like a toggle or a list or even a numeric field.  For toggles we now have to type `true` or `false`, list fields no longer show a list and numeric fields no longer restrict ot numeric keypad.

Issue #900, opened 2020-09-08
Bluetooth Sensor: Report nearby BLE Beacons


fti7:
**Describe the solution you'd like**
It would be very cool if the App could additionally report nearby BLE Beacons (Eddystone)
Since they will never appear in "unpaired devices"

https://developers.google.com/nearby/messages/android/get-beacon-messages

Would be very nice for Room/Presence Tracking.
Just Report the Eddystone UID + Signal/Distance of all Recieved ones

Hypfer:
Adding other beacons such as those used for contact tracing might also be an idea.

Since those change all the time, the only useful metric would probably be to have a total count of unique IDs around you

PerJarlemark:
I totally think this would take room tracking to a new level, by adding a static tracker device like an simple ESP32 to announce a static MAC or UUID to a specific room the Companion app could pick it up and announce the presence by its BLE and RSSI signal strength to Home Assistant,

Alfiegerner:
Another option, rather than sensing BLE messages, send them instead, and let roomassistant or ESP32 mqtt room devices pick up the room based location of the phone.

This has the benefit of keeping the Companion code quite dumb, i.e. every second send a BLE message, vs attempting to track distances via rssi of various local BLE beacons. 

There are a couple of Android apps that do this, e.g. Beacon Simulator and Beaconscope, but neither persist with Bluetooth or phone restarts. Beacon simulator is open source and on GitHub (mobile here so won't link sorry).

PerJarlemark:
@Alfiegerner could be an light option but would require extra hardware in your rooms, 
if it could detect existion devices (speakers/google home etc) you alredy have would make it more userfriendly.

Alfiegerner:
Hi @PerJarlemark 

Thanks for your consideration of this. I think lots of users of room-assistant and esp32-mqtt-room, and in general people who wanted to keep separate device tracking system so other items like bags, cars etc can be tracked would like this.

I'm mindful of not derailing this issue though - let me know if you'd like to create a new one.

Thanks!

dshokouhi:
Are there any new docs from Google on this? The linked docs seem to point to this feature being deprecated.

> Warning: The Google beacon platform is deprecated as of December 7, 2020. The platform will shut down on April 1, 2021. After this shut down, beacons will no longer be supported within the Nearby Messages API.

PerJarlemark:
@dshokouhi okey, didn´t know, but think it still can be build in to HA app

[...]

PR #1759, opened 2021-10-08
Breaking Change: Switch notification command parameters

dshokouhi:
## Summary

This PR addresses the issue of notification commands hijacking other notification parameters which were used for standard notifications. Now commands will use a unique parameter. There is really no way around this breaking change but given that its a parameter name change it should be relatively simple for all to change.

The biggest change is going to be that `title` will no longer be used for any command including TTS, this means all additional command data will need to be in the `data` element underneath `message`. In most cases `title` was switched to `command`. Providing examples below and in the docs to remove any confusion.

TTS

change `title` to `tts_text`
change `channel` to `media_stream`

```yaml
automation:
  - alias: "Fire Detected TTS alarm"
    trigger:
      ...
    action:
      - service: notify.mobile_app_<your_device_id_here>
        data:
          message: TTS
          data:
            ttl: 0
            priority: high
            media_stream: "alarm_stream"
            tts_text: "The house is on fire and the cat's stuck in the dryer!"
```

Device Commands

For DND, Ringer Mode, bluetooth, ble transmitter, high accuracy, webview

Change `title` to `command` 

```yaml
automation:
  - alias: Command ringer mode
    trigger:
      ...
    action:
      - service: notify.mobile_app_<your_device_id_here>
        data:
          message: "command_ringer_mode"
          data:
            command: "vibrate"
```

Broadcast Intent

change `channel` to `intent_package_name`
change `title` to `intent_action`
change `group` to `intent_extras`

Volume Level

change `title` to `command`
change `channel` to `media_stream`

Media

change `title` to `media_command`
change `channel` to `media_package_name`

Activity

change `channel` to `intent_package_name`
change `tag` to `intent_action`
change `title` to `intent_uri`
change `subject` to `intent_type`
change `group` to `intent_extras`

## Link to pull request in Documentation repository

Documentation: home-assistant/companion.home-assistant#589

## Any other notes

FCM: https://github.com/home-assistant/mobile-apps-fcm-push/pull/61

Leaving as a draft so we can merge this for the next release.  I don't think it makes sense to do backwards compatibility here as most of the changes are straight forward

PR #2472, opened 2022-04-24
Add Beacon Monitor

Adi146:
## Summary

This Pull Request adds a sensor for monitoring iBeacon. It will add a sensor which shows the distance to the closest beacon. There are all beacons in range listed in the attributes.

## Screenshots

Screenshot_20220424-204800
Screenshot_20220424-204806

Documentation: home-assistant/companion.home-assistant#736

## Any other notes

Issue: https://github.com/home-assistant/android/issues/900


common/src/main/res/values/strings.xml
```diff
@@ -426,6 +427,7 @@
...
     <string name="sensor_description_battery_temperature">The current battery temperature</string>
     <string name="sensor_description_bluetooth_ble_emitter">Send BLE iBeacon with configured interval, used to track presence around house, e.g. together with roomassistant, esp32-mqtt-room or espresence projects.\n\nWarning: this can affect battery life, particularly if the \"Transmitter power\" setting is set to High or \"Advertise Mode\" is set to Low latency.\nSettings allow for specifying:\n- \"UUID\" (standard UUID format), \"Major\" and \"Minor\" (should be 0 - 65535), to tailor identifiers and groups\n- \"Transmitter Power\" and \"Advertise Mode\" to help to preserve battery life (use lowest values if possible)#n - \"Measured Power\" to specify power measured at 1m (initial default -59)\n\nNote:\nAdditionally a separate setting exists (\"Enable Transmitter\") to stop or start transmitting.</string>
+    <string name="sensor_description_bluetooth_ble_beacon_distance">Distance to the nearest iBeacon, used to track presence around house, e.g. together with ESPHome beacons</string>
```
dshokouhi, line 430: We like to use the description as a place to describe the various settings a sensor could have. Could you write up about each setting offered making sure to mention units if applicable? Also if this feature will have a big impact on battery its also good to mention that here so users are aware of the implications.

dshokouhi, line 430: We like to use the description as a place to describe the various settings a sensor could have. Could you write up about each setting offered making sure to mention units if applicable? Also if this feature will have a big impact on battery its also good to mention that here so users are aware of the implications.

Adi146: I updated the description, let me know if something is not explained good enough or I missed something

dshokouhi:
Started taking a look at the debug APK today.  Just a couple of questions so far based on usability.

- What should the state of the sensor be when it does not detect any beacons within range? Like if we enable the sensor and are not near any beacons.
- Should the state of the sensor reflect the state of the monitor (setting) being enabled? The reason I say this is because users will want to know if the monitor is enabled when they want to control it via notification command. We do something similar for the beacon transmitter today. I think maybe combined with the first question we can probably introduce states like `not monitoring` or `monitoring` when the state is not the closest distance?

Adi146:
At the moment the sensor shows 0.0m if there is no beacon nearby. If the monitor is turned off it simply stops updating and keeps the last state. I agree, we should probably add some indication if it is monitoring or not but I don't think we should do that with the state because this would break the unit. What do you think about adding another attribute, like `state: enabled` or `state: disabled`?

dshokouhi:
> At the moment the sensor shows 0.0m if there is no beacon nearby.

About how long would you say it would take for this to show up? 2 minutes? If I enable the sensor after a fresh install and I do not have any beacons nearby the state of the sensor shows as enabled and I never see the sensor being registered or data being sent to HA for it.  Looking at the code it looks like it only sends an update if it has found a beacon before or if one expired but not if it didn't find anything at all on the first try.  Is that correct or did I misinterpret when we send updates?



> I agree, we should probably add some indication if it is monitoring or not but I don't think we should do that with the state because this would break the unit. What do you think about adding another attribute, like `state: enabled` or `state: disabled`?

I think that is fine as it still describes the state of the sensor, so users should expect the state to be 0.0m if monitoring is disabled correct? 

One more question :) Do we also need to account for bluetooth being turned off? Or is that a none issue? I guess I need to find a beacon to test this out better 😄 I ask because in the beacon transmitter sensor we do need to account for it.

Adi146:
> About how long would you say it would take for this to show up? 2 minutes? If I enable the sensor after a fresh install and I do not have any beacons nearby the state of the sensor shows as enabled and I never see the sensor being registered or data being sent to HA for it. Looking at the code it looks like it only sends an update if it has found a beacon before or if one expired but not if it didn't find anything at all on the first try. Is that correct or did I misinterpret when we send updates?

You got the sensor updates correctly and thats a case I didn't consider. I will have to add a sensor update on enable/disable for the state attribute anyways so this should fix it.

> I think that is fine as it still describes the state of the sensor, so users should expect the state to be 0.0m if monitoring is disabled correct?

Yes, if the sensor is disabled the state will be 0.0m and attribute `state` will be `disabled`. If there is no beacon in range the state will also be 0.0m but the attribute `state`will be `enabled`

> One more question :) Do we also need to account for bluetooth being turned off? Or is that a none issue?

Bluetooth should be enabled automatically when a scan period starts so this should be no big deal.

I hope, I can implement these changes later today.

dshokouhi:
> Yes, if the sensor is disabled the state will be 0.0m and attribute `state` will be `disabled`. If there is no beacon in range the state will also be 0.0m but the attribute `state`will be `enabled`

So in this particular case where the state value has not changed, the attribute will not update either. We only update sensors if the state has changed since the last update as attributes are supposed to describe the state of the sensor so they should not update when the state has not changed. Maybe its better if the state is just `0` when the monitor is disabled and `0.0` when its enabled so users can see it change if they enable it and are not near any beacons?

dshokouhi:
is there a beacon app I can use on another device to test with? I tried to use the apps transmitter sensor on a fire tablet but it doesnt seem to be picking it up on my Pixel in the same room but I am also not sure if its designed for that type of beacon :) I was thinking if we can use the apps own sensors it might be worth a write up to show it working between 2 devices using the app

madface303:
On android you can use beacon scope to send ibeacons

Adi146:
> > Yes, if the sensor is disabled the state will be 0.0m and attribute `state` will be `disabled`. If there is no beacon in range the state will also be 0.0m but the attribute `state`will be `enabled`
> 
> So in this particular case where the state value has not changed, the attribute will not update either. We only update sensors if the state has changed since the last update as attributes are supposed to describe the state of the sensor so they should not update when the state has not changed. Maybe its better if the state is just `0` when the monitor is disabled and `0.0` when its enabled so users can see it change if they enable it and are not near any beacons?

After all it seems like we need to account for if Bluetooth is on or off. What do you think of adding a second sensor which shows the state of the monitor (enabled, disabled, bluetooth off) like with it is implemented in the Beacon Transmitter?

madface303:
Just a question, is there anything special to do to detect a beacon? I have an esp32 beacon which is recognized by other scanners, but the debug apk i installed beside the normal app doesn't find it. The sensor sais monitoring but there are no attributes.
In the log i see an error about the permission, to test i gave the debug app every permission possible but the error persists.

```
04-29 20:46:47.258 23562 28427 I BluetoothAdapter: STATE_ON
04-29 20:46:47.258 23562 28427 D BluetoothLeScanner: could not find callback wrapper
04-29 20:46:48.255 23562 28427 I BluetoothAdapter: STATE_ON
04-29 20:46:48.255 23562 28427 D BluetoothLeScanner: could not find callback wrapper
04-29 20:46:48.258 23562 28427 I BluetoothAdapter: STATE_ON
04-29 20:46:48.259 23562 28427 I BluetoothAdapter: STATE_ON
04-29 20:46:48.259 23562 28427 I BluetoothAdapter: STATE_ON
04-29 20:46:48.260 23562 28427 D BluetoothLeScanner: Start Scan with callback
04-29 20:46:48.265 23562 28427 E CycledLeScannerForLollipop: Cannot start scan.  Security Exception: Need android.permission.BLUETOOTH_SCAN permission for AttributionSource { uid = 10393, packageName = io.homeassistant.companion.android.debug, attributionTag = null, token = android.os.BinderProxy@3bdb84a, next = null }: GattService registerScanner
```

I am running on a galaxy s21+ with samsung stock android 12

dshokouhi:
hmm I don't believe I noticed that permission error before but its possible that is why the BLE transmitter does not get picked up by other devices emitting it as I am on android 12 as well.

```
04-29 20:46:48.265 23562 28427 E CycledLeScannerForLollipop: Cannot start scan.  Security Exception: Need android.permission.BLUETOOTH_SCAN permission for AttributionSource { uid = 10393, packageName = io.homeassistant.companion.android.debug, attributionTag = null, token = android.os.BinderProxy@3bdb84a, next = null }: GattService registerScanner
```

This is a new permission in API level 31 so we probably just need to add it to our permission check as it is missing there: https://developer.android.com/reference/android/Manifest.permission#BLUETOOTH_SCAN

Adi146:
I added the permission, can you check again?

madface303:
Tried, but with this actual "action build" the sensor cannon be turned on. Nothing to see in the logs about the beacon scanner

Adi146:
> Tried, but with this actual "action build" the sensor cannon be turned on. Nothing to see in the logs about the beacon scanner

Thanks for the feedback. It should be fixed now, at least I could enable it with the emulator. I'm still using Android 11 on my phone so its a bit hard to test.

dshokouhi:
Working for me during initial testing. Good news too is that it is picking up the BLE transmitter beacon from the android app installed on other devices (wall mounted fire tablet 😃 ).

Great job!!  Will run some more tests and start looking at the code too.  

image

Lets make sure to update the docs to match the new state logic.

madface303:
Same for me, initial test is fine. Will do more practise tests tomorrow, but the permission problem seems to be solved!

madface303:
A short question how it is meant to be integrated into ha. I now have the sensor with 4 beacon attributes (only one of them is mine), i can easily make a template sensor of the attribute of my beacon, but when i leave this stays at the last distance submitted. From my opinion as a user it would be fine if we can a) filter our own beacons out directly in the app so that only my beacons are submitted to ha and b) the state of the sensor should be the distance again (and 0 if no beacon is in range) and the state of the monitor should go to the attributes. So even a user with a not so large experience (as me for example ;) ) can easily build an automation for being home/not_home. What are you thinking about?

Adi146:
> A short question how it is meant to be integrated into ha. I now have the sensor with 4 beacon attributes (only one of them is mine), i can easily make a template sensor of the attribute of my beacon, but when i leave this stays at the last distance submitted. From my opinion as a user it would be fine if we can

Here is my template sensor how I'm using it at the moment.
```
{% set beacon_uuid = states.sensor.smartphone_test_beacon_monitor.attributes | items | selectattr('0', 'in', '[bc59aec6-38b0-a4e9-58ce-ff107994091a,afe0fbaa-755e-64b5-1740-71570fa08335 2cc98e32-6ea7-e417-fb2f-baaa237ea789]') | sort(attribute=1) | first | first %}
{% if beacon_uuid == 'bc59aec6-38b0-a4e9-58ce-ff107994091a' or beacon_uuid == 'afe0fbaa-755e-64b5-1740-71570fa08335' %}
   Room1
{% elif beacon_uuid == '2cc98e32-6ea7-e417-fb2f-baaa237ea789' %}
  Bathroom
{% else %}
   Away
{% endif %}
```

I have three beacons of which two are in a bit larger room and one is in the bathroom. The template first filters for my beacons and then sorts them by the distance. After that it checks which beacon is the closest and reports the room I'm in.

>a) filter our own beacons out directly in the app so that only my beacons are submitted to ha

There are a lot of features which could be implemented later but for now I just want to get a basic sensor working. 

>b) the state of the sensor should be the distance again (and 0 if no beacon is in range) and the state of the monitor should go to the attributes. So even a user with a not so large experience (as me for example ;) ) can easily build an automation for being home/not_home. What are you thinking about?

I think there are a lot of use cases where you just want to check if a specific beacon is in range and in this case the state implementation as I had it before would be pretty useless. With the current implementation you could use a template like this

```
{% set distance = state_attr('sensor.smartphone_test_beacon_monitor', 'bc59aec6-38b0-a4e9-58ce-ff107994091a') %}
{% if distance != None %}
  in range
{% else %}
  out of range
{% endif %}
```

madface303:
This template would have taken me hours ;). Thanks a lot, for now i will create your sensor and do a long test run with it. But for my first experience it works every time i had a look at it, really good job. Another tool in the swif knife of home assistant companion!

app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
```diff
@@ -250,4 +327,29 @@ class BluetoothSensorManager : SensorManager {
             )
         )
     }
+
+    fun updateBeaconMonitoringSensor(context: Context) {
+        val icon = if (MonitoringManager.isMonitoring()) "mdi:bluetooth" else "mdi:bluetooth-off"
+
+        var state = if (!isBtOn(context)) "Bluetooth is turned off" else if (MonitoringManager.isMonitoring()) "Monitoring" else "Stopped"
```
dshokouhi, line 334: So I noticed that on a fresh install on android 12 if I dont enable the BT state sensor the state of this sensor remains at "Bluetooth is turned off"  eventhough BT is on for this device. As soon as I enable BT state sensor the monitor starts reporting beacons and the state updates too. I think we may need to change the check on `isBtOn()` to not be sensor specific like it is today. The BLE transmit sensor does not have this particular issue. Or we may just need to add more BT permissions for Android 12 devices to work around it.

Adi146: I changed it to the `BluetoothUtils.isOn(context)` function which seem to fix this case. However when I switch Bluetooth of in the emulator on Android 12 the sensor state keeps showing `Monitoring`. I'm not sure if this happens on real devices too. Could you please check this?

dshokouhi: latest debug APK works well on my pixel 6 pro :)  the state also updates immediately when I toggle BT on/off

Adi146:
@dshokouhi whats the state of the review? Are we ready to merge?

JBassett (changes requested):

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/MonitoringManager.kt
```diff
@@ -0,0 +1,68 @@
...
+            BeaconManager.setRssiFilterImplClass(KalmanFilter::class.java)
+        }
+        if (!beaconManager.isAnyConsumerBound) {
+            beaconManager.foregroundScanPeriod = scanPeriod
+            beaconManager.foregroundBetweenScanPeriod = scanInterval
+            beaconManager.backgroundScanPeriod = scanPeriod
+            beaconManager.backgroundBetweenScanPeriod = scanInterval
+
+            region = buildRegion()
+            GlobalScope.launch(Dispatchers.Main) {
```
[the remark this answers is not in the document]
Adi146: I'm not sure if this works because `observeForever`  has to run in the main thread. Correct me if I'm wrong.

common/src/main/java/io/homeassistant/companion/android/database/sensor/SensorDao.kt
```diff
@@ -72,6 +72,9 @@ interface SensorDao {
     @Query("UPDATE sensors SET last_sent_state = :state WHERE id = :sensorId")
     suspend fun updateLastSendState(sensorId: String, state: String)
 
+    @Query("UPDATE sensors SET last_sent_state = :state WHERE id = :sensorId")
+    fun updateLastSendStateSync(sensorId: String, state: String)
```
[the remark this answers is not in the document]
Adi146: We have to reset the `last_sent_state` before sending an sensor update, so that the attributes get updated in home-assistant.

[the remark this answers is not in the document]
JBassett: suspend functions are still run synchronously, they just don't block the thread they are on when waiting on I/O

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/MonitoringManager.kt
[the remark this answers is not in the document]
JBassett: You can still scope a coroutine that uses the main thread.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -365,6 +367,19 @@ class MessagingManager @Inject constructor(
                             }
                         }
                     }
+                    COMMAND_BEACON_MONITOR -> {
+                        if (!jsonData[TITLE].isNullOrEmpty() && jsonData[TITLE] in ENABLE_COMMANDS)
```
dshokouhi, line 371: Now that #1759 has been merged we will need to update this PR to switch from `TITLE` to `COMMAND`, we shoudl also update the documentation to reflect this change as well.

dshokouhi (changes requested):
We need to adjust the notification commands now that the breaking change PR was merged. Asides from this there is also the failing linter that needs to be resolved.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -377,6 +379,19 @@ class MessagingManager @Inject constructor(
...
```
dshokouhi, line 383: We also need to switch to `jsonData[COMMAND]` here as well, just like the other commands above/below :)

dshokouhi (changes requested):
The notification command was only partially updated :)

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
Adi146: ups I overlooked that

dshokouhi (approved):
I have tested the debug APK and its working as expected in my tests. Lets wait for another review to make sure the previous requested changes are also as expected.

dshokouhi:
@Adi146 just a heads up you have a conflict with a file here

app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
```diff
@@ -261,6 +342,33 @@ class BluetoothSensorManager : SensorManager {
         )
     }
 
+    fun updateBeaconMonitoringSensor(context: Context) {
```
dshokouhi, line 345: We should make sure none of the code below is reachable if the sensor is not enabled. May also want to do something similar with `updateBeaconMonitoringDevice()` up above. We do a similar check with other sensors so the code only runs if the sensor is enabled.

dshokouhi (changes requested):
Apologize for not catching this earlier, just a small change to make sure the code is only reachable if the sensor is not enabled.

app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
```diff
@@ -261,6 +342,33 @@ class BluetoothSensorManager : SensorManager {
...
+        if (BluetoothUtils.isOn(context) && monitoringManager.isMonitoring()) {
+            for (beacon: IBeacon in beaconMonitoringDevice.beacons) {
+                attr += Pair(beacon.uuid, beacon.distance)
+            }
+        }
+
+        // reset the last_sent_state of the sensor so it won't skip the update of attributes
+        val sensorDao = AppDatabase.getInstance(context).sensorDao()
+        runBlocking {
+            sensorDao.updateLastSentStateAndIcon(beaconMonitor.id, "", "")
```
jpelgrom, line 360: This function was updated in #2734 (which has been merged), please use `null` instead of `""` to force an update of the sensor

Yogranov:
Hey,
Im here to say i wanted this feature for a long time (literally started an background app project for temp workaround).
Ive downloaded this artifact APK, and it looks amazing! :D
It was little hard to make an automation with ranges (less than 10m), but it must merge to next version guys :P
Thanks @Adi146 !

Adi146:
I added the requested changes. It would be nice if we could finally merge this.

dshokouhi (approved):
All my requested changes have been met. We are planning on cutting a new release with existing beta features so this PR (given final review from others) will probably be included in the next beta cycle so it can be given a proper test cycle. We actually have other PRs we are holding off on as well.

nodnarb402:
I joined the beta track specifically for this feature. Thanks for working on this. Very much looking forward to it getting merged!

Issue #1206, opened 2020-11-19
Notification attributes (array/list) not expanded (android.textLines and android.people.list) / Stacked messages


Mincka:
**Home Assistant Android version:**
3.0.2-full

**Android version:**
10.0

**Phone model:**
Samsung S20+

**Home Assistant version:**
0.118

**Last working Home Assistant release (if known):**
n/a

**Description of problem:**
When you get multiple messages from a single app, they can be grouped together.
In this case, even if the messages are all visible in the notifications on the phone, the only captured information as `android.text` is something like "4 new messages". The text itself in then stored in `android.textLines` but this object needs to be converted to something that can be read.

Similar issue regarding `android.people.list` where it's not possible to use the attribute because the list is not expanded.

In my case, I play TTS (Google Cloud) on my Google Home when I get WhatsApp notifications.
However, if I don't open the WhatsApp conversations, the messages are stacked and I am only able to retrieve a generic message which is "4 new messages". I would prefer to be able to use the latest `android.textLines` or the `android.text` if the first one is empty / not present.

```
android.appInfo: ApplicationInfo{5767c9 com.whatsapp}
android.infoText: null
android.largeIcon: null
android.people.list: [android.app.Person@213eede0]
android.progress: 0
android.progressIndeterminate: false
android.progressMax: 0
android.reduced.images: true
android.remoteInputHistory: null
android.showChronometer: false
android.showWhen: true
android.subText: null
android.summaryText: ‎4 nouveaux messages
android.template: android.app.Notification$InboxStyle
android.text: ‎4 nouveaux messages
android.textLines: [Ljava.lang.CharSequence;@97a1bd0
android.title: John Doe
package: com.whatsapp
friendly_name: Téléphone Julien Dernière notification
icon: mdi:bell-ring
```

Maybe related:
https://stackoverflow.com/a/50868911
https://stackoverflow.com/a/49181208

Suggested implementation:
Since the attribute size is limited to 255 characters, it would not be possible to use the same field for all the lines. I think multiple attributes could be created based on the array size:
* android.textLines_0: Message 1
* android.textLines_1: Message 2
* android.people.list_0: John Doe
* android.people.list_1: Jane Doe

dshokouhi:
@Mincka can you try the latest beta with the new Active Notification sensor? I think it might provide each and every notification detail for you in the attributes.  I have seen that count jump when my whatsapp group is chatting away so I think it might be what your looking for :)

Mincka:
[...]

PR #2478, opened 2022-04-26
Breaking change: send sensor list attributes as list to server

### LINKED ISSUE #2474: Bluetooth sensor not showing correct lists
beta-2234-6c64b31f

**Android version:**
Android 12

**Phone model:**
Pixel 5

**Home Assistant version:**
2022.4.7

Unknown

**Description of problem:**
Bluetooth Connection sensor uses strings which look like a list, but are not, it would be nice if these were actually real lists

Current state in dev-tools > state
```yaml
connected_not_paired_devices: '[14:8F:21:FD:CE:BC]'
connected_paired_devices: '[E8:07:BF:13:99:AB]'
paired_devices: >-
  [00:7C:2D:12:42:A0, FA:2C:A7:D6:EB:41, 7C:B2:7D:3B:CD:4B, E0:67:81:01:F7:D3,
  C8:55:AF:52:29:20, 62:3B:4D:94:AE:8F, D6:DA:5C:AE:66:B3, E8:07:BF:13:99:AB,
  88:B4:36:44:B7:0B, 74:E5:F9:64:FA:8B, ED:22:91:7C:BE:DF, 04:4E:AF:C6:00:32,
  60:AB:D2:3C:85:66]
unit_of_measurement: connection(s)
icon: mdi:bluetooth
friendly_name: Pixel 5 Bluetooth Connection
```

I would expect:
```yaml
connected_not_paired_devices: 
  - 14:8F:21:FD:CE:BC
connected_paired_devices: 
  - E8:07:BF:13:99:AB
paired_devices:
  - 00:7C:2D:12:42:A0
  - FA:2C:A7:D6:EB:41
  - 7C:B2:7D:3B:CD:4B
  - E0:67:81:01:F7:D3
  - C8:55:AF:52:29:20
  - 62:3B:4D:94:AE:8F
  - D6:DA:5C:AE:66:B3
  - E8:07:BF:13:99:AB
  - 88:B4:36:44:B7:0B
  - 74:E5:F9:64:FA:8B
  - ED:22:91:7C:BE:DF
  - 04:4E:AF:C6:00:32
  - 60:AB:D2:3C:85:66
unit_of_measurement: connection(s)
icon: mdi:bluetooth
friendly_name: Pixel 5 Bluetooth Connection
```

```
```

**Screenshot of problem:**
image

**Additional information:**

jpelgrom:
## Summary

When the app collects sensor attributes, it is stored in the app's database as a string and converted back to it's original type when sent to the server, if possible. This PR updates the "if possible" part to also include lists. This will allow users to use array features in HA templating for these values.

Because this changes the format for attribute values, this is a **breaking change**. Based on a quick search, this will have changes for the following sensors:
 - Bluetooth Connection
 - Last Notification
 - Last Removed Notification
 - Active Notification Count

I've tested this with core 2022.4.7. Fixes #2474.

## Screenshots

n/a

n/a

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
@@ -129,12 +129,12 @@ class BluetoothSensorManager : SensorManager {
         var totalConnectedDevices = 0
         var connectedPairedDevices: List<String> = ArrayList()
         var connectedNotPairedDevices: List<String> = ArrayList()
-        var bondedString = ""
+        var pairedDevices: List<String> = ArrayList()
 
         if (checkPermission(context, bluetoothConnection.id)) {
 
             val bluetoothDevices = BluetoothUtils.getBluetoothDevices(context)
-            bondedString = bluetoothDevices.filter { b -> b.paired }.map { it.address }.toString()
+            pairedDevices = bluetoothDevices.filter { b -> b.paired }.map { it.address }
             connectedPairedDevices = bluetoothDevices.filter { b -> b.paired && b.connected }.map { it.address }
             connectedNotPairedDevices = bluetoothDevices.filter { b -> !b.paired && b.connected }.map { it.address }
             totalConnectedDevices = bluetoothDevices.filter { b -> b.connected }.count()
@@ -147,7 +147,7 @@ class BluetoothSensorManager : SensorManager {
             mapOf(
                 "connected_paired_devices" to connectedPairedDevices,
                 "connected_not_paired_devices" to connectedNotPairedDevices,
-                "paired_devices" to bondedString
+                "paired_devices" to pairedDevices
             )
         )
     }
--- a/app/src/main/java/io/homeassistant/companion/android/settings/sensor/views/SensorDetailView.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/sensor/views/SensorDetailView.kt
@@ -54,6 +54,8 @@ import androidx.compose.ui.text.input.KeyboardType
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.fasterxml.jackson.module.kotlin.readValue
 import com.mikepenz.iconics.IconicsDrawable
 import com.mikepenz.iconics.compose.Image
 import com.mikepenz.iconics.typeface.library.community.material.CommunityMaterial
@@ -76,6 +78,7 @@ fun SensorDetailView(
 ) {
     val context = LocalContext.current
     var sensorUpdateTypeInfo by remember { mutableStateOf(false) }
+    val jsonMapper by lazy { jacksonObjectMapper() }
 
     val sensorEnabled = viewModel.sensor?.sensor?.enabled
         ?: (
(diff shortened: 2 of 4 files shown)


dshokouhi:
> * Last Notification (technically, depends on extras but I haven't seen any lists in there)
> * Last Removed Notification (technically, depends on extras but I haven't seen any lists in there)

Not related to the changes here but on this subject we do have this long standing bug: https://github.com/home-assistant/android/issues/1206

I don't believe there are any other places other than what you have listed where this occurs

dshokouhi (approved):
Tested the debug APK and it checks out fine to me.

jpelgrom:
A concern raised [here](https://github.com/home-assistant/android/issues/1206#issuecomment-1111377153): the fallback to a string in case there is a ", " in the stringified list values may produce inconsistent attribute value types.

I'm not sure what the best solution is here;
 - drop this PR for consistency
 - continue with this PR and ask users to check the format themselves every time
- perhaps store data in the database as JSON/custom format which allows escaping values (so the ", " won't be a problem) but will increase complexity because we need to do more conversions

jpelgrom:
Updated to store lists as JSON in the database, which escapes any problematic characters, so now users can be sure that lists are always sent as lists. List values will always be booleans, ints, longs, floats or, if not one of the previous types, strings using `toString()`, for consistency with values outside lists.

I also looked into storing all other types as JSON, which might be interesting when #2482 is merged because that code will allow (nested) key/value maps as attributes. Unfortunately Jackson's serialization of some classes throws exceptions (for example, `Bitmap`) and I don't see a simple fix for that, so it is out of scope for this PR.

JBassett (approved):

JBassett (approved):

Mincka:
Can someone confirm this one works as expected for bluetooth connection or other sensors?
I get comma separated strings for bluetooth sensor or last_notification.

Hi,

I just checked in beta-2300-f33743dc published last week and now `android.textLines` is a string. It's not an array converted as a string, it just comma separated elements. I was expecting that this PR would convert the array do a list but it looks like that's not the case.

```
android.appInfo: ApplicationInfo{8fc5718 com.whatsapp}
android.infoText: null
android.largeIcon: null
android.people.list: android.app.Person@739669d8
android.progress: 0
android.progressIndeterminate: false
android.progressMax: 0
android.reduced.images: true
android.remoteInputHistory: null
android.showChronometer: false
android.showWhen: true
android.subText: null
android.summaryText: ‎7 nouveaux messages
android.template: android.app.Notification$InboxStyle
android.text: ‎7 nouveaux messages
android.textLines: Un, Deux, Trois, Quatre, Cinq, 📷 Photo, 🥰
android.title: Morgane Bardet
androidx.core.app.extra.COMPAT_TEMPLATE: androidx.core.app.NotificationCompat$InboxStyle
is_clearable: true
is_ongoing: false
package: com.whatsapp
post_time: 1652110784918
icon: mdi:bell-ring
friendly_name: Téléphone Julien Last Notification
```

jpelgrom:
I can confirm it is working correctly for me, tested just now with the Bluetooth Connection sensor on a build based on beta-2310. If they were still strings, I would expect to see `[]` around the values like you can see in the app.

Did you copy these values from the 'more info' panel or frontend dev tools states table? I think the frontend will only show lists as lists if you open the 'Set State' panel in the dev tools, otherwise it shows lists as a string with comma separated items.

Mincka:
Oh, indeed that's what I did. I can confirm it shows as a list using 'Set State'. Sorry about that and thanks again! I am going to have some fun with the for each loop now. 😛

PR #2488, opened 2022-04-28
Create a WidgetConfigureActivity base class

NotWoods:
## Summary
Adds a base class for the widget configure activities. As a result some logic can be shared - initially just the delete dialog.

Also updated the delete widget operation to use coroutines.


jpelgrom:
@NotWoods In case you hadn't noticed, ktlint doesn't like the code and there's also a merge conflict now

Aside from that this seems like a good improvement to me!

JBassett (approved):

Issue #2485, opened 2022-04-27
Add a sensor for high_accuracy_update_interval


Mincka:
**Is your feature request related to a problem? Please describe.**
We are now able to update the high_accuracy_update_interval remotely thanks to #2356. Thanks again for this, it's working great. :)
However, we don't know the current value so we can't tell if the value was modified locally or if the latest remote update failed.

**Describe the solution you'd like**
A dedicated sensor to get the current value of high_accuracy_update_interval 

**Describe alternatives you've considered, if any**
None. To be honest, I don't have a specific use case. The idea is just to have the ability to reflect the state of something that can be set either locally or remotely. For instance, we can enable or disable high accuracy with a command, and there is a sensor to return its state because it can be changed in different contexts. I can tell that's useful in this case because I have a retry logic in my automations if the mobile app does not return the updated value.

**Additional context**
From a UX point of view, I think it would be strange to have the high_accuracy_update_interval as a parameter of the "Background location" sensor AND a dedicated sensor that would reflect the same thing. If this sensor is implemented, maybe it would require a little bit a work around this.

Now there is a command, I think it would be better to move the parameter from "Background location" sensor because:
1. this attribute is hidden when background location is off
2. this attribute is disabled when high accuracy is off

But in both cases it can still be changed by the command remotely (which is great and useful). But I am a bit puzzled by the fact that the user cannot always see it. The parameter would be with its own sensor that could be either enabled or not like the others.

jpelgrom:
Reading this issue after looking at the linked PR, and I have some thoughts about "additional context".

Splitting high accuracy mode from background location could be a good idea to provide two distinct options (background location tracking and foreground location tracking) that are easier to understand for the user. This would reduce the number of visible settings and keep setting + state in one place, which might make it easier to understand.

At the same time, background location isn't actually a sensor but a device tracker, so it is already out of place. High accuracy mode also affects the background location. No visible attributes/settings in the app already happens for other commands (for example, persistent connection). Splitting the option creates even more questions: should we now provide two device trackers, how to handle attributes for trackers, ... 

I'm also thinking about future location tracking features, like only reporting zones instead of coordinates (like on iOS), or using another source like beacons or WiFi networks, which will raise similar requests and questions when commands are available to control their settings. Should those options be added to background location, or additional 'sensors'/device trackers? Should each option get additional sensors or should there be a 'location tracking' sensor with all information as attributes? 

[...]

PR #2490, opened 2022-04-28
Add a sensor for high accuracy update interval

dshokouhi:
## Summary

Fixes: #2485 the state will update as soon as the value changes which takes place when the service is started or restarted

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#737


jpelgrom:
Wouldn't it make more sense for this to be an attribute for the existing 'high accuracy mode' sensor? That sensor tells you if it's enabled or not, the update interval is a detail for that state.

dshokouhi:
> Wouldn't it make more sense for this to be an attribute for the existing 'high accuracy mode' sensor? That sensor tells you if it's enabled or not, the update interval is a detail for that state.

When high accuracy mode is enabled and the update interval changes we only call a restart and in this case the state will not change and the attribute value will not update. So I had opted to keep it separate to ensure the sensor is as up to date as possible.

JBassett (approved):

PR #2496, opened 2022-05-01
Add theming to widgets: Dynamic color (Material You) and Transparent

### LINKED ISSUE #1852: [Feature request] Android Widget Template font-size and background color
It would be great to implement possibility to change background color and trasparency of android Template Widget,
and also add the ability to change the font size with html-tags (like we can change color and style of the font (bold, etc) ).

This will create endless possibilities for changing the appearance, taking into account the existing html possibilities.

photo_2021-10-29_13-22-51

jpelgrom:
## Summary

Adds theming options to most widgets:
 - **Dynamic color**: uses colors from the device theme (eg. Material You) on supported Android 12+ devices. Default for new widgets on devices that support it.
 - **Light/dark theme**: the existing widget theme. Default for new widgets on devices that don't support dynamic color.
 - **Transparent**: removes the widget background, user can choose whether the content should be black or white.

The text size for template widget can now also be set.

More minor changes for Android 12 to improve the widget theme:

 - Album art now is correctly clipped by the widget corner radius
 - `android:previewLayout` is set to the dynamic color layout file, to provide dynamic previews and better match the widget that'll be added
 - Suggested size for entity state and template widgets is updated to 2x1, to match the preview images
 - When using dynamic color, the system widget corner radius is used

Resolves #1852: this PR adds options for background and font size, and additional formatting can be done using HTML tags.

Any widgets that are already added won't change with this PR, they are migrated to settings that match the existing widget theme. The camera widget doesn't get any theming options because you should only see a camera image most of the time. The media player widget doesn't get a transparent theme because floating text and buttons next to album art doesn't make sense.

## Screenshots

Theming options when creating a widget. Also shows the template widget text size option.
|Light|Dark|
|-----|-----|
|image|image|

When the 'Transparent' theme is selected, an additional option for the widget's text/icon color will be shown.
|Light|Dark|
|-----|-----|
|image|image|

Transparent widgets, from left to right: entity state, button, template
|'White' text color|'Black' text color|
|-----|-----|
|image|image|

Dynamic color widgets, a few examples. The colors used match widgets provided by Google apps. The corner radius for these widgets also matches the system value, which for this device is larger than the value normally used by the app. (Color varies slightly between widgets because it actually uses the widget location on the background!)
|Light|image|image|image|image|
|-----|-----|-----|-----|-----|
|Dark|image|image|image|image|

Launcher providing dynamic previews of widgets:
|Light|Dark|
|-----|-----|
|image|image|

Documentation: home-assistant/companion.home-assistant#740

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -184,7 +184,7 @@ open class HomeAssistantApplication : Application() {
                 IntentFilter(Intent.ACTION_TIME_TICK)
             )
 
-        // Update widgets when the screen turns on, updates are skipped if widgets were not added
+        // Update widgets when the screen turns on or configuration changes, updates are skipped if widgets were not added
         val buttonWidget = ButtonWidget()
         val entityWidget = EntityWidget()
         val mediaPlayerWidget = MediaPlayerControlsWidget()
@@ -193,6 +193,7 @@ open class HomeAssistantApplication : Application() {
         val screenIntentFilter = IntentFilter()
         screenIntentFilter.addAction(Intent.ACTION_SCREEN_ON)
         screenIntentFilter.addAction(Intent.ACTION_SCREEN_OFF)
+        screenIntentFilter.addAction(Intent.ACTION_CONFIGURATION_CHANGED)
 
         registerReceiver(buttonWidget, screenIntentFilter)
         registerReceiver(entityWidget, screenIntentFilter)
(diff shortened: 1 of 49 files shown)


JBassett (changes requested):
My only issue is the radio button alignment, it shouldn't just be right aligned.

jpelgrom:
> My only issue is the radio button alignment, it shouldn't just be right aligned.

Updated alignment for the text/icon color radio buttons to left aligned (for ltr locales), and switched to a dropdown menu for the theme selection because it didn't look right to me when changing the alignment.

JBassett (approved):

Issue #2499, opened 2022-05-02
BLE intermittent transmitter


ingrove:
**Home Assistant Android version:**
2022.3.0-full
**Android version:**
12
**Phone model:**
Google Pixel 5
**Home Assistant version:**
2022.4.7
**Last working Home Assistant release (if known):**

**Description of problem:**
The BLE beacon transmitter randomly stops/starts/disappears.
**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**
HA core log:
```
2022-05-02 14:45:00 DEBUG (MainThread) [homeassistant.components.mobile_app.webhook] Received webhook payload from Pixel 5 for type update_sensor_states: [{'unique_id': 'ble_emitter', 'state': '', 'type': 'sensor', 'icon': 'mdi:bluetooth-off', 'attributes': {'Advertise mode': 'lowLatency', 'Measured power': -59, 'Transmitting power': 'low', 'id': '008fd71e-a43a-4ab4-8e02-f969133f8380-100-1'}}]

2022-05-02 14:45:00 DEBUG (MainThread) [homeassistant.components.mobile_app.webhook] Received webhook payload from Pixel 5 for type update_sensor_states: [{'unique_id': 'ble_emitter', 'state': '', 'type': 'sensor', 'icon': 'mdi:bluetooth-off', 'attributes': {'Advertise mode': 'lowLatency', 'Measured power': -59, 'Transmitting power': 'low', 'id': '008fd71e-a43a-4ab4-8e02-f969133f8380-100-1'}}]

2022-05-02 14:45:59 DEBUG (MainThread) [homeassistant.components.mobile_app.webhook] Received webhook payload from Pixel 5 for type update_sensor_states: [{'unique_id': 'ble_emitter', 'state': 'Transmitting', 'type': 'sensor', 'icon': 'mdi:bluetooth', 'attributes': {'Advertise mode': 'lowLatency', 'Measured power': -59, 'Transmitting power': 'low', 'id': '008fd71e-a43a-4ab4-8e02-f969133f8380-100-1'}}]
```
Companion App log:
```
05-02 14:45:00.066 31838 31838 E mpanion.androi: Not starting debugger since process cannot load the jdwp agent.
05-02 14:45:00.068 31838 31838 D ProcessState: Binder ioctl to enable oneway spam detection failed: Invalid argument
[...]

PR #2500, opened 2022-05-03
BLE Transmit fix blank state and check device supports advertising

dshokouhi:
## Summary

Fixes: #2499 by sending the last state if the string is blank, the default will be `unknown` unless there is a better default to use? Either way we shouldn't send a blank string :)

Noticed one of my devices failed to start the transmitter and this is caused by the BT on the device not supporting advertising so adding a check for that.  Here is the failure I noticed:

```
[05-01 12:15:00.337 27750:28542 E/BeaconTransmitter]
Cannot start advertising due to exception
java.lang.NullPointerException: Attempt to invoke virtual method 'void android.bluetooth.le.BluetoothLeAdvertiser.startAdvertising(android.bluetooth.le.AdvertiseSettings, android.bluetooth.le.AdvertiseData, android.bluetooth.le.AdvertiseCallback)' on a null object reference
	at org.altbeacon.beacon.BeaconTransmitter.startAdvertising(BeaconTransmitter.java:213)
	at org.altbeacon.beacon.BeaconTransmitter.startAdvertising(BeaconTransmitter.java:159)
	at io.homeassistant.companion.android.bluetooth.ble.TransmitterManager.startTransmitting(TransmitterManager.kt:65)
	at io.homeassistant.companion.android.sensors.BluetoothSensorManager.updateBLESensor(BluetoothSensorManager.kt:236)
	at io.homeassistant.companion.android.sensors.BluetoothSensorManager.requestSensorUpdate(BluetoothSensorManager.kt:123)
	at io.homeassistant.companion.android.common.sensors.SensorManager$DefaultImpls.requestSensorUpdate(SensorManager.kt:94)
	at io.homeassistant.companion.android.sensors.BluetoothSensorManager.requestSensorUpdate(BluetoothSensorManager.kt:17)
	at io.homeassistant.companion.android.common.sensors.SensorReceiverBase.updateSensors(SensorReceiverBase.kt:144)
	at io.homeassistant.companion.android.common.sensors.SensorReceiverBase$onReceive$2.invokeSuspend(SensorReceiverBase.kt:106)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:749)
```

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
@@ -94,8 +94,7 @@ class BluetoothSensorManager : SensorManager {
     override val name: Int
         get() = commonR.string.sensor_name_bluetooth
     override fun getAvailableSensors(context: Context): List<SensorManager.BasicSensor> {
-        val list = listOf(bluetoothConnection, bluetoothState)
-        return if (supportsTransmitter(context)) list.plus(bleTransmitter) else list
+        return listOf(bluetoothConnection, bluetoothState, bleTransmitter)
     }
 
     override fun requiredPermissions(sensorId: String): Array<String> {
@@ -256,7 +255,8 @@ class BluetoothSensorManager : SensorManager {
                 "id" to bleTransmitterDevice.uuid + "-" + bleTransmitterDevice.major + "-" + bleTransmitterDevice.minor,
                 "Transmitting power" to bleTransmitterDevice.transmitPowerSetting,
                 "Advertise mode" to bleTransmitterDevice.advertiseModeSetting,
-                "Measured power" to bleTransmitterDevice.measuredPowerSetting
+                "Measured power" to bleTransmitterDevice.measuredPowerSetting,
+                "Supports transmitter" to supportsTransmitter(context)
             )
         )
     }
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
```diff
@@ -93,7 +94,8 @@ class BluetoothSensorManager : SensorManager {
     override val name: Int
         get() = commonR.string.sensor_name_bluetooth
     override fun getAvailableSensors(context: Context): List<SensorManager.BasicSensor> {
-        return listOf(bluetoothConnection, bluetoothState, bleTransmitter)
+        val list = listOf(bluetoothConnection, bluetoothState)
+        return if (supportsTransmitter(context)) list.plus(bleTransmitter) else list
```
jpelgrom, line 98: As a result of this check, the 'BLE Transmitter' sensor disappears on my device when Bluetooth is turned off. I don't think we should completely hide the sensor, perhaps a state of 'unavailable' or 'unsupported' for the sensor is better?

(also: you wouldn't be able to turn off the sensor now if it becomes unsupported)

dshokouhi: Hmm good point, maybe `unavailable` is the way to go then in these cases.

dshokouhi: So looks like this isn't a good check to add to see if the BLE transmitter is supported as it depends on BT being enabled. The current library actually catches this failure and also stops transmitting when an error is received so might be best to let the library to continue doing its thing. I will opt to add this as a attribute of the sensor to help users understand why the transmitter refuses to start because the sensor already accounts for notifying the user when BT is turned off.

JBassett (approved):

PR #2505, opened 2022-05-05
Allow setting push notification visibility/sensitivity

jpelgrom:
## Summary

This PR ~+ home-assistant/mobile-apps-fcm-push#73~ adds support for setting the visibility* of push notifications, which allows you to control what notification content is visible on the lock screen when the device is locked.

Example use cases:
 - setting visibility to `secret` to completely hide a notification from the lock screen
 - setting visibility to `public` to make a notification visible on the lock screen, when in the system settings you've set it to hide sensitive content from the lock screen

Users on Android 5-7 will benefit most from this change because the system setting only allows controlling the visibility of all notifications. On Android 8+ notification channels allow limiting visibility beyond the general system setting (first use case), but they don't allow users to make a notification more visible (second use case).

\*or sensitivity, as user-facing settings call it

Edit: it turns out the FCM code already has a `visibility` attribute even though it isn't used anywhere in the Android app?

## Screenshots

The following device is set to hide sensitive content from the lock screen in system settings > Notifications. The locked lock screen shows a private notification and a public notification. 
|Light|Dark|
|-----|-----|
|A lock screen showing two notifications, one collapsed with only the Home Assistant logo and 'Home Assistant - now' visible, the other one showing the Home Assistant logo and 'Public notification - 1m - Hello!', light theme|A lock screen showing two notifications, one collapsed with only the Home Assistant logo and 'Home Assistant - now' visible, the other one showing the Home Assistant logo and 'Public notification - now - Hello!', dark theme|

Adding a screenshot of a secret notification wouldn't be useful, as it is just an empty lock screen.

Documentation: home-assistant/companion.home-assistant#743

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -105,6 +105,7 @@ class MessagingManager @Inject constructor(
         const val IMAGE_URL = "image"
         const val ICON_URL = "icon_url"
         const val VIDEO_URL = "video"
+        const val VISIBILITY = "visibility"
         const val LED_COLOR = "ledColor"
         const val VIBRATION_PATTERN = "vibrationPattern"
         const val PERSISTENT = "persistent"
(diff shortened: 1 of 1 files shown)


dshokouhi (approved):
Looks good to me

JBassett (approved):

PR #2526, opened 2022-05-14
TLS Client Authentication Support

magicbuffer:
## Summary
This PR enables the Home Assistant app to perform TLS Client Authentication, typically when the Home Assistant instance is behind a reverse proxy. This setup ensure that only clients with a valid certificate can communicate with the Home Assistant backend, this considerably prevents the risks of Home Assistant being exploited for user with a publicly reachable instance.

## Screenshots
This screenshot show the user being prompted to select a certificate when the backend requests one.
cert

[Documentaiton](https://github.com/home-assistant/companion.home-assistant/pull/748)

## Any other notes
Thanks @mircoboschi for the original implementation (#2023 ) and the whole HA community for this great app!


homeassistant:
Hi @magicbuffer, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;android#2526).

Thanks!

functionpointer:
I have just tried commit 851f41413d8d8c8a20c2a0601ad270a1a0441b54 and i have to say it works great with nginx.

Finally i can access homeassistant from the public internet with no security issues :)
I hope this will get merged soon.

phoenix1589:
The TLS authentication is working well with this PR (thanks @magicbuffer @mircoboschi )! However this version doesn't list any location senors available on my OnePlus 5; unlike the latest version from the play store.

(I have enabled the location permission for this version)

jpelgrom:
> However this version doesn't list any location senors available

Check which APK you've downloaded as this PR doesn't change any sensors. It sounds like you're using the minimal version of the app which does not include location tracking.

phoenix1589:
> Check which APK you've downloaded as this PR doesn't change any sensors. It sounds like you're using the minimal version of the app which does not include location tracking.

🤦 thanks; yup I was using minimal. Full version installed and has location sensors.

kmccmk9:
Looks like the linting check failed, but these should be relatively simple to update with an additional commit from @magicbuffer 

```
/home/runner/work/android/android/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt:25:1 Wildcard import (cannot be auto-corrected)
/home/runner/work/android/android/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt:12:1 Wildcard import (cannot be auto-corrected)
/home/runner/work/android/android/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt:30:1 Unused import
/home/runner/work/android/android/app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt:1:1 File must end with a newline (\n)
/home/runner/work/android/android/app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt:13:1 Unused import
/home/runner/work/android/android/app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt:19:1 Unused import
/home/runner/work/android/android/app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt:21:1 Unused import
/home/runner/work/android/android/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt:1101:93 Missing newline before ")"
```

magicbuffer:
Should be better this time :crossed_fingers:

biscuit-316:
I tried the latest [app-full-debug.apk](https://github.com/home-assistant/android/actions/runs/2362220950). And it works well.

The first client certificate I used has expired. And it gives a nice error message "ERR_BAD_SSL_CLIENT_AUTH_CERT". But it only did it once. Now only the "unable to connect to Home Assistant" screen is displayed.
Also, the app does not display an error message when the client certificate is revoked. Firefox does give an error. Maybe the app is giving an error message now that [PR 2535](https://github.com/home-assistant/android/pull/2535) has been merged.

One thing I'm really missing is a setting/popup/sensor that can tell me when my client certificate expires. I couldn't find it in the android settings. nor in the Home Assistant app. I wish I could see the expiration date somewhere. Or else that the app would inform me in time with a pop-up or notification.

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -0,0 +1,119 @@
...
+        if (activity == null) return
+
+        var key: PrivateKey?
+        var chain: Array<X509Certificate>?
+
+        // Get the key and the chain (if the user previously chose)
+        runBlocking {
+            key = keyChainRepository.getPrivateKey()
+            chain = keyChainRepository.getCertificateChain()
+        }
```
JBassett, line 67: This feels wrong, can we launch this whole process in a coroutine and call `request.proceed` when we are ready?

app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
```diff
@@ -57,6 +62,10 @@ open class HomeAssistantApplication : Application() {
             }
         )
 
+        runBlocking {
+            keyChainRepository.load(applicationContext)
+        }
```
JBassett, line 67: This is going to affect everyones startup by running this blocking.  This should be launched on an IO coroutine scope like we do for `initCrashReporting`

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -0,0 +1,119 @@
...
+        while (!kcac.ready) {
+            Thread.sleep(200)
+        }
+
+        // load the key and the cert chain
+        runBlocking {
+            if (kcac.alias != null) {
+                keyChainRepository.load(activity.applicationContext, kcac.alias!!)
+            }
+        }
```
JBassett, line 115: `KeyChain.choosePrivateKeyAlias` has a callback, we should not be looping until it's ready.

JBassett, line 115: Again, this shouldn't be run blocking.

common/src/main/java/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -0,0 +1,95 @@
...
+                // block until a key is provided via the TLSWebView
+                while (true) {
+                    runBlocking {
+                        key = keyChainRepository.getPrivateKey()
+                    }
+
+                    if (key != null) break
+
+                    Thread.sleep(100)
+                }
```
JBassett, line 89: We should not be looping like this.

common/src/main/java/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -0,0 +1,95 @@
...
+                var chain: Array<X509Certificate>?
+
+                // block until a chain is provided via the TLSWebView
+                while (true) {
+                    runBlocking {
+                        chain = keyChainRepository.getCertificateChain()
+                    }
+
+                    if (chain != null) break
+                }
```
JBassett, line 72: We shouldn't be looping like this.

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+            Executors.newSingleThreadExecutor().execute {
+                if (chain == null) {
+                    chain = KeyChain.getCertificateChain(context, alias!!)
+                }
+                if (key == null) {
+                    key = KeyChain.getPrivateKey(context, alias!!)
+                }
+                isLoading = false
+            }
+        }
```
JBassett, line 50: Why are you spawning a thread for this?  Why can't we just use a coroutine?

JBassett (changes requested):

magicbuffer:
@JBassett Thanks for the code review, i am having a look at the various comments.
@biscuit-316 Hmm to be honest I did not get that far in the testing, i will have a look if the the various ssl error condition can be hardened.

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
magicbuffer: Not an Android/Kotlin so to be sure, I assume something like this would run non-blocking?
```
runBlocking {
    launch {
         <stuff>
```

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
magicbuffer: It was to be sure this code will not run out of the main thread. Do we have the guarantee the corouting won't be running out of the main thread or can we somehow force this?

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+import io.homeassistant.companion.android.common.data.prefs.PrefsRepository
+import java.security.PrivateKey
+import java.security.cert.X509Certificate
+import java.util.concurrent.Executors
+import javax.inject.Inject
+
+class KeyChainRepositoryImpl@Inject constructor() : KeyChainRepository {
+
+    @Inject
+    lateinit var prefsRepository: PrefsRepository
```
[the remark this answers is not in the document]
magicbuffer: This should be resolved.

app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
magicbuffer: This should be resolved

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
magicbuffer: This should be resolved

magicbuffer: Is it okay if it runs from a coroutine?

common/src/main/java/io/homeassistant/companion/android/common/data/TLSHelper.kt
magicbuffer: This should be resolved

common/src/main/java/io/homeassistant/companion/android/common/data/TLSHelper.kt
magicbuffer: This should be resolved

magicbuffer:
Something broke with the latest changes (during the initial onboarding the app will report an error, as a quick fix restarting the app will solve the connection issue and then the app will be good to go). I am looking into it.

magicbuffer:
The issue during onboarding seems resolved (hard to say for sure with race condition) gonna run some more test to be sure but it looks good on my end.

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
magicbuffer: I just realised I wrote a comment about this but it is not well located. In essence, calling `getCertificateChain` or `getPrivateKey` will block until a connection is received, if called from the main looper then there won't be any more dispatch and the app will hang. It is documented [here](https://developer.android.com/reference/android/security/KeyChain).
So it is important, but maybe there is a cleaner way to this (I am not familiar at all with the Android SDK and the Kotlin API)

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @RequiresApi(Build.VERSION_CODES.M)
+    private fun selectClientCert(activity: Activity, principals: Array<Principal>?, request: ClientCertRequest) {
+        var kcac = KeyChainAliasCallback { alias ->
+            if (alias != null) {
+                runBlocking {
+                    // Load the key and the chain
+                    keyChainRepository.load(activity.applicationContext, alias!!)
```
marazmarci, line 83: We already know here that `alias` is not `null`.
```suggestion
                    keyChainRepository.load(activity.applicationContext, alias)
```

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    private fun getActivity(context: Context?): Activity? {
+        if (context == null) {
+            return null
+        } else if (context is ContextWrapper) {
+            return if (context is Activity) {
+                context
+            } else {
+                getActivity((context).baseContext)
```
marazmarci, line 39:
```suggestion
                getActivity(context.baseContext)
```

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
marazmarci: I suggest wrapping the whole `load(Context)` method's body into a `withContext(Dispatchers.IO) { ... }`, like this:

```
override suspend fun load(context: Context): Boolean = withContext(Dispatchers.IO) {
```

And you can remove the `Executor` altogether, and make the logic sequential. I don't know if the 5 sec timeout is important, but if it is, I can help to come up with a coroutine-based solution for that.

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepository.kt
```diff
@@ -0,0 +1,17 @@
...
+import android.content.Context
+import java.security.PrivateKey
+import java.security.cert.X509Certificate
+
+interface KeyChainRepository {
+    suspend fun getAlias(): String?
+
+    suspend fun load(context: Context, alias: String): Boolean
+
+    suspend fun load(context: Context): Boolean
```
marazmarci, line 12: The `Boolean` return types of these methods are not used anywhere, as I saw. I think we could remove them.

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+        return isLoading
+    }
+
+    override suspend fun getPrivateKey(): PrivateKey? {
+        return key
+    }
+
+    override suspend fun getCertificateChain(): Array<X509Certificate>? {
+        return chain
+    }
```
marazmarci, line 64: The implementations of these methods seem pretty non-suspending, so I suggest removing the `suspend` modifiers. This will make our lives easier where they are used, because we won't need to be in a `suspending` scope to be able to call them. The same applies to the `getAlias()` method.

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+    lateinit var prefsRepository: PrefsRepository
+
+    private var alias: String? = null
+    private var key: PrivateKey? = null
+    private var chain: Array<X509Certificate>? = null
+    private var isLoading: Boolean = false
+
+    override suspend fun getAlias(): String? {
+        return alias
+    }
```
marazmarci, line 23: The `suspend` modifier could be removed from this method as well.

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
marazmarci: @magicbuffer no, the snippet you pasted won't run non-blocking! See this little [Kotlin playground snippet](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS42LjIxIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJjb2RlIjoiaW1wb3J0IGtvdGxpbnguY29yb3V0aW5lcy4qXG5cbmZ1biBtYWluKCkge1xuICAgIHJ1bkJsb2NraW5nIHtcbiAgICBcdGxhdW5jaCB7XG5cdFx0ICAgIHByaW50bG4oXCJpbnNpZGUgbGF1bmNoXCIpXG4gICAgICAgICAgICBkZWxheSgxMDAwTClcblx0XHQgICAgcHJpbnRsbihcImluc2lkZSBsYXVuY2ggYWZ0ZXIgMXNcIilcbiAgICAgICAgfVxuXHQgICAgcHJpbnRsbihcImFmdGVyIGxhdW5jaFwiKVxuXHR9XG4gICAgcHJpbnRsbihcImFmdGVyIHJ1bkJsb2NraW5nXCIpXG59In0=) I wrote as a demonstration.

As a solution, I can imagine something like obtaining/constructing a `CoroutineScope` in `TLSWebViewClient`, and calling `launch` on that.

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -0,0 +1,102 @@
...
+    override fun onReceivedClientCertRequest(view: WebView, request: ClientCertRequest) {
+        // Let the WebViewActivity know the endpoint requires TLS Client Auth
+        isTLSClientAuthNeeded = true
+
+        // Aim to obtain the private key for the whole lifecycle of the WebViewActivity
+        val activity = getActivity(view.context)
+        if (activity == null) return
+
+        runBlocking {
+            launch {
```
marazmarci, line 55: This `runBlocking` and `launch` call can be removed If you remove the `suspend` modifiers from `KeyChainRepository.getPrivateKey()` and `KeyChainRepository.getCertificateChain()` as I suggested in my [other comment](https://github.com/home-assistant/android/pull/2526/files/6a606252fde6eb8330072f437082868700884a28#r890575307).

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -0,0 +1,102 @@
...
+                }
+            }
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.M)
+    private fun selectClientCert(activity: Activity, principals: Array<Principal>?, request: ClientCertRequest) {
+        var kcac = KeyChainAliasCallback { alias ->
+            if (alias != null) {
+                runBlocking {
```
marazmarci, line 81: This will block the calling thread (as its name suggests 😉) and it's probably not a good idea.

One thing that can be done instead is the idea I wrote down at the end of [this comment](https://github.com/home-assistant/android/pull/2526#discussion_r890587980).

magicbuffer:
Apologies for the delay and thanks you @marazmarci for the review. I will have a look at these points.

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
magicbuffer: I like this idea, I will try to reuse that `lifecycleScope` before attempting to create a new one :+1:

magicbuffer: It is working like a charm :clap:  thank you!

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
magicbuffer: Thank! Reusing the 'lifecycleScope' + the IO Dispatcher did the trick just fine! Should be okay now.

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
magicbuffer: This should be okay

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
magicbuffer: Done

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
magicbuffer: Done

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepository.kt
magicbuffer: Done

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
magicbuffer: Done

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
magicbuffer: Done

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
magicbuffer: Thank you this worked well and the 5 sec timeout was purely arbitrary!

magicbuffer:
Thanks again for the great review, I did learned a couple of new stuff which is great! Let me know if the code is okay now. Once it is I will rebase again main. :+1:

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1088,6 +1092,17 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
             return
         isShowingError = true
 
+        // If the endpoint requires a TLS client certificate and we got a timeout, ignore the error
+        // It is likely due to the user not choosing a key yet
+        val tlsWebViewclient = webView.webViewClient as TLSWebViewClient
```
marazmarci, line 1097: Android Studio shows this error for the `.webViewClient` getter:
> Call requires API level 26 (current min is 21)

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -0,0 +1,106 @@
...
+    @RequiresApi(Build.VERSION_CODES.M)
+    override fun onReceivedClientCertRequest(view: WebView, request: ClientCertRequest) {
+        // Let the WebViewActivity know the endpoint requires TLS Client Auth
+        isTLSClientAuthNeeded = true
+
+        // Aim to obtain the private key for the whole lifecycle of the WebViewActivity
+        val activity = getActivity(view.context)
+        if (activity == null || activity !is AppCompatActivity) return
+
+        activity.lifecycleScope.launch {
```
marazmarci, line 55: we don't call anything `suspend`ing here, so `activity.lifecycleScope.launch { ... }` could be removed.

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
```diff
@@ -0,0 +1,58 @@
...
+        if (alias != null && !isLoading && (key == null || chain == null)) {
+            isLoading = true
+            if (chain == null) {
+                chain = KeyChain.getCertificateChain(context, alias!!)
+            }
+            if (key == null) {
+                key = KeyChain.getPrivateKey(context, alias!!)
+            }
+            isLoading = false
+        }
```
marazmarci, line 44: I'm wondering if this `isLoading` flag is the best way to handle mutual exclusion here.
If a previous loading is in progress, and there's a new `.load(...)` call, the second load request will be basically dropped. And also it's not atomic.

I think it would be better to use synchronization instead to solve this.
Suggestion: Extract the highlighted code to a new synchronized private method (annotated with `@Synchronized`).
(It's perfectly OK to block a coroutine on `Dispatchers.IO`, this dispatcher is designed for handling blocking operations)
This way every `load(...)` call would eventually lead to an actual call on `KeyChain`, so every time it's called, the caller will have up-to-date `KeyChain` information available by the time the `load` method returns.

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
```diff
@@ -0,0 +1,58 @@
...
+) : KeyChainRepository {
+
+    private var alias: String? = null
+    private var key: PrivateKey? = null
+    private var chain: Array<X509Certificate>? = null
+    private var isLoading: Boolean = false
+
+    override suspend fun load(context: Context, alias: String) {
+        if (alias == null) return
+
```
marazmarci, line 23: `alias` is not nullable, so this can be removed

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -0,0 +1,106 @@
...
+                }
+            }
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.M)
+    private fun selectClientCert(activity: Activity, principals: Array<Principal>?, request: ClientCertRequest) {
+        if (activity !is AppCompatActivity) {
+            return
+        }
```
marazmarci, line 80: I think we could be more strict, because it _should_ be an `AppCompatActivity`, right? What do you think?

But I'm also curious about @JBassett's opinion about this.

```suggestion
        require(activity is AppCompatActivity)
```

marazmarci (changes requested):

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
marazmarci: See my [comment](https://github.com/home-assistant/android/pull/2526#discussion_r901014727) on the newest changes.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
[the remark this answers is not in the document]
magicbuffer: What would be the best way to handle this? Check for the API version and if below 26 fallback on a standard `WebViewClient`?

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -49,38 +50,39 @@ open class TLSWebViewClient @Inject constructor(private var keyChainRepository:
...
             }
         }
     }
 
     @RequiresApi(Build.VERSION_CODES.M)
     private fun selectClientCert(activity: Activity, principals: Array<Principal>?, request: ClientCertRequest) {
+        if (activity !is AppCompatActivity) {
+            return
+        }
         var kcac = KeyChainAliasCallback { alias ->
```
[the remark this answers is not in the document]
magicbuffer: Done

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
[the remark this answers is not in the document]
magicbuffer: Done

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
magicbuffer: Should be okay, i wrapped the new synchronized `doLoad()` call inside a `runBlocking`, I hope that is what you meant.

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
magicbuffer: Done

app/src/main/java/io/homeassistant/companion/android/util/TLSWebViewClient.kt
magicbuffer: Looks fine be, done.

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
magicbuffer: And yes, my code was a hack at best, I am really glad you proposed a clean solution :+1:

magicbuffer:
Also a quick question about how to properly handle error conditions. Right now I am using flags on the `TLSWebViewClient`  (`isTLSClientAuthNeeded` and `hasUserDeniedAccess`) that can be queried from the `WebViewActivity` and `AuthenticationFragment`. Is this the proper way to proceed? I am thinking about that because I am adding a check for the validity of the certificate so that we can inform users when the certificate is not valid instead of bluntly failing and I am not sure I should keep adding more flags.

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
```diff
@@ -32,15 +29,8 @@ class KeyChainRepositoryImpl @Inject constructor(
...
-            if (chain == null) {
-                chain = KeyChain.getCertificateChain(context, alias!!)
-            }
-            if (key == null) {
-                key = KeyChain.getPrivateKey(context, alias!!)
-            }
-            isLoading = false
+        runBlocking {
+            doLoad(context)
         }
```
marazmarci, line 34: There's no need to wrap it in `runBlocking` because `doLoad` is not a `suspend`ing function. See the [documentation of `runBlocking`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/run-blocking.html)

```suggestion
        doLoad(context)
```

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1088,6 +1092,17 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
+        if (tlsWebViewclient != null &&
+            tlsWebViewclient.isTLSClientAuthNeeded &&
+            !tlsWebViewclient.hasUserDeniedAccess &&
+            errorType == io.homeassistant.companion.android.webview.WebView.ErrorType.TIMEOUT
```
marazmarci, line 1101: An import can be added for `ErrorType`:
```
import io.homeassistant.companion.android.webview.WebView.ErrorType
```

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1096,7 +1111,17 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                 waitForConnection()
             }
 
-        if (errorType == io.homeassistant.companion.android.webview.WebView.ErrorType.AUTHENTICATION) {
+        if (errorType == io.homeassistant.companion.android.webview.WebView.ErrorType.TIMEOUT && tlsWebViewclient.hasUserDeniedAccess) {
```
marazmarci, line 1114: An import can be added for `ErrorType`:
```
import io.homeassistant.companion.android.webview.WebView.ErrorType
```

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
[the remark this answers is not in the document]
marazmarci: As far as I understand these API-level things, what matters is to never allow the execution to reach a point where a class/method is used that is not present on the device's API level.
I think `TLSWebViewClient::onReceivedClientCertRequest` is OK, and doesn't need modifications (but pls confirm @JBassett).

For `WebViewActivity`, I think all we need is an if around the `TLSWebViewClient`-related part:
```
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
    // If the endpoint requires a TLS client certificate and we got a timeout, ignore the error
    // It is likely due to the user not choosing a key yet
    val tlsWebViewclient = webView.webViewClient as TLSWebViewClient
    if (tlsWebViewclient != null &&
        tlsWebViewclient.isTLSClientAuthNeeded &&
        !tlsWebViewclient.hasUserDeniedAccess &&
        errorType == ErrorType.TIMEOUT
    ) {
        return
    }
}
```

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1088,6 +1092,17 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
             return
         isShowingError = true
 
+        // If the endpoint requires a TLS client certificate and we got a timeout, ignore the error
+        // It is likely due to the user not choosing a key yet
+        val tlsWebViewclient = webView.webViewClient as TLSWebViewClient
+        if (tlsWebViewclient != null &&
```
marazmarci, line 1098: `tlsWebViewclient` is not nullable, so there's no need to do this null-check.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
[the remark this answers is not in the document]
marazmarci: update: I pasted the code part I wrote in my previous comment into Android Studio, and it showed me a warning telling me there's a compatibility layer util method for this: `WebViewCompat.getWebViewClient(webView)`

Here's how it could be used:
(disclaimer: I didn't try to run the code, I'm posting this blindly)

```
var tlsWebViewclient: TLSWebViewClient? = null
// If the endpoint requires a TLS client certificate and we got a timeout, ignore the error
// It is likely due to the user not choosing a key yet
if (WebViewFeature.isFeatureSupported(WebViewFeature.GET_WEB_VIEW_CLIENT)) {
    tlsWebViewclient = WebViewCompat.getWebViewClient(webView) as TLSWebViewClient
    if (tlsWebViewclient.isTLSClientAuthNeeded &&
        !tlsWebViewclient.hasUserDeniedAccess &&
        errorType == ErrorType.TIMEOUT
    ) {
        return
    }
}
```

And this way there's no need for checking the API level.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
magicbuffer: :+1:

magicbuffer:
I added some basic error handling in the case the client certificate is not valid (thanks @biscuit-316 ).

marazmarci (changes requested):
Thanks for the changes. LGTM besides some minor improvement points I added as comments. I still submit my review as "Request changes" to maintain the code quality in the project by not introducing new warnings into it. I hope you understand 🙂

BTW are you using Android Studio? I'm asking because most of my comments were based on Android Studio warnings 😄

common/src/main/java/io/homeassistant/companion/android/common/data/keychain/KeyChainRepositoryImpl.kt
```diff
@@ -0,0 +1,62 @@
+package io.homeassistant.companion.android.common.data.keychain
+
+import android.content.Context
+import android.security.KeyChain
+import io.homeassistant.companion.android.common.data.prefs.PrefsRepository
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.runBlocking
```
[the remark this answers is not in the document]
dshokouhi: You can easily detect these errors by running `gradlew ktlintCheck` inside Android Studio too.

Once this branch no longer has a merge conflict we will be able to run Github actions which will actually fail for things like this during the lint check. It will also generate a debug APK for others to test with.

magicbuffer:
> Thanks for the changes. LGTM besides some minor improvement points I added as comments. I still submit my review as "Request changes" to maintain the code quality in the project by not introducing new warnings into it. I hope you understand slightly_smiling_face
> 
> BTW are you using Android Studio? I'm asking because most of my comments were based on Android Studio warnings smile

No worries I understand perfectly and yes I use Android Studio; I should be more thorough at checking warnings, sorry to give you some additional work .. 😅
And thanks @dshokouhi for the linter tip, I was honestly too bored to looked that up by myself 😇
I'll make sure the next commit has no warnings and does piss of the linter :+1: 

And as stated previously, once you guys are happy with the code, i'll rebase against main.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1096,13 +1103,49 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
+                }
+            }
+        }
+
+        if (errorType == ErrorType.AUTHENTICATION) {
             alert.setMessage(commonR.string.error_auth_revoked)
             alert.setPositiveButton(android.R.string.ok) { _, _ ->
                 presenter.clearKnownUrls()
                 relaunchApp()
             }
```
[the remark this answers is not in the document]
magicbuffer: hmm indeed and yet I somehow tested that branch... Thank you that was well spotted!

JBassett (approved):
Get this building and I think we can merge.

magicbuffer:
There we go, rebased against master, no linting issues and afaik no warnings?! Also ran some tests and it is still working as expected :+1:

JBassett (approved):

Issue #2530, opened 2022-05-15
Sensor for current time zone reports wrong timezone


FrankLandrock:
Samsung A71 with Homeassistant companion 2022.3.0-full

It is May 2022, my Phone is located in Central Europe.

The current time zone sensor delivers this (taken from Developer-Tools/States): 

Sensor:                                                 
sensor.sm_a715f_current_time_zone

State:                                              
Central European Standard Time   

Attributes:
in_daylight_time: true	
time_zone_id: Europe/Berlin
time_zone_short: GMT+02:00
uses_daylight_time: true
utc_offset: 7200000
icon: mdi:map-clock
friendly_name: SM-A715F Current Time Zone

I am happy with the attributes, but the state is just wrong, it has to be "Central European Summer Time" (NOT "Standard"!). 
There is no such thing as "Central European Standard Time", it is either "Central European Time" (when daylightsaving=false) or "Central European Summer Time" (when daylightsaving=true).

dshokouhi:
The app in this case simply returns the API data from the TimeZone display name using the english locale. I do not think we can correct this as we simply return the data from the API.

https://developer.android.com/reference/java/util/TimeZone#getDisplayName(boolean,%20int,%20java.util.Locale)

https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/sensors/TimeZoneManager.kt#L52

FrankLandrock:
@dshokouhi Thanks for looking into that....

So, it looks like I have to report this issue on some other bug tracker, right? As it seems to be a general Android issue, rather than a HA-compnaion issue? Can you give me a hint on where to do that? 

P.S.: I am new to the show...

jpelgrom:
Time zone data is provided and updated by your device manufacturer. I'm also seeing these time zones names in an emulator so it's probably an error somewhere in a common data source (but I don't know which).

FYI, I'm submitting a fix to change the time zone name to a different value if the device is currently in summer/daylight savings time.

dshokouhi:
Are we good to close this now? I think the original issue is now solved in the latest beta.

PR #2536, opened 2022-05-19
Improve time zone sensor state

jpelgrom:
## Summary

This PR changes the time zone sensor's state to the 'current' name that reflects whether the time zone is in summer/daylight savings time or not (bonus: it 'fixes' 50% of #2530).

## Screenshots

The sensor previously had 'Central European Standard Time' as the state.
|Light|Dark|
|-----|-----|
|Time zone sensor showing 'Central European Summer Time' as the state, light mode|Time zone sensor showing 'Central European Summer Time' as the state, dark mode|

n/a

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/TimeZoneManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/TimeZoneManager.kt
@@ -49,18 +49,17 @@ class TimeZoneManager : SensorManager {
             return
 
         val timeZone: TimeZone = TimeZone.getDefault()
-        val currentZone = timeZone.getDisplayName(Locale.ENGLISH)
-        val date = Date()
+        val currentlyInDst = timeZone.inDaylightTime(Date())
 
         onSensorUpdated(
             context,
             currentTimeZone,
-            currentZone,
+            timeZone.getDisplayName(currentlyInDst, TimeZone.LONG, Locale.getDefault()),
             currentTimeZone.statelessIcon,
             mapOf(
-                "in_daylight_time" to timeZone.inDaylightTime(date),
+                "in_daylight_time" to currentlyInDst,
                 "time_zone_id" to timeZone.id,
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/TimeZoneManager.kt
```diff
@@ -49,18 +49,17 @@ class TimeZoneManager : SensorManager {
...
         val timeZone: TimeZone = TimeZone.getDefault()
-        val currentZone = timeZone.getDisplayName(Locale.ENGLISH)
-        val date = Date()
+        val currentlyInDst = timeZone.inDaylightTime(Date())
 
         onSensorUpdated(
             context,
             currentTimeZone,
-            currentZone,
+            timeZone.getDisplayName(currentlyInDst, TimeZone.LONG, Locale.getDefault()),
```
dshokouhi, line 57: Will the state still be in english if the user had a non english language selected? It looks like this will return the display name in whatever locale is selected and I believe core expects the state to be in english.

jpelgrom: No it will change to match the locale, that's kind of the point (and the reason I listed this as a breaking change for non-English users). Or was there a reason for always using English?

Core seems to treat the state like a string so any value is accepted? I've tested this and it seems to work fine, nothing in the logs either.

dshokouhi: For states core wants them provided in English and then the integration can provide the translations so they appear correct in the frontend. There's an ADR that covers this. Keeping it all in English also provides some consistency too.

https://github.com/home-assistant/architecture/blob/master/adr/0009-translations-2.0.md

jpelgrom: That's the guideline I was looking for, thanks!

dshokouhi (approved):

PR #2540, opened 2022-05-22
Update ManageTilesView for performance

NotWoods:
## Summary
Removes blocking calls and reduces recompositions in ManageTilesView. The viewmodel was rewritten to use private setters and suspend functions.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/qs/TileExtensions.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/qs/TileExtensions.kt
@@ -66,7 +66,7 @@ abstract class TileExtensions : TileService() {
         fun setTileData(context: Context, tileId: String, tile: Tile, integrationUseCase: IntegrationRepository): Boolean {
             Log.d(TAG, "Attempting to set tile data for tile ID: $tileId")
             val tileDao = AppDatabase.getInstance(context).tileDao()
-            val tileData = tileDao.get(tileId)
+            val tileData = tileDao.getSync(tileId)
             try {
                 return if (tileData != null) {
                     tile.label = tileData.label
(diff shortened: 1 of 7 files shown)


app/src/main/java/io/homeassistant/companion/android/qs/TileExtensions.kt
```diff
@@ -66,7 +66,7 @@ abstract class TileExtensions : TileService() {
         fun setTileData(context: Context, tileId: String, tile: Tile, integrationUseCase: IntegrationRepository): Boolean {
             Log.d(TAG, "Attempting to set tile data for tile ID: $tileId")
             val tileDao = AppDatabase.getInstance(context).tileDao()
-            val tileData = tileDao.get(tileId)
+            val tileData = tileDao.getSync(tileId)
```
JBassett, line 69: This whole function should become a suspend function but since all you are actually doing is a rename it's fine.

app/src/main/java/io/homeassistant/companion/android/qs/TileExtensions.kt
```diff
@@ -112,7 +112,7 @@ abstract class TileExtensions : TileService() {
         ) {
             Log.d(TAG, "Click detected for tile ID: $tileId")
             val tileDao = AppDatabase.getInstance(context).tileDao()
-            val tileData = tileDao.get(tileId)
+            val tileData = tileDao.getSync(tileId)
```
JBassett, line 115: This whole function should become a suspend function but since all you are actually doing is a rename it's fine.

JBassett (approved):

app/src/main/java/io/homeassistant/companion/android/qs/TileExtensions.kt
NotWoods: Updated in https://github.com/home-assistant/android/pull/2541

PR #2541, opened 2022-05-23
Add dagger module for daos

NotWoods:
## Summary

Add a dagger module for daos and appDatabase, so they can be injected like repository classes. This helps remove the need to pass context around too.

## Any other notes
https://svvashishtha.medium.com/using-room-with-hilt-cb57a1bc32f


JBassett (approved):
Just need to solve the merge conflicts and looks good!

PR #2542, opened 2022-05-23
Provide Compose preview for Play Store listing screens

jpelgrom:
## Summary

Adds Compose previews for screens that are in the Play Store listing, aka the onboarding screen and Wear OS 'scenes' screen.

## Screenshots

image
image

n/a


dshokouhi:
should we add the updated screenshots to the play folder?

https://github.com/home-assistant/android/blob/master/app/src/full/play/listings/en-US/graphics

Then once we merge this PR we can probably re-push to play store and see if they accept it? Really hope this is why its been rejected.

JBassett:
> should we add the updated screenshots to the play folder?

Yep, everything should be setup so changing those images will cause the play store listing to update as well.

> Then once we merge this PR we can probably re-push to play store and see if they accept it? Really hope this is why its been rejected.

Will do once merged!

jpelgrom:
I've only updated the Wear OS screenshot for now, because the phone preview gives me a much lower resolution than the existing screenshots and it still mostly matches (Google aren't going to complain about font weight I hope?).

Perhaps a future improvement would be to redo all the screenshots with a `demo` configuration + document the device/emulator so that they can easily be re-done if necessary.

dshokouhi (approved):

Issue #2545, opened 2022-05-25
Quest 2 sensors not reflecting when device is off


Sian-Lee-SA:
**Is your feature request related to a problem? Please describe.**

The app should be detecting whether the device is shutting down via flat battery or manual turning off and change sensors to reflect this state (especially in_use sensor)

**Describe the solution you'd like**

**Describe alternatives you've considered, if any**

Can do this with Tasker and webhook but seems like an unnecessary step

**Additional context**

dshokouhi:
If the headset is still being worn when it shut down then the intent that the app receives will never be so we cannot update the state there. Also the android shutdown event does not always give applications enough time to fire an event so this request may not be possible. Even the android app on the phone suffers from this as well. Luckily battery level is an easy to discern this.  If you take the headset off and then shut down manually that will report the correct state. This sensor relies on the Quest hardware to communicate when the headset is on or off so we should stick to that logic.

dshokouhi:
We could try to register for the shutdown event so users can see when that happens but android does not offer any guarantees for data to be sent or for apps to even be active so it may not be viable.

https://developer.android.com/reference/android/content/Intent#ACTION_SHUTDOWN

Sian-Lee-SA:
I felt that could be the case but with tasker seeming to do this very consistently then I assume the shutdown event that android would send should suffice. But my basis is that Tasker possibilities should be the same for any other app as it has no special privilege's.

dshokouhi:
So Tasker will be sending 1 command at that point....the app itself will get triggered to send an update for every enabled sensor so the likelihood of the app relaying that may not happen. We can add the intent to get notified when the shutdown occurs but as you have seen with Android documentation there is no guarantee that data will be allowed to be sent.

Sian-Lee-SA:
That's all understandable

dshokouhi:
For the time being go ahead and enable the Last Update sensor. Then on that same page add an intent under the settings and add the below text as the intent. Once you saved it close and open the app again. That is the sensor you will need to use to see if the shutdown is detected and this "workaround" is actually what the fix will become for most part :)

```
android.intent.action.ACTION_SHUTDOWN
```

This will also become the state to detect when a shutdown has an occurred. You will be able to see how reliable the fix will be while a PR gets ready.

Sian-Lee-SA:
[...]

PR #2572, opened 2022-06-01
Register for the shutdown intent to send 1 final sensor update before device restarts or powers down

dshokouhi:
## Summary

Fixes: #2545 the state of the last update sensor will reflect this intent

## Screenshots

image

## Any other notes

This intent will not be guaranteed however it does show up in testing

https://developer.android.com/reference/android/content/Intent#ACTION_SHUTDOWN

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -153,6 +153,12 @@ open class HomeAssistantApplication : Application() {
             )
         }
 
+        // Add a receiver for the shutdown event to attempt to send 1 final sensor update
+        registerReceiver(
+            sensorReceiver,
+            IntentFilter(Intent.ACTION_SHUTDOWN)
+        )
+
         // Register for all saved user intents
         val sensorDao = AppDatabase.getInstance(applicationContext).sensorDao()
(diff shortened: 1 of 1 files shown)


jpelgrom:
The issue specifically mentions the in_use sensor, to really fix it shouldn't this intent also be used to update any sensors where it makes sense (for example, in_use and is_interactive will always become false if this intent is received)?

dshokouhi:
> The issue specifically mentions the in_use sensor, to really fix it shouldn't this intent also be used to update any sensors where it makes sense (for example, in_use and is_interactive will always become false if this intent is received)?

I am personally not a fan of manipulating the data like that as those 2 sensors use values only retrieved by android. There are other sensors that users would expect to be updated like bluetooth, work profile, wifi or any other sensor in the future. Users may also just ask for all sensors to become `unavailable` as we have seen that before too like when nothing has been reported for hours. By keeping it separate we let users handle the shutdown event however they want to handle it. The sensors will continue to report the last data they have retrieved.

jpelgrom:
> Users may also just ask for all sensors to become `unavailable`

That actually makes a lot of sense to me, but would be a major breaking change.

Because we cannot guarantee that a state update is sent before shutdown, only updating the last update trigger (or at least trying to) is probably the best solution 👍

jpelgrom (approved):

Issue #1898, opened 2021-11-10
Add Wear OS complications


leroyboerefijn:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Add support for watchface complications. I'm thinking of the following:

- Entity state
- Entity attribute value
- Template value
- Template text

**Describe alternatives you've considered, if any**
N.a.

**Additional context**

barrymossel:
And sensor data please.

barrymossel:
More information here: https://developer.android.com/training/wearables/watch-faces/adding-complications

Vedeneb:
Are there plans on implementing this feature? If there isn't anybody working on it already, I might eventually have a look into this myself, I have developed a Wear OS with complication support before, however it was written in Java (never started to like Kotlin...) and I'm not familiar with the HA code base at all so it would definitely be a challenge :D

leroyboerefijn:
@Vedeneb I have it on my list to start working on, but cannot give you any guarantees on timing ;)

If you feel like contributing, then please feel free to work on it! I think the entity state widget will give an example of the type of information we could set in a complication. Also, we're always happy to support over on the discord :)

sepodele:
This would be awesome. Here's a link to 'Exposing your Data to Complications': https://developer.android.com/codelabs/data-sources#0

marazmarci:
I think it would be useful to be able to update the complication on demand, eg. with a notification command sent to the phone, that will forward it to the Wear OS app. But I admit that implementing a periodically updating complication that shows an entity's state would be simpler based on the already implemented Tile's code.

jpelgrom:
> I think it would be useful to be able to update the complication on demand, eg. with a notification command sent to the phone, that will forward it to the Wear OS app. 

If this is implemented, it should probably [match the notification command for Apple Watch users](https://companion.home-assistant.io/docs/apple-watch/complications#manual-updates) to make it easier to handle multiple devices / blueprints.

leroyboerefijn:
@marazmarci @jpelgrom I suggest that a new feature request is made for that once the complication PR is merged

dshokouhi:
Now that initial support has been added should we close this request or is there more to be done still?

leroyboerefijn:
My original plan was to also add attribute support instead of just state support. But maybe it makes sense to just close this indeed

dshokouhi:
ok lets close this one and get new requests created for anything else we want to add.

PR #2574, opened 2022-06-01
Entity state complication

leroyboerefijn:
## Summary

Implementation of entity state complication. This is the first step towards implementing #1898 fully.
When the complication is added, a config activity is opened in which the user can select an entity. This entity can be of any type (not only the ones supported by the app). Depending on the watch face it will then show the state of the entity and optionally the name and icon.

The complication updates automatically whenever the state changes.

## Screenshots

Screenshot_20220601_230037_debug
Screenshot_20220601_225948_sysui

Documentation: home-assistant/companion.home-assistant#760

## Any other notes

Next step is a complication that allows the user to select an attribute of the entity.


leroyboerefijn:
Had to fix a merge conflict with the database, so wasn't able to properly test the DB upgrade. Think it should be fine, since it is just a table added though

common/src/main/java/io/homeassistant/companion/android/database/wear/EntityStateComplicationsDao.kt
```diff
@@ -0,0 +1,24 @@
+package io.homeassistant.companion.android.database.wear
+
+import androidx.room.Dao
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+
+@Dao
+interface EntityStateComplicationsDao {
```
JBassett, line 9: This should be added to the `DatabaseModule.kt` file as well so it can be injected.

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationConfigActivity.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.wear.watchface.complications.datasource.ComplicationDataSourceService.Companion.EXTRA_CONFIG_DATA_SOURCE_COMPONENT
+import dagger.hilt.android.AndroidEntryPoint
+import io.homeassistant.companion.android.complications.views.LoadConfigView
+import javax.inject.Inject
+
+@AndroidEntryPoint
+class ComplicationConfigActivity : ComponentActivity() {
+
+    @Inject
+    lateinit var presenter: ComplicationConfigPresenter
```
JBassett, line 23: This class shouldn't exist.  Everything should be in the view or the view model.

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationConfigComponent.kt
```diff
@@ -0,0 +1,13 @@
...
+import dagger.Binds
+import dagger.Module
+import dagger.hilt.InstallIn
+import dagger.hilt.android.components.ActivityComponent
+
+@Module
+@InstallIn(ActivityComponent::class)
+abstract class ComplicationConfigComponent {
+    @Binds
+    abstract fun complicationConfigPresenter(complicationConfigPresenterImpl: ComplicationConfigPresenterImpl): ComplicationConfigPresenter
```
JBassett, line 12: No need for a presenter.

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationConfigPresenter.kt
```diff
@@ -0,0 +1,13 @@
+package io.homeassistant.companion.android.complications
+
+import io.homeassistant.companion.android.common.data.integration.Entity
+import io.homeassistant.companion.android.common.data.websocket.WebSocketState
+import kotlinx.coroutines.flow.Flow
+
+interface ComplicationConfigPresenter {
```
JBassett, line 7: All of these functions should be in the view model.

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationConfigViewModel.kt
```diff
@@ -0,0 +1,125 @@
...
+
+    enum class LoadingState {
+        LOADING, READY, ERROR
+    }
+
+    private lateinit var presenter: ComplicationConfigPresenter
+    val app = getApplication<HomeAssistantApplication>()
+    private val entityStateComplicationsDao = AppDatabase.getInstance(app.applicationContext).entityStateComplicationsDao()
+
+    // TODO: This is bad, do this instead: https://stackoverflow.com/questions/46283981/android-viewmodel-additional-arguments
```
JBassett, line 36: See I even said it was bad! :-p

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationReceiver.kt
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    private fun onScreenOff() {
+        mainScope.cancel()
+        entityUpdates = null
+    }
+
+    private suspend fun getAllComplicationIds(context: Context): List<Int> {
+        return AppDatabase.getInstance(context).entityStateComplicationsDao().getAll().map { it.id }
```
JBassett, line 92: entityStateComplicationsDao should be injected in now.

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationReceiver.kt
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    private suspend fun onEntityStateChanged(context: Context, entity: Entity<*>) {
+        Log.d(TAG, "Entity state changed for ${entity.entityId}")
+        AppDatabase.getInstance(context).entityStateComplicationsDao().getAll().forEach {
```
JBassett, line 97: entityStateComplicationsDao should be injected in now.

wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
```diff
@@ -0,0 +1,86 @@
...
+    @Inject
+    lateinit var integrationUseCase: IntegrationRepository
+
+    override suspend fun onComplicationRequest(request: ComplicationRequest): ComplicationData? {
+        if (request.complicationType != ComplicationType.SHORT_TEXT)
+            return null
+
+        val id = request.complicationInstanceId
+
+        val entityStateComplicationsDao = AppDatabase.getInstance(application.applicationContext).entityStateComplicationsDao()
```
JBassett, line 37: entityStateComplicationsDao should be injected in now.

JBassett (changes requested):

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationReceiver.kt
```diff
@@ -0,0 +1,134 @@
...
+import javax.inject.Inject
+
+@AndroidEntryPoint
+class ComplicationReceiver : BroadcastReceiver() {
+    private var entityUpdates: Flow<Entity<*>>? = null
+
+    @Inject
+    lateinit var integrationUseCase: IntegrationRepository
+    private val scope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate)
+    private var mainScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
```
marazmarci, line 30: Do we really need this `mainScope` property? I think it isn't very clear to have two `CoroutineScope` properties for `Dispatchers.Main`.

I think it would be better to use `scope.launch {...}` in `onScreenOn()` and store the `Job` return value as a property (eg. `screenOnJob: Job?`) and in `onScreenOff()`, call `screenOnJob?.cancel()` and set `screenOnJob = null`.

Or is this screen on/off handling a special case, where `Dispatchers.Main.immediate` cannot be used for some reason?

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationReceiver.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    override fun onReceive(context: Context, intent: Intent) {
+        val result = goAsync()
+
+        try {
+            when (intent.action.toString()) {
```
marazmarci, line 36: the `.toString()` is not necessary here, because `intent.action` is already a `String`, and the potential `"null"` (`String`) return value this extension can produce, won't be used here anyways.

common/src/main/res/values/strings.xml
```diff
@@ -102,9 +102,14 @@
     <string name="cancel">Cancel</string>
     <string name="changelog">View Full Change Log</string>
     <string name="checking_with_home_assistant">Checking with Home Assistant</string>
+    <string name="choose_entity">Choose entity</string>
     <string name="choose_server">Choose server</string>
     <string name="clear_favorites">Clear Favorites</string>
     <string name="color_temp">Color temperature: %1$d</string>
+    <string name="complication_entity_invalid">Invalid entity</string>
+    <string name="complication_entity_state_content_description">Entity state</string>
+    <string name="complication_entity_state_label">Complication</string>
```
jpelgrom, line 111: I think "Entity state" makes more sense, because this is shown without any context in a list titled 'Complications' when adding a complication (on the Wear OS 3 emulator):
A list of complication options, with the Home Assistant app option labeled 'Complication' alongside 'Battery' and 'Countdown to date'

wear/src/main/res/drawable/ic_lightbulb.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+  <path
+      android:fillColor="#FF000000"
```
jpelgrom, line 7: I'm seeing a black lightbulb on a grey background when choosing a complication because of this. Everywhere else the icon is being tinted, and I don't think Wear OS ever has a white/light background in native system menus, so maybe set this color to white to match other complications in the list?
A list of complication options with the app's complication showing a black lightbulb on a grey background, barely visible

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
```diff
@@ -8,9 +8,13 @@ import com.mikepenz.iconics.IconicsDrawable
 import com.mikepenz.iconics.typeface.IIcon
 import com.mikepenz.iconics.typeface.library.community.material.CommunityMaterial
 import io.homeassistant.companion.android.common.data.integration.Entity
+import io.homeassistant.companion.android.home.HomePresenterImpl
 import java.util.Calendar
 import io.homeassistant.companion.android.common.R as commonR
 
+fun stringForDomain(domain: String, context: Context): String? =
+    HomePresenterImpl.domainsWithNames[domain]?.let { context.getString(it) }
```
jpelgrom, line 16: This should probably be extended / adjusted because there are currently a lot of entities not supported by the app, which will now show up as something like "automation" or "device_tracker"

wear/src/main/java/io/homeassistant/companion/android/complications/views/ComplicationConfigMainView.kt
```diff
@@ -0,0 +1,135 @@
...
+                    hasSelected = complicationConfigViewModel.hasSelected.value,
+                    loaded = complicationConfigViewModel.loadingState.value == ComplicationConfigViewModel.LoadingState.READY,
+                    error = complicationConfigViewModel.loadingState.value == ComplicationConfigViewModel.LoadingState.ERROR,
+                    onChooseEntityClicked = {
+                        swipeDismissableNavController.navigate(SCREEN_CHOOSE_ENTITY)
+                    },
+                    onAcceptClicked = onAcceptClicked
+                )
+            }
+            composable(SCREEN_CHOOSE_ENTITY) {
```
jpelgrom, line 60: The title for this screen is currently "Choose shortcuts", but should be "Choose entity" to match the chip that was just tapped

wear/src/main/java/io/homeassistant/companion/android/complications/views/ComplicationConfigMainView.kt
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+}
+
+@Composable
+fun MainConfigView(
+    entity: SimplifiedEntity,
+    hasSelected: Boolean,
+    loaded: Boolean,
+    error: Boolean,
```
jpelgrom, line 82: These are both essentially the same thing: `complicationConfigViewModel.loadingState`. Wouldn't it be better to pass the loading state to the Composable, and use a local variable for `loaded`? (`error` is used just once!)

wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt
```diff
@@ -199,17 +199,18 @@ fun LoadHomePage(
...
                             mainViewModel.clearTileShortcut(shortcutEntitySelectionIndex)
                             TileService.getUpdater(context).requestUpdate(ShortcutsTile::class.java)
                             swipeDismissableNavController.navigateUp()
-                        },
-                        { entity ->
-                            mainViewModel.setTileShortcut(shortcutEntitySelectionIndex, entity)
-                            TileService.getUpdater(context).requestUpdate(ShortcutsTile::class.java)
-                            swipeDismissableNavController.navigateUp()
-                        }
+                        }, { entity ->
```
jpelgrom, line 209: For consistent indentation, please put the next function on a new line :)

wear/src/main/java/io/homeassistant/companion/android/home/views/ChooseEntityView.kt
```diff
@@ -19,44 +22,49 @@ import com.mikepenz.iconics.typeface.library.community.material.CommunityMateria
...
-import io.homeassistant.companion.android.home.MainViewModel
 import io.homeassistant.companion.android.theme.WearAppTheme
 import io.homeassistant.companion.android.util.getIcon
+import io.homeassistant.companion.android.util.stringForDomain
 import io.homeassistant.companion.android.common.R as commonR
 
 @Composable
 fun ChooseEntityView(
-    mainViewModel: MainViewModel,
+    context: Context,
```
jpelgrom, line 32: Don't pass a Context to the composable, use `LocalContext.current`

jpelgrom (commented):
So many comments already! I'm going to add even more.

One thing that's harder for me to test on an emulator: what is the battery impact of keeping the websocket connection alive while the screen is on? For the main app, it seems to depend on the device but I'd expect that for watches it could be quite severe given the small batteries and because we're listening for all updates. Would it be worth adding an option to allow users to specify how often they want to update the complications (instantly/websocket or every x min)?

marazmarci:
> One thing that's harder for me to test on an emulator: what is the battery impact of keeping the websocket connection alive while the screen is on? For the main app, it seems to depend on the device but I'd expect that for watches it could be quite severe given the small batteries and because we're listening for all updates. Would it be worth adding an option to allow users to specify how often they want to update the complications (instantly/websocket or every x min)?

@jpelgrom, it's a bit off-topic, but I think it's important to talk about this: I added a [comment](https://github.com/home-assistant/android/issues/1898#issuecomment-1119865201) on #1898 a few weeks ago. It's about being able to update the complication on-demand through the phone via a notification command. What do you think about this?

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationReceiver.kt
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    private suspend fun getAllComplicationIds(context: Context): List<Int> {
+        return AppDatabase.getInstance(context).entityStateComplicationsDao().getAll().map { it.id }
+    }
+
+    private suspend fun onEntityStateChanged(context: Context, entity: Entity<*>) {
+        Log.d(TAG, "Entity state changed for ${entity.entityId}")
+        AppDatabase.getInstance(context).entityStateComplicationsDao().getAll().forEach {
+            Log.d(TAG, it.entityId)
```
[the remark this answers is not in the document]
dshokouhi: Don't we already print all websocket messages to the logs anyways?

jpelgrom:
> It's about being able to update the complication on-demand through the phone via a notification command. What do you think about this?

The Wear OS app is standalone so it should be possible to use all functionality without a phone (unless there is a good reason not to support it, like the template tile). As a result, I don't think it is in the scope of this PR to add such a feature, but a good future improvement.

(Also, the existing code already allows updating a complication on demand by tapping on it.)

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationReceiver.kt
leroyboerefijn: You're absolutely right, didn't pay enough attention with all copy-pasting :P

leroyboerefijn:
> One thing that's harder for me to test on an emulator: what is the battery impact of keeping the websocket connection alive while the screen is on? For the main app, it seems to depend on the device but I'd expect that for watches it could be quite severe given the small batteries and because we're listening for all updates. Would it be worth adding an option to allow users to specify how often they want to update the complications (instantly/websocket or every x min)?

I've been testing it for two days now and definitely notice the battery impact a bit. For now I'm wondering if it will help to only update when the screen turns on and don't listen for additional changes (the screen is typically on for a short time only anyways).
I also tried setting the update frequency to 1 minute in the manifest, but the OS seems to not like updating more frequently than every ~5 min. I feel this is too long, so for faster updates we should implement our own jobscheduler/alarm or something right?

Any input on best way forward for updating?

wear/src/main/res/drawable/ic_lightbulb.xml
leroyboerefijn: Should definitely not be black indeed. What's interesting is that my samsung watch has a range of colored icons in this menu, so I'm thinking to go home assistant blue?

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
leroyboerefijn: True, but in the current implementation the keys of this list are taken as supported entities in `HomePresenter`, so maybe need to extend the existing list with all possible domains...

leroyboerefijn: By the way, the list of domains is basically endless, but I'll start with including the most common domains :)

jpelgrom: > in the current implementation the keys of this list are taken as supported entities in `HomePresenter`

Right, so I'd say you need to use another list for the strings or change the function that returns the supported domains 🙂

> By the way, the list of domains is basically endless

Good point. Perhaps a nicer fallback string would be a good alternative? (replace underscores with spaces, capitalize the string)

wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
```diff
@@ -0,0 +1,86 @@
...
+            contentDescription = PlainComplicationText.Builder(getText(R.string.complication_entity_state_content_description))
+                .build()
+        )
+            .setTapAction(ComplicationReceiver.getComplicationToggleIntent(this, request.complicationInstanceId))
+            .setMonochromaticImage(MonochromaticImage.Builder(Icon.createWithBitmap(iconBitmap)).build())
+            .setTitle(PlainComplicationText.Builder(attributes["friendly_name"] as String? ?: entity.entityId).build())
+            .build()
+    }
+
+    override fun getPreviewData(type: ComplicationType): ComplicationData =
```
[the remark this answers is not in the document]
leroyboerefijn: The user can have multiple of the same complication type, in the preview you don't know which entity is selected. The info from this function will only be displayed when changing/configuring watch faces, so I think it is desired to just have a quick call here. Let me know if you agree!

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationReceiver.kt
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    private fun updateComplication(context: Context, id: Int) {
+        scope.launch {
+            // Request an update for the complication that has just been toggled.
+            ComplicationDataSourceUpdateRequester
+                .create(
+                    context = context,
+                    complicationDataSourceComponent = ComponentName(context, "io.homeassistant.companion.android.complications.EntityStateDataSourceService")
```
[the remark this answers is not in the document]
leroyboerefijn: Nope, just me not thinking ;)

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationReceiver.kt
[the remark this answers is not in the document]
leroyboerefijn: I always tend to forget removing these debug logs haha...

leroyboerefijn:
> @jpelgrom, it's a bit off-topic, but I think it's important to talk about this: I added a [comment](https://github.com/home-assistant/android/issues/1898#issuecomment-1119865201) on #1898 a few weeks ago. It's about being able to update the complication on-demand through the phone via a notification command. What do you think about this?

I'd prefer to keep this functionality out of this PR

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
leroyboerefijn: > Good point. Perhaps a nicer fallback string would be a good alternative? (replace underscores with spaces, capitalize the string)

Good suggestion :)

leroyboerefijn:
Appreciate the comments everyone! Let me know if you have more comments on the latest version.

I think we should give it a spin for a few days to check the battery impact

wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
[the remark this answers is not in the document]
jpelgrom: I missed that you only get the `ComplicationType` and thought there was more information, in that case this is the best possible solution!

wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
```diff
@@ -0,0 +1,86 @@
...
+        val id = request.complicationInstanceId
+
+        val entityId = entityStateComplicationsDao.get(id)?.entityId
+            ?: return ShortTextComplicationData.Builder(
+                text = PlainComplicationText.Builder(getText(R.string.complication_entity_invalid)).build(),
+                contentDescription = PlainComplicationText.Builder(getText(R.string.complication_entity_state_content_description))
+                    .build()
+            ).build()
+
+        val entity = integrationUseCase.getEntity(entityId)
```
jpelgrom, line 46: An integration request also might throw an exception, maybe catch it + return `null` to indicate no update instead of crashing? Or doesn't the system mind uncaught exceptions and will the complication keep working?

jpelgrom (commented):
> For now I'm wondering if it will help to only update when the screen turns on and don't listen for additional changes (the screen is typically on for a short time only anyways).

I hadn't considered that option, but for a watch it makes a lot of sense. Could we combine this with OS updates every 15 minutes (via manifest) for example, to try to always have a recent state for always on displays / in case we cannot reach the server the moment the screen is turned on?

> I also tried setting the update frequency to 1 minute in the manifest, but the OS seems to not like updating more frequently than every ~5 min.

5 minutes is mentioned [in the documentation as the minimum enforced by the OS](https://developer.android.com/training/wearables/tiles/exposing-data-complications#:~:text=If%20you%20don%27t%20set%20UPDATE_PERIOD_SECONDS%20to%200%2C%20you%20must%20use%20a%20value%20of%20at%20least%20300%20(5%20minutes)%2C%20which%20is%20the%20minimum%20update%20period%20that%20the%20system%20enforces%20in%20order%20to%20preserve%20device%20battery%20life.)

> I feel this is too long, so for faster updates we should implement our own jobscheduler/alarm or something right?

JobScheduler/WorkManager enforces a minimum of 15 minutes between jobs, you could register for `Intent.ACTION_TIME_TICK` to update every minute like sensor updates in the main app. However, I think update on screen on + background updates triggered by the OS (see comment after first quote) would be good enough for most people and relatively battery friendly. If someone really wants live/more frequent updates, they can always tap the complication to update it or open the app for websocket updates.

Unrelated to update frequency: `getIcon` should be amended to provide icons for most common domains, anything not controllable in the app without a custom icon currently shows the cellphone icon. Not for this PR, just something to keep in mind.

wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
```diff
@@ -0,0 +1,86 @@
...
+                text = PlainComplicationText.Builder(getText(R.string.state_unknown)).build(),
+                contentDescription = PlainComplicationText.Builder(getText(R.string.complication_entity_state_content_description))
+                    .build()
+            ).build()
+
+        val attributes = entity.attributes as Map<*, *>
+        val icon = getIcon(entity, entity.domain, applicationContext) ?: CommunityMaterial.Icon.cmd_cellphone
+        val iconBitmap = IconicsDrawable(this, icon).apply {
+            colorInt = Color.WHITE
+            backgroundColor = IconicsColor.colorRes(io.homeassistant.companion.android.R.color.colorOverlay)
```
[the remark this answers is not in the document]
leroyboerefijn: Ah didn't have this problem on my watch, but the background was a copy-paste error. Can you try again?

wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
leroyboerefijn: Don't know if it cares about uncaught exceptions, but indeed simple change to make it catch it. Good call ;)

leroyboerefijn:
> Could we combine this with OS updates every 15 minutes (via manifest) for example, to try to always have a recent state for always on displays / in case we cannot reach the server the moment the screen is turned on?

If it works anything like the tiles, it will only use the manifest as a guideline and only update when it will actually show, in which case it is just about as fast. If you want I can still add 15 min in the manifest though.

> 5 minutes is mentioned [in the documentation as the minimum enforced by the OS](https://developer.android.com/training/wearables/tiles/exposing-data-complications#:~:text=If%20you%20don%27t%20set%20UPDATE_PERIOD_SECONDS%20to%200%2C%20you%20must%20use%20a%20value%20of%20at%20least%20300%20(5%20minutes)%2C%20which%20is%20the%20minimum%20update%20period%20that%20the%20system%20enforces%20in%20order%20to%20preserve%20device%20battery%20life.)

Oh right, that makes sense then for composable as well :)

> Unrelated to update frequency: `getIcon` should be amended to provide icons for most common domains, anything not controllable in the app without a custom icon currently shows the cellphone icon. Not for this PR, just something to keep in mind.

Yes that would be nice as well indeed! By the way, why is the default icon cellphone and can we not change it for something more relevant? Though I'm not sure what icon would make sense honestly...

leroyboerefijn:
FYI, today battery usage is much better, now that listening for updates during screen on is removed. Used 1.5% today for the complication + 1 time app update launching from AS.

jpelgrom:
> If it works anything like the tiles, it will only use the manifest as a guideline and only update when it will actually show, in which case it is just about as fast. If you want I can still add 15 min in the manifest though.

So I've just tested this with the emulator in a discharging state, and it does seem to make a difference when in always on / ambient mode; when set to 0 it doesn't update and when set to something else it also updates when the system decides to update (not sure what the schedule is, I've seen ~10 minutes between updates but also 31 minutes once). I expect the battery impact to be minimal enough that this might be a nice addition? Up to you to decide.

> By the way, why is the default icon cellphone and can we not change it for something more relevant?

If I had to guess it's because the default icon for sensors created by the app (if nothing is set) is a cellphone, but you've worked on the app for a longer time so you probably can also make some valid guesses :)

As a more generic alternative, I guess we could follow [the frontend](https://github.com/home-assistant/frontend/blob/91cd584b4b5dd57392f42aac7ee2fe9b5b7255c7/src/common/const.ts#L66) and use [mdi:bookmark](https://materialdesignicons.com/icon/bookmark)?

wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
```diff
@@ -0,0 +1,94 @@
...
+
+        val entity = try {
+            (
+                integrationUseCase.getEntity(entityId)
+                    ?: return ShortTextComplicationData.Builder(
+                        text = PlainComplicationText.Builder(getText(R.string.state_unknown)).build(),
+                        contentDescription = PlainComplicationText.Builder(getText(R.string.complication_entity_state_content_description))
+                            .build()
+                    ).build()
+                )
```
jpelgrom, line 57: The outer `(` `)` aren't needed

wear/src/main/java/io/homeassistant/companion/android/complications/ComplicationConfigViewModel.kt
```diff
@@ -0,0 +1,127 @@
...
+    companion object {
+        const val TAG = "ComplicationConfigViewModel"
+    }
+
+    enum class LoadingState {
+        LOADING, READY, ERROR
+    }
+
+    val app = getApplication<HomeAssistantApplication>()
+    private val entityStateComplicationsDao = AppDatabase.getInstance(app.applicationContext).entityStateComplicationsDao()
```
jpelgrom, line 41: This one should also be injected now

leroyboerefijn: Good catch, missed that one!

leroyboerefijn:
> So I've just tested this with the emulator in a discharging state, and it does seem to make a difference when in always on / ambient mode; when set to 0 it doesn't update and when set to something else it also updates when the system decides to update (not sure what the schedule is, I've seen ~10 minutes between updates but also 31 minutes once). I expect the battery impact to be minimal enough that this might be a nice addition? Up to you to decide.

I completely didn't think about the ambient mode case, added the 15min interval now in the manifest!

> As a more generic alternative, I guess we could follow [the frontend](https://github.com/home-assistant/frontend/blob/91cd584b4b5dd57392f42aac7ee2fe9b5b7255c7/src/common/const.ts#L66) and use [mdi:bookmark](https://materialdesignicons.com/icon/bookmark)?

Very nice overview! I will add all the default icons in this list as well and I propose to also update the default icon to bookmark then.

common/src/main/java/io/homeassistant/companion/android/database/wear/EntityStateComplicationsDao.kt
```diff
@@ -0,0 +1,24 @@
...
+    @Query("SELECT * FROM entityStateComplications WHERE id = :id")
+    suspend fun get(id: Int): EntityStateComplications?
+
+    @Query("SELECT * FROM entityStateComplications")
+    suspend fun getAll(): List<EntityStateComplications>
```
jpelgrom, line 14: This function isn't used

common/src/main/java/io/homeassistant/companion/android/database/wear/EntityStateComplicationsDao.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun add(entityStateComplications: EntityStateComplications)
+
+    @Query("DELETE FROM entityStateComplications WHERE id = :id")
+    suspend fun delete(id: Int)
+
+    @Query("DELETE FROM entityStateComplications")
+    suspend fun deleteAll()
```
jpelgrom, line 23: These functions aren't used

wear/src/main/java/io/homeassistant/companion/android/complications/views/ComplicationConfigMainView.kt
```diff
@@ -0,0 +1,139 @@
...
+                val iconBitmap = getIcon(
+                    entity?.icon,
+                    entity?.domain ?: "light",
+                    LocalContext.current
+                )
+                Chip(
+                    modifier = Modifier.fillMaxWidth(),
+                    icon = {
+                        Image(
+                            asset = iconBitmap ?: CommunityMaterial.Icon2.cmd_lightbulb,
```
jpelgrom, line 98: Shouldn't this also be updated to use the `mdi:bookmark` icon as a fallback?

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
```diff
@@ -56,15 +90,31 @@ fun getIcon(entity: Entity<Map<String, Any>>?, domain: String, context: Context)
                 CommunityMaterial.Icon2.cmd_light_switch
             }
             "input_button" -> CommunityMaterial.Icon2.cmd_gesture_tap_button
+            "input_datetime" -> CommunityMaterial.Icon.cmd_calendar_clock
```
jpelgrom, line 93: This one changes [depending on state](https://github.com/home-assistant/frontend/blob/ffc4ca5b568ad56754967462e9dca01bae2bd70d/src/common/entity/domain_icon.ts#L105-L111) in the frontend

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
```diff
@@ -56,15 +90,31 @@ fun getIcon(entity: Entity<Map<String, Any>>?, domain: String, context: Context)
...
+            "plant" -> CommunityMaterial.Icon2.cmd_flower
+            "proximity" -> CommunityMaterial.Icon.cmd_apple_safari
+            "remote" -> CommunityMaterial.Icon3.cmd_remote
             "scene" -> CommunityMaterial.Icon3.cmd_palette_outline // Different from frontend: outline version
+            "script" -> CommunityMaterial.Icon3.cmd_script_text_outline // Different from frontend: outline version
+            "select" -> CommunityMaterial.Icon2.cmd_format_list_bulleted
+            "sensor" -> CommunityMaterial.Icon.cmd_eye
+            "siren" -> CommunityMaterial.Icon.cmd_bullhorn
+            "simple_alarm" -> CommunityMaterial.Icon.cmd_bell
+            "sun" -> CommunityMaterial.Icon3.cmd_white_balance_sunny
```
jpelgrom, line 117: This one changes [if it is night](https://github.com/home-assistant/frontend/blob/ffc4ca5b568ad56754967462e9dca01bae2bd70d/src/common/entity/domain_icon.ts#L151-L154) in the frontend

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
```diff
@@ -74,7 +124,13 @@ fun getIcon(entity: Entity<Map<String, Any>>?, domain: String, context: Context)
...
                 CommunityMaterial.Icon2.cmd_light_switch
             }
-            else -> CommunityMaterial.Icon.cmd_cellphone
+            "timer" -> CommunityMaterial.Icon3.cmd_timer_outline
+            "updater" -> CommunityMaterial.Icon.cmd_cloud_upload
+            "vacuum" -> CommunityMaterial.Icon3.cmd_robot_vacuum
+            "water_heater" -> CommunityMaterial.Icon3.cmd_thermometer
+            "weather" -> CommunityMaterial.Icon3.cmd_weather_cloudy
+            "zone" -> CommunityMaterial.Icon3.cmd_map_marker_radius
+            else -> CommunityMaterial.Icon.cmd_bookmark
```
jpelgrom, line 133: For consistency I'd say the `CommunityMaterial.Icon.cmd_cellphone`s elsewhere also should be updated.

jpelgrom (commented):
Found some unused code and did a quick check for the icons for simple state-based changes, other than that it looks good to me!

jpelgrom (approved):

Issue #2565, opened 2022-05-30
Unable to view manage sensors, sensor update frequency, manage widgets or persistent connection due to app crash


bendikrb:
**Home Assistant Android version:**
beta-2366-31ab4e82-full

**Android version:**
12

**Phone model:**
Samsung Galaxy S21

**Home Assistant version:**
2022.6.0b2

**Last working Home Assistant release (if known):**
beta-2328-b7428d45

**Description of problem:**
Immediate crash when going in to "Manage sensors" or "Sensor update frequency".

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```
2022-05-31 01:27:11.706: java.lang.RuntimeException: Cannot create an instance of class io.homeassistant.companion.android.settings.SettingViewModel
	at androidx.lifecycle.ViewModelProvider$AndroidViewModelFactory.create(ViewModelProvider.java:269)
	at androidx.lifecycle.SavedStateViewModelFactory.create(SavedStateViewModelFactory.java:112)
	at androidx.lifecycle.ViewModelProvider.get(ViewModelProvider.java:185)
	at androidx.lifecycle.ViewModelProvider.get(ViewModelProvider.java:150)
	at androidx.lifecycle.ViewModelLazy.getValue(ViewModelProvider.kt:54)
	at androidx.lifecycle.ViewModelLazy.getValue(ViewModelProvider.kt:41)
	at io.homeassistant.companion.android.settings.sensor.SensorUpdateFrequencyFragment.getViewModel(SensorUpdateFrequencyFragment.kt:22)
	at io.homeassistant.companion.android.settings.sensor.SensorUpdateFrequencyFragment$onCreateView$1$1$1.invoke(SensorUpdateFrequencyFragment.kt:46)
	at io.homeassistant.companion.android.settings.sensor.SensorUpdateFrequencyFragment$onCreateView$1$1$1.invoke(SensorUpdateFrequencyFragment.kt:45)
	at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.jvm.kt:107)
	at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.jvm.kt:34)
	at androidx.compose.runtime.CompositionLocalKt.CompositionLocalProvider(CompositionLocal.kt:228)
[...]

PR #2575, opened 2022-06-01
Fix issues with initing view models.

JBassett:
## Summary

There was an issue with missing annotations needed by hilt.  Really fixes: https://github.com/home-assistant/android/issues/2565


dshokouhi (approved):
Gave the debug APK a quick test, looks to be ok

Issue #2582, opened 2022-06-02
Removed important hint in minimal build


z3r0r4:
# The Issue 

The String websocket_setting_description previously contained a hint that:

> if you are on a minimal build this needs to be always to consistently get push notifications.

https://github.com/home-assistant/android/blob/9861462a1732fc8e372c7d60db4b088a5d9a0c26/common/src/main/res/values/strings.xml#L749

Which was removed:
https://github.com/home-assistant/android/blob/dcfc590458467980945d83ab754d4ac9d09c13cd/common/src/main/res/values/strings.xml#L798

### Arguments
This caused me about an hour of debugging to understand why I couldnt receive push notifications on 2022.3.0-minimal.
It was caused by not setting the Persistent Connection to _Always_, which I only found out by downgrading and looking around until I saw the hint in 2022.2.0-minimal.

Future confusion and Issues as to why the notifications are not working will be avoided if the hint is added again.

# The Solution
Add a hint to the requirement again.

dshokouhi:
The hint was actually moved to the actual setting that is required for notifications to work all the time.  Also the old text was shown to all users, this new string is only shown to minimal users.

https://github.com/home-assistant/android/blob/0094f9163afa6a6f23da2b74826fb210a83aa449/common/src/main/res/values/strings.xml#L804

dshokouhi:
also the default setting for minimal users was always on "Always" so you may have changed that at some point in time

z3r0r4:
> The hint was actually moved to the actual setting that is required for notifications to work all the time. Also the old text was shown to all users, this new string is only shown to minimal users.
> 
> https://github.com/home-assistant/android/blob/0094f9163afa6a6f23da2b74826fb210a83aa449/common/src/main/res/values/strings.xml#L804

Atleast for me this didn't work/ suffice.
I also do not think that 

> Required to always receive notifications

sufficiently implies that its necessary for the minimal installation.
 
![Screenshot 2022.3.0-minimal](https://user-images.githubusercontent.com/38572382/171880264-e585d550-004b-499e-a2d2-bbb8b9d41251.jpg)

Thus there are no hints on this setting page that imply that the connection needs to be set to _Always_ to receive notifications on the minimal build.

Please note if I'm missing something.

z3r0r4:
> also the default setting for minimal users was always on "Always" so you may have changed that at some point in time

I did change it a long time ago as many casual users would due to battery concerns.
[...]

PR #2584, opened 2022-06-03
Use minimal specific string for home wifi option on persistent connection

dshokouhi:
## Summary

Fixes: #2582 to remove mention of Google so the text is more clear about whats required or not for when notifications can be received.

Making sure not to mention the minimal flavor but also making it clear how each option is intended to work.


common/src/main/res/values/strings.xml
```diff
@@ -815,6 +815,7 @@
     <string name="remote_debugging">WebView Remote Debugging</string>
     <string name="remote_debugging_summary">Allow remote debugging of WebView to troubleshoot Home Assistant frontend issues</string>
     <string name="websocket_setting_home_wifi">On Home WiFi Only\n\nNotifications will be delivered by Google only when not on Home WiFi.</string>
+    <string name="websocket_setting_home_wifi_minimal">On Home WiFi Only\n\nNotifications will only be delivered when on Home WiFi.</string>
```
jpelgrom, line 818: For consistency with other descriptions for the minimal version, maybe use "received" instead of "delivered"?

JBassett (approved):

PR #2585, opened 2022-06-04
Fix removing non-last list items from Last Update Trigger sensor's settings

marazmarci:
Fixes #2550 _Cannot add new Intent to Last Update Trigger sensor_

## Summary
Now the `Last Update Trigger` sensor's settings screen handles removing non-last list items correctly.

The fix is intended to be backwards compatible, so existing installations with incorrect state will be fixed automatically upon the first call of the `sendLastUpdate()` method (= by an update in the background, or when the user edits the list).

## Screen recordings
[video before the fix: reproduction of the bug](https://user-images.githubusercontent.com/1349654/171994081-0fede94b-c223-46db-93b6-f2c23244b48a.mp4)

[video after the fix: the bug cannot be reproduced anymore](https://user-images.githubusercontent.com/1349654/171994305-86835a33-8b1d-4f8f-a1ff-ada6da129341.mp4)

## Any other notes
I did some refactoring in the `sendLastUpdate()` method to clarify and optimize things a bit, but I tried to not go too far to keep the changes focused on the bugfix.

Useful command for testing if `Intent`s are received:
`adb shell am broadcast -a com.example.test1`
(don't forget to force close and reopen the app after adding new `Intent`s to the list!)

PR #2593, opened 2022-06-07
Update delete methods in daos

NotWoods:
## Summary
Some additional work to make Dao methods run non-blocking using coroutines. Converted some methods to use coroutines and removed some unused methods.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt
@@ -76,9 +76,12 @@ class LastUpdateManager : SensorManager {
         } else {
             sensorDao.add(SensorSetting(lastUpdate.id, SETTING_ADD_NEW_INTENT, "false", SensorSettingType.TOGGLE))
         }
-        for (setting in allSettings) {
-            if (setting.value == "")
-                sensorDao.removeSetting(lastUpdate.id, setting.name)
+
+        val settingsToRemove = allSettings
+            .filter { setting -> setting.value == "" }
+            .map { setting -> setting.name }
+        if (settingsToRemove.isNotEmpty()) {
+            sensorDao.removeSettings(lastUpdate.id, settingsToRemove)
         }
     }
 }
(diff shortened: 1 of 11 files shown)


common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt
```diff
@@ -76,9 +76,12 @@ class LastUpdateManager : SensorManager {
...
         }
-        for (setting in allSettings) {
-            if (setting.value == "")
-                sensorDao.removeSetting(lastUpdate.id, setting.name)
+
+        val settingsToRemove = allSettings
+            .filter { setting -> setting.value == "" }
+            .map { setting -> setting.name }
+        if (settingsToRemove.isNotEmpty()) {
+            sensorDao.removeSettings(lastUpdate.id, settingsToRemove)
```
marazmarci, line 84: This change will cause a conflict with my open PR: #2585 _Fix removing non-last list items from Last Update Trigger sensor's settings_

See [these lines](https://github.com/home-assistant/android/blob/f8f32951447caef3de1ce739e976313365f12d2e/common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt#L68-L74) in my changes

We should decide on who and how to resolve it.

marazmarci: This change will cause a conflict with my open PR: #2585 _Fix removing non-last list items from Last Update Trigger sensor's settings_

See [these lines](https://github.com/home-assistant/android/blob/f8f32951447caef3de1ce739e976313365f12d2e/common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt#L68-L74) in my changes.

We should decide on who and how to resolve it.
I think it would be easier to solve the conflict in your PR because I have several commits which would be harder to rebase. And my PR is already reviewed (but not yet formally approved) by one person (@jpelgrom).

marazmarci: You're right. In this case, I can take care of it, and rebase my branch when this PR is merged.

marazmarci: Then I'll mark my PR as `WIP`, so it won't get merged accidentally in the meantime.

jpelgrom (approved):

PR #2633, opened 2022-06-22
Bump wear compose to 1.0.0-rc02

leroyboerefijn:
## Summary
Bump androidx.wear.compose to version 1.0.0-rc02

This has a bunch of improvements behind the scenes and makes the app up to date with the latest API

Also updated the loading view with a spinner


jpelgrom:
It looks like there is now another version, 1.0.0-rc02 :)

dshokouhi:
> It looks like there is now another version, 1.0.0-rc02 :)

well it is a Wednesday :P

jpelgrom:
If you have selected any favorites, the 'loading' screen still looks like it did before (no circle spinning), is that intentional or did you forget to change it?

Something also appears to have changed in the rendering, as it looks subtly bolder and more padded. Not necessarily a bad thing but it no longer perfectly matches the system UI 🙁 (due to [this change](https://android-review.googlesource.com/c/platform/frameworks/support/+/1995450) and [this change](https://android-review.googlesource.com/c/platform/frameworks/support/+/2020417)
|Stable|This PR|
|-----|-----|
|wear-home-stable|wear-home-pull|

leroyboerefijn:
> If you have selected any favorites, the 'loading' screen still looks like it did before (no circle spinning), is that intentional or did you forget to change it?

Oh I missed that in testing! I'll have a look.

> Something also appears to have changed in the rendering, as it looks subtly bolder and more padded. Not necessarily a bad thing but it no longer perfectly matches the system UI 🙁

Yes, I think they adapted a number of things to match the latest material design guidelines. Didn't notice this level of detail by the way haha. OK to leave it like this?

leroyboerefijn:
So apparently we have a loading view in two places: `MainView` and `HomeView`. But I don't see why we would show a different loading view if we don't have favorites. I propose to have just one loading view in `MainView` and then show the favorites placeholder depending on whether favorites exist.

jpelgrom:
> Didn't notice this level of detail by the way haha. OK to leave it like this?

I noticed something was 'off' and then started comparing. HA mostly follows Material Design so definitely OK, just an observation!

> Or am I missing a specific edge case?

That would make loading invisible if you are somehwere else than home, but I don't think there are any deeplinks to the app so it should be fine. There is another loading chip in `EntityViewList` that seems unreachable to me.

leroyboerefijn:
> That would make loading invisible if you are somehwere else than home, but I don't think there are any deeplinks to the app so it should be fine.

Don't think so either!

> There is another loading chip in `EntityViewList` that seems unreachable to me.

Hmm that seems unreachable indeed. Even if it's reachable, the check doesn't necessarily mean that we're still loading. Could also be that the filtered list is empty. Maybe this should throw an error instead?

Doubting a bit whether to make this a spinner as well, let it throw an error, remove it, or leave it be 😅

JBassett (approved):

PR #2635, opened 2022-06-23
Only apply prioritize internal to WebView when location is disabled on the device

dshokouhi:
## Summary

We have been seeing a lot of users having issues with loading the HA frontend. In most of those cases it seems users have enabled Prioritize Internal even when they don't typically leave location off. This is not a big issue when using API calls because we silently fail and try other URLs. When it comes to WebView we can't reliably detect when it fails to load and the feature prioritize internal is for our users who actually leave location off on their device typically and are ok with the needing to refresh the window. This ends up causing users confusion as to why HA is not loading.

Now we will only try the internal URL if location is disabled on the device and prioritize internal is enabled, otherwise if location is enabled the app will continue to function as it does which includes background permission checks. Users will need to first grant the app location permissions in order to enable this feature so the check is only to ensure location is enabled.

I ran some tests locally with location on/off and wifi on/off and it seems to do URL switching as expected along with the prioritize internal option.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
@@ -9,6 +9,7 @@ import io.homeassistant.companion.android.common.data.authentication.Authenticat
 import io.homeassistant.companion.android.common.data.authentication.SessionState
 import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
 import io.homeassistant.companion.android.common.data.url.UrlRepository
+import io.homeassistant.companion.android.common.util.DisabledLocationHandler
 import io.homeassistant.companion.android.util.UrlHandler
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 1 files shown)


dshokouhi:
> If you doubt people read it they probably only read the title. When only reading the title, "Prioritize Internal URL" suggests that the app will _always_ use the external URL first _unless_ you enable the setting. So to really fix it the setting should probably be renamed instead of expanding the description.

Yea the name probably should be changed  but its tough to determine what would be suitable given the description.



> Suggestion: move it to the Home WiFi Network dialog to reduce the chance someone enables it (or perhaps the dialog should be a page instead). Below the network list, add the setting and change from a toggle to a dropdown "When unable to determine if connected to Home WiFi network": "Use Home Assistant URL first (recommended)" (default) or "Use internal connection URL first (use this if you typically turn location access off)" (= current setting prioritize internal). I think this will make it clearer that most people shouldn't change it, and for the people that turn location services off it is still accessible.

I had thought about moving it, this PR was really more for a quick fix solution to the issue why is it not working. While we can move it and prevent new users from enabling it, existing users who have not complained will still continue to see the feature as broken.  Thats why I wanted to first fix the heart of the issue where the page doesn't load as you would expect.



> That would be nice for code quality/consistency, but most people won't notice the difference because the app loops so does it matter?

Exactly, plus to add the location call to `getApiUrls` would involve needing to pull in `context` from a lot of different places. Given that its a silent failure I don't expect many users to even notice a difference here unless you look at the logs. If the WebView were able to reliably switch URLs we probably wouldn't have any complaints from users on that subject.

My thoughts were with the changes here users wouldn't notice it being broken as the switching will happen as they would expect.

JBassett (approved):

PR #2638, opened 2022-06-24
Fix Storage sensors to update continuously and display correct size unit

### LINKED ISSUE #2637: Internal Storage sensor only updating once on app start
**Home Assistant Android version:** beta-2422-766e879d-full

**Android version:** 12

**Phone model:** S20FE 5G, S21 5G

**Home Assistant version:** 2022.6.6

**Last working Home Assistant release (if known):** Judging based on the code, it's been like this since the feature's introduction.

**Description of problem:** The Internal Storage sensor should be updated regularly. Instead, it's only updated on app start, because the values are stored in a `companion object` of `StorageSensorManager`, and only calculated once when the `companion object` is being initialized.

```

```

**Screenshot of problem:** 

**Additional information:**

marazmarci:
Fixes #2637 

## Summary

The state of the Internal Storage sensor was stored in the `companion object` of `StorageSensorManager`. It was calculated once when the `companion object` was initialized. I fixed it and also did some refactoring.

Based on the code, the Free and the Total attributes' unit was supposed to be dynamic (KB / MB / GB), but it always displayed GB because the logic was flawed. Now it can not only output 0GB for smaller values (<1GB) but also 345MB or 640KB for example.

## Any other notes

This app may come in handy for testing this:
[Fill Storage – Google Play](https://play.google.com/store/apps/details?id=com.arumcomm.fillstorage)

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/StorageSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/StorageSensorManager.kt
@@ -33,19 +33,11 @@ class StorageSensorManager : SensorManager {
             stateClass = SensorManager.STATE_CLASS_MEASUREMENT,
             entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC
         )
-        val path: File = Environment.getDataDirectory()
-        private val stat = StatFs(path.path)
-        var availableBlocks = stat.availableBlocksLong
-        var blockSize = stat.blockSizeLong
-        var totalBlocks = stat.blockCountLong
-        var blockSizeSD = 0L
-        var availableBlocksSD = 0L
-        var totalBlocksSD = 0L
-        private var externalPath: File? = null
 
-        private fun externalMemoryAvailable(context: Context): File? {
+        private fun getExternalStoragePathIfAvailable(context: Context): File? {
             val pathsSD = context.getExternalFilesDirs(null)
             var removable: Boolean
+            var externalPath: File? = null
             Log.d(TAG, "PATHS SD ${pathsSD.size}")
             for (item in pathsSD) {
                 if (item != null) {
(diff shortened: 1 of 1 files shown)


jpelgrom:
Please don't change the line endings from `CRLF` to `LF`, Git will consider all lines to have changed (as you can also see on GitHub) which results in a messy history

marazmarci:
@jpelgrom I changed the line endings back to `CRLF`.

app/src/main/java/io/homeassistant/companion/android/sensors/StorageSensorManager.kt
```diff
@@ -113,29 +104,19 @@ class StorageSensorManager : SensorManager {
...
             context,
             externalStorage,
-            percentFreeExternal,
+            externalStorageStats?.percentage ?: 0,
             externalStorage.statelessIcon,
             mapOf(
-                "free_external_storage" to freeExternalStorage,
-                "total_external_storage" to totalExternalStorage
+                "free_external_storage" to (externalStorageStats?.freeBytes ?: "No SD Card"),
+                "total_external_storage" to (externalStorageStats?.totalBytes ?: "No SD Card")
```
marazmarci, line 119: Off-topic:
These attribute names don't match the ones written in the [documentation](https://companion.home-assistant.io/docs/core/sensors/#storage-sensor)

dshokouhi: We should update the documentation then, I thought we had updated this a while back but may have missed the documentation. Attribute names should be all lower case with underscores for best HA core compatibility. There may or may not be other places to correct now :)

app/src/main/java/io/homeassistant/companion/android/sensors/StorageSensorManager.kt
```diff
@@ -125,29 +125,21 @@ class StorageSensorManager : SensorManager {
...
-        var commaOffset = resultBuffer.length - 3
-        while (commaOffset > 0) {
-            resultBuffer.insert(commaOffset, ',')
-            commaOffset -= 3
-        }
-
-        resultBuffer.append(suffix)
-        return resultBuffer.toString()
+        val sizeWithThousandsSeparator = String.format("%,d", sizeLong)
+        return "$sizeWithThousandsSeparator$suffix"
```
marazmarci, line 142: I just found that this method needs some improvement, but it isn't strictly related to this PR's scope.

Is it OK to have them in this PR or should I create a new PR for them?

jpelgrom: I think it's fine to update the scope from "fix updating the storage sensors" to "fix the storage sensors", strictly speaking refactoring is already out of scope 😉

marazmarci:
I edited the PR's name and description and also reworded the commit message of 1f9e57f8 (new: 6980a2e9).

app/src/main/java/io/homeassistant/companion/android/sensors/StorageSensorManager.kt
marazmarci: I edited the PR's name and description (and also reworded the commit message of 1f9e57f8 → new: 6980a2e9)

JBassett (approved):

Issue #2551, opened 2022-05-29
WearOS template tile html tags


chertvl:
Is it possible to add HTML Tags feature to template tile for WearOS app? It would be great to stylize my favorite feature a little bit (font color, bold/italic, etc, like it realized in Template widget for Android).

Thanks!
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered, if any**

**Additional context**

jpelgrom:
Based on a quick search and some experimentation, I don't think this is feasible.

The function used right now for text [only accepts a plain String and settings for style (no mixed styling)](https://developer.android.com/reference/kotlin/androidx/wear/tiles/LayoutElementBuilders.Text.Builder?hl=en), and there isn't really a good way to convert HTML to [the required format that accepts multiple 'text parts with styling'](https://developer.android.com/reference/kotlin/androidx/wear/tiles/LayoutElementBuilders.Spannable.Builder?hl=en).

Issue #2157, opened 2022-01-18
Template Tile null pointer exception on non-null parameter renderedTemplate


dshokouhi:
**Home Assistant Android version:**
io.homeassistant.companion.android@beta-1916-8a8772d5

**Android version:**
11

**Phone model:**
SM-G970F

**Home Assistant version:**
Not sure but probably not important here

**Last working Home Assistant release (if known):**
N/A new feature

**Description of problem:**

Looks to be an unhandled exception here according to Sentry so I imagine the app crashes on this device when the user first navigates to the template tile screen in phone settings.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```
java.lang.NullPointerException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkNotNullParameter, parameter renderedTemplate
    at io.homeassistant.companion.android.settings.wear.views.SettingsWearTemplateTileKt.SettingsWearTemplateTile
    at io.homeassistant.companion.android.settings.wear.views.SettingsWearHomeViewKt$LoadSettingsHomeView$1$1$3.invoke(SettingsWearHomeView.kt:35)
    at io.homeassistant.companion.android.settings.wear.views.SettingsWearHomeViewKt$LoadSettingsHomeView$1$1$3.invoke(SettingsWearHomeView.kt:34)
    at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.jvm.kt:116)
    at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.jvm.kt:34)
    at androidx.navigation.compose.NavHostKt$NavHost$4$1.invoke(NavHost.kt:150)
    at androidx.navigation.compose.NavHostKt$NavHost$4$1.invoke(NavHost.kt:149)
    at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.jvm.kt:107)
    at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.jvm.kt:34)
    at androidx.compose.runtime.CompositionLocalKt.CompositionLocalProvider(CompositionLocal.kt:228)
    at androidx.compose.runtime.saveable.SaveableStateHolderImpl.SaveableStateProvider(SaveableStateHolder.kt:84)
[...]

PR #2653, opened 2022-07-01
Implement html styling on the template tile

leroyboerefijn:
## Summary

Parse HTML for the template tile, similar to the template widget. Supports the following styling:

- bold/italic/underline
- color
- size (with \<big> and \<small> tags)
- headers (\<h1>, \<h2>, etc.)

The following snippet will result in the screenshot:
```html
This <b>is <i>an</i> example</b><br><big>Of HTML</big><br>Of HTML<br><span style='color:#03a9f4'><small>And markups</small></span>
```

**Breaking change: new lines in the form of `\n` are no longer supported. Need to use `<br>` for a new line**

## Screenshots

Screenshot_20220701_103651_sysui

Screenshot_20220701-104000

Documentation: home-assistant/companion.home-assistant#767

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTile.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTile.kt
@@ -119,7 +119,7 @@ fun SettingsWearTemplateTile(
 }
 
 private fun parseHtml(renderedText: String) = buildAnnotatedString {
-    val renderedSpanned = fromHtml(renderedText, FROM_HTML_MODE_LEGACY)
+    val renderedSpanned = fromHtml(renderedText.replace("\n", "<br>\n"), FROM_HTML_MODE_LEGACY)
     append(renderedSpanned.toString())
     renderedSpanned.getSpans(0, renderedSpanned.length, CharacterStyle::class.java).forEach { span ->
         val start = renderedSpanned.getSpanStart(span)
--- a/app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTile.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTile.kt
@@ -1,5 +1,12 @@
 package io.homeassistant.companion.android.settings.wear.views
 
+import android.graphics.Typeface
+import android.text.style.AbsoluteSizeSpan
+import android.text.style.CharacterStyle
+import android.text.style.ForegroundColorSpan
+import android.text.style.RelativeSizeSpan
+import android.text.style.StyleSpan
+import android.text.style.UnderlineSpan
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
@@ -19,12 +26,20 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.ColorFilter
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.colorResource
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.SpanStyle
+import androidx.compose.ui.text.buildAnnotatedString
+import androidx.compose.ui.text.font.FontStyle
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
+import androidx.core.text.HtmlCompat.FROM_HTML_MODE_LEGACY
+import androidx.core.text.HtmlCompat.fromHtml
 import com.mikepenz.iconics.compose.Image
 import com.mikepenz.iconics.typeface.library.community.material.CommunityMaterial
 import io.homeassistant.companion.android.util.IntervalToString
@@ -47,8 +62,8 @@ fun SettingsWearTemplateTile(
                 docsLink = WEAR_DOCS_LINK
             )
         }
-    ) {
-        Column(Modifier.padding(all = 16.dp)) {
+    ) { padding ->
+        Column(Modifier.padding(padding).padding(all = 16.dp)) {
             Row(verticalAlignment = Alignment.CenterVertically) {
                 Image(
                     asset = CommunityMaterial.Icon3.cmd_timer_cog,
@@ -95,10 +110,33 @@ fun SettingsWearTemplateTile(
                 maxLines = 10
             )
             Text(
-                renderedTemplate,
+                parseHtml(renderedTemplate),
                 fontSize = 12.sp,
                 modifier = Modifier.padding(top = 8.dp)
             )
+        }
+    }
+}
+
+private fun parseHtml(renderedText: String) = buildAnnotatedString {
+    val renderedSpanned = fromHtml(renderedText, FROM_HTML_MODE_LEGACY)
+    append(renderedSpanned.toString())
+    renderedSpanned.getSpans(0, renderedSpanned.length, CharacterStyle::class.java).forEach { span ->
+        val start = renderedSpanned.getSpanStart(span)
+        val end = renderedSpanned.getSpanEnd(span)
+        when (span) {
+            is AbsoluteSizeSpan -> addStyle(SpanStyle(fontSize = span.size.sp), start, end)
+            is ForegroundColorSpan -> addStyle(SpanStyle(color = Color(span.foregroundColor)), start, end)
+            is RelativeSizeSpan -> {
+                val defaultSize = 12
+                addStyle(SpanStyle(fontSize = (span.sizeChange * defaultSize).sp), start, end)
+            }
+            is StyleSpan -> when (span.style) {
+                Typeface.BOLD -> addStyle(SpanStyle(fontWeight = FontWeight.Bold), start, end)
+                Typeface.ITALIC -> addStyle(SpanStyle(fontStyle = FontStyle.Italic), start, end)
+                Typeface.BOLD_ITALIC -> addStyle(SpanStyle(fontWeight = FontWeight.Bold, fontStyle = FontStyle.Italic), start, end)
+            }
+            is UnderlineSpan -> addStyle(SpanStyle(textDecoration = TextDecoration.Underline), start, end)
         }
     }
 }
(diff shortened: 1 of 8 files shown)


app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTile.kt
```diff
@@ -47,8 +62,8 @@ fun SettingsWearTemplateTile(
                 docsLink = WEAR_DOCS_LINK
             )
         }
-    ) {
-        Column(Modifier.padding(all = 16.dp)) {
+    ) { padding ->
+        Column(Modifier.padding(padding).padding(all = 16.dp)) {
```
leroyboerefijn, line 66: If not using the padding of the `Scaffold`, then the entire block becomes red in AS. However, the `Scaffold` padding does not give the right margin all around...

common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
```diff
@@ -167,7 +167,7 @@ class IntegrationRepositoryImpl @Inject constructor(
                         "render_template",
                         mapOf("template" to Template(template, variables))
                     )
-                ).getValue("template")
+                ).getValue("template") ?: throw NullPointerException("Null returned in template")
```
leroyboerefijn, line 170: I ran into an error I didn't see before. The app would crash when HA returns null for the template. According to AS this can never be null, but still this change was required to stop the app from crashing

wear/src/main/java/io/homeassistant/companion/android/tiles/TemplateTile.kt
```diff
@@ -152,4 +163,60 @@ class TemplateTile : TileService() {
...
+                            is ForegroundColorSpan -> setColor(
+                                ColorBuilders.ColorProp.Builder()
+                                    .setArgb(it.foregroundColor)
+                                    .build()
+                            )
+                            is RelativeSizeSpan -> {
+                                val defaultSize = 16 // https://developer.android.com/training/wearables/design/typography
+                                setSize(
+                                    DimensionBuilders.SpProp.Builder()
+                                        .setValue(it.sizeChange * defaultSize)
```
leroyboerefijn, line 192: Would love to do this more smartly, with real relative sizes. Then I won't have to assume the default size and it would allow for nested size tags. Any tips are welcome :)

jpelgrom:
You did the manual work to implement #2551! ❤️

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTile.kt
jpelgrom: I believe this was a change in lint, there are lots of `Scaffold`s without using the padding in the codebase and they all show up in red now (but you can still compile the app without problems).

A cleaner solution while stil ensuring that the required `Scaffold` padding is used would be:
- calculating the new totals using `calculateStartPadding` etc from the parent and then `coerceAtLeast(16)` for all 4 sides, for example: `padding.calculateStartPadding(LocalLayoutDirection.current).value.coerceAtLeast(16f).dp`
- switching to a `LazyColumn` and using the parent value for the modifier, and setting your existing padding of 16 dp as the `contentPadding`

common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
jpelgrom: Do you have an example of a template that returns `null`?

If null is a valid result from the server, wouldn't it make more sense to change the return type to `String?`?

leroyboerefijn: Yes I do. I store my content in a binary sensor attribute for easier editing. So in my tile settings I just use something like

`{{ state_attr('binary_sensor.watch_template_test', 'template_text') }}`

I received the null when I started to add the `_test` suffix. So I got a non-existing entity id: `binary_sensor.watch_template_`

So it seems to return null when the template is invalid and/or when using an invalid entity id. Not sure on the exact conditions. I'm also fine with returning `String?` and handling the null elsewhere. However, the `getValue` seems to return a non-nullable `String`.

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTile.kt
jpelgrom: 'Stacking' modifiers like this works so I'd say the option you've already implemented is a good fit for this screen 👍

leroyboerefijn: Alright! Thanks for you feedback 🙂

common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
jpelgrom: It looks like the server actually returns `null` while indicating it was succesful, so I'd keep that and not throw an error. The developer tools in the frontend also show 'null' instead of an error.

```
2022-07-01 13:35:19.210 15428-15566/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: --> POST https://[domain]/api/webhook/[webhook id]
2022-07-01 13:35:19.210 15428-15566/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: Content-Type: application/json; charset=UTF-8
2022-07-01 13:35:19.210 15428-15566/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: Content-Length: 148
2022-07-01 13:35:19.210 15428-15566/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: {"type":"render_template","data":{"template":{"template":"{{ state_attr('binary_sensor.watch_', 'template_text') }} ","variables":{}}}}
2022-07-01 13:35:19.210 15428-15566/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: --> END POST (148-byte body)
2022-07-01 13:35:19.242 15428-15566/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: <-- 200 OK https://[domain]/api/webhook/[webhook id] (31ms)
2022-07-01 13:35:19.242 15428-15566/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: Content-Type: application/json; charset=utf-8
2022-07-01 13:35:19.242 15428-15566/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: Content-Length: 18
2022-07-01 13:35:19.242 15428-15566/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: Date: Fri, 01 Jul 2022 11:35:19 GMT
2022-07-01 13:35:19.242 15428-15566/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: Server: Python/3.9 aiohttp/3.8.1
2022-07-01 13:35:19.243 15428-15566/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: {"template": null}
2022-07-01 13:35:19.243 15428-15566/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: <-- END HTTP (18-byte body)
```

> However, the getValue seems to return a non-nullable `String`

You can use `getOrDefault("template", null)`.

leroyboerefijn: So `getValue` either returns the value or an exception, while `get` will return the value or null if it doesn't exist. So it seems that switching to `get` (or `getOrDefault`) will change the behavior, in that we won't get an exception if the return value didn't contain the key, but just get null instead.

I think we will then still get an exception when we cannot connect to the server, so we can show a different message for when catching an exception versus receiving `null`. Does that make sense to implement?

jpelgrom: But the key does exist (see the server response in my previous comment), so does it correctly recognize it as `null` or throw an exception because you're trying to cast null to a non-nullable String?

> I think we will then still get an exception when we cannot connect to the server, so we can show a different message for when catching an exception versus receiving `null`. Does that make sense to implement?

Yes, you're correct and good idea!

app/src/main/java/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -216,10 +216,10 @@ class TemplateWidgetConfigureActivity : BaseWidgetConfigureActivity() {
             var enabled: Boolean
             withContext(Dispatchers.IO) {
                 try {
-                    templateText = integrationUseCase.renderTemplate(template, mapOf())
+                    templateText = integrationUseCase.renderTemplate(template, mapOf()) ?: getString(commonR.string.template_error)
                     enabled = true
                 } catch (e: Exception) {
-                    templateText = "Error in template"
+                    templateText = getString(commonR.string.error_connection_failed)
```
jpelgrom, line 222: I'm not sure this is a good error message. When starting to type brackets on a phone, after typing "{{" the preview shows "Unable to connect to Home Assistant." while the problem isn't related to the network but to the template text.

Server response for "{{" is: `{"template": {"error": "TemplateSyntaxError: unexpected 'end of template'"}}`

leroyboerefijn: Oh, didn't test it properly then. And annoying that it sometimes gives an error message and sometimes null... I'll update it somewhere this weekend

common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
leroyboerefijn: Btw, seem related to #2157

leroyboerefijn:
Added support for new line characters. @JBassett maybe this makes it easier to merge :)

jpelgrom:
`MessagingManager` has a regex for all kinds of newline characters, you could use that to support even more:
https://github.com/home-assistant/android/blob/f9ceeb0d651074675ef833f23040af8e0a53aca9/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt#L1225-L1226

JBassett (approved):

fabcop:
Hi there, new here, can someone show how to view a sensor state in WearOs template?

jpelgrom:
You can use [template syntax](https://www.home-assistant.io/docs/configuration/templating/) like `{{ states('sensor.something') }}`.

Old PRs aren't the place for these questions. In the future, try asking on the forums :)

PR #2663, opened 2022-07-08
Add camera device control

jpelgrom:
## Summary

Add support for `camera` entities to device controls. The control is only supported on Android 12+.

## Screenshots

As usual with device controls, the appearance depends on the manufacturer. These screenshots are for Google's implementation, note that if the camera isn't active they decided to not show the snapshot:
|State: `idle`|State: `recording` or `streaming`|
|-----|-----|
|A device control tile with a camera icon, the state 'Idle', the title 'RTSP camera' and a gray background|A device control tile with a camera icon and the state 'Streaming', no title is shown and the background of the tile is a still image of the stream|

Documentation: home-assistant/companion.home-assistant#769

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/controls/CameraControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/CameraControl.kt
@@ -49,7 +49,7 @@ object CameraControl : HaControl {
         val icon = if (image != null) {
             Icon.createWithBitmap(image)
         } else {
-            Icon.createWithResource(context, R.drawable.ic_sync_problem)
+            Icon.createWithResource(context, R.drawable.control_camera_placeholder)
         }
         control.setControlTemplate(
             ThumbnailTemplate(
--- a/app/src/main/java/io/homeassistant/companion/android/controls/CameraControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/CameraControl.kt
@@ -1,0 +1,92 @@
+package io.homeassistant.companion.android.controls
+
+import android.content.Context
+import android.graphics.Bitmap
+import android.graphics.BitmapFactory
+import android.graphics.drawable.Icon
+import android.os.Build
+import android.service.controls.Control
+import android.service.controls.DeviceTypes
+import android.service.controls.actions.ControlAction
+import android.service.controls.templates.ThumbnailTemplate
+import android.util.Log
+import androidx.annotation.RequiresApi
+import io.homeassistant.companion.android.R
+import io.homeassistant.companion.android.common.data.integration.Entity
+import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
+import io.homeassistant.companion.android.common.data.websocket.impl.entities.AreaRegistryResponse
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
+import kotlinx.coroutines.withTimeoutOrNull
+import java.net.URL
+import java.util.concurrent.TimeUnit
+import io.homeassistant.companion.android.common.R as commonR
+
+@RequiresApi(Build.VERSION_CODES.S)
+object CameraControl : HaControl {
+    private const val TAG = "CameraControl"
+
+    override fun provideControlFeatures(
+        context: Context,
+        control: Control.StatefulBuilder,
+        entity: Entity<Map<String, Any>>,
+        area: AreaRegistryResponse?,
+        baseUrl: String?
+    ): Control.StatefulBuilder {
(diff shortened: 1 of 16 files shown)


app/src/main/java/io/homeassistant/companion/android/controls/CameraControl.kt
```diff
@@ -0,0 +1,92 @@
...
+        integrationRepository: IntegrationRepository,
+        action: ControlAction
+    ): Boolean {
+        // No action is received, Android immediately invokes long press
+        return true
+    }
+
+    private fun getThumbnail(path: String): Bitmap? = runBlocking {
+        var image: Bitmap? = null
+        withTimeoutOrNull(TimeUnit.SECONDS.toMillis(2)) {
```
dshokouhi, line 81: So I've been testing the debug APK with a few different cameras. Some cameras I think take a bit longer to respond and the image doesn't get downloaded (there is no error) but when I look at the token URL the image loads. Should we consider increasing this timeout? I am testing at home so it could be slower on mobile data too?

Whats strange is that the cameras with the state `streaming` load the image while the ones listed as `idle` do not load even though the URL given in the attributes loads an image.  Here is a state output example of one where the image doesn't load but the URL in a browser loads an image.

```
2022-07-08 10:25:15.279 18377-20320/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: {"entity_id": "camera.frontyard_driveway", "state": "idle", "attributes": {"access_token": "TOKEN", "frontend_stream_type": "hls", "entity_picture": "/api/camera_proxy/camera.frontyard_driveway?token=TOKEN", "friendly_name": "Frontyard Driveway", "supported_features": 2}, "last_changed": "2022-06-25T22:36:59.523680+00:00", "last_updated": "2022-07-08T17:22:17.124588+00:00", "context": {"id": "965ca585ef4a57d848453d8fcc128371", "parent_id": null, "user_id": null}}
```

For sanity sake here is the state output for a working camera:

```
2022-07-08 10:25:15.199 18377-20320/io.homeassistant.companion.android.debug I/okhttp.OkHttpClient: {"entity_id": "camera.den_camera", "state": "streaming", "attributes": {"access_token": "TOKEN", "model_name": "IP2M-841B", "brand": "Amcrest", "motion_detection": true, "frontend_stream_type": "hls", "audio": "on", "motion_recording": "on", "color_bw": "auto", "entity_picture": "/api/camera_proxy/camera.den_camera?token=TOKEN", "friendly_name": "Den Camera", "supported_features": 3}, "last_changed": "2022-07-08T05:03:57.796163+00:00", "last_updated": "2022-07-08T17:22:17.128454+00:00", "context": {"id": "4028f79c0afed2a9c4c26da096b3b97b", "parent_id": null, "user_id": null}}
```

image

jpelgrom: > while the ones listed as idle do not load

As mentioned right before the screenshots, Google seems to have decided to not show an image if the camera state is `idle`. The app will always provide the image. If you don't see any errors in the logs the image has been downloaded.

We could 'lie' and always say that the camera is active, but I'm not sure about that. 

> Should we consider increasing this timeout?

The timeout was added because right now controls are loaded sequentially, and images must be added when creating the controls. As a result adding lots of cameras could significantly increase the loading time, especially if you have to wait for the default network timeout (10s) for each control.

If you're worried that it is too short, it could be increased to another value, or I could work on loading the controls asynchronously but that will create more changes.

dshokouhi: > As mentioned right before the screenshots, Google seems to have decided to not show an image if the camera is inactive (I've mapped this to the state `idle`). The app will always provide the image. If you don't see any errors in the logs the image has been downloaded.

ah totally missed that part, oops 🤦‍♂️ 



> We could 'lie' and always say that the camera is active, but I'm not sure about that.

yea its fine, we should match our states and stick to what is supported natively.



> The timeout was added because right now controls are loaded sequentially, and images must be added when creating the controls. As a result adding lots of cameras could significantly increase the loading time, especially if you have to wait for the default network timeout (10s) for each control.

ok for some reason I thought the 2 seconds overrided that :) thanks for clearing it up. I also just tested on mobile and the images still loaded but not sure how it fairs in really low data speeds lol



> If you're worried that it is too short, it could be increased to another value, or I could work on loading the controls asynchronously but that will create more changes.

My initial testing is fine as is with multiple cameras so I think for a first PR its perfect, may need to see if any adjustments are needed after users start to utilize the feature.

dshokouhi (approved):
Looks good to me, tested debug APK and images loaded fine

JBassett (approved):

PR #2670, opened 2022-07-11
Minor location update fixes

dshokouhi:
## Summary

- Stops using deprecated priority constants introduced in latest [location dependency update](https://developers.google.com/android/guides/releases#june_07_2022)
- Stops using deprecated constructor usage, not sure when this was deprecated
- Adds a short delay when high accuracy mode is disabling so we can ensure the request to remove location updates has completed before we register for normal background updates. During testing I found that when I was manually enabling and disabling high accuracy mode from settings the updates were not coming in consistently (if at all) anymore and it turns out we were requesting updates before the service was successfully destroyed. After adding this delay the frequency in updates returned.

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
@@ -21,6 +21,7 @@ import com.google.android.gms.location.LocationCallback
 import com.google.android.gms.location.LocationRequest
 import com.google.android.gms.location.LocationResult
 import com.google.android.gms.location.LocationServices
+import com.google.android.gms.location.Priority
 import dagger.hilt.android.AndroidEntryPoint
 import io.homeassistant.companion.android.common.bluetooth.BluetoothUtils
 import io.homeassistant.companion.android.common.data.integration.Entity
@@ -37,6 +38,7 @@ import io.homeassistant.companion.android.location.HighAccuracyLocationService
 import io.homeassistant.companion.android.notifications.MessagingManager
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import io.homeassistant.companion.android.common.R as commonR
@@ -265,6 +267,7 @@ class LocationSensorManager : LocationSensorManagerBase() {
                     } else {
                         Log.d(TAG, "High accuracy mode parameters changed. Disable high accuracy mode.")
                         stopHighAccuracyService()
+                        delay(500)
                         requestLocationUpdates()
                     }
                 }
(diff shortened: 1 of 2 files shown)


app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -265,6 +267,7 @@ class LocationSensorManager : LocationSensorManagerBase() {
                     } else {
                         Log.d(TAG, "High accuracy mode parameters changed. Disable high accuracy mode.")
                         stopHighAccuracyService()
+                        delay(500)
```
jpelgrom, line 270: Do you know what exactly caused the issue? Because location updates are received through intents, if I had to guess maybe it's a race condition with the `PendingIntent` we create so perhaps the foreground service intent shouldn't use the same ID as the background service intent to make sure they don't influence each other / cancel out? Both services request `removeLocationUpdates` with (usually) basically the same intent.

I'm unable to reproduce the bug so I can't test it, but simply adding a delay feels like a bad hack and should be avoided.

dshokouhi: > Do you know what exactly caused the issue?

I think it is a speed issue, not entirely sure if its caused by the other changes in this PR moving away from deprecated methods?



> Because location updates are received through intents, if I had to guess maybe it's a race condition with the `PendingIntent` we create so perhaps the foreground service intent shouldn't use the same ID as the background service intent to make sure they don't influence each other / cancel out? Both services call `requestLocationUpdates` and `removeLocationUpdates` with (usually) the same intent.

That is a good point, let me try and test that out here, woudl be a quick change to use a different intent that hits the same function.



> I'm unable to reproduce the bug so I can't test it, but simply adding a delay feels like a bad hack and should be avoided.

I agree, I wasn't really sure how to proceed here but when I looked at my logs it was definitely an issue. With this delay I continue to see updates flowing in every minute but without it I don't see updates coming so it could be the intent conflict you mentioned.

dshokouhi: working well without a delay, as always thank you @jpelgrom 🙏

jpelgrom: I was actually referring to the `requestCode` (currently `0`), but this also works to make the intents unique!

dshokouhi: Oh hmm I don't see `requestCode` in either of the location files lol do you think we should still change it given I no longer see the delayed issue?

jpelgrom: `requestCode` is the [second parameter for `PendingIntent.getBroadcast`](https://developer.android.com/reference/kotlin/android/app/PendingIntent?hl=en#getbroadcast)

No need to change it now, it works and might come in handy in the future to be able to handle high accuracy updates differently.

JBassett (approved):

Issue #2673, opened 2022-07-13
Last update sensor no longer showing settings on new installs


dshokouhi:
**Home Assistant Android version:**

latest master

**Android version:**
12
**Phone model:**
Pixel 6 pro
**Home Assistant version:**
2022.3.x
**Last working Home Assistant release (if known):**
Might be before #2585 
**Description of problem:**

On a fresh install the last update sensor settings page is not showing the add new intent button so users can add intents. On an upgraded install with 1 intent registered it does work.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```

**Screenshot of problem:**

![image](https://user-images.githubusercontent.com/1634145/178769884-9694492c-de9b-4b4b-95af-9296e490379a.png)

**Additional information:**

Reference: https://community.home-assistant.io/t/last-update-trigger-sensor/438968/

dshokouhi:
@marazmarci FYI in case this was introduced with #2585

dshokouhi:
Fix incoming, just need to add the toggle if we don't need to create a new intent.

marazmarci:
And there's another bug that I discovered recently, but I didn't have time yet to report it:

When there are >10x Intents, the order will change every time there's a sensor update. (It's caused by the alphabetical ordering of the numbers: "Intent 1" > "Intent 10")

And I think it can also cause data loss if the user is editing an Intent and there's an update happening in the background in the meantime, and the user hits save, because it will edit a different item in the list.

dshokouhi:
Oh wow I have not noticed that yet, I only use 2 intents personally haha

marazmarci:
@dshokouhi I created an issue for the aforementioned bug: #2703

PR #2674, opened 2022-07-13
Fix last update add new intent toggle not showing up on new installs

dshokouhi:
## Summary

Fixes: #2673 

Need to make sure on a new install the setting toggle shows up as well.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt
@@ -92,16 +92,14 @@ class LastUpdateManager : SensorManager {
         if (addNewIntentToggle == null) {
             // add the toggle if it was not already added.
             sensorDao.add(SensorSetting(lastUpdate.id, SETTING_ADD_NEW_INTENT, "false", SensorSettingType.TOGGLE))
-        } else {
-            if (addNewIntentToggle.value == "true") {
-                val newIntentSettingOrdinal = intentSettings.size + 1
-                val newIntentSettingName = "$INTENT_SETTING_PREFIX$newIntentSettingOrdinal:"
-                if (allSettings.none { it.name == newIntentSettingName }) {
-                    // turn off the toggle:
-                    sensorDao.add(SensorSetting(lastUpdate.id, SETTING_ADD_NEW_INTENT, "false", SensorSettingType.TOGGLE))
-                    // add the new Intent:
-                    sensorDao.add(SensorSetting(lastUpdate.id, newIntentSettingName, intentAction, SensorSettingType.STRING))
-                }
+        } else if (addNewIntentToggle.value == "true") {
+            val newIntentSettingOrdinal = intentSettings.size + 1
+            val newIntentSettingName = "$INTENT_SETTING_PREFIX$newIntentSettingOrdinal:"
+            if (allSettings.none { it.name == newIntentSettingName }) {
+                // turn off the toggle:
+                sensorDao.add(SensorSetting(lastUpdate.id, SETTING_ADD_NEW_INTENT, "false", SensorSettingType.TOGGLE))
+                // add the new Intent:
+                sensorDao.add(SensorSetting(lastUpdate.id, newIntentSettingName, intentAction, SensorSettingType.STRING))
             }
         }
     }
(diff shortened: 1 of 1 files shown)


jpelgrom (approved):

common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt
```diff
@@ -98,6 +98,9 @@ class LastUpdateManager : SensorManager {
                 // add the new Intent:
                 sensorDao.add(SensorSetting(lastUpdate.id, newIntentSettingName, intentAction, SensorSettingType.STRING))
             }
+        } else {
+            // add the toggle if it was not already added.
+            sensorDao.add(SensorSetting(lastUpdate.id, SETTING_ADD_NEW_INTENT, "false", SensorSettingType.TOGGLE))
```
marazmarci, line 103: The execution will reach the `else` branch when ...
a. there's no toggle yet
b. **there's already a toggle, with `false` value**

So I think this would cause a DB write operation on every sensor update.

Maybe we could do something like this to prevent this:

```kotlin
val addNewIntentToggle = allSettings.firstOrNull { it.name == SETTING_ADD_NEW_INTENT }
if (addNewIntentToggle == null) {
    // add the toggle if it was not already added.
    sensorDao.add(SensorSetting(lastUpdate.id, SETTING_ADD_NEW_INTENT, "false", SensorSettingType.TOGGLE))
} else {
    if (addNewIntentToggle.value == "true") {
        val newIntentSettingOrdinal = intentSettings.size + 1
        val newIntentSettingName = "$INTENT_SETTING_PREFIX$newIntentSettingOrdinal:"
        if (allSettings.none { it.name == newIntentSettingName }) {
            // turn off the toggle:
            sensorDao.add(SensorSetting(lastUpdate.id, SETTING_ADD_NEW_INTENT, "false", SensorSettingType.TOGGLE))
            // add the new Intent:
            sensorDao.add(SensorSetting(lastUpdate.id, newIntentSettingName, intentAction, SensorSettingType.STRING))
        }
    }
}
```

marazmarci (changes requested):

common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt
dshokouhi: Good call, the old code always wrote to the DB here.  Just tested this code and its working as expected on a new install. Will update  now.

marazmarci: Sorry, I didn't state it explicitly that I edited my code suggestion. It's up to you if you like the new version more or the already committed version.

common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt
```diff
@@ -88,15 +88,20 @@ class LastUpdateManager : SensorManager {
...
-                // turn off the toggle:
-                sensorDao.add(SensorSetting(lastUpdate.id, SETTING_ADD_NEW_INTENT, "false", SensorSettingType.TOGGLE))
-                // add the new Intent:
-                sensorDao.add(SensorSetting(lastUpdate.id, newIntentSettingName, intentAction, SensorSettingType.STRING))
+        val addNewIntentToggle = allSettings.firstOrNull { it.name == SETTING_ADD_NEW_INTENT }
+        if (addNewIntentToggle == null) {
+            // add the toggle if it was not already added.
+            sensorDao.add(SensorSetting(lastUpdate.id, SETTING_ADD_NEW_INTENT, "false", SensorSettingType.TOGGLE))
+        } else {
+            if (addNewIntentToggle.value == "true") {
```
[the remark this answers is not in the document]
dshokouhi: efficiency applied :)

common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt
dshokouhi: > Sorry, I didn't notify you that I edited my code suggestion. It's up to you if you like the new version more or the already committed version.

LOL I just saw this, lets leave it as is for now to get the fix out there. We can look into improving this later I think :)

JBassett (approved):

PR #2682, opened 2022-07-15
Android 13 notification changes + target

jpelgrom:
## Summary

This PR includes updates to handle the [notification changes in Android 13](https://developer.android.com/about/versions/13/changes/notification-permission). It will also change the app's target Android version to Android 13, as being able to request notification permission when we want instead of when Android decides to, requires targeting Android 13.

A new screen is added to onboarding that requests the user to grant the notification permission on Android 13+ or when using the minimal version, and the app settings have also been updated to check if notifications are enabled on Android 8+.  I've opted to show the screen in onboarding to minimal version users on _all_ Android versions; on Android 13 denying the notification permission will also turn off the websocket service, so to provide a consistent experience I believe it makes sense to show it on all versions.

Based on my testing disabling notifications doesn't produce any major issues, this behaves the same as if notifications were manually turned off on Android 12 (Android silently blocks them, the app doesn't really notice). 

<details>
<summary>There are a few notifications sent by the app that might be missed when the permission is denied. All of these notifications are for failures that will still be visible to the user, either because something isn't changing as expected or something breaks.</summary>

 - notification command received but the app is missing the required permission(s)
 - location disabled might result in degraded experience, on app startup
 - sensor sync from core tried to enable a sensor but the app is missing the required permission(s)
 - database migration failed
</details>

---

Targeting Android 13 also results in other behavior changes:

 - [x] External storage permission is no longer supported -> #2691
 - [x] Intent filter changes -> Wear dependency bumped in #2910, otherwise it doesn't impact the app much [as described in more detail here](https://github.com/home-assistant/android/pull/2901#issuecomment-1255538200)
 - [x] WebView [force dark changes](https://developer.android.com/about/versions/13/behavior-changes-13#webview-color-theme) -> #2923
 - [x] Turning Bluetooth on/off [is deprecated](https://developer.android.com/reference/kotlin/android/bluetooth/BluetoothAdapter?hl=en#enable) -> adjusted to post notification on Android 13 as invalid in 1d7d70b2d5be61116d712ac88fb95f11a568dd09

## Screenshots

|     |Light|Dark|
|-----|-----|-----|
|Permission screen in onboarding, shown after entering device name. This will be skipped when using the full version and on Android <13 or permission already granted.|image|image|
|Tapping Continue will show the system's permission prompt.|image|image|
|If the permission is denied, the app will inform you where you can change it|image|image|
|In the app's settings, the 'Notification Channels' setting will be replaced by the 'Notification Permission' setting if the permission is denied/all channels are turned off. Selecting it takes you to the system list of app notification channels where the permission can be granted/they can be turned on (on Android 8+).|image|image|

Documentation: home-assistant/companion.home-assistant#831

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -23,7 +23,7 @@ android {
     defaultConfig {
         applicationId = "io.homeassistant.companion.android"
         minSdk = 21
-        targetSdk = 31
+        targetSdk = 33
 
         versionName = System.getenv("VERSION") ?: "LOCAL"
         versionCode = System.getenv("VERSION_CODE")?.toIntOrNull() ?: 1
--- a/app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearViewModel.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearViewModel.kt
@@ -175,13 +175,15 @@ class SettingsWearViewModel @Inject constructor(
         url: String,
         authCode: String,
         deviceName: String,
-        deviceTrackingEnabled: Boolean
+        deviceTrackingEnabled: Boolean,
+        notificationsEnabled: Boolean
     ) {
         val putDataRequest = PutDataMapRequest.create("/authenticate").run {
             dataMap.putString("URL", url)
             dataMap.putString("AuthCode", authCode)
             dataMap.putString("DeviceName", deviceName)
             dataMap.putBoolean("LocationTracking", deviceTrackingEnabled)
+            dataMap.putBoolean("Notifications", notificationsEnabled)
             setUrgent()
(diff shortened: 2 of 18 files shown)


jpelgrom:
The screen in onboarding is based on the iOS onboarding screen for their notification permission, but a bit more Android-ified.

I tried to make it fit in with the other screens, but every screen is using different styles and padding which makes it hard. If this PR is accepted I'll probably do another update to make onboarding more consistent. Quick example:
|image|image|image|
|-----|-----|-----|

dshokouhi:
Looks the issue has been marked as complete as of today! https://issuetracker.google.com/issues/235538840#comment23

dshokouhi (approved):

JBassett (approved):

PR #2690, opened 2022-07-18
Deprecate Android WebView quick bar gesture support

### LINKED ISSUE #2686: 3-fingers gesture (WebViewActivity?) doesnt work anymore
**Home Assistant Android version:** beta-2478-0883ada4-full

**Android version:** Android 12, OneUI 4.1

**Phone model:** Samsung Galaxy S10

**Last working Home Assistant release (if known):** a long time ago, idk

**Description of problem:** 3-fingers swype down does not work anymore for me (and for my friend with Xiaomi Redmi 8 Note Pro also), maybe problem still present since https://github.com/home-assistant/android/pull/2396
Any non-documented breaking-changes with this coolest feature with last releases? 
I dont know reason.

```

```

**Screenshot of problem:**

https://user-images.githubusercontent.com/38353584/179396514-d865d473-a2df-48ef-812a-4b6a260974f9.mp4

jpelgrom:
## Summary

A three finger swipe gesture was added as a workaround to open the quick bar while it was only possible with a physical keyboard; Home Assistant 2022.5 added a button for it to the toolbar and a later update broke support for the app's workaround (see [this comment](https://github.com/home-assistant/android/issues/2686#issuecomment-1186577056)). Because it is no longer functional on recent versions and an alternative is available, add a comment in the code to mark it for removal.

Closes #2686

## Screenshots

n/a

The documentation is related but doesn't depend on this PR being merged to be true

Documentation: home-assistant/companion.home-assistant#777

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -245,7 +245,7 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
 
         webView = binding.webview
         webView.apply {
-            // TODO This quick bar workaround only works on Home Assistant core versions up to 2022.7
+            // TODO This quick bar workaround only works on Home Assistant core versions <2022.7
             // If not 'fixed' or officially supported: should be removed in Android 2023.1 (GitHub: #2690)
             setOnTouchListener(object : OnSwipeListener() {
                 override fun onSwipe(
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -245,6 +245,8 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
 
         webView = binding.webview
         webView.apply {
+            // TODO This quick bar workaround only works on Home Assistant core versions up to 2022.7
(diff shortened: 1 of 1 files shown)


dshokouhi:
So I wonder how long we should consider to keep things backwards compatible, I believe this feature is 2 months old now but I am not sure what is considered a sane amount of time to wait.

jpelgrom:
Home Assistant moves fast and the feature is already broken in recent versions, but I understand the hesitation.

Perhaps a better first step would be to mark the feature as deprecated in the docs (also add note that it is no longer working on recent versions) and give users more time to migrate, for example 6 months since introduction -> remove in 2022.11? There doesn't appear to be an official policy for deprecation but looking at analytics.home-assistant.io 6 months seems reasonable to me.

It's also possible to leave the feature as it is, and only update the docs. It's not actively breaking something and the code was recently updated to make it work correctly.

dshokouhi:
I think adding in a note in the docs and maybe even a code comment to remove it once time has passed would be good.

If we really want to keep this feature then we will need to enhance the HA frontend so the external bus can be leveraged to trigger the quick bar.  We still need that to be done to avoid our more info pop-up hack :) I created a feature request for that a while ago lol.

jpelgrom:
Alright, let's make it less destructive :) I've checked which frontend version includes the problematic change and added a comment in the code + docs.

(Tried 'fixing' it with more workarounds as well but wasn't able to get it to work.)

dshokouhi (approved):

JBassett (approved):

PR #2691, opened 2022-07-19
Remove external storage permission for logs, completely on Android 10+

jpelgrom:
## Summary

When targeting Android 13, [the `WRITE_EXTERNAL_STORAGE` permission will stop working](https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions) (requesting it simply does nothing), which is currently used for saving the logs. Turns out it actually isn't required at all since Android 4.4 for the directory we're using according to the documentation! The permission request has been removed on Android 6 and above, still requesting it on Android 5 due to [a possible bug](https://issuetracker.google.com/issues/37013142) for some devices (the permission will automatically be granted on that version and it is still used for downloads in that case).

As a result, the `WRITE_EXTERNAL_STORAGE` permission is only required for devices on Android <10 to handle Downloads. The manifest has been updated to reflect this.

## Screenshots

n/a

n/a

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -32,7 +32,8 @@
     <uses-permission android:name="com.android.alarm.permission.SET_ALARM" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
     <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"
+        android:maxSdkVersion="28"/>
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
     <uses-permission android:name="android.permission.PACKAGE_USAGE_STATS"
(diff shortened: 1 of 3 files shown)


jpelgrom:
Interesting, I didn't get that error on Android 7, 12 or 13. Maybe Android is caching the permission somewhere, even though it is no longer visible in the UI?
Should be fixable by adding it to the file provider's paths, change coming soon!

dshokouhi:
Wasn't it renamed to files and media permission?

dshokouhi:
Oh so this comes after #2682 then right?

jpelgrom:
> This error shows up when I attempt to share the logs. I thought we needed the permission for this?

Managed to reproduce it on an emulator! Changed back to the latest beta of the app and granted external storage permission, and this error still shows up in the logs. Does this actually affect sharing for you? It isn't fatal for me and the apps that I shared the file to seem to be able to read the file (tried Gmail and Messages).

I'm guessing it's this bug: https://issuetracker.google.com/issues/151386328, comment 12 includes a workaround that I could test but I'd rather do that in a separate PR if sharing still works with this exception in the logs.

> Oh so this comes after https://github.com/home-assistant/android/pull/2682 then right?

No, before #2682. This change works without changing the target version, keeps the other PR smaller, and also benefits users on older Android versions.

dshokouhi (approved):
Looks good to me

JBassett (approved):

Issue #2696, opened 2022-07-21
Sensor: "bluetooth_connection" attributes - Capture Bluetooth friendly device names


mkanet:
**Is your feature request related to a problem? Please describe.**

Currently, the **bluetooth_connection sensor** displays unfriendly/unrecognizable mac addresses in attributes.  _(See below example)._  

sensor.sm_n960u_bluetooth_connection - [attributes]
```
state_class: measurement
connected_not_paired_devices:
  - 68:A8:E1:F1:69:26
connected_paired_devices:
  - 30:C0:1B:2D:07:24
paired_devices:
  - 54:27:1E:FC:B0:A6
  - 00:02:72:C9:59:33
  - 30:C0:1B:2D:07:24
  - C4:4F:33:6E:04:23
  - 5C:EB:68:5E:1F:78
  - 5C:EB:68:5E:13:89
  - FC:69:47:28:71:33
unit_of_measurement: connection(s)
icon: mdi:bluetooth
friendly_name: SM-N960U Bluetooth Connection
```

**Describe the solution you'd like**

Capture friendly Bluetooth device name in addition to MAC addresses)

sensor.sm_n960u_bluetooth_connection - [attributes]
```
state_class: measurement
connected_not_paired_devices:
  - 68:A8:E1:F1:69:26 (HTPC)
connected_paired_devices:
  - 30:C0:1B:2D:07:24 (JBL Flip 4)
paired_devices:
  - 54:27:1E:FC:B0:A6 (AccousticSleep)
  - 00:02:72:C9:59:33 (T2-R)
  - 30:C0:1B:2D:07:24 (JBL Flip 4)
  - C4:4F:33:6E:04:23 (HandsFreeLink)
  - 5C:EB:68:5E:1F:78 (Litter-Robot)
  - 5C:EB:68:5E:13:89 (Envy Photo 7800 series)
  - FC:69:47:28:71:33 (T2-L)
unit_of_measurement: connection(s)
icon: mdi:bluetooth
friendly_name: SM-N960U Bluetooth Connection
```

PR #2697, opened 2022-07-21
Breaking Change: Add friendly BT name next to the mac address if we have it

dshokouhi:
## Summary

Fixes: #2696 

This is a breaking change because the friendly name will now be added to the Mac address if we have it, otherwise it will print the address again.

New format will be: `MAC (Name)`

Also cleaned up some warnings and removed a commented line

## Screenshots

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
@@ -5,6 +5,7 @@ import android.content.Context
 import android.os.Build
 import io.homeassistant.companion.android.bluetooth.ble.IBeaconTransmitter
 import io.homeassistant.companion.android.bluetooth.ble.TransmitterManager
+import io.homeassistant.companion.android.common.bluetooth.BluetoothDevice
 import io.homeassistant.companion.android.common.bluetooth.BluetoothUtils
 import io.homeassistant.companion.android.common.bluetooth.BluetoothUtils.supportsTransmitter
 import io.homeassistant.companion.android.common.sensors.SensorManager
@@ -135,9 +136,9 @@ class BluetoothSensorManager : SensorManager {
         if (checkPermission(context, bluetoothConnection.id)) {
 
             val bluetoothDevices = BluetoothUtils.getBluetoothDevices(context, true)
-            pairedDevices = bluetoothDevices.filter { b -> b.paired }.map { "${it.address} (${it.name})" }
-            connectedPairedDevices = bluetoothDevices.filter { b -> b.paired && b.connected }.map { "${it.address} (${it.name})" }
-            connectedNotPairedDevices = bluetoothDevices.filter { b -> !b.paired && b.connected }.map { "${it.address} (${it.name})" }
+            pairedDevices = bluetoothDevices.filter { b -> b.paired }.map { checkNameAddress(it) }
+            connectedPairedDevices = bluetoothDevices.filter { b -> b.paired && b.connected }.map { checkNameAddress(it) }
+            connectedNotPairedDevices = bluetoothDevices.filter { b -> !b.paired && b.connected }.map { checkNameAddress(it) }
             totalConnectedDevices = bluetoothDevices.count { b -> b.connected }
         }
         onSensorUpdated(
@@ -258,5 +259,9 @@ class BluetoothSensorManager : SensorManager {
                 "Supports transmitter" to supportsTransmitter(context)
             )
         )
+    }
+
+    private fun checkNameAddress(bt: BluetoothDevice): String {
+        return if (bt.address != bt.name) "${bt.address} (${bt.name})" else bt.address
     }
 }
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
```diff
@@ -136,10 +135,10 @@ class BluetoothSensorManager : SensorManager {
         if (checkPermission(context, bluetoothConnection.id)) {
 
             val bluetoothDevices = BluetoothUtils.getBluetoothDevices(context, true)
-            pairedDevices = bluetoothDevices.filter { b -> b.paired }.map { it.address }
-            connectedPairedDevices = bluetoothDevices.filter { b -> b.paired && b.connected }.map { it.address }
-            connectedNotPairedDevices = bluetoothDevices.filter { b -> !b.paired && b.connected }.map { it.address }
-            totalConnectedDevices = bluetoothDevices.filter { b -> b.connected }.count()
+            pairedDevices = bluetoothDevices.filter { b -> b.paired }.map { "${it.address} (${it.name})" }
```
jpelgrom, line 138: If we don't have a name, the `name` property will be set to the device's address. This could result in a string like "A1:B2:C3:D4:E5:F6 (A1:B2:C3:D4:E5:F6)". 

This seems undesirable to me, so perhaps add a check it is not the same (or make name nullable, if possible)?

dshokouhi: ok I will opt to only add it only if we have a name, this method is used in other places so I would rather not modify it :)

jpelgrom (approved):
Works correctly, and users can make it a non-breaking change by splitting the sensor attribute strings on spaces (" ") and using the first value

JBassett (approved):

Arquiteto:
> Works correctly, and users can make it a non-breaking change by splitting the sensor attribute strings on spaces (" ") and using the first value

Can you elaborate on that? I'll give you an example of what was working for me earlier:

I had a sensor that calculated distance traveled by my cars, based on me and my wifes phone connecting to its BT beginning like this:
```
            {% set bt_mac =  '12:34:56:78:9A:BC'  %}                                    
            {% set bt_connection_sensor_1 = 'sensor.my_phone_bluetooth_connection' %}   
            {% set phone_owner_id_1 = "My phone" %}                                         
            {% set bt_connection_sensor_2 = 'sensor.wifes_phone_bluetooth_connection' %}     
            {% set phone_owner_id_2 = "Wife's phone" %}                                         
            {% set distance_sensor_id = 'sensor.car_traveled distance' %}               


            {# Check which phone changed location and is connected to cars BT #}
            {% if ((bt_mac in state_attr(bt_connection_sensor_1, 'connected_paired_devices')) and (trigger.id == phone_owner_id_1)) 
               or ((bt_mac in state_attr(bt_connection_sensor_2, 'connected_paired_devices')) and not (bt_mac in state_attr(bt_connection_sensor_1, 'connected_paired_devices')) and (trigger.id == phone_owner_id_2))  %}
```

Now, for my phone the friendly name is available, but it's not for my wife, so I can't just change the "bt_mac" variable to include friendly name. Connected paired devices attribute returns a list, so I can't just use .split(" ") on it (at least it's not working for me in template dev tools). And when in my car I always have more then 1 device connected, as I use a smartwatch.

So what I'm really looking for is to find if there is a device with mac, for example '12:34:56:78:9A:BC' in a list looking like ['12:34:56:78:9A:BC (CAR RADIO)', 'DE:F1:23:45:67:89 (SMARTWACH (MODEL))']

The only way I currently came up with is (and that's for just 1 phone)
```
{% set bt_mac = "12:34:56:78:9A:BC" %}
{% set myphone_bt_devices = state_attr('sensor.my_phone_bluetooth_connection', 'connected_paired_devices') %}

{% set myphone_in_car = namespace(found=false) %}
{% for bt_device in myphone_bt_devices  %}
  {% if bt_mac in bt_device.split(" ") %}
    {% set myphone_in_car.found = true %}
  {% endif %}
{% endfor %}

{% if myphone_in_car.found %}
```

Which is waaay more complicated than what was working before:
```

{% set bt_mac = "12:34:56:78:9A:BC" %}
{% set bt_connection_sensor_1 = 'sensor.my_phone_bluetooth_connection' %}

{% if bt_mac in state_attr(bt_connection_sensor_1, 'connected_paired_devices') %}
```

Do you have a simplier way of checking, did I miss something?


And I also have some automations based on phone connecting with car, so I have to rewrite those too...

Arquiteto:
Okay, I managed to find a simplier fix with the help of my FB group. It just requires converting the list to a string. I leave it here for someone who might encounter simillar problem

```
            {% set bt_mac =  '12:34:56:78:9A:BC'  %}                                    
            {% set bt_connection_sensor_1 = 'sensor.my_phone_bluetooth_connection' %}   
     
            {# Check which phone changed location and is connected to cars BT #}
            {% if ((bt_mac in state_attr(bt_connection_sensor_1, 'connected_paired_devices')) | string %}
```

PR #2702, opened 2022-07-23
Automatically delete 'stale'/old widgets from database

jpelgrom:
## Summary

Automatically deletes widgets from the database that are no longer bound to the widget provider but that are still in the database when asked to update all widgets. This will remove unnecessary updates (and websocket connections!) due to 'stale'/old widgets that aren't exposed to the user. The PR also removes the "Delete widget" button on the widget configuration screen as it is no longer required.

No issue to link here, but "remove unused widgets from Settings > Companion app > Manage Widgets" has been a 'solution' for multiple issues in the past.

## Screenshots

The "Delete widget" button has been removed when updating a widget:
|Light|Dark|
|-----|-----|
|Settings for a template widget, with the only button being 'Update Widget' and no longer also including 'Delete Widget', light mode|Settings for a template widget, with the only button being 'Update Widget' and no longer also including 'Delete Widget', dark mode|

n/a, the documentation didn't mention the delete functionality

## Any other notes

Testing to make sure widgets aren't accidentally deleted: I've been running this code for a few hours now and tried several things to imitate running it for a longer period of time (manually triggering doze, killing the launcher, rebooting a few times).

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/BaseWidgetConfigureActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/BaseWidgetConfigureActivity.kt
@@ -1,48 +1,16 @@
 package io.homeassistant.companion.android.widgets
 
 import android.appwidget.AppWidgetManager
-import android.content.Context
-import android.view.View
 import android.widget.Toast
-import androidx.lifecycle.lifecycleScope
 import io.homeassistant.companion.android.BaseActivity
 import io.homeassistant.companion.android.common.R
 import io.homeassistant.companion.android.database.widget.WidgetDao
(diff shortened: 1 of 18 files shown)


dshokouhi:
This is awesome, I believe the only time we have seen widgets get removed from the home screen but still exist in the database when we refactored widgets. So if you test by renaming any widget folder you should be able to reproduce that :) (Assuming you haven't already done that test lol)

jpelgrom:
I hadn't tested that scenario! Can confirm that when you rename the class or package Android removes the widgets but doesn't notify the app. On the next screen on intent when all widgets are updated this is correctly detected and the widgets are removed from the database 👍

JBassett (approved):

PR #2704, opened 2022-07-23
Fix unwanted reordering of Last Update Sensor's Intent list when there are 10+ entries

### LINKED ISSUE #2703: Last Update Sensor's Intent list is reordered on every update when there are 10+ entries
**Home Assistant Android version:** beta-2500-f8d0d461-full

**Android version:** 12

**Phone model:** Samsung S20FE

**Last working Home Assistant release (if known):** -

**Description of problem:**

The Last Update Sensor's list of Intents is reordered on every update when there are 10+ entries. It's caused by the lexicographic ordering of numbers (contained in the `name` DB field).

<!--**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```-->

**Screen recording of the problem:**

https://user-images.githubusercontent.com/1349654/180609826-58899331-2936-45a4-a331-5ffb985af978.mp4


<!--**Additional information:**-->

### PR DESCRIPTION
**Pull request by marazmarci**
## Summary 
Fixes #2703

The Last Update Sensor's list of Intents was reordered on every update when there were 10+ entries. It was caused by the lexicographic ordering of numbers (contained in the `name` DB field). It was caused by the sequence continuity detection I added recently because I haven't tested it with more than 10 entries. 

I fixed it by sorting it ~~like it is displayed on the UI~~ numerically.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt
@@ -73,6 +73,8 @@ class LastUpdateManager : SensorManager {
         }
         val intentSettings = allSettings.filter {
             it.name.startsWith(INTENT_SETTING_PREFIX)
+        }.sortedBy {
+            it.name.removeSuffix(":").substringAfterLast(':').toInt()
         }
         val isSequenceContinuous = intentSettings.withIndex().all { (index, setting) ->
             val ordinal = setting.name.removePrefix(INTENT_SETTING_PREFIX).removeSuffix(":").toInt()

### CODE REVIEW

File: common/src/main/java/io/homeassistant/companion/android/common/sensors/LastUpdateManager.kt [hunk 1]
```diff
@@ -71,9 +71,12 @@ class LastUpdateManager : SensorManager {
         if (settingsToRemove.isNotEmpty()) {
             sensorDao.removeSettings(lastUpdate.id, settingsToRemove.map { it.name })
         }
-        val intentSettings = allSettings.filter {
+        val intentSettingsUnsorted = allSettings.filter {
             it.name.startsWith(INTENT_SETTING_PREFIX)
         }
+        val intentSettings = intentSettingsUnsorted.sortedBy {
```
jpelgrom, line 77: Is there a specific reason you're introducing the `intentSettingsUnsorted` variable, instead of chaining the calls (`allSettings.filter { ... }.sortedBy { ... }`)? As far as I can see it is only used for sorting here.

marazmarci: Good point, there's no specific reason. Fixed.

jpelgrom (approved):
> I fixed it by sorting it like it is displayed on the UI

Not really, the UI will show 10, 11, 12, ... before 1 as well.

However it does correctly prevent unwanted reordering and follows the numbers ☺️

marazmarci:
> > I fixed it by sorting it like it is displayed on the UI
> 
> Not really, the UI will show 10, 11, 12, ... before 1 as well.
> 
> However it does correctly prevent unwanted reordering and follows the numbers ☺️

Yeah, right, thanks for the correction! I edited the description 😄

JBassett (approved):

Issue #2701, opened 2022-07-22
Notification Command clear_notification should trigger mobile_app_notification_cleared event


soloam:
**Is your feature request related to a problem? Please describe.**
When A "clear_notification" is triggered, I would like to catch that action (It could be originated by other terminal) and I would like to catch that event to make the automation act, for example stop waiting in all devices.

**Describe the solution you'd like**
When the app recieves a "clear_notification" command, it should trigger a "mobile_app_notification_cleared" event, or another event. It could contain in the data that it was triggered by the service

**Describe alternatives you've considered, if any**
None

**Additional context**
All included above

dshokouhi:
Why is this request necessary? As long as you use critical notification format it will clear immediately and you can clear from all devices.

soloam:
I have automatons that act on some messages being cleared, and without this event, or some that I can l monitor, I can't know that a notifications was cleared when it's from the app. Sure I could do a script that calls the clear notification and trigger a event, but I think that it would be the best solution that if a notifications is cleared (even that it's from a service) that it should trigger this event (or other with a different name)

dshokouhi:
> I have automatons that act on some messages being cleared, and without this event, or some that I can l monitor, I can't know that a notifications was cleared when it's from the app.

But you don't need an event here, simply sending the message is good enough on your part to know it was cleared as that is your action to clear it. The event being sent is sent based on user and/or device behavior, the clear command is already something you use to trigger from HA.

jpelgrom:
> But you don't need an event here

I can imagine two relevant scenarios:

 - It's possible that the notification is cleared either by the user or by an automation, being able to use the same trigger for both could be useful.
 - It's possible that an automation requests clearing the notification but it was already cleared.

dshokouhi:
So now we would need to check if the notification still exists and then send the event in those cases? That begins to overcomplicate the clear command which simply clears the notification as soon as the command is received regardless of the notification being present.

Now I can see a benefit in the app sending an event to confirm receipt of any notification but not just for a notification clear command being received and processed. That to me seems repetitive, if you send an automation to clear the notification then you should be assured its already cleared as that was the purpose of the command. Its a fair assumption to continue with the script as the command gets processed immediately, it will add a delay in that command just to send an event.  The original event gets sent when a user performs an action to clear the notification, HA sending an automation is identical to that behavior.

soloam:
[...]

PR #2705, opened 2022-07-23
Fire an event when a notification is received

dshokouhi:
## Summary

Fixes: #2701 by firing an event anytime a notification is received

The event type will be `mobile_app_notification_received`

example data: 

```
{
    "event_type": "mobile_app_notification_received",
    "data": {
        "message": "test",
        "device_id": "DEVICE_ID"
    },
    "origin": "REMOTE",
    "time_fired": "2022-07-23T15:31:33.320831+00:00",
    "context": {
        "id": "ID",
        "parent_id": null,
        "user_id": "USER_ID"
    }
}
```

Documentation: home-assistant/companion.home-assistant#779

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -258,6 +258,17 @@ class MessagingManager @Inject constructor(
             NotificationItem(0, now, jsonData[MESSAGE].toString(), jsonObject.toString(), source)
         notificationDao.add(notificationRow)
 
+        mainScope.launch {
+            if (integrationUseCase.getNotificationReceivedEvent()) {
+                try {
+                    integrationUseCase.fireEvent("mobile_app_notification_received", jsonData)
+                } catch (e: Exception) {
+                    Log.e(TAG, "Unable to send notification received event", e)
+                }
+            } else {
+                // No op
+            }
+        }
         when {
             jsonData[MESSAGE] == REQUEST_LOCATION_UPDATE -> {
                 Log.d(TAG, "Request location update")
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -259,15 +259,10 @@ class MessagingManager @Inject constructor(
         notificationDao.add(notificationRow)
 
         mainScope.launch {
-            if (integrationUseCase.getNotificationReceivedEvent()) {
-                try {
-                    integrationUseCase.fireEvent("mobile_app_notification_received", jsonData)
-                } catch (e: Exception) {
-                    Log.e(TAG, "Unable to send notification received event", e)
-                    Toast.makeText(context, commonR.string.notification_received_error, Toast.LENGTH_LONG).show()
-                }
-            } else {
-                // No op
+            try {
+                integrationUseCase.fireEvent("mobile_app_notification_received", jsonData)
+            } catch (e: Exception) {
+                Log.e(TAG, "Unable to send notification received event", e)
             }
         }
         when {
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -264,6 +264,7 @@ class MessagingManager @Inject constructor(
                     integrationUseCase.fireEvent("mobile_app_notification_received", jsonData)
                 } catch (e: Exception) {
                     Log.e(TAG, "Unable to send notification received event", e)
+                    Toast.makeText(context, commonR.string.notification_received_error, Toast.LENGTH_LONG).show()
                 }
             } else {
                 // No op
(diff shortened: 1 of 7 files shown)


app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -258,6 +258,18 @@ class MessagingManager @Inject constructor(
             NotificationItem(0, now, jsonData[MESSAGE].toString(), jsonObject.toString(), source)
         notificationDao.add(notificationRow)
 
+        mainScope.launch {
+            if (integrationUseCase.getNotificationReceivedEvent()) {
+                try {
+                    integrationUseCase.fireEvent("mobile_app_notification_received", jsonData)
+                } catch (e: Exception) {
+                    Log.e(TAG, "Unable to send notification received event", e)
+                    Toast.makeText(context, commonR.string.notification_received_error, Toast.LENGTH_LONG).show()
```
jpelgrom, line 267: I'm not a big fan of using Toast messages:
 - They lack context. It's hard to know which notification this message belongs to, especially if the notification was a command, triggered by an automation and/or multiple notifications were delivered shortly after another (for example due to doze). This is only made worse by the duration and how Toasts are queued to display.
 - They are not actionable. So sending the event failed, what should I do now? You can't manually send it again. Manually doing something when there's no confirmation event defeats the purpose of the event, that should be handled by the automation.

As a result, the only result of the message is that you get frustrated and/or have to open the app to find out more and do something about it.

I'd say simply printing the error in the logs is enough. If you feel like a more visible error is important, maybe save the result as a boolean in the database with the specific notification and show it in the notification history details?

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -258,6 +258,18 @@ class MessagingManager @Inject constructor(
             NotificationItem(0, now, jsonData[MESSAGE].toString(), jsonObject.toString(), source)
         notificationDao.add(notificationRow)
 
+        mainScope.launch {
+            if (integrationUseCase.getNotificationReceivedEvent()) {
+                try {
+                    integrationUseCase.fireEvent("mobile_app_notification_received", jsonData)
```
jpelgrom, line 264: For consistency with other [app events](https://companion.home-assistant.io/docs/integrations/app-events) and the mentioned HTML5 event, suggestion: `mobile_app.notification_received`.

(I know there is `mobile_app_notification_action`, but that is the exception!)

jpelgrom (commented):
Why make this optional, and not enabled for everyone? In the issue you mentioned that this was at least somewhat inspired by [the `html5_notification.received` event](https://www.home-assistant.io/integrations/html5/#received-event), and that isn't a setting you have to toggle. If you want to use it, you can use it, if you don't want to use it, you can simply ignore the event.

I can't think of any major downsides, whereas making it optional will introduce yet another setting to the app.

dshokouhi:
> Why make this optional, and not enabled for everyone? In the issue you mentioned that this was at least somewhat inspired by [the `html5_notification.received` event](https://www.home-assistant.io/integrations/html5/#received-event), and that isn't a setting you have to toggle. If you want to use it, you can use it, if you don't want to use it, you can simply ignore the event.
> 
> I can't think of any major downsides, whereas making it optional will introduce yet another setting to the app.

I was thinking of making it a setting to be battery friendly but the app had already woken up so I agree it doesn't need to be a setting. I'll update this PR when I get a chance.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
dshokouhi: Ok I can remove the toast message, we should probably do the same in the other PR for the clear event failure in this case.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
dshokouhi: There is also `mobile_app_notification_cleared` 😉

jpelgrom: OK, another one that's different, but there are still more events with a `.` 😉 For consistency with other notification events `_` like you did might make more sense, feel free to mark this as resolved if you want.

> we don't list any of the notification events here

Yeah I don't really understand why, they're still events sent by the app?

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
jpelgrom: Are you referring to #2576? These arguments still apply but are a little weaker in that case because it is the result of a direct action by the user (but still very much in favor of removing the toast there! ).

dshokouhi: Ok I think it makes sense, some of our Toast messages were from before the in app log viewer. Now that we've had it for sometime it's an ok place to tell users where to go.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
dshokouhi: I'm not sure why I didn't add them when I updated that page lol.

JBassett (approved):

Issue #2470, opened 2022-04-23
Modification of the UUID, the Minor and the Major Bluetooth in HA Companion from HA server.


AxokGit:
**Is your feature request related to a problem? Please describe.**

I would like to change the UUID, minor and major of the Bluetooth iBeacon transmitter on an Android device from a Home Assistant server service. On the one hand for anonymity reasons (rotating UUID) and for security reasons.
And if possible, a system that would modify the UUID autonomously without dependence on the Home Assistant server (more details in the following)

**Describe the solution you'd like**

To do this we can use the Notification Commands like changing the transmission power. There would be one command for the UUID, one for the minor and one for the major.

**Describe alternatives you've considered, if any**

Another solution would be to modify our UUID autonomously on the Android application side without dependency on the server (thus functional without internet on the phone) using the Google Authentificator principle. From a given key and a time stamp, a key could be generated and would allow to have an unpredictable rotating UUID.

**Additional context**

My idea would be to unlock a front door by identifying ourselves from the signal our phone emits in Bluetooth. By rotating our UUID more or less frequently and unpredictably, this would prevent someone from impersonating us by listening to their UUID otherwise by duplicating the signal, anyone could enter.
In order to make this system battery efficient, we could imagine a system that changes our UUID using the principle of Google Authentificator. Everything would be on the phone without having an internet connection to update the UUID.

Hellis81:
Love the idea.
The only thing I would want to add is a communication back to HA that the app has changed the UUID, or make the UUID a sensor so that the automation that sends the new UUID can make sure it's set or try again.

dshokouhi:
The UUID is already an attribute for the sensor, this request is for a notification command to change the UUID which will also reflect in the attribute.

https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt#L255

lapodomo:
had the same idea , would be nice if we could change the UUID (temporary) by notification command.
if we use a esp32 to detect the original UID then send a generated UUID by the esp to the phone so the esp can check if the generated uid matches the UUID send out.

PR #2713, opened 2022-07-25
Allow UUID, Major and Minor to be changed via notification command

dshokouhi:
## Summary

Fixes: #2470 by introducing new parameters for BLE command.

- `ble_set_uuid` with `ble_uuid` to change the UUID
- `ble_set_major` with `ble_major` to change Major
- `ble_set_minor` with `ble_minor` to change Minor

Documentation: home-assistant/companion.home-assistant#783

## Any other notes

FCM: https://github.com/home-assistant/mobile-apps-fcm-push/pull/84

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -80,6 +80,7 @@ import org.json.JSONObject
 import java.net.URL
 import java.net.URLDecoder
 import java.util.Locale
+import java.util.UUID
 import javax.inject.Inject
 import io.homeassistant.companion.android.common.R as commonR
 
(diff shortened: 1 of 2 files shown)


jpelgrom:
While changing the settings works, I found that I had to actively stop and start the transmitter to have them used. Is this intentional? I would have expected it to change automatically and immediately when receiving the command.

dshokouhi:
> While changing the settings works, I found that I had to actively stop and start the transmitter to have them used. Is this intentional? I would have expected it to change automatically and immediately when receiving the command.

According to the code, it should restart next time the sensor checks in for an update.

https://github.com/home-assistant/android/blob/c1b4ea5626767e3358b205576a523918b25f5250/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt#L206-L212

  Let me try and test it out by forcing a sensor update after the DB was updated.

jpelgrom:
> According to the code, it should restart next time the sensor checks in for an update.

That's also what I was able to find, guess I wasn't missing anything then. As I said I was expecting it to work immediately, so not having to wait on the SensorWorker, but if that is by design it's OK.

dshokouhi:
So had to make sure the transmitter restarts properly, but also need to make sure the attributes can be sent during the restart. The restart happens very fast, the state change is actually not sent to HA  but we can see in the logs that it restarted properly.

```
2022-07-26 09:17:12.134 16226-16226/io.homeassistant.companion.android.debug I/BeaconTransmitter: Advertisement start succeeded.
```

jpelgrom (approved):
Works as expected

JBassett (approved):

PR #2716, opened 2022-07-27
Improve logging and error messages during onboarding

dshokouhi:
## Summary

Noticed in our Authentication Fragment that we were not printing errors, we should still print errors in case our error messages are not completely correct to help with troubleshooting.

I have also changed the generic error message to include error content like status code and description

## Screenshots

image

## Any other notes

Ref: https://discord.com/channels/330944238910963714/377667559303938050/1001786919665012736

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
@@ -83,19 +83,24 @@ class AuthenticationFragment : Fragment() {
                                     error: WebResourceError?
                                 ) {
                                     super.onReceivedError(view, request, error)
-                                    Log.e(TAG, "onReceivedError: Status Code: ${error?.errorCode} Description: ${error?.description}")
-                                    showError(
-                                        requireContext().getString(
-                                            commonR.string.error_http_generic,
-                                            error?.errorCode,
-                                            if (error?.description.isNullOrBlank())
-                                                commonR.string.no_description
-                                            else
-                                                error?.description
-                                        ),
-                                        null,
-                                        error
-                                    )
+                                    if (request?.url?.toString() == authUrl) {
+                                        Log.e(
+                                            TAG,
+                                            "onReceivedError: Status Code: ${error?.errorCode} Description: ${error?.description}"
+                                        )
+                                        showError(
+                                            requireContext().getString(
+                                                commonR.string.error_http_generic,
+                                                error?.errorCode,
+                                                if (error?.description.isNullOrBlank())
+                                                    commonR.string.no_description
+                                                else
+                                                    error?.description
+                                            ),
+                                            null,
+                                            error
+                                        )
+                                    }
                                 }
 
                                 override fun onReceivedHttpError(
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
@@ -197,9 +197,11 @@ class AuthenticationFragment : Fragment() {
                                 WebViewClient.ERROR_PROXY_AUTHENTICATION -> requireContext().getString(commonR.string.webview_error_PROXY_AUTHENTICATION)
                                 WebViewClient.ERROR_UNSUPPORTED_AUTH_SCHEME -> requireContext().getString(commonR.string.webview_error_AUTH_SCHEME)
                                 WebViewClient.ERROR_HOST_LOOKUP -> requireContext().getString(commonR.string.webview_error_HOST_LOOKUP)
+                                else -> message
                             }
+                        } else {
+                            message
                         }
-                        message
                     }
                 }
             )
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
@@ -47,6 +47,8 @@ class AuthenticationFragment : Fragment() {
 
     private val viewModel by activityViewModels<OnboardingViewModel>()
 
+    private var authUrl: String? = null
+
     @Inject
     lateinit var themesManager: ThemesManager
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
```diff
@@ -78,6 +78,7 @@ class AuthenticationFragment : Fragment() {
                                     error: WebResourceError?
                                 ) {
                                     super.onReceivedError(view, request, error)
+                                    Log.e(TAG, "onReceivedError: $error")
```
jpelgrom, line 81: This will result in a fairly useless error message like:
```
2022-07-27 18:13:26.538 4100-4100/io.homeassistant.companion.android.debug E/AuthenticationFragment: onReceivedError: pE0@93b3f73
```
It should use the `getErrorCode()` and `getDescription()` functions to provide something usable.

app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
```diff
@@ -87,6 +88,7 @@ class AuthenticationFragment : Fragment() {
                                     errorResponse: WebResourceResponse?
                                 ) {
                                     super.onReceivedHttpError(view, request, errorResponse)
+                                    Log.e(TAG, "onReceivedHttpError: $errorResponse")
```
jpelgrom, line 91: This will result in a fairly useless error message like:
```
2022-07-27 18:22:59.669 4100-4100/io.homeassistant.companion.android.debug E/AuthenticationFragment: onReceivedHttpError: android.webkit.WebResourceResponse@18c35d7
```
It should use the `getStatusCode()`and `getReasonPhrase()` functions to provide something usable.

app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
```diff
@@ -78,6 +78,9 @@ class AuthenticationFragment : Fragment() {
                                     error: WebResourceError?
                                 ) {
                                     super.onReceivedError(view, request, error)
+                                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
jpelgrom, line 81: This [entire function](https://developer.android.com/reference/android/webkit/WebViewClient?hl=en#onReceivedError(android.webkit.WebView,%20android.webkit.WebResourceRequest,%20android.webkit.WebResourceError)) was added on API 23, no need to include the if statement here

dshokouhi: android studio was complaining, I'll add a required flag on the function to make it happy :)

jpelgrom: Android Studio does complain if you remove the version check without adding `@RequiresApi(Build.VERSION_CODES.M)` to the function, but it will build the app 🤔

dshokouhi: Yea shouldn't be highlighted in red if we can still build IMO

jpelgrom:
I don't know if you're done yet, but maybe it's a good idea to also use show the error code / description in `onReceivedError`? The docs even mention that the description is localized for the user.

dshokouhi:
> I don't know if you're done yet, but maybe it's a good idea to also use show the error code / description in `onReceivedError`? The docs even mention that the description is localized for the user.

yea good call we should cover all our bases, the SSL error message should be good to go I hope 😁

app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
```diff
@@ -155,21 +182,21 @@ class AuthenticationFragment : Fragment() {
...
-                                WebViewClient.ERROR_UNSUPPORTED_AUTH_SCHEME -> commonR.string.webview_error_AUTH_SCHEME
-                                WebViewClient.ERROR_HOST_LOOKUP -> commonR.string.webview_error_HOST_LOOKUP
+                                    requireContext().getString(commonR.string.webview_error_FAILED_SSL_HANDSHAKE)
+                                WebViewClient.ERROR_AUTHENTICATION -> requireContext().getString(commonR.string.webview_error_AUTHENTICATION)
+                                WebViewClient.ERROR_PROXY_AUTHENTICATION -> requireContext().getString(commonR.string.webview_error_PROXY_AUTHENTICATION)
+                                WebViewClient.ERROR_UNSUPPORTED_AUTH_SCHEME -> requireContext().getString(commonR.string.webview_error_AUTH_SCHEME)
+                                WebViewClient.ERROR_HOST_LOOKUP -> requireContext().getString(commonR.string.webview_error_HOST_LOOKUP)
                             }
                         }
                         message
```
jpelgrom, line 202: Found a bug, this will overwrite the specific error set by the error code above and should be placed in else branches. For example, on SSL handshake errors it will now show error code -11 instead of the specific handshake message.

jpelgrom (commented):
These "Descriptions are localized, and thus can be used for communicating the problem to the user" strings are not what I was expecting:
A dialog showing the text 'Unable to connect to Home Assistant. There was an error loading Home Assistant. Please review the connection settings and try again. Error Code: -6 Description: net::ERR_CONNECTION_REFUSED'.

Still better than a more generic error in my opinion.

> the SSL error message should be good to go I hope 😁

Yes I believe we had already covered all possible `SslError`s with specific messages 👍

app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
dshokouhi: These strings are already in an `else` branch after all the SSL specific error messages. Is that what you mean? Just above the condition

jpelgrom: No it looks like inside this block it should also use an else? Without it when sending a `WebViewClient.ERROR_FAILED_SSL_HANDSHAKE` I'm getting the original message you provided, if I change it to something like this I correctly get the SSL handshake error:
```kotlin
else -> {
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
        when (error?.errorCode) {
           ... -> ...
           else -> message
        }
    } else {
        message
    }
}
```

dshokouhi: ooh gotcha, yea seems like it was always there. For some reason when I look at it the way it was before it looked _mostly_ correct lol

dshokouhi: usually android studio is good on calling out a missing else branch like that but seems it doesnt always complain when it should

jpelgrom: Maybe it was a Kotlin feature in that case we're not aware of? Doesn't matter in the end, it works as expected now.

jpelgrom (approved):

agners (approved):

app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
```diff
@@ -68,17 +71,31 @@ class AuthenticationFragment : Fragment() {
...
 
+                                @RequiresApi(Build.VERSION_CODES.M)
                                 override fun onReceivedError(
                                     view: WebView?,
                                     request: WebResourceRequest?,
                                     error: WebResourceError?
                                 ) {
                                     super.onReceivedError(view, request, error)
-                                    showError(commonR.string.webview_error, null, error)
+                                    Log.e(TAG, "onReceivedError: Status Code: ${error?.errorCode} Description: ${error?.description}")
```
jpelgrom, line 86: The `if (request?.url?.toString() == authUrl)` check should probably also be added to here, it will report non-HTTP errors (like timeouts) for all requests here.

dshokouhi: > The `if (request?.url?.toString() == authUrl)` check should probably also be added to here, it will report non-HTTP errors (like timeouts) for all requests here.

Good call, this would match our current `WebViewActivity` as well. Well not 100% but close enough lol.



> And maybe also for SSL errors? Although that one is unlikely to fail only partially.

We could add it but I don't see the same check done in `WebViewActivity` so not sure if its needed?

jpelgrom: > We could add it but I don't see the same check done in `WebViewActivity` so not sure if its needed?

Yep it's a bit much for something that is unlikely to happen, if there haven't been any complaints about the behavior in the `WebViewActivity` it should be fine as it is.

JBassett (approved):

PR #2718, opened 2022-07-27
Improve discovery reliability for some devices

jpelgrom:
## Summary

After complaining on Discord that discovery of Home Assistant instances on Android wasn't working for me most of the time, it turns out the solution was a quick search away. Some devices, most notably Google Pixels, require manually requesting a multicast lock in order for DNS-SD to work reliably (see [this comment](https://github.com/andriydruk/RxDNSSD/issues/38#issuecomment-392206040)). After including this change, it now works 100% of the time for me, hopefully also for others.

## Screenshots

n/a, discovery isn't a new interface

n/a

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -7,6 +7,7 @@
 
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+    <uses-permission android:name="android.permission.CHANGE_WIFI_MULTICAST_STATE" />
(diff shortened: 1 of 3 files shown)


dshokouhi (approved):
Tested debug APK don't see any issues. I did notice my test instance occasionally fails to resolve with error 3 but I am pretty sure that issue has always been there. Sometimes it resolves and other times not.

jpelgrom:
[`NsdManager.FAILURE_ALREADY_ACTIVE`](https://developer.android.com/reference/kotlin/android/net/nsd/NsdManager?hl=en#failure_already_active)'s value is 3. Not sure why it would happen, but these changes only handle discovery, not resolving, so it shouldn't happen more than it already did.

JBassett (approved):

Issue #2726, opened 2022-07-29
Connected not paired devices 


guarddog13:
**Is your feature request related to a problem? Please describe.**

The Bluetooth sensor recognizes connected not paired devices but they don't show up as devices that can trigger high accuracy mode within the app. 

**Describe the solution you'd like**

Could this be set up in the companion app so that i can choose these connected not paired devices? 

**Describe alternatives you've considered, if any**

I currently use a template automation based on the mac address from the Bluetooth sensor. 
**Additional context**

![Screenshot_20220729-084451_Home Assistant](https://user-images.githubusercontent.com/31947215/181765172-4e8af650-214c-4ab0-b3d6-524167c18207.jpg)

![Screenshot_20220729-084940_Home Assistant](https://user-images.githubusercontent.com/31947215/181765225-1c6fa994-3723-4604-82db-8fbc2513b079.jpg)

guarddog13:
I have a couple of use cases with this, especially if I Fi can get their collars to connect to the bases better than they do phones. Devices like the Fi collar and the Roav charger I use in my truck are connected via BLE.

1. My thought when Fi pushes their update is that when my Husky isn't connected to his base High Accuracy Mode is on, on my phone. This does one of two things...it ensures that I'm tracked when he's with me, this keeps the Fi app on the ready working with Google Fused location. If he takes off as he's prone to do then I have at least the distance of my BLE to keep faster than 60 second tracking. He tends to stay in a visible range and bolts when you get close. Any time I can gain is gold as we live near a busy freeway and highway. 

2. The other is to have the app see the Roav devices to trigger the Companion as mentioned before. I work in a rural county and having an automation do it may not always be possible...although I was surprised to see the high accuracy automation run with nearly zero internet connection....when a webhook didn't go to my server in the exact same spot.

dshokouhi:
Can you test out the debug APK from: https://github.com/home-assistant/android/actions/runs/2764064862 make sure to use the full version. Should let you select the device. Make sure the devices are connected when you select the device because thats the only way the app knows it exists :)

guarddog13:
That seems to work. When the Roav charger is on it connects and gives it a name. When it's off it gives it it's MAC addy. The only thing and this may because of it being a debug app...

I had to open the app to get it to recognize the toggles. If the app was already open, I had to close and reopen it to recognize the toggles.

dshokouhi:
[...]

PR #2731, opened 2022-07-30
Allow connected not paired devices to trigger high accuracy mode

dshokouhi:
## Summary

Fixes: #2726 

Can't think of a reason why we would want to exclude them, we did have quite a bit of code improvement here so might be why it was a quick switch :)

With this change I can now select my Mi Band 7 which is a connected but not paired device. One thing to keep in mind is that the device needs to be currently connected to select it in the list. Otherwise we don't know it exists. Paired devices will always show up.

Documentation: home-assistant/companion.home-assistant# Not sure if its completely necessary as the docs do not mention the device has to be paired.

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
@@ -409,7 +409,7 @@ class LocationSensorManager : LocationSensorManagerBase() {
         if (highAccuracyModeBTDevices.isNotEmpty()) {
             constraintsUsed = true
 
-            val bluetoothDevices = BluetoothUtils.getBluetoothDevices(latestContext)
+            val bluetoothDevices = BluetoothUtils.getBluetoothDevices(latestContext, true)
 
             // If any of the stored devices aren't a Bluetooth device address, try to match them to a device
             var updatedBtDeviceNames = false
--- a/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
@@ -288,7 +288,7 @@ class SensorDetailViewModel @Inject constructor(
             SensorSettingType.LIST ->
                 setting.entries
             SensorSettingType.LIST_BLUETOOTH ->
-                BluetoothUtils.getBluetoothDevices(getApplication()).map { it.address }
+                BluetoothUtils.getBluetoothDevices(getApplication(), true).map { it.address }
             else ->
                 emptyList()
         }


jpelgrom:
Shouldn't settings also be adjusted to make them selectable?

https://github.com/home-assistant/android/blob/90a9a27e5e371445f75481e7a93609528357d3ff/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt#L291

(I don't have a good way to test this, it looks like all my Bluetooth devices that show up as connected are either paired only or paired and not paired at the same time?)

dshokouhi:
> Shouldn't settings also be adjusted to make them selectable?

Good catch, thank you

jpelgrom:
Looking at my own suggestion again, it doesn't make much sense. Changing the code here doesn't do much as the keys also need labels to match.

However, more importantly, `BluetoothUtils.getBluetoothDevices` already always returns both paired _and_ connected not paired devices. The parameter specified is `allowDuplicates` - aka if the device is already considered paired should it appear again when connected via GATT? That doesn't matter for the settings view.

So does this add anything for the constraint? Is it possible to have a device that was previously paired, doesn't appear in the paired list as connected but does appear in the connected not paired list?

dshokouhi:
> So does this add anything for the constraint?

At least in my own setup the amount of devices does not change from the original commit. I only have 1 device that is connected not paired. It did not show up until the first commit. Looking at the code here I figured it needed the correct key just in case? It doesn't seem to hurt this use case and this setting is the only place its used lol.

> Is it possible to have a device that was previously paired, doesn't appear in the paired list as connected but does appear in the connected not paired list?

No, that shouldn't be possible if its connected the device is connected, pairing doesn't matter here. IIRC the mi band could be paired to android but then it will show up in `connected_paired_devices` and `connected_not_paired_devices` due to its pairing with its own app.

> (sorry for the additional work @dshokouhi - should've caught it this morning!)

all good 😄

jpelgrom:
> I only have 1 device that is connected not paired. It did not show up until the first commit. Looking at the code here I figured it needed the correct key just in case? It doesn't seem to hurt this use case and this setting is the only place its used lol.

That is curious because the only added devices by changing the parameter are paired + connected via GATT devices. Devices that aren't paired are always included, no matter if it's `true` or `false`. Are you sure it was connected when you tested it without changes?

> IIRC the mi band could be paired to android but then it will show up in connected_paired_devices and connected_not_paired_devices due to its pairing with its own app.

Looking at it again I believe maybe the function is incorrect? When a device is returned from [`bluetoothManager.getConnectedDevices`](https://github.com/home-assistant/android/blob/90a9a27e5e371445f75481e7a93609528357d3ff/common/src/main/java/io/homeassistant/companion/android/common/bluetooth/BluetoothUtils.kt#L34) the app [always treats it as unpaired](https://github.com/home-assistant/android/blob/90a9a27e5e371445f75481e7a93609528357d3ff/common/src/main/java/io/homeassistant/companion/android/common/bluetooth/BluetoothUtils.kt#L42), without checking the `bondState`. Testing with my device that shows up in both `connected_paired_devices` and `connected_not_paired_devices`, it turns out the `bondState` is actually `BOND_BONDED` for both...

dshokouhi:
> Are you sure it was connected when you tested it without changes?

hmm great question, looks like it was not connected in the moment that I did that check. Seems like the connection isn't that stable lol 😅 

I need to be more diligent with these details.

Closing this PR for now as it seems to be unnecessary for the use case

PR #2734, opened 2022-07-30
Fix Last Notification sensor not updating when notification is repeated with the same content

### LINKED ISSUE #2733: Last Notification sensor doesn't update when notification is repeated with same content
**Home Assistant Android version:** [beta-2534-90a9a27e](https://github.com/home-assistant/android/releases/tag/beta-2534-90a9a27e)

**Android version:** Android 12

**Phone model:** Samsung S21

**Last working Home Assistant release (if known):** -

**Description of problem:**

When an app's package name is registered in the Last Notification sensor, and the sends a notification, then later it sends a notification with exactly the same contents, the Last Notification sensor won't update its state, nor the `post_time` attribute.

For example, it can be reproduced with [Pushbullet](https://play.google.com/store/apps/details?id=com.pushbullet.android) (`com.pushbullet.android`):
Mirroring → Send a test notification.

marazmarci:
Fixes #2733

## Summary
I added a `forceUpdate: Boolean` flag to `Sensor`, and `NotificationSensorManager` sets it to `true` when a notification is received. The flag is cleared after an update (in `SensorDao::updateLastSentStateAndIconAndClearForceUpdateFlag(...)`).

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
@@ -297,8 +297,8 @@ abstract class SensorReceiverBase : BroadcastReceiver() {
                     val sensor = sensorDao.get(it.uniqueId)
                     if (sensor != null) {
                         sensor.registered = null
-                        sensor.lastSentState = ""
-                        sensor.lastSentIcon = ""
+                        sensor.lastSentState = null
+                        sensor.lastSentIcon = null
                         sensorDao.update(sensor)
                     }
                 }
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -750,8 +750,8 @@ class MessagingManager @Inject constructor(
                     mainScope.launch {
                         sensorDao.updateLastSentStateAndIcon(
                             BluetoothSensorManager.bleTransmitter.id,
-                            "",
-                            ""
+                            null,
+                            null
                         )
                     }
                     BluetoothSensorManager().requestSensorUpdate(context)
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
@@ -154,7 +154,8 @@ class NotificationSensorManager : NotificationListenerService(), SensorManager {
             lastNotification,
             state.toString().take(255),
             lastNotification.statelessIcon,
-            attr
+            attr,
+            forceUpdate = true,
         )
 
         // Need to send update!
@@ -211,7 +212,8 @@ class NotificationSensorManager : NotificationListenerService(), SensorManager {
             lastRemovedNotification,
             state.toString().take(255),
             lastRemovedNotification.statelessIcon,
-            attr
+            attr,
+            forceUpdate = true,
         )
 
         // Need to send update!
--- a/common/schemas/io.homeassistant.companion.android.database.AppDatabase/31.json
+++ b/common/schemas/io.homeassistant.companion.android.database.AppDatabase/31.json
@@ -1,0 +1,713 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 31,
+    "identityHash": "87e521cae11434d2d6208760f5e82b50",
+    "entities": [
+      {
+        "tableName": "sensor_attributes",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`sensor_id` TEXT NOT NULL, `name` TEXT NOT NULL, `value` TEXT NOT NULL, `value_type` TEXT NOT NULL, PRIMARY KEY(`sensor_id`, `name`))",
+        "fields": [
+          {
+            "fieldPath": "sensorId",
+            "columnName": "sensor_id",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "name",
+            "columnName": "name",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "value",
+            "columnName": "value",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "valueType",
+            "columnName": "value_type",
+            "affinity": "TEXT",
+            "notNull": true
+          }
+        ],
+        "primaryKey": {
+          "columnNames": [
+            "sensor_id",
+            "name"
+          ],
+          "autoGenerate": false
+        },
+        "indices": [],
+        "foreignKeys": []
+      },
+      {
+        "tableName": "Authentication_List",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`host` TEXT NOT NULL, `Username` TEXT NOT NULL, `Password` TEXT NOT NULL, PRIMARY KEY(`host`))",
+        "fields": [
+          {
+            "fieldPath": "host",
+            "columnName": "host",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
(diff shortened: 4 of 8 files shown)


common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
```diff
@@ -299,6 +299,7 @@ abstract class SensorReceiverBase : BroadcastReceiver() {
                         sensor.registered = null
                         sensor.lastSentState = ""
                         sensor.lastSentIcon = ""
+                        sensor.forceUpdate = false // TODO is this correct?
```
marazmarci, line 302: I need assistance with this. What do you think, should we set it to `false` here?

dshokouhi:
I was thinking about this issue the other day. What do we think about comparing the `post_time` attribute value and using that to determine if we should clear the last sent state? Every notification gets a post time.

marazmarci:
> I was thinking about this issue the other day. What do we think about comparing the `post_time` attribute value and using that to determine if we should clear the last sent state? Every notification gets a post time.

[Hmm](https://cultofthepartyparrot.com/parrots/hd/hmmparrot.gif), you've got a point there.
I would go even further:
We should delete `last_sent_state` every time we receive a new notification in `onNotificationPosted`. And this way I think we won't need any other modifications. Let me PoC it quickly.

(Some additional thoughts: It would be better if notifications would be sent as HA events, or at least have it as an option.)

jpelgrom:
For something that isn't relevant to most sensors, with the latest commits it is much simpler and seems to achieve the same goal 👍

dshokouhi:
> (Some additional thoughts: It would be better if notifications would be sent as HA events, or at least have it as an option.)

I have thought about this before, we could make it an option to do so but then it makes me wonder whats the point in the sensor because a user can just save the event data into their own custom sensor. The data would feel duplicated here.

jpelgrom:
> > (Some additional thoughts: It would be better if notifications would be sent as HA events, or at least have it as an option.)
> 
> I have thought about this before, we could make it an option to do so but then it makes me wonder whats the point in the sensor because a user can just save the event data into their own custom sensor. The data would feel duplicated here.

I can understand both use cases but agree that it'd unnecessarily duplicate data. Sensor states are recorded whereas events are not (they are ephermal). While you can easily listen for sensor updates like an event, saving events to a custom sensor would require more work.

dshokouhi (approved):
Much better approach than what I was thinking 😃 will make it easier to apply to other sensors in the future as well. I am not sure if any other existing sensor will benefit from this at the moment.

marazmarci:
I realized a potential problem:

When the notification's text is an empty `String` (`""`), it won't be sent because it will be equal to `lastSentState = ""`.

Proposal: make the `lastSentState` field nullable (`String?`), and set it to `null` when `forceUpdate = true`

jpelgrom:
Setting `lastSentIcon` to "" would work, as the app will never set it to an empty string. You could do `lastSentState` and `lastSentIcon` to match how an update is forced if sending an update fails.

Your latest changes include a database change but don't include the schema file required for the AutoMigration.

marazmarci:
> Setting `lastSentIcon` to "" would work, as the app will never set it to an empty string. You could do `lastSentState` and `lastSentIcon` to match how an update is forced if sending an update fails.

Yeah, I thought about it also, but it seems like a quick'n'dirty workaround

jpelgrom:
> Yeah, I thought about it also, but it seems like a quick'n'dirty workaround

Not more of a workaround than resetting the state to "" to trigger an update is? If it's done via `forceUpdate = true` it's fine in my opinion.

marazmarci:
> > Yeah, I thought about it also, but it seems like a quick'n'dirty workaround
> 
> Not more of a workaround than resetting the state to "" to trigger an update is? If it's done via `forceUpdate = true` it's fine in my opinion.

Hmm, true 😄 Implemented in 5e9cfc38

> Your latest changes include a database change but don't include the schema file required for the AutoMigration.

Fixed in 8e3b7d4b, then reverted in 12db1289 in favor of the `lastSentIcon = ""` approach.

(I'll clean up the commit history whenever I'll mark it as ready for review.)

dshokouhi:
> When the notification's text is an empty `String` (`""`), it won't be sent because it will be equal to `lastSentState = ""`.

actually if there is no text we default to package name

https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt#L150

we also should not be sending a blank state for the sensor even in those cases, so we should prevent it if its actually possible to get blank text with the field. I didn't think its possible to get that field and have it be empty.

marazmarci:
> I didn't think its possible to get that field and have it be empty.

Let me try that!

marazmarci:
I was able to produce a state where `state = ""`:
1. I added the prod HA app to the notification allow list of the debug HA app.
2. I sent a notification to the prod HA app with these service call parameters:
```
service: notify.mobile_app_s21
data:
  message: ""
```
3. The debug app sent `""` as the state to the server (with `lastSentIcon = ""` in place)

Without `lastSentIcon = ""`, the app won't send the state to the server.

dshokouhi:
Hmm for our purpose we really shouldn't be sending a blank state for any sensor. Not that it should be fixed here but we should enforce that for the sensor. For HA core I am surprised that notification sent out because `message` is supposed to be required but I guess it doesn't check that the string is empty 😬

marazmarci:
For me, it feels like it would be a cleaner approach to have the `lastSentState` as a nullable field, regardless of if we should allow `""` as a state or not.

Currently the default value is `lastSentState = ""`, which should be `lastSentState = null` IMHO.

WDYT?

dshokouhi:
> For me, it feels like it would be a cleaner approach to have the `lastSentState` as a nullable field, regardless of if we should allow `""` as a state or not.
> 
> Currently the default value is `lastSentState = ""`, which should be `lastSentState = null` IMHO. (see [8e3b7d4](https://github.com/home-assistant/android/commit/8e3b7d4b401de98125b734b9b264c625aaf46046))
> 
> WDYT?

I agree 👍

marazmarci:
And btw I also think that it's not necessarily a problem if we allow `""` as a state for this sensor, because Android allows posting a notification with empty text, so the sensor should be able to reflect that.

Or is there any other aspect that I may have missed? 🤔

dshokouhi:
> And btw I also think that it's not necessarily a problem if we allow `""` as a state for this sensor, because Android allows posting a notification with empty text, so the sensor should be able to reflect that.
> 
> Or is there any other aspect that I may have missed? 🤔

I can't remember the exact issue number but the user pointed out a blank state messes with the HA frontend graphs The user ends up seeing gaps in the graphs. Looking at the data docs I don't see it explicitly mentioning its required to be filled in. The concern is more on the HA core side of things than Android in this case.

marazmarci:
> > For me, it feels like it would be a cleaner approach to have the `lastSentState` as a nullable field, regardless of if we should allow `""` as a state or not.
> > Currently the default value is `lastSentState = ""`, which should be `lastSentState = null` IMHO. (see [8e3b7d4](https://github.com/home-assistant/android/commit/8e3b7d4b401de98125b734b9b264c625aaf46046))
> > WDYT?
> 
> I agree 👍

Implemented nullable `lastSentState` in 6b90716c
(and nullable `lastSentIcon` in 1431a29e as a bonus)

jpelgrom:
As `null` instead of `""` is now used to force an update, this usage should also be changed:

https://github.com/home-assistant/android/blob/90a9a27e5e371445f75481e7a93609528357d3ff/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt#L749-L756

marazmarci:
> As `null` instead of `""` is now used to force an update, this usage should also be changed:
> 
> https://github.com/home-assistant/android/blob/90a9a27e5e371445f75481e7a93609528357d3ff/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt#L749-L756

Done in 1c77b5ac.

marazmarci:
I squashed the commits. I think the PR is now ready for review.

JBassett (approved):

Issue #2736, opened 2022-07-30
Bluetooth Connection sensor incorrectly assumes connected devices are not paired


jpelgrom:
**Home Assistant Android version:** 2022.6.0-full, beta-2534-90a9a27e-full

**Android version:** 13

**Phone model:** Pixel 4a

**Home Assistant version:** not relevant

**Last working Home Assistant release (if known):** looking at the history of BluetoothUtils, it has always been broken

**Description of problem:** Discovered [while reviewing a PR](https://github.com/home-assistant/android/pull/2731#issuecomment-1200296124): paired devices that are connected via GATT will show up both in `connected_paired_devices` and `connected_not_paired_devices` because for `connected_not_paired_devices` it actually isn't checked if the device is paired or not:

https://github.com/home-assistant/android/blob/90a9a27e5e371445f75481e7a93609528357d3ff/common/src/main/java/io/homeassistant/companion/android/common/bluetooth/BluetoothUtils.kt#L34

https://github.com/home-assistant/android/blob/90a9a27e5e371445f75481e7a93609528357d3ff/common/src/main/java/io/homeassistant/companion/android/common/bluetooth/BluetoothUtils.kt#L39-L44

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```
n/a
```

**Screenshot of problem:**

![App settings showing sensor attributes, with one device showing up in both `connected_paired_devices` and `connected_not_paired_devices`](https://user-images.githubusercontent.com/8148535/181996906-37616b89-14a7-4702-919f-59cdf54b9407.png)

(the device addresses are the same, just blacked out for privacy)

**Additional information:** After adding debug logging into the app to check the `bondState` of the device, it shows up as `BOND_BONDED` in both cases. This makes sense because they share the same hardware address and [the developer documentation](https://developer.android.com/reference/kotlin/android/bluetooth/BluetoothDevice?hl=en) says that it is: "(...) a thin wrapper for a Bluetooth hardware address".

dshokouhi:
> After adding debug logging into the app to check the `bondState` of the device, it shows up as `BOND_BONDED` in both cases.

ran a test myself on my Mi Band 7, still shows up as unpaired if we check the `bondState` instead of assuming its `false`

jpelgrom:
> ran a test myself on my Mi Band 7, still shows up as unpaired if we check the `bondState` instead of assuming its `false`

[...]

PR #2738, opened 2022-07-31
Actually check the paired state instead of assuming it

dshokouhi:
## Summary

Fixes: #2736 by actually checking the bonded state. I am also suppressing the lint message as its not valid in our use case and makes the page look like it won't compile. The main app already requests the proper permission and its not necessary for the Wear app.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/bluetooth/BluetoothUtils.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/bluetooth/BluetoothUtils.kt
@@ -1,5 +1,6 @@
 package io.homeassistant.companion.android.common.bluetooth
 
+import android.annotation.SuppressLint
 import android.bluetooth.BluetoothManager
 import android.bluetooth.BluetoothProfile
 import android.content.Context
@@ -7,6 +8,7 @@ import androidx.core.content.getSystemService
 import java.lang.reflect.Method
 
 object BluetoothUtils {
+    @SuppressLint("MissingPermission")
     fun getBluetoothDevices(context: Context, allowDuplicates: Boolean = false): List<BluetoothDevice> {
         val devices: MutableList<BluetoothDevice> = ArrayList()
 
@@ -26,7 +28,7 @@ object BluetoothUtils {
                         BluetoothDevice(
                             btDev.address,
                             name,
-                            true,
+                            btDev.bondState == android.bluetooth.BluetoothDevice.BOND_BONDED,
                             isConnected(btDev)
                         )
                     )
@@ -39,7 +41,7 @@ object BluetoothUtils {
                         BluetoothDevice(
                             btDev.address,
                             name,
-                            false,
+                            btDev.bondState == android.bluetooth.BluetoothDevice.BOND_BONDED,
                             isConnected(btDev)
                         )
                     )


jpelgrom:
Now a single device is showing up multiple times in the same attribute in the Bluetooth Connection sensor for me:

Sensor attributes showing a single device appearing multiple times in one list

I think a check should be added to make sure that even if `allowDuplicates` is `true`, it should filter out devices where the `BluetoothDevice` data class is exactly the same.

dshokouhi:
> I think a check should be added to make sure that even if `allowDuplicates` is `true`, it should filter out devices where the `BluetoothDevice` data class is exactly the same.

I think a better way instead of adding another condition is to just call `distinct()` on the lists.  At least in my setup it works but curious if you still get duplicates :)

jpelgrom:
`distinct` works for the attributes but still results in an incorrect state / total of connected devices, because the device is still included in the list and there is no filter/distinct on the count.
The Bluetooth Connection sensor showing "2 connection(s)" as a state but only 1 connected device in the attributes

dshokouhi:
ok now i think it should be good 🤞

jpelgrom:
Unfortunately, I don't think it is. It works for me, but [Android's `BluetoothDevice.equals`](https://android.googlesource.com/platform/packages/modules/Bluetooth/+/refs/heads/master/framework/java/android/bluetooth/BluetoothDevice.java#1380) only compares the device address, so if I understood your comments in #2736 correctly, that Mi Band would now never show up as connected again (it is paired, but the connected device isn't recognized as paired)? If it is only comparing the device address we might as well remove the `allowDuplicates` parameter.

Assuming my interpretation of your comments in the issue is correct, suggestion:
```kotlin
for (btDev in btConnectedDevices) {
    if (!allowDuplicates && devices.any { it.address == btDev.address }) continue
    
    val name = btDev.name ?: btDev.address
    val device = BluetoothDevice(
        btDev.address,
        name,
        btDev.bondState == android.bluetooth.BluetoothDevice.BOND_BONDED,
        isConnected(btDev)
    )
    if (!devices.contains(device)) devices.add(device)
}
```

dshokouhi:
> so if I understood your comments in #2736 correctly, that Mi Band would now never show up as connected again (it is paired, but the connected device isn't recognized as paired)?

it actually did show up as connected when I tested the latest change. In my setup the attributes have always been correct in each of the changes, which makes the testing part difficult lol.  Even when I test your provided code just now I see no changes.

My Mi band is actually not a paired device (not paired to the OS but paired to its own app) so it should only show up as `connected_not_paired_devices` which it always has with each change.

> If it is only comparing the device address we might as well remove the `allowDuplicates` parameter.

I always took it as allow duplicates was meant more to hide the non-paired devices. At least thats how I initially read it.  With that said IMO the device address is really the only thing that counts.

jpelgrom:
> My Mi band is actually not a paired device (not paired to the OS but paired to its own app) so it should only show up as `connected_not_paired_devices` which it always has with each change.

In that case I didn't understand your comments, sorry. For connected, not paired devices there should indeed be no changes 🙈

> > If it is only comparing the device address we might as well remove the `allowDuplicates` parameter.
> 
> I always took it as allow duplicates was meant more to hide the non-paired devices. At least thats how I initially read it. With that said IMO the device address is really the only thing that counts.

No it was added in #2591 to stop the settings from crashing due to device addresses being included twice in the list, while still allowing this sensor to get a list with a device with the same address showing up in both the paired connected and not paired connected lists. As you can see it doesn't do anything for non-paired devices.

Considering the issue this PR solves and the confusion it creates, looking back at it I think it is no longer necessary? The app will now always check paired + connected status irregardless of the source.

dshokouhi:
> Considering the issue this PR solves and the confusion it creates, looking back at it I think it is no longer necessary? The app will now always check paired + connected status irregardless of the source.

one of those things were code wise it doesnt look right, until you look at the end result and realize why it was like that 😅 

So I guess I should close the PR then?

jpelgrom:
> So I guess I should close the PR then?

No the PR is still good to have! Keep the changes from the first commit, remove `allowDuplicates` and always treat it as true / always apply this:

https://github.com/home-assistant/android/blob/90a9a27e5e371445f75481e7a93609528357d3ff/common/src/main/java/io/homeassistant/companion/android/common/bluetooth/BluetoothUtils.kt#L36

jpelgrom (approved):
Thanks for your patience @dshokouhi ❤️

Tested and resolves the issue

dshokouhi:
> Thanks for your patience @dshokouhi ❤️
> 
> Tested and resolves the issue

Thanks for your help in testing and code suggestions 😃

PR #2746, opened 2022-08-02
Make sure to fallback to background updates when disable is selected in high accuracy notification

dshokouhi:
## Summary

Noticed during testing of other PRs that the disable action button in the high accuracy notification only stops the service but it never registers for background updates again. This is not the case when high accuracy is automated either in app or controlled via notification command. As the user has location tracking enabled we should make sure to trigger background updates again when disable is selected.

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/location/HighAccuracyLocationReceiver.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/location/HighAccuracyLocationReceiver.kt
@@ -17,6 +17,10 @@ class HighAccuracyLocationReceiver : BroadcastReceiver() {
                 {
                     HighAccuracyLocationService.stopService(context)
                     LocationSensorManager.setHighAccuracyModeSetting(context, false)
+                    LocationSensorManager.isBackgroundLocationSetup = false
+                    val locationIntent = Intent(context, LocationSensorManager::class.java)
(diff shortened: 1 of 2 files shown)


app/src/full/java/io/homeassistant/companion/android/location/HighAccuracyLocationReceiver.kt
```diff
@@ -17,6 +17,10 @@ class HighAccuracyLocationReceiver : BroadcastReceiver() {
                 {
                     HighAccuracyLocationService.stopService(context)
                     LocationSensorManager.setHighAccuracyModeSetting(context, false)
+                    LocationSensorManager.isBackgroundLocationSetup = false
```
jpelgrom, line 20: Is this required? `LocationSensorManager.ACTION_REQUEST_LOCATION_UPDATES` will go to `setupLocationTracking()` -> `setupBackgroundLocation(true, ?)` -> high accuracy mode is different -> it will try to stop the service and call `requestLocationUpdates()`, which always registers an intent. No override necessary as far as I see.

dshokouhi: Totally unnecessary, thank you :) 🙏

jpelgrom (approved):
Works based on the logs and 2 background location updates (and a lot of duplicates) after tapping 'Disable' on the notification

Issue #2462, opened 2022-04-16
Added backup and restore function


martusi61:
I've searched quite a bit without finding a save and restore settings feature in the companion app, so I guess you're not there. However, it would be very useful to have it available, especially for the numerous widgets and links that can be created in the latest versions.

dshokouhi:
The app supports android auto backup which does restore upon reinstalling the application.

martusi61:
> L'app supporta il backup automatico di Android che viene ripristinato dopo aver reinstallato l'applicazione.

Very nice, but it didn't work with my tablet. My Teclast M40 broke down, so I bought another new one, when I installed it I chose to restore from Google backup of the old tablet, in fact he restored everything including App companion of which he brought only the URL of the home assistant, all the other settings had been reset.

martusi61:
Perhaps it would also be the case to add a manual import export.

PR #2747, opened 2022-08-02
Don't include session in backup/transfer

jpelgrom:
## Summary

A user on Discord mentioned that [they were automatically logged in on a new device](https://discord.com/channels/330944238910963714/562408603345092636/1003940648157708288) which surprised them. Session data is device specific and shouldn't be restored or transferred, so this PR excludes session data from the automatic backup/transfer.

Not including the session data will also show onboarding on start, which prevents issues with re-using the mobile app integration registration.

## Screenshots

n/a

n/a

### DIFF
--- a/app/src/main/res/xml/backup_rules.xml
+++ b/app/src/main/res/xml/backup_rules.xml
@@ -2,4 +2,5 @@
 <full-backup-content>
     <include domain="database" path="." />
     <include domain="sharedpref" path="." />
+    <exclude domain="sharedpref" path="session_0.xml" />
 </full-backup-content>
--- a/app/src/main/res/xml/backup_rules_android12.xml
+++ b/app/src/main/res/xml/backup_rules_android12.xml
@@ -3,9 +3,11 @@
     <cloud-backup>
         <include domain="database" path="." />
         <include domain="sharedpref" path="." />
+        <exclude domain="sharedpref" path="session_0.xml" />
     </cloud-backup>
     <device-transfer>
         <include domain="database" path="." />
         <include domain="sharedpref" path="." />
+        <exclude domain="sharedpref" path="session_0.xml" />
     </device-transfer>
 </data-extraction-rules>
--- a/wear/src/main/AndroidManifest.xml
+++ b/wear/src/main/AndroidManifest.xml
@@ -17,7 +17,8 @@
         android:networkSecurityConfig="@xml/network_security_config"
         android:supportsRtl="true"
         android:theme="@style/Theme.HomeAssistant"
-        android:fullBackupContent="@xml/backup_rules">
+        android:fullBackupContent="@xml/backup_rules"
+        android:dataExtractionRules="@xml/backup_rules_android12">
         <!-- Start things like SensorWorker on device boot -->
         <receiver android:name=".sensors.SensorReceiver"
             android:exported="true">
--- a/wear/src/main/res/xml/backup_rules.xml
+++ b/wear/src/main/res/xml/backup_rules.xml
@@ -2,4 +2,5 @@
 <full-backup-content>
     <include domain="database" path="." />
     <include domain="sharedpref" path="." />
+    <exclude domain="sharedpref" path="session_0.xml" />
 </full-backup-content>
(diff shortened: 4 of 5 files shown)


dshokouhi:
Should we just disable auto backup? It seems to cause more headaches than anything.

jpelgrom:
> Should we just disable auto backup? It seems to cause more headaches than anything.

Backups are nice to have though, and it's not impossible to make them work.

If we decide to remove auto backup, that also closes feature request #2462? It doesn't make sense to me to only support manual backup.

dshokouhi:
> > Should we just disable auto backup? It seems to cause more headaches than anything.
> 
> Backups are nice to have though, and it's not impossible to make them work.
> 
> If we decide to remove auto backup, that also closes feature request #2462? It doesn't make sense to me to only support manual backup.

So for me the only time I find it works well the way it is today is if I upgraded my device and stopped using my old phone immediately. In those cases the app starts up all data is there and I don't even have to login or enable additional sensors. I don't see that as a bad thing like the user in discord did. It becomes an issue when a user does not factory reset old phone and continues to open and use the app on the old phone.  Now with that said things like widgets may not function well and I am pretty sure need to be recreated as the ID's change during a device transfer. I also don't think app shortcuts get added back either.

jpelgrom:
We can't dictate what users do after restoring a backup re. the 'old' session, so we have to support all scenarios. Keeping your settings / sensors is good to have! Why remove those advantages by completely disabling backup if it also can be solved by asking users to log in again?

> Now with that said things like widgets may not function well and I am pretty sure need to be recreated as the ID's change during a device transfer. 

Last time I checked IDs will be migrated over to a new device.

dshokouhi:
> We can't dictate what users do after restoring a backup re. the 'old' session, so we have to support all scenarios. Keeping your settings / sensors is good to have! Why remove those advantages by completely disabling backup if it also can be solved by asking users to log in again?

Well when auto backup does its restore its more obvious that it restored by the user not having to login. If we remove the session then the user still needs to complete onboarding right? At that point how does a user know that everything else was restored? Do we show the old device name and check off location tracking to show we still have that data during onboarding? It becomes a bit more hidden that there was a restore at this point.

jpelgrom:
> If we remove the session then the user still needs to complete onboarding right? (...) Do we show the old device name and check off location tracking to show we still have that data during onboarding?

Yes the user would need to complete onboarding again, and it will look just like a fresh install during that process. It currently does not show stored data.

> It becomes a bit more hidden that there was a restore at this point.

Is that a bad thing?

As the feature request I linked to indicates, there is a demand to be able to restore settings. The current approach tries to keep everything which doesn't work in all scenarios, and completely removing backups feels like overreacting.

dshokouhi:
> > It becomes a bit more hidden that there was a restore at this point.
> 
> Is that a bad thing?

not until another user complains lol



> As the feature request I linked to indicates, there is a demand to be able to restore settings. The current approach tries to keep everything which doesn't work in all scenarios

The current request is actually more a bug report as they say its not working at all but this PR contradicts that. Then later on they request for manual import and export but that is actually not related to android auto backup because we can't manually force it to upload. Auto backup has its own schedule and constraints. I honestly took that request as the app having its own thing for users to export and save elsewhere to restore.

dshokouhi (approved):

jpelgrom:
> The current request is actually more a bug report as they say its not working at all but this PR contradicts that.

While I don't know why the auto backup didn't work for them, I can personally confirm it works (it restores old data for the debug app frequently). The fact that they made a bug report still indicates the demand. 

> Then later on they request for manual import and export but that is actually not related to android auto backup (...)

Yes it is a bug report turned feature request, and even if it's not strictly related if we can't support the auto backup I don't see why a manual backup would be much different - it's the same settings!

dshokouhi:
> While I don't know why the auto backup didn't work for them, I can personally confirm it works (it restores old data for the debug app frequently). The fact that they made a bug report still indicates the demand.

yup same here it works for me too.  Getting logs from something like that will be difficult too, not sure how to proceed there.

JBassett (approved):

PR #2757, opened 2022-08-04
Update geocode location sensor with location updates that are sent to HA

dshokouhi:
## Summary

Noticed that we had some code to update the geocoded sensor but it only updated the DB and did not update HA. Moved the logic to only update the sensor if location is being sent to HA instead of when it was just received. Also only sending the state if it has changed since the last update so we don't send too many updates like in high accuracy mode.

As this sensor should be kept updated with location updates it made sense to only update this one sensor. Updating all sensors when high accuracy mode is on does not sound like something that would be battery friendly. Updating 1 sensor since we are awake and sending an update should have minimal impact.

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
@@ -765,21 +765,21 @@ class LocationSensorManager : LocationSensorManagerBase() {
         lastLocationSend = now
         lastUpdateLocation = updateLocation.gps.contentToString()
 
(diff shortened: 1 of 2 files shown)


jpelgrom:
Not a big fan of duplicating sensor update logic into a random manager class. 

After sending the update `lastSentState` should be updated, right now it'll keep updating until the periodic sensor update completes.

dshokouhi:
> Not a big fan of duplicating sensor update logic into a random manager class.

I agree, moved it to `SensorReceiverBase` so it can be reused




> After sending the update `lastSentState` should be updated, right now it'll keep updating until the periodic sensor update completes.

totally missed that, also made sure to more closely match the original update logic in the function

JBassett (approved):

PR #2766, opened 2022-08-06
Update Wear OS companion app packages

jpelgrom:
## Summary

Wear OS 3 devices use a dedicated companion app instead of one app by Google (like Wear OS 1/2). Due to how the HA app is setup, it only queries for packages at first to show the settings item. This PR updates the list of packages for the first non-Samsung, very expensive Wear OS 3 smartwatch.

## Screenshots

n/a

n/a

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
@@ -286,9 +286,13 @@ class SettingsFragment constructor(
         }
 
         val pm = requireContext().packageManager
-        val hasWearApp = pm.getLaunchIntentForPackage("com.google.android.wearable.app")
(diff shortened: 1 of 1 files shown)


dshokouhi (approved):
In about 2-3 months time we will probably be adding another app to the list for the Pixel Watch, not sure why Google went this route to let each manufacturer do their own thing.

jpelgrom:
> In about 2-3 months time we will probably be adding another app to the list for the Pixel Watch, not sure why Google went this route to let each manufacturer do their own thing.

Instead of creating a giant list of companion apps over the next few months/years, the app should probably create a class that can do the detection using the Google libraries, and returning nothing on non-full versions of the app, similar to how Sentry/crash handling is handled. I don't have a real device to test with though and I don't want to break things, which is why I've submitted this quick fix for now.

PR #2772, opened 2022-08-07
Fix sensorDao not initialized error after geocoded update

dshokouhi:
## Summary

Fixes the following error, by passing in the `sensorDao` variable instead of relying on the lateinit modifier. I was able to reproduce the error by force stopping the app and opening it again. Must be an edge case where we update location before we update sensors completely and initialize the variable.

```
[08-06 11:26:33.414 26944:11135 E/LocBroadcastReceiver]
Could not update location.
kotlin.UninitializedPropertyAccessException: lateinit property sensorDao has not been initialized
	at io.homeassistant.companion.android.common.sensors.SensorReceiverBase.getSensorDao(SensorReceiverBase.kt:57)
	at io.homeassistant.companion.android.common.sensors.SensorReceiverBase.updateSensor(SensorReceiverBase.kt:340)
	at io.homeassistant.companion.android.common.sensors.SensorReceiverBase$updateSensor$1.invokeSuspend(Unknown Source:19)
```

## Any other notes

Ref: https://community.home-assistant.io/t/location-updating-doesnt-work-very-serious-problem-for-my-automations/437206/88?u=dshokouhi

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
@@ -29,6 +29,7 @@ import io.homeassistant.companion.android.common.data.integration.UpdateLocation
 import io.homeassistant.companion.android.common.data.integration.ZoneAttributes
 import io.homeassistant.companion.android.common.sensors.LocationSensorManagerBase
 import io.homeassistant.companion.android.common.sensors.SensorManager
+import io.homeassistant.companion.android.common.sensors.SensorReceiverBase
 import io.homeassistant.companion.android.common.util.DisabledLocationHandler
 import io.homeassistant.companion.android.database.AppDatabase
 import io.homeassistant.companion.android.database.sensor.Attribute
@@ -765,24 +766,17 @@ class LocationSensorManager : LocationSensorManagerBase() {
         lastLocationSend = now
         lastUpdateLocation = updateLocation.gps.contentToString()
 
-        val geoSensorManager = SensorReceiver.MANAGERS.firstOrNull { it.getAvailableSensors(latestContext).any { s -> s.name == commonR.string.basic_sensor_name_geolocation } }
-        val sensorDao = AppDatabase.getInstance(latestContext).sensorDao()
-        val geoSensor = sensorDao.getFull(GeocodeSensorManager.geocodedLocation.id)
-
         ioScope.launch {
             try {
                 integrationUseCase.updateLocation(updateLocation)
                 Log.d(TAG, "Location update sent successfully")
 
                 // Update Geocoded Location Sensor
-                SensorReceiver().updateSensor(
-                    latestContext,
-                    integrationUseCase,
-                    geoSensor,
-                    geoSensorManager,
-                    GeocodeSensorManager.geocodedLocation,
-                    sensorDao
-                )
+                val intent = Intent(latestContext, SensorReceiverBase::class.java)
+                intent.action = SensorReceiverBase.ACTION_UPDATE_SENSOR
+                intent.putExtra("basic_sensor", GeocodeSensorManager.geocodedLocation)
+                GeocodeSensorManager().requestSensorUpdate(latestContext)
+                latestContext.sendBroadcast(intent)
             } catch (e: Exception) {
                 Log.e(TAG, "Could not update location.", e)
             }
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
@@ -766,7 +766,8 @@ class LocationSensorManager : LocationSensorManagerBase() {
         lastUpdateLocation = updateLocation.gps.contentToString()
 
         val geoSensorManager = SensorReceiver.MANAGERS.firstOrNull { it.getAvailableSensors(latestContext).any { s -> s.name == commonR.string.basic_sensor_name_geolocation } }
-        val geoSensor = AppDatabase.getInstance(latestContext).sensorDao().getFull(GeocodeSensorManager.geocodedLocation.id)
+        val sensorDao = AppDatabase.getInstance(latestContext).sensorDao()
+        val geoSensor = sensorDao.getFull(GeocodeSensorManager.geocodedLocation.id)
 
         ioScope.launch {
             try {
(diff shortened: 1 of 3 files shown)


jpelgrom:
> Must be an edge case where we update location before we update sensors completely and initialize the variable.

I have been thinking about this and don't quite follow. Shouldn't Hilt initialize this as soon as you create the class? There is no `sensorDao = ...` in the class.

dshokouhi:
> > Must be an edge case where we update location before we update sensors completely and initialize the variable.
> 
> I have been thinking about this and don't quite follow. Shouldn't Hilt initialize this as soon as you create the class? There is no `sensorDao = ...` in the class.

that I am not sure, I assume when I did the force stop that the update tried to send before the sensor worker kicked in. That might explain why I didnt see it during initial testing?

jpelgrom:
@dshokouhi Now that there has been a stable release with these changes, how often does this appear in Sentry?

dshokouhi:
> @dshokouhi Now that there has been a stable release with these changes, how often does this appear in Sentry?

It's actually a handled exception so it's not being sent to sentry.

jpelgrom:
Decided to look at this in more detail, and it turns out that Hilt does injection for `BroadcastReceiver`s in the `onReceive` function. Because the app creates the class but doesn't call `onReceive`, the fields are never injected.

It is relatively easy to reproduce by adding a custom intent action to update the sensor in the `LocationSensorManager`, and immediately calling `SensorReceiver().updateSensor(...)`, which skips the need for location updates. Then send a broadcast with the action to trigger it, for example: `adb shell am broadcast -a io.homeassistant.companion.android.test.geocoded io.homeassistant.companion.android.debug`.

While this change will prevent the exception, it isn't pretty. It'd be better to add another action to `SensorReceiver`'s intent filter to update an individual sensor so that `onReceive` is used 🙂

dshokouhi:
Ok that makes sense and explains why it only happens in certain cases. So instead of calling the method directly we need to send a intent to process the data. I'll get around to that in next couple of days since I'm traveling today. 

Thanks for looking into this further!

dshokouhi:
Ok cleaned up the update method a bit to take less information as not all the data was easy to send over in the intent. Really only need the `BasicSensor` to handle the individual sensor update.

app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -765,22 +766,17 @@ class LocationSensorManager : LocationSensorManagerBase() {
...
-                SensorReceiver().updateSensor(
-                    latestContext,
-                    integrationUseCase,
-                    geoSensor,
-                    geoSensorManager,
-                    GeocodeSensorManager.geocodedLocation
-                )
+                val intent = Intent(latestContext, SensorReceiverBase::class.java)
+                intent.action = SensorReceiverBase.ACTION_UPDATE_SENSOR
+                intent.putExtra("basic_sensor", GeocodeSensorManager.geocodedLocation)
```
jpelgrom, line 777: `"basic_sensor"` should be a constant to prevent typos, in case this function is used more in the future.

Do we really need to send the entire sensor in the intent? Wouldn't it be better to only send the ID, and handle retrieving the `BasicSensor` class in the sensor receiver?

app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -765,22 +766,17 @@ class LocationSensorManager : LocationSensorManagerBase() {
...
 
                 // Update Geocoded Location Sensor
-                SensorReceiver().updateSensor(
-                    latestContext,
-                    integrationUseCase,
-                    geoSensor,
-                    geoSensorManager,
-                    GeocodeSensorManager.geocodedLocation
-                )
+                val intent = Intent(latestContext, SensorReceiverBase::class.java)
```
jpelgrom, line 775: `SensorReceiverBase` isn't annotated with `@AndroidEntryPoint` so it this intent won't trigger injection and doesn't solve the crash. Using the base is like an implementation detail, so I think the app should send it to `SensorReceiver` instead (which is annotated).

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
```diff
@@ -320,14 +330,12 @@ abstract class SensorReceiverBase : BroadcastReceiver() {
...
-        context: Context,
+    private suspend fun updateSensor(
         integrationUseCase: IntegrationRepository,
         fullSensor: SensorWithAttributes?,
-        sensorManager: SensorManager?,
-        basicSensor: SensorManager.BasicSensor
+        basicSensor: SensorManager.BasicSensor,
+        sensorDao: SensorDao
     ) {
-        sensorManager?.requestSensorUpdate(context)
```
jpelgrom, line 330: Why move updating the sensor out of the `updateSensor` function?

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
```diff
@@ -320,14 +330,12 @@ abstract class SensorReceiverBase : BroadcastReceiver() {
...
 
-    suspend fun updateSensor(
-        context: Context,
+    private suspend fun updateSensor(
         integrationUseCase: IntegrationRepository,
         fullSensor: SensorWithAttributes?,
-        sensorManager: SensorManager?,
-        basicSensor: SensorManager.BasicSensor
+        basicSensor: SensorManager.BasicSensor,
+        sensorDao: SensorDao
```
jpelgrom, line 337: It should no longer be necessary to pass `integrationUseCase` and `sensorDao` in here, as it isn't used like a static function anymore.

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
dshokouhi: I was not seeing the correct approach earlier 🙈

app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
dshokouhi: Ah good point, not sure why I tried to stuff the intent with everything. Will update to send only the ID to retrieve the `SensorManager` and `BasicSensor`

JBassett (approved):

PR #2776, opened 2022-08-08
Fix unable to Toast on QS tiles being clicked error

dshokouhi:
## Summary

Fixes the following sentry errors:

```
java.lang.NullPointerException: Can't toast on a thread that has not called Looper.prepare()
    at com.android.internal.util.Preconditions.checkNotNull(Preconditions.java:167)
    at android.widget.Toast.getLooper(Toast.java:265)
    at android.widget.Toast.<init>(Toast.java:250)
    at android.widget.Toast.makeText(Toast.java:865)
    at android.widget.Toast.makeText(Toast.java:853)
    at android.widget.Toast.makeText(Toast.java:892)
    at io.homeassistant.companion.android.qs.TileExtensions$tileClicked$2.invokeSuspend(TileExtensions.kt:152)
```

and

```
java.lang.RuntimeException: Can't toast on a thread that has not called Looper.prepare()
    at android.widget.Toast$TN.<init>(Toast.java:474)
    at android.widget.Toast.<init>(Toast.java:142)
    at android.widget.Toast.makeText(Toast.java:333)
    at android.widget.Toast.makeText(Toast.java:323)
    at android.widget.Toast.makeText(Toast.java:370)
    at io.homeassistant.companion.android.qs.TileExtensions$tileClicked$2.invokeSuspend(TileExtensions.kt:152)
```

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/qs/TileExtensions.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/qs/TileExtensions.kt
@@ -4,8 +4,6 @@ import android.content.Context
 import android.graphics.Bitmap
 import android.graphics.drawable.Icon
 import android.os.Build
-import android.os.Handler
-import android.os.Looper
 import android.service.quicksettings.Tile
 import android.service.quicksettings.TileService
 import android.util.Log
@@ -151,7 +149,7 @@ abstract class TileExtensions : TileService() {
                     Log.d(TAG, "Service call sent for tile ID: $tileId")
                 } catch (e: Exception) {
                     Log.e(TAG, "Unable to call service for tile ID: $tileId", e)
-                    Handler(Looper.getMainLooper()).post {
+                    withContext(Dispatchers.Main) {
                         Toast.makeText(
                             context,
                             commonR.string.service_call_failure,
@@ -167,7 +165,7 @@ abstract class TileExtensions : TileService() {
             tile.state = Tile.STATE_UNAVAILABLE
             tile.updateTile()
             Log.d(TAG, "No tile data found for tile ID: $tileId")
-            Handler(Looper.getMainLooper()).post {
+            withContext(Dispatchers.Main) {
                 Toast.makeText(
                     context,
                     commonR.string.tile_data_missing,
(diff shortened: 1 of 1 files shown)


jpelgrom:
It works, but considering coroutines are used wouldn't using `withContext(Dispatchers.Main) { ... }` be a cleaner fix?

(Also the UX of this isn't great, the toast isn't visible without collapsing the notification drawer)

dshokouhi:
> It works, but considering coroutines are used wouldn't using `withContext(Dispatchers.Main) { ... }` be a cleaner fix?

I thought that at first but if you notice the first `Toast` message where the actual error points to is already wrapped in `withContext()` so I took it as we need to use `Looper` still. So you're saying we should use 2 `withContext` calls within each other here?



> (Also the UX of this isn't great, the toast isn't visible without collapsing the notification drawer)

I agree, we could temporarily change the tile but then we would need to change it back after some delay. Posting a notification if it fails feels a bit intrusive too.

jpelgrom:
> I thought that at first but if you notice the first `Toast` message where the actual error points to is already wrapped in `withContext()` so I took it as we need to use `Looper` still. So you're saying we should use 2 `withContext` calls within each other here?

The outer one is `withContext(Dispatchers.IO)` right? It looks like a classic example of "the app will crash if you do UI stuff on a non-UI thread" to me.

> > (Also the UX of this isn't great, the toast isn't visible without collapsing the notification drawer)
> 
> I agree, we could temporarily change the tile but then we would need to change it back after some delay. Posting a notification if it fails feels a bit intrusive too.

Yes the options for feedback are limited with tiles. Not something for this PR :)

dshokouhi:
> The outer one is `withContext(Dispatchers.IO)` right?

Correct, so this should be changed to use the main dispatcher there then?

jpelgrom:
> > The outer one is `withContext(Dispatchers.IO)` right?
> 
> Correct, so this should be changed to use the main dispatcher there then?

I suggest replacing `Handler(Looper.getMainLooper()).post {` with `withContext(Dispatchers.Main) {`, only the Toast needs to be posted from the UI thread. I don't know why `withContext(Dispatchers.IO)` was originally used but presumably it was necessary.

dshokouhi:
> I don't know why `withContext(Dispatchers.IO)` was originally used but presumably it was necessary.

It was using `runBlocking{}` before which was an inappropriate blocking call here. I had switched to `withContext()` based on a recommendation I saw on Stackoverflow. Its necessary because we need to wait for some output to properly update the tile based on the response.

jpelgrom (approved):

JBassett (approved):

Issue #2706, opened 2022-07-24
Fetched image previews in notifications do not use mTLS


charlesmunger:
**Home Assistant Android version:**
version beta-2500-f8d0d461-full

**Android version:**
12

**Phone model:**
Pixel 6

**Home Assistant version:**
2022.7.6

**Last working Home Assistant release (if known):**

**Description of problem:**
Notification `image` URLs do not use mTLS and so they don't load if mTLS is required. 

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```
07-23 19:41:29.328 23688  3325 D UrlRepository: localUrl is: true and usesInternalSsid is: false
07-23 19:41:29.328 23688  3325 D UrlRepository: Using external URL
07-23 19:41:30.236 23688  3325 E MessagingService: Couldn't download image for notification
07-23 19:41:30.236 23688  3325 E MessagingService: java.io.FileNotFoundException: https://<url redacted>:8808/api/frigate/notifications/1658616858.243185-7e068v/thumbnail.jpg?format=android
07-23 19:41:30.236 23688  3325 E MessagingService: 	at com.android.okhttp.internal.huc.HttpURLConnectionImpl.getInputStream(HttpURLConnectionImpl.java:255)
07-23 19:41:30.236 23688  3325 E MessagingService: 	at com.android.okhttp.internal.huc.DelegatingHttpsURLConnection.getInputStream(DelegatingHttpsURLConnection.java:211)
07-23 19:41:30.236 23688  3325 E MessagingService: 	at com.android.okhttp.internal.huc.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:30)
07-23 19:41:30.236 23688  3325 E MessagingService: 	at io.homeassistant.companion.android.notifications.MessagingManager$getImageBitmap$2.invokeSuspend(MessagingManager.kt:1276)
07-23 19:41:30.236 23688  3325 E MessagingService: 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
07-23 19:41:30.236 23688  3325 E MessagingService: 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
07-23 19:41:30.236 23688  3325 E MessagingService: 	at kotlinx.coroutines.internal.LimitedDispatcher.run(LimitedDispatcher.kt:42)
[...]

PR #2779, opened 2022-08-08
Use OkHttpClient to download notification videos

jpelgrom:
## Summary

This PR fixes downloading notification videos from servers that have enabled TLS client certificate authentication, [a bug reported in a comment on a related bug here](https://github.com/home-assistant/android/issues/2706#issuecomment-1207245382).

The app will now handle the downloading so that it is possible to provide the necessary authentication, instead of asking the system to do it. The file is removed as soon as the frames are extracted.

## Screenshots

n/a, notifications will show videos again instead of not including them

n/a

## Any other notes

cc @Rogue136198, as you reported this bug I'd appreciate it if you could also verify that the change works

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -79,6 +79,8 @@ import kotlinx.coroutines.withContext
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import org.json.JSONObject
+import java.io.File
+import java.io.FileOutputStream
 import java.net.URL
 import java.net.URLDecoder
 import java.util.Locale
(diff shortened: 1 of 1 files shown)


Rogue136198:
Just tested and everything is working as expected. 

I tested without mTLS and the performance seems to even be a bit better than before. Notifications appear noticeably faster now.

jpelgrom:
> I tested without mTLS

I'm confused, did you als test _with mTLS_? That is what was causing the bug after all.

Rogue136198:
> > I tested without mTLS
> 
> I'm confused, did you als test _with mTLS_? That is what was causing the bug after all.

Oh yes! Sorry I should have been more clear. I tested with mTLS and it works flawlessly. Without mTLS was just for A/B testing on performance when I noticed it felt faster.

dshokouhi (approved):

JBassett (approved):

PR #2795, opened 2022-08-16
Don't attempt to unsubscribe from event when the ID is null

dshokouhi:
## Summary

Found the following sentry error so avoiding the unsubscription when the ID is null

```
java.lang.NullPointerException: null
    at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl$subscribeToEventsForType$2$1$1$1.invokeSuspend(WebSocketRepositoryImpl.kt:206)
```

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
@@ -198,20 +198,17 @@ class WebSocketRepositoryImpl @Inject constructor(
                 eventSubscriptionFlow[eventType] = callbackFlow<T> {
                     eventSubscriptionProducerScope[eventType] = this as ProducerScope<Any>
                     awaitClose {
-                        if (lastResponseID[eventType] == null) {
-                            eventSubscriptionProducerScope.remove(eventType)
-                            eventSubscriptionFlow.remove(eventType)
-                            return@awaitClose
-                        }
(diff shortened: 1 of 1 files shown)


jpelgrom:
I guess this could happen if re-subscribing fails in `handleClosingSocket()`? Otherwise it should be caught by the null check before starting a subscription.

While it might not be possible to unsubscribe correctly, the app should still clean up `eventSubscriptionProducerScope[eventType]` and `eventSubscriptionFlow[eventType]` if they're set to prevent additional re-subscriptions.

dshokouhi:
> I guess this could happen if re-subscribing fails in `handleClosingSocket()`? Otherwise it should be caught by the null check before starting a subscription.

Yea not entirely sure but I do remember you mentioning in the previous PR that this might happen 😬 



> While it might not be possible to unsubscribe correctly, the app should still clean up `eventSubscriptionProducerScope[eventType]` and `eventSubscriptionFlow[eventType]` if they're set to prevent additional re-subscriptions.

good call, this is now done :)

jpelgrom:
Any reason for duplicating code? I think it is nicer to wrap the if statement around the `ioScope.launch` block, which is the only code we need to skip.

dshokouhi:
> Any reason for duplicating code? I think it is nicer to wrap the if statement around the `ioScope.launch` block, which is the only code we need to skip.

umm probably was a bit early for me when I submitted the change lol 😅 

Looking now, probably don't need to return at all and only unsbscribe when not null :)

jpelgrom (approved):

JBassett (approved):

Issue #2796, opened 2022-08-16
Could we get a sensor to indicate fast-charge v normal charge


kenguest:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
It would be useful if the android app could report whether a phone is charging over a fast-charge connection.
Use cases are for presence detection (only fast charger is in the home office, standard chargers elsewhere), and can also indicate if a fast-charge cable is worn out and only delivering a standard charge.

The entity could be sensor.<phone_name>_is_fast_charge (=0/1)

**Describe alternatives you've considered, if any**

**Additional context**

dshokouhi:
Looks like in this case we need to look at the current supplied to `BatteryManager` this means the sensor will not be instant, but when paired with any other battery sensor it should basically be instant.  Here is a stackoverflow post I found describing how to achieve this.

https://stackoverflow.com/questions/38912585/how-to-test-battery-charging-speed

If we do add this it will probably just be a raw value so the user can make their own decision. Seems there is no direct API to determine if fast charging or not. Will the raw value be helpful to you?

In my own personal setup I use a wireless charger in the bedroom and that helps me differentiate between my other chargers because of the `charger_type` sensor

dshokouhi:
We might be able to cheat and decipher fast charging using a native API that tells us how much longer to charge the device? 😂 

https://developer.android.com/reference/android/os/BatteryManager#computeChargeTimeRemaining()

jpelgrom:
> We might be able to cheat and decipher fast charging using a native API that tells us how much longer to charge the device? 😂 

This might be unreliable because 'fast' charging usually slows down when charging is closer to 100%.

dshokouhi:
> This might be unreliable because 'fast' charging usually slows down when charging is closer to 100%.

in that case then even going by the `current` will not suffice right?

https://developer.android.com/reference/android/os/BatteryManager#BATTERY_PROPERTY_CURRENT_NOW

jpelgrom:
I don't know what exactly decreases (volts or amperes), sounds like something to experiment with 🙂. It also might be worth it to look at the AOSP source to find out how it determines the 'fast charging'/'charging'/'charging slowly' text on the lock screen - it will probably not cover OEM-specific fast charging but it is a good starting point.

dshokouhi:
looks like they go by wattage, so both LOL. At least thats what this led me to when I found the string `battery_info_status_charging_fast`

[...]

PR #2799, opened 2022-08-17
Add a battery power sensor

dshokouhi:
## Summary

Fixes: #2796 by adding a new sensor which takes in the current and voltage and sends the wattage back to HA with current and voltage as the attributes.

## Screenshots

Discharging:
image

Charging via USB;
image

Fast charging:
image

Documentation: home-assistant/companion.home-assistant#793

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/BatterySensorManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/BatterySensorManager.kt
@@ -71,12 +71,12 @@ class BatterySensorManager : SensorManager {
             entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC
         )
 
-        private val chargingWattage = SensorManager.BasicSensor(
-            "charging_wattage",
+        private val batteryPower = SensorManager.BasicSensor(
+            "battery_power",
             "sensor",
-            commonR.string.basic_sensor_name_charging_wattage,
-            commonR.string.sensor_description_charging_wattage,
-            "mdi:battery-charging",
+            commonR.string.basic_sensor_name_battery_power,
+            commonR.string.sensor_description_battery_power,
+            "mdi:battery-plus",
             "power",
             unitOfMeasurement = "W",
             stateClass = SensorManager.STATE_CLASS_MEASUREMENT,
@@ -109,7 +109,7 @@ class BatterySensorManager : SensorManager {
             chargerTypeState,
             batteryHealthState,
             batteryTemperature,
-            chargingWattage
+            batteryPower
         )
     }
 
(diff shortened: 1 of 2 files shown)


jpelgrom:
Is "Charging Wattage" the right name for this sensor? It also reports values when not charging.

I wonder how accurate this is - I'm getting 3-5W less than a power plug with monitoring shows.

dshokouhi:
> Is "Charging Wattage" the right name for this sensor?

it was the best name I could come up with unless we want to call it battery wattage?



> It also reports values when not charging.

negative values mean the device is not charging, positive values is when its charging. I decided against setting it to 0 as it can probably indicate heavier drain? the API is still returning values and it seems odd to set the state to 0 or something else when `current` is negative :) I even doubted myself and googled if negative wattage is expected 😂 



> I wonder how accurate this is - I'm getting 3-5W less than a power plug with monitoring shows.

I have no idea but if you notice the value changes a lot over a short period of time so it may not be meaningful if it updates every 15 minutes. The other variables in AOSP are inaccessible to the app so I had to work with what I could get lol.

jpelgrom:
> > Is "Charging Wattage" the right name for this sensor?
> 
> it was the best name I could come up with unless we want to call it battery wattage?

Watt is the unit for electric power, so maybe "battery power"? A plug which provides an entity that shows watts also shows up as [device name] power in HA.

> negative values mean the device is not charging, positive values is when its charging. I decided against setting it to 0 as it can probably indicate heavier drain? the API is still returning values and it seems odd to set the state to 0 or something else when `current` is negative :) I even doubted myself and googled if negative wattage is expected 😂

Maybe the icon should reflect that? `mdi:battery-plus` and `mdi:battery-minus` instead of a generic battery or charging battery.

> I have no idea but if you notice the value changes a lot over a short period of time so it may not be meaningful if it updates every 15 minutes. The other variables in AOSP are inaccessible to the app so I had to work with what I could get lol.

I didn't notice that, probably a good idea to mention it in the sensor description. I'll keep this version installed to see how it changes over time and whether the sensor could be useful or not.

dshokouhi:
> Watt is the unit for electric power, so maybe "battery power"? A plug which provides an entity that shows watts also shows up as [device name] power in HA.

battery power sounds much better 😄 

> Maybe the icon should reflect that? `mdi:battery-plus` and `mdi:battery-minus` instead of a generic battery or charging battery.
> 
> >

Yup that is better icon usage, thank you



> I didn't notice that, probably a good idea to mention it in the sensor description. I'll keep this version installed to see how it changes over time and whether the sensor could be useful or not.

ok will update the description to mention  to get the most of the sensor its best paired with "Fast while charging" frequency updates

JBassett (approved):

Issue #840, opened 2020-08-29
Lock app function only when not at home


masterfish1:
**Is your feature request related to a problem? Please describe.**
I really like the idea of the "Lock app" function but don't feel it necessary to be active at home or in certain zones, for example.

**Describe the solution you'd like**
An option to only have "Lock app" activated when at home, perhaps simply based on whether the phone is connected to the Home Network SSID, for example. 

**Describe alternatives you've considered, if any**
None

**Additional context**
N/A

masterfish1:
Still hoping this can be possible.

dshokouhi:
So I saw that you mentioned zones in the original request however that won't work well for our minimal users who cannot have a location tracker. I think it makes sense to use the connected wifi SSID defined as the Home SSID in App Configuration for this. Based on the title that seems to be the goal here anyways :)

kongo09:
Wouldn't the home SSID not be enough for a device tracker to identify the home zone?

dshokouhi:
> Wouldn't the home SSID not be enough for a device tracker to identify the home zone?

these users can already use the wifi connection sensor to translate if they are at home, the device tracker from `mobile_app` itself is tied to GPS location and is only available in the full flavor of the app.

kongo09:
One other option would be to tie this to the HA URL used.
- internal URL: no lock
- external URL: lock
No complexities with zones, trackers and the like.

dshokouhi:
> One other option would be to tie this to the HA URL used.
> - internal URL: no lock
> - external URL: lock
> 
> No complexities with zones, trackers and the like. 

This would actually follow the same conditions the URL switching does today. Internal URL gets used based on the connected wifi SSID.

alexeiw123:
> One other option would be to tie this to the HA URL used.
> 
> * internal URL: no lock
> * external URL: lock
> 
> No complexities with zones, trackers and the like.

Some users access home assistant using the same URL internally and externally, using NAT loopback.

This feature would be great.

PR #2800, opened 2022-08-17
Option to bypass app lock on home network

RoboMagus:
## Summary

Adds the option to bypass the app (fingerprint) lock when on the home WiFi network.
Implements #840 

## Screenshots

Screenshot_20220817-105358_Home Assistant
Screenshot_20220817-105413_Home Assistant

## Any other notes

* This has been a long open requested feature, so I decided to dable in Kotlin for the first time myself. Please review well, as I'm not at all familiar in this environment!
* I noticed the use of `isInternal()` to check if the user is on their home WiFi network, but this does not work well if the internal URL is _implicitly_ set to the same as the external URL. For the user it is unclear when that's the case. I feel this should be addressed in another issue.


app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -429,6 +434,7 @@ class SettingsFragment constructor(
         val success = result == Authenticator.SUCCESS
         val switchLock = findPreference<SwitchPreference>("app_lock")
         switchLock?.isChecked = success
+        findPreference<SwitchPreference>("app_lock_home_bypass")?.isVisible = success
```
dshokouhi, line 437: We should probably hide or keep this disabled if the user did not configure home wifi SSIDs yet. That way the behavior follows the rest of the settings page like internal connection.

RoboMagus: Thanks for the feedback. I've made changes to keep this setting disabled when no home networks are configured.

dshokouhi (approved):
Looks good to me, gave the debug APK a quick check and functions as expected.

JBassett (approved):

PR #2801, opened 2022-08-17
Support multiple subscriptions of same type in websocket and use in template widget

jpelgrom:
## Summary

This PR updates the app's websocket implementation to support multiple subscriptions of the same type, and partially implements #2769.

**1. Websocket update**

The app will now not only track the `type` of a subscription but also any additional data, in order to support multiple subscriptions where the type is the same but the data is not (like templates, or specific entity updates). 

The logic for these subscriptions is the same as before: if an identical subscription already exists, it will return the existing Flow instead of starting another subscription.

**2. Template widget update**

The changes mentioned above are used to implement the first part of #2769: template widgets now use a subscription for the template instead of refreshing on any state change. Because there might be multiple instances of the widget which start subscriptions, a single `CoroutineScope` is stored in the companion object to tie these subscriptions to - based on my understanding this should be fine but please let me know if it is not!

As a result the `TemplateWidget` no longer implements `BaseWidgetProvider`, and some duplicate code had to be added.

## Screenshots

n/a, no visual changes

Documentation: home-assistant/companion.home-assistant#796

## Any other notes

I opted to combine changes 1 and 2 in this PR, because without change 2 nothing will actually use the new subscription support. Change 2 is all included in commit 4bcf375ef5316ff5c460e5ddabab053f2ca5c884 though.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -161,7 +161,7 @@
                 <action android:name="android.bluetooth.device.action.ACL_DISCONNECTED" />
                 <action android:name="io.homeassistant.companion.android.background.REQUEST_SENSORS_UPDATE" />
                 <action android:name="android.appwidget.action.APPWIDGET_UPDATE" />
-                <action android:name="io.homeassistant.companion.android.widgets.template.BaseWidgetProvider.RECEIVE_DATA" />
+                <action android:name="io.homeassistant.companion.android.widgets.template.TemplateWidget.RECEIVE_DATA" />
             </intent-filter>
 
             <meta-data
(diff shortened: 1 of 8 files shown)


jpelgrom:
Does anyone know why there are all these unrelated intents in the widget's intent receiver?

https://github.com/home-assistant/android/blob/e7dcdacda0090110aa9e86a09335d0172c3395dc/app/src/main/AndroidManifest.xml#L156-L162

dshokouhi:
> Does anyone know why there are all these unrelated intents in the widget's intent receiver?

I think this was when we first added widgets as it was a best effort to keep them as up to date as possible. Then shortly after was when we added the screen on updates and now websocket updates :)

dshokouhi:
So far initial testing looks good. This will cut down on a lot of events! Really impressive.

This PR made me realize that my current template was inefficient so to anyone not seeing instant updates this means you need to optimize your templates 😄 . To determine if your template is actually inefficient test the output in Developer Tools > Templates and if you don't see instant updates there then that means you need to optimize it.

https://www.home-assistant.io/integrations/template/#rate-limiting-updates

dshokouhi (approved):
In my testing things look good here, it is nice to see less state updates coming for the template widget. I haven't observed any issues, just less noise when the screen is on 😃

JBassett (approved):

PR #2802, opened 2022-08-17
Only update geocode sensor if accuracy and time are valid, like location updates

dshokouhi:
## Summary

Noticed that we were updating the Geocoded state `unknown` even when accuracy has not been met, just like location updates we should skip these updates completely instead of sending this state.

I also noticed that we did no time evaluation on the report, so also doing a 5 minute check here as well.


jpelgrom (approved):

dshokouhi:
> This can be considered a bug fix or a breaking change, as some users may rely on the sensor to change to `Unknown` if accuracy isn't met. Documentation suggests that updates would be skipped but looking at the history that has never been the case:
> 
> > Users may adjust [the minimum accuracy] to fit their own needs if they find (...) not enough reports.
> 
> Either way I approve but it should be more than a generic bug fix in the changelog.

I agree, we should call it out since the behavior changed here.  My hope with this change is that users should see less `unknown` states that get sent anytime accuracy was not met.  Hopefully the data will be more consistent to the device tracker. 

One other improvement idea I had was to possibly just feed in the GPS coordinate from the device tracker to really keep them in sync.

JBassett (approved):

Issue #2688, opened 2022-07-18
WearOS - Selecting entity to use in complication needs to be faster


sepodele:
**Is your feature request related to a problem? Please describe.**

The new WearOS complication feature is awesome but when I change between watch faces I have to reselect what entities I want to use in the complication. When you have 1000 entities, scrolling through the list takes forever.

**Describe the solution you'd like**

1. Groups (Automations, Binary Sensor, Buttons, etc.) should be collapsed, not expanded by default.
2. Favorites set in the companion app are not shown here. It should be and it should be a group itself sorted before the other groups.
3. A search option to find the entity you need fast.

Vedeneb:
Can confirm this, for me it's even worse: I can't select any entity because the app just freezes after loading the list of entities. After 30 seconds or so the app finally crashes. Currently in the cleanup process of my home assistant instance with more than 1000 entities as well :)

dshokouhi:
> After 30 seconds or so the app finally crashes.

Can you please open a new issue and get us the crash logs from this experience? We should not be crashing the app.

sygys:
The wear part of the app is crashing and lagging all the time for me. no matter what i do. its due to the large amount of entities i have. They can never be loaded all into the watch. it tries to load the entities and after some time it crashes back to the main menu on my galaxy watch 4. the approach here in the wear app is all wrong. it tries to load all entities from home assistant. In my case this means it tries to load 2800 entities and show them in a list. this is never going to work on a watch. 

we need to either being able to select entities in the browser or in the phone app that will be shown in the wear app. or it shouldnt load all the entities but only a part of the, or hell maybe make a search function in which you send a search request to home assistant on the phone to give only a select few results in a list on the watch. this wear app is fun if you are a beginning user with 100 entities but its plain useless when you have people like me with their whole home automated.

dshokouhi:
> The wear part of the app is crashing and lagging all the time for me. no matter what i do. its due to the large amount of entities i have. They can never be loaded all into the watch. it tries to load the entities and after some time it crashes back to the main menu on my galaxy watch 4. the approach here in the wear app is all wrong. it tries to load all entities from home assistant. In my case this means it tries to load 2800 entities and show them in a list. this is never going to work on a watch.

Please get us the logs so we can see the cause of the crash, there are other users who also have a high amount of entities and the selection of entities loads fine for them. Also that is a bug and not a feature request so it should be filed separately with the requested information.  You will need to access logcat from the watch to do this.



[...]

PR #2804, opened 2022-08-18
Faster entity selection Wear OS

leroyboerefijn:
## Summary
Implements #2688 by starting with all collapsed headers for entity selection, and adding favorites section. I propose to make the search function into a separate FR/PR.

I also replaced the +/- symbols for the material expand and collapse icons.

## Screenshots
Screenshot_20220818_220348_debug

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/views/ChooseEntityView.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/views/ChooseEntityView.kt
@@ -78,15 +78,14 @@ fun ChooseEntityView(
                 if (expandedFavorites) {
                     items(favoriteEntityIds.value.size) { index ->
                         val favoriteEntityID = favoriteEntityIds.value[index].split(",")[0]
-                        for (entity in entitiesByDomain.flatMap { (key, values) -> values }) {
-                            if (entity.entityId == favoriteEntityID) {
+                        entitiesByDomain.flatMap { (_, values) -> values }
+                            .firstOrNull { it.entityId == favoriteEntityID }
+                            ?.let {
                                 ChooseEntityChip(
-                                    entity = entity,
+                                    entity = it,
                                     onEntitySelected = onEntitySelected
                                 )
-                                break // no need to search further
                             }
-                        }
                     }
                 }
             }
--- a/wear/src/main/java/io/homeassistant/companion/android/views/ChooseEntityView.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/views/ChooseEntityView.kt
@@ -3,6 +3,11 @@ package io.homeassistant.companion.android.views
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.State
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.runtime.setValue
 import androidx.compose.runtime.snapshots.SnapshotStateList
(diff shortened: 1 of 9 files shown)


dshokouhi:
> I propose to make the search function into a separate FR/PR.

I agree, that is a much larger task.

I was hopeful this may fix my bug but the crashing is still there :( hopefully this makes life easier for some users :)

leroyboerefijn:
> I was hopeful this may fix my bug but the crashing is still there :( hopefully this makes life easier for some users :)

Yeah was hoping so too, but I'm still hopeful we can figure out a way to reduce recompositions:)

leroyboerefijn:
@dshokouhi, can you check if the latest version fixes the crash for you? I removed the unneeded live updates that @jpelgrom found

wear/src/main/java/io/homeassistant/companion/android/views/ExpandableListHeader.kt
```diff
@@ -40,9 +46,13 @@ fun ExpandableListHeader(
             .clickable { onExpandChanged(!expanded) }
     ) {
         Row {
-            val plusMinus = if (expanded) "-" else "+"
             Text(
-                text = "$string\u2001$plusMinus"
+                text = string
+            )
+            Spacer(modifier = Modifier.width(4.dp))
+            Icon(
```
jpelgrom, line 53: Instead of bundling new XML drawables, you can also use the Material icons that are already bundled with the app in the Android-Iconics library:

```kotlin
Image(
    asset = if (expanded) CommunityMaterial.Icon.cmd_chevron_up else CommunityMaterial.Icon.cmd_chevron_down,
    contentDescription = stringResource(if (expanded) commonR.string.collapse else commonR.string.expand),
    colorFilter = ColorFilter.tint(LocalContentColor.current)
)
```

wear/src/main/java/io/homeassistant/companion/android/views/ChooseEntityView.kt
```diff
@@ -59,6 +66,31 @@ fun ChooseEntityView(
...
+                        val favoriteEntityID = favoriteEntityIds.value[index].split(",")[0]
+                        for (entity in entitiesByDomain.flatMap { (key, values) -> values }) {
+                            if (entity.entityId == favoriteEntityID) {
+                                ChooseEntityChip(
+                                    entity = entity,
+                                    onEntitySelected = onEntitySelected
+                                )
+                                break // no need to search further
+                            }
+                        }
```
jpelgrom, line 89: Using a for loop and breaking it works but using Kotlin language features makes it cleaner in my opinion:

```suggestion
                        entitiesByDomain.flatMap { (_, values) -> values }
                            .firstOrNull { it.entityId == favoriteEntityID }
                            ?.let {
                                ChooseEntityChip(
                                    entity = it,
                                    onEntitySelected = onEntitySelected
                                )
                            }
```

dshokouhi:
> @dshokouhi, can you check if the latest version fixes the crash for you? I removed the unneeded live updates that @jpelgrom found

There we go its loading now and I was able to select the complication! It seems to be a lot faster too!  Great work!!

wear/src/main/java/io/homeassistant/companion/android/views/ExpandableListHeader.kt
leroyboerefijn: Good suggestion, did not think about it

wear/src/main/java/io/homeassistant/companion/android/views/ChooseEntityView.kt
leroyboerefijn: Yes, like this style as well 😊
I did this one now, will do the same in `MainView` later today as well!

jpelgrom (approved):

dshokouhi (approved):
looks good to me

PR #2805, opened 2022-08-18
Add exclude recents flag to avoid showing multiple open activities in recents

dshokouhi:
## Summary

We had a few closed issues and user complaints before and I was just reading the description for this flag and its behavior works well with our webview intents which require a new task to be created in order to properly load and change the URL. I don't think there are any other places to add this. In device controls we use a different flag so it is not impacted by showing multiple activities in recent apps.

I decided against adding this to sensor detail intent as I figured those users may actually need to navigate back to sensor settings faster if they need to make any further changes outside of granting permissions.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -1587,6 +1587,7 @@ class MessagingManager @Inject constructor(
             intent.putExtra("fragment", NOTIFICATION_HISTORY)
         intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
         intent.addFlags(Intent.FLAG_ACTIVITY_MULTIPLE_TASK)
+        intent.addFlags(Intent.FLAG_ACTIVITY_EXCLUDE_FROM_RECENTS)
 
         return PendingIntent.getActivity(
             context,
@@ -1886,6 +1887,7 @@ class MessagingManager @Inject constructor(
             else
                 WebViewActivity.newInstance(context, title)
             intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
+            intent.addFlags(Intent.FLAG_ACTIVITY_EXCLUDE_FROM_RECENTS)
             context.startActivity(intent)
         } catch (e: Exception) {
             Log.e(TAG, "Unable to open webview", e)
--- a/app/src/main/java/io/homeassistant/companion/android/settings/shortcuts/ManageShortcutsViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/shortcuts/ManageShortcutsViewModel.kt
@@ -109,6 +109,7 @@ class ManageShortcutsViewModel @Inject constructor(
         )
         intent.action = shortcutPath
         intent.addFlags(Intent.FLAG_ACTIVITY_MULTIPLE_TASK)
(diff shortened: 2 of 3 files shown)


jpelgrom:
> our webview intents (...) require a new task to be created in order to properly load and change the URL

I haven't extensively tested, but is there any reason why they can't be handled using `onNewIntent()`? This PR also changes one, possibly two intents for settings which presumably aren't affected by the webview?

The change works as described but if the app is opened using one of these intents and wasn't open in recents, the user will end up with no Home Assistant in their recent apps which is unexpected.

dshokouhi:
> > our webview intents (...) require a new task to be created in order to properly load and change the URL
> 
> I haven't extensively tested, but is there any reason why they can't be handled using `onNewIntent()`? This PR also changes one, possibly two intents for settings which presumably aren't affected by the webview?
> 
> The change works as described but if the app is opened using one of these intents and wasn't open in recents, the user will end up with no Home Assistant in their recent apps which is unexpected.

So based on the description of the required flag I don't think its going to be reliable for our Webview instance

`If set, the activity will not be launched if it is already running at the top of the history stack.`

In all my testing this causes issues when trying to navigate from dashboard to dashboard and entity to entity using shortcuts and notifications only. They eventually get stuck and it reused the previous activity which this sounds like. We actually needed that new instance to be created to reliably navigate to the new path.

Ideally I think what we need is a flag that will remove the other activities and start a new instance at the top of the history stack.

I agree it may not be desirable but its also weird to see multiple instances of the app side by side.

The reason why I added it to the settings intents is because those also launch with the same flags which means a user can also have multiple of them in the stack. Even just showing 1 webview and 1 settings in recents will probably confuse a user as they expect to see 1 instance per app.

jpelgrom:
I still don't quite follow why re-using the same activity / when received in `onNewIntent()` loading the new pages and adding it to the history will cause "it" (?) to get stuck, but this is not the place to discuss it.

> Ideally I think what we need is a flag that will remove the other activities and start a new instance at the top of the history stack.

`Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK`? Assuming the app doesn't start even more new tasks during regular use.

dshokouhi:
> I still don't quite follow why re-using the same activity / when received in `onNewIntent()` loading the new pages and adding it to the history will cause "it" (?) to get stuck, but this is not the place to discuss it.

This seems like it would be a larger change. I have tried to implement on `onNewIntent()` by forcing the websocket content intent to have only the `SINGLE_TOP` flag as required for `onNewIntent()` however I still saw multiple instances of the app. I think we may need to make more changes for `onNewIntent()` to work properly so might be best for a new PR? 

> `Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK`? Assuming the app doesn't start even more new tasks during regular use.

Unfortunately this does not work and results in new entries in the stack

jpelgrom:
> I have tried to implement on `onNewIntent()` by forcing the websocket content intent to have only the `SINGLE_TOP` flag as required for `onNewIntent()` however I still saw multiple instances of the app. I think we may need to make more changes for `onNewIntent()` to work properly so might be best for a new PR?

It also works with other flags, see [the documentation](https://developer.android.com/guide/components/activities/tasks-and-back-stack#IntentFlagsForTasks).

Let's keep this PR as it is though, based on your description this will 'solve' a problem users currently experience without major downsides, using `onNewIntent()` can be done in the future with more testing.

jpelgrom (approved):

JBassett (approved):

PR #2809, opened 2022-08-20
Breaking Change: Add some more attributes to geocoded sensor and check that device supports it

dshokouhi:
## Summary

Breaking Change: The app will now combine `latitude` and `longitude` into the `location` attribute just like iOS and old app behavior.

- Checks that the device actually supports the geocoder otherwise hide it from users. The sensor was probably already broken for them anyways 😅 
- Add attributes for language, feature name, phone, premises and URL

Documentation: home-assistant/companion.home-assistant#794


app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
```diff
@@ -98,14 +102,19 @@ class GeocodeSensorManager : SensorManager {
                     "administrative_area" to it.adminArea,
                     "country" to it.countryName,
                     "iso_country_code" to it.countryCode,
+                    "language" to it.locale.displayLanguage,
```
jpelgrom, line 105: This always seems to match the device language, is that useful information to have here?

dshokouhi: If you leave the country it should update to match the visiting locale. At least that's how I took the description lol 

https://developer.android.com/reference/java/util/Locale#getDisplayLanguage()

https://developer.android.com/reference/android/location/Address#getLocale()

jpelgrom: I think it matches the locale used to generate the Address object. Otherwise changing the device language shouldn't change this value, but it does.

For example, I'm in The Netherlands and have my device set to English. The geocoder shows `language` right now as "English". If your idea was correct that should be "Dutch", correct? If I change the device language to German it'll show "Deutsch" for `language`, etc.

dshokouhi: Ah that is strange, I must've misunderstood it when it said `Returns the Locale associated with this address.` I took it as it applies to that `address` like all other attributes lol. Will remove this attribute then :)

jpelgrom (approved):

dshokouhi:
I am debating on making this PR a breaking change to help fix/alleviate an issue https://github.com/home-assistant/frontend/issues/11970 

What are our thoughts on either removing the attributes of `latitude` and `longitude` or renaming them to `address_latitude` or something similar?

I spoke with @balloob  regarding the above issue and it seems the best and only fix is to avoid using those attributes so the map behaves as expected.

jpelgrom:
It looks like the iOS app sends an array in the attributes named `Location` where the first value is latitude, second value is longitude:

https://github.com/home-assistant/iOS/blob/ef9e0f3965b1adf0af936a8a6d9291ee90e0ab0f/Sources/Shared/API/Webhook/Sensors/GeocoderSensor.swift#L102

If we're going to make it a breaking change, we might as well align it? Or does the issue also apply to iOS?

dshokouhi:
We actually used to do that but I think when we started the attribute conversions we ended up splitting them like they are today. Now that attributes send the proper last format we should go back to the old way. Good call

jpelgrom (approved):

hofst:
@dshokouhi This change broke the ability to visualize geocoded locations in custom MapCards inside dashboards. Wouldn't it make more sense to to change all apps (including ios) to have lattitude and longitude as separate attributes? That not only causes fewer problems but makes also the usage of those values easier (without requiring string parsing)

dshokouhi:
> @dshokouhi This change broke the ability to visualize geocoded locations in custom MapCards inside dashboards. Wouldn't it make more sense to to change all apps (including ios) to have lattitude and longitude as separate attributes? That not only causes fewer problems but makes also the usage of those values easier (without requiring string parsing)

Please do not comment on merged PRs

PR #2842, opened 2022-09-02
Update change log week of 9/2

dshokouhi:
## Summary

First update for the next release.

I opted to mention the template widget update as a breaking change in case there are other templates that needed to be adjusted like mine.

Also attempting to give a bit more info on the changes, not too much but hopefully enough to answer some more questions regarding the breaking changes.

### DIFF
--- a/app/src/main/res/xml/changelog_master.xml
+++ b/app/src/main/res/xml/changelog_master.xml
@@ -3,7 +3,7 @@
     tools:ignore="MissingDefaultResource">
     <release version="2022.9 - Phone" versioncode="2">
         <change>&lt;b&gt;Breaking Change:&lt;/b&gt; Geocoded Location sensor attributes for latitude and longitude have been combined into location, the sensor will no longer show on the map. Additional attributes also added to the sensor.</change>
-        <change>&lt;b&gt;Breaking Change:&lt;/b&gt; Template widgets will only update when the template detects a change. If you are not seeing frequent updates then you are hit by rate limiting and will need to readjust the template. Test the template in Developer Tools > Templates to check updates are as expected. Old behavior would update any time any entity had a state change, this change will reduce traffic to the device.</change>
+        <change>&lt;b&gt;Breaking Change:&lt;/b&gt; Template widgets will only update when the template detects a change. If you are not seeing frequent updates then you are hit by rate limiting and will need to readjust the template.</change>
         <change>Bug fixes, performance improvements, design tweaks and dependency updates</change>
     </release>
     <release version="2022.9 - Wear" versioncode="1">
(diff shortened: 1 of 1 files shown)


jpelgrom:
> Also attempting to give a bit more info on the changes, not too much but hopefully enough to answer some more questions regarding the breaking changes.

I'm not sure this is the right way forward - the dialog already fills half the screen for me and is dense, and we're only one week into a new release cycle.
A screenshot of the app, with the 'Change Log' dialog filling half of the screen

Merging this is fine*, but the last update shows that users already aren't reading all the info in the dialog and I don't expect adding more text will change that. Perhaps the changelog should be moved to the companion website and/or we should be more selective in what is highlighted in this dialog?

\* though there is still time for more changes before the next beta on Sunday

dshokouhi:
> I'm not sure this is the right way forward - the dialog already fills half the screen for me and is dense, and we're only one week into a new release cycle.

So we had complaints on the forum that the last breaking change mention didn't mean much to users even though we clearly said everything changed check the docs for the updated format. So I was trying something new for the users who actually read this changelog.



> Merging this is fine*, but the last update shows that users already aren't reading all the info in the dialog and I don't expect adding more text will change that.

Its tough because you have users who look forward to the change log as it tells them about new features and they will appreciate more info. Then you have users who skip the change log and then panic when an issue occurs. Its going to be impossible to satisfy all users here.



> Perhaps the changelog should be moved to the companion website and/or we should be more selective in what is highlighted in this dialog?

Adding to the companion docs would only help the users who actually read the docs which really not too many users do. Even this last time when we told users to check the docs for the updated formats they skipped the examples and were still confused.

This last breaking change was a one time thing that had a quick fix, for the most part we do deprecate features and provide a better workaround. So I don't think there will be too many complaints next time but as with every release you always have users who don't read things and complain when things break. In our case we provide a lot of tools for users to succeed but they still don't know those tools exist. We can't detect when a user has auto update enabled but at the same time our app is not the only app susceptible to breaking changes (Kodi being a big one that always bites users on each major version). We do provide a current version sensor but we still have users who don't know that sensor exists.

Even if we decide to post a notification of breaking changes we will still get complaints from users because they will say the notification is meaningless to those who don't manage the server. This is kinda a lose-lose situation for us. We can definitely experiment with adding something to the docs but we probably want the iOS team on board too otherwise we will get other complaints.

jpelgrom:
I agree we can't provide a perfect solution for all users, but do feel like more info about major changes for those who want it can also be provided by, for example, a link instead of a wall of text that is shown to everyone.

Just try to not describe all changes this extensively, at that point we might as well send them to the GitHub changelog with all merged PRs 😉

dshokouhi:
ok let me remove the last couple of sentences since the docs mention it already. Maybe we can place a link to the docs in the change log if we think it will help? Although we already link to the docs in multiple places in settings already so not sure how much that will help.

jpelgrom:
Sorry I didn't get back to this in time - no nice changelog for the latest Play Store beta I guess. Removing those last few sentences makes it a lot better already 👍

JBassett (approved):

Issue #2846, opened 2022-09-03
Battery current does not always report in microamperes


CamFlyerCH:
**Home Assistant Android version:** 2022.8.0-full

**Android version:** 12

**Phone model:** Samsung Galaxy S21 5G / SM-G991B/DS

**Home Assistant version:** 2022.8.7

**Last working Home Assistant release (if known):**

**Description of problem:**
The sensor battery power measures the current 1000x to small. Hence the watt are also 1000x to small und unusable. Same issue during charging and discharging. I compared the values to a current meter in my charger and to an app AccuBattery pro.
Voltage seams fine !

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```

**Screenshot of problem:**
![image](https://user-images.githubusercontent.com/25889580/188281646-95701ee9-e7d1-4329-8961-844b1802cc36.png)

![image](https://user-images.githubusercontent.com/25889580/188281678-8337d2dd-82bd-46a4-a088-da5e7b1aa488.png)

**Additional information:**

dshokouhi:
The app is actually converting microamps to amps.

https://github.com/home-assistant/android/blob/ef5ec73a877f735112ac2a192e4717b83ecd7f69/common/src/main/java/io/homeassistant/companion/android/common/sensors/BatterySensorManager.kt#L323

https://developer.android.com/reference/android/os/BatteryManager#BATTERY_PROPERTY_CURRENT_NOW

It seems some manufacturers may not be providing microamp values? On my pixel the conversion is correct otherwise it would be very high.

![image](https://user-images.githubusercontent.com/1634145/188281966-eb641b40-58cd-4b73-8b6f-1e7a7b10a97e.png)

On my Amazon tablet its also completely different

![image](https://user-images.githubusercontent.com/1634145/188282076-601ead53-4d8b-4b0a-a7aa-64a4b178891e.png)

CamFlyerCH:
After some searching it seams to be a known issue with this value ! 
[https://play.google.com/store/apps/details?id=com.competit.battery_monitor](https://play.google.com/store/apps/details?id=com.competit.battery_monitor)
This app has the same issue.

[...]

PR #2848, opened 2022-09-04
Add a setting to battery power sensor to convert current when it is not in microamps

dshokouhi:
## Summary

Fixes: #2846 by adding a setting so the user can adjust the divisor for the units their device supplies. Seems we have may have at least 3 different units even though android documentation says the value will be in micoramps.

Documentation: home-assistant/companion.home-assistant#804


JBassett (approved):
Gotta love the consistency...

CamFlyerCH:
I tested this with the latest beta APK and it works perfectly. I had to change the divider from 1000000 to 1000 for my Galaxy S21 and now get the correct current and also with that also the right power values. There is a reasonable difference while charging between my charging cable's display that shows 13 W and the value in HA that reads 10.51 W. I think this the loss in the charging circuits and the reason why the phone heats up during charging. Thanks for this great and quick improvement !

PR #2856, opened 2022-09-05
Add item ID to active notification attributes to keep them unique

dshokouhi:
## Summary

Was helping a user on discord who mentioned they did not see all notification data in the attributes for active notification sensor, turns out we need to add the ID as well to keep the attributes unique otherwise they will overwrite each other.

Ref: https://discord.com/channels/330944238910963714/562408603345092636/1016456857574985738

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
@@ -227,15 +227,15 @@ class NotificationSensorManager : NotificationListenerService(), SensorManager {
         try {
             val attr: MutableMap<String, Any?> = mutableMapOf()
             for (item in activeNotifications) {
-                attr += mappedBundle(item.notification.extras, "_${item.packageName}").orEmpty()
-                    .plus(item.packageName + "_" + item.id + "_post_time" to item.postTime)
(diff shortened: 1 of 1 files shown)


jpelgrom (approved):
It works

(though I'm not sure how useful it is and if it meets the current standards, dumping so many attributes in a sensor and with slightly different formats for package name + id depending on where the attribute is from)

dshokouhi:
I havent seen anything in the docs about too many attributes, I normally refer to the [data docs](https://data.home-assistant.io/docs/states) to make sure the DB will be fine. The attributes in this case describe the state as best as possible. At least now we are not missing any attribute data. I think in this case the user wants to display the notifications from the device.

ScottG489:
Yeah I think it's great that this information is available.

I do somewhat agree with @jpelgrom, but I'm not sure there's an alternative. Sensors and attributes don't provide a great way to create a data structure. I almost wonder if it would be better to cram them all into a single attribute as json or something. I might even like that idea less, but it would at least make it easier to get at the information in a structured way.

Perhaps we could consider documenting the format we're using to create attributes?

JBassett (approved):
Approving for now since people who are using this really need to know what they are doing already.  However, I agree we are starting to get out of the spirit of what a `sensor` is for Android

PR #2860, opened 2022-09-06
Improve adding quick settings tile

jpelgrom:
## Summary

This PR bundles a few improvements to the quick settings tiles setup process:

 - Instead of disabling a tile in the QS panel when it hasn't been setup, make it inactive and deeplink to the specific tile's settings when tapping it. This makes it easier to set up the tile if you haven't browsed the app settings, and also helps with the Toast message not appearing / being hidden behind the QS panel on recent Android versions.
 - When using Android 13, [use the new quick settings placement API](https://developer.android.com/about/versions/13/features#quick-settings) to allow adding a tile from the app's settings. If the tile was already added it will update like before.
 - Don't show the tile subtitle option on Android versions before 10, as it simply isn't available for those versions.

## Screenshots

A tile that hasn't been setup will now be inactive instead of disabled (appearance depends on Android version):
An inactive tile with the Home Assistant logo, titled 'Title 1' and the subtitle 'Requires setup'

After tapping it, the settings will open for that specific tile with a message that the tile requires setup before use:
|Light|Dark|
|-----|-----|
|Tile settings, with a message 'You need to set up the tile before using it' at the bottom of the screen, light mode.|Tile settings, with a message 'You need to set up the tile before using it' at the bottom of the screen, dark mode.|

When using Android 13, you will now see the 'Update Tile Data' button replaced with 'Add Tile' if the tile hasn't been added to the quick settings panel yet:
|Light|Dark|
|-----|-----|
|Tile settings that show a blue button 'Add Tile' in the place of where before 'Update Tile Data' was shown, light mode.|Tile settings that show a blue button 'Add Tile' in the place of where before 'Update Tile Data' was shown, dark mode.|

After tapping on 'Add Tile', the system will show a dialog asking if you want to add the tile, with a basic preview:
|Light|Dark|
|-----|-----|
|A dialog on top of the app titled 'Home Assistant wants to add the following tile to Quick Settings', with a preview of the tile that was setup in the app and buttons titled 'Do not add tile' and 'Add tile', light mode.|A dialog on top of the app titled 'Home Assistant wants to add the following tile to Quick Settings', with a preview of the tile that was setup in the app and buttons titled 'Do not add tile' and 'Add tile', dark mode.|

Documentation: home-assistant/companion.home-assistant#808

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/qs/Tile10Service.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/qs/Tile10Service.kt
@@ -8,7 +8,7 @@ import androidx.annotation.RequiresApi
 class Tile10Service : TileExtensions() {
 
     companion object {
-        private const val TILE_ID = "tile_10"
+        const val TILE_ID = "tile_10"
     }
 
     override fun getTile(): Tile? {
--- a/app/src/main/java/io/homeassistant/companion/android/qs/Tile11Service.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/qs/Tile11Service.kt
@@ -8,7 +8,7 @@ import androidx.annotation.RequiresApi
 class Tile11Service : TileExtensions() {
 
     companion object {
-        private const val TILE_ID = "tile_11"
+        const val TILE_ID = "tile_11"
     }
 
     override fun getTile(): Tile? {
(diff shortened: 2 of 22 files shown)


dshokouhi (approved):
Looks good to me, tested debug APK and it works well.

I did spot a bug in Android 13 or at least on my device where I couldn't drag up a new tile beyond a certain point however, that happened for other apps as well like AutoNotifications which adds 40 tiles. Definitely not something introduced here but just mentioning it in case anyone else see's that issue during testing.

JBassett (approved):

PR #2861, opened 2022-09-06
Don't collapse lists by default if there's only one list

### LINKED ISSUE #2858: Wear os companion areas and domains wont load anything
Running latest beta home assistant wear os app will show favorites fine but when opening the chips for areas or domains like light they just show a black screen with the name and no entities unable to scroll or do anything other than go back.

All entities shows everything OK and works.

jpelgrom:
## Summary

If there is only one collapsible list, the app shouldn't collapse it by default as the app may not use a collapsible header in that case. Fixes #2858.

## Screenshots

n/a

n/a

## Any other notes

cc @leroyboerefijn somehow we missed this in #2804


leroyboerefijn:
Good find! Completely missed that if-else statement before...

dshokouhi (approved):
Thanks for the quick fix!

JBassett (approved):

Issue #2876, opened 2022-09-13
Player is drawing over notification and navigation bar when in full screen


felipecrs:
**Home Assistant Android version:** beta-2692-8713bb96-full

**Android version:** 12

**Phone model:** Samsung S20 FE 5G

**Home Assistant version:** 2022.9.2

**Last working Home Assistant release (if known):** N/A

**Description of problem:** When entering in full screen by clicking the full screen icon in any of the players, such as the Picture Glance one, the player draws over the notification area and also over the the navigation bar, while not hiding any of them.

I initially submitted this issue to https://github.com/dermotduffy/frigate-hass-card/issues/817, but then we realized there is nothing which can be made from either the card or the HA frontend, since APIs for handling these bars in Android does not seem to be exposed by the browser.

- https://developer.android.com/develop/ui/views/layout/immersive
- https://developer.android.com/develop/ui/views/layout/display-cutout

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):** N/A

**Screenshot of problem:**

![ha](https://user-images.githubusercontent.com/29582865/188486805-9bb75add-fda2-42a2-b0e4-a26cccab694e.jpg)

https://user-images.githubusercontent.com/29582865/188494573-156cb2ce-09c8-4e06-83cc-b6e826a471ce.mp4

**Additional information:**

I have two suggestions:

1. Somehow wrap these Android APIs and expose them to the HA frontend, which then can control based on their desire/contents. This can take a very long time to implement.
2. As a short term measure, when the frontend requests full screen, the Android app should automatically hide the navigation bar. The Android app should also NOT allow contents to be displayed over notch, since it may produce undesirable behaviors (it should only be allowed when number 1 gets implemented).

dshokouhi:
Can you reproduce the issue in chrome for android on the same device?

felipecrs:
Slightly: on the same device, on Chrome, the navigation bar and the notification bar hide properly. Ideally it should also NOT draw over the notch as well, but that's ok since it can be controlled by CSS (https://www.w3.org/TR/css-round-display-1/#viewport-fit-descriptor).

felipecrs:
https://user-images.githubusercontent.com/29582865/189950548-d8d86029-b63f-466d-bfbf-7d40d664f70d.mp4

dshokouhi:
[...]

PR #2878, opened 2022-09-13
Fix full screen mode for all devices

dshokouhi:
## Summary

Fixes: #2876 by using the [recommended methods](https://developer.android.com/develop/ui/views/layout/immersive) from Google, most likely these were updated after we had this code. Changed the theme to reflect for displays with cutouts too based on [updated recommendation](https://developer.android.com/develop/ui/views/layout/display-cutout).

Tested on my pixel 6 pro by enabling the setting and also keeping the setting disabled and take a video into full screen. Also tested on an android 7 device to ensure the changes still work on older devices.

FWIW current code still continues to work on android 7 devices so this issue seems to impact newer version of android.

### DIFF
--- a/app/src/main/res/values-v27/styles.xml
+++ b/app/src/main/res/values-v27/styles.xml
@@ -14,4 +14,18 @@
         <item name="android:windowLightStatusBar">@bool/isLightMode</item>
         <item name="android:windowLightNavigationBar">@bool/isLightMode</item>
     </style>
+
+    <style name="Theme.HomeAssistant" parent="Theme.MaterialComponents.DayNight.NoActionBar">
+        <item name="colorPrimary">@color/colorPrimary</item>
+        <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
+        <item name="colorAccent">@color/colorAccent</item>
+        <item name="colorSecondary">@color/colorAccent</item>
+        <item name="colorSecondaryVariant">@color/colorAccent</item>
+        <item name="colorButtonNormal">@color/colorPrimary</item>
+        <item name="android:colorBackground">@color/colorBackground</item>
+        <item name="android:navigationBarColor">@color/colorPrimaryDark</item>
+        <item name="alertDialogTheme">@style/Theme.HomeAssistant.Dialog.Alert</item>
+        <item name="actionBarPopupTheme">@style/ThemeOverlay.MaterialComponents.Dark</item>
+        <item name="android:windowLayoutInDisplayCutoutMode">shortEdges</item>
+    </style>
 </resources>
--- a/app/src/main/res/values-v27/styles.xml
+++ b/app/src/main/res/values-v27/styles.xml
@@ -15,17 +15,4 @@
         <item name="android:windowLightNavigationBar">@bool/isLightMode</item>
     </style>
 
-    <style name="Theme.HomeAssistant" parent="Theme.MaterialComponents.DayNight.NoActionBar">
-        <item name="colorPrimary">@color/colorPrimary</item>
-        <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
-        <item name="colorAccent">@color/colorAccent</item>
-        <item name="colorSecondary">@color/colorAccent</item>
-        <item name="colorSecondaryVariant">@color/colorAccent</item>
-        <item name="colorButtonNormal">@color/colorPrimary</item>
-        <item name="android:colorBackground">@color/colorBackground</item>
-        <item name="android:navigationBarColor">@color/colorPrimaryDark</item>
-        <item name="alertDialogTheme">@style/Theme.HomeAssistant.Dialog.Alert</item>
-        <item name="actionBarPopupTheme">@style/ThemeOverlay.MaterialComponents.Dark</item>
-        <item name="android:windowLayoutInDisplayCutoutMode">shortEdges</item>
-    </style>
 </resources>
--- a/app/src/main/res/values-v27/styles.xml
+++ b/app/src/main/res/values-v27/styles.xml
@@ -14,5 +14,4 @@
         <item name="android:windowLightStatusBar">@bool/isLightMode</item>
         <item name="android:windowLightNavigationBar">@bool/isLightMode</item>
     </style>
-
 </resources>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values-v27/styles.xml
```diff
@@ -14,4 +14,18 @@
...
+        <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
+        <item name="colorAccent">@color/colorAccent</item>
+        <item name="colorSecondary">@color/colorAccent</item>
+        <item name="colorSecondaryVariant">@color/colorAccent</item>
+        <item name="colorButtonNormal">@color/colorPrimary</item>
+        <item name="android:colorBackground">@color/colorBackground</item>
+        <item name="android:navigationBarColor">@color/colorPrimaryDark</item>
+        <item name="alertDialogTheme">@style/Theme.HomeAssistant.Dialog.Alert</item>
+        <item name="actionBarPopupTheme">@style/ThemeOverlay.MaterialComponents.Dark</item>
+        <item name="android:windowLayoutInDisplayCutoutMode">shortEdges</item>
```
jpelgrom, line 29: Is `shortEdges` the right value here (rendering content behind cutouts)? The app doesn't enable rendering behind cutouts for the webview, does it? Even without enabling full screen mode, I now see a difference in landscape mode (there is a cutout in the bottom left):
|Stable app|This PR|
|-----|-----|
|image|image|

jpelgrom (commented):
The new `windowInsetsController` code is a lot nicer!

app/src/main/res/values-v27/styles.xml
```diff
@@ -14,4 +14,18 @@
         <item name="android:windowLightStatusBar">@bool/isLightMode</item>
         <item name="android:windowLightNavigationBar">@bool/isLightMode</item>
     </style>
+
+    <style name="Theme.HomeAssistant" parent="Theme.MaterialComponents.DayNight.NoActionBar">
```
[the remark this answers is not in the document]
dshokouhi: ah this was done by android studio magic, I just clicked the link to make it happy LOL

app/src/main/res/values-v27/styles.xml
dshokouhi: I wasn't sure about this one honestly, I just know that if we don't add this line then the theme does not push up to the top and it looked odd so wanted to fix that :)

What option do you think works best?

jpelgrom: It's a black bar (system hides UI) or white/dark grey bar (system shows UI: window background) in the area for me, so I don't quite follow?

dshokouhi: So if we don't use `shortEdges` or use the new theme line item I end up seeing a black bar where the system bar which looks odd

image

With the new line and set to `shortEdges` it looks like:

image

I am not sure if we can get away with the new landscape behavior when a cut out is used, that seems to be new behavior

jpelgrom: It's black + dark blue vs dark grey + dark blue in dark mode. They're closer but it's not a major difference in my opinion (especially on a phone with lower brightness, where it's barely noticeable). 

The difference is more obvious in light mode: you get a black or white bar + HA blue toolbar. No matter which option you choose, it looks odd. So why change it?

dshokouhi: ok will leave the theme as is then :)

app/src/main/res/values-v27/styles.xml
```diff
@@ -14,4 +14,5 @@
         <item name="android:windowLightStatusBar">@bool/isLightMode</item>
         <item name="android:windowLightNavigationBar">@bool/isLightMode</item>
     </style>
+
```
jpelgrom, line 17: No need for a new empty line ;)

dshokouhi: image

felipecrs:
Just a small feedback. Before:

image

After:

image

None of them are in full-screen. I think the previous one looks slightly better.

dshokouhi:
@felipecrs did you test with the latest commit APK? I ended up reverting the theme changes in the latest commits

https://github.com/home-assistant/android/actions/runs/3048225313

felipecrs:
Oh yeah. The latest one looks fine. Thank you! My last feedback is:

Do you think these CSS attributes have any effect while using the companion app? https://www.w3.org/TR/css-round-display-1/#viewport-fit-descriptor

I'm asking, because I wonder if we can code the Frigate Card in such a way that, when it requests full screen, it will before evaluate whether it's safe or not to draw over the cutout and disable drawing over the cutout if not safe (i.e. when there is a potential button that would be drawn over it).

If they are not effective for the use case I described above, I would suggest that the safest approach for the companion app, when requesting full-screen, would be not to draw over the cutout since it does not know what are the contents that will be drawn on it.

dshokouhi:
> Do you think these CSS attributes have any effect while using the companion app? https://www.w3.org/TR/css-round-display-1/#viewport-fit-descriptor

Anything that works in Chrome for Android should in theory work for the companion app as it uses Android System WebView which is based on Chromium.

The Companion App also offers a setting to enable [Remote Debugging](https://companion.home-assistant.io/docs/integrations/android-webview#remote-debugging) so you can inspect things like easier.

felipecrs:
> Anything that works in Chrome for Android should in theory work for the companion app as it uses Android System WebView which is based on Chromium.

Ok, just to mention, they now behave a little differently:


https://user-images.githubusercontent.com/29582865/190012577-5dea5d19-47f6-4944-8203-0ce4d759c637.mp4

Which for me is completely okay (I prefer that contents are not drawn on top of the notch), but I'm not sure if that was the intended behavior.

jpelgrom:
> ~Do you think these CSS attributes have any effect while using the companion app? https://www.w3.org/TR/css-round-display-1/#viewport-fit-descriptor~

The `@viewport` tag has been [deprecated for some time now and is no longer supported](https://developer.mozilla.org/en-US/docs/Web/CSS/@viewport#browser_compatibility) by the Android WebView. The modern alternative for what you're describing, `safe-area-inset-*` values, should be supported [according to Google, but no one seems to be able to get them to work reliably](https://bugs.chromium.org/p/chromium/issues/detail?id=1094366) (I've also personally tried and failed 😔).

dshokouhi:
> The Android app never displays on top of the notch, while Chrome does. Which for me is completely okay (I prefer that contents are not drawn on top of the notch), but I'm not sure if that was the intended behavior.

I think thats to be expected and ok, the new code now follows Google's own recommendation for immersive mode. I also agree its good not to do anything in the notch area.

jpelgrom (approved):
Tested on Android 13 with a camera cutout in the corner of the display, and on Android 8.1 with no cutouts, and it looks like it is working as expected. 

I don't have any fancy fullscreen interfaces that I can trigger from within the dashboard but user reports suggest that is working well.

felipecrs (approved):

JBassett (approved):

PR #2883, opened 2022-09-15
Add support for using BSSIDs for internal connection

### LINKED ISSUE #2880: Check for BSSID when enabling Internal URL

I have three places where they all share a common SSID/Password, this was done to easily change wifi devices between locations, but then the internal URL only works in one of these locations and requiring me to change settings every time (add/remove SSID) in order for HA app to work properly.

PS: This screen recently got a nice interface, thanks!

Allow me to include a combination of SSID and/or BSSID (AP mac address) to check if it should connect locally or not. This information is already available in the app sensors. The BSSID is different between any access point (even in mesh systems) so that should be pretty much enough.

Maybe a toggle to enable and disable internal URL through HA automations?
**Additional context**

Thanks in advance!

jpelgrom:
## Summary

Add support for entering BSSIDs to use for detecting when the internal connection URL should be used. These are entered and saved like any regular SSID, but contain the prefix "BSSID:". The SSID settings screen will recognize the prefix and apply some formatting to make it easier to see that they've been recognized.

Fixes #2880

## Screenshots

The Home Network WiFi SSID screen with a BSSID alongside a regular SSID. It will remove the "BSSID:" prefix and instead shows the value in a monospace font. Active network detection (blue icon) still works.
|Light|Dark|
|-----|-----|
|The Home Network WiFi SSID screen as described above, light mode.|The Home Network WiFi SSID screen as described above, dark mode.|

Note that the main screen still shows the raw value, I think this is fine.
|Light|Dark|
|-----|-----|
|A settings row with the text 'Home Network WiFi SSID', value 'BSSID:1A:2B:3C:4D:5E:6F, home-assistant-wifi', light mode.|A settings row with the text 'Home Network WiFi SSID', value 'BSSID:1A:2B:3C:4D:5E:6F, home-assistant-wifi', dark mode.|

Documentation: home-assistant/companion.home-assistant#813

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/ssid/SsidFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/ssid/SsidFragment.kt
@@ -34,6 +34,7 @@ class SsidFragment : Fragment() {
                         wifiSsids = viewModel.wifiSsids,
                         prioritizeInternal = viewModel.prioritizeInternal,
                         activeSsid = viewModel.activeSsid,
+                        activeBssid = viewModel.activeBssid,
                         onAddWifiSsid = viewModel::addHomeWifiSsid,
                         onRemoveWifiSsid = viewModel::removeHomeWifiSsid,
                         onSetPrioritize = viewModel::setPrioritize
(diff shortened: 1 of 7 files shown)


jpelgrom:
> Do you think we need to update the description on the Home WiFi SSID page to mention using a BSSID?

Not sure, it is a bit of a pro feature and it would add a lot of additional text to the description. I've added a link to the documentation on this screen which does describe it.

> Not sure if we need to state the obvious as well that a user should remove the SSID if its used in more than 1 location and only use BSSIDs?

I agree it is obivous and don't think it should be added. If someone decides to use a BSSID I assume they know that they want to use a specific AP and will remove the SSID.

dshokouhi (approved):
Tested the debug APK and it works for me!

JBassett (approved):

PR #2888, opened 2022-09-16
Add support for intent scheme URI and deep links in clickAction and actionable notifications

### LINKED ISSUE #2885: Add Ability to add App Links that can be opened.
Possibly, slightly incorrect behaviour from the app, but its also not documented supported behaviour. See below.

Would like to be able to add app links to notifications so that apps can be opened using their prescribed links.

i.e. Opening BG Stats app requires using a url like so:

_bgstats://app.bgstatsapp.com/addPlay.html?gameId=36218_

However at the moment sending this link in a notification and clicking it results in HA opening and then getting a 404 error.
Home Assistant should only open on relative likes with /lovelace/... and then open other links as per the system, or do nothing.

http links opening browser for example.

app::// links do work as documented, but unfortunately cannot send additional parameters with these kind of links.

I'm going to assume here BG Stats for example are doing things correctly, and there should be greater scope for HA to support link types in notifications, but happy to be corrected if wrong.

No alternatives.

**Additional context**
Example of requirements etc. can be found at [BG Stats Support](https://www.bgstatsapp.com/support/push-plays-to-bg-stats-from-other-apps-or-websites/) page.

dshokouhi:
## Summary

Fixes #2885 by allowing users to use an intent scheme URI or an app link so the app can handle these actions after user interaction as opposed to them happening as soon as they are received on the device via `command_activity` and `command_broadcast_intent`.

```
service: notify.mobile_app_pixel_6_pro
data:
  message: test
  data:
    actions:
    - action: URI
      title: test
      uri: "intent://scan/#Intent;scheme=zxing;package=com.google.zxing.client.android;end"
```

```
service: notify.mobile_app_pixel_6_pro
data:
  message: test
  data:
    clickAction: "intent://scan/#Intent;scheme=zxing;package=com.google.zxing.client.android;end"
```

for deep link testing I used our own internal `homeassistant://navigate` as I dont know of any apps I have installed that do deep links 😂 

```
service: notify.mobile_app_pixel_6_pro
data:
  message: test
  data:
    actions:
    - action: URI
      title: test
      uri: "deep-link://example://link/to/content"
```

```
service: notify.mobile_app_pixel_6_pro
data:
  message: test
  data:
    clickAction: "deep-link://example://link/to/content"
```

Also making sure to specify the `package` for `app://` prefix, it still worked beforehand but also does not hurt adding more data into its intent.

Documentation: home-assistant/companion.home-assistant#815

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -102,7 +102,7 @@ class MessagingManager @Inject constructor(
         const val TAG = "MessagingService"
 
         const val APP_PREFIX = "app://"
-        const val APP_LINK_PREFIX = "app-link://"
+        const val DEEP_LINK_PREFIX = "deep-link://"
         const val INTENT_PREFIX = "intent://"
         const val MARKET_PREFIX = "https://play.google.com/store/apps/details?id="
         const val SETTINGS_PREFIX = "settings://"
@@ -1599,11 +1599,11 @@ class MessagingManager @Inject constructor(
                 else
                     WebViewActivity.newInstance(context)
             }
-            UrlHandler.isAbsoluteUrl(uri) || uri.startsWith(APP_LINK_PREFIX) -> {
+            UrlHandler.isAbsoluteUrl(uri) || uri.startsWith(DEEP_LINK_PREFIX) -> {
                 Intent(Intent.ACTION_VIEW).apply {
                     this.data = Uri.parse(
-                        if (uri.startsWith(APP_LINK_PREFIX))
-                            uri.removePrefix(APP_LINK_PREFIX)
+                        if (uri.startsWith(DEEP_LINK_PREFIX))
+                            uri.removePrefix(DEEP_LINK_PREFIX)
                         else
                             uri
                     )
@@ -1616,7 +1616,7 @@ class MessagingManager @Inject constructor(
 
         if (uri.startsWith(SETTINGS_PREFIX) && uri.substringAfter(SETTINGS_PREFIX) == NOTIFICATION_HISTORY)
             intent.putExtra("fragment", NOTIFICATION_HISTORY)
-        if (uri.startsWith(INTENT_PREFIX) || uri.startsWith(APP_PREFIX) || uri.startsWith(APP_LINK_PREFIX))
+        if (uri.startsWith(INTENT_PREFIX) || uri.startsWith(APP_PREFIX) || uri.startsWith(DEEP_LINK_PREFIX))
             intent.`package`?.let {
                 context.packageManager.getLaunchIntentForPackage(
                     it
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -102,6 +102,7 @@ class MessagingManager @Inject constructor(
         const val TAG = "MessagingService"
 
         const val APP_PREFIX = "app://"
+        const val INTENT_PREFIX = "intent://"
         const val MARKET_PREFIX = "https://play.google.com/store/apps/details?id="
         const val SETTINGS_PREFIX = "settings://"
         const val NOTIFICATION_HISTORY = "notification_history"
@@ -1588,6 +1589,9 @@ class MessagingManager @Inject constructor(
             uri.startsWith(APP_PREFIX) -> {
                 context.packageManager.getLaunchIntentForPackage(uri.substringAfter(APP_PREFIX))
             }
+            uri.startsWith(INTENT_PREFIX) -> {
+                Intent.parseUri(uri, Intent.URI_INTENT_SCHEME)
+            }
             uri.startsWith(SETTINGS_PREFIX) -> {
                 if (uri.substringAfter(SETTINGS_PREFIX) == NOTIFICATION_HISTORY)
                     SettingsActivity.newInstance(context)
@@ -1606,6 +1610,13 @@ class MessagingManager @Inject constructor(
 
         if (uri.startsWith(SETTINGS_PREFIX) && uri.substringAfter(SETTINGS_PREFIX) == NOTIFICATION_HISTORY)
             intent.putExtra("fragment", NOTIFICATION_HISTORY)
+        if (uri.startsWith(INTENT_PREFIX) || uri.startsWith(APP_PREFIX))
+            intent.`package`?.let {
+                context.packageManager.getLaunchIntentForPackage(
+                    it
+                )
+            }
+
         intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
         intent.addFlags(Intent.FLAG_ACTIVITY_MULTIPLE_TASK)
         intent.addFlags(Intent.FLAG_ACTIVITY_EXCLUDE_FROM_RECENTS)
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -102,6 +102,7 @@ class MessagingManager @Inject constructor(
         const val TAG = "MessagingService"
 
         const val APP_PREFIX = "app://"
+        const val APP_LINK_PREFIX = "app-link://"
         const val INTENT_PREFIX = "intent://"
         const val MARKET_PREFIX = "https://play.google.com/store/apps/details?id="
         const val SETTINGS_PREFIX = "settings://"
@@ -1598,9 +1599,14 @@ class MessagingManager @Inject constructor(
                 else
                     WebViewActivity.newInstance(context)
             }
-            UrlHandler.isAbsoluteUrl(uri) -> {
+            UrlHandler.isAbsoluteUrl(uri) || uri.startsWith(APP_LINK_PREFIX) -> {
                 Intent(Intent.ACTION_VIEW).apply {
-                    this.data = Uri.parse(uri)
+                    this.data = Uri.parse(
+                        if (uri.startsWith(APP_LINK_PREFIX))
+                            uri.removePrefix(APP_LINK_PREFIX)
+                        else
+                            uri
+                    )
                 }
             }
             else -> {
@@ -1610,7 +1616,7 @@ class MessagingManager @Inject constructor(
 
         if (uri.startsWith(SETTINGS_PREFIX) && uri.substringAfter(SETTINGS_PREFIX) == NOTIFICATION_HISTORY)
             intent.putExtra("fragment", NOTIFICATION_HISTORY)
-        if (uri.startsWith(INTENT_PREFIX) || uri.startsWith(APP_PREFIX))
+        if (uri.startsWith(INTENT_PREFIX) || uri.startsWith(APP_PREFIX) || uri.startsWith(APP_LINK_PREFIX))
             intent.`package`?.let {
                 context.packageManager.getLaunchIntentForPackage(
                     it
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -1631,13 +1631,12 @@ class MessagingManager @Inject constructor(
                         it
                     )
                 }
-                if (intentPackage != null)
-                    intent
-                else {
+                if (intentPackage == null && (!intent.`package`.isNullOrEmpty() || uri.startsWith(APP_PREFIX))) {
                     val marketIntent = Intent(Intent.ACTION_VIEW)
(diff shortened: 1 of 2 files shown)


jpelgrom:
> I don't really see a way for us to validate these custom protocols that are defined by the app themselves, android does not have a native way to detect this

Wouldn't checking if the intent resolves be enough?

```kotlin
val valid = intent.resolveActivity(context.packageManager) != null
```

dshokouhi:
> > I don't really see a way for us to validate these custom protocols that are defined by the app themselves, android does not have a native way to detect this
> 
> Wouldn't checking if the intent resolves be enough?
> 
> ```kotlin
> val valid = intent.resolveActivity(context.packageManager) != null
> ```

hmm I am not really sure how that would fit in with existing logic because we default the intent to the app when its null. We already do a protocol check for `http` and `https` earlier as well

https://github.com/home-assistant/android/blob/7a41d44872343f0f16badb6c1d42a5d936b6add6/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt#L1585-L1609

dshokouhi:
I suppose we could modify:

https://github.com/home-assistant/android/blob/7a41d44872343f0f16badb6c1d42a5d936b6add6/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt#L1601

to be `uri.contains("://")` a bit hacky but all the deep links and URLs support `ACTION_VIEW`

we already default to the webview activity and any path set I doubt will have `://` in it 😂

jpelgrom:
I hadn't thought about how to implement it yet :) Now that I think about it this probably isn't going to work correctly anyway because it won't work for broadcasts (= no activity).

dshokouhi:
ok well at least with this current code we could work around the issue temporarily instead of not working at all 😂

dshokouhi:
I am going to add another prefix for deep links specifically since there is no native way to detect things, this will actually solve the request :)

jpelgrom:
Is 'App links' the correct name for what it is supporting? The documentation describes them as "leverag[ing] HTTP URLs and association with a website", which was already possible, and your example is specifically about using a custom scheme.

dshokouhi:
I felt like App Link sounds more user friendly but I can definitely update it to deep links to be more precise 😃 will update the docs link to be more specific to: https://developer.android.com/training/app-links#deep-links

jpelgrom:
Technically the documentation describes "_Android_ App Links" so you're not wrong but it is somewhat confusing as a developer...  Agreed we should focus on the user though, so updating the documentation link to be more specific is a good solution.

dshokouhi:
LOL I committed too fast 🤦‍♂️

dshokouhi:
Lets just stick with deep links as its more precise, because a normal URL will still be handled the same.  Honestly I suck at naming things and am often on the fence with what I ultimately decided 😂

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -102,6 +102,8 @@ class MessagingManager @Inject constructor(
         const val TAG = "MessagingService"
 
         const val APP_PREFIX = "app://"
+        const val DEEP_LINK_PREFIX = "deep-link://"
+        const val INTENT_PREFIX = "intent://"
```
jpelgrom, line 106: This is incorrect (or at least, it is used incorrectly): only the `intent:` part is always part of a intent URI, `//` is not required. 

Example: the output of the following Intent `toUri` from the debug app:
```kotlin
Intent(context, NfcSetupActivity::class.java).toUri(Intent.URI_INTENT_SCHEME)
```
is:
```
intent:#Intent;component=io.homeassistant.companion.android.debug/io.homeassistant.companion.android.nfc.NfcSetupActivity;end
```

dshokouhi: This is the same format that we use for the WebView activity.

We can even see the same format used in the anchor link example in the chrome docs.

jpelgrom: The Chrome docs also show that only `intent:` is always included, the `//` in their example is from the next line "HOST/URI-path" which is marked optional

dshokouhi: Well I would hate to introduce a breaking change and/or have 2 different methods used in each place. The code still works FWIW.

jpelgrom: But it doesn't currently work? Taking the (valid) intent URI from my previous comment I'm sending a notification to the app on commit 29f61e770060df255d42a30d310e5890628d9d78:

```yaml
service: notify.mobile_app_pixeldev
data:
  message: test
  data:
    actions:
      - action: URI
        title: test
        uri: >-
          intent:#Intent;component=io.homeassistant.companion.android.debug/io.homeassistant.companion.android.nfc.NfcSetupActivity;end
```

Tapping this action doesn't open the NFC activity as expected but instead it opens the webview (`else` branch, and it tries to append it to my instance's URL? `http://instance/intent:?external_auth=1#Intent;component=io.homeassistant.companion.android.debug/io.homeassistant.companion.android.nfc.NfcSetupActivity;end`). If I change `INTENT_PREFIX` to only `intent:`, it works as expected.

> Well I would hate to introduce a breaking change

Intent URI support is new in this PR, so no breaking change yet?

dshokouhi: > But it doesn't currently work?

Try the URI example I used up above in the PR description, it does indeed launch the scanner app to scan something.


> > Well I would hate to introduce a breaking change
> 
> Intent URI support is new in this PR, so no breaking change yet?

Its only new for notifications but not in webview.

We currently already support `intent://` in [webview links](https://companion.home-assistant.io/docs/integrations/android-webview#links) so for consistency I would like them to remain the same so a user can use the same code in both places.

jpelgrom: Guess I found a bug, the same intent link doesn't work in the webview :)

To be clear, I don't think this would be a breaking change as it's checking for the start of the link and `intent://` and `intent:` both start with `intent:`.

dshokouhi: I suppose we could just drop the `//` from the prefix and it should still work for all cases.  For the example purpose I will keep it alone so a user can see a working example.  

Thanks for clearing that up 😅 I was stuck on what we did previously as the correct way since we had it for so long.

dshokouhi: Good catch, thanks for sticking with me while I was stuck in my old ways

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -1606,6 +1616,13 @@ class MessagingManager @Inject constructor(
 
         if (uri.startsWith(SETTINGS_PREFIX) && uri.substringAfter(SETTINGS_PREFIX) == NOTIFICATION_HISTORY)
             intent.putExtra("fragment", NOTIFICATION_HISTORY)
+        if (uri.startsWith(INTENT_PREFIX) || uri.startsWith(APP_PREFIX) || uri.startsWith(DEEP_LINK_PREFIX))
+            intent.`package`?.let {
+                context.packageManager.getLaunchIntentForPackage(
+                    it
+                )
+            }
```
jpelgrom, line 1624: Is this required? You don't seem to be using the output and intent, app and deep links seem to work fine without them.

dshokouhi: This was carried over from WebViewActivity, I can't remember why I had used it previously but I dont think it hurts to set the package explicitly if we can?

dshokouhi: Looking at the docs for [setPackage()](https://developer.android.com/reference/android/content/Intent#setPackage(java.lang.String)) I see now that it was added in cases where it might be needed.

```
(Usually optional) Set an explicit application package name that limits the components this Intent will resolve to.
```

"(Usually optional)" 😂

jpelgrom: I don't see where you're setting it on the intent though - or are the backticks similar to `=` somehow?

dshokouhi: Oh I see now, yes you are right its not used here 🤦‍♂️ will remove

dshokouhi: Actually now I remember, this was used in `WebViewActivity` to detect when an app did not exist on the device to load the play store so the user can install the app and consume the link. Will add it back properly. Thanks for spotting this one too. 😓

jpelgrom:
The latest changes break deep links for me, whereas previously it worked correctly. 

<details>
<summary>Example using Google Clock app</summary>

```yaml
service: notify.mobile_app_pixeldev
data:
  message: test
  data:
    actions:
      - action: URI
        title: test
        uri: deep-link://clock-app://com.google.android.deskclock/stopwatch
```
</details>

<details>
<summary>Example using Home Assistant app</summary>

```yaml
service: notify.mobile_app_pixeldev
data:
  message: test
  data:
    actions:
      - action: URI
        title: test
        uri: deep-link://homeassistant://navigate/lovelace/0
```
</details>

dshokouhi:
So it's taking you to the market for deep links? Or nowhere?

jpelgrom:
> So it's taking you to the market for deep links? Or nowhere?

It opened the Play Store with a screen "Something went wrong / Try again".

Unfortunately it looks like the package check still marks the intent URI I previously generated as invalid, probably because it doesn't specify the `package` item, and it tries to open the Play Store: `intent:#Intent;component=io.homeassistant.companion.android.debug/io.homeassistant.companion.android.nfc.NfcSetupActivity;end`.

dshokouhi:
> > So it's taking you to the market for deep links? Or nowhere?
> 
> It opened the Play Store with a screen "Something went wrong / Try again".
> 
> Unfortunately it looks like the package check still marks the intent URI I previously generated as invalid, probably because it doesn't specify the `package` item, and it tries to open the Play Store: `intent:#Intent;component=io.homeassistant.companion.android.debug/io.homeassistant.companion.android.nfc.NfcSetupActivity;end`.

I imagine you see the same behavior in WebViewActivity? I wouldn't want to get rid of the code personally because its nice to direct the user to installing the item rather than doing nothing which is what happens when the app is not installed. Tough to decide whats correct or not.

jpelgrom:
> I imagine you see the same behavior in WebViewActivity?

The markdown card seems to filter the links out so I can't test it easily 😅 but I assume it'll be the same

> its nice to direct the user to installing the item rather than doing nothing which is what happens when the app is not installed

I agree, but right now it opens the Play Store with an error message because the URL is missing the package name. Maybe only send the user to the Play Store if the app is not installed _and_ the package name is known?

dshokouhi:
I don't think we can detect if the package name is known when the launch intent can't resolve properly? The only reason it works for `app://` is because the user explicitly gave it to us.  So it sounds like we either silently do nothing or load the play store when the intent URI did not include the package.

dshokouhi:
> Can't we simply check for `intent.package != null`?

That check happened [earlier](https://github.com/dshokouhi/home-assistant-android/blob/action_intent_scheme/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt#L1634) I thought but maybe not used properly? 

just now I found we can use `context.packageManager.queryIntentActivities(intent, 0)[0].activityInfo.packageName` which does resolve the package name for us but gives us a list so do you think its safe to go by the first package in the list? Obviously it won't work with your example because the debug package is not in the app store but it will prevent a null value.

jpelgrom:
> That check happened [earlier](https://github.com/dshokouhi/home-assistant-android/blob/action_intent_scheme/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt#L1634) I thought but maybe not used properly?

It can return `null` if

1. there is no package set (because of `?.let`)
2. there is a package set, but no launch intent can be found (`context.packageManager.getLaunchIntentForPackage` returns `null`)

We should probably only open the Play Store in situation 2. My example is `null` because of situation 1, in which the package isn't known.

> just now I found we can use `context.packageManager.queryIntentActivities(intent, 0)[0].activityInfo.packageName` which does resolve the package name for us but gives us a list so do you think its safe to go by the first package in the list?

It will work for my example but if the user is trying to send an intent that doesn't open an activity it won't work (while it still may be valid)

dshokouhi:
ok think I got it this time 🤞

jpelgrom (approved):
You got it :)

Thanks for looking at all the weird edge cases I found

dshokouhi:
> Thanks for looking at all the weird edge cases I found

As always I appreciate the time you take to look at the PRs and to find these edge cases and old bugs :)

JBassett (approved):

Issue #2816, opened 2022-08-24
screen brightness control


gmcaesar:
**Is your feature request related to a problem? Please describe.**

I can control so much of the phone through notification but not the brightness

**Describe the solution you'd like**

command notification to turn on and off auto brightness and one to set the brightness %

**Describe alternatives you've considered, if any**

**Additional context**

dshokouhi:
I believe the permission required here is a protected permission that the app may not qualify for.

gmcaesar:
could it not be a permissions request? i know I can do it with other apps, but i'd rather not send mqtt to automate and send control notifications

gmcaesar:
opps accidently closed.... but could this not be added with a permission request? i know I can give admin permissions to apps, I just would rather not have to send mqtt to automate and control notifications to the companion when one app could do it all

dshokouhi:
this is what I meant

![image](https://user-images.githubusercontent.com/1634145/189234111-7aab8abd-994c-4e7d-a465-7cac79debd3d.png)

We may not qualify for it given that we do not qualify for things related to the phone call log and SMS sending etc...

We can keep this open for now but it is indeed a protected permission.

marazmarci:
I think this functionality can be hacked together with Tasker 🙂
You can send an Intent with the `command_broadcast_intent` notification command (with the desired brightness included as an extra), and catch it in Tasker using its `Intent Received` event handler, then use the `Display Brightness` action.

dshokouhi:
> I think this functionality can be hacked together with Tasker 🙂 You can send an Intent with the `command_broadcast_intent` notification command (with the desired brightness included as an extra), and catch it in Tasker using its `Intent Received` event handler, then use the `Display Brightness` action.

Yea this would be a valid workaround if we don't qualify for the permission, its still worth it to give it a shot I think. IMO our app should qualify for all permissions that tasker qualifies for but unfortunately that was not the case when we tried to add the last caller sensor.

PR #2894, opened 2022-09-19
Add commands to control screen brightness and a sensor to get its value

dshokouhi:
## Summary

Fixes: #2816 by adding 2 new commands.

-  `command_auto_screen_brightness` which accepts the parameter `command` for either `turn_on` or `turn_off` values to control if automatic screen brightness is on or not
-  Also adds `command_screen_brightness_level` which accepts the parameter `command` that needs to be an integer between `0` and `255`, if a non-integer value is sent then we will post the notification otherwise the app will default to either `0` or `255` whichever is closer.  
- The app will prompt the user to enable the Write Settings permission if not granted in order for the commands to work. 
- The notification will also post as normal if there was an error writing the value to the DB per the [docs](https://developer.android.com/reference/android/provider/Settings.System#putInt(android.content.ContentResolver,%20java.lang.String,%20int)).
- A sensor has also been added to get the current screen brightness setting as well as if automatic is enabled.

```
service: notify.mobile_app_pixel_6_pro
data:
  message: command_screen_brightness_level
  data:
    command: 150
```

```
service: notify.mobile_app_pixel_6_pro
data:
  message: command_auto_screen_brightness
  data:
    command: turn_on
```

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#824

## Any other notes

FCM: https://github.com/home-assistant/mobile-apps-fcm-push/pull/93

Testing did seem a bit odd at times where values despite being correct resulted in some inconsistencies which I suspect is due to automatic being enabled. For auto brightness, although my pixel did not turn off "adaptive brightness" I was able to replicate the dimming of the display when enabled and can confirm the screen did not dim when I turned it off.  It is consistent enough for users to use.


app/src/main/java/io/homeassistant/companion/android/sensors/SettingsSensorManager.kt
```diff
@@ -0,0 +1,73 @@
...
+class SettingsSensorManager : SensorManager {
+    companion object {
+        private const val TAG = "ScreenBrightness"
+
+        val screenBrightness = SensorManager.BasicSensor(
+            "screen_brightness",
+            "sensor",
+            commonR.string.basic_sensor_name_screen_brightness,
+            commonR.string.sensor_description_screen_brightness,
+        )
```
jpelgrom, line 18: The sensor should have a stateless icon so something is shown before enabling it

app/src/main/java/io/homeassistant/companion/android/sensors/SettingsSensorManager.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    override fun docsLink(): String {
+        return "https://companion.home-assistant.io/docs/core/sensors#screen-brightness-sensor"
+    }
+
+    override val enabledByDefault: Boolean
+        get() = false
+    override val name: Int
+        get() = commonR.string.sensor_name_device_settings
```
jpelgrom, line 28: While I understand this name from a technical poin of view, it doesn't make any sense to me as a normal user. Aren't most of the sensors showing the state of a setting and as a result "Device Settings Sensors"?

Because there isn't a lot of logic to 'support' the `Settings.System` API, isn't it better to call this one something like "Display Sensors" instead?

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -1996,6 +2034,31 @@ class MessagingManager @Inject constructor(
...
+            if (data[MESSAGE].toString() == COMMAND_SCREEN_BRIGHTNESS_LEVEL)
+                Settings.System.SCREEN_BRIGHTNESS
+            else
+                Settings.System.SCREEN_BRIGHTNESS_MODE,
+            if (data[MESSAGE].toString() == COMMAND_SCREEN_BRIGHTNESS_LEVEL)
+                when {
+                    command!!.toInt() < 0 -> 0
+                    command.toInt() > 255 -> 255
+                    else -> command.toInt()
+                }
```
jpelgrom, line 2051: Kotlin has a nice function for this: [`coerceIn`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/coerce-in.html)

```suggestion
                command!!.toInt().coerceIn(0..255)
```

app/src/main/java/io/homeassistant/companion/android/sensors/SettingsSensorManager.kt
```diff
@@ -0,0 +1,73 @@
...
+                Settings.System.getInt(context.contentResolver, Settings.System.SCREEN_BRIGHTNESS)
+            auto = Settings.System.getInt(
+                context.contentResolver,
+                Settings.System.SCREEN_BRIGHTNESS_MODE
+            ) == Settings.System.SCREEN_BRIGHTNESS_MODE_AUTOMATIC
+        } catch (e: Exception) {
+            Log.e(TAG, "Unable to get screen brightness settings", e)
+        }
+
+        val icon = if (auto) "mdi:brightness-auto" else "mdi:brightness-5"
```
jpelgrom, line 62: Should we maybe use `mdi:brightness-6` when auto brightness is off to match the system settings app?

jpelgrom:
With my display at 46% brightness, auto brightness off, I'm getting 19 as a state for the sensor... Either it isn't working correctly or this isn't a linear scale as expected.

dshokouhi:
> With my display at 46% brightness, auto brightness off, I'm getting 19 as a state for the sensor... Either it isn't working correctly or this isn't a linear scale as expected.

is that with adaptive display on? that seems to do additional screen brightness things from what I noticed

app/src/main/java/io/homeassistant/companion/android/sensors/SettingsSensorManager.kt
dshokouhi: My thoughts was that we may be showing more values from `Settings.*` in here, I can change it if we feel the only things of value are screen related settings? In reality we are just reading the raw values of the pre-defined constants, whatever they may be.

jpelgrom:
I believe adaptive brightness = auto brightness, or at least the automatic attribute changes when I turn adaptive brightness off in the system settings.

dshokouhi:
> > is that with adaptive display on? that seems to do additional screen brightness things from what I noticed
> 
> I believe adaptive brightness = auto brightness, or at least the automatic attribute changes when I turn adaptive brightness off in the system settings.

Yea I saw that too but when I ran the commands I did not see the toggle change so I assume it does more. The brightness slider does update accordingly most of the time. This one is kinda funky due to its reliance on `Settings.System` it seems

app/src/main/java/io/homeassistant/companion/android/sensors/SettingsSensorManager.kt
jpelgrom: But will users understand that it is from `Settings.*`? That shouldn't be required in my opinion and as a result the name isn't good. 

(We can get vibration or WiFi or font scaling info from the same API but if the app were to all add it here it wouldn't be a logical group for non-tech people.)

dshokouhi: Thats a good point, will update to "Display Sensors" since thats more correct here

JBassett (approved):

thatso:
To give an idea of some values for those coming late to the party: Android 12 on a Poco F3 with adaptive brightness off and the slider setting on the lower side at about 25% out of 100% gave a `screen_brightness` sensor reading of 175. After a recent upgrade to Android 13 with the same slider setting the sensor now reads a value as low as 6. :astonished:

PR #2899, opened 2022-09-20
Breaking Change: Only send notification received event if a notification wants the confirmation

dshokouhi:
## Summary

There has been a large increase in data traffic from the app to HA since the last release. This PR now requires a new notification parameter to be set so the app can send the received event. Users will need to set `confirmation: true` for the notification they wish to get confirmation the notification was received. This will allow us to cut down on traffic and let the user specify which notifications they actually want the event for.

example:

```
message: Test
data:
  confirmation: true
```

Documentation: home-assistant/companion.home-assistant#818

## Any other notes

FCM: https://github.com/home-assistant/mobile-apps-fcm-push/pull/92


JBassett (changes requested):
Can you also add this to the changelog so we can get this in Beta ASAP?

JBassett (approved):

PR #2901, opened 2022-09-21
Allow custom intents register for categories in case they are required

dshokouhi:
## Summary

I noticed in Android 13 that there is a behavior change in [intents](https://developer.android.com/about/versions/13/behavior-changes-all#intents) which in some cases will cause the app not to receive an intent.

This PR extends the intent setting value to contain categories that an intent could have. As an intent can have multiple categories I have added in support for this as well.

The new format will be `intent,category1,category2` to allow for the app to continue working as is and to let users modify the setting to include categories if required. Once the app targets SDK 33 I suspect we may be more susceptible to this issue but as of now we can reproduce this behavior using Tasker.

Documentation: home-assistant/companion.home-assistant#821


common/src/main/res/values/strings.xml
```diff
@@ -493,7 +493,7 @@
...
+    <string name="sensor_description_last_update">The intent for the last update that was sent, periodic updates will show as \"SensorWorker\". Enabling the \"Add New Intent\" toggle will create 1 setting to allow you to register for a broadcast intent. The toggle will switch back to off once a new setting is created so you will need to turn it back on to save more intents. You can also clear out the setting value to remove the setting in the next update. If you are on Android 13+ and you notice an intent longer works then you will need to add categories that the intent expects. To do this you will need to add each category after the intent with a \",\" repeating until there are no more categories. For example an intent with 2 categories will be: \"intent,category1,category2\" as the setting value. You must restart the application after making changes to these settings to take effect.</string>
```
jpelgrom, line 496: > If you are on Android 13+ and you notice an intent longer works

You forgot "no". However, this should be written for both first time users and upgrading users and will also need to account for new intents. I'm thinking something like "When you're using Android 13+ and you're not receiving all intents..."?

> then you will need to add categories that the intent expects.

The intent doesn't expect categories, but includes/sends them. The intent receiver expects categories (but let's not make it more complicated by introducing intent receivers here).

> add each category after the intent 

after the intent action

> with a \",\" repeating until there are no more categories.

Easier: "seperated by a ','."? Or would that mean something different?

jpelgrom (commented):
Is a comma a safe separator to use that won't conflict with values? I know most apps use dots and underscores in actions but we wouldn't want to create an accidental breaking change.

dshokouhi:
> Is a comma a safe separator to use that won't conflict with values? I know most apps use dots and underscores in actions but we wouldn't want to create an accidental breaking change.

It's hard to tell what's safe or not here honestly. In terms of what's documented on Android all examples include dots and underscores so it felt safe. It's also what's used for intent extras in our notification command.

common/src/main/res/values/strings.xml
dshokouhi: > You forgot "no". However, this should be written for both first time users and upgrading users and will also need to account for new intents. I'm thinking something like "When you're using Android 13+ and you're not receiving all intents..."?

good catch, will update




> after the intent action

I actually didn't specify `action` here because it was not used earlier. I will update the description to make it more explicit.



> Easier: "seperated by a ','."? Or would that mean something different?

👍 no I think thats good. The example immediately after this should clear up any confusion......i hope 😃

jpelgrom: >> then you will need to add categories that the intent expects.
>
> The intent doesn't expect categories, but includes/sends them. The intent receiver expects categories (but let's not make it more complicated by introducing intent receivers here).

Looks like you missed this comment, or didn't share why you don't agree.

dshokouhi: Oh I didn't think it was meant to be applied since you mentioned not to make it more complicated :)  

For our purposes I thought the expectation would be enough to tell a user that its required in order to get the intent registered.

jpelgrom: I understand the confusion now, let's keep it as it is :)

jpelgrom:
Testing this and it's not working like expected. I have registered "ha_test,android.intent.category.DEFAULT" in the app, yet intents with only the action ha_test and no category set are still received. Am I misunderstanding something here?

dshokouhi:
thats to be expected actually....you can register for more categories than the intent calls for and it won't make a difference....its only when the intent contains a category does it need to match

jpelgrom:
Right I didn't understand the changes correctly. Did a lot more testing and reading and I think it's actually a lot less complicated than we think.

The Android 13 change only applies to explicit intents; intents sent _directly_ to the app (package = io.homeassistant.companion.android(.debug), class = io.homeassistant.companion.android.sensors.SensorReceiver). On older Android versions/when not targeting Android 13 any intents sent directly to the app will be received even if it doesn't match the intent filter. The upcoming Android change is for these intents, now there will have to be a filter for them. However there is no need to send intents directly to the HA app, the last update sensor allows registering any intent.

Most intents users register probably aren't sent directly to the HA app so they aren't impacted by the change, they are implicit intents for which these requirements already existed.

This PR is more like a new feature 🙂. Implicit intents can only be received when registered correctly. Previously the intent filter would only accept an action, and if the sender included a category it wouldn't be handled by the app (tested on Android 11). Now you can specify the category and receive these intents as well.

tl;dr This PR works and is more of a new feature. Only apps specifically sending an intent to HA will be impacted by the Android 13 change.

---

Two more notes about the description:
 - as explained above the category filtering doesn't apply exclusively to Android 13, so maybe it can be more general?
 - on a phone screen it is a very large block of text, maybe put the info about categories in a new paragraph (`\n\n`) like in the docs?

dshokouhi:
Thank you for doing that extra research into what actually is going on. That makes complete sense as to why we didn't hear any complaints after the update. I'll update the description to make to clean it up as well :)

jpelgrom (approved):

jpelgrom:
Don't forget to update the docs PR as well :)

dshokouhi:
> Don't forget to update the docs PR as well :)

ok updated :) thank you for calling that out, I would've forgotten 😂

JBassett (approved):

PR #2829, opened 2022-08-27
Support websocket message coalescing

jpelgrom:
## Summary

Implement support for home-assistant/core#77238. Core will now combine multiple responses in 1 single websocket message if it's faster. Translation: it might return a JSON array of objects instead of a single object.

The logging for websocket has also been adjusted to handle these changes. To keep the logs readable and better match 'normal' network requests, it will no longer print the entire body on non-debug builds.

Example output (debug):

```
2022-08-27 18:10:52.323 28760-28760/io.homeassistant.companion.android.debug D/WebSocketRepository: Sending message 1: {type=supported_features, id=1, features={coalesce_messages=1}}
2022-08-27 18:10:52.332 28760-28760/io.homeassistant.companion.android.debug D/WebSocketRepository: Sending message 2: {type=config/device_registry/list, id=2}
2022-08-27 18:10:52.332 28760-28760/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number 2 sent
2022-08-27 18:10:52.334 28760-28760/io.homeassistant.companion.android.debug D/WebSocketRepository: Sending message 3: {type=config/entity_registry/list, id=3}
2022-08-27 18:10:52.335 28760-28760/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number 3 sent
2022-08-27 18:10:52.337 28760-28760/io.homeassistant.companion.android.debug D/WebSocketRepository: Sending message 4: {type=get_states, id=4}
2022-08-27 18:10:52.337 28760-28760/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number 4 sent
2022-08-27 18:10:52.338 28760-28760/io.homeassistant.companion.android.debug D/WebSocketRepository: Sending message 5: {type=config/area_registry/list, id=5}
2022-08-27 18:10:52.339 28760-28760/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number 5 sent
2022-08-27 18:10:52.488 28760-28796/io.homeassistant.companion.android.debug D/WebSocketRepository: Websocket: onMessage (text: [{"id":1,"type":"result","success":true,"result":null},{"id":2,"type":"result","success":true,"result":[...]},{"id":3,"type":"result","success":true,"result":[...
2022-08-27 18:10:52.598 28760-28796/io.homeassistant.companion.android.debug D/WebSocketRepository: Message number 1 received
[...]

Issue #2889, opened 2022-09-16
Sentry error: Null Pointer Exception for subscribeTo state changes


dshokouhi:
**Home Assistant Android version:**

beta-2692

**Android version:**

12

**Phone model:**

SM-F936U1

**Home Assistant version:**

unknown

**Last working Home Assistant release (if known):**

unknown

**Description of problem:**

Found the following sentry error that started to show up in the latest beta in the play store, not impacting too many users but more than just a few.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```
java.lang.NullPointerException: null
    at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.subscribeTo(WebSocketRepositoryImpl.kt:256)
    at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.subscribeToTrigger(WebSocketRepositoryImpl.kt:203)
    at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.getStateChanges(WebSocketRepositoryImpl.kt:182)
    at io.homeassistant.companion.android.common.data.integration.impl.IntegrationRepositoryImpl.getEntityUpdates(IntegrationRepositoryImpl.kt:615)
    at io.homeassistant.companion.android.widgets.BaseWidgetProvider$onScreenOn$1.invokeSuspend(BaseWidgetProvider.kt:108)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
    at kotlinx.coroutines.internal.ScopeCoroutine.afterResume(Scopes.kt:33)
    at kotlinx.coroutines.AbstractCoroutine.resumeWith(AbstractCoroutine.kt:102)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:46)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
    at android.os.Handler.handleCallback(Handler.java:938)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at android.os.Looper.loopOnce(Looper.java:226)
    at android.os.Looper.loop(Looper.java:313)
    at android.app.ActivityThread.main(ActivityThread.java:8855)
    at java.lang.reflect.Method.invoke(Method.java)
[...]

PR #2902, opened 2022-09-21
Fix websocket subscription race/dropping first message

jpelgrom:
## Summary

This PR fixes a potential race condition for subscriptions on the websocket, where the first message might be dropped if it is delivered very soon after the subscription confirmation because the `eventSubscriptionFlow`/`eventSubscriptionProducerScope` haven't been set up yet. While not new, it is more likely to happen after #2829 because the server can send messages faster.

As this issue also applied to the notification flow, **I've merged the notification subscription logic with the more general subscription logic in this PR** so it can also receive the fix without duplicating code.

Inspired by [home-assistant-js-websocket](https://github.com/home-assistant/home-assistant-js-websocket), the app will now set up the response + subscription handlers immediately after _sending_ the initial message. To make sure no messages are dropped, it'll also use a `Channel` to buffer event messages while not yet ready to receive them. The `Channel` will be collected inside the existing `callbackFlow` that is only active while there are subscribers.

If the app receives events for subscriptions that aren't tracked, it will now send an unsubscribe message.

## Screenshots

n/a

n/a

## Any other notes

The easiest option to reproduce the issue is by introducing a delay [here](https://github.com/home-assistant/android/blob/7ab914a8bb47ad84a62c435d19f2b26fb1171064/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt#L232), for example 10 seconds. Then, do something that will generate an event after subscribing and note that it isn't handled by the app.

You can easily see this in the Wear app by triggering a state change after opening the app. In the stable version of the app the state change won't be reflected, with this PR it will be after the delay.

Re: https://github.com/home-assistant/android/blob/1449ff540938e8823eba6c9b26d86eb8ab20c4b3/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt#L287

Yes, it turns out it works the same as unsubscribing from any other subscription :)

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/WebSocketRequest.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/WebSocketRequest.kt
@@ -10,6 +10,8 @@ import kotlinx.coroutines.flow.SharedFlow
  * received, or sent for a subscription) on the websocket connection.
  * @param message Map that holds the websocket message contents
  * @param eventFlow Flow (using callbackFlow) that will emit events for a subscription, else `null`
+ * @param eventTimeout timeout in milliseconds for ending the subscription when the flow is no
+ * longer collected
  * @param onEvent Channel that can receive events for a subscription, else `null`
  * @param onResponse Continuation for the initial response to this message. Don't set this when
  * creating this class, it will be set when a message is sent on the websocket.
@@ -17,6 +19,7 @@ import kotlinx.coroutines.flow.SharedFlow
 data class WebSocketRequest(
     val message: Map<*, *>,
     val eventFlow: SharedFlow<Any>? = null,
+    val eventTimeout: Long = 0L,
     val onEvent: Channel<Any>? = null,
     var onResponse: CancellableContinuation<SocketResponse>? = null
 )
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/WebSocketRequest.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/WebSocketRequest.kt
@@ -1,0 +1,22 @@
+package io.homeassistant.companion.android.common.data.websocket
+
+import io.homeassistant.companion.android.common.data.websocket.impl.entities.SocketResponse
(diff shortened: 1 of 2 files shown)


dshokouhi:
I was looking on sentry and noticed an error which I think may actually be fixed by this PR. If you think its separate I can make a new issue but for now here is the stacktrace from beta 2732 (note: according to sentry this is still the same error from https://github.com/home-assistant/android/issues/2889 just a different stacktrace based on the events from the release tag)

```
java.lang.NullPointerException: null cannot be cast to non-null type kotlinx.coroutines.flow.Flow<T of io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.subscribeTo>
    at java.util.Objects.requireNonNull(Objects.java:245)
    at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.subscribeTo(WebSocketRepositoryImpl.kt:262)
    at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.subscribeToTrigger(WebSocketRepositoryImpl.kt:203)
    at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.getStateChanges(WebSocketRepositoryImpl.kt:182)
    at io.homeassistant.companion.android.common.data.integration.impl.IntegrationRepositoryImpl.getEntityUpdates(IntegrationRepositoryImpl.kt:615)
    at io.homeassistant.companion.android.widgets.BaseWidgetProvider$onScreenOn$1.invokeSuspend(BaseWidgetProvider.kt:108)
    at android.os.Handler.handleCallback(Handler.java:938)
    at android.os.Looper.loopOnce(Looper.java:226)
    at android.os.Looper.loop(Looper.java:313)
    at android.app.ActivityThread.main(ActivityThread.java:8669)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1135)
```

JBassett (approved):

Issue #2908, opened 2022-09-23
Uniquely identifying (i)Beacons: Beacon Monitor vs HA Core implementation for iBeacons


Mincka:
**Is your feature request related to a problem? Please describe.**
The current implementation of the Beacon Monitor only shows the UUID as an attribute. As per the upcoming **i**Beacon [integration](https://github.com/home-assistant/home-assistant.io/pull/24178) and its [documentation](https://github.com/home-assistant/home-assistant.io/blob/next/source/_integrations/ibeacon.markdown) by @dbraco, he suggests the following setup for iBeacons:

> iBeacon Devices are tracked by a combination of the following data:
> 
> UUID (universally unique identifier) is a 128-bit identifier that is generally set the same for all iBeacons at the same physical location.
> Major is an integer to differentiate between iBeacons with the same UUID.
> Minor is an integer to differentiate between iBeacons with the same UUID and Major value.
> MAC address (except for devices with a randomized MAC address)
> Consider setting up your iBeacons with a schema similar to the following:
> 
> uuid=UUID major=1000 minor=1000 Downstairs Front Room
> uuid=UUID major=1000 minor=1001 Downstairs Bathroom
> uuid=UUID major=2000 minor=1001 Upstairs Main Bedroom
> uuid=UUID major=2000 minor=1002 Upstairs Guest Bedroom
> uuid=UUID major=3000 minor=1000 Attic

The same logic was also presented in the code of the iBeacon [lib](https://github.com/AltBeacon/android-beacon-library/blob/957ffd4ff9c214f6996135a36e03ed3c1487172c/src/com/radiusnetworks/ibeacon/IBeacon.java):
> An iBeacon is identified by a three part identifier based on the fields
> proximityUUID - a string UUID typically identifying the owner of a number of ibeacons
> major - a 16 bit integer indicating a group of iBeacons
> minor - a 16 bit integer identifying a single iBeacon

but was later changed to something more general:
> A Beacon is identified by a unique multi-part identifier, with the first of the ordered identifiers being more significant for the purposes of grouping beacons.

So we can see that the general idea is to use major / minor to differentiate iBeacons.

This means that in this case the Beacon Monitor will not be able to differentiate (i)Beacons with the same UUID since the Major / Minor information is not captured by the Beacon Monitor, only the `id1` of a beacon (which is the UUID of the iBeacon).

[...]

PR #2909, opened 2022-09-23
Breaking Change: Add major and minor to UUID for beacon attribute in beacon monitor

dshokouhi:
## Summary

Fixes: #2908 by adding the `major` and `minor` for the beacon.

Listing as a breaking change since the attribute is changing

## Any other notes

Tested using the BLE transmitter on other devices to make sure it continues to work and supply valid data.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
@@ -9,6 +9,7 @@ import io.homeassistant.companion.android.bluetooth.ble.IBeaconTransmitter
 import io.homeassistant.companion.android.bluetooth.ble.KalmanFilter
 import io.homeassistant.companion.android.bluetooth.ble.MonitoringManager
 import io.homeassistant.companion.android.bluetooth.ble.TransmitterManager
+import io.homeassistant.companion.android.bluetooth.ble.name
 import io.homeassistant.companion.android.common.bluetooth.BluetoothDevice
 import io.homeassistant.companion.android.common.bluetooth.BluetoothUtils
 import io.homeassistant.companion.android.common.bluetooth.BluetoothUtils.supportsTransmitter
@@ -336,7 +337,7 @@ class BluetoothSensorManager : SensorManager {
             if (state != "") state else lastState,
             icon,
             mapOf(
-                "id" to bleTransmitterDevice.uuid + "-" + bleTransmitterDevice.major + "-" + bleTransmitterDevice.minor,
+                "id" to name(bleTransmitterDevice.uuid, bleTransmitterDevice.major, bleTransmitterDevice.minor),
                 "Transmitting power" to bleTransmitterDevice.transmitPowerSetting,
                 "Advertise mode" to bleTransmitterDevice.advertiseModeSetting,
                 "Measured power" to bleTransmitterDevice.measuredPowerSetting,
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
```diff
@@ -353,27 +352,22 @@ class BluetoothSensorManager : SensorManager {
 
         val icon = if (monitoringManager.isMonitoring()) "mdi:bluetooth" else "mdi:bluetooth-off"
 
-        var state = if (!BluetoothUtils.isOn(context)) "Bluetooth is turned off" else if (monitoringManager.isMonitoring()) "Monitoring" else "Stopped"
+        val state = if (!BluetoothUtils.isOn(context)) "Bluetooth is turned off" else if (monitoringManager.isMonitoring()) "Monitoring" else "Stopped"
 
-        var attr: Map<String, Any?> = mapOf()
+        val attr: MutableMap<String, Any?> = mutableMapOf()
```
jpelgrom, line 357: Try using type inference if possible.

```suggestion
        val attr = mutableMapOf<String, Any?>()
```

jpelgrom (commented):
Should the name format be aligned with core? At first glance, it appears to be [`UUID major.minor`](https://github.com/home-assistant/core/blob/a495df9759ee973620690ee8083d73181547934c/homeassistant/components/ibeacon/coordinator.py#L65) if no specific name is set instead of `UUID-major-minor`.

Also you're joining strings in 4 places, it might be time to create an extension property `name` to avoid the potential for typos in the future.

dshokouhi:
> Should the name format be aligned with core? At first glance, it appears to be [`UUID major.minor`](https://github.com/home-assistant/core/blob/a495df9759ee973620690ee8083d73181547934c/homeassistant/components/ibeacon/coordinator.py#L65) if no specific name is set instead of `UUID-major-minor`.

ah thanks for pointing that out, then maybe we should also update the `ID` attribute for the BLE transmitter to keep it inline?



> Also you're joining strings in 4 places, it might be time to create an extension property `name` to avoid the potential for typos in the future.

good call we should standardize it :)

app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
dshokouhi: will add in my next commit, thank you 🙏

jpelgrom:
> > Should the name format be aligned with core? At first glance, it appears to be [`UUID major.minor`](https://github.com/home-assistant/core/blob/a495df9759ee973620690ee8083d73181547934c/homeassistant/components/ibeacon/coordinator.py#L65) if no specific name is set instead of `UUID-major-minor`.
> 
> ah thanks for pointing that out, then we should also update the `ID` attribute for the BLE transmitter to keep it inline

I hadn't seen that one 🤔 It does explain why you used `-` but I still prefer using the same format as core.

> > Also you're joining strings in 4 places, it might be time to create an extension property `name` to avoid the potential for typos in the future.
> 
> good call we should standardize it :)

Now it is a function available for every class, I was thinking more something like this but that won't work now as it needs to support two classes :)

```kotlin
val IBeacon.name: String
    get() = "$uuid $major.$minor"
```

dshokouhi:
> Now it is a function available for every class, I was thinking more something like this but that won't work now as it needs to support two classes :)

I figure we may also use this in more places so its good that it can used by the app

jpelgrom (approved):

JBassett (approved):

PR #2916, opened 2022-09-25
Only check for beacons if the index is present

dshokouhi:
## Summary

Noticed the below sentry error in the latest beta so preventing accessing an index that will not exist so the app does not crash. I thought the [`assert()`](https://github.com/home-assistant/android/blob/1fb159999720b6065a3feb8198b344e267765750/app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt#L61)  above would take care of this but I guess not? Do we need that check?

```
java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
    at java.util.ArrayList.get(ArrayList.java:437)
    at io.homeassistant.companion.android.bluetooth.ble.IBeaconMonitor.setBeacons(IBeaconMonitor.kt:64)
    at io.homeassistant.companion.android.bluetooth.ble.MonitoringManager$startMonitoring$2.invokeSuspend$lambda-0(MonitoringManager.kt:53)
    at io.homeassistant.companion.android.bluetooth.ble.MonitoringManager$startMonitoring$2.$r8$lambda$a-4DQIY1f5dyfoiXoAHs7lcceLI
    at io.homeassistant.companion.android.bluetooth.ble.MonitoringManager$startMonitoring$2$$ExternalSyntheticLambda0.onChanged
    at androidx.lifecycle.LiveData.considerNotify(LiveData.java:133)
    at androidx.lifecycle.LiveData.dispatchingValue(LiveData.java:151)
    at androidx.lifecycle.LiveData.setValue(LiveData.java:309)
    at androidx.lifecycle.MutableLiveData.setValue(MutableLiveData.java:50)
    at androidx.lifecycle.LiveData$1.run(LiveData.java:93)
    at android.os.Handler.handleCallback(Handler.java:938)
    at android.os.Looper.loopOnce(Looper.java:226)
    at android.os.Looper.loop(Looper.java:313)
    at android.app.ActivityThread.main(ActivityThread.java:8751)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1135)
```

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt
@@ -60,12 +60,13 @@ class IBeaconMonitor {
         }
         assert(sorted.count() == beacons.count())
         beacons.forEachIndexed foreach@{ i, existingBeacon ->
-
-            if (name(sorted[i].uuid, sorted[i].major, sorted[i].minor) != name(existingBeacon.uuid, existingBeacon.major, existingBeacon.minor) || // the distance order switched
-                kotlin.math.abs(sorted[i].distance - existingBeacon.distance) > 0.5 // the distance difference is greater than 0.5m
-            ) {
-                requireUpdate = true
-                return@foreach
+            if (i <= sorted.size) {
+                if (name(sorted[i].uuid, sorted[i].major, sorted[i].minor) != name(existingBeacon.uuid, existingBeacon.major, existingBeacon.minor) || // the distance order switched
+                    kotlin.math.abs(sorted[i].distance - existingBeacon.distance) > 0.5 // the distance difference is greater than 0.5m
+                ) {
+                    requireUpdate = true
+                    return@foreach
+                }
             }
         }
         if (requireUpdate) {
--- a/app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt
@@ -59,7 +59,7 @@ class IBeaconMonitor {
             return
         }
         beacons.forEachIndexed foreach@{ i, existingBeacon ->
-            if (i <= sorted.size) {
+            if (i < sorted.size) {
                 if (name(sorted[i].uuid, sorted[i].major, sorted[i].minor) != name(existingBeacon.uuid, existingBeacon.major, existingBeacon.minor) || // the distance order switched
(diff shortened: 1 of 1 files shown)


jpelgrom:
> I thought the [`assert()`](https://github.com/home-assistant/android/blob/1fb159999720b6065a3feb8198b344e267765750/app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt#L61) above would take care of this but I guess not?

Only ["if the value is false and runtime assertions have been enabled on the JVM using the `-ea` JVM option"](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/assert.html#assert), I don't think that option is present.

> Do we need that check?

Good question, I don't see the code catching the exception so presumably this crashes the monitoring job? @Adi146 Could you clarify what is intended here?

Adi146:
The intention was to split the logic when we update the beacon list in three steps.
1. when a new beacon was found
2. when we can find a beacon for x scan cycles. We can not throw it away immediately because the scan job is sometime a bit flacky and we might have just missed it for this scan
3. when the distance order of the beacon did change or there was a major distance change on one of the beacons.

Right at the moment, I can not figure it out in which case the exception might be triggered as we already cover both array size changes in the first two steps.

dshokouhi:
Ok should I remove the `assert()` check for now since it doesn't seem to be stopping the code look it should? I am not entirely sure the fix I have proposed is correct but its better to not error out when we know we don't have to :)

Adi146:
Yeah I think we can just jump out of the method if the condition of the assert it true. I'm not entirely sure if this fixes the error but I think it would be worth a try. If it doesn't fix it we might have to lock the list while this method runs, it might be caused by a race condition

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt
```diff
@@ -58,14 +58,14 @@ class IBeaconMonitor {
...
         }
-        assert(sorted.count() == beacons.count())
         beacons.forEachIndexed foreach@{ i, existingBeacon ->
-
-            if (name(sorted[i].uuid, sorted[i].major, sorted[i].minor) != name(existingBeacon.uuid, existingBeacon.major, existingBeacon.minor) || // the distance order switched
-                kotlin.math.abs(sorted[i].distance - existingBeacon.distance) > 0.5 // the distance difference is greater than 0.5m
-            ) {
-                requireUpdate = true
-                return@foreach
+            if (i <= sorted.size) {
```
jpelgrom, line 62: Indexes are 0-based

```suggestion
            if (i < sorted.size) {
```

jpelgrom (approved):

JBassett (approved):

PR #2918, opened 2022-09-25
Fix BLE monitor new beacon check

jpelgrom:
## Summary

Fixes a bug introduced in #2909 where the app keeps thinking there are new beacons because it is only checking for the UUID while the app now stores beacons using UUID+major+minor.

## Screenshots

n/a

n/a


dshokouhi (approved):
Good catch thank you

JBassett (approved):

PR #2920, opened 2022-09-25
Breaking Change: Add bluetooth connect permission to beacon monitor to prevent log spam from library

dshokouhi:
## Summary

Noticed the following error spamming logcat while testing other PR fixes so adding in the permission to avoid that.

```
2022-09-25 08:59:35.789  2898-3408  BeaconParser            io....stant.companion.android.debug  D  Cannot read device name without Manifest.permission.BLUETOOTH_CONNECT
```

Marking as a breaking change as the sensor may not be enabled if permissions are not granted.


jpelgrom (approved):
> Marking as a breaking change as the sensor may not be enabled if permissions are not granted.

The permission will almost certainly be bundled in the general Nearby devices permission group so while technically a breaking change most users won't notice.

JBassett (approved):

PR #2922, opened 2022-09-25
Show custom icon if set for the entity in device controls

dshokouhi:
## Summary

If the `icon` attribute is present then the app will use that icon to display in device controls. The color of the icon is controlled by the state of the entity in an effort to match device type behavior.

## Screenshots

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
@@ -9,10 +9,12 @@ import android.service.controls.Control
 import android.service.controls.actions.ControlAction
 import androidx.annotation.RequiresApi
 import com.mikepenz.iconics.IconicsDrawable
+import com.mikepenz.iconics.utils.sizeDp
 import com.mikepenz.iconics.utils.toAndroidIconCompat
 import io.homeassistant.companion.android.common.R
 import io.homeassistant.companion.android.common.data.integration.Entity
 import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
+import io.homeassistant.companion.android.common.data.integration.domain
 import io.homeassistant.companion.android.common.data.websocket.impl.entities.AreaRegistryResponse
 import io.homeassistant.companion.android.webview.WebViewActivity
 
@@ -57,9 +59,11 @@ interface HaControl {
         ) {
             val iconName = entity.attributes["icon"]!!.toString().split(':')[1]
             val iconDrawable =
-                IconicsDrawable(context, "cmd-$iconName")
+                IconicsDrawable(context, "cmd-$iconName").apply {
+                    sizeDp = 48
+                }
             val colorTint = when (entity.state) {
-                "on" -> "#FDD663"
+                "on" -> if (entity.domain == "light") "#FDD663" else "#8AB4F8"
                 "off", "unavailable", "unknown" -> "#A6A6A7"
                 else -> "#8AB4F8"
             }
--- a/app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
@@ -64,16 +64,22 @@ interface HaControl {
                 IconicsDrawable(context, "cmd-$iconName").apply {
                     sizeDp = 48
                 }
-            val colorTint = when {
-                entity.domain == "light" && entity.state == "on" -> R.color.colorDeviceControlsLightOn
-                entity.domain == "camera" -> R.color.colorDeviceControlsCamera
-                entity.domain == "climate" && entity.state == "heat" -> R.color.colorDeviceControlsThermostatHeat
-                entity.state in listOf("off", "unavailable", "unknown") -> R.color.colorDeviceControlsOff
-                else -> R.color.colorDeviceControlsDefaultOn
-            }
+            if (iconDrawable.icon != null) {
+                val colorTint = when {
+                    entity.domain == "light" && entity.state == "on" -> R.color.colorDeviceControlsLightOn
+                    entity.domain == "camera" -> R.color.colorDeviceControlsCamera
+                    entity.domain == "climate" && entity.state == "heat" -> R.color.colorDeviceControlsThermostatHeat
+                    entity.state in listOf(
+                        "off",
+                        "unavailable",
+                        "unknown"
+                    ) -> R.color.colorDeviceControlsOff
+                    else -> R.color.colorDeviceControlsDefaultOn
+                }
 
-            iconDrawable.setTint(ContextCompat.getColor(context, colorTint))
-            control.setCustomIcon(iconDrawable.toAndroidIconCompat().toIcon(context))
+                iconDrawable.setTint(ContextCompat.getColor(context, colorTint))
+                control.setCustomIcon(iconDrawable.toAndroidIconCompat().toIcon(context))
+            }
         }
 
         return provideControlFeatures(context, control, entity, area, baseUrl).build()
--- a/app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
@@ -3,10 +3,17 @@ package io.homeassistant.companion.android.controls
 import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
+import android.graphics.Color
 import android.os.Build
 import android.service.controls.Control
 import android.service.controls.actions.ControlAction
+import androidx.annotation.ColorInt
 import androidx.annotation.RequiresApi
+import androidx.compose.material.MaterialTheme
+import androidx.core.graphics.drawable.toIcon
+import com.google.android.material.color.MaterialColors
+import com.mikepenz.iconics.IconicsDrawable
+import com.mikepenz.iconics.utils.toAndroidIconCompat
 import io.homeassistant.companion.android.common.R
 import io.homeassistant.companion.android.common.data.integration.Entity
 import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
@@ -48,6 +55,15 @@ interface HaControl {
         )
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
             control.setAuthRequired(authRequired)
+        }
+        if (entity.attributes["icon"]?.toString()?.startsWith("mdi:") == true &&
+            !entity.attributes["icon"]?.toString()?.substringAfter(":").isNullOrBlank()
+        ) {
+            val iconName = entity.attributes["icon"]!!.toString().split(':')[1]
+            val iconDrawable =
+                IconicsDrawable(context, "cmd-$iconName")
+            iconDrawable.setTint(context.resources.getColor(R.color.colorAccent))
+            control.setCustomIcon(iconDrawable.toAndroidIconCompat().toIcon(context))
         }
 
         return provideControlFeatures(context, control, entity, area, baseUrl).build()
--- a/app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
@@ -7,11 +7,7 @@ import android.graphics.Color
 import android.os.Build
 import android.service.controls.Control
 import android.service.controls.actions.ControlAction
-import androidx.annotation.ColorInt
 import androidx.annotation.RequiresApi
-import androidx.compose.material.MaterialTheme
-import androidx.core.graphics.drawable.toIcon
-import com.google.android.material.color.MaterialColors
 import com.mikepenz.iconics.IconicsDrawable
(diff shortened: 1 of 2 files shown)


jpelgrom:
Also if anyone with a device with an interface that isn't close to stock/Google Android (for example Samsung or Huawei) is reading this, could you check the app and see if the colors are somewhat close to the controls without a custom icon?

If there's a major difference - or even worse, the icons are now hard to see - I feel like there should be manufacturer specific colors in this PR to keep the controls usable and nice for everyone.

dshokouhi:
I will leave as a draft until we get more confirmation from users and make the logic better

SkechyWolf:
Custom icons seem to have a weird radial gradient in the background Will follow up with a photo

SkechyWolf:
Looks fine in dark mode 
also icons or at least the car looks slightly more blurry

jpelgrom:
> also icons or at least the car looks slightly more blurry 

That's probably because no size is set when creating the icon, in that case it will default to a size (I think 20x20 or 24x24 dp) which is presumably smaller than the icon size on your device.

dshokouhi:
hmm so you think we will need to calculate the size of the icon? or is there a good "safe" size we can go with?

jpelgrom:
> hmm so you think we will need to calculate the size of the icon?

I don't think we can as the app only supplies data, no views.

> or is there a good "safe" size we can go with?

Probably something a little larger, maybe 48x48dp? That should scale nicely from the current size, and looks like it should be good enough (the screenshot doesn't look very low res).

dshokouhi:
> My input booleans are now yellow like a light when on, that doesn't seem right. Instead of matching it to only the state value, you should probably match it to [state _and_ device type like the system](https://android.googlesource.com/platform/frameworks/base/+/766bd4cedbd00fe300f8eb693001395c380fc067/packages/SystemUI/src/com/android/systemui/controls/ui/RenderInfo.kt#87).

Is it me or is only the light device type set to a different color for `on` / `off` ? The others seem to be the default blue. I know thermostat and camera have different ones but those seem to be special cases.



> It's somewhat annoying that the icon will always use one color, using a state list like the default icons [is simply ignored](https://android.googlesource.com/platform/frameworks/base/+/6ffa5fcc40963b0c80ec80b7694decb8d886f1eb/packages/SystemUI/src/com/android/systemui/controls/management/ControlAdapter.kt#289). This will mix colored with grey icons in the list of new controls. To better match how the system shows colorful icons for all controls in the list of new controls, maybe you could always act as if they were in an active state in that case? (I'm thinking the best way to do this would be to add a new property in the `createControl` function that indicates if we're dealing with stateful or stateless controls.)

I still don't see how we can reliably detect when we are in the add controls screen vs when a control has been added. It seems that with custom icons we can't do that and will need to go by the state of the device. I personally don't mind the intermixed colors because its already a mixture of yellow, blue and grey when you include cameras, input booleans and lights on that screen. In fact today it also adds our app icon for scenes and what not. I would be in favor of letting the icon update the color with the state as opposed not to updating at all.

jpelgrom:
> Is it me or is only the light device type set to a different color for `on` / `off` ? The others seem to be the default blue. I know thermostat and camera have different ones but those seem to be special cases.

Yes it seems to be blue for most, yellow for lights, white for camera icons, and a few options for thermostats/climate entities 👍 We do support cameras and climate entities (as a thermostat) though so I think it's not just a special case and we should include them.

> I still don't see how we can reliably detect when we are in the add controls screen vs when a control has been added. (...)

Isn't it add controls screen = created from `createPublisherForAllAvailable()`, normal use/controls that are added = created from `createPublisherFor(controlIds)`?

> its already a mixture of yellow, blue and grey (...) it also adds our app icon for scenes and what not

Good point, maybe it isn't as noticeable as I said. If you feel like it's not feasible to support feel free to leave it out.

app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
```diff
@@ -49,6 +54,22 @@ interface HaControl {
...
+            val iconName = entity.attributes["icon"]!!.toString().split(':')[1]
+            val iconDrawable =
+                IconicsDrawable(context, "cmd-$iconName").apply {
+                    sizeDp = 48
+                }
+            val colorTint = when (entity.state) {
+                "on" -> if (entity.domain == "light") "#FDD663" else "#8AB4F8"
+                "off", "unavailable", "unknown" -> "#A6A6A7"
+                else -> "#8AB4F8"
+            }
```
jpelgrom, line 69: Why not use color resources to give it a more descriptive name and a preview in Android Studio next to the line numbers?

dshokouhi:
Ok made an attempt to match everything from:

https://android.googlesource.com/platform/frameworks/base/+/766bd4cedbd00fe300f8eb693001395c380fc067/packages/SystemUI/src/com/android/systemui/controls/ui/RenderInfo.kt#87

to https://cs.android.com/android/platform/superproject/+/master:frameworks/base/packages/SystemUI/res/values/colors.xml

also moved to color resources since it looks better thank you for that suggestion :)



> > I still don't see how we can reliably detect when we are in the add controls screen vs when a control has been added. (...)
> 
> Isn't it add controls screen = created from `createPublisherForAllAvailable()`, normal use/controls that are added = created from `createPublisherFor(controlIds)`?
> 
> > its already a mixture of yellow, blue and grey (...) it also adds our app icon for scenes and what not
> 
> Good point, maybe it isn't as noticeable as I said. If you feel like it's not feasible to support feel free to leave it as it is.

Thanks for pointing that out, I was trying to keep the logic within `HaControl.kt` as i think we can handle the logic here instead of splitting it up which looks like would be a bit of a refactor to get that right. I personally like the simplicity of keeping the logic in this one file :)

app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
dshokouhi: great suggestion, thank you 🙏

app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
```diff
@@ -49,6 +56,25 @@ interface HaControl {
...
+                }
+            val colorTint = when {
+                entity.domain == "light" && entity.state == "on" -> context.getString(R.color.colorDeviceControlsLightOn)
+                entity.domain == "camera" -> context.getString(R.color.colorDeviceControlsOff)
+                entity.domain == "climate" && entity.state == "heat" -> context.getString(R.color.colorDeviceControlsThermostatHeat)
+                entity.state == "on" -> context.getString(R.color.colorDeviceControlsDefaultOn)
+                entity.state == "off" || entity.state == "unavailable" || entity.state == "unknown" -> context.getString(R.color.colorDeviceControlsOff)
+                else -> context.getString(R.color.colorDeviceControlsDefaultOn)
+            }
+            iconDrawable.setTint(Color.parseColor(colorTint))
```
jpelgrom, line 75: Don't get a color resource as a string 😬 You can simply set `colorTint` to the `R.color.colorDeviceControlsSomething`, and then use `ContextCompat.getColor(context, colorTint)` here instead.

app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
```diff
@@ -49,6 +56,25 @@ interface HaControl {
...
+            val iconName = entity.attributes["icon"]!!.toString().split(':')[1]
+            val iconDrawable =
+                IconicsDrawable(context, "cmd-$iconName").apply {
+                    sizeDp = 48
+                }
+            val colorTint = when {
+                entity.domain == "light" && entity.state == "on" -> context.getString(R.color.colorDeviceControlsLightOn)
+                entity.domain == "camera" -> context.getString(R.color.colorDeviceControlsOff)
+                entity.domain == "climate" && entity.state == "heat" -> context.getString(R.color.colorDeviceControlsThermostatHeat)
+                entity.state == "on" -> context.getString(R.color.colorDeviceControlsDefaultOn)
```
jpelgrom, line 71: This branch isn't necessary and handled by `else`

app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
```diff
@@ -49,6 +56,25 @@ interface HaControl {
...
+                entity.state == "off" || entity.state == "unavailable" || entity.state == "unknown" -> context.getString(R.color.colorDeviceControlsOff)
```
jpelgrom, line 72:
```suggestion
                entity.state in listOf("off", "unavailable", "unknown") -> context.getString(R.color.colorDeviceControlsOff)
```

jpelgrom (commented):
Looks like the icon became a bit larger, there was [1dp padding](https://github.com/mikepenz/Android-Iconics/blob/0e298f9fff213eed62e007c4cefa7311acdb26ce/iconics-core/src/main/java/com/mikepenz/iconics/utils/IconicsDrawableExtensions.kt#L110). Will most users notice? :)

common/src/main/res/values/colors.xml
```diff
@@ -27,4 +27,8 @@
     <color name="colorCodeBackground">#f5f5f5</color>
     <color name="colorAlertWarning">#1fffa600</color>
     <color name="colorOnAlertWarning">#ffa600</color>
+    <color name="colorDeviceControlsLightOn">#FDD663</color>
+    <color name="colorDeviceControlsOff">#A6A6A7</color>
```
jpelgrom, line 31: This does not match [the system value](https://android.googlesource.com/platform/frameworks/base/+/766bd4cedbd00fe300f8eb693001395c380fc067/packages/SystemUI/src/com/android/systemui/controls/ui/RenderInfo.kt#87), which is #9AA0A6 instead. I guess this is the first manufacturer difference...

dshokouhi: those are so close to each other LOL 🤦

jpelgrom: Then you shouldn't have said you'd matched the system colors 😉

app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt
```diff
@@ -49,6 +56,25 @@ interface HaControl {
...
+            !entity.attributes["icon"]?.toString()?.substringAfter(":").isNullOrBlank()
+        ) {
+            val iconName = entity.attributes["icon"]!!.toString().split(':')[1]
+            val iconDrawable =
+                IconicsDrawable(context, "cmd-$iconName").apply {
+                    sizeDp = 48
+                }
+            val colorTint = when {
+                entity.domain == "light" && entity.state == "on" -> context.getString(R.color.colorDeviceControlsLightOn)
+                entity.domain == "camera" -> context.getString(R.color.colorDeviceControlsOff)
```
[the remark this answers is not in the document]
dshokouhi: Oh good catch, must've skipped it due to `grey` being in the name lol

jpelgrom:
Looks good to me on a Pixel device! (Thanks for all the quick responses to review comments, no matter how minor they were.)

dshokouhi:
> Looks good to me on a Pixel device! 

🎉 

> (Thanks for all the quick responses to review comments, no matter how minor they were.)

I like to have a nice balance between perfection and simplicity 😛 

As always appreciate the attention to detail in the review 🙏

SkechyWolf:
Looks good only slight color difference of lights (the yellow) in dark mode (manufacturer difference) and that radial gradient in light mode don't know what causes that

jpelgrom:
If the icon that is specified is not supported by the app (because the library is a bit old 😅 working on getting it updated), the device control tile will show a blank space for the icon. Should this be handled by falling back to the system icon instead? I think it looks OK on a Pixel but can imagine it looking less OK on a Samsung device based on the screenshot above.
Device control tile without an icon and the text 'Off' offset roughly 1/3 from the left

(`iconDrawable.icon` will be `null`, test with something that was added in the last 4 releases or so like `mdi:sign-yield`)

SkechyWolf:
Here is how it looks on a Samsung device

dshokouhi:
> If the icon that is specified is not supported by the app (because the library is a bit old 😅 working on getting it updated), the device control tile will show a blank space for the icon.

yea same thing happens with `notification_icon` when a user sets it to something that has no icon



> (`iconDrawable.icon` will be `null`, test with something that was added in the last 4 releases or so like `mdi:sign-yield`)

I will take a look at this next chance thank you 🙏

dshokouhi:
Ok updated and it works well, now the icon will only set if it exists in the library so even if a user had `mdi:abcd` which resolves to nothing we will not override the icon.

I think this improvement can also be used in other places but to keep this PR small I will not opt to fix those just yet as some of them may require some refactoring.

jpelgrom (approved):

jpelgrom:
> I think this improvement can also be used in other places but to keep this PR small I will not opt to fix those just yet as some of them may require some refactoring.

Yes that should be done in separate PRs to make it easier to review. Also in existing places it could be considered a breaking change if someone currently depends on an invalid value to make the icon blank.

JBassett (approved):

PR #1950, opened 2021-11-21
Feature/show launch screen behind webview

LasseRosenow:
## Summary
This is the new implementation of #1943

After the launch screen the `WebViewActivity` is shown. But the `WebViewActivity` needs some time until the html of the web frontend is loaded. So there is a blank screen between the launch screen and when the frontend is rendered.
To fix this, I changed the background color of the webview to transparent and made the the added the launch screen drawable as the background of the `WebViewActivity`. (On android 12 it will just be a solid color for now).

Also I removed the blue status and navigation bars from the `WebViewActivity` in light mode. This is another step in removing any theme specific colors from the LaunchScreen until the app is drawn and the correct theme color is being extracted from the  web app.

This greatly improves the startup experience of the app and removes some annoying blinking issues when launching.

## Videos
<table>
<tr>
<th>Before</th><th>After</th>
</tr>
<tr>
<td>

https://user-images.githubusercontent.com/10547444/142776946-751ac65e-aae4-43de-8bb2-c46e3d254cdd.mp4

</td>
<td>

https://user-images.githubusercontent.com/10547444/142776947-bbbb3046-7d2f-4eb0-a63d-d572bad2711b.mp4

</td>
</tr>
</table>

### Before
<table>
<tr>
<th>Launch Screen</th><th>WebView</th>
</tr>
<tr>
<td>

before_Moment

</td>
<td>

before_Moment2

</td>
</tr>
</table>

### After
<table>
<tr>
<th>Launch Screen</th><th>WebView</th>
</tr>
<tr>
<td>

before_Moment

</td>
<td>

before_Moment

</td>
</tr>
</table>

## Link to pull request in Documentation repository

Documentation: home-assistant/companion.home-assistant#

PR #2923, opened 2022-09-26
Simplify themes

jpelgrom:
## Summary

This PR simplifies themes by:

 - **Merging the 'Follow Home Assistant' and 'Follow System Settings' theme options**
    This was originally introduced in #1501 to be able to control the system bar colors, but that functionality was removed in #1950. As the two options currently behave identically and there haven't been any complaints, I'm merging them back in to one setting 'Follow System Settings'.
    The merged option will be called 'Follow System Settings' because 'Follow Home Assistant' makes it sound like the app will use the theme setting from the frontend/profile which is not correct.
 - **Removing webview force dark settings**
   While the developer documentation [suggests that you use these 'force' settings](https://developer.android.com/develop/ui/views/layout/webapps/dark-theme), they are no longer required for the app. With the latest version of the webview and all the app's dependencies, the frontend will follow the app's theme setting without the need to set force dark (so dark mode app + auto frontend = dark mode frontend, or dark mode app + light frontend = light mode frontend simply works without any additional settings). I have tested this on Android 8.1 (no system level dark theme), Android 11 and Android 13.
    When the app starts targeting Android 13, the existing setting also wouldn't be possible to set [as it will only allow managing algorithmic darkening](https://developer.android.com/jetpack/androidx/releases/webkit#1.5.0) which was already disabled by setting the strategy to [`DARK_STRATEGY_WEB_THEME_DARKENING_ONLY`](https://developer.android.com/reference/kotlin/androidx/webkit/WebSettingsCompat#DARK_STRATEGY_WEB_THEME_DARKENING_ONLY()).

## Screenshots

n/a

n/a

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
@@ -66,7 +66,7 @@ class AuthenticationFragment : Fragment() {
                 MdcTheme {
                     AndroidView({
                         WebView(requireContext()).apply {
-                            themesManager.setThemeForWebView(requireContext(), settings)
+                            themesManager.setThemeForWebView()
                             settings.javaScriptEnabled = true
                             settings.domStorageEnabled = true
                             settings.userAgentString = settings.userAgentString + " ${HomeAssistantApis.USER_AGENT_STRING}"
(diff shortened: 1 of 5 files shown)


dshokouhi (approved):
Looks good to me, nice to simplify things :)

JBassett (approved):

jpelgrom:
@bcutter Instead of commenting on a PR, could you create a new issue and provide the requested information? Especially more detailed device information will be useful. Thanks!

PR #2936, opened 2022-10-01
Default to app icon if icon is not available for notification

dshokouhi:
## Summary

Default back to the app icon if the MDI icon does not exist in the library.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -1068,8 +1068,11 @@ class MessagingManager @Inject constructor(
         if (data[NOTIFICATION_ICON]?.startsWith("mdi:") == true && !data[NOTIFICATION_ICON]?.substringAfter("mdi:").isNullOrBlank() && Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
             val iconName = data[NOTIFICATION_ICON]!!.split(":")[1]
             val iconDrawable =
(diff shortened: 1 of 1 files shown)


jpelgrom (approved):
Nice improvement!

Based on a quick search, other places this could be useful:

 - Media player widget shows a source icon if set using MDI or mdi:cast otherwise, but assumes if there is an `icon` value it is valid
 - Wear OS entity chips show an icon, currently it actually handles null by not doing anything/keeping the placeholder ([mdi:bookmark](https://materialdesignicons.com/icon/bookmark)), use a domain icon or switch to showing nothing like the frontend?

dshokouhi:
> Media player widget shows a source icon if set using MDI or mdi:cast otherwise, but assumes if there is an icon value it is valid

So for this we should probably default back to `mdi:cast` in case its null

> * Wear OS entity chips show an icon, currently it actually handles null by not doing anything/keeping the placeholder ([mdi:bookmark](https://materialdesignicons.com/icon/bookmark)), use a domain icon or switch to showing nothing like the frontend?

oh I didn't think it handled `null` based on https://github.com/home-assistant/android/blob/50d387a726b9fadcada0f6ba8aea511b57cd3df0/wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt#L50

jpelgrom:
> oh I didn't think it handled `null` based on (...)

Not there, but in Compose it is handled:

https://github.com/home-assistant/android/blob/203e6868d10aca697babc2b61a98a5b5e91d5358/wear/src/main/java/io/homeassistant/companion/android/home/views/EntityUi.kt#L60

JBassett (approved):

PR #2938, opened 2022-10-01
Update change log week of 10/1 for beta

dshokouhi:
## Summary

Bumps release to 2022.10.1 in case we release again this month, resets the changelog for next release.

### DIFF
--- a/app/src/main/res/xml/changelog_master.xml
+++ b/app/src/main/res/xml/changelog_master.xml
@@ -2,6 +2,13 @@
 <changelog xmlns:tools="http://schemas.android.com/tools"
     tools:ignore="MissingDefaultResource">
     <release version="2022.10.1 - Phone" versioncode="2">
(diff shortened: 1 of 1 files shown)


jpelgrom:
Here's a new issue: beta users probably won't see the full changelog for 2022.10 considering the releases are only a few hours apart. Should we maybe keep the 2022.10 changes in the app's changelog as well for one more week?

dshokouhi:
> Here's a new issue: beta users probably won't see the full changelog for 2022.10 considering the releases are only a few hours apart. Should we maybe keep the 2022.10 changes in the app's changelog as well for one more week?

good point! will restore those changes above since they are breaking changes and we dont want them to miss that

jpelgrom (approved):

JBassett:
If we're going to leave them in can we leave them in a separate release block?

dshokouhi:
> 

let me see how it looks

dshokouhi:
image

JBassett (approved):

PR #2940, opened 2022-10-02
Set the media player widget icon if available, otherwise default to cast icon

dshokouhi:
## Summary

Only adds the custom icon if it is available in the library, otherwise we stick to the cast icon

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/media_player_controls/MediaPlayerControlsWidget.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/media_player_controls/MediaPlayerControlsWidget.kt
@@ -16,6 +16,7 @@ import android.widget.Toast
 import androidx.core.content.getSystemService
 import com.google.android.material.color.DynamicColors
 import com.mikepenz.iconics.IconicsDrawable
+import com.mikepenz.iconics.typeface.library.community.material.CommunityMaterial
 import com.squareup.picasso.Picasso
 import dagger.hilt.android.AndroidEntryPoint
 import io.homeassistant.companion.android.BuildConfig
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/widgets/media_player_controls/MediaPlayerControlsWidget.kt
```diff
@@ -231,15 +230,16 @@ class MediaPlayerControlsWidget : BaseWidgetProvider() {
                     )
                 }
 
-                if (icon == null || !icon.startsWith("mdi") || !icon.contains(":")) {
-                    icon = "mdi:cast"
+                var iconBitmap = IconicsDrawable(context, "cmd-cast").toBitmap()
```
jpelgrom, line 233: Minor performance improvement: the icon is known, so instead of using the string `"cmd-cast"` which will search all icons you can just specify `CommunityMaterial.Icon.cmd_cast`.

dshokouhi: > Minor performance improvement: the icon is known, so instead of using the string `"cmd-cast"` which will search all icons you can just specify `CommunityMaterial.Icon.cmd_cast`.

This doesn't seem to allow me to set it `toBitmap()` which is required for `setImageViewBitmap()` which is the only method we can use thats compatible with all supported API levels.  Unless I am missing something?

jpelgrom: This works for me:
```suggestion
                var iconBitmap = IconicsDrawable(context, CommunityMaterial.Icon.cmd_cast).toBitmap()
```

(+ import for `com.mikepenz.iconics.typeface.library.community.material.CommunityMaterial`)

dshokouhi: oh thats what you meant 😅

jpelgrom (approved):

JBassett (approved):

Issue #2932, opened 2022-09-29
Beacon Transmitter / Monitor - Ability to advertise and get additional attributes such as the name


Mincka:
**Is your feature request related to a problem? Please describe.**
It's possible to set a name on BlueCharm beacons. This name is retrieved by the HA Core Integration.
This information is retrieved from the `service_info.name` in the [HA Integration](https://github.com/home-assistant/core/blob/dev/homeassistant/components/ibeacon/coordinator.py#L62).

**Describe the solution you'd like**
Would it be possible to set and broadcast this information using the mobile app? It would be more user friendly.

I am not sure about the requirements for this parameter. In Bluecharm documentation, its says "16 or less ASCII characters", but on the mobile application itself, it says "17 or less ASCII characters"

**Describe alternatives you've considered, if any**
Just use UUID / Major / Minor as the unique identifier. Rename the device and the sensors in HA, but this applies only to HA, no to other beacon scanners that could benefit from name reading.

**Additional context**
Would it be relevant to also request the retrieval of this information using the Beacon Monitor? At this stage, I don't have any idea about how this information should be retrieved. I think we are hitting the limits of the current implementation of the Beacon Monitor. I agree with the new consensus about the ID format, but how can we get additional attributes with this format in the app?

```
a92ee200-5501-11e4-916c-0800200c9a66_12555_14421: 9.94
a92ee200-5501-11e4-916c-0800200c9a66_11475_24820: 11.43
```

Today it's about the "name" but tomorrow, users could also request TLM attributes to get battery level, temperature, ADV count, time since the beacon is up... Maybe I am pushing too far for TLM since it's another protocol. Still, I think it's relevant for the name. We could also break down each part as an attribute. May I suggest to format the beacons as a list of UUID_Major_Minor with attributes?

![image](https://user-images.githubusercontent.com/6184289/193002196-06db71bd-041b-4857-8bd1-bc2560c09452.png)
(Screenshot of the KBeacon app that retrieves information from a Bluecharm beacon)

```
icon: mdi:bluetooth
friendly_name: Téléphone Julien Beacon Monitor
beacons:
- 13370000-4368-6172-6d42-6561636f6e74_1000_1001
  distance: 0
  name: Julien
  uuid: 13370000-4368-6172-6d42-6561636f6e74
  major: 1000
  minor:  1001
- 13370000-4368-6172-6d42-6561636f6e74_1000_1002
  distance: 2.4
  name: Dog
[...]

PR #2941, opened 2022-10-02
Breaking Change: Retrieve beacon name for beacon monitor sensor

dshokouhi:
## Summary

Fixes: #2932 by retrieving the `name` in the Beacon Monitor.

This will be a breaking change as we are changing the attribute format

Leaving as a draft to get feedback and also need some help in making the array look better in the attributes. Not really sure what type of conversion we need to do there. As of now its a string 🤷‍♂️ 

Example of new attributes in current state of this PR:

```
icon: mdi:bluetooth
friendly_name: Pixel 6 Pro Beacon Monitor
1a01a2b8-13a7-4c98-a732-496554720c3d_100_1: {distance=1.1, name=null, uuid=1a01a2b8-13a7-4c98-a732-496554720c3d, major=100, minor=1}
50765cb7-d9ea-4e21-99a4-fa879613a492_15841_33291: {distance=18.54, name=null, uuid=50765cb7-d9ea-4e21-99a4-fa879613a492, major=15841, minor=33291}
9d05c5cc-b742-4b1d-96b8-6b8626854119_100_1: {distance=4.76, name=null, uuid=9d05c5cc-b742-4b1d-96b8-6b8626854119, major=100, minor=1}
```

## Screenshots

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt
@@ -15,6 +15,7 @@ data class IBeacon(
     var distance: Double,
     var rssi: Double,
     var skippedUpdated: Int,
+    var name: String?,
 )
 
 class IBeaconMonitor {
@@ -38,10 +39,11 @@ class IBeaconMonitor {
             val minor = newBeacon.id3.toString()
             val distance = round(newBeacon.distance * 100) / 100
             val rssi = newBeacon.runningAverageRssi
+            val name = newBeacon.bluetoothName
             if (!tmp.contains(name(uuid, major, minor))) { // we found a new beacon
                 requireUpdate = true
             }
-            tmp += Pair(name(uuid, major, minor), IBeacon(uuid, major, minor, distance, rssi, 0))
+            tmp += Pair(name(uuid, major, minor), IBeacon(uuid, major, minor, distance, rssi, 0, name))
         }
         val sorted = sort(tmp.values).toMutableList()
         if (requireUpdate) {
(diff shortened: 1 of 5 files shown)


leroyloren:
I'm afraid this would again complicate/break the work in e.g. NodeRED, do you think it's good to use attributes in an attribute?
How important are individual values?
I would expect the app to give me a list of available beacons with a unique ID to trigger the action, not other sub-attributes.
I think the most important thing is to announce the availability and possibly the distance, everything else is not necessary.

Mincka:
> everything else is not necessary.

Sorry but this is not about what you need, or what works or not out of the box with NodeRed.
The beacons are providing additional information that is not be limited to their name or their distance.

If you don't need this additional data, can't you use a simple regex or something similar to filter them out in NodeRed?

We don't know yet what will be the final format to hold this information but as far as we respect general guidance and practices, I am fine with that. I fully support the request to avoid spaces in the name of the attributes. Beyond that, I suggested another format in the original issue that seems more "YAMLish", but maybe not better than this original draft proposal.

In your case, I think it would be even better if you could specify the "sub attribute" for your beacon, something like:
`new_state.attributes.1a01a2b8-13a7-4c98-a732-496554720c3d_100_1.distance`
or
`new_state.attributes.1a01a2b8-13a7-4c98-a732-496554720c3d_100_1[distance]`

leroyloren:
> In your case, I think it would be even better if you could specify the "sub attribute" for your beacon, something like: `new_state.attributes.1a01a2b8-13a7-4c98-a732-496554720c3d_100_1.distance` or `new_state.attributes.1a01a2b8-13a7-4c98-a732-496554720c3d_100_1[distance]`

This could very well be 👍

https://github.com/zachowj/node-red-contrib-home-assistant-websocket/discussions/667

dshokouhi:
update: Looks like we cannot change the `name` as its supposed to already be transmitting the friendly device name however I am not sure if it is.

https://github.com/AltBeacon/android-beacon-library/issues/987#issuecomment-653904387

`bluetoothName is only populated (sometimes) after scanning, and only if the BLE device detected with that beacon has also recently advertised its name in a separate packet.`

Will need to update this PR to reflect that.

dshokouhi:
closing this PR as the approach is not right for this

PR #2956, opened 2022-10-07
Add permissions requests to wear OS sensors and add network sensors

dshokouhi:
## Summary

Adds permissions requests to Wear OS when a user enables a sensor, also adds Network Sensors as a start.

WTH: https://community.home-assistant.io/t/wth-dont-we-have-more-sensor-in-wear-os/469354

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#834

## Any other notes

This was a "quick and dirty" approach to just get some more sensors added as permissions is holding up adding some of the sensors.  Some of this code may be stale but wanted to get some extra eyes on this.

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
@@ -279,7 +279,6 @@ class MainViewModel @Inject constructor(
         isEnabled: Boolean
     ) {
         sensorDao.setSensorsEnabled(listOf(basicSensor.id), isEnabled)
-        sensorDao.updateLastSentStateAndIcon(basicSensor.id, null, null)
         SensorWorker.start(getApplication())
     }
 
--- a/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
@@ -21,6 +21,7 @@ import io.homeassistant.companion.android.data.SimplifiedEntity
 import io.homeassistant.companion.android.database.sensor.SensorDao
 import io.homeassistant.companion.android.database.wear.FavoritesDao
 import io.homeassistant.companion.android.database.wear.getAllFlow
+import io.homeassistant.companion.android.sensors.SensorWorker
 import io.homeassistant.companion.android.util.RegistriesDataHandler
 import kotlinx.coroutines.async
 import kotlinx.coroutines.flow.Flow
@@ -278,6 +279,8 @@ class MainViewModel @Inject constructor(
         isEnabled: Boolean
     ) {
         sensorDao.setSensorsEnabled(listOf(basicSensor.id), isEnabled)
+        sensorDao.updateLastSentStateAndIcon(basicSensor.id, null, null)
+        SensorWorker.start(getApplication())
     }
 
     fun getAreaForEntity(entityId: String): AreaRegistryResponse? =
--- a/wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
@@ -25,12 +25,11 @@ fun SensorUi(
     basicSensor: SensorManager.BasicSensor,
     onSensorClicked: (String, Boolean) -> Unit,
 ) {
-    var checked = sensor?.enabled == true
+    val checked = sensor?.enabled == true
 
     val backgroundRequest =
         rememberLauncherForActivityResult(ActivityResultContracts.RequestPermission()) {
             onSensorClicked(basicSensor.id, it)
-            checked = sensor?.enabled == true && it
         }
 
     val permissionLaunch = rememberLauncherForActivityResult(
@@ -50,7 +49,6 @@ fun SensorUi(
                 allGranted = false
         }
         onSensorClicked(basicSensor.id, allGranted)
-        checked = sensor?.enabled == true && allGranted
     }
 
     val perm = manager.checkPermission(LocalContext.current, basicSensor.id)
--- a/wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
@@ -32,14 +32,17 @@ fun SensorUi(
     val permissionLaunch = rememberLauncherForActivityResult(
         ActivityResultContracts.RequestMultiplePermissions()
     ) { isGranted ->
+        var allGranted = true
         isGranted.forEach {
             if (manager.requiredPermissions(basicSensor.id)
                 .contains(Manifest.permission.ACCESS_FINE_LOCATION) && manager.requiredPermissions(basicSensor.id)
                     .contains(Manifest.permission.ACCESS_BACKGROUND_LOCATION) && it.key == Manifest.permission.ACCESS_FINE_LOCATION
             )
                 backgroundRequest.launch(Manifest.permission.ACCESS_BACKGROUND_LOCATION)
-            checked = sensor?.enabled == true && it.value
+            if (!it.value)
+                allGranted = false
         }
+        checked = sensor?.enabled == true && allGranted
     }
 
     val perm = manager.checkPermission(LocalContext.current, basicSensor.id)
--- a/wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
@@ -26,10 +26,20 @@ fun SensorUi(
 ) {
     var checked = sensor?.enabled == true
 
+    val backgroundRequest =
+        rememberLauncherForActivityResult(ActivityResultContracts.RequestPermission()) {}
+
     val permissionLaunch = rememberLauncherForActivityResult(
         ActivityResultContracts.RequestMultiplePermissions()
     ) { isGranted ->
-        isGranted.forEach { checked = sensor?.enabled == true && it.value }
+        isGranted.forEach {
+            if (manager.requiredPermissions(basicSensor.id)
+                .contains(Manifest.permission.ACCESS_FINE_LOCATION) && manager.requiredPermissions(basicSensor.id)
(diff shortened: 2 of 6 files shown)


jpelgrom:
Foreground/background location permissions also need to be handeld seperately on Wear OS. Using the Wear OS 3 emulator (API 30), after tapping on WiFi Connection it is instantly denied with logcat showing:

```
2022-10-07 21:28:34.559 12392-12392/com.google.android.permissioncontroller E/GrantPermissionsActivity: Apps targeting 30 must have foreground permission before requesting background and must request background on its own.
```

dshokouhi:
> Foreground/background location permissions also need to be handeld separately on Wear OS. Using the Wear OS 3 emulator (API 30), after tapping on WiFi Connection it is instantly denied with logcat showing:
> 
> ```
> 2022-10-07 21:28:34.559 12392-12392/com.google.android.permissioncontroller E/GrantPermissionsActivity: Apps targeting 30 must have foreground permission before requesting background and must request background on its own.
> ```

Thanks for that! I submitted a quick fix for that but its not pretty as of the moment.

jpelgrom:
Still not working - while it will ask for while in use location permission, after granting it I still cannot toggle the sensor on. You need to request `Manifest.permission.ACCESS_BACKGROUND_LOCATION` from the permission result, because when tapping the toggle it will ask for everything except `Manifest.permission.ACCESS_BACKGROUND_LOCATION`.

Also not a problem yet, but `checked` should only be set to `true` if _all_ permissions _for the sensor_ are granted, not looping through the request results one by one?

It's getting very close to having too much logic in the UI layer.

dshokouhi:
> Still not working - while it will ask for while in use location permission, after granting it I still cannot toggle the sensor on. You need to request `Manifest.permission.ACCESS_BACKGROUND_LOCATION` from the permission result, because when tapping the toggle it will ask for everything except `Manifest.permission.ACCESS_BACKGROUND_LOCATION`.
> 
> Also not a problem yet, but `checked` should only be set to `true` if _all_ permissions _for the sensor_ are granted, not looping through the request results one by one?

Thanks hopefully its all a bit better after the latest commits :)



> It's getting very close to having too much logic in the UI layer.

I agree, its a bit difficult to do the permissions properly without refactoring things a bit I think. I was hoping to perform the checks here on the page where we enable sensors as it seemed to be the most practical.

dshokouhi:
For anyone who wants to test this issue, please be aware of this issue which may impact the initial state of the sensor https://github.com/home-assistant/core/issues/79837

you will need to trigger a second update

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
```diff
@@ -278,6 +279,8 @@ class MainViewModel @Inject constructor(
         isEnabled: Boolean
     ) {
         sensorDao.setSensorsEnabled(listOf(basicSensor.id), isEnabled)
+        sensorDao.updateLastSentStateAndIcon(basicSensor.id, null, null)
```
jpelgrom, line 282: This line isn't necessary because the sensor DAO will already [set the state and icon to null](https://github.com/home-assistant/android/blob/93844ffea03be01e35c618114073b0e26fd47f39/common/src/main/java/io/homeassistant/companion/android/database/sensor/SensorDao.kt#L88).

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
```diff
@@ -21,13 +24,43 @@ fun SensorUi(
...
+        rememberLauncherForActivityResult(ActivityResultContracts.RequestPermission()) {}
+
+    val permissionLaunch = rememberLauncherForActivityResult(
+        ActivityResultContracts.RequestMultiplePermissions()
+    ) { isGranted ->
+        var allGranted = true
+        isGranted.forEach {
+            if (manager.requiredPermissions(basicSensor.id)
+                .contains(Manifest.permission.ACCESS_FINE_LOCATION) && manager.requiredPermissions(basicSensor.id)
+                    .contains(Manifest.permission.ACCESS_BACKGROUND_LOCATION) && it.key == Manifest.permission.ACCESS_FINE_LOCATION
```
jpelgrom, line 39: Splitting conditions with `contains` over two lines is very confusing, any reason why you did this?
```
                                                   manager.requiredPermissions(basicSensor.id)
.contains(Manifest.permission.ACCESS_BACKGROUND_LOCATION)
```

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
```diff
@@ -21,13 +24,43 @@ fun SensorUi(
...
+            )
+                backgroundRequest.launch(Manifest.permission.ACCESS_BACKGROUND_LOCATION)
```
jpelgrom, line 41: As this permission only exists on API level 29 or higher, a version check should be added [like in the main app](https://github.com/home-assistant/android/blob/93844ffea03be01e35c618114073b0e26fd47f39/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt#L345-L350).

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
```diff
@@ -21,13 +24,43 @@ fun SensorUi(
     basicSensor: SensorManager.BasicSensor,
     onSensorClicked: (String, Boolean) -> Unit,
 ) {
+    var checked = sensor?.enabled == true
+
+    val backgroundRequest =
+        rememberLauncherForActivityResult(ActivityResultContracts.RequestPermission()) {}
```
jpelgrom, line 30: In the callback here the sensor that was being toggled should be enabled if all required permissions are granted.

jpelgrom (commented):
Enabling the network sensors now works! (The state remains Unknown but that is probably due to the emulator, not permissions.)

> > It's getting very close to having too much logic in the UI layer.
>
> I agree, its a bit difficult to do the permissions properly without refactoring things a bit I think.

So why not refactor things a bit, instead of the current code? 😉

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
dshokouhi: I think this got added when I was discovered the HA core issue, will remove it

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
dshokouhi: > Splitting conditions with `contains` over two lines is very confusing, any reason why you did this?

yea android studio split it up like that and for some reason I kept it lol

dshokouhi:
> (The state remains Unknown but that is probably due to the emulator, not permissions.)

yea to get the state to change and properly update is not easy in teh emulator, on my watch I had BT connected then turned wifi off and it finally sent the first update. Definitely seems to be a HA core issue based on what I had filed.

> So why not refactor things a bit, instead of the current code? 😉

I don't think at this point there is too much logic here, and in terms of permissions this might be about it as most of the special permissions on the phone side are not there on Wear OS like Usage Stats.

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
dshokouhi: This call back is really only used for background permissions as its the only one where we remove and request later so there isn't a need to check all other permissions since it will be the last one.

jpelgrom: But right now it means that the sensor won't be enabled after granting the required permission (because the callback isn't used), you have to tap it again to enable it.

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
```diff
@@ -21,13 +25,46 @@ fun SensorUi(
...
+        ActivityResultContracts.RequestMultiplePermissions()
+    ) { isGranted ->
+        var allGranted = true
+        isGranted.forEach {
+            if (
+                manager.requiredPermissions(basicSensor.id).contains(Manifest.permission.ACCESS_FINE_LOCATION) &&
+                manager.requiredPermissions(basicSensor.id).contains(Manifest.permission.ACCESS_BACKGROUND_LOCATION) &&
+                it.key == Manifest.permission.ACCESS_FINE_LOCATION && Build.VERSION.SDK_INT >= Build.VERSION_CODES.R
+            )
+                backgroundRequest.launch(Manifest.permission.ACCESS_BACKGROUND_LOCATION)
```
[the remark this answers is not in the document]
dshokouhi: Its required to be in the loop because on certain devices anytime fine location is requested we also request background location and we need to detect if we are ready for background permissions after fine location was granted.

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
```diff
@@ -21,13 +25,46 @@ fun SensorUi(
...
+            if (!it.value)
+                allGranted = false
+        }
+        checked = sensor?.enabled == true && allGranted
```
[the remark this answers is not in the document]
dshokouhi: Ah good catch, should be fixed in the latest commit :) thanks for testing!

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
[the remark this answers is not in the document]
jpelgrom: All permissions are already granted when the callback is called, it isn't required to loop over them 😉

Now that the return is added it is fine, you'd still need a loop to check if everything is granted.

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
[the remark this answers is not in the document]
jpelgrom: Now you don't need to update `checked` any more because it will recompose when the sensor is enabled (and `checked` can be changed to `val`)

jpelgrom (approved):
Great work, hopefully this will make it easier to share more sensors soon!

JBassett (approved):

PR #2957, opened 2022-10-07
Add pixel watch companion to list of Wear OS apps

dshokouhi:
## Summary

First PR for the Pixel Watch! 😂 

https://play.google.com/store/apps/details?id=com.google.android.apps.wear.companion&hl=en_US&gl=US

## Link to pull request in Documentation repository

Documentation: home-assistant/companion.home-assistant#

PR #2958, opened 2022-10-07
Don't use a hardcoded list of apps for Wear OS detection

jpelgrom:
## Summary

As Wear OS 3 watches are going to use an unique app for each brand, the HA app should use the Google APIs for detecting watches instead of hardcoding a list of companion apps to be more future-proof.

I have only tested this with an emulator (and without an emulator, for when there is no watch) which works well, but I'd appreciate it if anyone with a real watch could test it as well.

## Screenshots

n/a

n/a

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearDetection.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearDetection.kt
@@ -1,0 +1,25 @@
+package io.homeassistant.companion.android.settings.wear
+
+import android.content.Context
+import android.util.Log
+import com.google.android.gms.wearable.Wearable
+import kotlinx.coroutines.tasks.await
+
+object SettingsWearDetection {
+
+    private const val TAG = "SettingsWearDetection"
+
+    /**
+     * Returns if there are any Wear OS devices connected to this device. It does **not** indicate
+     * if they have the Home Assistant app installed.
+     */
+    suspend fun hasAnyNodes(context: Context): Boolean {
+        return try {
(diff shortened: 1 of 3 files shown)


jpelgrom:
I know #2957 was just merged, but whatever the next brand with a Wear OS 3 watch is will benefit from this :)

dshokouhi:
so with this change, the app will no longer notify a user that the watch is not connected. Personally I think that is ok but some users may find it odd that the setting is now missing completely when their watch is dead or not connected.

https://github.com/home-assistant/android/blob/a711c659a4c331b04580a420488342dac4fcf095/app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearActivity.kt#L164-L168

If only 1 app was used this would work but since there will be others in the future it will be difficult to handle this case

dshokouhi (approved):
Works for me with my Fossil Gen 5, also this PR made me realize we need to update the messaging on the Wear OS settings landing page as beta is no longer required to use the app. We can do that in a future PR :)

jpelgrom:
> so with this change, the app will no longer notify a user that the watch is not connected.

I didn't consider that (the emulator always shows up as connected 😅), but as far as I'm aware there is no function to get paired watches. The current approach also isn't perfect because if you install a companion app the settings item will show up even if you don't have a watch.

dshokouhi:
yea admittedly the old approach was taken from a [sample](https://github.com/android/wear-os-samples/tree/main/WearVerifyRemoteApp)😂  and I thought it was nice to inform the user that while they "might" have a wear OS device based on the installed app its not connected and therefore cannot be controlled.

JBassett (approved):

PR #2961, opened 2022-10-09
Add On Body sensor for Wear OS

dshokouhi:
## Summary

A new hardware sensor thats specific to Wear OS. Unlike the other hardware sensors for this one I opted to keep the listener registered all the time so we can send updates as and when they happen. I figure out of all the sensors this one is most important to keep up to date.

https://developer.android.com/reference/android/hardware/SensorEvent#sensor.type_low_latency_offbody_detect: (may need to scroll down to the Values section the copied link seems odd)

https://developer.android.com/reference/android/hardware/Sensor#TYPE_LOW_LATENCY_OFFBODY_DETECT

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#836

## Any other notes

I pulled in the sensor manager display fix from #2956 so this one can be ready to merge as well

I am not 100% sold on the name but also calling the sensor "on wrist" doesn't make that much sense since the sensor can be placed against an object and it may think its detected. On Body felt the most appropriate given the type of sensor.

### DIFF
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -931,4 +931,6 @@
     <string name="sensor_name_display_sensors">Display Sensors</string>
     <string name="sensor_name_screen_off_timeout">Screen Off Timeout</string>
     <string name="sensor_description_screen_off_timeout">The current duration of idle time before the screen turns off in milliseconds</string>
+    <string name="sensor_name_on_body">On Body Sensor</string>
+    <string name="sensor_description_on_body">Indicates whether the device is on body</string>
 </resources>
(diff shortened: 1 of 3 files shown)


jpelgrom:
I'm surprised this doesn't require any additional permissions...

The class looks good but I don't see a way to verify it works in the emulator, so I'll leave this one for someone else to review.

dshokouhi:
> I'm surprised this doesn't require any additional permissions...

yea I think only steps and heartrate have special permissions, personally I was surprised I just found out about this sensor and that it worked. I was hopeful it would work for in pocket detection for the phone but my pixel 6 pro did not have this sensor so I assume none of the phones do.

JBassett (approved):

PR #2962, opened 2022-10-11
Convert Play tasks to coroutines

NotWoods:
## Summary
Small change to use Kotlin coroutines in place of some callbacks from Play Services.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -144,7 +144,7 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-reflect:1.6.21")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.4")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.6.4")
-    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-play-services:1.6.4")
+    "fullImplementation"("org.jetbrains.kotlinx:kotlinx-coroutines-play-services:1.6.4")
 
     implementation("com.google.dagger:hilt-android:2.44")
     kapt("com.google.dagger:hilt-android-compiler:2.44")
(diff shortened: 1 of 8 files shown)


app/build.gradle.kts
```diff
@@ -144,6 +144,7 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-reflect:1.6.21")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.4")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.6.4")
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-play-services:1.6.4")
```
jpelgrom, line 147: This will introduce a [dependency on the closed source `com.google.android.gms:play-services-tasks`](https://github.com/Kotlin/kotlinx.coroutines/blob/master/integration/kotlinx-coroutines-play-services/build.gradle.kts#L11) library in all versions, including minimal.

```suggestion
    "fullImplementation"("org.jetbrains.kotlinx:kotlinx-coroutines-play-services:1.6.4")
```

jpelgrom (changes requested):

wear/build.gradle.kts
```diff
@@ -93,6 +93,7 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.6.21")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.6.4")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-guava:1.6.4")
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-play-services:1.6.4")
```
jpelgrom, line 96: In the latest commit you removed the use of `await()` for the phone listener task, which means this dependency is now unused. Why was it removed / shouldn't this dependency be removed as well now?

NotWoods: It was removed since I didn't see a fullImplementation equivalent in this module (although I have no idea if a no-closed-source app is possible on Wear OS). I forgot to remove it here

jpelgrom: There is no equivalent for the Wear module, as Wear without Google services doesn't exist. In other words, no need to remove it for Wear.

jpelgrom (approved):
Thanks for the coroutine improvements (and other code quality changes that weren't mentioned)!

JBassett (approved):

PR #2964, opened 2022-10-11
Fix permission requests for beacon monitor

dshokouhi:
## Summary

Noticed that we were not requesting permissions properly based on the android version for beacon monitor, so correcting it so the sensor can be enabled on older versions of android.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
@@ -148,6 +148,14 @@ class BluetoothSensorManager : SensorManager {
                     Manifest.permission.BLUETOOTH_CONNECT
                 )
             }
+            (sensorId == beaconMonitor.id && Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) -> {
+                arrayOf(
+                    Manifest.permission.BLUETOOTH,
+                    Manifest.permission.BLUETOOTH_ADMIN,
+                    Manifest.permission.ACCESS_FINE_LOCATION,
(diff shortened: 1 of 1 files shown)


dshokouhi:
> * `Manifest.permission.BLUETOOTH` shouldn't be required on Android 12 and above (>= S), as it has been replaced by the new more specific permissions that are listed

Good call, still works without this permission.



> * Surely this also fixes background scanning for newer devices, not just older devices?

yup I realized it wasn't right when I compared it to LocationSensorManager so needed to make these adjustments :)

dshokouhi:
ah found a few more places to update based on the docs...will run some tests on my end and make another commit

jpelgrom:
> ah found a few more places to update based on the docs...will run some tests on my end and make another commit

If this is about removing `Manifest.permission.BLUETOOTH` (and `Manifest.permission.BLUETOOTH_ADMIN`) everywhere when using Android 12 or newer, also don't forget to add the `maxSdkVersion` to the manifest.

dshokouhi:
> > ah found a few more places to update based on the docs...will run some tests on my end and make another commit
> 
> If this is about removing `Manifest.permission.BLUETOOTH` (and `Manifest.permission.BLUETOOTH_ADMIN`) everywhere when using Android 12 or newer, also don't forget to add the `maxSdkVersion` to the manifest.

yup was getting that in there too :) this should now cover the permissions

jpelgrom (approved):
The beacon monitor permissions on <Q make sense. 

Also tested on Android 13 for the removal of 'old' Bluetooth permissions and seems to work correctly.

JBassett (approved):

blakadder:
Tested on 8.1, works fine now!

PR #2965, opened 2022-10-12
Improve settings when using Home Assistant Cloud

jpelgrom:
## Summary

This PR aims to improve the app's connection settings when using Home Assistant Cloud by:
- **refreshing Home Assistant Cloud information after registration**; currently information is only retrieved during registration but the `get_config` webhook provides up-to-date information about the cloudhook and remote UI URL, which means the app can now detect if Cloud is added after registration and switch to it without requiring the user to manually copy and paste URLs
- **hiding technical details**; the user doesn't care about their Nabu Casa URL, so hide it behind a simple switch instead to better match how it is (not) presented elsewhere in the frontend UI
- **making it easy to switch it off and go back to local access**; Home Assistant Cloud information is stored separately instead of overwriting the normal URL, so disabling it by choice in the settings or when the instance tells us that there is no longer a subscription is seamless and doesn't require setup or looking up IP addresses

## Screenshots

|     |Light|Dark|
|-----|-----|-----|
|The main settings screen now shows 'Use Home Assistant Cloud' when using cloud for the connection|App connection settings with an item "Home Assistant URL: Use Home Assistant Cloud", light mode.|App connection settings with an item "Home Assistant URL: Use Home Assistant Cloud", dark mode.|
|Using Home Assistant Cloud|A settings screen titled "Home Assistant URL", with a toggle "Use Home Assistant Cloud" on, light mode.|A settings screen titled "Home Assistant URL", with a toggle "Use Home Assistant Cloud" on, dark mode.|
|Not using Home Assistant Cloud|A settings screen titled "Home Assistant URL", with a toggle "Use Home Assistant Cloud" off and a text input with an IP address below it, light mode.|A settings screen titled "Home Assistant URL", with a toggle "Use Home Assistant Cloud" off and a text input with an IP address below it, dark mode.|
|No Home Assistant Cloud on this instance|A settings screen titled "Home Assistant URL", with only a text input with an IP address, light mode.|A settings screen titled "Home Assistant URL", with only a text input with an IP address, dark mode.|
|Input validation|A settings screen titled "Home Assistant URL", with a toggle "Use Home Assistant Cloud" off and a text input with an IP address below it, highlighted in red with the error text "Unable to parse your URL. It should look like this: https://example.com.", light mode.|A settings screen titled "Home Assistant URL", with a toggle "Use Home Assistant Cloud" off and a text input with an IP address below it, highlighted in red with the error text "Unable to parse your URL. It should look like this: https://example.com.", dark mode.|

n/a

## Any other notes

I originally intended to change up more of the connection settings to unify and simplify it, but in the end decided against it because the home network SSIDs are also used independently in other places (websocket, app lock). This partly explains why there is so much (UI) code for relatively small changes.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
@@ -608,7 +608,7 @@ class SettingsFragment constructor(
         super.onResume()
         activity?.title = getString(commonR.string.companion_app)
 
-        presenter.updateInternalUrlStatus()
         presenter.updateExternalUrlStatus()
+        presenter.updateInternalUrlStatus()
     }
 }
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
@@ -43,6 +43,7 @@ import io.homeassistant.companion.android.settings.sensor.SensorSettingsFragment
 import io.homeassistant.companion.android.settings.sensor.SensorUpdateFrequencyFragment
 import io.homeassistant.companion.android.settings.shortcuts.ManageShortcutsSettingsFragment
 import io.homeassistant.companion.android.settings.ssid.SsidFragment
+import io.homeassistant.companion.android.settings.url.ExternalUrlFragment
 import io.homeassistant.companion.android.settings.wear.SettingsWearActivity
 import io.homeassistant.companion.android.settings.websocket.WebsocketSettingFragment
 import io.homeassistant.companion.android.settings.widgets.ManageWidgetsSettingsFragment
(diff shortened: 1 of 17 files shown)


dshokouhi:
Is the long term plan for the new connection page to be used for more stuff? Right now it seems like a lot of empty space.

I do agree that Home WiFi SSID has become more than what it was originally intended for.  With that said it still kinda makes sense to keep it all in the same place when we consider thing like multi server support where the Home Wifi SSID may differ from instance to instance. I know we don't have multi server support yet but if you were looking for a better reason to combine them I think thats it 😛

jpelgrom:
> Is the long term plan for the new connection page to be used for more stuff? Right now it seems like a lot of empty space.

Yeah it's a lot of empty space, even more so when you have a larger phone or don't have Cloud. At the same time, I don't want to add another toggle to the main list (which is already quite long, and for which it might not be clear which setting it affects), and I don't think putting it in a dialog 'works'.

So originally in addition to the Cloud stuff I had other ideas:

 - URL status: does it work / which url is used right now / "changing this URL won't affect your current connection"
 - Merging all the connection settings to one screen (maybe including persistent connection, which currently isn't in the connection category?), which combined with the above makes it easier to troubleshoot and keeps settings in one place

However, I couldn't figure out a good way to do this without making it more complicated for new users or users with a simple setup, and as previously mentioned the home WiFi stuff is also used elsewhere. I do think the Cloud stuff is an improvement though which is why I submitted this PR as it is.

At the moment I don't have a specific plan for more stuff but it can happen in the future. If someone has a good idea please share it.

> (...) With that said it still kinda makes sense to keep it all in the same place when we consider thing like multi server support where the Home Wifi SSID may differ from instance to instance. I know we don't have multi server support yet but if you were looking for a better reason to combine them I think thats it 😛

For multi server we definitely need to treat SSIDs as a server-specific setting. However it still isn't exclusively used for the internal connection URL and I don't want to change it now and then change it again when multi server support is added because there's something I haven't considered now.

dshokouhi (approved):
Looks good to me

PR #2969, opened 2022-10-16
Support zone name only location tracking

jpelgrom:
## Summary

Add a setting to the background location tracking that allows for sending only the zone name or `not_home` for the current location, instead of the exact GPS location. For existing users sent data doesn't change after updating (setting will be set to exact by default). This feature was implemented in the iOS app earlier this year.

## Screenshots

|     |Light|Dark|
|-----|-----|-----|
|A new setting for the background location 'sensor' and updated description that mentions this setting|Background Location settings with a new option titled 'Location Sent', current value 'Zone name only', light mode|Background Location settings with a new option titled 'Location Sent', current value 'Zone name only', dark mode|
|Setting options|A dialog showing two options for the 'Location Sent' setting: 'Exact' and 'Zone name only' (selected), light mode|A dialog showing two options for the 'Location Sent' setting: 'Exact' and 'Zone name only' (selected), dark mode|

On the server `person` entities and `device_tracker`s will remove most attributes when a zone-only location update is received, and for example will no longer display a map in the frontend.

Frontend dev tools showing a person entity with a state of 'home' but no GPS attributes
Frontend person card showing only a state but no map

Documentation: home-assistant/companion.home-assistant#841


app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -540,6 +552,28 @@ class LocationSensorManager : LocationSensorManagerBase() {
         ).toBoolean()
     }
 
+    private suspend fun setupSendLocationAsSetting() {
+        val supported = integrationUseCase.isHomeAssistantVersionAtLeast(2022, 2, 0)
```
dshokouhi, line 556: this release was 8 months ago, do we really need to check this still?

jpelgrom: Why not? It prevents potential reports of it not working and isn't that much effort as we need to get/create the setting anyway.

dshokouhi: Ok do you think its better to hide the setting in this case then? Also we should probably update the sensor description to mention this new setting too.

jpelgrom: I don't think we can hide settings - do you mean not creating the setting when on <2022.2?

dshokouhi: yea thats what I meant 😂

jpelgrom: Updated setting to only be added when on >= 2022.2, and added a short description of the setting (+ now the setting name actually matches the documentation).

Like you mentioned most users will probably be on a supported version, so I think it is fine that the description always mentions the version even though a few users don't have the option.

JBassett (approved):

Issue #2709, opened 2022-07-24
Update Material Design Icons to 7.0.96


jpelgrom:
**Is your feature request related to a problem? Please describe.**

On Discord there was [a message](https://discord.com/channels/330944238910963714/377667559303938050/9996141964361892340) about Material Design Icons 7.0.96, which will be included in the next frontend release and includes [breaking changes](https://dev.materialdesignicons.com/changelog#version-7.0.96). The app should be updated to the latest version to handle these changes and provide a good experience.

**Describe the solution you'd like**

Update the Material Design Icons included in the app to version 7.0.96 or newer.

One icon used by the app should be renamed to `mdi:microphone-message`:
https://github.com/home-assistant/android/blob/f9ceeb0d651074675ef833f23040af8e0a53aca9/app/src/main/java/io/homeassistant/companion/android/sensors/AudioSensorManager.kt#L221

MDI icons are used and provided by two libraries:

1. [Android-Iconics](https://github.com/mikepenz/Android-Iconics), currently on 6.4.95 and no update available. Main + Wear app. <br> It shouldn't be too hard to create a PR here and update it, [@leroyboerefijn suggested he might do it](https://discord.com/channels/330944238910963714/377667559303938050/999753602224623716).

2. [Icon Picker Dialog](https://github.com/maltaisn/icondialoglib), currently on 5.3.45 and no update available - library hasn't been updated in ~2 years. Main app only. <br> Updating it is difficult because it depends on [this file](https://github.com/PeterShaggyNoble/MDI-Sandbox/blob/master/json/icons.json) which also is out of date, and there doesn't appear to be any activity. @goyney, a MDI contributor, offered to look into it [and said](https://discord.com/channels/330944238910963714/377667559303938050/1000565719991332905):

> Okay, so we have no connection to either of those libraries, so unfortunately, I have no pull to get them updated. I can tell you though, that they should have never been using that PeterShaggyNoble link. His work wasn't anything official and he had personal things come up that has severely limit his time to help out now or in the future.
>
> My suggestion would be to either fork or create a new library and use the official libraries we provide, specifically @mdi/svg. It contains meta.json that will always be up-to-date with out latest releases: https://github.com/Templarian/MaterialDesign-SVG/blob/master/meta.json
> 
> It can be installed and version controlled via NPM.

**Describe alternatives you've considered, if any**

No realistic alternatives

**Additional context**

[...]

PR #2970, opened 2022-10-17
Add icon meta file

NotWoods:
## Summary
Adds code to start migrating off the icon dialog library, which is out of date. This change adds a mapping from the lib's integer IDs to regular MDI icon names.

I generated the meta file in [NotWoods/home-assistant-android-icon-map](https://github.com/NotWoods/home-assistant-android-icon-map)

## Screenshots
N/A

N/A

## Any other notes
Related to #2709


jpelgrom:
Thanks for your work so far!

I see a lot of unused data in the json file, if not needed data should be removed to limit the binary size. Do you intend to use this for something else in the future?

If so, presumably this file will need to be updated in the future with new MDI releases? Would it make sense to also include the code to generate the file in this repository?

If not, there is a lot of data that could be removed as we only really need the icon dialog lib ID and MDI name. To remove the need to load this data each time, maybe even migrate the database data to store the MDI name instead?

NotWoods:
I'm planning to use the aliases and tags for the icon dialog. I wasn't able to see a clear way to run migrations using asynchronous code, but I agree that it makes sense to store the icon name instead. Happy to include the generating code inside this repo but wasn't sure what language to use.

jpelgrom:
> I wasn't able to see a clear way to run migrations using asynchronous code

I don't think that is possible, as the database schema can't be changed while the database is open. Room will actively throw errors if the database doesn't match the schema. Most users don't have any widgets/tiles setup or only a few, so I think the impact of doing this synchronously (= on app launch, which is when Room performs migrations) is acceptable.

> Happy to include the generating code inside this repo but wasn't sure what language to use.

Most developers will use Android Studio / IntelliJ IDEA to develop and build the app so if you can make it easy to run from there that would be nice. Your existing Javascript code also looks good to me if that is what you prefer, but maybe you could add a little more info to the README explaining how to setup the environment to run it for those not familiar with JS?

NotWoods:
Replacing with #3016. I'll include a python script for generation later, since its used with the main repo

jpelgrom:
I know it's still draft but did you drop the tags/alias filtering idea? The JSON file in there is a lot simpler / just a map for conversion, not metadata.

NotWoods:
I dropped it for now because of the size of the PR; I'd rather get this ready first and add additional search filters later. I think the metadata will end up in a separate file so it can be loaded independently of the migration code

PR #2980, opened 2022-10-21
Improve monochrome icon

Dubzer:
## Summary
I noticed that the Home Assistant monochrome app icon looked weird compared to the others. So, I re-generated it with Android Studio to better match the guideline grid.

## Screenshots

before

after


Dubzer:
I'm not sure about the documentation. Is it required here?

jpelgrom:
I agree it is good to better match other app's icon sizes, and it looks better in your screenshot, but it isn't working for me when building with the branch you supplied. The new drawables seem to be exactly the same as the existing foreground icon drawable?

Also does `ic_launcher_monochrome.xml` need to be duplicated?

> I'm not sure about the documentation. Is it required here?

No, a change to the documentation wouldn't be required for this PR.

dshokouhi:
I think the debug app icon should still be different than the production app icon, if we keep them the same it will be hard to distinguish between the 2 when themed icons are used on the device.

jpelgrom:
> I think the debug app icon should still be different than the production app icon

I think they're currently also the same when using themed icons? It seems to be limited to the home screen, so not impossible to tell them apart (still using the default color in the app drawer, overview, ...).

Dubzer:
> The new drawables seem to be exactly the same as the existing foreground icon drawable?

This was not intended. Fixed!

Also, I removed debug version of the icon.

jpelgrom:
Tested again and now the icon appears to match the screenshot.

@dshokouhi Do you feel like there should be a different monochrome icon for the debug builds? Not sure how to interpret your 👍 on my last comment.

dshokouhi:
> @dshokouhi Do you feel like there should be a different monochrome icon for the debug builds? Not sure how to interpret your 👍 on my last comment.

My 👍 was in response to the icon already being used. I didn't realize that. Also you are correct that they will still be distinguishable there.

jpelgrom (approved):

JBassett (approved):

Issue #2208, opened 2022-01-28
Add Wear OS next_alarm sensor


deliantwo:
**Is your feature request related to a problem? Please describe.**

I would like to use the next_alarm sensor in some automation. Since I use the alarm on my smartwatch there's no alarm on my phone.

**Describe the solution you'd like**

Add the next_alarm sensor on the wear app

**Describe alternatives you've considered, if any**

I've heard you can do something with tasker but I would like a first-party solution

**Additional context**

dshokouhi:
This will come sometime after #2033 has been finished and merged. After that we will be reviewing all phone app sensors and adding the ones which work. We have to get a UI for enabling sensors and adding sensor settings before we can add this one.

janstadt:
Any updates on new sensors being added for wearos devices @dshokouhi now that #2033 has been merged? Would love to get the geo_location sensor on the GW4 going. :-)

dshokouhi:
For this sensor I think it may be ok for initial release not to have sensor settings because on a wearable an allow list will not be a good experience. Hopefully it won't result in too much unwanted data but we will need to see how it fairs for users.

Issue #2308, opened 2022-02-22
Wear OS - Sensors - Add sleep and/or bedtime mode


Ulrar:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Two new sensors : 
- Bedtime mode enabled : true / false
- User asleep : true / false

Samsung's GW4 is very good at telling precisely when I fall asleep and wake up, and the latest version supposedly can trigger smart things automations based on that.
I don't know if there's any way to get that information out of their services without using smart things but it'd be a very handy sensor.

A sensor on whether or not bedtime mode is enabled might be easier to implement, I don't know if that's a Samsung only feature or not but I could see automations trigger when I enable or disable it. Maybe turn off all the lights and enable alarms when enabling / going to sleep then disable the alarm when disabling it / waking up, or stuff like that.

**Describe alternatives you've considered, if any**
None

**Additional context**
None

dshokouhi:
Are these sensors documented on the official Wear OS docs or is this a samsung specific feature?

Ulrar:
I don't expect the sleep status to be, I believe that's a samsung health feature and I don't know if it's even possible to access. I certainly wouldn't where to look for that.

For the bedtime mode again I don't know if that's a samsung feature or a wear os one, but AutoWear seems to be able to trigger based on it so presumably there is a way to get that status, see [this reddit thread](https://www.reddit.com/r/tasker/comments/ptx4e0/galaxy_watch_4_gw4_turn_onoff_bedtime_mode_on/).

dshokouhi:
> For the bedtime mode again I don't know if that's a samsung feature or a wear os one, but AutoWear seems to be able to trigger based on it so presumably there is a way to get that status, see [this reddit thread](https://www.reddit.com/r/tasker/comments/ptx4e0/galaxy_watch_4_gw4_turn_onoff_bedtime_mode_on/).

autowear mimics end user behavior to do what it needs to do, the app will not do that

if we cant find official developer documentation then these sensors may not get added.

Ulrar:
I know essentially nothing about this, I'm just a user so nevermind if it can't be done I suppose, I was just hoping for it.

If and when #2183 gets implemented, room presence could be a decent alternative to know when users are in the bedroom at or after sleep time. So I guess I'll wait on that one instead.

Thanks

dshokouhi:
If you have the phone app, then consider enabling the Sleep Confidence sensor. You may find it useful for this purpose :)

[...]

PR #2981, opened 2022-10-21
Add more Wear OS sensors

dshokouhi:
## Summary

Fixes: #2208 and #2308 

Adds the following new sensors specific to wear OS:

- Bedtime mode (only available on Wear OS 3)
- Theater mode

Migrates the following sensor managers from phone app:
- App Sensors
- Audio Sensors
- DND
- Next Alarm
- Last Update
- Steps sensor
- Power sensor

## Any other notes

All tested and working on phone and wear OS, leaving as a draft to not cause a conflict with #2961


wear/src/main/java/io/homeassistant/companion/android/sensors/TheaterModeSensorManager.kt
```diff
@@ -0,0 +1,64 @@
...
+        private const val TAG = "TheaterSensor"
+
+        val theaterMode = SensorManager.BasicSensor(
+            "theater_mode",
+            "binary_sensor",
+            commonR.string.sensor_name_theater_mode,
+            commonR.string.sensor_description_theater_mode,
+            "mdi:movie-open",
+            entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC,
+            updateType = SensorManager.BasicSensor.UpdateType.INTENT
```
jpelgrom, line 20: [On Discord](https://discord.com/channels/330944238910963714/377667559303938050/1033053536503726203) you mentioned that this sensor isn't linked to DND after all. Is there another intent the app is using for this? If not the `updateType` should be changed (even if this is not visible in the app yet).

dshokouhi: Ah forgot to update this bit, this one can actually update with the interactive sensor but it should have normal update type since its not direct

dshokouhi: I will also update bedtime mode to be more correct as the intents are linked to a different sensor.

jpelgrom: That would be a good comment if we get decide to add sensor descriptions on the watch

dshokouhi: ok updated the description as well for both sensors :)

JBassett (approved):

PR #2982, opened 2022-10-21
Add caching of favorites in Wear OS, resulting in prettier loading

mads03dk:
## Summary
Added local caching of the `friendly_name` and `icon` properties in the WearOS app, thus avoiding using the entity ID and a bookmark icon whilst waiting for the server to send the current states. Aimed to achieve a prettier visual while waiting for the connection to be established

## Screenshots
Before, whilst waiting for a connection:
Screenshot_20221021_194139_android

Now, whilst waiting for a connection:
Screenshot_20221021_194142_android

Settings:
image

## Any other notes
Only favorites are cached, since they're the first entities to be shown. Cached is updated every time the entity is received from the server or added/removed from the list of favorites. Some internals of the `getIcon` function has also been updated to allow direct usage of icon-strings

Please note;
This is my very first go at native app development (hereby Kotlin), so I expect some - if not all - of my changes to be suboptimal. I'm really open to suggestions and would love to get some constructive feedback, but don't go too harsh, haha!

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
@@ -43,6 +43,8 @@ import io.homeassistant.companion.android.database.wear.EntityStateComplications
 import io.homeassistant.companion.android.database.wear.EntityStateComplicationsDao
 import io.homeassistant.companion.android.database.wear.Favorites
 import io.homeassistant.companion.android.database.wear.FavoritesDao
+import io.homeassistant.companion.android.database.wear.Caches
+import io.homeassistant.companion.android.database.wear.CachesDao
 import io.homeassistant.companion.android.database.widget.ButtonWidgetDao
 import io.homeassistant.companion.android.database.widget.ButtonWidgetEntity
 import io.homeassistant.companion.android.database.widget.CameraWidgetDao
@@ -71,6 +73,7 @@ import io.homeassistant.companion.android.common.R as commonR
         NotificationItem::class,
         TileEntity::class,
         Favorites::class,
+        Caches::class,
         EntityStateComplications::class,
         Setting::class
     ],
@@ -105,6 +108,7 @@ abstract class AppDatabase : RoomDatabase() {
     abstract fun notificationDao(): NotificationDao
     abstract fun tileDao(): TileDao
     abstract fun favoritesDao(): FavoritesDao
+    abstract fun cachesDao(): CachesDao
     abstract fun entityStateComplicationsDao(): EntityStateComplicationsDao
     abstract fun settingsDao(): SettingsDao
 
@@ -153,7 +157,8 @@ abstract class AppDatabase : RoomDatabase() {
                     MIGRATION_20_21,
                     MIGRATION_21_22,
                     MIGRATION_22_23,
-                    MIGRATION_23_24
+                    MIGRATION_23_24,
+                    MIGRATION_24_25,
                 )
                 .fallbackToDestructiveMigration()
                 .build()
@@ -509,6 +514,12 @@ abstract class AppDatabase : RoomDatabase() {
         private val MIGRATION_23_24 = object : Migration(23, 24) {
             override fun migrate(database: SupportSQLiteDatabase) {
                 database.execSQL("ALTER TABLE `settings` ADD `sensorUpdateFrequency` TEXT NOT NULL DEFAULT 'NORMAL'")
+            }
+        }
+
+        private val MIGRATION_24_25 = object : Migration(24, 25) {
+            override fun migrate(database: SupportSQLiteDatabase) {
+                database.execSQL("CREATE TABLE IF NOT EXISTS `caches` (`id` TEXT PRIMARY KEY NOT NULL, `friendly_name` TEXT NOT NULL, `icon` TEXT)")
             }
         }
 
--- a/common/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
@@ -41,10 +41,10 @@ import io.homeassistant.companion.android.database.settings.Setting
 import io.homeassistant.companion.android.database.settings.SettingsDao
 import io.homeassistant.companion.android.database.wear.EntityStateComplications
 import io.homeassistant.companion.android.database.wear.EntityStateComplicationsDao
+import io.homeassistant.companion.android.database.wear.FavoriteCaches
+import io.homeassistant.companion.android.database.wear.FavoriteCachesDao
 import io.homeassistant.companion.android.database.wear.Favorites
 import io.homeassistant.companion.android.database.wear.FavoritesDao
-import io.homeassistant.companion.android.database.wear.Caches
-import io.homeassistant.companion.android.database.wear.CachesDao
 import io.homeassistant.companion.android.database.widget.ButtonWidgetDao
 import io.homeassistant.companion.android.database.widget.ButtonWidgetEntity
 import io.homeassistant.companion.android.database.widget.CameraWidgetDao
@@ -73,11 +73,11 @@ import io.homeassistant.companion.android.common.R as commonR
         NotificationItem::class,
         TileEntity::class,
         Favorites::class,
-        Caches::class,
+        FavoriteCaches::class,
         EntityStateComplications::class,
         Setting::class
     ],
-    version = 33,
+    version = 34,
     autoMigrations = [
         AutoMigration(from = 24, to = 25),
         AutoMigration(from = 25, to = 26),
@@ -88,6 +88,7 @@ import io.homeassistant.companion.android.common.R as commonR
         AutoMigration(from = 30, to = 31),
         AutoMigration(from = 31, to = 32),
         AutoMigration(from = 32, to = 33),
+        AutoMigration(from = 33, to = 34)
     ]
 )
 @TypeConverters(
@@ -108,7 +109,7 @@ abstract class AppDatabase : RoomDatabase() {
     abstract fun notificationDao(): NotificationDao
     abstract fun tileDao(): TileDao
     abstract fun favoritesDao(): FavoritesDao
-    abstract fun cachesDao(): CachesDao
+    abstract fun favoriteCachesDao(): FavoriteCachesDao
     abstract fun entityStateComplicationsDao(): EntityStateComplicationsDao
     abstract fun settingsDao(): SettingsDao
 
@@ -157,8 +158,7 @@ abstract class AppDatabase : RoomDatabase() {
                     MIGRATION_20_21,
                     MIGRATION_21_22,
                     MIGRATION_22_23,
-                    MIGRATION_23_24,
-                    MIGRATION_24_25,
+                    MIGRATION_23_24
                 )
                 .fallbackToDestructiveMigration()
                 .build()
@@ -514,12 +514,6 @@ abstract class AppDatabase : RoomDatabase() {
         private val MIGRATION_23_24 = object : Migration(23, 24) {
             override fun migrate(database: SupportSQLiteDatabase) {
                 database.execSQL("ALTER TABLE `settings` ADD `sensorUpdateFrequency` TEXT NOT NULL DEFAULT 'NORMAL'")
-            }
-        }
-
-        private val MIGRATION_24_25 = object : Migration(24, 25) {
-            override fun migrate(database: SupportSQLiteDatabase) {
-                database.execSQL("CREATE TABLE IF NOT EXISTS `caches` (`id` TEXT PRIMARY KEY NOT NULL, `friendly_name` TEXT NOT NULL, `icon` TEXT)")
             }
         }
 
--- a/common/src/main/java/io/homeassistant/companion/android/database/wear/Caches.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/database/wear/Caches.kt
@@ -1,0 +1,19 @@
+package io.homeassistant.companion.android.database.wear
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+/**
+ * Represents a favorite entity
+ */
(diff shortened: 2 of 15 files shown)


jpelgrom:
Hi 👋 This is a really nice idea

For database changes please increase the version as you're changing something for it. Right now, upgrades of existing installs won't work. When increasing the version this also means that the existing JSON files that describe the database don't change, and you can probably use an `AutoMigration` instead of writing SQL yourself.

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
```diff
@@ -31,7 +31,7 @@ fun stringForDomain(domain: String, context: Context): String? =
         )
         )[domain]?.let { context.getString(it) }
 
-fun getIcon(icon: String?, domain: String, context: Context): IIcon? {
+fun getIcon(icon: String?, domain: String, context: Context?): IIcon? {
```
jpelgrom, line 34: Why are you changing `Context` to `Context?` here? It should still always be available

common/src/main/java/io/homeassistant/companion/android/database/wear/Caches.kt
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Represents a favorite entity
+ */
+@Entity(tableName = "cache")
+data class Caches(
+    @PrimaryKey
+    @ColumnInfo(name = "id")
+    val id: String,
+    @ColumnInfo(name = "friendly_name")
+    val friendly_name: String,
```
jpelgrom, line 16:
```suggestion
    val friendlyName: String,
```

jpelgrom (commented):
In addition to the more general comment about database changes above, a few more specific notes. If anything is unclear feel free to add a comment :)

common/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
```diff
@@ -512,6 +517,12 @@ abstract class AppDatabase : RoomDatabase() {
             }
         }
 
+        private val MIGRATION_24_25 = object : Migration(24, 25) {
+            override fun migrate(database: SupportSQLiteDatabase) {
+                database.execSQL("CREATE TABLE IF NOT EXISTS `caches` (`id` TEXT PRIMARY KEY NOT NULL, `friendly_name` TEXT NOT NULL, `icon` TEXT)")
+            }
+        }
+
```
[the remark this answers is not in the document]
mads03dk: Thanks for the specifics! Implementing

common/src/main/java/io/homeassistant/companion/android/database/wear/Caches.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+/**
+ * Represents a favorite entity
+ */
+@Entity(tableName = "cache")
+data class Caches(
```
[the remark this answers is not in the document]
mads03dk: The idea was to start out with just caching favorites, as I currently don't see the need to cache any other entities. However, a general name like `cache` provides the future possibility of caching other entities as well. Let me know what you think.

The comment on line 7-9 is mere a silly mistake by me not paying attention when copying the contents of the `Favorites.kt` file, sorry for that haha, fixing as we speak :)

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
```diff
@@ -136,9 +141,14 @@ class MainViewModel @Inject constructor(
                 }
                 deviceRegistry = getDeviceRegistry.await()
                 entityRegistry = getEntityRegistry.await()
+                val allFavorites = favoritesDao.getAll()
```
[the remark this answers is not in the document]
mads03dk: Doing so, thanks!

common/src/main/java/io/homeassistant/companion/android/database/wear/Caches.kt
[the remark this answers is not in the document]
jpelgrom: > However, a general name like cache provides the future possibility of caching other entities as well. 

Correct, but this data class is specifically for favorited entities. If we want to cache something else in the future, other data/properties will have to be stored so probably a new data class would be used instead of adding it to this existing data class.

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
mads03dk: Oops! Something I forgot to remove when trying out different methods. Fixing.

wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt
```diff
@@ -149,8 +150,10 @@ fun LoadHomePage(
                 ) { entityId, isSelected ->
                     if (isSelected) {
                         mainViewModel.addFavoriteEntity(entityId)
+                        mainViewModel.addCachedEntity(entityId)
```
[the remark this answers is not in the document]
mads03dk: Implementing

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
```diff
@@ -43,11 +43,10 @@ fun getIcon(icon: String?, domain: String, context: Context): IIcon? {
     return getIcon(simpleEntity as Entity<Map<String, Any>>, domain, context)
 }
 
-fun getIcon(entity: Entity<Map<String, Any>>?, domain: String, context: Context): IIcon? {
+fun getIcon(entity: Entity<Map<String, Any>>?, domain: String, context: Context?): IIcon? {
```
jpelgrom, line 46: Same question as for the other function, why are you changing `Context` to `Context?` here? `LocalContext.current`,  which is what is passed in, should always be non-null.

common/src/main/java/io/homeassistant/companion/android/database/wear/FavoriteCachesDao.kt
```diff
@@ -0,0 +1,22 @@
...
+import androidx.room.Dao
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+
+@Dao
+interface FavoriteCachesDao {
+
+    @Query("SELECT * FROM favorite_cache where id = :id LIMIT 1")
+    fun get(id: String): FavoriteCaches
```
jpelgrom, line 12: There might not be a favorite for the provided ID

```suggestion
    fun get(id: String): FavoriteCaches?
```

common/src/main/java/io/homeassistant/companion/android/database/wear/FavoriteCachesDao.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import androidx.room.Dao
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+
+@Dao
+interface FavoriteCachesDao {
+
+    @Query("SELECT * FROM favorite_cache where id = :id LIMIT 1")
```
jpelgrom, line 11: The `LIMIT 1` isn't necessary as ID is the primary key and unique

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
```diff
@@ -358,17 +366,46 @@ class MainViewModel @Inject constructor(
...
 
     fun removeFavoriteEntity(entityId: String) {
         viewModelScope.launch {
             favoritesDao.delete(entityId)
+            removeCachedFavorite(entityId)
+        }
+    }
+
+    fun getCachedEntity(entityId: String): FavoriteCaches =
+        FavoriteCachesDao.get(entityId)
```
jpelgrom, line 381: 1. This might return `null` if there is no favorite.

2. This is called directly from the composable function, so please implement some form of caching so that the app isn't constantly reading from the database. For example implement a `getAll()` function in the DAO that is used when the viewmodel is initialized, and return values from that list.

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
```diff
@@ -31,6 +33,7 @@ import javax.inject.Inject
 @HiltViewModel
 class MainViewModel @Inject constructor(
     private val favoritesDao: FavoritesDao,
+    private val FavoriteCachesDao: FavoriteCachesDao,
```
jpelgrom, line 36: Variable names usually start with a lowercase letter :)

```suggestion
    private val favoriteCachesDao: FavoriteCachesDao,
```

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
mads03dk: I seem to have misunderstood the use of `?` in Kotlin, a beginner mistake, I assume. Fixing.

wear/src/main/java/io/homeassistant/companion/android/util/CommonFunctions.kt
```diff
@@ -143,6 +142,14 @@ fun getIcon(entity: Entity<Map<String, Any>>?, domain: String, context: Context)
     }
 }
 
+fun getIconFromMdiString(icon: String, context: Context?): IIcon {
```
jpelgrom, line 145: Due to the latest changes `Context?` here can now also be `Context`, and lines 147-149 can be removed.

mads03dk: Aahh, got it! I have much better understanding of null assertions in Kotlin now, after all of these changes.
Many thanks for the tyour time and help :)

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
```diff
@@ -358,17 +367,46 @@ class MainViewModel @Inject constructor(
...
         viewModelScope.launch {
             favoritesDao.addToEnd(entityId)
+            addCachedFavorite(entityId)
         }
     }
 
     fun removeFavoriteEntity(entityId: String) {
         viewModelScope.launch {
             favoritesDao.delete(entityId)
+            removeCachedFavorite(entityId)
```
jpelgrom, line 377: This function is only used here and doesn't do anything special, so you could call `favoriteCachesDao.delete(entityId)` here to simplify things a bit.

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
```diff
@@ -358,17 +367,46 @@ class MainViewModel @Inject constructor(
...
+
+    fun getCachedEntity(entityId: String): FavoriteCaches? =
+        favoriteCaches.find { it.id == entityId }
+
+    private fun addCachedFavorite(entityId: String) {
+        viewModelScope.launch {
+            val entity = entities[entityId]
+            val attributes = entity?.attributes as Map<*, *>
+            val icon = attributes["icon"] as String?
+            favoriteCachesDao.add(FavoriteCaches(entityId, attributes["friendly_name"].toString(), icon))
```
jpelgrom, line 389: In case the friendly name isn't set nothing should be stored instead of the string `"null"`, which would be even uglier in my opinion. That will allow the app to show the entity ID instead like it does now.

(If you agree, make sure to also update the data class to reflect this.)

wear/src/main/java/io/homeassistant/companion/android/home/views/MainView.kt
```diff
@@ -89,19 +89,20 @@ fun MainView(
                         items(favoriteEntityIds.size) { index ->
                             val favoriteEntityID = favoriteEntityIds[index].split(",")[0]
                             if (mainViewModel.entities.isEmpty()) {
-                                // Use a normal chip when we don't have the state of the entity
+                                // when we don't have the state of the entity, create a Chip from cache as we don't have the state yet
+                                val cached = mainViewModel.getCachedEntity((favoriteEntityID))
```
jpelgrom, line 93: No need to use double parantheses

```suggestion
                                val cached = mainViewModel.getCachedEntity(favoriteEntityID)
```

jpelgrom:
I've been thinking, do we need a clear cache button in the settings? The app handles adding/updating/deleting on load, when the favorites are edited and on logout.

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
jpelgrom: Storing the entity ID _as_ the friendly name when there is no friendly name also works :)

jpelgrom:
> I've been thinking, do we need a clear cache button in the settings? The app handles adding/updating/deleting on load, when the favorites are edited and on logout.

@mads03dk What is your opinion on this ^ ?

jpelgrom:
Yes it also seems unnecessary to me, so feel free to remove it :)

common/src/main/res/values/strings.xml
```diff
@@ -97,6 +97,8 @@
     <string name="buttons">Buttons</string>
     <string name="button_forward">Next</string>
     <string name="button_widget_desc">Call any service</string>
+    <string name="cache">Cache</string>
+    <string name="cache_clear">Clear Cache</string>
```
jpelgrom, line 101: Don't forget to remove the strings now that the button has been removed

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
```diff
@@ -358,17 +367,41 @@ class MainViewModel @Inject constructor(
...
     }
 
     fun logout() {
         homePresenter.onLogoutClicked()
+
+        // also clear cache when logging out
+        clearCache()
+    }
+
+    fun clearCache() {
```
jpelgrom, line 401: This can be `private` now

jpelgrom (approved):
Thanks, nice first contribution!

JBassett (approved):

PR #2983, opened 2022-10-21
Change log 10/21

dshokouhi:


JBassett (approved):

jpelgrom:
The Android 13 target also means that the Bluetooth notification on/off command will stop working on devices running Android 13, which should be marked as a breaking change.

dshokouhi:
> The Android 13 target also means that the Bluetooth notification on/off command will stop working on devices running Android 13, which should be marked as a breaking change.

good catch!!! updated 🙏

PR #2984, opened 2022-10-22
Add previews to Wear OS compose and sort sensor by name

dshokouhi:
## Summary

Adds some previews for some of our Wear OS compose files, also sorts sensors by the name

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/home/views/SensorManagerUi.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/views/SensorManagerUi.kt
@@ -3,6 +3,7 @@ package io.homeassistant.companion.android.home.views
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.wear.compose.material.PositionIndicator
 import androidx.wear.compose.material.Scaffold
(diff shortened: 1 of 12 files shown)


wear/src/main/java/io/homeassistant/companion/android/home/views/SensorManagerUi.kt
```diff
@@ -28,7 +32,9 @@ fun SensorManagerUi(
             },
             timeText = { TimeText(!scalingLazyListState.isScrollInProgress) }
         ) {
-            val availableSensors = sensorManager.getAvailableSensors(LocalContext.current)
+            val availableSensors = sensorManager
+                .getAvailableSensors(LocalContext.current)
+                .sortedBy { stringResource(it.name) }
```
jpelgrom, line 37: This doesn't change so I believe it can be stored in a `by remember { ... }`

Also shouldn't the sensor _managers_ be sorted by name as well?

dshokouhi: > This doesn't change so I believe it can be stored in a `by remember { ... }`

looks like we can save it, will update



> Also shouldn't the sensor _managers_ be sorted by name as well?

its already being done [here](https://github.com/home-assistant/android/blob/7c792c4d4bada99cc35cf6378b000d82b4002df8/wear/src/main/java/io/homeassistant/companion/android/home/views/SensorsView.kt#L69)

jpelgrom:
Some of these are weird previews... Should the annotation be updated to show a more typical Wear form factor where it makes sense (`@Preview(device = Devices.WEAR_OS_LARGE_ROUND)`)?

|Now|image|image|
|---|---|---|
|`device`|image|image|

dshokouhi:
> Some of these are weird previews... Should the annotation be updated to show a more typical Wear form factor where it makes sense (`@Preview(device = Devices.WEAR_OS_LARGE_ROUND)`)?

yup I agree, will update all our Wear OS previews to be device specific :)

jpelgrom (approved):

JBassett (approved):

PR #2989, opened 2022-10-23
Revert/set force dark again for webview

### LINKED ISSUE #2985: Companion app no longer uses dark theme if companion app settings are "Dark" or "Follow System Settings" but Home Assistant is set to "Auto"
**Home Assistant Android version:** 2022.10.2-full

**Android version:** 10

**Phone model:** Pixel 3a

**Home Assistant version:** 2022.10.5

**Last working Home Assistant release (if known):** note that it's the companion app that's the problem, and the last known working version of that is 2022.10.0-full

**Description of problem:** 
In Android Companion App > Settings > Companion App > Theme I choose either "Dark" or "Follow System Settings" (which should be equivalent to "Dark" on my system). While the companion app settings screen stays in dark mode, the rest of Home Assistant within the app is light. (Home Assistant's theme itself is set to "Auto" in "My User" > Theme.)

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):** N/A

```

```

**Screenshot of problem:**
The settings page is dark, which is correct:
settings page is dark

...but hitting the back button and going to the rest of Home Assistant in the companion app ignores the chosen "Dark" theme.
but the rest of home assistant is light

**Additional information:**
[PR 2923](https://github.com/home-assistant/android/pull/2923) is almost certainly the culprit, and I'm guessing it has something to do with "Removing webview force dark settings"

jpelgrom:
## Summary

Revert some of the changes in #2923 in an attempt to fix #2985. The WebView doesn't use the app's light/dark theme light as expected because users are running a webview version which works differently or are running an outdated/unupdateable version (which IMO is the bigger issue here, but who am I to judge).

Documentation and logs suggest this won't work anymore but as the affected devices don't work correctly in the first place it is worth a shot. There don't appear to be any downsides to adding this code aside from the log statement.

```
2022-10-23 15:51:12.299 18646-18646 cr_SupportWebSettings   io....stant.companion.android.debug  W  setForceDarkBehavior() is a no-op in an app with targetSdkVersion>=T
2022-10-23 15:51:12.300 18646-18646 cr_SupportWebSettings   io....stant.companion.android.debug  W  setForceDark() is a no-op in an app with targetSdkVersion>=T
```

Confirmed 'working' by multiple users in the issue.

## Screenshots

n/a

n/a

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
@@ -66,7 +66,7 @@ class AuthenticationFragment : Fragment() {
                 MdcTheme {
                     AndroidView({
                         WebView(requireContext()).apply {
-                            themesManager.setThemeForWebView()
+                            themesManager.setThemeForWebView(requireContext(), settings)
                             settings.javaScriptEnabled = true
                             settings.domStorageEnabled = true
                             settings.userAgentString = settings.userAgentString + " ${HomeAssistantApis.USER_AGENT_STRING}"
--- a/app/src/main/java/io/homeassistant/companion/android/themes/ThemesManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/themes/ThemesManager.kt
@@ -1,7 +1,12 @@
 package io.homeassistant.companion.android.themes
 
+import android.content.Context
+import android.content.res.Configuration
 import android.os.Build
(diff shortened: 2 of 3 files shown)


dshokouhi (approved):
This is a good fix, we can't account for some of the android forks where the user cannot update WebView independently. I also see some users confirmed it is working.

emansom:
This broke setting dark mode through automations for me on Android 9.

PR #2991, opened 2022-10-23
Remove dependency that caused beta to fail to publish.

JBassett:
## Summary

The removed dependecy was causing us to declare the `com.google.android.gms.permission.AD_ID` permission.  Since we don't actually use the permission and you need to manually declare it if used we are removing it.

## Screenshots

N/A

N/A

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -163,7 +163,6 @@ dependencies {
 
     "fullImplementation"("com.google.android.gms:play-services-location:20.0.0")
     "fullImplementation"(platform("com.google.firebase:firebase-bom:30.4.1"))
-    "fullImplementation"("com.google.firebase:firebase-analytics")
     "fullImplementation"("com.google.firebase:firebase-messaging")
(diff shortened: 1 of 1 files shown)


jpelgrom:
Won't this break the rate limit details (successful/errors)? Google says analytics is required for [delivery reporting](https://firebase.google.com/docs/cloud-messaging/android/client#set_up_the_sdk).

JBassett:
> Won't this break the rate limit details (successful/errors)? Google says analytics is required for [delivery reporting](https://firebase.google.com/docs/cloud-messaging/android/client#set_up_the_sdk).

The successful / error numbers are handled by this endpoint `https://mobile-apps.home-assistant.io/api/checkRateLimits` and aren't part of Firebase at all.

jpelgrom:
I see now, I checked the 'server' code and it isn't using Firebase to get these numbers but simply stores them when sending. Makes you wonder why analytics weren't removed before :)

dshokouhi (approved):
Looks good to me don't have a chance to check the APK but as long as fcm still works don't think we ever needed it lol.

PR #2993, opened 2022-10-24
Add a touch lock sensor to wear os

dshokouhi:
## Summary

Adds a sensor to represent the state of the touchscreen being locked (AKA wet mode on some device).  This is a special mode where the user must press and hold the crown/power button for 2 seconds to re-enable touch. The system broadcasts a intent housing an intent which contains an extra for the state.

```
2022-10-24 14:37:41.329 10921-10921 SensorReceiver          io....stant.companion.android.debug  D  Event data: {nowisconvenient=false, timestamp=1666647461141, intent=android.os.UpdateLock.UPDATE_LOCK_CHANGED}
```

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#844

## Any other notes

Also updating the docs to point to the correct place for the other new Wear OS sensors :)

Looking in AOSP it seems we do have a global setting here however I think the intent approach is probably better as it updates faster.

https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/provider/Settings.java#17531

if we think we should change it I am ok with that, hopefully the intent itself registers on other devices 🤞

### DIFF
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -940,4 +940,6 @@
     <string name="sensor_description_bedtime_mode">Whether bedtime mode is enabled on the device. For best results enable either Do Not Disturb or Interactive sensor.</string>
     <string name="sensor_name_theater_mode">Theater Mode</string>
     <string name="sensor_description_theater_mode">Whether theater mode is enabled on the device. For best results enable the interactive sensor.</string>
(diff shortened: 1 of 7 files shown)


dshokouhi:
Marking as a draft as it seems this method triggers when in settings and other apps

dshokouhi:
closing in favor of a future PR to be submitted shortly

PR #2994, opened 2022-10-25
Add Wet Mode Sensor

SkechyWolf:
## Summary

Adds a Wet Mode sensor to Wear. This is a special mode where the user must press and hold the crown/power button for 2 seconds to re-enable touch. May also be referred as Touch Lock or Water Lock.
## Screenshots

image

Documentation: home-assistant/companion.home-assistant#845


dshokouhi (approved):
This works better, the approach I ran into wasn't correct :) Tested on my Pixel Watch and it updates as expected with touch lock being enabled/disabled.

JBassett (approved):

PR #2995, opened 2022-10-25
Add user activity sensor from health services for Wear OS

dshokouhi:
## Summary

Adds the first of [many](https://developer.android.com/reference/kotlin/androidx/health/services/client/data/DataType#public-companion-properties_1) sensors for the new Health Services API. For this first sensor I chose activity state as I know a lot of users want to know when they fall asleep so hopefully this works fast enough for them.  I ran some tests using the synthetic method Google [recommended](https://developer.android.com/training/wearables/health-services/synthetic-data#enable) and the state updates as expected. Will need to consider a good way to register and unregister for updates as a user enables or disables the various sensors they can have from this manager.

This sensor has 4 possible states: asleep, exercise, passive or unknown

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#846

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
@@ -60,12 +60,6 @@ class HealthServicesSensorManager : SensorManager {
 
     override fun requestSensorUpdate(context: Context) {
         latestContext = context
-
-        if (healthClient == null) healthClient = HealthServices.getClient(latestContext)
-        if (passiveMonitoringClient == null) passiveMonitoringClient = healthClient?.passiveMonitoringClient
-        passiveListenerConfig = PassiveListenerConfig.builder()
-            .setShouldUserActivityInfoBeRequested(isEnabled(latestContext, userActivityState.id))
-            .build()
         updateUserActivityState()
     }
 
@@ -75,6 +69,12 @@ class HealthServicesSensorManager : SensorManager {
             callBackRegistered = false
             return
         }
+
+        if (healthClient == null) healthClient = HealthServices.getClient(latestContext)
+        if (passiveMonitoringClient == null) passiveMonitoringClient = healthClient?.passiveMonitoringClient
+        passiveListenerConfig = PassiveListenerConfig.builder()
+            .setShouldUserActivityInfoBeRequested(isEnabled(latestContext, userActivityState.id))
+            .build()
 
         val passiveListenerCallback: PassiveListenerCallback = object : PassiveListenerCallback {
             override fun onUserActivityInfoReceived(info: UserActivityInfo) {
--- a/wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
@@ -1,0 +1,115 @@
+package io.homeassistant.companion.android.sensors
+
+import android.Manifest
+import android.content.Context
+import android.os.Build
+import android.util.Log
+import androidx.annotation.RequiresApi
+import androidx.health.services.client.HealthServices
+import androidx.health.services.client.HealthServicesClient
+import androidx.health.services.client.PassiveListenerCallback
+import androidx.health.services.client.PassiveMonitoringClient
+import androidx.health.services.client.data.PassiveListenerConfig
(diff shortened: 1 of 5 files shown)


dshokouhi:
Real world testing passed 😂 

image

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
```diff
@@ -0,0 +1,115 @@
...
+                    when (info.userActivityState) {
+                        UserActivityState.USER_ACTIVITY_ASLEEP -> "asleep"
+                        UserActivityState.USER_ACTIVITY_PASSIVE -> "passive"
+                        UserActivityState.USER_ACTIVITY_EXERCISE -> "exercise"
+                        else -> "unknown"
+                    },
+                    when (info.userActivityState) {
+                        UserActivityState.USER_ACTIVITY_EXERCISE -> "mdi:run"
+                        UserActivityState.USER_ACTIVITY_ASLEEP -> "mdi:sleep"
+                        UserActivityState.USER_ACTIVITY_PASSIVE -> "mdi:walk"
```
jpelgrom, line 94: Walking != passive in my opinion... Suggestion: something that doesn't imply movement like [`mdi:human-handsdown`](https://materialdesignicons.com/icon/human-handsdown)? (The documentation says this is when "[t]he user is awake but is not currently exercising", so it could be correct, but passive implies less.)

wear/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
```diff
@@ -49,6 +50,10 @@ class SensorReceiver : SensorReceiverBase() {
             StepsSensorManager(),
             TheaterModeSensorManager()
         )
+        val MANAGERS = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R)
```
jpelgrom, line 53: Any specific reason you're doing it like this? The main app also has SDK limited sensor managers but still uses one list.

jpelgrom (commented):
Should [`onPermissionLost`](https://developer.android.com/reference/kotlin/androidx/health/services/client/PassiveListenerCallback#onPermissionLost()) also be implemented in the callback to disable the sensor as soon as possible?

wear/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
dshokouhi: Lint was failing so it was either the current implementation or we skipped the lint check here. I decided against skipping the lint check but in reality we can ignore it as the code will not be reached.

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
dshokouhi: good suggestion, sometimes its not easy to find these icons lol

wear/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
jpelgrom: Missed that one! Can't that be solved by using `@RequiresApi(Build.VERSION_CODES.R)`, like for various other sensors?

> but in reality we can ignore it

We can't as it is a requirement to merge PRs 🙃

dshokouhi: > Missed that one! Can't that be solved by using `@RequiresApi(Build.VERSION_CODES.R)`, like for various other sensors?

its already there in `HealthServicesSensorManager` but it wasn't good enough. For some reason it had a different complaint here than it did for DND sensor in the common folder.



> We can't as it is a requirement to merge PRs 🙃

lol I meant to suppress lint like one of Android Studio recommendations lol

dshokouhi:
> Should [`onPermissionLost`](https://developer.android.com/reference/kotlin/androidx/health/services/client/PassiveListenerCallback#onPermissionLost()) also be implemented in the callback to disable the sensor as soon as possible?

yup will add that and also will add `onRegistrationFailed` so we can log the request and set the callback registration to false

jpelgrom (approved):
Seems to register correctly on the Wear OS 2/3 emulators. State is always "passive" which is correct for a computer ;) but confirmed to be changing and working in the comments above.

JBassett:
@dshokouhi needs a conflict fix.

JBassett (approved):

PR #3001, opened 2022-10-26
Fix global vars in bedtime and theater mode

SkechyWolf:
## Summary

Adds a check in the Theater & Bedtime mode sensors for Samsung as they have different globals for those
## Screenshots

|  image | image | 
|-----------|-----------|

Documentation: N/A


dshokouhi (approved):
Thanks for fixing the doc links!

I think its safe to assume that the galaxy watch 5 will use the same globals as the galaxy watch 4. Its also "safe" to consider the Pixel Watch as a reference device here so its good to default to it. Hopefully the other Wear OS devices use the same variables.

PR #3003, opened 2022-10-26
Add unit tests!

NotWoods:
## Summary
Sets up some infrastructure for unit testing repository classes!

## Screenshots
N/A

## Any other notes
https://developer.android.com/training/testing/local-tests
https://mockk.io/
https://developer.android.com/kotlin/coroutines/test

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -117,3 +117,31 @@ jobs:
       uses: actions/upload-artifact@v3
       with:
         path: ./**/*.apk
+
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3.12.0
+        with:
+          distribution: 'adopt'
+          java-version: '17'
+
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Mock google-services.json
+        run: |
+          cp .github/mock-google-services.json app/google-services.json
+          cp .github/mock-google-services.json wear/google-services.json
+          cp .github/mock-google-services.json automotive/google-services.json
+
+      - name: Run Tests
+        run: ./gradlew test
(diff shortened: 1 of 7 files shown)


common/src/test/java/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryTest.kt
```diff
@@ -0,0 +1,120 @@
...
+            }
+        }
+
+        override suspend fun getBoolean(key: String) = when (key) {
+            "crash_reporting" -> crashReporting
+            else -> false
+        }
+    }
+
+    private lateinit var prefsRepository: PrefsRepository
```
nick2525, line 16: It is better to initialize it here, when use lateinit

NotWoods: Since we want a fresh state in every test, I initialize in the `@Before` block instead so the initializer is run before each test rather than once before all tests.

nick2525: In my tests every val was created many times, please try it under log or debug

NotWoods: I didn't notice anything odd in my tests. `prefsRepository` should be created many times

jpelgrom:
Thanks for updating these because everything changed (and patience while the original PR sat ignored for months), but if we want to merge this and take it seriously, the CI runs for PRs should be updated to run tests.

common/src/test/java/io/homeassistant/companion/android/common/data/HomeAssistantVersionTest.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    @Test
+    fun fromDateVersionTest() {
+        assertEquals(
+            HomeAssistantVersion(2021, 10, 12),
+            HomeAssistantVersion.fromString("2021.10.12")
+        )
+        assertEquals(
+            HomeAssistantVersion(2023, 1, 1),
+            HomeAssistantVersion.fromString("2023.1.01.dev")
```
jpelgrom, line 23: More realistic if you want to simulate a dev version: `2023.1.0.dev20221231`

.github/workflows/pr.yml
```diff
@@ -117,3 +117,31 @@ jobs:
...
+            ${{ runner.os }}-gradle-
+
+      - name: Mock google-services.json
+        run: |
+          cp .github/mock-google-services.json app/google-services.json
+          cp .github/mock-google-services.json wear/google-services.json
+          cp .github/mock-google-services.json automotive/google-services.json
+
+      - name: Run Tests
+        run: ./gradlew test
```
jpelgrom, line 147: CI testing works, but is taking some time. I don't think you need to test release builds? The main PR build also only uses debug, and there is basically no debug-specific code.

```suggestion
        run: ./gradlew testDebugUnitTest
```

jpelgrom:
Closing as this seems abandoned

PR #3005, opened 2022-10-27
Add daily floors sensor for Wear OS3 devices

dshokouhi:
## Summary

Adds a daily floors sensors registered with health services. Updates are pushed to the app and we will send them as they come in, as long as we have actual data to send. Its very hard to test this as we don't really get many events, not even when initially registered. But I did indeed get an update of `0.0` as I did not walk any floors 😛 

Also adds "a few" exercise icons 😄 and attempts to set things up to easily add more sensors to detect when to re-register.

The method `getAvailableSensors()` is now a suspend method so we can do a proper compatibility check for health services sensors and make them available as supported. I have updated the phone and wear OS app to account for this change by moving some things to the view models. I also noticed that the capability check at one point took 17 seconds to complete so I added a loader to the Wear OS app in case that happens for a user so they don't think its an error.

I have also opted to print all the new data points received in case we need to do more analysis on the data that we are collecting.

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#851

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
@@ -17,6 +17,7 @@ import dagger.hilt.android.lifecycle.HiltViewModel
 import io.homeassistant.companion.android.common.bluetooth.BluetoothUtils
 import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
 import io.homeassistant.companion.android.common.sensors.NetworkSensorManager
+import io.homeassistant.companion.android.common.sensors.SensorManager
 import io.homeassistant.companion.android.common.util.DisabledLocationHandler
 import io.homeassistant.companion.android.database.sensor.SensorDao
 import io.homeassistant.companion.android.database.sensor.SensorSetting
@@ -75,10 +76,8 @@ class SensorDetailViewModel @Inject constructor(
     private val _permissionSnackbar = MutableSharedFlow<PermissionSnackbar>()
     var permissionSnackbar = _permissionSnackbar.asSharedFlow()
 
-    val sensorManager = SensorReceiver.MANAGERS
-        .find { it.getAvailableSensors(getApplication()).any { sensor -> sensor.id == sensorId } }
-    val basicSensor = sensorManager?.getAvailableSensors(getApplication())
-        ?.find { it.id == sensorId }
+    var sensorManager: SensorManager? = getSensorManagers()
+    var basicSensor: SensorManager.BasicSensor? = getBasicSensors()
 
     var sensor by mutableStateOf<SensorWithAttributes?>(null)
         private set
@@ -115,10 +114,25 @@ class SensorDetailViewModel @Inject constructor(
         }
     }
 
+    private fun getSensorManagers(): SensorManager? {
+        viewModelScope.launch {
+            sensorManager = SensorReceiver.MANAGERS
+                .find { it.getAvailableSensors(getApplication(), null).any { sensor -> sensor.id == sensorId } }
+        }
+        return sensorManager
+    }
+
+    private fun getBasicSensors(): SensorManager.BasicSensor? {
+        viewModelScope.launch {
+            basicSensor = sensorManager?.getAvailableSensors(getApplication(), null)
+                ?.find { it.id == sensorId }
+        }
+        return basicSensor
+    }
     private suspend fun checkSensorEnabled(sensor: SensorWithAttributes?) {
         if (sensorManager != null && basicSensor != null && sensor != null) {
             sensorCheckedEnabled = true
-            val hasPermission = sensorManager.checkPermission(getApplication(), basicSensor.id)
+            val hasPermission = sensorManager!!.checkPermission(getApplication(), basicSensor!!.id)
             val enabled = sensor.sensor.enabled && hasPermission
             updateSensorEntity(enabled)
         }
@@ -133,14 +147,14 @@ class SensorDetailViewModel @Inject constructor(
                 if ((fineLocation || coarseLocation) &&
                     !DisabledLocationHandler.isLocationEnabled(getApplication())
                 ) {
-                    val sensorName = basicSensor?.let { getApplication<Application>().getString(basicSensor.name) }.orEmpty()
+                    val sensorName = basicSensor?.let { getApplication<Application>().getString(basicSensor!!.name) }.orEmpty()
                     locationPermissionRequests.value = LocationPermissionsDialog(block = true, sensors = arrayOf(sensorName))
                     return
                 } else {
-                    if (!sensorManager.checkPermission(getApplication(), sensorId)) {
+                    if (!sensorManager!!.checkPermission(getApplication(), sensorId)) {
                         if (sensorManager is NetworkSensorManager) {
                             locationPermissionRequests.value = LocationPermissionsDialog(block = false, sensors = emptyArray(), permissions = permissions)
-                        } else if (sensorManager is LastAppSensorManager && !sensorManager.checkUsageStatsPermission(getApplication())) {
+                        } else if (sensorManager is LastAppSensorManager && !sensorManager!!.checkUsageStatsPermission(getApplication())) {
                             permissionRequests.value = permissions
                         } else {
                             permissionRequests.value = permissions
--- a/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
@@ -76,13 +76,17 @@ class SensorDetailViewModel @Inject constructor(
     private val _permissionSnackbar = MutableSharedFlow<PermissionSnackbar>()
     var permissionSnackbar = _permissionSnackbar.asSharedFlow()
 
-    var sensorManager: SensorManager? = runBlocking {
+    val sensorManager: SensorManager? = runBlocking {
         SensorReceiver.MANAGERS
             .find {
-                it.getAvailableSensors(getApplication(), null).any { sensor -> sensor.id == sensorId }
+                it.getAvailableSensors(getApplication()).any { sensor -> sensor.id == sensorId }
             }
     }
-    var basicSensor: SensorManager.BasicSensor? = getBasicSensors()
+
+    val basicSensor: SensorManager.BasicSensor? = runBlocking {
+        sensorManager?.getAvailableSensors(getApplication())
+            ?.find { it.id == sensorId }
+    }
 
     var sensor by mutableStateOf<SensorWithAttributes?>(null)
         private set
@@ -119,17 +123,10 @@ class SensorDetailViewModel @Inject constructor(
         }
     }
 
-    private fun getBasicSensors(): SensorManager.BasicSensor? {
-        viewModelScope.launch {
-            basicSensor = sensorManager?.getAvailableSensors(getApplication(), null)
-                ?.find { it.id == sensorId }
-        }
-        return basicSensor
-    }
     private suspend fun checkSensorEnabled(sensor: SensorWithAttributes?) {
         if (sensorManager != null && basicSensor != null && sensor != null) {
             sensorCheckedEnabled = true
-            val hasPermission = sensorManager!!.checkPermission(getApplication(), basicSensor!!.id)
+            val hasPermission = sensorManager.checkPermission(getApplication(), basicSensor.id)
             val enabled = sensor.sensor.enabled && hasPermission
             updateSensorEntity(enabled)
         }
@@ -144,14 +141,18 @@ class SensorDetailViewModel @Inject constructor(
                 if ((fineLocation || coarseLocation) &&
                     !DisabledLocationHandler.isLocationEnabled(getApplication())
                 ) {
-                    val sensorName = basicSensor?.let { getApplication<Application>().getString(basicSensor!!.name) }.orEmpty()
+                    val sensorName = basicSensor?.let {
+                        getApplication<Application>().getString(
+                            basicSensor.name
+                        )
+                    }.orEmpty()
                     locationPermissionRequests.value = LocationPermissionsDialog(block = true, sensors = arrayOf(sensorName))
                     return
                 } else {
-                    if (!sensorManager!!.checkPermission(getApplication(), sensorId)) {
+                    if (!sensorManager.checkPermission(getApplication(), sensorId)) {
                         if (sensorManager is NetworkSensorManager) {
                             locationPermissionRequests.value = LocationPermissionsDialog(block = false, sensors = emptyArray(), permissions = permissions)
-                        } else if (sensorManager is LastAppSensorManager && !sensorManager!!.checkUsageStatsPermission(getApplication())) {
+                        } else if (sensorManager is LastAppSensorManager && !sensorManager.checkUsageStatsPermission(getApplication())) {
                             permissionRequests.value = permissions
                         } else {
                             permissionRequests.value = permissions
--- a/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
@@ -76,7 +76,12 @@ class SensorDetailViewModel @Inject constructor(
     private val _permissionSnackbar = MutableSharedFlow<PermissionSnackbar>()
     var permissionSnackbar = _permissionSnackbar.asSharedFlow()
 
-    var sensorManager: SensorManager? = getSensorManagers()
+    var sensorManager: SensorManager? = runBlocking {
+        SensorReceiver.MANAGERS
+            .find {
+                it.getAvailableSensors(getApplication(), null).any { sensor -> sensor.id == sensorId }
+            }
+    }
     var basicSensor: SensorManager.BasicSensor? = getBasicSensors()
 
     var sensor by mutableStateOf<SensorWithAttributes?>(null)
@@ -112,14 +117,6 @@ class SensorDetailViewModel @Inject constructor(
                 if (!sensorCheckedEnabled) checkSensorEnabled(it)
             }
         }
-    }
-
-    private fun getSensorManagers(): SensorManager? {
-        viewModelScope.launch {
-            sensorManager = SensorReceiver.MANAGERS
-                .find { it.getAvailableSensors(getApplication(), null).any { sensor -> sensor.id == sensorId } }
-        }
-        return sensorManager
     }
 
     private fun getBasicSensors(): SensorManager.BasicSensor? {
--- a/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorSettingsViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorSettingsViewModel.kt
@@ -49,7 +49,7 @@ class SensorSettingsViewModel @Inject constructor(
     fun updateManagers(sensorManager: SensorManager) {
         val context = getApplication<HomeAssistantApplication>().applicationContext
         viewModelScope.launch {
-            availableSensors = sensorManager.getAvailableSensors(context, null)
+            availableSensors = sensorManager.getAvailableSensors(context)
                 .filter { basicSensor ->
                     sensors.any { basicSensor.id == it.value.id }
                 }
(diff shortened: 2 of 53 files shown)


wear/src/main/java/io/homeassistant/companion/android/home/views/SensorManagerUi.kt
```diff
@@ -49,21 +48,36 @@ fun SensorManagerUi(
...
+                                .padding(vertical = 32.dp),
+                            horizontalAlignment = Alignment.CenterHorizontally,
+                            verticalArrangement = Arrangement.Center
+                        ) {
+                            ListHeader(id = R.string.loading)
+                            CircularProgressIndicator()
+                        }
+                    }
+                } else {
+                    items(allAvailSensors!!.size, { allAvailSensors[it].id }) { index ->
```
jpelgrom, line 70: `!!` could result in a crash as the `else` branch is also used when `allAvailSensors` is `null`

wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt
```diff
@@ -222,8 +222,10 @@ fun LoadHomePage(
                 val sensorManager = getSensorManagers().first { sensorManager ->
                     sensorManager.id() == sensorManagerId
                 }
+                mainViewModel.updateAllSensors(sensorManager)
                 SensorManagerUi(
                     allSensors = mainViewModel.sensors.value,
+                    allAvailSensors = mainViewModel.availableSensors,
```
jpelgrom, line 228: Assume I'm first looking at sensor manager A, which loads its data into `allAvailSensors`, and then I go back and want to look at sensor manager B. In that case I think the data for A would still be in here and you could see incorrect sensors while the data for B is being loaded?

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
```diff
@@ -31,13 +41,26 @@ class HealthServicesSensorManager : SensorManager {
             entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC,
             updateType = SensorManager.BasicSensor.UpdateType.INTENT
         )
+        private val dailyFloors = SensorManager.BasicSensor(
+            "daily_floors",
+            "sensor",
+            commonR.string.sensor_name_daily_floors,
+            commonR.string.sensor_description_daily_floors,
+            "mdi:stairs",
+            unitOfMeasurement = "floors",
```
jpelgrom, line 50: Shouldn't this string be localizable? It isn't really an unit.

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
```diff
@@ -48,8 +71,20 @@ class HealthServicesSensorManager : SensorManager {
...
+            healthClient = HealthServices.getClient(latestContext)
+        if (passiveMonitoringClient == null)
+            passiveMonitoringClient = healthClient?.passiveMonitoringClient
+        if (passiveMonitoringCapabilities == null)
+            passiveMonitoringCapabilities = passiveMonitoringClient?.getCapabilitiesAsync()?.await()
+
+        val supportedSensors = mutableListOf(userActivityState)
+
+        if (passiveMonitoringCapabilities?.supportedDataTypesPassiveMonitoring?.contains(DataType.FLOORS_DAILY) == true)
+            supportedSensors += supportedSensors.plus(dailyFloors)
```
jpelgrom, line 86: Doesn't this create duplicates? I think this means the existing list which has `userActivityState` _plus [`userActivityState`, `dailyFloors`]_.

```suggestion
            supportedSensors += dailyFloors
```

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
```diff
@@ -62,22 +97,36 @@ class HealthServicesSensorManager : SensorManager {
...
+
+        if (!activityState && !isEnabled(latestContext, dailyFloors.id)) {
+            clearHealthServicesCallBack()
             return
         }
 
         if (healthClient == null) healthClient = HealthServices.getClient(latestContext)
         if (passiveMonitoringClient == null) passiveMonitoringClient = healthClient?.passiveMonitoringClient
+
+        val dataTypes = emptySet<DataType<*, *>>().toMutableSet()
```
jpelgrom, line 114:
```suggestion
        val dataTypes = mutableSetOf<DataType<*, *>>()
```

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
```diff
@@ -62,22 +97,36 @@ class HealthServicesSensorManager : SensorManager {
...
     }
 
-    private fun updateUserActivityState() {
-        if (!isEnabled(latestContext, userActivityState.id)) {
-            passiveMonitoringClient?.clearPassiveListenerCallbackAsync()
-            callBackRegistered = false
+    private fun updateHealthServices() {
+        val activityState = isEnabled(latestContext, userActivityState.id)
+
+        if (!activityState && !isEnabled(latestContext, dailyFloors.id)) {
```
jpelgrom, line 106: Daily floors enabled is also used later to determine the data types to be included, so for consistency maybe put it in a variable like activity state enabled?

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
```diff
@@ -90,21 +139,61 @@ class HealthServicesSensorManager : SensorManager {
...
+                    }
+                    onSensorUpdated(
+                        latestContext,
+                        dailyFloors,
+                        floorsDaily[lastIndex].value,
+                        dailyFloors.statelessIcon,
+                        mapOf()
+                    )
+
+                    SensorWorker.start(latestContext)
```
jpelgrom, line 193: For when the app might collect more data ponts, I think this should be moved out of the daily floors if branch.

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
```diff
@@ -124,6 +218,61 @@ class HealthServicesSensorManager : SensorManager {
...
+                    ExerciseType.BIKING, ExerciseType.BIKING_STATIONARY, ExerciseType.MOUNTAIN_BIKING -> "mdi:bike"
+                    ExerciseType.SWIMMING_POOL, ExerciseType.SWIMMING_OPEN_WATER -> "mdi:swim"
+                    ExerciseType.BASEBALL -> "mdi:baseball"
+                    ExerciseType.BASKETBALL -> "mdi:basketball"
+                    ExerciseType.FOOTBALL_AMERICAN -> "mdi:football"
+                    ExerciseType.FOOTBALL_AUSTRALIAN -> "mdi:football-australian"
+                    ExerciseType.SOCCER -> "mdi:soccer"
+                    ExerciseType.SKATING, ExerciseType.INLINE_SKATING -> "mdi:skate"
+                    ExerciseType.ROLLER_SKATING -> "mdi:roller-skate"
+                    ExerciseType.SCUBA_DIVING -> "mdi:diving-scuba"
```
jpelgrom, line 246: This was [renamed in MDI 7.0.96](https://dev.materialdesignicons.com/changelog#version-7.0.96)

```suggestion
                    ExerciseType.SCUBA_DIVING -> "mdi:diving-scuba-mask"
```

jpelgrom:
> an additional suspend method (identical to `requestSensorUpdate`)

That one isn't suspending 😉

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
dshokouhi: We don't localize `unitOfMeasurement` anywhere though so would be strange to do it in one place.

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
dshokouhi: Good suggestion :) will make it easier for others to add sensors later

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
dshokouhi: Well I want to make sure we have the data point because there have been times where we get no data even though the method was called.

app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
```diff
@@ -115,10 +114,25 @@ class SensorDetailViewModel @Inject constructor(
         }
     }
 
+    private fun getSensorManagers(): SensorManager? {
```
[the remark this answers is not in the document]
dshokouhi: I try not to use `runBlocking{}` as much so this was force of habit, thanks for the suggestion will change to this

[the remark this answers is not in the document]
jpelgrom: Good habit, but this viewmodel assumes a value so it's either `runBlocking` or a bigger update.

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
jpelgrom: Are there strings like this used anywhere else? It is not a standard unit so if the install is not in English "floors" will look out of place.

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
jpelgrom: I see, in that case this is a good solution for now.

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
dshokouhi: yes steps and also bluetooth connection, I believe if `mobile_app` integration provides the strings in core it should pick up the translations from there just like some state updates

wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt
dshokouhi: Just tested and this is not the case here, navigating between each manager still only shows its respective available sensor list :)

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorManager.kt
```diff
@@ -99,7 +99,13 @@ interface SensorManager {
      */
     fun requestSensorUpdate(context: Context)
 
-    fun getAvailableSensors(context: Context): List<BasicSensor>
+    suspend fun getAvailableSensors(context: Context, intent: Intent?): List<BasicSensor> {
```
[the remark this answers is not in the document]
dshokouhi: Looks like I will still need to send `null` in several places to use the correct method, unless I need to just drop the other method so they all use the new `suspend` method?

app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorSettingsViewModel.kt
```diff
@@ -43,22 +46,37 @@ class SensorSettingsViewModel @Inject constructor(
         }
     }
 
+    fun updateManagers(sensorManager: SensorManager) {
```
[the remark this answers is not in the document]
dshokouhi: Made an attempt to move the data over to the `init` block however I think due to the `sensors` variable being blank I am unable to generate the `map` that we use in the composable function thus resulting in a blank list of sensors.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -711,7 +712,9 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
 
         SensorWorker.start(this)
         WebsocketManager.start(this)
-        checkAndWarnForDisabledLocation()
+        ioScope.launch {
```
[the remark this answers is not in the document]
dshokouhi: ok I believe I have this setup properly now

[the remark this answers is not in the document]
jpelgrom: I don't think it is - you're calling `repeatOnLifecycle` in `onResume` which will already be repeated. As it is already in `onResume` that check isn't needed, so either move it to `onCreate` or remove the repeating part of it:

```kotlin
lifecycleScope.launch {
    checkAndWarnForDisabledLocation()
}
```

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorManager.kt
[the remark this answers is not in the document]
jpelgrom: Hadn't thought of the conflict, sorry. I believe you're using the new method almost everywhere right now so dropping the other method/keeping the `suspend` method with a default value sounds good.

wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt
jpelgrom: In testing that is also what I see, but do you agree it could happen? Take for example the 17 seconds you experienced on first load, what if that was sensor manager B?

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
jpelgrom: Testing with the Bluetooth connection sensor and I'm not seeing a translation - the state shows up as "1 connection(s)" even when HA is not in English. 

Not a problem for this PR to solve so considering that is already used elsewhere doing it like this is fine for now 👍

app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
```diff
@@ -76,7 +76,12 @@ class SensorDetailViewModel @Inject constructor(
     private val _permissionSnackbar = MutableSharedFlow<PermissionSnackbar>()
     var permissionSnackbar = _permissionSnackbar.asSharedFlow()
 
-    var sensorManager: SensorManager? = getSensorManagers()
+    var sensorManager: SensorManager? = runBlocking {
```
jpelgrom, line 79: This can be a `val` now, and all the `!!` can be removed again

app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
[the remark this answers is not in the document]
jpelgrom: This also applies to `basicSensor`/`getBasicSensors()`, hadn't mentioned it because it seemed obvious to me but I guess not.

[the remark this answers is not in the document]
jpelgrom: This also applies to `basicSensor`/`getBasicSensors()`, hadn't mentioned it because it seemed obvious to me but I guess not.

wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt
dshokouhi: Honestly I am not sure, almost seems like compose is smart enough to see the multiple instances and keep separate values. That is a big guess on my part :)

jpelgrom: Health services don't seem to work on the emulator so I can confirm that it keeps the old values

https://user-images.githubusercontent.com/8148535/198901045-77b6b65a-58ad-405b-82eb-d25fbabf3081.mp4

dshokouhi: hmm trying to think of whats the best way to handle this so the view model can continue to provide the data 🤔  I suppose I could clear the data on each update call to reset it?

dshokouhi: ok this should be fixed in the latest commit now :) appreciate the video and testing!

app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorSettingsViewModel.kt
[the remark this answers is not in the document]
jpelgrom: So while I said "include it in the `init` block", due to how the list is structured that means you'd have to create it in the `filterSensorsList()` function as that is when it changes and you have all the required data. The 'old' implementation in Compose without a suspend function works because it is actually observing the `sensors` variable you're talking about :).


```kotlin
private suspend fun filterSensorsList() {
    // setup context, sensormanager list, basicsensor list
    // create the `sensors` variable from sensormanagers, basicsensors, `sensorsList` and query/filter
    // create the [sensormanager, basicsensor] map from sensormanagers, basicsensors, values in `sensors`
}
```

(sorry for the late response)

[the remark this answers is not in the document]
dshokouhi: Thank you very much :) it has been moved and working as expected now!

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
dshokouhi: Accidentally missed this one earlier when I fixed a similar issue down below :)

app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorSettingsViewModel.kt
```diff
@@ -94,5 +86,14 @@ class SensorSettingsViewModel @Inject constructor(
                     .mapNotNull { sensor -> sensorsList.firstOrNull { it.id == sensor.id } }
             }
             .associateBy { it.id }
+
+        allSensors = managers.associateWith { manager ->
+            availableSensors = manager.getAvailableSensors(app)
```
jpelgrom, line 91: I think you can immediately return this, no need to put it in a variable?

dshokouhi: We access the data from the variable once its ready though?

dshokouhi: oh sorry I see what you mean now :)

jpelgrom: The variable is local / only available in this function, assigned here and then immediately returned if I understand correctly. So what does the variable add vs. returning the assigned value directly?

dshokouhi: You are correct, my bad I misinterpreted which line the comment was about 😅

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
```diff
@@ -90,21 +141,61 @@ class HealthServicesSensorManager : SensorManager {
...
+                            if (dataPoint is IntervalDataPoint) {
+                                val endTime = dataPoint.getEndInstant(bootInstant)
+                                Log.d(
+                                    TAG,
+                                    "Data for ${dataType.name} index: $indexPoint with value: ${dataPoint.value} end time: ${endTime.toEpochMilli()}"
+                                )
+                            }
+                        }
+                    }
+                }
```
jpelgrom, line 177: This doesn't appear to get used, is this just for logging or is there another reason for doing this? (The emulator doesn't create data so I can't test this, but the daily floors specific loop below seems to do almost the same and that result is used.)

jpelgrom:
Just one more comment, otherwise everything is looking good to me!

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
dshokouhi: Yup just for logging to help out when adding additional sensors from health services.

jpelgrom (approved):
Tested the sensor manager related changes, and daily floors sensor looks good to me (but I can't fully test it on the emulator)

JBassett (approved):

Issue #3006, opened 2022-10-27
App won't connect to HA via SSID if internet is down and mobile data is on


habuild:
**Home Assistant Android version:**
2022.10.2-full
**Android version:**
Emui 12.0.0
**Phone model:**
Huawei alp-l29 
**Home Assistant version:**
2022.10.5
**Last working Home Assistant release (if known):**

**Description of problem:**
If home internet is down and mobile data and wifi are active the app won't connect to the local address. Turning off mobile data allows it to connect to the local address. Local address is IP not FQDN. 
**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```
--------- beginning of main
10-27 16:50:45.486  7824  4137 E SensorReceiver: Error while getting core config to sync sensor status
10-27 16:50:45.486  7824  4137 E SensorReceiver: io.homeassistant.companion.android.common.data.integration.IntegrationException: java.net.ConnectException: Failed to connect to /192.168.0.5:8123
10-27 16:50:45.486  7824  4137 E SensorReceiver: 	at io.homeassistant.companion.android.common.data.integration.impl.IntegrationRepositoryImpl.getConfig(IntegrationRepositoryImpl.kt:547)
10-27 16:50:45.486  7824  4137 E SensorReceiver: 	at io.homeassistant.companion.android.common.data.integration.impl.IntegrationRepositoryImpl$getConfig$1.invokeSuspend(Unknown Source:14)
10-27 16:50:45.486  7824  4137 E SensorReceiver: 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
10-27 16:50:45.486  7824  4137 E SensorReceiver: 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
10-27 16:50:45.486  7824  4137 E SensorReceiver: 	at kotlinx.coroutines.internal.LimitedDispatcher.run(LimitedDispatcher.kt:42)
10-27 16:50:45.486  7824  4137 E SensorReceiver: 	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:95)
10-27 16:50:45.486  7824  4137 E SensorReceiver: 	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
[...]

Issue #2532, opened 2022-05-16
Recognize ethernet as Home Network


5A52:
**Is your feature request related to a problem? Please describe.**

I use samsung phone as a phone and Desktop PC, and if it is connected to WIFI (undocked) it uses internal HA url. But When it is connected to ethernet (docked, same network bridged to wifi) app swiсhes to external HA url.

**Describe the solution you'd like**

Add a checkbox to recognize (or not) ethernet as a home network

**Describe alternatives you've considered, if any**

Checkbox or any other option, at least possibility to add "eth0" to list of home wifi networks

**Additional context**

Not only phones with docking possibilities, different android devices are able to use ethernet port.  I am attaching a screenshot of similar functionality implemented in "FolderSync" syncronization app.
![Screenshot_20220516-162547_FolderSync Pro](https://user-images.githubusercontent.com/65124087/168592787-c4d9f274-d1d6-4f7e-bf2f-ae3c251ce251.jpg)

marazmarci:
I think the ideas described in #3480 can be relevant here, as for Ethernet, there's no equivalent single network identifier that is similar to a WiFi access point's BSSID and is available through the Android API. So in case of an Ethernet connection, using the IP range can be our only information to rely on, maybe? 🤷🏻‍♂️

marazmarci:
But note that using IP ranges to detect the home network can be a security concern if the HA server uses HTTP on the LAN.

dshokouhi:
I think for ethernet we can probably use NetworkInterface API to pull the hardware address to properly determine it?

https://developer.android.com/reference/java/net/NetworkInterface#getHardwareAddress()

marazmarci:
Yeah, but you cannot expect a device with Ethernet to be stationary (I mean not mobile).

For example a user may move from place to place and connect to Ethernet at multiple places, to different networks.

5A52:
> I think the ideas about IP ranges described in #3480 can be relevant here, as for Ethernet, there's no equivalent single network identifier that is similar to a WiFi access point's BSSID and is available through the Android API. So in case of an Ethernet connection, using the IP range can be our only information to rely on, maybe? 🤷🏻‍♂️

In case of checkbox "use ethernet" I think it is inought to check if internal url is online, and if not use external.

> Yeah, but you cannot expect a device with Ethernet to be stationary (I mean not mobile).
For example a user may move from place to place and connect to Ethernet at multiple places, to different networks.

[...]

PR #3015, opened 2022-10-29
Only use internal URL when data is also using Wi-Fi

jpelgrom:
## Summary

Add a check to the external/internal URL logic to make sure that the device is not only connected to a Home Wi-Fi network, but is also using that network for data for using the internal URL. This better matches the expected behavior. Inspired by #3006. 

A device can be connected to Wi-Fi but may be using a different (usually mobile) network for data for example if the internet is down or the network is considered "low quality". Expected log output in these situations is something like:
```
2022-10-29 15:32:21.317 11938-12081 UrlRepository           io....stant.companion.android.debug  D  localUrl is: true, usesInternalSsid is: true, usesWifi is: false
2022-10-29 15:32:21.318 11938-12081 UrlRepository           io....stant.companion.android.debug  D  Using external URL
```

## Screenshots

n/a

n/a

## Any other notes

Testing note: my Pixel on Android 13 simply hides the Wi-Fi icon from the status bar / shows the 4G data icon when Wi-Fi is not used for data, even though it may be connected. Check the system settings app to be sure.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/ssid/SsidFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/ssid/SsidFragment.kt
@@ -46,6 +46,7 @@ class SsidFragment : Fragment() {
                     SsidView(
                         wifiSsids = viewModel.wifiSsids,
                         prioritizeInternal = viewModel.prioritizeInternal,
+                        usingWifi = viewModel.usingWifi,
                         activeSsid = viewModel.activeSsid,
                         activeBssid = viewModel.activeBssid,
                         onAddWifiSsid = viewModel::addHomeWifiSsid,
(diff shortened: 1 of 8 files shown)


dshokouhi (approved):
Nice improvement!

marazmarci:
Thanks for this improvement! It will surely be useful.

I was wondering about a less frequent use case, when the phone is connected to Ethernet and the device has cellular connection as a fallback. Here in this PR, the changes only take WiFi into account, so in the Ethernet use case, the same problem will arise.

jpelgrom:
> I was wondering about a less frequent use case, when the phone is connected to Ethernet and the device has cellular connection as a fallback. Here in this PR, the changes only take WiFi into account, so in the Ethernet use case, the same problem will arise.

Currently ethernet won't be recognized as a home network, so I think this would be part of feature request #2532?

marazmarci:
> Currently ethernet won't be recognized as a home network, so I think this would be part of feature request #2532?

Oh wow, yes, I saw this FR earlier, but forgot about it.

Yes, I also think this should be part of it.

JBassett (approved):

PR #3022, opened 2022-10-30
More granular sensor update frequency

TheKingOfAtlantis:
## Summary

The aim of this PR is to enable more granular control over the frequency of sensor updates (at least those which are updated regularly by the worker) and bring greater feature parity with the iOS counterpart of this app.

This feature was initially motivated out of frustration that my only real choose with the frequency of updates was either 1 or 15mins, which felt either too short and thus using up an "unnecessary" amount of my battery (admittedly this is a bit subjective) or too long missing out on events which occur within the time between updates.

This PR does a couple things to try and resolve this:
 1. Provides the user with a dropdown containing various update frequencies (this could be modified further to text field instead giving the user further flexibility) which are then used to specify the period of the SensorWorker
 2. And (In the spirit of granularity and flexibility) provides the user with the ability to specify different update frequencies when the device is charging vs when it is running off its batteries

As such the following main changes have occured:
 1. The radio menu which allows the user to modify the frequency of updates has been replace with two dropdown menus (one of battery and the other for powered/charging)
 2. The strings relating to sensor updates have been replaced with ones which are better fit for having more granular and separated update frequency control
 3. The app no longer registers the worker receiver with the ACTION_TIME_TICK which provided minute by minute updates as this role should be taken over by adjusting the frequency on the periodic worker
 4. The periodic worker is given a frequency to use rather than just running every 15mins

### Light Mode
| Sensor Frequency Settings | Sensor Frequency Chip | Sensor Frequency Info Dialog |
|---------------------------|-----------------------|------------------------------|
| image | image | image |

### Dark Mode
| Sensor Frequency Settings | Sensor Frequency Chip | Sensor Frequency Info Dialog |
|---------------------------|-----------------------|------------------------------|
| image | image | image |

_Dark mode also highlights how we show a more concise description of update frequency when both battery and charging/powered frequencies are the same_

TODO

### Todo
A couple tasks still need to be done before I am happy with calling this done which are:
 - [ ] Tests: Need to properly ensure that updates work as expected (especially that we are no longer registering to ACTION_TIME_TICK for minute by minute updates)
 - [ ] Documentation: The companion app documentation needs to reflect the changes made (if necessary)
 - [ ] WearOS: PR had been wholly focused on the phone side and the wearable aspect may need to be addressed 
 - [ ] Make reasonable limitations with regards to not allowing the user to pick a higher frequency update schedule when on battery than when charging


jpelgrom:
I know that testing is still in the TODO checklist, but did you consider that the sensor worker [won't accept a frequency](https://developer.android.com/reference/kotlin/androidx/work/PeriodicWorkRequest.Builder?hl=en) that is [less than 15 minutes](https://android.googlesource.com/platform/frameworks/support/+/androidx-main/work/work-runtime/src/main/java/androidx/work/PeriodicWorkRequest.kt#183)? That is the current reason for 15 minutes using the worker _or_ 1 minute using the intent options.

TheKingOfAtlantis:
Ah... I see, I wasn't aware of the limitation (seeing as android dev is only a hobby of mine).

Well could we instead run multiple workers which are offset give the illusion of more frequent workers avoiding the period limitation OR reinstate the ACTION_TIME_TICK and cancel the job immediately if the minute is not the modulus of our frequency

jpelgrom:
Workers aren't guaranteed to run exactly on time, only within the time period (so once somewhere between minute 0 and 15). So if you plan multiple workers, even when offset, the system might execute them all at the same time to save resources/battery.

> reinstate the ACTION_TIME_TICK and cancel the job immediately if the minute is not the modulus of our frequency

That would come closest to what you're describing, but would mean giving up the benefits of the worker like more efficient battery usage. Also how would you determine what is "our" modulus? The device can reboot or the clock can change or the device may pause the app to save battery or ...

dshokouhi:
I would say for the purpose of this PR it may make sense to make the 15 minute adjustable like 15, 30, 45 etc... that way the worker can run as would be expected.  For 5 and 10 minutes, probably best to keep track of time and perform updates as needed based on the TIME_TICK intent?

TheKingOfAtlantis:
The split approach is definitely the way forward
I've just done a quick preliminary implementation of using modulus for high frequency updates but I think using the time of the last update would be better (if there exists a way already to do so without the last update sensor)

Unfortunately I might not be able to do much until possibly next weekend but feel free to contribute changes to the PR if you find the time

jpelgrom:
Closing as stale / due lack of interest from community

PR #3024, opened 2022-10-30
Add setting to start the app always on the first dashboard view

chriss158:
## Summary

This PR adds a setting to always start on the first view of the user dashboard. 

Currently it is quite annoying if you close the app while you are in the settings and then later when you open the app again, because you want to just turn off a light, your are again back in the settings. Now you have to close the settings first and then select the proper view to turn off the light.

Also this is helpful when you are on your last view in the dashboard. Just to set a setting you need to change only once in a year. If you then open the app again you will be right on the first view on your dashboard. 

I tried to do the switch to the first view of the dashboard as fast as possible. I dont wan't to have a loading screen. Because of that, i decided to load the first view when we close (Event `onPause`) the app. Then the first view is already loaded on opening the app. Also i use javascript to load the first view, because when we do a `loadUrl` on `/lovelace` the webview will do a complete reload of the page.

The javascript just do a click on the first element in the home assistant sidebar.
image

## Screenshots

image
image

Documentation: home-assistant/companion.home-assistant#853


jpelgrom:
This setting breaks the expected navigation when opening a screen provided by the app (for example, companion app settings) and in some cases this can break flows (for example, NFC tag writing from the frontend). When pressing back, instead of returning back to the frontend where the user was, it will show a dashboard.

The short flash of a dashboard loading opening before showing an activity also isn't desirable IMO.

If the above can be fixed, would it make sense to align this with the default dashboard profile setting instead? For most users that will be the same dashboard/view but not necessarily.

chriss158:
> This setting breaks the expected navigation when opening a screen provided by the app (for example, companion app settings) and in some cases this can break flows (for example, NFC tag writing from the frontend). When pressing back, instead of returning back to the frontend where the user was, it will show a dashboard.

You are right. The navigation flow breaks. Never thought of the navigation flow. Maybe i just remove this one to fix this issue. After all, you're not in the app's settings every day. So you also don't have to deal with other side effects that I/we don't think about at the moment.

> The short flash of a dashboard loading/opening before showing an activity also isn't desirable IMO.

I don't see  dashboard loading/opening flash. When do you see that on exactly? Maybe this is depending on the device. Using a Pixel 6 Pro

> If the above can be fixed, would it make sense to align this with the default dashboard profile setting instead? For most users that will be the same dashboard/view but not necessarily.

Do you mean we should show an additional select box to let the user decide on which dashboard and/or on which view on the dashboard the user will see when the app is opened?

jpelgrom:
> Maybe i just remove this one to fix this issue.

While navigating to/from the settings now works as expected again, writing NFC tags from the frontend still returns you to the dashboard unexpectedly.

> (...) So you also don't have to deal with other side effects that I/we don't think about at the moment.

I don't quite understand why we don't have to deal with side effects? The entire point is that there are side effects, while the setting description says that you're returned to the dashboard (only) when the app is opened, not while using the app.

> I don't see dashboard loading/opening flash. When do you see that on exactly? Maybe this is depending on the device. Using a Pixel 6 Pro

I'm using a Pixel 4a, so not extremely fast and not slow either, but I don't think the hardware matters much, more likely the server/network speed. I've recorded a short video to illustrate it, after tapping on companion app settings you can see the dashboard appear/flash shortly before the settings are shown:

https://user-images.githubusercontent.com/8148535/199835257-f0551184-49df-4aed-b618-14ddb81126e3.mp4

> Do you mean we should show an additional select box to let the user decide on which dashboard and/or on which view on the dashboard the user will see when the app is opened?

Not a selection (that would require reading menu data, keeping it updated, emulating clicks probably won't be enough, ...), but reading + using the value of the profile setting if possible:

A screenshot of the profile setting 'Dashboard - Pick a default dashboard for this device.'

I haven't done any investigation to see if this is relatively easy to support, but it is basically the same setting as you're adding, but it controls what is shown on app start when the app is **not** in the background.

chriss158:
> While navigating to/from the settings now works as expected again, writing NFC tags from the frontend still returns you to the dashboard unexpectedly.

Oh missed that there is an additional NFC tag setting in the frontend. Fixed by not allowing to call the first view while on the config page and config subpages of HA. However, i'm not sure if we should allow to call the first view if we are on the config dashboard page of HA (not the subpages). 

> I'm using a Pixel 4a, so not extremely fast and not slow either, but I don't think the hardware matters much, more likely the server/network speed. I've recorded a short video to illustrate it, after tapping on companion app settings you can see the dashboard appear/flash shortly before the settings are shown:

Good catch.  I hardly noticed that. Fixed by counting the open activities and just call the first view if there is no activity, which is in a start state, left. 

> I haven't done any investigation to see if this is relatively easy to support, but it is basically the same setting as you're adding, but it controls what is shown on app start when the app is not in the background.

So this would be basically the same select box as in the profile settings. only this controls which dashboard is shown on app start? Sorry if I still do not quite understand. 
Initially i thought of that too, but i wanted to keep it simple. But i nevertheless look tomorrow how i could implement this.

jpelgrom:
> So this would be basically the same select box as in the profile settings. only this controls which dashboard is shown on app start? Sorry if I still dont' quite understand. Initially i thought of that too, but i wanted to keep it simple. But i nevertheless look tomorrow how i could implement this.

No additional select box, just using that value. So instead of going to the first menu item when the app is put in the background, go to [the menu item for the selected dashboard in the profile settings].

If the first item in the menu _is_ the same as the selected dashboard, the result is the same.

If the first item in the menu _isn't_ the same as the selected dashboard, the current behavior results in different behavior depending on the app state when reopened. If it is still in the background, the first menu item is shown, but if it isn't/the app was completely stopped, the profile settings' dashboard is shown.

chriss158:
> If the first item in the menu is the same as the selected dashboard, the result is the same.
>
> If the first item in the menu isn't the same as the selected dashboard, the current behavior results in different behavior depending on the app state when reopened. If it is still in the background, the first menu item is shown, but if it isn't/the app was completely stopped, the profile settings' dashboard is shown.

If i set the default dashboard in the profile settings, it will be always the first item in the menu. Even if the dashboard is configured as "Not shown in sidebar". Then the dashboard will be named just "Overview". Also if the dashboard is configured as "Show in sidebar" then it will be also the first item in the menu, but named after the dashboards selected title.
At least this is the behavior on my instance. 
That's why i didn't understand it the first time. 

grafik

jpelgrom:
> If i set the default dashboard in the profile settings, it will be always the first item in the menu. 

By default, yes, but it is possible to re-order the menu items in which case this is no longer true.

chriss158:
Ah right you can rearrange it. Never used it before. I will have a look

chriss158:
Luckily the default dashboard is in the local storage. So this pretty easy to get. Also i didn't find any API to get the default dashboard. Only to fetch all dashboards. 

Thanks for review @jpelgrom

jpelgrom:
Thanks for your patience as well while I struggled to explain myself properly :)

The comment for the code that checks if the current page is in settings suggests that all pages - except the landing page - shouldn't change when the app is resumed, but this isn't working correctly for some (for example the automations list and areas). I think it matches on any `/dashboard` at the end of the URL (area settings: `/config/areas/dashboard`) instead of only on `/config/dashboard`.

Is naming `LivecycleHandler` with a `v` instead of a `f` intentional?

chriss158:
> The comment for the code that checks if the current page is in settings suggests that all pages - except the landing page - shouldn't change when the app is resumed, but this isn't working correctly for some (for example the automations list and areas). I think it matches on any /dashboard at the end of the URL (area settings: /config/areas/dashboard) instead of only on /config/dashboard.

I tried it a thousand times. I don't know how i overlooked this :dizzy_face: 

> Is naming LivecycleHandler with a v instead of a f intentional?

You are absolulty right 👍 Thanks. Fixed both issues.

jpelgrom (approved):
Tested - seems to work correctly and like expected. Using custom JS for a feature like this is a risk as we've seen with the [quick bar gesture](https://companion.home-assistant.io/docs/integrations/android-webview/#swipe-gestures) but I think the menu should be OK.

jpelgrom:
Linking this here for future reference: after approving the PR, [there were some comments on Discord](https://discord.com/channels/330944238910963714/377667559303938050/1039300917410942996) from me, @chriss158, @dshokouhi showing how manipulating the frontend without official support can break and later un-break without app changes and how the 'correct' way to do something like this would be using the external bus, but it is still fine for various reasons (which mostly can be summarised to "custom code like this is already in the app elsewhere").

dshokouhi (approved):
Looks good to me :)

JBassett (approved):

PR #3025, opened 2022-10-31
Don't check Wifi SSID when home wifi list is empty

dshokouhi:
## Summary

If home wifi SSID is empty then do not bother checking the connected SSID or BSSID as we don't need to access the data.

## Any other notes

Ref: https://discord.com/channels/330944238910963714/562408603345092636/1036401886842064956

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/url/UrlRepositoryImpl.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/url/UrlRepositoryImpl.kt
@@ -152,9 +152,11 @@ class UrlRepositoryImpl @Inject constructor(
     }
 
     override suspend fun isHomeWifiSsid(): Boolean {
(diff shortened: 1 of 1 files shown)


jpelgrom:
I wonder if the changes in #3015 will be considered 'location access' by this user's software? If so these changes will work but only until that PR is merged 😅

dshokouhi:
> I wonder if the changes in #3015 will be considered 'location access' by this user's software? If so these changes will work but only until that PR is merged 😅

Well as long as we avoid any internal URL checks if home WiFi isn't set I guess it should be fine?

jpelgrom:
> Well as long as we avoid any internal URL checks if home WiFi isn't set I guess it should be fine?

Did a quick test and determining if the active network is WiFi also works without location access so yes it should be fine 👍

jpelgrom (approved):

JBassett (approved):

PR #3026, opened 2022-10-31
Track time between hardware sensor updates to restart the listener as needed

dshokouhi:
## Summary

Noticed at times some hardware sensors would get its state stuck and thus needing a force stop of the app to reset the variable to continue to register for updates. I realized the state is stuck because other sensors were updating its state but these ones were not. So now adding a check where if 1 minute has passed and we are still registered on the next sensor update to reset the variable so we can get new updates.

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/LightSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/LightSensorManager.kt
@@ -17,6 +17,7 @@ class LightSensorManager : SensorManager, SensorEventListener {
 
         private const val TAG = "LightSensor"
         private var isListenerRegistered = false
+        private var listenerLastRegistered = 0
         private val lightSensor = SensorManager.BasicSensor(
             "light_sensor",
             "sensor",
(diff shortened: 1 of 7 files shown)


jpelgrom:
Would it make sense to add the `60000` as a (Android) sensor manager timeout constant in the `SensorManager` class, so that it can easily be changed in the future if it turns out it needs to be a different value?

The `OnBodySensorManager` also uses `android.hardware.SensorManager`, but you're not resetting the callback for it. Should the sensor in that class be updated to `UpdateType.INTENT` instead, or is it intended to update periodically (as stated) but doesn't it matter if there is no value within 60s?

dshokouhi:
> Would it make sense to add the `60000` as a (Android) sensor manager timeout constant in the `SensorManager` class, so that it can easily be changed in the future if it turns out it needs to be a different value?

yup that does make sense, will add it :)



> Should the sensor in that class be updated to `UpdateType.INTENT` instead, or is it intended to update periodically (as stated) but doesn't it matter if there is no value within 60s?

Intent is probably the best fit here because the sensor reports within 5 seconds. There is no reason to track the time between updates because we never unregister the listener like the other hardware sensors. This one reports as soon as it detects a state change :)  I will update the type too.

jpelgrom (approved):

JBassett (approved):

PR #3030, opened 2022-11-02
Fix location updates on core <2022.2

jpelgrom:
## Summary

It turns out that #2969 breaks location updates for people on core <2022.2, which wasn't intended. This PR fixes that.

Always including `location_name` breaks compatibility with previous versions as it will reject updates: `Received invalid webhook payload: string value is None for dictionary value @ data['location_name']. Got None`. Updates would look like this:

```json
{"type":"update_location","data":{"gps":[1, 2],"gps_accuracy":13,"location_name":null,"speed":0,"altitude":47,"course":7,"vertical_accuracy":1}}
```

Now, updates on core <2022.2 and when using the 'Exact' setting look like this (again):

```json
{"type":"update_location","data":{"gps":[1, 2],"gps_accuracy":13,"speed":0,"altitude":47,"course":7,"vertical_accuracy":1}}
```

And 'Zone name only' updates will only include the location name:

```json
{"type":"update_location","data":{"location_name":"not_home"}}
```

Only including non-null values works correctly on newer versions (it still clears the GPS data / switches to zone when sending just the `location_name`, and clears the forced zone when sending GPS data) and restores compatibility with older versions. Tested on:
 - core 2021.12.10 - last version before app zone only support
 - core 2022.2.0 - first version with app zone only support
 - core 2022.10.5 - latest version while testing

## Screenshots

n/a

n/a

## Any other notes

Reported [on Discord](https://discord.com/channels/330944238910963714/562408603345092636/1037390991193297027)

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/entities/UpdateLocationRequest.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/entities/UpdateLocationRequest.kt
@@ -1,5 +1,8 @@
 package io.homeassistant.companion.android.common.data.integration.impl.entities
 
+import com.fasterxml.jackson.annotation.JsonInclude
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
 data class UpdateLocationRequest(
     val gps: Array<Double>?,
     val gpsAccuracy: Int?,


dshokouhi (approved):
Thanks for the quick fix!

JBassett (approved):

PR #3054, opened 2022-11-09
Add Daily Calories, Distance and Steps sensors for health services

dshokouhi:
## Summary

Add 3 new sensors from Health Services:

* daily calories - measured in kcal
* daily distance - measured in meters
* daily steps

Documentation: home-assistant/companion.home-assistant#859

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
@@ -50,6 +50,36 @@ class HealthServicesSensorManager : SensorManager {
             entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC,
             updateType = SensorManager.BasicSensor.UpdateType.INTENT
         )
+        private val dailyDistance = SensorManager.BasicSensor(
+            "daily_distance",
+            "sensor",
+            commonR.string.sensor_name_daily_distance,
+            commonR.string.sensor_description_daily_distance,
+            "mdi:map-marker-distance",
+            unitOfMeasurement = "m",
+            entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC,
+            updateType = SensorManager.BasicSensor.UpdateType.INTENT
+        )
+        private val dailyCalories = SensorManager.BasicSensor(
+            "daily_calories",
+            "sensor",
+            commonR.string.sensor_name_daily_calories,
+            commonR.string.sensor_description_daily_calories,
+            "mdi:fire",
+            unitOfMeasurement = "kcal",
+            entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC,
+            updateType = SensorManager.BasicSensor.UpdateType.INTENT
+        )
+        private val dailySteps = SensorManager.BasicSensor(
+            "daily_steps",
+            "sensor",
+            commonR.string.sensor_name_daily_steps,
+            commonR.string.sensor_description_daily_steps,
+            "mdi:shoe-print",
+            unitOfMeasurement = "steps",
+            entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC,
+            updateType = SensorManager.BasicSensor.UpdateType.INTENT
+        )
     }
 
     private lateinit var latestContext: Context
@@ -85,6 +115,12 @@ class HealthServicesSensorManager : SensorManager {
 
         if (passiveMonitoringCapabilities?.supportedDataTypesPassiveMonitoring?.contains(DataType.FLOORS_DAILY) == true)
             supportedSensors += dailyFloors
+        if (passiveMonitoringCapabilities?.supportedDataTypesPassiveMonitoring?.contains(DataType.DISTANCE_DAILY) == true)
+            supportedSensors += dailyDistance
+        if (passiveMonitoringCapabilities?.supportedDataTypesPassiveMonitoring?.contains(DataType.CALORIES_DAILY) == true)
+            supportedSensors += dailyCalories
+        if (passiveMonitoringCapabilities?.supportedDataTypesPassiveMonitoring?.contains(DataType.STEPS_DAILY) == true)
+            supportedSensors += dailySteps
         return supportedSensors
     }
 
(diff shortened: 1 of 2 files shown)


dshokouhi:
> Daily steps 'conflicts' with the existing steps (since last reboot) sensor. Did you consider limiting availability of the existing steps sensor to Wear OS 2 devices to prevent two very similar sensors? Or is the feature simply not available on real Wear OS 3 watches?

So personally speaking in my testing while I did see Daily Steps updating, its very very infrequent when compared to the existing hardware steps sensor. Its to the point where you will see the hardware sensor updating more frequently than the daily steps sensor so that might lead to some user confusion in itself if a user tries to compare the 2.  In our docs we also mention using the Utility Meter integration to get a daily count so that may be another consideration. Personally speaking the data comes from a separate API so I think its worth it to let the user decide which sensor they want to stick with. I think most users expect to see Daily Steps so its nice to add it here :)

Health services itself is only available on Wear OS3 devices so only those users would see the potentially duplicate sensor.

On the subject of duplicating sensors. I opted to add a hardware sensor for Heart Rate as opposed to Health Services simply because on my device I was getting updates way too frequently from Health Services and that wouldn't be good on battery life :) It felt like a better route to go by scheduled updates there.



> (A similar question would arise if the `ActivitySensorManager` from the main app was added to the Wear app.)

Personally I think this sensor has its own uses as well since it uses a different API (available on Wear OS 2) so you get different data. The states are also completely different :)

jpelgrom:
> So personally speaking in my testing while I did see Daily Steps updating, its very very infrequent when compared to the existing hardware steps sensor. (...)

Can't test that on an emulator so thanks for the clarification, in that case it makes sense to provide both 👍

Another potential issue: wouldn't a user expect the 'Steps Sensor' sensor manager to also show the daily steps sensor?

> On the subject of duplicating sensors. I opted to add a hardware sensor for Heart Rate as opposed to Health Services simply because on my device I was getting updates way too frequently from Health Services and that wouldn't be good on battery life :)

In that case there's a good reason to add only one option, but in this case it seemed to me that both were providing similar data, which is why I asked.

dshokouhi:
> Another potential issue: wouldn't a user expect the 'Steps Sensor' sensor manager to also show the daily steps sensor?

Thats a good question, maybe in a future PR we combine the hardware sensors (`SensorEventListener`) so its more obvious the sensor reads the value from the hardware directly? 



> In that case there's a good reason to add only one option, but in this case it seemed to me that both were providing similar data, which is why I asked.

they are indeed going to be similar lol, the real question for the user will be which is more accurate 😂

jpelgrom:
> Thats a good question, maybe in a future PR we combine the hardware sensors (`SensorEventListener`) so its more obvious the sensor reads the value from the hardware directly? 

Most users will just want their step count and don't care which API is used to retrieve it. Especially on the watch where you have to tap on a sensor manager first to see the sensors that makes it hard to have it be easily discoverable.

SkechyWolf:
Personally, I would use the daily steps more for counting toward my step goal since that resets at midnight every night but that is just one person's opinion

dshokouhi:
> Most users will just want their step count and don't care which API is used to retrieve it. Especially on the watch where you have to tap on a sensor manager first to see the sensors that makes it hard to have it be easily discoverable.

yea discoverability is not the greatest here thats why I was thinking renaming the manager will help make it more clear. Google classifies the sensor as a "Motion Sensor" but that doesn't quite fit with our terminology lol. That was why I was thinking the term `Hardware` makes the most sense since we are reading the raw value.  Health services doesn't just use the hardware sensor to get its data (not specifically referring to daily steps just in general)



> Personally, I would use the daily steps more for counting toward my step goal since that resets at midnight every night but that is just one person's opinion

the issue with the daily steps though is we can't give out any guarantee teh sensor will have a final update before it resets. We cannot control that. At least with the hardware sensor we will check at a normal interval what it is. So they both have benefits and drawbacks.

jpelgrom:
> > Personally, I would use the daily steps more for counting toward my step goal since that resets at midnight every night but that is just one person's opinion
> 
> the issue with the daily steps though is we can't give out any guarantee teh sensor will have a final update before it resets. We cannot control that.

For a step goal you could also [subscribe to the goal, if known](https://developer.android.com/training/wearables/health-services/passive#passive-goals), to make sure at least it is updated when it is reached at the last minute.

dshokouhi:
> For a step goal you could also [subscribe to the goal, if known](https://developer.android.com/training/wearables/health-services/passive#passive-goals), to make sure at least it is updated when it is reached at the last minute.

yup I think for Goals to properly work we need sensor settings in place because a user needs to specify a goal type and amount to be notified.  We also need to consider how it would work, separate sensor or part of the existing? Do we fire an event when this happens?

jpelgrom:
> We also need to consider how it would work, separate sensor or part of the existing? Do we fire an event when this happens?

I would treat it as a data point for the daily steps sensor that triggers an update. However as this requires settings I'd say it is out of scope for this PR ;)

wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
```diff
@@ -175,25 +224,14 @@ class HealthServicesSensorManager : SensorManager {
...
-                        floorsDaily[lastIndex].value,
-                        dailyFloors.statelessIcon,
-                        mapOf()
-                    )
 
+                if (
+                    processDataPoint(floorsDaily, dailyFloors) ||
+                    processDataPoint(distanceDaily, dailyDistance) ||
+                    processDataPoint(caloriesDaily, dailyCalories) ||
+                    processDataPoint(stepsDaily, dailySteps)
```
jpelgrom, line 232: Stacking the processing functions like this means that if the app receives multiple data point types at once, other data points aren't processed: it will stop executing functions once one of these returns `true`. For example, if the app were to receive distance and steps at the same time, only the distance data points would be processed.

dshokouhi: Ah I probably didn't catch this in testing as I only got 1 data type per update. I'll submit a change shortly so we can ensure each data point is processed before sending the update.

jpelgrom: I don't know if it actually happens but think the app should handle it just in case sensor updates are batched

dshokouhi: yea they could be batched if we decide to switch from the callback to the service. I went with the callback only because its said to be faster.

`You can receive passive data through a service or a callback (or both). A service allows your app to receive data in the background when no part of your app is visible in the foreground. When you receive data in the background it is delivered in batches. The callback receives data at a slightly faster rate, but only while the app is running and the callback is successfully notified.`

as we are a foreground service the callback works just fine :) except for when we are hit by doze lol but thats out of our control

dshokouhi: > Also the app isn't using a foreground service all the time for the callback so it is probably still "in the background" as far as Android is concerned.

i havent seen any issues with using the callback, updates still come in as expected throughout the day.

jpelgrom (approved):
Code looks good to me and the app doesn't crash (unable to test data points on the emulator but that shouldn't matter)

JBassett (approved):

PR #3069, opened 2022-11-14
Only update health services if sensor state actually changed and throttle updates to once every 15 minutes to match the worker and prevent heavy updates

dshokouhi:
## Summary

Now that I have been doing some more long term testing I came to realize a few things:

- Every device has its own update interval, for example a pixel watch will update steps and distance anytime a change is detected. I have seen cases where there are multiple updates per minute when I am just walking around. Calories on a pixel watch also updates once per minute which is a very high rate for a wearable, probably the calorie counting bug mentioned in the blogs. On a Galaxy Watch 4 the updates are more sporadic, they still come in but a different interval altogether.
- Health services will send us the same value numerous times, most obvious with daily floors. Steps and distance don't appear to have this issue and Calories will that always changes :)
- At times it may appear health services sensors get "stuck". In fact what I have noticed is that the updates start at a particular interval and then slowly go longer until they stop coming in. I assume we are exhausting some kind of doze/app stand by bucket because if I whitelist our app using ADB I do not see that issue.

So in this PR we will keep track of the updates and only send them when enough time has passed (15 minutes to match our sensor worker). The sending really only impacts when these sensors have an update. We still store the state so it can be sent during the normal interval. This helps cut down on the amount of resources we use by not updating all sensors and making a network call every time.

The `sendUpdate` variable is also dependent upon the state of the sensor actually changing, there is no reason to waste resources on a sensor update if nothing has changed.

User Activity State is not impacted by this as it only updates upon a state change.

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
@@ -90,6 +90,7 @@ class HealthServicesSensorManager : SensorManager {
     private var callBackRegistered = false
     private var dataTypesRegistered = emptySet<DataType<*, *>>()
     private var activityStateRegistered = false
+    private var lastUpdateSent = 0L
 
     override fun docsLink(): String {
         return "https://companion.home-assistant.io/docs/wear-os/sensors#health-services"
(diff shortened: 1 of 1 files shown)


dshokouhi:
One thing that I did notice yesterday during testing was that it appeared our Sensor Worker stopped running, even when the app was opened as I was expecting to see battery level update. It did not update when I opened the app as we do indeed start our worker then. I decided to try and whitelist the app from doze and upon connecting to WiFi our network state change intent kicked in and the battery level state updated. So it does seem there is some kind of doze restriction occurring at times. This is probably going to vary from user to user.

dshokouhi:
testing a different approach so marking as a draft, my next attempt will be to just store the data received and send it out as part of normal update interval.

dshokouhi:
Closing in favor of #3076

PR #3076, opened 2022-11-16
Store health services updates and send them with worker updates

dshokouhi:
## Summary

This is a different approach compared to #3069 where instead of pushing data point updates every 15 minutes, we just send them along with our worker updates.

I ran a test for 24 hours and while the worker did not stop during the day, it seemed to have stopped shortly after enabling bedtime mode. What actually happens is that Android starts to throttle our jobs for an undetermined length of time, updates from intents are still coming through. The jobs end up getting canceled as soon as they are created.  Placing the app in the whitelist does prevent this behavior and in my testing is decent on battery life.

Its going to be difficult to find a happy medium here and this behavior will vary from user to user.

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
@@ -48,7 +48,7 @@ class HealthServicesSensorManager : SensorManager {
             "mdi:stairs",
             unitOfMeasurement = "floors",
             entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC,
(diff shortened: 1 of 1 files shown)


jpelgrom:
If the update type is set to `CUSTOM` it should be documented in my opinion, even if the Wear app doesn't display it (yet). In this case maybe `WORKER` would also be fine as it is still sends the most recent value?

dshokouhi:
> If the update type is set to `CUSTOM` it should be documented in my opinion, even if the Wear app doesn't display it (yet). In this case maybe `WORKER` would also be fine as it is still sends the most recent value?

I think worker may make the most sense here since we send updates with the current schedule. I'll update the types to be more correct.

jpelgrom (approved):
Hopefully this provides a good balance between battery life and timely updates

JBassett (approved):

PR #3080, opened 2022-11-17
Update and fix in-app language picker

### LINKED ISSUE #2821: Language issue in latest app update
2022.8.0

**Android version:**
12

**Phone model:**
Asus ZenFone 8 

**Home Assistant version:**
2022.8.6

2022.8.6

**Description of problem:**
Funny language for selected options in last HA Android app update. See enclosed example. This also goes for copy/paste menu.

```

```

**Screenshot of problem:**
Screenshot_20220826-111953276

**Additional information:**

### LINKED ISSUE #2832: Changing app language doesn't work
**Home Assistant Android version:** 2022.6.0-full, 2022.8.0-full

**Android version:** tested on Android 8.1, Android 11 and Android 13

**Phone model:** emulator, Google Pixel 2, Google Pixel 4a

**Home Assistant version:** doesn't matter

**Last working Home Assistant release (if known):** probably beta-1696? see below

**Description of problem:**

The option to change the app's language in the companion app settings doesn't seem to work. Selecting a different value will recreate the activity, and so does the webview activity when I go back, but the language stays the same. Even after a force stop, the language doesn't change.

Looking at what this option does, I've found [`getContextWrapper` in the `LanguagesManager`](https://github.com/home-assistant/android/blob/84aa4454b731e4058cce083d947c27a4d90c5469/app/src/main/java/io/homeassistant/companion/android/settings/language/LanguagesManager.kt#L90-L103), which seems to be the only place where the language setting is actually used, however this function is only ['used' in the `BaseActivity`](https://github.com/home-assistant/android/blob/84aa4454b731e4058cce083d947c27a4d90c5469/app/src/main/java/io/homeassistant/companion/android/BaseActivity.kt#L11) and was commented out when switching to Hilt (672e49b6fbd48baff466f0a0ba7dfb3292768602).

I feel like I'm missing something as it seems unlikely that no one noticed this in the past 10 months since the change was made, but either way it currently doesn't work for me.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):** there is nothing relevant in the logs

**Screenshot of problem:** n/a

**Additional information:**

 - #2821 could be related - if the app doesn't implement language switching (as far as I can find), it is weird to see it switching language by itself
 - There has been [some discussion about this on Discord in #devs_mobile_apps](https://discord.com/channels/330944238910963714/377667559303938050/1013726264512761877)

jpelgrom:
## Summary

Switch to the new AndroidX features for app-specific language support and integration with Android 13's native app language setting. 

In addition to the setting not doing anything one of the issues with the previous implementation was detecting/selecting languages when using the app bundle (Play Store releases) only showing languages selected in the system settings, as it only downloaded splits for what was selected there. I opted to disable the language splits instead of integrating [Google Play Core additional feature delivery (for language resources)](https://developer.android.com/guide/playcore/feature-delivery/on-demand#lang_resources) as it improves the experience of language switching (it is a lot faster and offers a consistent experience for all Android versions and flavors) and prevents a lot of additional complexity for very little gain (all language files add up to ~1.5MB, Play Core dependencies would also increase the size and require handling downloads/unavailability/errors/flavors).

Fixes #2832, probably also fixes #2821 as the app language is now explicitly set.

## Screenshots

|     |Light|Dark|
|-----|-----|-----|
|In-app language picker|A dialog in the app titled 'Language', with the option 'Device language' and a long list of languages, light mode|A dialog in the app titled 'Language', with the option 'Device language' and a long list of languages, dark mode|
|Android 13 language picker|Android system settings showing a screen titled 'App Language' with the Home Assistant app name and icon and a list of languages, light mode|Android system settings showing a screen titled 'App Language' with the Home Assistant app name and icon and a list of languages, dark mode|

n/a

## Any other notes

The Android 13 picker will show the languages provided by the app, but also all possible countries/regions for those languages. If the user chooses a specific country for a language while the app doesn't specify it (for example Spanish (Mexico) is selected in the system settings, but the app only says it supports Spanish), the in-app picker won't reflect the selected language to prevent changing the setting or overriding it in case the specific country is added in the future.

The frontend appears to keep the initial language that is used/heavily cache the language, so it might not change after changing the language in the app settings. I've verified that the `Accept-Language` header is updated, so I think this is a deliberate choice by the frontend team.

For testing you'll need to manually add string files and update `common/src/main/res/xml/locales_config.xml` as the languages aren't added to PR builds.


dshokouhi:
> The frontend appears to keep the initial language that is used/heavily cache the language, so it might not change after changing the language in the app settings.

Indeed the frontend does very heavy caching not just here but in all aspects to help with faster loading times. We had other bugs about similar behavior.

dshokouhi (approved):
Tested things out on my end and I don't see any issues here, nice to see the new android 13 language selection added too!

JBassett (approved):

PR #3081, opened 2022-11-18
Skip heart rate value updates when below 0

dshokouhi:
## Summary

We had a user mention on [discord](https://discord.com/channels/330944238910963714/562408603345092636/1042882512789176320) that their heart rate sensor will show a value of `-1` so lets skip any value below `0` as we are measuring in `bpm` so `0` will be the lowest acceptable value.

Hopefully we don't have to account for a crazy high value 😆 


jpelgrom (approved):
Considered if we should send `unavailable` as a state instead of skipping, but a negative value isn't that different from unreliable/no contact states which are already skipped. If someone wants to know if the value is 'current' the on body sensor can be used.

JBassett (approved):

PR #3087, opened 2022-11-19
Reduce usage of job scheduling to avoid throttling

dshokouhi:
## Summary

I have been noticing at times when the app is not on the doze whitelist that sometimes the sensor worker will not start/update, even when we open the app.

WorkManagers `doWork()` function calls `JobScheduler.schedule()` I noticed the following note in the docs:

`
Note: Scheduling a job can have a high cost, even if it's rescheduling the same job and the job didn't execute, especially on platform versions before version [Build.VERSION_CODES.Q](https://developer.android.com/reference/android/os/Build.VERSION_CODES#Q). As such, the system may throttle calls to this API if calls are made too frequently in a short amount of time.
`

Sure enough when the issue was happening on my Pixel Watch the worker calls were always immediately being canceled up until I would force stop the app.  I believe on a wearable device this issue may be more prominent  as this throttling probably happens at a faster interval.

So in this PR I am looking at all of our usage of `SensorWorker.start()` and replacing it with an intent based update so we don't `schedule` our job so often so we can respect the limit imposed above.

- Any sensor that causes an update will use this new method.
- We will start the worker upon app launch `onResume` but when we exit the app we will call the new method. This is to prevent possible throttling if the app were open and put in the background several times causing too many schedules.
- When we enable sensors we will also call this new method to prevent throttling in case many sensors were added back to back.

I don't really know what qualifies the throttling but in my opinion we should really just schedule the job and let it do its thing, instead of constantly rescheduling it. We probably never noticed the above issue as we always keep the app on the whitelist, especially if using location tracking.

I have tested the phone side and updates still happen as expected. I am in the process of testing on my Pixel Watch to see if there are any issues so leaving as a draft to get feedback.  Please let me know if you think I have removed too many instances, my idea was to use the worker sparingly so we never get throttled.

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/ActivitySensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/ActivitySensorManager.kt
@@ -135,7 +135,7 @@ class ActivitySensorManager : BroadcastReceiver(), SensorManager {
                 )
 
                 // Send the update immediately
-                SensorWorker.start(context)
+                SensorReceiver.updateAllSensors(context)
             }
         }
         if (SleepSegmentEvent.hasEvents(intent) && isEnabled(context, sleepSegment.id)) {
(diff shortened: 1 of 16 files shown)


dshokouhi:
Been testing this PR on the Pixel Watch with the app not on the doze whitelist and the sensor worker is still running as expected instead of stopping overnight. This looks good to me now as I no longer saw the issue of the worker being throttled. Marking as ready for review :)

JBassett (approved):

PR #3092, opened 2022-11-22
Update a bunch of Github Actions

JBassett:
## Summary

We are using several deprecated Github Actions, this should resolve those issues.

## Screenshots

N/A

N/A

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -112,6 +112,6 @@ jobs:
         fi
     
     - name: Archive Debug Build
-      uses: kittaakos/upload-artifact-as-is@v0
+      uses: actions/upload-artifact@v3
       with:
         path: ./**/*.apk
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yml
```diff
@@ -112,6 +112,6 @@ jobs:
         fi
     
     - name: Archive Debug Build
-      uses: kittaakos/upload-artifact-as-is@v0
+      uses: actions/upload-artifact@v3
```
jpelgrom, line 115: The PR flow now includes one file - artifact, 82 MB - instead of all APKs

jpelgrom: It's less convenient, but not a major problem. Just wanted to point it out :)

JBassett: That action was archived and throwing warnings so figured it was safer to use a more default one.

dshokouhi (approved):
Looks good to me, too bad there is no action to expand the artifact 😛

Issue #3093, opened 2022-11-23
Weird Spacing in Button Widget


tanishqmanuja:
**Is your feature request related to a problem? Please describe.**

The spacing/margin/padding in button widget seems very unnatural to look at. Although it may not impact the working anyhow but significantly drops the polished impression we usually expect from great apps.

**Describe the solution you'd like**

- The top and bottom padding/margin should be equal.
- If the Label text is empty string i.e. "", It should not be rendered or 'android:visibility' should be adjusted. Then the icon will be centered in the widget.

**Describe alternatives you've considered, if any**

N/A

**Additional context**

![a](https://user-images.githubusercontent.com/40914272/203506493-8a783105-12b8-4eec-88c9-7f276e9a8ddd.jpg)
![b](https://user-images.githubusercontent.com/40914272/203506523-3363912d-22fc-4bfd-93bf-041489837d42.jpg)

jpelgrom:
I agree these don't look great everywhere. The preview images for older Android versions do look like they are correctly spaced, so I'm guessing different screen sizes / no text wasn't considered originally 😉

jpelgrom:
Playing with this:

 - Centering the icon when there is no label should be relatively easy to do.
 - "Equal spacing" when there is a label is a bit more difficult. With your example it'd look OK, but what if the device uses a taller grid? Should there be a larger gap in the middle (1), or should a larger top/bottom padding be used (2)? Especially on tablets even the smallest 1x1 widget can already be quite large. The current solution (3), while not perfect, at least comes somewhat close to the layout of an app icon and label. Example for demonstration, with exaggerated height:

![A screenshot showing widgets with layouts 1, 2, and 3 from the text above](https://user-images.githubusercontent.com/8148535/203639598-7870061a-04a0-4741-8c1f-88c9856e146e.png)

tanishqmanuja:
As per UX concerns option (2) should be out of the question because the widgets support transparent background also, this makes apparently invisible touch spots.
For the same reason option (1) is scores the highest on UX as user knows the touch area perfectly, but option (3) looks the best in my opinion 😁.

tanishqmanuja:
> Playing with this:
> 
>  - Centering the icon when there is no label should be relatively easy to do.
[...]

PR #3099, opened 2022-11-25
Button Widget: Uniform Spacing

tanishqmanuja:
## Summary

Button widget had asymmetrical top and bottom spacing. This PR tries to fix that.
Related issue: https://github.com/home-assistant/android/issues/3093

## Screenshots

Portrait:
image

Landscape:
image

## Any other notes

- The icon named old is from stock Home Assistant App kept for comparison with the newer ones.
- The only small downside of this approach is that on multiline text the icon becomes slightly smaller as compared to stock multi line icon (BTW stock multiline widget is not present in the screenshots above)

In order to fully solve this only downside with even better but complex approach is to use multiple layouts:
https://developer.android.com/develop/ui/views/appwidgets/layouts#provide-responsive-layouts

### DIFF
--- a/app/src/main/res/layout/widget_button.xml
+++ b/app/src/main/res/layout/widget_button.xml
@@ -14,6 +14,7 @@
         android:id="@+id/widgetImageButtonLayout"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
+        android:paddingVertical="4dp"
         android:orientation="vertical">
 
         <ImageView
@@ -31,18 +32,17 @@
             android:id="@+id/widgetLabelLayout"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
-            android:layout_margin="2dp"
-            android:minHeight="32sp">
+            android:layout_margin="2dp">
 
             <TextView
                 android:id="@+id/widgetLabel"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:layout_gravity="center"
+                android:layout_gravity="bottom|center_horizontal"
                 android:text="@string/widget_label_placeholder_text_button"
                 android:textSize="12sp"
                 android:textAlignment="center"
-                android:gravity="center"
+                android:gravity="top|center_horizontal"
                 android:textColor="?colorWidgetOnBackground"
                 android:minLines="1"
                 android:maxLines="2" />
--- a/app/src/main/res/layout/widget_button.xml
+++ b/app/src/main/res/layout/widget_button.xml
@@ -38,11 +38,11 @@
                 android:id="@+id/widgetLabel"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:layout_gravity="bottom|center_horizontal"
+                android:layout_gravity="center"
                 android:text="@string/widget_label_placeholder_text_button"
                 android:textSize="12sp"
                 android:textAlignment="center"
-                android:gravity="top|center_horizontal"
+                android:gravity="center"
                 android:textColor="?colorWidgetOnBackground"
                 android:minLines="1"
                 android:maxLines="2" />
--- a/app/src/main/res/layout/widget_static.xml
+++ b/app/src/main/res/layout/widget_static.xml
@@ -46,12 +46,12 @@
                 android:id="@+id/widgetLabel"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:layout_gravity="bottom|center_horizontal"
+                android:layout_gravity="center"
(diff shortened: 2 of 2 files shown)


jpelgrom:
This still isn't quite equal spacing, unless I'm missing something. The root view `widgetLayout` has 4dp padding all around, and the label view `widgetLabelLayout` has 2dp padding all around, so in total there is now 4dp padding at the top but 6dp padding at the bottom.

There is also an additional 4dp vertical padding for `widgetImageButtonLayout` - why is that? It contributes to the smaller icon you're describing.

The entity state widget uses very similar spacing for it's elements, if the button widget is updated I feel like that one should be too in this PR to keep their layouts aligned.

In combination with your previous PR, should this close the issue?

tanishqmanuja:
Maybe not perfectly symmetrical (with difference of 2dp that too only on some devices).

Top Spacing -> `4dp` from Widget Layout + `4dp` from Vertical padding of Linear Layout
Bottom Spacing -> `4dp` from Widget Layout + `4dp` from Vertical padding of Linear Layout + `2dp` from margin of Label Layout (to protect text character like 'j', 'g' which in my experience leaks on user fonts in Custom ROMs)

TLDR:
 - So the original 4dp from Widget Layout was already there before my PR, so I didn't touch it. I added extra 4dp to make total 8dp which is the `Widget Corner Radius`.
- `In combination with your previous PR, should this close the issue?` Yess :)

PS:
2dp not adding to bottom spacing when enough space is provided according to layout bounds.
image

tanishqmanuja:
@jpelgrom, Regarding the entity state widget what should I do ?
BTW in my testing I found stock widget is already broken in multiline scenarios for landscape mode.

tanishqmanuja:
Commit https://github.com/home-assistant/android/pull/3099/commits/f3a964e031aefa806b74f838949f1553d228027a,
Mirror the changes from button widget to entity state widget.

Screenshot:
image

jpelgrom:
> I added extra 4dp to make total 8dp which is the `Widget Corner Radius`.

The widget corner radius, especially when using the Dynamic color theme like you do, will depend on the widget theme, Android version, launcher, device and/or manufacturer. If you want to use the widget corner radius (which can be >8dp), you can use `?dimenWidgetCornerRadius`.

> Regarding the entity state widget what should I do ?

I think changes to mirror the button widget are good 👍

> I found stock widget is already broken in multiline scenarios for landscape mode

There is a big chance it will be too small when using a modern phone, but for less tall devices / tablets it is less of a problem. Right now the text size is configurable by users so it's not easily solved without adding more options, changing it to a less exact option like 'small/medium/large' or adding a completely different layout, which is IMO not in the scope for this PR.

(Adding more padding like you've done _does_ increase the chances of there not being enough space for the text.)

tanishqmanuja:
So is the PR complete or something more should be done?

app/src/main/res/layout/widget_button.xml
```diff
@@ -31,18 +32,17 @@
...
                 android:id="@+id/widgetLabel"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:layout_gravity="center"
+                android:layout_gravity="bottom|center_horizontal"
                 android:text="@string/widget_label_placeholder_text_button"
                 android:textSize="12sp"
                 android:textAlignment="center"
-                android:gravity="center"
+                android:gravity="top|center_horizontal"
```
jpelgrom, line 45: As the TextView height is `wrap_content`, aligning it to the top doesn't make any difference so this line doesn't need to be changed.

app/src/main/res/layout/widget_button.xml
```diff
@@ -31,18 +32,17 @@
...
-            android:layout_margin="2dp"
-            android:minHeight="32sp">
+            android:layout_margin="2dp">
 
             <TextView
                 android:id="@+id/widgetLabel"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:layout_gravity="center"
+                android:layout_gravity="bottom|center_horizontal"
```
jpelgrom, line 41: As the parent LinearLayout height is `wrap_content` and no longer has a minimum height, aligning it to the bottom doesn't make any difference so this line doesn't need to be changed.

app/src/main/res/layout/widget_button.xml
tanishqmanuja: Thanks for mentioning

jpelgrom (approved):

jpelgrom:
Not uniform but close enough when there is a single line of text 😉

JBassett (approved):

PR #3104, opened 2022-11-25
Fix current version sensor so it properly reports the flavor

dshokouhi:
## Summary

Just realized that the current version sensor was impacted by the move to the `common` module and no longer reports the flavor but instead reports the version code. This PR fixes this by moving the update function to the appropriate location so the flavor is pulled correctly.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/AppSensorManagerBase.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/AppSensorManagerBase.kt
@@ -136,7 +136,23 @@ abstract class AppSensorManagerBase : SensorManager {
         }
     }
 
-    abstract fun updateCurrentVersion(context: Context)
+    abstract fun getCurrentVersion(): String
+
+    private fun updateCurrentVersion(context: Context) {
+
+        if (!isEnabled(context, currentVersion.id))
+            return
+
+        val state = getCurrentVersion()
+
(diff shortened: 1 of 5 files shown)


common/src/main/java/io/homeassistant/companion/android/common/sensors/AppSensorManagerBase.kt
```diff
@@ -137,21 +136,7 @@ class AppSensorManager : SensorManager {
         }
     }
 
-    private fun updateCurrentVersion(context: Context) {
-
-        if (!isEnabled(context, currentVersion.id))
-            return
-
-        val state = BuildConfig.VERSION_NAME
```
jpelgrom, line 145: This is the only part that is unique to the module/flavor, so to limit duplicate code I propose not removing the entire function here but getting this value from a function

```kotlin
// in updateCurrentVersion
val state = getCurrentVersion()

// defined as
abstract fun getCurrentVersion(): String
```

which is implemented in the app/wear `AppSensorManager`s:

```kotlin
override fun getCurrentVersion(): String = BuildConfig.VERSION_NAME
```

dshokouhi: that is a lot cleaner, thank you 🙏

jpelgrom (approved):

JBassett (approved):

PR #3105, opened 2022-11-26
Add deeplinks to Wear OS sensor sync notifications

jpelgrom:
## Summary

When enabling a sensor on the server that requires additional permissions, the app will post a notification to let you know that it needs to be enabled from the app. The main app added an intent to that notification to open the sensor's settings when tapped, this PR adds a similar intent to the Wear OS app.

## Screenshots

The notification that is posted when a sensor requires additional permissions (this isn't new):
A notification on a watch by the Home Assistant app titled 'WiFi State', description 'To enable this sensor, turn it on in the app set...'

It now includes an intent which will show up as a 'Open on watch' button (using the Wear OS 3 emulator):
A notification on a watch by the Home Assistant app telling the user that the WiFi State sensor needs to be enabled in the app, with buttons labeled 'Clear' and 'Open on watch'

Tapping on it will open the sensor manager's settings screen where the sensor can be enabled and permission(s) can be granted.

n/a

## Any other notes

Testing note: after launching the app from the notification, the app in the recents screen seems to use this intent for subsequent launches until launched from the main app list, which is technically correct but might be unexpected.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
@@ -2,6 +2,7 @@ package io.homeassistant.companion.android.sensors
 
 import android.annotation.SuppressLint
 import android.app.NotificationManager
+import android.app.PendingIntent
 import android.bluetooth.BluetoothAdapter
 import android.content.Context
 import android.content.Intent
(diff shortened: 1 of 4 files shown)


dshokouhi (approved):
Nice improvement!!

JBassett (approved):

PR #3115, opened 2022-11-29
Migrate more sensors to Wear OS

dshokouhi:
## Summary

Add more sensors to the Wear OS app (if supported):

- Display Sensorss
- Keyguard sensors
- Last reboot
- Light sensor
- Mobile data
- Phone state
- Pressure
- Proximity
- Storage
- Time zone
- Traffic stats

Documentation: home-assistant/companion.home-assistant#866

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
@@ -14,13 +14,24 @@ import io.homeassistant.companion.android.BuildConfig
 import io.homeassistant.companion.android.common.sensors.AudioSensorManager
 import io.homeassistant.companion.android.common.sensors.BatterySensorManager
 import io.homeassistant.companion.android.common.sensors.DNDSensorManager
+import io.homeassistant.companion.android.common.sensors.DisplaySensorManager
+import io.homeassistant.companion.android.common.sensors.KeyguardSensorManager
+import io.homeassistant.companion.android.common.sensors.LastRebootSensorManager
 import io.homeassistant.companion.android.common.sensors.LastUpdateManager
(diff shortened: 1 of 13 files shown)


jpelgrom:
So many sensors...

It's hard for me to determine if these sensors are all reasonable and work correctly, but one I can think of: the phone state sensor will show if there is an incoming call, which can be true on a watch even if it does not have cellular connectivity itself (`PackageManager.FEATURE_TELEPHONY`). Does it work on non-cellular watches, or do all watches have this feature?

Internal/external storage are both available on the emulator, is this a weird emulator bug or does this also show on real watches? External is supposed to be SD cards according to our docs which no watches support.

dshokouhi:
> the phone state sensor will show if there is an incoming call, which can be true on a watch even if it does not have cellular connectivity itself (`PackageManager.FEATURE_TELEPHONY`). Does it work on non-cellular watches, or do all watches have this feature?

on my non-LTE pixel watch it does not show up. Ringing can be taken from the audio mode sensor in this case :)



> Internal/external storage are both available on the emulator, is this a weird emulator bug or does this also show on real watches? External is supposed to be SD cards according to our docs which no watches support.

This might be a bug in the existing sensor logic because its there on a Pixel phone as well. I thought we had a check to see if its available but seems we are including this sensor no matter what all the time 😬

dshokouhi:
> This might be a bug in the existing sensor logic because its there on a Pixel phone as well. I thought we had a check to see if its available but seems we are including this sensor no matter what all the time 😬

So this is not a bug, we actually mention in our docs that if the device has no SD card we will mention it.

https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-app-battery-drain

JBassett (approved):

Issue #3120, opened 2022-11-29
Quick Settings (Tiles) state not refreshing properly


icegothic:
**Home Assistant Android version:** 
2022.11.0-full

**Android version:** 
13 (or 12)

**Phone model:** 
Galaxy S22

**Home Assistant version:** 
2022.11.4

**Last working Home Assistant release (if known):** 
It never worked

**Description of problem:** 
The state of the Quick Settings (Tiles) seems not to refresh properly (not changing color) when tapped. 

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**
[homeassistant_companion_log_10-29-2022_11-20-21.txt](https://github.com/home-assistant/android/files/10115003/homeassistant_companion_log_10-29-2022_11-20-21.txt)

```

```

**Screenshot of problem:**
Here's a video of the issue, We can see that the state updates well when when we turn off the light, but it does not update when turning on unless we reopen the panel

https://user-images.githubusercontent.com/16883752/204583356-0248dd2e-0655-44e4-89ce-3e5d6d614f13.mp4

**Additional information:**
Notes: The light is toggling properly but the state (color) of the tile is not refreshed; we have to close and reopen the QuickSettings panel to see the real state of the device on the tile. Note: I noticed that Tasker can do it with QuickTask tiles (https://play.google.com/store/apps/details?id=com.balda.quicktask) but they are not as reactive as HA Companion tiles.

dshokouhi:
why was the issue template removed here? All the requested information is important especially the part about logs. The active state of the tile will update however its possible something failed.

icegothic:
Hi, sorry for the template; I fixed it and added a video of the issue and the log file. Is there a way I can provide more information?

icegothic:
How can I make it more obvious so the state of the tile update when the lights turns on?  (sorry I'm kinda new to Home Assistant)

dshokouhi:
have you considered using device controls? They are identical to tiles but way more interactive as they let you control the brightness level etc...

icegothic:
That seems nice but I never could find the option to activate/access those... :-(

icegothic:
I'll check it out and see if it fits my needs :-) thank you for the suggestion

dshokouhi:
There is a quick setting tile called "Device Controls" you can get to it from there. Samsung may have also moved things around.

https://companion.home-assistant.io/docs/integrations/android-device-controls

[...]

PR #3121, opened 2022-11-29
Update quick setting tiles with state updates

dshokouhi:
## Summary

Fixes: #3120 by switching to state change trigger subscription for the tile. We are still doing an initial state update to make sure the state is current when the tile is first in view.  These tiles will now always have its active state in sync with the state of the entity, even when interacting with them.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/qs/TileExtensions.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/qs/TileExtensions.kt
@@ -25,8 +25,8 @@ import io.homeassistant.companion.android.database.qs.TileDao
 import io.homeassistant.companion.android.database.qs.TileEntity
 import io.homeassistant.companion.android.database.qs.isSetup
 import io.homeassistant.companion.android.settings.SettingsActivity
-import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
@@ -50,7 +50,7 @@ abstract class TileExtensions : TileService() {
 
     private val mainScope = MainScope()
 
-    private var stateUpdateJob: CoroutineScope? = null
+    private var stateUpdateJob: Job? = null
 
     override fun onClick() {
         super.onClick()
@@ -88,12 +88,11 @@ abstract class TileExtensions : TileService() {
     override fun onStartListening() {
         super.onStartListening()
         Log.d(TAG, "Tile: ${getTileId()} is in view")
-        stateUpdateJob = MainScope()
         getTile()?.let { tile ->
             mainScope.launch {
                 setTileData(getTileId(), tile)
             }
-            stateUpdateJob?.launch {
+            stateUpdateJob = mainScope.launch {
                 val tileData = tileDao.get(getTileId())
                 if (tileData != null && tileData.isSetup && tileData.entityId.split('.')[0] in toggleDomainsWithLock)
                     integrationUseCase.getEntityUpdates(listOf(tileData.entityId))?.collect {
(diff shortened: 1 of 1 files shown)


jpelgrom:
This works initially for me, but not if I retry it later. Android probably doesn't remove the tile service which means `mainScope` is never reset. To reproduce: add tile from app, open quick settings, tap to toggle works, close quick settings, re-open quick settings, tap to toggle no longer works.

`toggleDomains.plus("lock")` is now used in 3 places, it should probably be put in a variable to make it easier to update in the future.

dshokouhi:
> Android probably doesn't remove the tile service which means `mainScope` is never reset.

so you never see the log line letting us know the tile listening stopped? Or do you think its caused by canceling all `mainScope` instead of just this one?

jpelgrom:
Log example:

```
[open quick settings]

2022-11-29 22:42:10.017 23416-23416/io.homeassistant.companion.android.debug D/TileExtensions: Tile: tile_1 is in view

[clicking, websocket, updates coming in, ...]
[close quick settings]

2022-11-29 22:42:13.155 23416-23416/io.homeassistant.companion.android.debug D/TileExtensions: Tile: tile_1 is no longer in view

[more websocket: wrapping up and closing]
[open quick settings]

2022-11-29 22:42:16.850 23416-23416/io.homeassistant.companion.android.debug D/TileExtensions: Tile: tile_1 is in view

[no websocket or updates]
```

> Or do you think its caused by canceling all `mainScope` instead of just this one?

I think the scope is shared across all tiles and stays alive for some time after closing quick settings, so it won't work anymore as it has been cancelled.

dshokouhi:
> I think the scope is shared across all tiles and stays alive for some time after closing quick settings, so it won't work anymore as it has been cancelled.

thats strange shouldnt it be starting and stopping when I tell it to?  thought these checks woudl handle it. By using a different scope allows the toggle to continue working but not the state changes. Should I just let the service cancel it when it deems necessary?

jpelgrom:
> thats strange shouldnt it be starting and stopping when I tell it to?

`launch` on a cancelled scope doesn't work, and the scope is only created when the class is initialized.

> Should I just let the service cancel it when it deems necessary?

That sounds reasonable, but in that case we do need to keep track of the subscription jobs by tile ID and cancel those when appropriate, like in `onTileRemoved`.  I'm not sure when the system will remove the class/cancel the scope so better not stay subscribed for longer than required.

dshokouhi:
Hmm then maybe its better that I separate the job and only create the `MainScope()` in `onStartListening` so I can effectively cancel it in `onStopListening` 🤔

dshokouhi:
Yup that seems to have done the trick and now the steps you provided are working as expected :)

app/src/main/java/io/homeassistant/companion/android/qs/TileExtensions.kt
```diff
@@ -49,6 +50,8 @@ abstract class TileExtensions : TileService() {
 
     private val mainScope = MainScope()
 
+    private var stateUpdateJob: CoroutineScope? = null
```
jpelgrom, line 53: This is an entire scope, not a `Job`. Is there a reason not the use the `mainScope` for this, so we are sure that the job inside it is also cancelled when `mainScope` is cancelled for other reasons (like in `onDestroy`)?

This could be `private var stateUpdateJob: Job? = null` and it can be assigned by using `stateUpdateJob = mainScope.launch { ... }`.

dshokouhi: No you are right it should be a `Job` and it works as expected with your suggestion. Thank you 🙏

jpelgrom (approved):
Nice improvement, this also solves the problem with locks potentially showing an unexpected state due to the locking/unlocking state

JBassett (approved):

Lord-Memester:
for me, this has stopped working. I apologize for being unfamiliar with the logging procedure, but I'm curious if anyone else has encountered this recently. Perhaps time for a new issue to be made?

PR #3122, opened 2022-11-29
Hide external storage sensor if device does not have it

dshokouhi:
## Summary

Hide the external storage sensor if it does not exist.
 
## Any other notes

ref: https://github.com/home-assistant/android/pull/3115#issuecomment-1331260740


jpelgrom:
How does this work for devices with an SD card slot, but there isn't always an SD card inserted? Do those devices always have an external storage path or will we be adding/removing the sensor constantly?

I think we need a docs update as well, as this [suggests that the external storage sensor is always available](https://companion.home-assistant.io/docs/core/sensors#storage-sensor):

> for devices without a SD card it will reflect `No SD Card`.

dshokouhi:
Lets just close this and leave it as is since the docs already mention the existing behavior.

Issue #3119, opened 2022-11-29
Quick Setting (Tile) vibration feedback


icegothic:
Hi, 

It would be nice to have a vibration feedback (customizable in duration) when pressing on one of the companion app’s quick setting (tile).

Note: I used to do that with Tasker & QuickTask but Home Assistant Companion tiles are way more reactive.

dshokouhi:
vibration duration does not sound like something we should add, we should just use a success or failure vibration if the user enabled it.

icegothic:
Could you point me to a link explaining how to do this?

dshokouhi:
this was more about implementation details for an app update more than something for the user to do :)

icegothic:
Oh ok :-)

PR #3123, opened 2022-11-29
Add new option to vibrate when quick settings tile is clicked

dshokouhi:
## Summary

Fixes: #3119 by introducing a new per tile setting (off by default) so the user can turn on vibration. As soon as the tile is clicked we will vibrate, then if the call fails we will vibrate again for longer.

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#868

## Any other notes

Leaving as a draft until after #3121  to avoid a conflict. Submitting early to get feedback.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
@@ -10,12 +10,12 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.Button
-import androidx.compose.material.Checkbox
 import androidx.compose.material.Divider
 import androidx.compose.material.DropdownMenu
 import androidx.compose.material.DropdownMenuItem
 import androidx.compose.material.OutlinedButton
 import androidx.compose.material.Scaffold
+import androidx.compose.material.Switch
 import androidx.compose.material.Text
 import androidx.compose.material.TextField
 import androidx.compose.material.rememberScaffoldState
@@ -161,7 +161,7 @@ fun ManageTilesView(
                         text = stringResource(R.string.tile_vibrate),
                         fontSize = 15.sp
                     )
-                    Checkbox(
+                    Switch(
                         checked = viewModel.selectedShouldVibrate,
                         onCheckedChange = { viewModel.selectedShouldVibrate = it }
                     )
--- a/app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
@@ -10,6 +10,7 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.Button
+import androidx.compose.material.Checkbox
 import androidx.compose.material.Divider
 import androidx.compose.material.DropdownMenu
 import androidx.compose.material.DropdownMenuItem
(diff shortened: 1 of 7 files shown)


icegothic:
If I may, is'nt 500ms and 1000ms a little long for vibrations? When I suggested this feature, I was thinking more of a haptic feedback to confirm that I tapped the tile properly. When I used Tasker&QuickTask to manage my tiles, I was using a 5ms to 20ms vibration.

dshokouhi:
> If I may, is'nt 500ms and 1000ms a little long for vibrations?

This matches our haptic feedback durations in our WebView Activity for success and failure.



> When I suggested this feature, I was thinking more of a haptic feedback to confirm that I tapped the tile properly

I would love to use the actual success and failure haptic feedback constants but those are only available inside an activity so we cannot utilize them in quick settings so to make it compatible with all devices we go with a set duration.

icegothic:
Is there a way to make the duration customizable by the user (in ms)?

dshokouhi:
> Is there a way to make the duration customizable by the user (in ms)?

I want to avoid adding so many options and if we do that here then it will be expected to do it in our WebView Activity.

icegothic:
I get that :-) but half a second vibration for every tile tap is way to long for me, I'll end up not using the feature I believe :-(

dshokouhi:
> I get that :-) but half a second vibration for every tile tap is way to long for me, I'll end up not using the feature I believe :-(

if we can find the appropriate vibration length for the haptic constants then it would make sense to update both areas. Its important for the app to be consistent in places. Every user is going ot have their own idea what vibration is too much or too little so it won't be easy to please everyone.

icegothic:
Allright, I'll wait for the release and see what it looks like :-) thanx for your consideration.

dshokouhi:
@icegothic I made a change that should make this more suitable for your use case and others. If the device supports it (which yours does) we will now use a vibration effect that indicates a click. On older devices we will continue to vibrate for longer.

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
```diff
@@ -154,6 +155,17 @@ fun ManageTilesView(
                         }
                     }
                 }
+
+                Row(verticalAlignment = Alignment.CenterVertically) {
+                    Text(
+                        text = stringResource(R.string.tile_vibrate),
+                        fontSize = 15.sp
+                    )
+                    Checkbox(
```
jpelgrom, line 164: Use a `Switch`, checkboxes are for selecting [one or more in a list](https://m3.material.io/components/checkbox/overview)

common/src/main/res/values/strings.xml
```diff
@@ -955,4 +955,5 @@
     <string name="sensor_description_daily_calories">The total number of calories over a day (including both BMR and active calories), where the previous day ends and a new day begins at 12:00 AM local time.</string>
     <string name="sensor_name_daily_steps">Daily Steps</string>
     <string name="sensor_description_daily_steps">The total step count over a day, where the previous day ends and a new day begins at 12:00 AM local time.</string>
+    <string name="tile_vibrate">Should vibrate when tile is clicked?</string>
```
jpelgrom, line 958: Why is this a question and not simply "Vibrate when clicked"?

jpelgrom: It is still a question 😉

Other on/off settings don't have a question mark

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
```diff
@@ -154,6 +155,17 @@ fun ManageTilesView(
...
+
+                Row(verticalAlignment = Alignment.CenterVertically) {
+                    Text(
+                        text = stringResource(R.string.tile_vibrate),
+                        fontSize = 15.sp
+                    )
+                    Switch(
+                        checked = viewModel.selectedShouldVibrate,
+                        onCheckedChange = { viewModel.selectedShouldVibrate = it }
+                    )
```
jpelgrom, line 167: The switch's "thumb" lacks contrast when off and using dark mode:

Setting 'Vibrate when clicked' with a switch, where the switch "thumb"/circle blends into the background

This is solved everywhere else by specifying a different color like this:
```kotlin
colors = SwitchDefaults.colors(uncheckedThumbColor = colorResource(commonR.color.colorSwitchUncheckedThumb))
```

(In light mode it is easier to see due to the shadow)

jpelgrom (commented):
I think the aim was to use the same vibration as clicks in the frontend, but this one feels different on my device (Pixel 4a): much shorter and less buzzing (it's an improvement IMO!).

app/src/main/java/io/homeassistant/companion/android/settings/qs/views/ManageTilesView.kt
dshokouhi: Thank you that does look better!

jpelgrom (approved):

JBassett (approved):

Issue #3130, opened 2022-12-02
Monochrome icon not available


T-bond:
**Home Assistant Android version:**
2022.11.0-minimal

**Android version:**
13

**Phone model:**
SM-G988B

**Home Assistant version:**
2022.11.2

**Last working Home Assistant release (if known):**
-

**Description of problem:**
The monochrome icon is not showing on supported launcher and OS.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```

**Screenshot of problem:**
![SmartSelect_20221202_172603_One UI Home](https://user-images.githubusercontent.com/3778556/205355621-12362469-3379-4a6a-a46f-cfddf89dfe64.jpg)

**Additional information:**
I will submit a fix for it.

PR #3131, opened 2022-12-02
Fix monochrome icon on Android 13+

T-bond:
Signed-off-by: Tamás Nagy (T-bond) <tbondvagyok@gmail.com>

Closes: #3130

## Summary

Fix monochrome icon not showing on supported launcher and OS, according to Android Developers documentation:
[Adaptive icons](https://developer.android.com/develop/ui/views/launch/icon_design_adaptive)

## Screenshots

SmartSelect_20221202_185726_One UI Home

Not relevant

## Any other notes

I did not update the debug icon with the same change, for it to be easily differentiable from the normal application.
(Maybe an updated icon for example with a `DEV` label on it would be sufficent?)


dshokouhi:
> I did not update the debug icon with the same change, for it to be easily differentiable from the normal application.
> (Maybe an updated icon for example with a `DEV` label on it would be sufficent?)

Good question, personally I like it to be red and obviously different as in most cases you dont want to keep a debug build installed all the time. They are usually only meant for testing.

JBassett (approved):

PR #3132, opened 2022-12-03
Add initial Matter commissioning support

jpelgrom:
## Summary

Add initial Matter commissioning support to the app. This support includes two flows:
- initiated by the frontend (note that this flow currently does not seem to be able to complete due to a server-side error)
- as a share target for other apps

Requires up-to-date Play Services and core 2022.12+

This is an initial version to enable beta testing. There are still some rough edges to handle before including this in a stable release, especially related to error handling by the app.

## Screenshots

Commissioning from the frontend (the error at the end is because the server encounters an exception and does not send back a response - if successful Google will show a success screen to enter a name for use with Google's fabric, and the app will resume from where it left off):

https://user-images.githubusercontent.com/8148535/205457312-feff2e98-91ee-4d17-83eb-f3f6feb31566.mp4

Shared to the Home Assistant app from Google Home (preview version):

https://user-images.githubusercontent.com/8148535/205457324-50ab9997-e0e1-4717-8b89-02c6636f7e6d.mp4

n/a


dshokouhi (approved):
Tested debug APK and do not see any issues. I do not have a matter device to test with but I imagine we will be in a bug fixing phase so I think its ok to merge so it can be in the beta push tonight.

app/src/main/java/io/homeassistant/companion/android/matter/MatterCommissioningRequest.kt
```diff
@@ -0,0 +1,11 @@
+package io.homeassistant.companion.android.matter
+
+object MatterCommissioningRequest {
+
+    enum class Status {
```
JBassett, line 5: Why are we wrapping this enum in an object?

jpelgrom: Because originally there was more in the object, and I forgot about this when cleaning it up. I'll remove the object and rename the enum.

JBassett (approved):

PR #3133, opened 2022-12-03
Update change log week of 12/3

dshokouhi:


jpelgrom:
"Subscribe to state updates/supported entities" is quite technical, wouldn't it be enough to say they now update live/less data is used and the app is more responsive?

> Notify user to grant sensor permissions when enabled from frontend

This isn't new, the app already notified but now added the action to go directly to the settings

dshokouhi:
> "Subscribe to state updates/supported entities" is quite technical, wouldn't it be enough to say they now update live/less data is used and the app is more responsive?

yup thats better, went a bit too technical there lol



> This isn't new, the app already notified but now added the action to go directly to the settings

ok updated :)

JBassett (approved):

Issue #1364, opened 2021-02-16
okhttp/4.9.0 user agent is blocked in potential bot crawling lists


e-minguez:
**Home Assistant Android version:**
2021.1.1-full

**Android version:**
10

**Phone model:**
OP5T

**Home Assistant version:**
2021.1.4

**Last working Home Assistant release (if known):**

**Description of problem:**
The user agent used by the android app (okhttp/4.9.0) seems to be pretty generic and it is being blocked in some blacklists (see https://github.com/JayBizzle/Crawler-Detect/issues/437)

I guess it would be better to modify the user-agent with something less generic to avoid being blocked.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```
n/a
```

**Screenshot of problem:**

**Additional information:**

bunkerized-nginx uses [these lists](https://github.com/bunkerity/bunkerized-nginx/blob/master/scripts/user-agents.sh), and it blocks access to home-assistant when trying to log in as:

```
2021/02/16 12:55:07 [warn] 24951#24951: *35 [lua] access_by_lua(main-lua.conf:166):80: [BLOCK] User-Agent okhttp/4.9.0 is blacklisted, client: 10.0.2.100, server: www.example.com, request: "POST /auth/token HTTP/2.0", host: "www.example.com"
```

dshokouhi:
What's wrong with just removing it from blocked list? It is up to every user who uses a block list to ensure it blocks what they want and to modify it for their ow needs.  Not to mention there can be other things dependent on the user agent string such as the HA frontend to understand the android app is talking to it.  I don't think this is something the app should be fixing.

e-minguez:
> What's wrong with just removing it from blocked list? It is up to every user who uses a block list to ensure it blocks what they want and to modify it for their ow needs. Not to mention there can be other things dependent on the user agent string such as the HA frontend to understand the android app is talking to it. I don't think this is something the app should be fixing.

There is nothing wrong with it but I guess it would be better if instead using a generic user agent, the app uses its own u-a so you can identify it properly (something like `home-assistant-android <arch>-<version>`)

dshokouhi:
So when you say its blocked are you seeing the frontend load at all? What parts of the app work and doesn't work? Do sensors update?

e-minguez:
> So when you say its blocked are you seeing the frontend load at all? What parts of the app work and doesn't work? Do sensors update?

[...]

PR #3137, opened 2022-12-05
use custom user-agent in websocket requests

alexbilevskiy:
## Summary

Fix user-agent in websocket requests (such as device controls tile in quick settings), use custom values insteadof `okhttp/4.10.0`  
pretty much the same reason as in #1364 and #1418

## Any other notes

Sorry, I dont have development env, I hope github actions build will work

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
@@ -372,7 +372,7 @@ class WebSocketRepositoryImpl @Inject constructor(
 
             try {
                 connection = okHttpClient.newWebSocket(
-                    Request.Builder().url(urlString).build(),
+                    Request.Builder().url(urlString).header(USER_AGENT, USER_AGENT_STRING).build(),
                     this
                 ).also {
                     // Preemptively send auth


dshokouhi:
I thought we already handled this here for all of our network calls.

https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/data/HomeAssistantApis.kt#L42

jpelgrom:
> I thought we already handled this here for all of our network calls.

Looks like interceptors don't work on websocket requests (also see [this StackOverflow post](https://stackoverflow.com/a/48856490/4214819)). As we're only adding one header the suggested change should be OK in my opinion.

> I hope github actions build will work

It currently doesn't because of actions/setup-java#422, but when that is fixed it still won't work because:

 - you need to add an import statement for the added constants
 - the `USER_AGENT` constant is currently private

I'd recommend editing with an IDE like Android Studio to help you catch errors like these.

dshokouhi:
> > I thought we already handled this here for all of our network calls.
> 
> Looks like interceptors don't work on websocket requests (also see [this StackOverflow post](https://stackoverflow.com/a/48856490/4214819)). As we're only adding one header the suggested change should be OK in my opinion.

In that case this PR is a good find!

alexbilevskiy:
> it still won't work because:

fixed


> I'd recommend editing with an IDE like Android Studio to help you catch errors like these.

Did so. No errors reported.

dshokouhi (approved):
Looks good to me! Nice catch.

JBassett (approved):

PR #3143, opened 2022-12-07
Proof of Concept: Add voice search to Wear OS home screen

dshokouhi:
## Summary

This PR is a proof of concept. I am looking to get some feedback to see if this is something worth investing more time in. If users will find this is something they want to utilize then we should look into a home screen redesign to make it all fit better. For example its tough to decide if this should be above favorite entities or not. With that said lets see what people think about the feature. It will have some bugs but the concept is working so its ready for some feedback :)

As of now its a very basic search. User will click on search which will prompt voice input, chip is only visible after all entities are loaded. The results are then compared against the friendly name attribute for all entities and whatever is present will be shown. If we think search should include areas and domains then I think that should be a separate PR after this one. My idea was that sometimes you may want to look at another entity and dont want to scroll to find it, this should make it easier.

## Screenshots

image

image

image


dshokouhi:
More than likely this PR will be updated after #3158 since that will be a good place to show these results instead of on the home screen.

dshokouhi:
Going to close this for now as the focus is currently on the conversation tile and its corresponding activity.

PR #3158, opened 2022-12-16
Add a tile to start a conversation

dshokouhi:
## Summary

Adds a tile to start a conversation

search_tile_example

Which listens and then transcribes the text to send over to the updated conversation API

Screenshot_20221216_082921

Once confirmed the transcription is sent to the API, and the spoken text and results are displayed in a new activity

Screenshot_20221216_082935

## Any other notes

For testing you need to be on the latest HA core `dev` branch with the `conversation` integration enabled, otherwise the voice prompt will not trigger and a message will be shown notifying the requirements are not met yet.

Probably still have more clean up todo but the code is functional and ready for feedback :)

CC: @balloob @synesthesiam


common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
```diff
@@ -179,6 +180,17 @@ class WebSocketRepositoryImpl @Inject constructor(
         }
     }
 
+    override suspend fun getConversation(speech: String): ConversationResponse? {
+        val socketResponse = sendMessage(
+            mapOf(
```
balloob, line 185: For future, do we have ability to send the language of the speech to text / watch. This will be helpful in case the watch language doesn't match the language configured in HA

dshokouhi: Looks like it should be possible using another API once we get the text.

https://developers.google.com/ml-kit/language/identification/android

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
```diff
@@ -179,6 +180,17 @@ class WebSocketRepositoryImpl @Inject constructor(
         }
     }
 
+    override suspend fun getConversation(speech: String): ConversationResponse? {
```
balloob, line 183: For future PR, you get a `conversastion_id` back (`str | null`) in `ConversationResponse`. As long as the activity is open, it would be nice if we can pass that back in. Some conversation agent in the future will be able to do dialogues.

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
JBassett: I would think that whatever language the app is using would be enough?  I don't think we need the ML stuff for that.

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/entities/ConversationSpeechResponse.kt
```diff
@@ -0,0 +1,12 @@
+package io.homeassistant.companion.android.common.data.websocket.impl.entities
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class ConversationSpeechResponse(
```
balloob, line 6: before merging this, we need to update the websocket response to align with the Rest response.

The Rest response has wrapped this in a dict like this https://github.com/home-assistant/frontend/blob/dev/src/data/conversation.ts#L47

wear/src/main/AndroidManifest.xml
```diff
@@ -102,6 +105,18 @@
             <meta-data android:name="androidx.wear.tiles.PREVIEW"
                 android:resource="@drawable/template_tile_example" />
         </service>
+        <service
+            android:name=".tiles.SearchTile"
```
balloob, line 109: Should we name it `ConversationTile` ?

dshokouhi: yea let me replace `Search` with `Conversation` throughout to make it easier :)

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/entities/ConversationSpeechResponse.kt
dshokouhi: ok and that will be a change in HA core which the app will need to reflect too correct?

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
balloob: Yeah that should definitely be enough.

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/entities/ConversationSpeechResponse.kt
balloob: yes.

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
dshokouhi: oh I was thinking we needed to detect the language spoken which may not match what language the watch is in?

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/entities/ConversationSpeechResponse.kt
balloob: nevermind, I looked at the wrong model. this is ok.

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
balloob: we need to just get the language that the text is in. Which is based on what the users keyboard is set to I guess. But we can assume 95% certain that it matches the app language

balloob (approved):

jpelgrom (commented):
> Probably still have more clean up todo

Not sure if you want to do it in this PR or in a followup, but here are comments for clean up :)

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
jpelgrom: > oh I was thinking we needed to detect the language spoken which may not match what language the watch is in?

The docs [suggest](https://developer.android.com/reference/android/speech/RecognizerIntent#EXTRA_LANGUAGE) that it will always recognize speech based on the language returned by `Locale.getDefault()`.

wear/build.gradle.kts
```diff
@@ -123,6 +123,7 @@ dependencies {
 
     implementation("com.google.guava:guava:31.1-android")
     implementation("androidx.wear.tiles:tiles:1.1.0")
+    implementation("androidx.wear.tiles:tiles-material:1.1.0")
```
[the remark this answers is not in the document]
dshokouhi: ah I added this when I was thinking of trying to use a `chip` or `button` but then settled on the `box` :) Will remove this.

wear/src/main/java/io/homeassistant/companion/android/conversation/views/ConversationView.kt
```diff
@@ -0,0 +1,57 @@
...
+import androidx.wear.compose.material.Text
+import androidx.wear.compose.material.rememberScalingLazyListState
+import io.homeassistant.companion.android.common.R
+import io.homeassistant.companion.android.conversation.ConversationViewModel
+import io.homeassistant.companion.android.home.views.TimeText
+import io.homeassistant.companion.android.theme.WearAppTheme
+import io.homeassistant.companion.android.views.ThemeLazyColumn
+
+@Composable
+fun SearchResultView(
```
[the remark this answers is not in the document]
dshokouhi: ah good catch :) will rename this

common/src/main/res/values/strings.xml
```diff
@@ -965,4 +965,8 @@
     <string name="sensor_name_daily_steps">Daily Steps</string>
     <string name="sensor_description_daily_steps">The total step count over a day, where the previous day ends and a new day begins at 12:00 AM local time.</string>
     <string name="tile_vibrate">Vibrate when clicked</string>
+    <string name="no_results">No results yet</string>
+    <string name="speak">Speak</string>
+    <string name="no_conversation_support">You must be at least on Home Assistant 2023.01 and have the conversation integration enabled</string>
```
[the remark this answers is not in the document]
dshokouhi: 🙈  will update this

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
dshokouhi: I will add a `TODO` to keep track of this request :)

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
dshokouhi: Thanks all :) I will add a `TODO` note for this so we can take care of this in future

wear/src/main/java/io/homeassistant/companion/android/conversation/ConversationViewModel.kt
```diff
@@ -0,0 +1,40 @@
...
+import javax.inject.Inject
+
+@HiltViewModel
+class ConversationViewModel @Inject constructor(
+    application: Application,
+    private val integrationUseCase: IntegrationRepository,
+    private val webSocketRepository: WebSocketRepository
+) : AndroidViewModel(application) {
+
+    var speechResult = mutableStateOf("")
```
jpelgrom, line 20: By using `by mutableStateOf("")` instead of `= mutableStateOf("")` you can stop using `.value` to access/update these variables.

(This will break updating the speech result from the activity, which shouldn't have been possible in the first place and should use a function in the viewmodel.)

dshokouhi: ok done, updated all variables and now updating the results in a function instead of the variable directly

jpelgrom: Now that you're using a function why remove `private set` for `speechResult`?

dshokouhi: Somewhere along the way android studio asked me to make it public, I will add it back.

jpelgrom (approved):
Not seeing any major issues, looks good for an MVP to allow others to start testing.

JBassett (approved):

Issue #2183, opened 2022-01-22
Add iBeacon capability to the Wear OS app


rpatel3001:
**Is your feature request related to a problem? Please describe.**

Room presence is a fun addition to a home automation setup, and the phone app beacon has been working well for me, but I sometimes leave my phone in a different room. Tracking a watch that's always on my wrist would be nice.

**Describe the solution you'd like**

An additional setting for the companion app to enable iBeacon advertisements, similar to the phone app.

**Describe alternatives you've considered, if any**

I'm going to tinker with writing my own app as the APIs for this don't seem too convoluted, but I've not done it before and it's not scalable.

**Additional context**

dshokouhi:
All current phone sensors will be evaluated once the wear OS app has a sensor UI for enabling and adding settings.  Will come sometime after #2033

rpatel3001:
Got it, thanks

rpatel3001:
If anyone else is interested in this and lands at this issue, I have a working proof of concept at https://github.com/rpatel3001/BeaconWear

Credit to the good folks at @AltBeacon, this was forked from https://github.com/davidgyoung/android-beacon-library-reference-kotlin and modified with code from https://altbeacon.github.io/android-beacon-library/beacon-transmitter.html.

There's still a ton of cruft for scanning and absolutely no user interface, but I'll probably clean it up eventually.

wifiuk:
be handy to have beacon capability on the home assistant wear os app for sure.

maartendamen:
Would love to see this in comination with ESPresense. 
@rpatel3001 can you please explain a bit more about your PoC. Can I use it on my GW4 by just compiling and installing it?

rpatel3001:
@maartendamen there is an apk available [here](https://github.com/rpatel3001/BeaconWear/releases/tag/first release). The UI doesn't actually do anything, the UUID and reference power are hard coded. You can clone the repo and build in Android Studio if you need to change those values.

maartendamen:
> @maartendamen there is an apk available [here](https://github.com/rpatel3001/BeaconWear/releases/tag/first_release). The UI doesn't actually do anything, the UUID and reference power are hard coded. You can clone the repo and build in Android Studio if you need to change those values.

Very cool, I got it working. Thanks for your efforts!
I will do some writing about this on my blog soon if you don't mind.

rpatel3001:
> Very cool, I got it working. Thanks for your efforts! I will do some writing about this on my blog soon if you don't mind.

[...]

PR #3168, opened 2022-12-20
Migrate bluetooth sensors to common and add to wear OS

dshokouhi:
## Summary

Migrates over the bluetooth sensors so they can be used in Wear OS.

I know that previously I had mentioned we need to have settings and notifications for these sensors to work properly, I came to notice that a lot of users in the request (#2183) just wanted something to work and use it right away. Given the interest I wanted to see if we can get some testers to do some battery life testing.

Its important to note that I have not tested this on my own watch yet (I plan to do so this weekend or next week but maybe sooner), testing was done on the phone but the code should still work on Wear OS so would be good to get some early feedback on this. This will also tell us how badly we would need settings and commands for these sensors or even if a dedicated tile to control the transmitter/monitor will suffice.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -51,6 +51,7 @@ import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.common.data.authentication.AuthenticationRepository
 import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
 import io.homeassistant.companion.android.common.data.url.UrlRepository
+import io.homeassistant.companion.android.common.sensors.BluetoothSensorManager
 import io.homeassistant.companion.android.common.util.cancel
 import io.homeassistant.companion.android.common.util.cancelGroupIfNeeded
 import io.homeassistant.companion.android.common.util.generalChannel
@@ -60,7 +61,6 @@ import io.homeassistant.companion.android.database.notification.NotificationItem
 import io.homeassistant.companion.android.database.sensor.SensorDao
 import io.homeassistant.companion.android.database.settings.SettingsDao
 import io.homeassistant.companion.android.database.settings.WebsocketSetting
-import io.homeassistant.companion.android.sensors.BluetoothSensorManager
 import io.homeassistant.companion.android.sensors.LocationSensorManager
 import io.homeassistant.companion.android.sensors.NotificationSensorManager
 import io.homeassistant.companion.android.sensors.SensorReceiver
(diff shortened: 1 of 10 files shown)


dshokouhi:
Had several users confirm things are working well for them and their intention is to just leave the sensors enabled as the battery drain seemed minimal to them. Marked as ready for review.

Docs have been updated to mention sensor settings are not functional on Wear OS.

jpelgrom (changes requested):
The Wear app manifest should be updated to include the various Bluetooth permissions that may be required (presumably they're granted by default on Wear OS as no one has complained yet, despite the merged manifest only including 2/5?).

dshokouhi:
> The Wear app manifest should be updated to include the various Bluetooth permissions that may be required (presumably they're granted by default on Wear OS as no one has complained yet, despite the merged manifest only including 2/5?).

Good call, I also forgot to register for the BT intents. Will add it all now to match the phone app.

dshokouhi:
> match the phone app.

Actually we should not match the permissions as we are not targeting SDK 33 so some of the new BT permissions do not exist yet. I realized this because the sensors that are working are not getting enabled because the app does not need all the new BT restrictions just yet. Will adjust for the current state of the app and the fact that there are no Wear OS devices running SDK 33. We should only worry about the 3 new permissions once we Wear OS app changes its target SDK.

The phone app already targets SDK 33 so the new permissions and restrictions on old permissions work as expected.

jpelgrom:
> we are not targeting SDK 33 so some of the new BT permissions do not exist yet

The new permissions [are for SDK 31+](https://developer.android.com/guide/topics/connectivity/bluetooth/permissions?hl=en#declare-android12-or-higher) actually. The app targets SDK 32 currently so they should work, a bit unexpected that they don't and impact SDK <31 but then again there aren't actually any watches/emulators with SDK >= 31.

When using the Bluetooth permissions [like in the main manifest](https://github.com/home-assistant/android/blob/afd8a21fbd6a09a1ecc7ed6ec474af3e4e0cfe11/app/src/main/AndroidManifest.xml#L10-L16), sensors do stay enabled for me in the emulator and aren't behaving differently (but that isn't a 100% confirmation as the emulator doesn't really expose Bluetooth data). Are you sure that the sensors not being enabled was because of the `maxSdkVersion` attribute?

dshokouhi:
> When using the Bluetooth permissions [like in the main manifest](https://github.com/home-assistant/android/blob/afd8a21fbd6a09a1ecc7ed6ec474af3e4e0cfe11/app/src/main/AndroidManifest.xml#L10-L16), sensors do stay enabled for me in the emulator and aren't behaving differently (but that isn't a 100% confirmation as the emulator doesn't really expose Bluetooth data). Are you sure that the sensors not being enabled was because of the `maxSdkVersion` attribute?

So what I did for my test was to copy over the 5 BT permissions including the `maxSdkVersion` attribute. The transmitter and monitor sensors did not stay enabled even after granting the permissions. The initially turned on and sent a state but then turned off and stayed off until I made the most recent changes committed here.

dshokouhi:
Ran another test and now its behaving as expected, weird. Will update again.

jpelgrom (approved):

JBassett (approved):

PR #3178, opened 2022-12-28
Add UUID filter to Beacon Monitor

chatziko:
## Summary

The recently added Beacon Monitor is hugely useful, however the fact that it reports __all beacons with no exception__ can lead to excessive traffic (and battery usage). It's quite common to see dozens of beacons throughout a day, of which just a couple might be relevant.

This PR implements a UUID filter to report only selected beacons. I tried to achieve a balance between features and ease-of-use:
- UUIDs are selected from a list containing all currently visible beacons (and those selected in the past, even if they are not currently visible). So one needs to be in proximity to add a beacon to the filter (but on the upside there is no need to manually enter UUIDs, which sucks).

- Only the UUID is matched, not the major/minor. Otherwise beacons that change major/minor would be a problem. Also there is no wildcard support (tricky to implement a nice UI for).

- There is a boolean flag to turn the filter from include (only report selected beacons) to exclude (report all but the selected).

I also fixed a couple of edge-case bugs I discovered:

- If no updates were made, the `skippedUpdated` counter was always being incremented even if a beacon was visible (it would reset to 0 only in case `requireUpdate == true`). So the counter could exceed `MAX_SKIPPED_UPDATED` causing the beacon to be removed the very first time it went missing.
- In the edge case when an update was present every single time, `MAX_SKIPPED_UPDATED` would never be enforced.

Finally the PR contains a tiny bit of polishing and small performance tweaks (given that updates run every second). Still, the overall PR is relatively simple.

I would be happy to modify the code, of course, if someone has a better idea about this feature.

## Screenshots

uuid-filter-2

uuid-filter-1

Documentation: home-assistant/companion.home-assistant#876


app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt
```diff
@@ -21,17 +21,32 @@ data class IBeacon(
 class IBeaconMonitor {
     lateinit var sensorManager: BluetoothSensorManager
     var beacons: List<IBeacon> = listOf()
+    var lastSeenBeacons: Collection<Beacon> = listOf()
+    private var uuidFilter: List<String> = listOf()
```
jpelgrom, line 25:
```suggestion
    private var uuidFilter = listOf<String>()
```

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt
```diff
@@ -21,17 +21,32 @@ data class IBeacon(
...
+    private var uuidFilterExclude: Boolean = false
```
jpelgrom, line 26: This type can be inferred

```suggestion
    private var uuidFilterExclude = false
```

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt
```diff
@@ -82,4 +91,14 @@ class IBeaconMonitor {
         sensorManager.updateBeaconMonitoringSensor(context)
         SensorReceiver.updateAllSensors(context)
     }
+
+    public fun setUUIDFilter(uuidFilter: List<String>, uuidFilterExclude: Boolean) {
```
jpelgrom, line 95: `public` isn't necessary

```suggestion
    fun setUUIDFilter(uuidFilter: List<String>, uuidFilterExclude: Boolean) {
```

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt
```diff
@@ -82,4 +91,14 @@ class IBeaconMonitor {
...
+        this.uuidFilter = uuidFilter
+        this.uuidFilterExclude = uuidFilterExclude
+
+        // existing beacons are only filtered when the filter changes
+        beacons
+            .filter { beacon -> ignoreBeacon(beacon.uuid) }
```
jpelgrom, line 101: This will return a new list, ie the original `beacons` is not updated. As you're not doing anything with the result I think you want to update the original list?

app/src/main/java/io/homeassistant/companion/android/sensors/BluetoothSensorManager.kt
```diff
@@ -301,6 +303,10 @@ class BluetoothSensorManager : SensorManager {
         val scanInterval = getSetting(context, beaconMonitor, SETTING_BEACON_MONITOR_SCAN_INTERVAL, SensorSettingType.NUMBER, DEFAULT_BEACON_MONITOR_SCAN_INTERVAL).toLongOrNull() ?: DEFAULT_BEACON_MONITOR_SCAN_INTERVAL.toLong()
         KalmanFilter.maxIterations = getSetting(context, beaconMonitor, SETTING_BEACON_MONITOR_FILTER_ITERATIONS, SensorSettingType.NUMBER, DEFAULT_BEACON_MONITOR_FILTER_ITERATIONS).toIntOrNull() ?: DEFAULT_BEACON_MONITOR_FILTER_ITERATIONS.toInt()
         KalmanFilter.rssiMultiplier = getSetting(context, beaconMonitor, SETTING_BEACON_MONITOR_FILTER_RSSI_MULTIPLIER, SensorSettingType.NUMBER, DEFAULT_BEACON_MONITOR_FILTER_RSSI_MULTIPLIER).toDoubleOrNull() ?: DEFAULT_BEACON_MONITOR_FILTER_RSSI_MULTIPLIER.toDouble()
+        beaconMonitoringDevice.setUUIDFilter(
+            getSetting(context, beaconMonitor, SETTING_BEACON_MONITOR_UUID_FILTER, SensorSettingType.LIST_BEACONS, "").split(", ").filter { v -> v.isNotEmpty() },
```
jpelgrom, line 307: Instead of `filter { v -> v.isNotEmpty() }`, you could use `filter { it.isNotEmpty() }` to simplify the code.

jpelgrom (commented):
I'm not in a position to really test if these changes work as explained, but do have a few suggestions to make the code more Kotlin-y :)

app/src/main/java/io/homeassistant/companion/android/bluetooth/ble/IBeaconMonitor.kt
chatziko: Indeed, but the new list will contain references to the same objects as the old list, so it will modify the `skippedUpdated` field of those objects. I could do it with loop/if:
```
for (beacon in beacons)
    if (ignoreBeacon(beacon.uuid))
        beacon.skippedUpdated = MAX_SKIPPED_UPDATED
```
but I thought the "higher-order" (filter/forEach) style is a bit cleaner:
```
beacons
    .filter { ignoreBeacon(it.uuid) }
    .forEach { it.skippedUpdated = MAX_SKIPPED_UPDATED }
```

chatziko:
> I'm not in a position to really test if these changes work as explained, but do have a few suggestions to make the code more Kotlin-y :)

Thanks for the feedback, I implemeted the suggestions (first time writing in kotlin, I think I like it :smile:)

jpelgrom:
Code looks good to me now.

@dshokouhi I think you've used the beacon monitor with multiple beacons in the past, would you be able to test this?

dshokouhi:
Spoke too soon, functionally works as described. I am able to get a dynamic list of currently available UUIDs and can either include/exclude them.

I do wonder if we should disable the "Exclude selected UUIDs" toggle when the UUID filter is empty because it doesn't change anything functionally when the list is empty and its toggled. In fact that option really only applies when there is a UUID filter list correct?

chatziko:
> I do wonder if we should disable the "Exclude selected UUIDs" toggle when the UUID filter is empty because it doesn't change anything functionally when the list is empty and its toggled. In fact that option really only applies when there is a UUID filter list correct?

Indeed, it has no effect on an empty list.

I thought about disabling it, but I didn't know whether it's best from a UI perpective (will users understand why it's disabled?). I can surely add it.

dshokouhi:
> I thought about disabling it, but I didn't know whether it's best from a UI perpective (will users understand why it's disabled?). I can surely add it.

We do disable settings for high accuracy mode as the options do not apply until it is enabled. We could also just mention in the sensor description that the option has no effect when the UUID list is empty to avoid more code changes :)

chatziko:
Ok I implemented disabling the option. This involved creating a `CoroutineScope` so have a quick look that I'm doing this properly.

dshokouhi (approved):
Looks good to me, tested the debug APK and it functions as desired. The exclude option is also disabled properly when it cannot be used.

JBassett (approved):
Need to fix merge conflicts, then looks good.

chatziko:
> Need to fix merge conflicts, then looks good.

I fixed the conflicts (they were caused by the recent db63a38cd5fc2b26ecd02f8cb8b157bcd0aa110d)

Issue #791, opened 2020-08-19
Add Android Auto support


TECHNOFAB11:
Inspired by [this](https://www.reddit.com/r/homeassistant/comments/icplnv/running_home_assistant_on_android_auto/) Reddit post, is it possible to add Android Auto functionality?

Developer Guide: https://developer.android.com/training/cars

Problem is that there are only Media and Messaging apps supported, any ideas how to still show the Home Assistant interface? It should probably be less interactive tho, so people don't use it while driving etc. But for now it would suffice if the web interface could be shown, even if only as proof of concept

myhrmans:
Google does not allow that sadly. Hence why I made the guide that I linked in that post. All apps must be allowed by Google and an app like Home Assistant will sadly never be allowed since the possibilities with Home Assistant are endless.

TECHNOFAB11:
Sorry to bother you guys again, but I just wanted to let y'all know that Smarthings works with Android Auto: https://youtu.be/Bk1aT0_PqUk

In my opinion the UI isn't the best and could easily distract the driver. I'm not sure why Google allowed it, but if they allow this then there is a good chance that they'll allow Home Assistant too (you guys are way better at UI than what Smarthings is doing :smiley: )

wrkn4alivn:
See this [blog post](https://android-developers.googleblog.com/2021/04/start-your-engines-launch-new-android.html) and related media coverage.

It seems like there could be a path recently opened up for some Auto functionality for HA. I think there are some use cases that maintain the principle of only allowing driver-relevant interactions, in a minimally distractive way. 

There's no reason to allow unlimited possibilities in the production version of HA Android. An ideal MVP would be a single screen with up to two elements: cover/door (for garage or gate), or camera snapshot.

I would love the Android app even more!

katsaplias:
bump

OliverSkivdal:
Support for only Home Assistant media browser in Android Auto would be amazing if google allows it!

dshokouhi:
Looks like a new article got posted! I guess they will now accept us?

https://developer.android.com/training/cars/apps/iot

GSzabados:
@JBassett, thanks for the answer. And please excuse me for my ignorance, but when is the next beta due? I managed to join beta in the Play Store.

Thanks again!

dshokouhi:
beta gets pushed every Saturday night 9pm PST and then it depends on Play store review, you'll need to have some patience as it may not be available for a few days after the push while we wait for approval.  

[...]

PR #3211, opened 2023-01-11
Initial Android Auto Support

JBassett:
## Summary

This PR will add the base level of Android Auto support.  This should allow several domains to be toggled from a vehicles head unit.
Fixes: https://github.com/home-assistant/android/issues/791

## Screenshots

Screenshot 2023-01-11 at 15 43 44
c3.png)

Documentation: None

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/vehicle/MainVehicleScreen.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/vehicle/MainVehicleScreen.kt
@@ -1,0 +1,95 @@
+package io.homeassistant.companion.android.vehicle
+
+import android.os.Build
+import android.util.Log
+import androidx.annotation.RequiresApi
+import androidx.car.app.CarContext
+import androidx.car.app.Screen
+import androidx.car.app.model.Action
+import androidx.car.app.model.CarColor
+import androidx.car.app.model.CarIcon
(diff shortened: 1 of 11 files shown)


app/src/main/java/io/homeassistant/companion/android/vehicle/MainVehicleScreen.kt
```diff
@@ -0,0 +1,91 @@
...
+                        )
+                    }
+                    .build()
+            )
+        }
+
+        // TODO: Add row for zones so we can start navigation?
+
+        return ListTemplate.Builder()
+            .setTitle("Home Assistant")
```
dshokouhi, line 86: Should use a string here :)

JBassett: Fixed.

dshokouhi (approved):
Although I cannot test the android auto part, the app did launch. I say lets go for initial approval so we can fix whatever else we need to :)

GSzabados:
@dshokouhi, can you please give some advice how this can be tested? Or when is it going to be part of the official release?

JBassett:
So the fun part about Android Auto is the app must be distributed by Google Play in order to be used on actual hardware.  This should be testable by the masses after our next Play Store beta for those who are enrolled.

PR #3221, opened 2023-01-13
Initial commit of notification support on Wear OS

dshokouhi:
## Summary

This PR adds initial support for `message` and `title` only as there is a lot to move and test. The goal here is to get notifications sending to the watch directly as an MVP. 

I tried to keep the changes minimal here but I am sure there will be room for improvements.

```
message: Message
title: Title
```

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#884

## Any other notes

More than likely the next PR after this will be to add device commands for the BLE sensors as those are probably going to be the most important here :)

Its important to note that this will not be testable using the debug APKs as they are from FCM. I have not given too much thought about the warning pop-up we do but given all Wear OS devices will have FCM I don't think it will be too much of a concern. If FCM does not work then functionality will default to existing behavior which is a no-op.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ If you are looking for documentation around the companion applications check out
  - `io.homeassistant.companion.android.minimal`
  - `io.homeassistant.companion.android.minimal.debug`
 
-5. Now download the `google-services.json` file and put it in the _home-assistant-Android/app_ folder. This file contains the configuration of the whole project (all four applications). ([You can also use the mock services file instead of generating your own](/.github/mock-google-services.json). The file should contain client IDs for all packages listed above for debugging to work properly.  **If you do not generate your own file FCM push notification will never work, only websocket notifications will**)
+5. Now download the `google-services.json` file and put it in the _home-assistant-Android/app_ and _home-assistant-Android/wear_ folder. This file contains the configuration of the whole project (all four applications). ([You can also use the mock services file instead of generating your own](/.github/mock-google-services.json). The file should contain client IDs for all packages listed above for debugging to work properly.  **If you do not generate your own file FCM push notification will never work, only websocket notifications will**)
 6. Start Android Studio, open your source code folder and check if the Gradle build will be successful using Build/Make Module "App". You might have to install the right Android SDK via Tools/SDK Manager first.
 7. Run `gradlew assembleDebug` to build all debug versions, this might take a while.
 8. If the build is successful, you can run the app by doing the following: click **Run** -> **Run 'app'**.
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -143,9 +143,6 @@ dependencies {
     implementation("org.altbeacon:android-beacon-library:2.19.5")
     implementation("com.maltaisn:icondialog:3.3.0")
     implementation("com.maltaisn:iconpack-community-material:5.3.45")
-    implementation("com.vdurmont:emoji-java:5.1.1") {
-        exclude(group = "org.json", module = "json")
-    }
 
     implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.20")
     implementation("org.jetbrains.kotlin:kotlin-reflect:1.7.20")
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -168,7 +168,7 @@ dependencies {
 
     "fullImplementation"("com.google.android.gms:play-services-location:21.0.1")
     "fullImplementation"("com.google.android.gms:play-services-home:16.0.0")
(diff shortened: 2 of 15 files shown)


jpelgrom:
I'm not sold on dumping everything in a related package as functions + tons of imports. Especially for the constants putting them in an object (`NotificationData`?) would clean up the usage. Maybe I'm thinking too much in Java but when there's comments above a set of constants that suggests to me that they should be organized as such.

While it is unlikely that Wear will ever get another entrypoint for notifications, why put all the code in `FirebaseMessagingService`? Splitting it up to keep the Firebase code separate from the notification handling code results in a smaller file, making it easier to read and compare with the main app.

dshokouhi:
> Especially for the constants putting them in an object (`NotificationData`?) would clean up the usage. Maybe I'm thinking too much in Java but when there's comments above a set of constants that suggests to me that they should be organized as such.

This makes sense to me I will move it all to an object to reduce the amount of imports



> While it is unlikely that Wear will ever get another entrypoint for notifications, why put all the code in `FirebaseMessagingService`?

I just called it `FirebaseMessagingService` as most (not all) of the code applies to Wear OS as we need to handle new tokens etc... Also not all is supported here so I kept it all in the same file. We can rename it if we want to make it more generic?

> Splitting it up to keep the Firebase code separate from the notification handling code results in a smaller file, making it easier to read and compare with the main app.

Not sure I follow here? The phone and Wear app will need to differentiate between what they can and can't support like video handling and various device commands that are phone specific. I thought the split I did would be sufficient as phone and watch will have separate logic they need to maintain. We can share some things still.

jpelgrom:
> I just called it `FirebaseMessagingService` as most (not all) of the code applies to Wear OS as we need to handle new tokens etc... Also not all is supported here so I kept it all in the same file. We can rename it if we want to make it more generic?
>
> (...)
>
> Not sure I follow here? The phone and Wear app will need to differentiate between what they can and can't support like video handling and various device commands that are phone specific.

Nothing wrong with the name. Obviously the Firebase-specific code needs to live somewhere. Trying to explain it differently: why not introduce a `MessagingManager`-like class for the `wear` module as well that deals with the functional stuff (notifications)? `FirebaseMessagingService` can do the housekeeping for receiving push data and refreshing tokens. By doing that it better mimics what is done in the main app, which makes it easy to compare code and introduce features in the same way.

> I thought the split I did would be sufficient as phone and watch will have separate logic they need to maintain. We can share some things still.

Yes the current approach for that is fine :)

dshokouhi:
> Trying to explain it differently: why not introduce a `MessagingManager`-like class for the `wear` module as well that deals with the functional stuff (notifications)? `FirebaseMessagingService` can do the housekeeping for receiving push data and refreshing tokens. By doing that it better mimics what is done in the main app, which makes it easy to compare code and introduce features in the same way.

Thanks thats more clear now!

jpelgrom:
Minor issue, building initially failed for me with:

```
Execution failed for task ':wear:processDebugGoogleServices'.
> File google-services.json is missing. The Google Services Plugin cannot function without it. 
   Searched Location: 
  \home-assistant-android\wear\src\debug\google-services.json
  \home-assistant-android\wear\src\google-services.json
  \home-assistant-android\wear\google-services.json
```

In my opinion the README should be updated to mention this, it currently only refers to the main `app` module.

The app is also [always sending `"push_websocket_channel": true` in the registration JSON](https://github.com/home-assistant/android/blob/5cd74c0179f13b617de517273ec860215a8259ba/common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt#L762), that should probably be changed for the Wear app.

jpelgrom:
Looks like push notifications are working 🥳 Tested an upgrade as well and it is handled automatically during re-registration.

A screenshot of a notification from the Home Assistant Wear app, titled "Hello!" and with the message "Via the cloud ☁️"

dshokouhi:
Awesome next commit I will update the readme and also for device registration will set it to `false` from wear OS as the phone app is always going to be `true` there. Have code changes ready just want to test before committing.

jpelgrom:
The latest Firebase BoM is 31.1.1, I don't see a reason to use an older version.

Other code looks good to start building notification support on to me!

dshokouhi:
> The latest Firebase BoM is 31.1.1, I don't see a reason to use an older version.

ah I copied and pasted the main app version, let me bump both of them :) 

Surprised android studio doesn't highlight it like the others 🤔

jpelgrom (approved):
I think this is a good start, let's see how it scales once more notification features are supported

JBassett (approved):

Issue #2684, opened 2022-07-16
Unable to connect to Home Assistant error when playing video opened from notification


Rogue136198:
**Home Assistant Android version:**
beta-2484-0e0994de-full

**Android version:**
Android 12
Android 13 (Beta)

**Phone model:**
Pixel 4a (5G)
Pixel 5 (in a VM)

**Home Assistant version:**
2022.7.5

**Last working Home Assistant release (if known):**
N/A

**Description of problem:**
When playing a video clip opened from a notification I receive the error "Unable to connect to Home Assistant" The video continues to play behind the error message

The issue is present on my primary Pixel 4a (5G) running Android 13 Beta 4 and the beta release of HA. This issue was present when I was running Android 12

The issue is also present on a Pixel 5 VM running Android 12 stable

To further clarify, the issue happens when accessing a video from /api. The URL format is as follows
`/api/frigate/notifications/{{id}}/{{camera}}/clip.mp4`

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

Debug Log from Android Studio VM
[android_studio_log.txt](https://github.com/home-assistant/android/files/9126161/android_studio_log.txt)

Debug Log from In App
[homeassistant_companion_log_6-16-2022_12-9-36.txt](https://github.com/home-assistant/android/files/9126160/homeassistant_companion_log_6-16-2022_12-9-36.txt)

**Screenshot of problem:**
![image](https://user-images.githubusercontent.com/14829730/179362008-01f4c9eb-6eee-4321-b368-cfff36e92426.png)

**Additional information:**
- I have reviewed Issue #2325 but I was still unable to solve the issue by restarting Mosquitto. To further clarify, I've rebooted everything.
- I am running Home Assistant behind nginx using the SWAG docker container. I have confirmed this issue is present when connecting directly to the HA docker IP.
- I am using the new mTLS option but the issue was present before implementing this. Also the issue is present when accessing the docker IP directly thus bypassing the mTLS auth.
- I have not ruled out that this could be an issue with the Frigate Integration

dshokouhi:
Duplicate of https://github.com/home-assistant/android/issues/1672

Rogue136198:
> Duplicate of #1672

Can you please explain how this is duplicated? This issue is regarding opening /api in app not /media

[...]

PR #3232, opened 2023-01-16
Skip showing connection prompt when loaded URL is local or api path

dshokouhi:
## Summary

Fixes: #2684 by skipping the connection prompt when the loaded URL is either `local` or `api` as these pages will not have external bus communication.


jpelgrom:
Is it safe to assume `local` or `api` will always be the first segment of the URL, to limit false positives?

dshokouhi:
> Is it safe to assume `local` or `api` will always be the first segment of the URL, to limit false positives?

I think that's a good idea, I considered it but wasn't sure. Now that you mention it, probably better to be more precise here. Only thing I can't think of at this point is if any other path should be skipped.

jpelgrom (approved):
DId a quick test with a local file and works as intended!

> I can't think of at this point is if any other path should be skipped.

Right now I also can't think of any, and if there is it's easy to add later.

JBassett (approved):

PR #3233, opened 2023-01-17
Conversation tile design changes

dshokouhi:
## Summary

A few more improvements to the conversation tile.

*  Added speech bubbles to align the text more like a conversation
*  If haptic feedback is enabled we will vibrate upon displaying the response
*  Close the activity when the screen turns off and we have received a response
*  Adds a preview of the speech bubbles

## Screenshots

image

image

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/conversation/ConversationActivity.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/conversation/ConversationActivity.kt
@@ -62,9 +62,8 @@ class ConversationActivity : ComponentActivity() {
     override fun onPause() {
         super.onPause()
         val pm = applicationContext.getSystemService<PowerManager>()
-        if (pm != null) {
-            if (!pm.isInteractive)
-                finish()
+        if (pm?.isInteractive == false && conversationViewModel.conversationResult.isNotEmpty()) {
+            finish()
         }
     }
 }
--- a/wear/src/main/java/io/homeassistant/companion/android/conversation/ConversationActivity.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/conversation/ConversationActivity.kt
@@ -3,11 +3,13 @@ package io.homeassistant.companion.android.conversation
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
+import android.os.PowerManager
 import android.speech.RecognizerIntent
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.activity.viewModels
+import androidx.core.content.getSystemService
 import androidx.lifecycle.lifecycleScope
 import dagger.hilt.android.AndroidEntryPoint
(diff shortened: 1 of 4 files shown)


wear/src/main/java/io/homeassistant/companion/android/conversation/ConversationActivity.kt
```diff
@@ -56,4 +58,13 @@ class ConversationActivity : ComponentActivity() {
...
         }
     }
+
+    override fun onPause() {
+        super.onPause()
+        val pm = applicationContext.getSystemService<PowerManager>()
+        if (pm != null) {
+            if (!pm.isInteractive)
+                finish()
+        }
```
jpelgrom, line 68:
```suggestion
        if (pm?.isInteractive == false) {
            finish()
        }
```

wear/src/main/java/io/homeassistant/companion/android/conversation/views/ConversationView.kt
```diff
@@ -30,28 +43,81 @@ fun ConversationResultView(
...
                 }
                 if (conversationViewModel.conversationResult.isNotEmpty())
                     item {
-                        Text(
-                            text = conversationViewModel.conversationResult,
-                            modifier = Modifier.padding(top = 8.dp, start = 32.dp)
-                        )
+                        Row {
+                            if (conversationViewModel.isHapticEnabled.value)
+                                LocalHapticFeedback.current.performHapticFeedback(HapticFeedbackType.LongPress)
```
jpelgrom, line 69: Haptic feedback will be performed on every recomposition while there is a result. This should use a side-effect instead so it is executed only once.

dshokouhi: Looks like a `SideEffect` will actually run on every recomposition as well so I guess a `LaunchedEffect` is more proper to handle that? I didnt think this page would recompose again after displaying the text since we don't do anything else and the activity closes when the screen turns off

jpelgrom:
Something I just thought of: if the watch is still waiting for the server to get a conversation result the activity shouldn't be finished when the screen turns off.

dshokouhi:
> Something I just thought of: if the watch is still waiting for the server to get a conversation result the activity shouldn't be finished when the screen turns off.

ah good catch, I will add a check to make sure we have already received a response!

balloob:
Can we add some padding on the opposite side of the text bubble so they don't take 100% width? That will make it more clear to which side they belong when the square corner is cut off.

CleanShot 2023-01-18 at 10 11 53

dshokouhi:
> Can we add some padding on the opposite side of the text bubble so they don't take 100% width? That will make it more clear to which side they belong when the square corner is cut off.
> 
> CleanShot 2023-01-18 at 10 11 53

This comment made me realize the spacer was not working as intended. This is how this screen looks now with the latest changes

image

jpelgrom (approved):

balloob (approved):

JBassett (approved):

PR #3236, opened 2023-01-17
Clear beacons when the monitor stops

dshokouhi:
## Summary

Fixes: #3010 by clearing the stored list of `beacons` and `lastSeenBeacons` every time the monitor stops. Also skipping the first observer values as they are cached from live data.


dshokouhi:
Closing this PR as the approach is not correct https://github.com/home-assistant/android/issues/3010#issuecomment-1387437327

dshokouhi:
Actually going to reopen this as the beacon variables should still be cleared out in addition to skipping the first observer values as mentioned in the linked issue from the above comment

dshokouhi:
will leave as a draft now in case theres another approach we can use but almost feels like the proper fix here will be from the library

PR #3239, opened 2023-01-18
Wi-Fi sensor improvements

anyuta1166:
## Summary

This request addresses 2 issues at once, because they are interconnected:

1) Using wifiManager.scanResults to get signal strength is very unreliable. I often see "-1" on some devices even if Wi-Fi is connected. Use conInfo.rssi for a reliable result.

Also, wifiManager.scanResults is not available if Location is disabled on a device, but conInfo.rssi is available in such case.

2) There is an issue on devices running Android 10+ (tested on various Android 10 and Android 11 devices). If Wi-Fi is connected, but Location is disabled on a device,  getNetworkId() returns -1 and Home Assistant app thinks that we are not connected. But some information - IP address, link speed, frequency, signal strength (via getRssi()) is actually still available. If NetworkId == -1, we can additionally check LinkSpeed to determine that we are connected and provide IP address, link speed, frequency and signal strength.

Please note that if Location is disabled on a device, wifiManager.scanResults is not available, but conInfo.rssi is available, that's why I've merged this into one commit.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
@@ -182,20 +182,27 @@ class NetworkSensorManager : SensorManager {
 
         var conInfo: WifiInfo? = null
         var ssid = "Unknown"
+        var connected = false
 
         if (checkPermission(context, wifiConnection.id)) {
             val wifiManager =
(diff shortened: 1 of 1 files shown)


dshokouhi (approved):
Looks good to me, didn't realize we were hitting an odd issue like that. Thanks for the PR.

JBassett (approved):

Issue #296, opened 2020-01-23
Support for control 2 Home Assistance


ghost:
Hi
Is any chance to add support for two instance Home Assistant on other address ?

dshokouhi:
Even better if it can handle one of those instances being down usually like a dev instance

jes1417:
I was hoping for this as well, in my use case I have an instance at home and at work and I wanted to control both of them. I ended up getting this working by setting up a work profile on my phone which shows me to have separate apps, works perfectly!

balloob:
We are still working on building out basic functionality of the app, so this is not on the roadmap. Once we feel like we're in a good place we would gladly accept a pull request!

tescher:
I'd like to give a +1 for this enhancement also.

mishoco:
This is a very important feature... any update for this?

T-bond:
If anybody like me on a Samsung phone without multiple profiles:
You can use GPlay and F-Droid to install 2 instances (F-Droid is only the minimal favour)
(Or maybe use Secure Folder)

I hope this can get implemented in the near future, as I can't think about anything missing, other than this.
So thank you everybody who worked on this project!

@balloob may I ask you, is the project reached into a state to allow pull request for this? Or is there an estimate for it?
Thank you, very much!

teras:
Hello.
I think the problem is not solved like this.
With two instances we have two notifications, double resources and the auto-update is a problem especially with f-droid. And what if I need three or more instances?

I strongly believe that this issue should be reopened. The similar iOS issue is still pending :grimacing:

balazs111:
Similar  issue here. Already has 2 system and the third is coming.

robbiet480:
Hey all, I ran a poll on our Twitter account this week trying to determine how many people actually have multiple instances. Based on the results of the poll we (@JBassett, @dshokouhi and I) still feel that there’s not enough people who would benefit from the immense effort it would take to rework the app core to support multiple instances vs the effort is takes for developers to keep adding features and fixing things.

As I’ve said before, if a developer came along that wanted to add this functionality (in concert with Justin and others) and was willing to maintain it, we’d be open to that. But for now, there’s far more people who benefit from incremental changes than dropping everything to support a small but vocal minority.

Sorry y’all. Will revisit this again in the future.

tescher:
Twitter poll? Who here monitors the HA Twitter feed?

robbiet480:
[...]

PR #3241, opened 2023-01-18
Split app-level and server-level settings

jpelgrom:
## Summary

Preparing for multiserver support. The app currently mixes settings related to general app features and for the server, both in the UI and in code. Once we start doing multiple servers this will create issues. This PR:
- introduces a new screen in the app settings for server-level settings*
- moves all app-level settings storage out of the `IntegrationRepository`, which will be for server-specific functions, into the existing `PrefsRepository` (which seems like an appropriate place) or the new `WearPrefsRepository`
- cleans up a few things in settings for fragment transactions and permission handling to modern code

\* Notes for the first item in the list:
 - Eventually I expect sensor settings will have server-level settings, but for now I'm keeping it on app-level to keep things simple.
 - The new screen only changes the location in the UI, most of it is based on the existing settings UI and code changes for storing the data on a server-level will be in a future even larger PR.

## Screenshots

At the top of the settings, there is a new 'Servers & Devices' section which will list the server + Wear OS settings if applicable. In the future more servers can be shown here as well as the option to add a new server.
|Light|Dark|
|-----|-----|
|Android app settings with a section 'Servers & Devices', showing a list item 'Pixeldev' with an IP address and 'Wear OS Settings', light mode|Android app settings with a section 'Servers & Devices', showing a list item 'Pixeldev' with an IP address and 'Wear OS Settings', dark mode|

The server settings include connection settings, device registration, app lock and persistent connection.
|Light|Dark|
|-----|-----|
|Android app settings with the title 'Server Settings', showing a list of settings as described that were previously in the main settings list, light mode|Android app settings with the title 'Server Settings', showing a list of settings as described that were previously in the main settings list, dark mode|

I believe this isn't necessary as none of these settings are described in detail, and the iOS app does something similar in settings which isn't causing any issues.

## Any other notes

Related to #296


dshokouhi (approved):
I really like this split! Tested out the debug APK and it works and makes sense to have things split as proposed. Code wise looks ok just a lot of restructuring :)

JBassett (approved):

PR #3246, opened 2023-01-20
Fix conversation tile vibrate after settings split

jpelgrom:
## Summary

Fix build after merging #3233 and #3241 within several minutes, causing a conflict as a new feature is using preferences that were moved at the same time.

## Screenshots

n/a

n/a


dshokouhi (approved):
Thanks for the quick fix!

JBassett (approved):

PR #3247, opened 2023-01-20
Update change log week of 1/20

dshokouhi:


jpelgrom:
> Initial support for notifications sent directly to watch service call.

To me this suggests it was already possible to send notifications to a watch, which can be solved by using "sending". Is "service call" necessary?

dshokouhi:
> To me this suggests it was already possible to send notifications to a watch, which can be solved by using "sending".

Technically by default it is because notifications are relayed, I wanted to make it obvious that the phone will not receive like they are used to and that in order to use the feature you have to use the corresponding service call

jpelgrom:
I see now, maybe using "only" instead of "directly" is easier to understand? Not sure but I definitely didn't get it the first time even though I'm familiar with the PR :)

jpelgrom (approved):

PR #3253, opened 2023-01-21
Add ble transmitter and beacon monitor notification commands to Wear OS

dshokouhi:
## Summary

Begins the migration process of notification commands to work on Wear OS. I do not expect many of the commands to carry over to Wear OS. A lot of the commands require special permissions  that simply don't exist in Wear OS nor is there an activity to let the user grant the permission. Commands should be accepted on a case by case basis if they will work.

In this PR I have moved over `command_ble_transmitter` and `command_beacon_monitor` I have also moved over the command data checks so they can reused by phone and watch.

Documentation: home-assistant/companion.home-assistant#889


dshokouhi:
out of all the other notification commands, the only one I can see we can add is `command_update_sensors` all others do not seem possible given the platform limitations or feature availability. No Draw Over Other Apps permission, No DND Permission, No Notification permission, No Write Settings permission and no persistent connection or location updates.

jpelgrom:
Code looks OK.

In the `app` module the `sendNotification` function is always called in a job that is launched from the `mainScope`, but you're not doing that for the `wear` module. Any reason for not doing this?

jpelgrom (approved):
Commands arrive and seem to change settings on the watch. Also appears to be working as it did before the changes when using the main app.

JBassett (approved):

PR #3254, opened 2023-01-23
Make conversation activity launchable

dshokouhi:
## Summary

Makes the new conversation activity launchable so it can show up in wearable app launcher as well as Recents. This will also allow for button remapping apps to let users map directly to the activity.

Adds a new icon set with debug images so they can continue to be distinguished.

Added a registration check so we can let the user know they are not logged in.

Updates tile selection screen to reflect "Assist" instead of "Conversation"

## Screenshots

image

image


jpelgrom:
Do other apps with multiple activities prefix them with their app name? "HA: Assist" looks out of place.

The launcher foreground/background images should use vectors, like the main app icon. The Wear app doesn't support SDK <26 so I believe we don't need pngs?

dshokouhi:
> Do other apps with multiple activities prefix them with their app name? "HA: Assist" looks out of place.

Fitbit has 2, for the name I actually asked @balloob in a side chat since "Home Assistant Assist" did not have a nice ring to it and the current icon with just "Assist" may not be known its from HA.



> The launcher foreground/background images should use vectors, like the main app icon. The Wear app doesn't support SDK <26 so I believe we don't need pngs?

possibly, admittedly I used [IconKitchen](https://icon.kitchen) to generate all the images and copy/pasted :)

jpelgrom:
> for the name I actually asked balloob in a side chat since "Home Assistant Assist" did not have a nice ring to it and the current icon with just "Assist" may not be known its from HA.

I agree it's probably the best of the three options you mentioned. 

> possibly, admittedly I used [IconKitchen](https://icon.kitchen) to generate all the images and copy/pasted :)

Surely there are other tools or a template for something like Figma that will allow you to get vector layers :)

dshokouhi:
> Surely there are other tools or a template for something like Figma that will allow you to get vector layers :)

it was the first one i found in my google search and it looked nice since it spits out the images for you in all its necessary folders, do you have another tool thats similar that also provides the correct files needed? I have never used Figma and it did not seem to easily supply those launcher icons from first look 😅

jpelgrom:
Have you tried the built-in tool in Android Studio? When unchecking all the legacy things in options it generates three files, which look pretty close to your result. Afterwards you only need to do some light editing (debug/main flavors, use color directly) to match the main app.

image
image
image

dshokouhi:
Thanks thats a lot easier! Not sure why I didn't consider the image asset tool lol

jpelgrom:
Even more files can be removed ☺️

 - `wear/src/debug/res/drawable/ic_assist_launcher_foreground.xml` is the same as the main (non-debug) version, 
 - `wear/src/debug/res/values/ic_assist_launcher_background.xml` and `wear/src/main/res/values/ic_assist_launcher_background.xml` be removed, use the colors [directly in the launcher icon file](https://github.com/home-assistant/android/blob/96cf4f82a3d28ce9702fe637a844740385e51162/app/src/main/res/mipmap-anydpi-v26/ic_launcher.xml#L3)
 - `wear/src/main/res/drawable/ic_assist_launcher.png` is unused

wear/src/main/AndroidManifest.xml
```diff
@@ -87,7 +87,16 @@
 
         <activity
             android:name=".conversation.ConversationActivity"
-            android:exported="true" />
+            android:icon="@mipmap/ic_assist_launcher"
+            android:roundIcon="@mipmap/ic_assist_launcher"
```
jpelgrom, line 91: As it is the same as the regular icon, [it can be removed](https://developer.android.com/develop/ui/views/launch/icon_design_adaptive#:~:text=define%20an%20icon%20drawable%20resource%20using%20the%20android%3AroundIcon%20attribute%E2%80%94but%20only%20if%20you%20require%20a%20different%20icon%20asset%20for%20circular%20masks)

dshokouhi: thank you!

jpelgrom (approved):

JBassett (approved):

Issue #3243, opened 2023-01-19
Wear OS Favorite Entities selections keep disappearing


Qoheleth:
**Home Assistant Android version:** 2022.12.0-full

**Android version:** 13

**Phone model:** Galaxy Fold 4 (SM-F936U)

**Home Assistant version:**

**Last working Home Assistant release (if known):** This has been an issue since I've tried using it (~2022.6)

**Description of problem:** When I set these on my phone, they appear as they’re supposed to on my watch (Samsung Galaxy Watch 5). Some time later, they start to disappear from my watch, often several at a time, but not all at once, inevitably they all disappear, sometimes even while I'm viewing them on the watch.
If I go back to the Companion app on my phone and re-select them, they’ll show up again until they start disappearing again some time later. This process my take a day or two to start, but always finishes within a few minutes.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```

**Screenshot of problem:** (Left as it should be. Right after disappearence)
![ha-watch](https://user-images.githubusercontent.com/2237808/213516767-1d32d23e-1fb4-4cd4-ae07-813180301bfd.png)

**Additional information:** none

dshokouhi:
Interesting the only time the favorites _should_ be removed is when the server does not respond with them in the list of entities we retrieved

https://github.com/home-assistant/android/blob/5cd74c0179f13b617de517273ec860215a8259ba/wear/src/main/java/io/homeassistant/companion/android/home/views/MainView.kt#L126

Strange so some responses don't have the entities but some do? Wonder what would cause that 🤔 either way not a good experience if actual entities are getting removed.

dshokouhi:
Confirmed with HA core developers that websocket responses are not paginated so the response should always have all entities. This is going to be tough to investigate where the potential loss is. One thing I wonder, when this issue happens do you see the entities in their respective area/domain after the removal? Part of me wonders if instead of trying to delete the entity we keep it as a `chip` without state data as its possible the chip will update later on with state data. We do offer a delete all option so users can clean up entities on their own without us trying to delete it when we believe the entity no longer exists

Qoheleth:
Just a follow up with more data:
When my favorites disappear. I also can't find some entities under Areas. (For example, I can only see a scene in my Living Room, and only a light in my Dining Room.)  I can, however, find them all under More Entities which groups them by type (buttons, covers, etc.)

[...]

PR #3269, opened 2023-01-26
Don't delete favorites instead show cached entity

dshokouhi:
## Summary

Fixes: #3243 by defaulting to a cached favorite entity chip instead of deleting a favorite. While I am not sure what the actual cause of the loss of favorites is, we can at least not delete the favorite and for the time being keep the removal in the hands of the user. I think it is a better user experience to keep the favorites so users do not need to continue to add them back. Asides from the one user report I have seen mentions on Discord and Reddit so this issue seems to impact quite a bit of users that use this feature.


dshokouhi:
Closing for now in favor of #3276 as it has a better approach I feel. No need to create a cached chip, the entities come back immediately.

PR #3273, opened 2023-01-26
Allow navigation to sensors in android auto if they contain proper attributes

dshokouhi:
## Summary

I originally thought that it was any entity but now I realize we limited the domains. Adding in `sensor` domain in case a user has a sensor or created a template sensor so it can also be navigated to.

Documentation: home-assistant/companion.home-assistant#896


jpelgrom (approved):
The docs should also be updated, as they incorrectly state only `zone` and `device_tracker` entities are supported.

dshokouhi:
> The docs should also be updated, as they incorrectly state only `zone` and `device_tracker` entities are supported.

ah good catch, docs PR attached

JBassett (approved):

PR #3276, opened 2023-01-27
Add new option to only show favorites on home screen

dshokouhi:
## Summary

Based on some feedback that I have seen from users only wanting to see their selected entities. I am adding a new setting to only show favorites on the home screen in the Wear OS app. The hope is that needing to load less data will make things feel faster for people who only want to control a certain set of entities from the app.

This also contains a different approach to #3269 as I have seen that favorites are not really removed, they actually get added back almost immediately. So here instead of creating a cached favorite, just let the app recreate the chip.

## Screenshots

image

image

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepositoryImpl.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepositoryImpl.kt
@@ -21,6 +21,7 @@ class WearPrefsRepositoryImpl @Inject constructor(
         private const val PREF_TILE_TEMPLATE_REFRESH_INTERVAL = "tile_template_refresh_interval"
         private const val PREF_WEAR_HAPTIC_FEEDBACK = "wear_haptic_feedback"
         private const val PREF_WEAR_TOAST_CONFIRMATION = "wear_toast_confirmation"
+        private const val PREF_WEAR_FAVORITES_ONLY = "wear_favorites_only"
     }
 
     init {
@@ -44,6 +45,9 @@ class WearPrefsRepositoryImpl @Inject constructor(
                 }
                 integrationStorage.getBooleanOrNull(PREF_WEAR_TOAST_CONFIRMATION)?.let {
                     localStorage.putBoolean(PREF_WEAR_TOAST_CONFIRMATION, it)
+                }
+                integrationStorage.getBooleanOrNull(PREF_WEAR_FAVORITES_ONLY)?.let {
+                    localStorage.putBoolean(PREF_WEAR_FAVORITES_ONLY, it)
                 }
 
                 localStorage.putInt(MIGRATION_PREF, MIGRATION_VERSION)
@@ -100,5 +104,13 @@ class WearPrefsRepositoryImpl @Inject constructor(
 
     override suspend fun setWearToastConfirmation(enabled: Boolean) {
         localStorage.putBoolean(PREF_WEAR_TOAST_CONFIRMATION, enabled)
+    }
+
+    override suspend fun getWearFavoritesOnly(): Boolean {
+        return localStorage.getBoolean(PREF_WEAR_FAVORITES_ONLY)
+    }
+
+    override suspend fun setWearFavoritesOnly(enabled: Boolean) {
+        localStorage.putBoolean(PREF_WEAR_FAVORITES_ONLY, enabled)
     }
 }
--- a/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
@@ -108,6 +108,8 @@ class MainViewModel @Inject constructor(
         private set
     var templateTileRefreshInterval = mutableStateOf(0)
         private set
+    var isFavoritesOnly = mutableStateOf(false)
+        private set
 
     fun supportedDomains(): List<String> = HomePresenterImpl.supportedDomains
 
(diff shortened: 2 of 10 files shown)


common/src/main/java/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepositoryImpl.kt
```diff
@@ -45,6 +46,9 @@ class WearPrefsRepositoryImpl @Inject constructor(
                 integrationStorage.getBooleanOrNull(PREF_WEAR_TOAST_CONFIRMATION)?.let {
                     localStorage.putBoolean(PREF_WEAR_TOAST_CONFIRMATION, it)
                 }
+                integrationStorage.getBooleanOrNull(PREF_WEAR_FAVORITES_ONLY)?.let {
+                    localStorage.putBoolean(PREF_WEAR_FAVORITES_ONLY, it)
+                }
```
jpelgrom, line 51: As it is a new preference you don't need to add a migration for it

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
```diff
@@ -108,6 +108,8 @@ class MainViewModel @Inject constructor(
         private set
     var templateTileRefreshInterval = mutableStateOf(0)
         private set
+    var isFavoritesOnly = mutableStateOf(false)
```
jpelgrom, line 111: All settings are using `=`, but wouldn't it be nice to use `by` instead so you don't need to use `.value` everywhere?

wear/src/main/java/io/homeassistant/companion/android/home/views/MainView.kt
```diff
@@ -54,10 +54,9 @@ fun MainView(
     onEntityLongClicked: (String) -> Unit,
     onRetryLoadEntitiesClicked: () -> Unit,
     onSettingsClicked: () -> Unit,
-    onTestClicked: (entityLists: Map<String, List<Entity<*>>>, listOrder: List<String>, filter: (Entity<*>) -> (Boolean)) -> Unit,
+    onTestClicked: (entityLists: Map<String, List<Entity<*>>>, listOrder: List<String>, filter: (Entity<*>) -> Boolean) -> Unit,
```
jpelgrom, line 57: If you're editing this please give it a proper name ☺️

wear/src/main/java/io/homeassistant/companion/android/home/views/SettingsView.kt
```diff
@@ -105,6 +107,29 @@ fun SettingsView(
...
                 }
+                item {
+                    ToggleChip(
+                        modifier = Modifier.fillMaxWidth(),
+                        checked = isFavoritesOnly,
+                        onCheckedChange = { setFavoritesOnly(it) },
+                        label = { Text(stringResource(commonR.string.only_favorites)) },
+                        toggleControl = {
+                            Icon(
+                                imageVector = ToggleChipDefaults.checkboxIcon(isFavoritesOnly),
```
jpelgrom, line 118: I'm not sure a checkbox is the appropriate control here, as those are generally used for [one or more items from a set](https://m3.material.io/components/checkbox/overview). For feedback options it's a set but here it's [on or off, standalone](https://m3.material.io/components/switch/overview).

jpelgrom (commented):
Looking at the screenshot, there should be padding inbetween favorites and the settings chip.

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
dshokouhi: Sure, let me change this new one I added here and then lets update them all in a follow up PR since that will touch a few different places :)

dshokouhi:
> Looking at the screenshot, there should be padding inbetween favorites and the settings chip.

ok added some padding there when all other items are removed :)

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
```diff
@@ -224,7 +233,7 @@ class MainViewModel @Inject constructor(
     }
 
     suspend fun entityRegistryUpdates() {
-        if (!homePresenter.isConnected())
+        if (!homePresenter.isConnected() || isFavoritesOnly)
```
jpelgrom, line 236: Why not do updates even though you do load the entity registry initially? (l191)

wear/src/main/java/io/homeassistant/companion/android/home/views/MainView.kt
```diff
@@ -123,195 +122,218 @@ fun MainView(
...
-                                    colors = ChipDefaults.secondaryChipColors()
-                                )
-                            }
-                        }
                     }
                 }
 
+                if (mainViewModel.isFavoritesOnly)
+                    item {
+                        Spacer(Modifier.padding(4.dp))
```
jpelgrom, line 334: The main list is currently using a [32dp spacer before the all entities/settings chips](https://github.com/home-assistant/android/blob/bd822afabad97aae8dfeb1d4cec5229216f63bb0/wear/src/main/java/io/homeassistant/companion/android/home/views/MainView.kt#L284-L286), let's use the same size everywhere.

jpelgrom (commented):
If all that is shown is favorites, do we still need the 'Favorites' title?

The experience when no favorites are set is a bit weird. I think we should either:
- disable the option when there are no favorites (including when all favorites are removed!)
- add an empty state that directly links to the set favorite entities screen

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
dshokouhi: I was only interested in the initial list of entities, but reading this comment. Yea we should make sure the list is up to date if modified when the app is open :)

dshokouhi:
> * disable the option when there are no favorites (including when all favorites are removed!)

totally forgot to add this back when I was creating the setting chip lol, thanks for the reminder 😄

dshokouhi:
> If all that is shown is favorites, do we still need the 'Favorites' title?

I think so because users can still collapse the header to quickly get to Settings if they desire. We may have users adding more favorites as a result of this new feature 😁

JBassett (approved):

PR #3297, opened 2023-02-01
Update issue template to be more specific

dshokouhi:
## Summary

Some minor updates to request phone and watch info. Also requesting YAML when troubleshooting notifications. Updates traceback section so users will always add logs instead of us always requesting them.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -15,11 +15,12 @@ assignees: ''
   DO NOT DELETE ANY TEXT from this template! All requested information is important.
 -->
 
-**Home Assistant Android version (Phone and/or Watch):**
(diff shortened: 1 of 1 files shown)


jpelgrom:
What about tablets 😉 ?

Would it maybe be a good idea to add comments `<!-- -->` instead of adding more info between `( )`, or switching to the bug report flow like core is using with rich text/links?

dshokouhi:
> What about tablets 😉 ?

haha well it did say phone before 😛 just wanted to clarify both is preferable in case we need it

> Would it maybe be a good idea to add comments  instead of adding more info between ( ) to keep the visible issue less cluttered, 

good call, will update

> or switching to the bug report flow like core is using with rich text/links?

last I asked it was not enabled for our repo yet

jpelgrom:
> If you are reporting an issue for (...) Android Auto please list both devices

The phone and the head unit? It all runs on the phone.

jpelgrom (approved):

JBassett (approved):

PR #3315, opened 2023-02-07
Update sensor enabled by default logic to be more granular

dshokouhi:
## Summary

As we have had more battery sensors get added the amount of sensors enabled by default has grown. Some of the battery sensors were created based on requests and probably not used by majority of users.  Updating the logic to move from disabled at the manager level to be enabled at the individual sensor level. Default is still `false`

So now the new default sensors will be:

- `battery_level`
- `battery_state`
- `charger_type`

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#907

## Any other notes

Not sure if this should be marked as a breaking change since this only impacts new installs.

If you are planning on testing this PR make sure to use a device name never used in your HA instance otherwise you will see the old sensors enabled by default.

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/ActivitySensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/ActivitySensorManager.kt
@@ -117,7 +117,7 @@ class ActivitySensorManager : BroadcastReceiver(), SensorManager {
 
     private fun handleSleepUpdate(intent: Intent, context: Context) {
         Log.d(TAG, "Received sleep update")
-        if (SleepClassifyEvent.hasEvents(intent) && isEnabled(context, sleepConfidence.id)) {
+        if (SleepClassifyEvent.hasEvents(intent) && isEnabled(context, sleepConfidence)) {
             Log.d(TAG, "Sleep classify event detected")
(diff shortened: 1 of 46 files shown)


jpelgrom:
> Not sure if this should be marked as a breaking change since this only impacts new installs.

Existing users won't notice a difference so it is not a breaking change :)

jpelgrom:
If we're keeping `charger_type`, I'd also keep `is_charging`, or if the aim is to add as few sensors as possible I expect more people will simply want to know "is my device charging?" than "how is my device charging?".

(Why isn't the charger type an attribute?)

dshokouhi:
> If we're keeping `charger_type`, I'd also keep `is_charging`, or if the aim is to add as few sensors as possible I expect more people will simply want to know "is my device charging?" than "how is my device charging?".

The goal is to have less sensors that are not redundant. In this case `is_charging` is actually derived from `battery_state` so it makes sense to include more details than the opposite to begin with. Personally I like the `full` state because then you know your battery is done charging. 100% != a full battery :)

jpelgrom (approved):

JBassett (approved):

JBassett:
Once merge conflict is fixed it will be merged.

PR #3316, opened 2023-02-07
Default to carrier name when SIM sensor display name is null

dshokouhi:
## Summary

Had a user point out that their e-SIM was showing `unavailable` despite using a SIM. Turns out there is data but the `displayName` is `null` so now we will default to the `carrierName` instead of showing `unavailable`

We may need to mark this as a breaking change due to this behavior change.

## Any other notes

Ref: https://community.home-assistant.io/t/e-sim-support/483806/17?u=dshokouhi


jpelgrom (approved):
When trying to edit the SIM name in system settings it actually sets it to the carrier name for me when I leave it empty, so this change seems logical.

JBassett (approved):

PR #3321, opened 2023-02-08
Add activity sensor manager to Wear OS

dshokouhi:
## Summary

This PR is attempting to bring over the Activity sensor manager which uses Google Services API to Wear OS.  I have not found a good way to not have so much code duplication but I was able to centralize all the Google specific API code and its corresponding functions so they can be reused to reduce the amount of duplication. I ran into issues with the minimal build with my current approach of attempting to migrate the entire class so I opted to migrate the Google specific parts so the full and minimal builds can continue to behave as they do.

Testing wise all 3 sensors continue to work on my Pixel 7 pro

On my Pixel Watch so far I have only seen the Detected Activity sensor updating at a random interval, not as frequent as the phone. Both sleep sensors do not update and I wonder if that API was added to Wear OS. There are no errors during registration and the API reports it was successful. As of the past 5 hours I have not seen any updates for them while the activity sensor updates at its own odd interval. These APIs are subject to doze so that probably explains why the interval is not as expected when compared to the phone.

Wanted to submit this as a draft to get feedback on the approach.  Also would be good if some Wear OS users tested the new sensors to confirm my suspicions around the API. I have a feeling the Sleep API may not work on Wear OS 3 as those devices allow for sleep detecting using Health Services which we currently offer. I would like a Wear OS 2 to user to also test and see if the 2 sleep sensors update at all in a 24 hour period. The more testing the devices the merrier.

## Screenshots

image


dshokouhi:
About 24 hours later both `sleep_confidence` and `sleep_segment` have not updated at all. `detected_activity` is updating

image

based on user feedback we may need to hide or remove those 2 sensors if they are not functional on Wear OS for any user

image

image

dshokouhi:
Update: based on user feedback so far Sleep API sensors are non-functional for Samsung Galaxy Watches as well, so as of now not supported on Wear OS 3. Currently waiting on a Wear OS 2 user to also test and confirm before updating this PR to remove the 2 non-functional sleep sensors.

dshokouhi:
Closing this PR as I am not a fan of my own approach and I think doing it the correct way will probably require more time than I have.

Issue #2220, opened 2022-01-31
Companion app with Nabu Casa URL


ACJUK:
**Home Assistant Android version:**
2022.1.1-full

**Android version:**
12

**Phone model:**
Google Pixel 3a

**Home Assistant version:**
core-2021.12.10

**Last working Home Assistant release (if known):**
Not known

**Description of problem:**
When logging into the app and using a Nabu Casa URL, the process goes to login and then the dashboard without asking for the phone model and asking for location services to be turned on.  This means that the mobile app integration doesn't get installed and none of the tacking entities get created

In turn, this also means that the Companion App settings option doesn't appear under Configuration.

I've tried this on 2 different servers.  One outside my network and one inside.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

```

```

**Screenshot of problem:**

**Additional information:**

dshokouhi:
This isn't necessarily related to Nabu Casa yet. We still need the logs as mentioned in the issue template. I can tell you that I have setup and used Nabu Casa on other devices with success as well so the issue still might be something else.

ACJUK:
[logcat_01-31-2022_21-55-34.txt](https://github.com/home-assistant/android/files/7974515/logcat_01-31-2022_21-55-34.txt)

I ran the command, went through setting up, and browsed to one of the pages that comes up blank.  I had a look but I have no idea what I'm looking at, I'm afraid!

dshokouhi:
we need to see the logs from when you register with the app using the external URL please, so clear data and then login to the app using the URL that doesn't work and attempt to go to configuration > Companion App and then get the logs for us please

ACJUK:
[logcat_01-31-2022_22-04-16.txt](https://github.com/home-assistant/android/files/7974570/logcat_01-31-2022_22-04-16.txt)

Sorry, I thought that's what I did.

This one, I started recording, switched to the companion app, entered the Nabu Casa URL, logged in, went to config.  Stopped recording

dshokouhi:
Noticed the following post on the forums

https://community.home-assistant.io/t/new-android-device-does-not-keep-connection-info-between-uses-with-the-companion-app/382766/25?u=dshokouhi

Can you check what URL you used @ACJUK did you by chance include the full URL to a lovelace path instead of just the base nabu casa URL? it should end in `nabu.casa`

Issue #1651, opened 2021-07-29
Path support for saving HAAS URL on the initial, setup screen


juliandroid:
I have HASS setup with NGINX reverse proxy that works properly with URL like this:
https://haas.domain.com/secretpath/ 

Basically NGINX converts that secretpath to HttpOnly Secure cookie that will be returned back with the response.

Since HASS don't support https://haas.domain.com/path/ installations, the very next request will be called directly to the https://haas.domain.com/ but since the cookie is provided by the browser the NGINX will forward the request correctly to the reverse proxy.

That works great in browser, so you only need once to open the https://haas.domain.com/secretpath/ and after that the cookie helps to keep the HASS secure.

The only problem is that when I want to use the Android application the path section is trimmed by the saveUrl method and would be great if the path section is also reconstructed by the Builder. If the path is provided as part of the URL then this setup would work correctly.

The benefit of having this setup is to mitigate the attacks that happens since the sub domain is exposed by the public DNS and all my public services typically face DoS attacks time to time. Hiding HAAS will help greatly from security perspective.

```kt
override suspend fun saveUrl(url: String, isInternal: Boolean?) {
        val trimUrl = if (url == "") null else try {
            val httpUrl = url.toHttpUrl()
            HttpUrl.Builder()
                .scheme(httpUrl.scheme)
                .host(httpUrl.host)
                .port(httpUrl.port)
                .toString()
        } catch (e: IllegalArgumentException) {
            throw MalformedHttpUrlException(
                e.message
            )
        }
        localStorage.putString(if (isInternal ?: isInternal()) PREF_LOCAL_URL else PREF_REMOTE_URL, trimUrl)
    }
  ```
  
 Current NGINX PoC configuration looks like this:
 
```plain
# https server section below

server {
# ....
	set $secretKey shhtKeepItLow;

        location ~* ^/(\w*?)/? {
                set $secret $1;
                set $authenticated 0;
                if ($secret = $secretKey) {
                        set $authenticated 1;
                }
                if ($http_cookie ~* "shhtKeepItLow=1") {
                        set $authenticated 1;
                }
                if ($authenticated = 0) {
                        return 404;
                }

            rewrite /shhtKeepItLow/(.*) /$1 break;

            proxy_pass http://192.168.0.2:8123;    # my local HAAS instance
            proxy_set_header Host $host;
            proxy_redirect http:// https://;
            proxy_http_version 1.1;
[...]

PR #3322, opened 2023-02-09
Only use the base URL during onboarding if the URL is Nabu Casa

dshokouhi:
## Summary

Noticed that during onboarding if a user enters a URL with say the lovelace path like a dashboard from a bookmark then the frontend will not present the correct login screen. This in turn means the app does not save any data upon exit and the user needs to onboard again. Even the Companion App menu item is missing.

example: `https://example.com/lovelace/home`

image

With this change we now force the URL to be just the base URL, there may be a better way to construct the actual base URL but this is what I found as of now :)

Fixes: #2220

## Any other notes

Ref: https://community.home-assistant.io/t/new-android-device-does-not-keep-connection-info-between-uses-with-the-companion-app/382766/25?u=dshokouhi

CC: @bramkragten - I am not sure if its correct that the frontend is serving the incorrect login page. Notice in the screenshot how the user is prompted to save the login.

CC:  @home-assistant/ios-developers - in case a similar fix should be applied elsewhere. I am not sure if iOS also looks at the URL to remove everything but the base URL. I imagine its a common use case for users to copy and paste the URL if discovery did not work.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
@@ -165,8 +165,11 @@ class AuthenticationFragment : Fragment() {
     }
 
     private fun buildAuthUrl(base: String): String {
+        val url = base.toHttpUrl().toUrl()
+        val port = if (url.port > 0) ":${url.port}" else ""
+        val baseUrl = "${url.protocol}://${url.host}$port"
         return try {
-            base.toHttpUrl()
+            baseUrl.toHttpUrl()
                 .newBuilder()
                 .addPathSegments("auth/authorize")
                 .addEncodedQueryParameter("response_type", "code")
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
@@ -166,8 +166,8 @@ class AuthenticationFragment : Fragment() {
     }
 
     private fun buildAuthUrl(base: String): String {
-        val url = base.toHttpUrl()
         return try {
+            val url = base.toHttpUrl()
             HttpUrl.Builder()
                 .scheme(url.scheme)
                 .host(url.host)
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
@@ -168,7 +168,7 @@ class AuthenticationFragment : Fragment() {
     private fun buildAuthUrl(base: String): String {
         return try {
             val url = base.toHttpUrl()
-            if (url.toString().contains("ui.nabu.casa", true)) {
+            if (url.host.endsWith("ui.nabu.casa", true)) {
                 HttpUrl.Builder()
                     .scheme(url.scheme)
                     .host(url.host)
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
@@ -33,6 +33,7 @@ import io.homeassistant.companion.android.onboarding.integration.MobileAppIntegr
 import io.homeassistant.companion.android.themes.ThemesManager
 import io.homeassistant.companion.android.util.TLSWebViewClient
 import io.homeassistant.companion.android.util.isStarted
+import okhttp3.HttpUrl
 import okhttp3.HttpUrl.Companion.toHttpUrl
 import javax.inject.Inject
 import io.homeassistant.companion.android.common.R as commonR
@@ -165,12 +166,12 @@ class AuthenticationFragment : Fragment() {
     }
 
     private fun buildAuthUrl(base: String): String {
-        val url = base.toHttpUrl().toUrl()
-        val port = if (url.port > 0) ":${url.port}" else ""
-        val baseUrl = "${url.protocol}://${url.host}$port"
+        val url = base.toHttpUrl()
         return try {
-            baseUrl.toHttpUrl()
-                .newBuilder()
+            HttpUrl.Builder()
+                .scheme(url.scheme)
+                .host(url.host)
+                .port(url.port)
                 .addPathSegments("auth/authorize")
                 .addEncodedQueryParameter("response_type", "code")
                 .addEncodedQueryParameter("client_id", AuthenticationService.CLIENT_ID)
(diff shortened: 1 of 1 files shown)


jpelgrom:
This conflicts with #1651, but isn't a major issue in my opinion as HA doesn't support installs in subdirectories and the app cannot properly support reverse proxies.

>there may be a better way to construct the actual base URL

`HttpUrl.Builder`? It's already used when adding segments and after onboarding completes when [saving URLs](https://github.com/home-assistant/android/blob/c89184bf31f101efc09a4cd1cc9842783f9549eb/common/src/main/java/io/homeassistant/companion/android/common/data/url/UrlRepositoryImpl.kt#L110).

```kotlin
val url = base.toHttpUrl()
HttpUrl.Builder()
    .scheme(url.scheme)
    .host(url.host)
    .port(url.port)
    .addPathSegments("auth/authorize")
    .addEncodedQueryParameter("response_type", "code")
    .addEncodedQueryParameter("client_id", AuthenticationService.CLIENT_ID)
    .addEncodedQueryParameter("redirect_uri", AUTH_CALLBACK)
    .build()
    .toString()
```

dshokouhi:
> This conflicts with #1651, but isn't a major issue in my opinion as HA doesn't support installs in subdirectories and the app cannot properly support reverse proxies.

hmm in that case we may need to close that request then, in that case the HA frontend would really need to handle the login page properly so we can drop this change. Until then though I think its more user friendly to keep this change as I can see the copy and pasting happening more often than reported.



> `HttpUrl.Builder`? It's already used when adding segments and after onboarding completes when [saving URLs](https://github.com/home-assistant/android/blob/c89184bf31f101efc09a4cd1cc9842783f9549eb/common/src/main/java/io/homeassistant/companion/android/common/data/url/UrlRepositoryImpl.kt#L110).

thank you 🙏

app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
```diff
@@ -162,9 +163,12 @@ class AuthenticationFragment : Fragment() {
     }
 
     private fun buildAuthUrl(base: String): String {
+        val url = base.toHttpUrl()
```
jpelgrom, line 166: `toHttpUrl()` will throw an exception if not valid so better to move it inside the `try`/`catch` block

jpelgrom:
> This in turn means the app does not save any data upon exit and the user needs to onboard again. Even the Companion App menu item is missing.

Decided to replicate the described issue: onboarding never completes which is why nothing is saved and menu item isn't shown. Same issue applies to the iOS app but it is more obvious that onboarding doesn't complete there because of the [sheet](https://user-images.githubusercontent.com/74188/185290789-ca69b486-96d9-432f-ad46-d975889d45e3.png) design.

dshokouhi:
> > This in turn means the app does not save any data upon exit and the user needs to onboard again. Even the Companion App menu item is missing.
> 
> Decided to replicate the described issue: onboarding never completes which is why nothing is saved and menu item isn't shown. Same issue applies to the iOS app but it is more obvious that onboarding doesn't complete there because of the [sheet](https://user-images.githubusercontent.com/74188/185290789-ca69b486-96d9-432f-ad46-d975889d45e3.png) design.

Yea it seems that when this issue occurs not even the external bus communicates with the app. Seems like maybe a frontend change could be needed but the fix here should prevent the issue from occurring in the frontend since we load the correct auth page.

jpelgrom:
> not even the external bus communicates with the app

That's not implemented in the onboarding webview.

> Seems like maybe a frontend change could be needed but the fix here should prevent the issue from occurring in the frontend since we load the correct auth page.

While this PR does prevent the issue from happening, it might break non-standard setups which is why I'm conflicted on whether we should do this or change the frontend instead.

dshokouhi:
> > not even the external bus communicates with the app
> 
> That's not implemented in the onboarding webview.

oh duh, we never leave that webview 😅 



> > Seems like maybe a frontend change could be needed but the fix here should prevent the issue from occurring in the frontend since we load the correct auth page.
> 
> While this PR does prevent the issue from happening, it might break non-standard setups which is why I'm conflicted on whether we should do this or change the frontend instead.

@bramkragten what are your thoughts on how to best proceed here? Seems iOS also has this issue so a frontend fix is probably most beneficial.

dshokouhi:
We should probably come up with a decision here as I am still finding myself helping users who keep hitting this issue. Here is the latest behavior described by a user.

https://community.home-assistant.io/t/new-samsung-phone-ha-app-get-to-loading-data-then-blank-screen/560420

JBassett:
Doing this will break anyone who hosts HA on a path.  If this was the root: `https://example.com/ha` then we would try to auth at `https://example.com/auth/authorize` instead of the correct `https://example.com/ha/auth/authorize`

felipecrs:
I have a suggestion:

1. frontend to support a magic query parameter on any route, and when present, redirect to auth. Examples: 
  1. https://ha:8123/my-dashboard/?redirect-to-auth (frontend should parse and redirect to https://ha:8123/auth)
  2. https://ha:8123/ha/any/given/path?redirect-to-auth (frontend should parse and redirect to https://ha:8123/ha/auth)
2. The app to automatically append `?redirect-to-auth` when attempting to first login.

dshokouhi:
thanks for the suggestion @felipecrs  I think at this point we need to first wait for the frontend to do their part and let us know what if anything we need to do on our end. 

 I am going to close this PR as the approach is not correct per the comment above.

dshokouhi:
Reopening this PR as we are going to target NC URLs only here

app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
```diff
@@ -163,14 +164,28 @@ class AuthenticationFragment : Fragment() {
...
-            base.toHttpUrl()
-                .newBuilder()
-                .addPathSegments("auth/authorize")
-                .addEncodedQueryParameter("response_type", "code")
-                .addEncodedQueryParameter("client_id", AuthenticationService.CLIENT_ID)
-                .addEncodedQueryParameter("redirect_uri", AUTH_CALLBACK)
-                .build()
-                .toString()
+            val url = base.toHttpUrl()
+            if (url.toString().contains("ui.nabu.casa", true)) {
```
jpelgrom, line 168: Why not check if `url.host` ends with it? String comparisons should be avoided.

jpelgrom:
Could you rebase this? Even if there are no merge conflicts, it is based on the last commit before multiserver so good to test with a more recent code base 😅

Image showing where this PR branched from master, and the first commit after it branched being multiserver on February 15, 2023

jpelgrom:
Now to deduplicate code, you could do something like:

```kotlin
// First create the builder either from scratch for NC or from the input otherwise
val builder = if (url.host.endsWith("ui.nabu.casa", true)) {
    HttpUrl.Builder().scheme(url.scheme).host(url.host).port(url.port)
} else {
    url.newBuilder()
}
// Then continue with the common parts
builder.addPathSegments("auth/authorize")
// etc
```

(sorry if this is a lot of small steps, thinking of it as we go)

dshokouhi:
> (sorry if this is a lot of small steps, thinking of it as we go)

no worries appreciate the help and guidance :)

jpelgrom (approved):
Good to have for HA Cloud, which is the only standardised network setup

JBassett (approved):

Issue #3324, opened 2023-02-09
Wear OS add actions to direct notifications


seanmccabe:
**Is your feature request related to a problem? Please describe.**
Not a problem.

**Describe the solution you'd like**
Add actionable notifications to Wear OS notifications.
The app is stated that it should work standalone, but this important feature appears to be missing.
Actionable notifications work when they come through via the phone, would be great if direct notifications can also have actions.

**Describe alternatives you've considered, if any**
Nothing applicable.

dshokouhi:
This will be an interesting one to port over as this feature is tied to many features not found in the Wear OS app.

I don't think we will be able to support [`action: URI`](https://companion.home-assistant.io/docs/notifications/actionable-notifications#uri-values) here given the limitations.

Is the end goal here to get back an event once clicked/replied to?

seanmccabe:
Yer the intent is to send an event back to HA, i.e. to trigger an automation.

Not fussed on the action URI.

Example, have an automation to say my watch has been disconnected from ym phone for 5 minutes, here I want to fire an automation to the watch directly asking if to arm away mode. Need a button to confirm that action back to HA.

PR #3329, opened 2023-02-09
Add support for actionable notifications on Wear OS directly

dshokouhi:
## Summary

Fixes: #3324 by adding support for actionable notifications.

A few things to note:

- Given the dependency on `messagingManager` and that we need variables initialized from the receiver I was unable to move `NotficationActionReceiver` as is so have opted to copy it over.
- `URI` action type is not supported at the moment, only reply and actions.
- Updating reply history on my Pixel Watch results in a weird notification, almost like a bug in how the pixel watch displays things. Same code works fine on the phone so feels like a platform difference.

Documentation: home-assistant/companion.home-assistant#908


jpelgrom:
That's quite a lot of duplicated code... Did you consider making a shared base class like we do for the sensor worker, with the `app`/`wear` `MessagingManager` exposed using an abstract value/function?

dshokouhi:
> That's quite a lot of duplicated code...

I know, not a fan but I also have not come up with a good and viable solution. There are too many dependencies on things in the `app` module that causes a conflict trying to migrate to `common`.

common/src/main/java/io/homeassistant/companion/android/common/notifications/NotificationFunctions.kt
```diff
@@ -274,3 +290,101 @@ fun handleText(
...
+        action.build()
+    )
+        .build()
+}
+
+fun createPendingIntent(
+    context: Context,
+    requestCode: Int,
+    intent: Intent,
+    type: String
```
jpelgrom, line 326: As this is only used as a flag to check if it is a reply or not a boolean would be better. If you have future plans use an enum or constants :)

jpelgrom (commented):
Considering the duplicated receiver again: if `MessagingManager` is the only app/wear specific dependency, couldn't we move it to the common module with an abstract function to be implemented in the app/wear module that uses the `MessagingManager`? By doing it like that the common module doesn't actually know about the messaging, and the only duplicated code is ~8 lines instead of the entire class.

dshokouhi:
leaving as a draft while I look into using abstract functions

dshokouhi:
Closing as I currently do not have enough free time to look into the requested changes

PR #3333, opened 2023-02-15
Revert "Add Matter module preferred to manifest"

JBassett:
Reverts home-assistant/android#3319

This caused us to be unable to upload the app to the Play Store or Firebase.


dshokouhi (approved):
Dont think we could've foreseen that error without trying lol

JBassett:
> Dont think we could've foreseen that error without trying lol

Agreed.

jpelgrom:
Why on earth is it in the example app then? If you have the exact error it might be good to open an issue with Google for this.

PR #3346, opened 2023-02-17
Server gestures and chooser

jpelgrom:
## Summary

Make switching between multiple servers faster with gestures when using the webview (same as on iOS):
 - Swipe left/right to go to the previous/next server in the list
 - Swipe up to open a server chooser where you can select one from the list

(The server chooser is set up to be used in multiple places, but it's only being used for this gesture right now.)

## Screenshots

|     |Light|Dark|
|-----|-----|-----|
|Swiping up shows the new server chooser|Home Assistant interface with a bottom sheet on top titled 'Select server' visible, with two options 'Home' and 'Dev', light mode.|Home Assistant interface with a bottom sheet on top titled 'Select server' visible, with two options 'Home' and 'Dev', dark mode.|
|A hint is added about the gestures underneath the Activate button|Server settings in the app, with the text 'Tip: Quickly activate using a three-finger swipe left, right, or up when viewing a server.' visible below the 'Activate' option, light mode.|Server settings in the app, with the text 'Tip: Quickly activate using a three-finger swipe left, right, or up when viewing a server.' visible below the 'Activate' option, dark mode.|

Documentation: home-assistant/companion.home-assistant#912

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/util/compose/ModalBottomSheet.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/util/compose/ModalBottomSheet.kt
@@ -1,0 +1,66 @@
+package io.homeassistant.companion.android.util.compose
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.Surface
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.res.colorResource
+import androidx.compose.ui.res.dimensionResource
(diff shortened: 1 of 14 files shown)


dshokouhi (approved):
Tested the debug APK and it works as expected, nice feature!

jpelgrom:
Looks like using the chooser will not always clear the history as expected. Changing this to draft while I try to fix it.

app/src/main/java/io/homeassistant/companion/android/util/compose/ModalBottomSheet.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.compose.ui.unit.dp
+import io.homeassistant.companion.android.R
+import io.homeassistant.companion.android.common.R as commonR
+
+/**
+ * A Material 3-style modal bottom sheet with an optional handle, for use with a
+ * [com.google.android.material.bottomsheet.BottomSheetDialogFragment].
+ */
+@Composable
+fun ModalBottomSheet(
```
JBassett, line 30: I know this in alpha but whenever we upgrade we should use the actual library: https://developer.android.com/jetpack/androidx/releases/compose-material3#1.1.0-alpha06

JBassett (approved):

app/src/main/java/io/homeassistant/companion/android/util/compose/ModalBottomSheet.kt
jpelgrom: Material 3 doesn't really play nice with Material 2 which we currently use... that's another project for the future :)

JBassett (approved):

PR #3349, opened 2023-02-18
Sync Thread datasets when starting Matter commissioning

jpelgrom:
## Summary

When starting Matter commissioning, automatically check if there are Thread datasets on the device or in core but not the other, and if so sync them. This new dataset should be marked as 'preferred' by Play Services/core as it is the only dataset available.

The app currently doesn't try to update or delete datasets automatically, for now this should be handled from the Thread config panel (not yet implemented).

## Screenshots

There are no visible changes for importing a dataset to the device / exporting from Home Assistant, this happens in the background.

When exporting a dataset from the device / importing into Home Assistant, a permission dialog will be shown.

Frontend flow: see [the blog post about Thread earlier this month](https://www.home-assistant.io/blog/2023/02/08/state-of-matter-and-thread/#thread), including the [demo video](https://www.youtube.com/watch?v=Fk0n0r0eKcE) for the full flow.
A dialog on top of the frontend titled 'Allow Home Assistant to access your home network?', with the message 'Devices managed by this app can connect to your home network'

Shared device flow: the permission dialog will be shown after confirming to import the device.
|Light|Dark|
|-----|-----|
|A dialog on top of the shared device flow titled 'Allow Home Assistant to access your home network?', with the message 'Devices managed by this app can connect to your home network', light mode|A dialog on top of the shared device flow titled 'Allow Home Assistant to access your home network?', with the message 'Devices managed by this app can connect to your home network', dark mode|

n/a

## Any other notes

For testing:

 - Requires core >= 2023.3.
 - Thread datasets are stored in Play Services. Removing all stored data for Play Services will clear stored credentials, but probably also Matter and Thread support modules so you may need to reinstall Google Home while connected to a charger to get those back quickly.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -168,6 +168,7 @@ dependencies {
 
     "fullImplementation"("com.google.android.gms:play-services-location:21.0.1")
     "fullImplementation"("com.google.android.gms:play-services-home:16.0.0")
+    "fullImplementation"("com.google.android.gms:play-services-threadnetwork:16.0.0-beta02")
     "fullImplementation"(platform("com.google.firebase:firebase-bom:31.1.1"))
     "fullImplementation"("com.google.firebase:firebase-messaging")
     "fullImplementation"("io.sentry:sentry-android:6.14.0")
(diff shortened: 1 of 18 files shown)


dshokouhi (approved):
Code looks good to me, I dont have a matter/thread device so can't really test it but I am sure it works :)

JBassett (approved):
Don't have a matter device to validate but code looks good.

PR #3352, opened 2023-02-20
feature/landscape add setting for fix landscape orientation

dmiales:
This PR adds a setting to fix the app's landscape orientation. This is extremely convenient for lovelace (yaml) display when we do not want to create a separate version of lovelace for mobile devices.
Example.
Looks awful in portrait orientation:

And in the landscape is perfect:

We fix the landscape orientation in the settings

And the user gets a beautiful board


jpelgrom:
Why make this an app feature? Every device has a rotation lock setting.

dmiales:
> Why make this an app feature? Every device has a rotation lock setting.

We cannot fix the webview (or the application as a whole) with the standard android settings.
I mean, take your phone out of your pocket and the app is already in landscape orientation.
Many use for this application, such as rotation control. But it's inconvenient to use.

dshokouhi:
What if a user wants to lock it in portrait instead of landscape?

dmiales:
> Что, если пользователь захочет зафиксировать его в портретном, а не в альбомном режиме?

Interesting. Perhaps in the first iteration, we will see the community's reaction to the landscape fix. In the future, I can implement a drop-down menu with a choice (default/landscape/portrait)

dshokouhi:
> > What if a user wants to lock it in portrait instead of landscape?
> 
> Interesting. Perhaps in the first iteration, we will see the community's reaction to the landscape fix. In the future, I can implement a drop-down menu with a choice (default/landscape/portrait)

Well it just feels incomplete to me to only do landscape. System settings let you switch between auto, portrait and landscape for the whole device

dmiales:
> Ну, мне просто кажется неполным делать только пейзаж

Ok, I'll make an orientation choice, thanks ✅

dmiales:
Continued: https://github.com/home-assistant/android/pull/3362

PR #3353, opened 2023-02-20
feature/minimumFontSize Proportional font styles are very blurry on l…

dmiales:
Proportional font styles are very blurry on low resolution devices. 

Problem:
**minimumFontSize** default is 8
https://developer.android.com/reference/android/webkit/WebSettings#setMinimumFontSize(int)
Example when style **font-size: 1vw** is automatically cast to **8px** (device: honor 20 pro)

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -295,6 +295,7 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                 }
             })
 
+            settings.minimumFontSize = 5
             settings.javaScriptEnabled = true
             settings.domStorageEnabled = true
             settings.displayZoomControls = false


dshokouhi:
Shouldn't this be something controlled from the HA frontend as opposed to the app?

dmiales:
> Shouldn't this be something controlled from the HA frontend as opposed to the app?

Logically yes. But in this case HA is a browser. And the minimum font size is a browser limitation. There is a similar one in firefox https://support.mozilla.org/en-US/kb/font-size-and-zoom-increase-size-of-web-pages or chrome https://mockitt.wondershare.com/guide /solution-for-making-font-size-smaller-than-12-pixels-in-chrome.html

dshokouhi:
> > Shouldn't this be something controlled from the HA frontend as opposed to the app?
> 
> Logically yes. But in this case HA is a browser. And the minimum font size is a browser limitation. There is a similar one in firefox https://support.mozilla.org/en-US/kb/font-size-and-zoom-increase-size-of-web-pages or chrome https://mockitt.wondershare.com/guide /solution-for-making-font-size-smaller-than-12-pixels-in-chrome.html

Yes but this only applies to the app. When you place this code with frontend it will work in the app, in any browser and even iOS. It makes more sense to place this code in the location that benefits everyone instead of just the Android app.

dmiales:
> Yes but this only applies to the app. When you place this code with frontend it will work in the app, in any browser and even iOS. It makes more sense to place this code in the location that benefits everyone instead of just the Android app.

I don't know how it works in iOS. I don't have a device to check the minimum font size. But if we consider other platforms, for example, mobile browsers (moreover, on devices with a small resolution, there is no such problem on large ones), the user can change the minimum size in the browser settings. But in our HA application, there is no such possibility. We don't have webview settings from the interface.

>  It makes more sense to place this code in the location that benefits everyone instead of just the Android app.
Since this is a browser (physical) setting (it cannot be changed with styles) - it changes directly in the application

dshokouhi (approved):
Ran some local tests using a font size renderer thanks to @jpelgrom advice :) Looks ok to me and does not impede on standard HA screens.

Before change

image

After change

image

JBassett (approved):

PR #3357, opened 2023-02-21
Enable/disable sensors per-server

jpelgrom:
## Summary

Allow users to enable/disable sensors (including location tracking) on a server level. By default sensors are the same on all servers, and I expect this will be fine for the majority of users, but it is a nice feature to have.

Syncing enabled state from the server (obviously) works. This change also speeds up sensor updates when one of the servers is down, as it doesn't have to wait on each server to check enabled before sending the state.

## Screenshots

When using multiple servers, a new icon will be added to the toolbar to expand the settings which will show toggles for each server. If not all servers are using the same enabled state, it will always be expanded.

|Light|Dark|
|-----|-----|
|Sensor settings screen for 'Detected Activity' with no major changes, two arrows pointing away from another are visible in the toolbar, light mode.|Sensor settings screen for 'Detected Activity' with no major changes, two arrows pointing away from another are visible in the toolbar, dark mode.|
|Sensor settings screen for 'Detected Activity' with two toggles labeled 'Home: Enable server' and 'Dev: Enable server', two arrows pointing towards each other are visible in the toolbar, light mode.|Sensor settings screen for 'Detected Activity' with two toggles labeled 'Home: Enable server' and 'Dev: Enable server', two arrows pointing towards each other are visible in the toolbar, dark mode.|

Documentation: home-assistant/companion.home-assistant#914

## Any other notes

The code in the sensor receiver is now similar to [pre-multiserver code](https://github.com/home-assistant/android/blob/76690b371abe11c6d3b5cfb13dd3ee3276f4eb8b/common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt#L195) again.

I have also tried other designs where the expand/collapse control wasn't placed in the toolbar or checkboxes were used, but this often felt cluttered or wasn't as good when only using one server.
Concept design where a down/up arrow is visible directly left to the 'Enable server' text

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
@@ -149,7 +149,7 @@ class LocationSensorManager : LocationSensorManagerBase() {
         private var zones = mutableMapOf<Int, Array<Entity<ZoneAttributes>>>()
         private var zonesLastReceived = mutableMapOf<Int, Long>()
 
-        private var geofenceRegistered = mutableListOf<Int>()
+        private var geofenceRegistered = mutableSetOf<Int>()
 
         private var lastHighAccuracyMode = false
         private var lastHighAccuracyUpdateInterval = DEFAULT_UPDATE_INTERVAL_HA_SECONDS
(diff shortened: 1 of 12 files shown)


dshokouhi (approved):
Tested debug APK and it works well here, great feature!

JBassett (approved):

PR #3366, opened 2023-02-24
Enable location tracking toggle in additional server onboarding

jpelgrom:
## Summary

Enable the toggle for location tracking in onboarding when adding additional servers (when using the full flavor). This was previously disabled as sensors were shared between all servers but the app supports server-level sensor enabling now.

## Screenshots

It's the same toggle as in initial onboarding

n/a, covered by note about multi-server sensors


dshokouhi (approved):
Looks good to me!

JBassett (approved):

Issue #3367, opened 2023-02-25
Beacon monitor (BackgroundScanPeriod)


leroyloren:
**Is your feature request related to a problem? Please describe.**
The background beacon setting does not respect the set value

**Describe the solution you'd like**
To match the scan period to the settings, at the moment the set scan interval only seems to work when the application is in the foreground

**Describe alternatives you've considered, if any**
If the scan interval is set to e.g. 1s, scan both in the foreground and in the background.

**Additional context**
https://altbeacon.github.io/android-beacon-library/javadoc/org/altbeacon/beacon/BeaconManager.html#setBackgroundMode(boolean)

dshokouhi:
This method is deprecated and I think we already do this in the code.

leroyloren:
Well it's hard to say, but it scans in the background for about 10+ sec, definitely not according to the set period.

dshokouhi:
Are you using the beacon monitor sensor to determine that? Or something else?

We already have the sensor as part of a foreground service too.

leroyloren:
Yes, beacon monitor sensor, de/activated by command_beacon_monitor

dshokouhi:
Have you checked the logs to determine an error is not showing up at that time?

leroyloren:
There are no errors, only the scan is considerably delayed, the expected 1s period is realistically min. 10s

dshokouhi:
Are you seeing this behavior on phone or watch?

leroyloren:
Phone

dshokouhi:
Looking at the code we are indeed using the same scan period and interval for both foreground and background

https://github.com/home-assistant/android/blob/db63a38cd5fc2b26ecd02f8cb8b157bcd0aa110d/common/src/main/java/io/homeassistant/companion/android/common/bluetooth/ble/MonitoringManager.kt#L45-L48

Looking at altbeacon docs we may need to look into using a foreground service scan but this will mean a notification will be present. Your symptoms seem to be described here as well.

https://altbeacon.github.io/android-beacon-library/foreground-service.html

We do have our own foreground service for updating sensors but we may need a second one based on what I am reading here.

dshokouhi:
Since it sounds like you are able to reproduce this pretty reliably I would appreciate it if you could test the debug APK by downloading the artifact from the below link.

https://github.com/home-assistant/android/actions/runs/4271958665

[...]

PR #3369, opened 2023-02-25
Use a foreground service for reliable beacon scanning in background

dshokouhi:
## Summary

Fixes: #3367 by using a foreground service to make beacon scanning more reliable in the background. Keeping this as a draft to gain feedback to see if this should be a separate setting or not. Although it changes the behavior making it more reliable should be the default IMO. ~~Still need to do some clean up (like strings) but before doing so~~ want some end user feedback to see if the monitor continues to scan in the background.

The [library docs](https://altbeacon.github.io/android-beacon-library/foreground-service.html) also mention this code is mainly for the beacon monitor but can also benefit the transmitter if used in conjunction so this may fix 2 bugs in one like #3159 

During testing I also noticed at times the monitor may not stop scanning if the sensor was disabled so adding a check for that as well.

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#916


dshokouhi:
Marking PR as ready for review as we have received confirmation from the user that the behavior is now as they would expect.

jpelgrom:
If the app is going to post a persistent notification, maybe it's a good idea to include an action to stop scanning / go to the settings where you can stop scanning?

dshokouhi:
> If the app is going to post a persistent notification,

In this case its actually not persistent :) I wasn't sure if we should make it persistent in this case. Our sensor worker does not appear to be persistent either.



> maybe it's a good idea to include an action to stop scanning / go to the settings where you can stop scanning?

I agree, good idea to add an action to stop scanning, I'll look into that.

dshokouhi:
Actually I forgot in Android 13 by default users can dismiss a notification unless we mark it as ongoing, so should we do that here? Personally I like letting the user dismiss it as they can still stop the service in Android 13 using the foreground service manager. Some may prefer the convenience of the new action button always being there too so not sure whats the best way to proceed. We could just keep it as is and modify it based on beta feedback as well.

https://developer.android.com/guide/components/foreground-services#user-dismiss-notification

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
```diff
@@ -102,6 +103,11 @@ abstract class SensorReceiverBase : BroadcastReceiver() {
             }
         }
 
+        if (intent.action == ACTION_STOP_BEACON_SCANNING) {
+            BluetoothSensorManager.enableDisableBeaconMonitor(context, false)
+            return
```
jpelgrom, line 108: If the stop action is used, should the sensor be allowed to update instead of returning before updates are sent?

common/src/main/res/values/strings.xml
```diff
@@ -1041,4 +1041,6 @@
     <string name="not_registered">Please launch the Home Assistant app and login before you can use the assist feature.</string>
     <string name="ha_assist">HA: Assist</string>
     <string name="only_favorites">Only Show Favorites</string>
+    <string name="beacon_scanning">Beacon Scanning</string>
```
jpelgrom, line 1044: To match the sensor name:
```suggestion
    <string name="beacon_scanning">Beacon Monitor Scanning</string>
```

common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
dshokouhi: We actually update sensors inside the above method so this was added to prevent 2 runs

jpelgrom: Looked at the wrong function, ignore the previous comment 😅

common/src/main/res/values/strings.xml
leroyloren: and stop_scanning maybe only stop/disable same as high accurancy action notification?

obrazek

in case of translation it can be a string 🤔

dshokouhi: There is already an action like that

jpelgrom: I think @leroyloren was suggesting changing the action to 'Stop' or 'Disable', so a more generic string that may already be translated and short like other notifications with actions posted by the app. 

(Existing actions for other notifications are 'Disable' or 'Settings')

jpelgrom (approved):

JBassett (approved):

PR #3380, opened 2023-02-28
Improve sensor re-registration on app version change for offline server

jpelgrom:
## Summary

Prevent the `SensorWorker` runs from taking a long time if a server is offline when the app version changes due to re-registration of all sensors (~15 min, runs sequentially), by skipping these re-registrations if one fails with an exception that indicates the server can't be reached.

This isn't a new issue but more obvious with multiserver as it is more likely to have a server that is frequently offline/unreachable. The app version doesn't change with debug builds so this was spotted by @dshokouhi with updates on the internal test track :)

## Screenshots

n/a

n/a


dshokouhi (approved):
Looks good to me, thank you for the quick fix :)

Issue #3384, opened 2023-03-02
BLE Beacon Transmitter Notification Command Not updating sensor until Companion app is restarted


birdwing:
**Home Assistant Android app version(s):**
2023.1.1-full

**Android version(s):**
13

**Device model(s):**
Pixel 7

**Home Assistant version:**
2023.3.0

**Last working Home Assistant release (if known):**

**Description of problem, include YAML if issue is related to notifications:**
Notification Command for turning on/off BLE Beacon Transmitter doesn't work until companion app is restarted and the "Updating Sensors" home assistant notification comes up.

Here is the YAML I am using to turn on/off the BLE Beacon Transmitter
```yaml
service: notify.mobile_app_aron_s_phone
data:
  message: command_ble_transmitter
  data:
    command: turn_on
```
```yaml
service: notify.mobile_app_aron_s_phone
data:
  message: command_ble_transmitter
  data:
    command: turn_off
```

Steps to reproduce (these are the steps captured in the logs posted below)
_Note: I am watching the BLE Transmitter sensor in a web browser on my computer to watch for updates as I test notification commands_

1. Start with BLE Beacon Transmitter enabled (transmitting).
2. Send notification command to Turn off BLE Beacon Transmitter.
3. Nothing happens, sensor didn't change
4. Open HA Companion app
5. Updating Sensors notification appears, BLE Beacon Transmitter now changes to Stopped.
6. With app Still open
7. Send notification command to Turn On BLE Beacon Transmitter.
8. Nothing Happens, sensor didn't change
9. Close HA Companion app, Nothing happens with the sensor
10. Open HA Companion app
11. Updating Sensors notification appears, BLE Beacon Transmitter now changes to Transmitting.

It appears that turning On or off the BLE Beacon Transmitter doesn't actually update the sensor until the ```io.homeassistant.companion.android.UPDATE_SENSOR``` intent is received (when the app is opened).
Even though It received and processed the command according to earlier entries in the logs.

I am not sure if the BLE Transmitter is turning on/off but just not updating the sensor, or if the BLE Transmitter itself is not turning on/off until the app is restarted. I will check on this when I get home and have my wifes phone to test with (to see if the transmitter turns on and is picked up by her phone, even though the sensor hasn't updated to say it's transmitting).

It is possible that the transmitter is turning on, but that the sensor itself is simply not updating. Which makes using the sensor to detect if the notification was received impossible. 

**Companion App Logs:**

<details>
<summary>Logs</summary>

```
[...]

PR #3385, opened 2023-03-02
Update sensors when BLE transmitter turned on and off

dshokouhi:
## Summary

Fixes: #3384 by updating sensors anytime the command is processed 

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/notifications/DeviceCommands.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/notifications/DeviceCommands.kt
@@ -164,8 +164,8 @@ fun commandBleTransmitter(
                 null
             )
         }
-        BluetoothSensorManager().requestSensorUpdate(context)
-        SensorUpdateReceiver.updateSensors(context)
     }
+    BluetoothSensorManager().requestSensorUpdate(context)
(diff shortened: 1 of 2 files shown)


jpelgrom:
Is there a particular reason for not adding this to `enableDisableBLETransmitter` instead, to match how beacon monitor commands are processed? That function already handles starting/stopping on change and updating the state.

dshokouhi:
> Is there a particular reason for not adding this to `enableDisableBLETransmitter` instead, to match how beacon monitor commands are processed? That function already handles starting/stopping on change and updating the state.

Beacon Monitor only allows for enable and disable, the BLE transmitter also allows changing other sensor settings which get reflected in the attributes so we also want to update that.

jpelgrom:
OK should the starting/stopping be removed from `enableDisableBLETransmitter` then as it's also handled in the sensor update? Less duplicate code.

dshokouhi:
> OK should the starting/stopping be removed from `enableDisableBLETransmitter` then as it's also handled in the sensor update? Less duplicate code.

Good call, just tested removing that code and it seems to work as intended

jpelgrom (approved):
Tested and it appears to be working correctly

JBassett (approved):

PR #3387, opened 2023-03-02
Update Play Store listing for upcoming play policy changes for Wear OS

dshokouhi:
## Summary

Upcoming play policy changes want us to update the app description and to provide more screenshots. I felt that it made sense to call out some phone features along with Android Auto support since we need to list features for Wear OS. I have added screenshots here of the tiles. One of our screenshots should be adjusted for onboarding but I want to hold off on that for now since we may change onboarding again. I also think we may want to update our Template tile example to make it more like an example like the shortcut tile example :)

https://developer.android.com/docs/quality-guidelines/wear-app-quality#upcoming

```
Your app listing on Google Play Store must adhere to the following:

Accurately list the latest build and features.
Mention Wear OS.
Do not mention Android Wear.
Be localized in languages offered by the app.
```

and

```
Your app listing on Google Play Store must:

Contain at least one screenshot that accurately depicts the current version of the app on Wear OS.
Provide screenshots showing only the app interface.
Not position the screenshots within device frames, or include additional text, graphics, or backgrounds that are not part of the interface of the app.
Include screenshots with a 1:1 aspect ratio.
If your app offers Tiles, then we recommend sharing a screenshot of Tiles functionality.
```

## Any other notes

I am unsure if I have the format correct here.

Whats the best way for us to get the description localized? Maybe there is a way for us to add this to Lokalize so the community can help there? Admittedly I am not the best for integrated the 2 😁


balloob (approved):

jpelgrom:
> One of our screenshots should be adjusted for onboarding

Or drop the remaining onboarding screenshot, it's not that interesting?

dshokouhi:
> > One of our screenshots should be adjusted for onboarding
> 
> Or drop the remaining onboarding screenshot, it's not that interesting?

I would prefer to keep the first screen a reviewer will see instead of removing it :)

dshokouhi:
I am going ot update phone screenshots as well to be consistent with Wear OS

jpelgrom:
I like the idea of updating phone screenshots but we should at least make sure they all have the same size. We should try following [guidelines](https://support.google.com/googleplay/android-developer/answer/9866151?hl=en), especially:

> Minimum dimension: 320px
not met for widgets screenshots, why not show an actual homescreen instead of a cropped screenshot

> - Edit excess elements in the notification bar before submitting. Do not show service providers or notifications. The battery, WiFi, and cell service logos should be full.
by using demo mode in developer settings

dshokouhi:
@jpelgrom  as always appreciate the attention to detail 🙏

dshokouhi:
> I still wouldn't cut off the bottom of the screenshot for widgets / update the screenshot `1.png` to have the same size as other screenshots.

ok in that case might as well add them all!

jpelgrom:
Phone screenshot 1 still has a different aspect ratio, as it wasn't retaken. Phone screenshot 4 has notification icons in the status bar (demo mode bug? easy to edit the screenshot as the background is black).

jpelgrom (approved):

JBassett (approved):

PR #3393, opened 2023-03-03
Google dependency bumps

jpelgrom:
## Summary

Update various Google dependencies that aren't picked up by dependabot to their latest stable versions. Did a local build and didn't encounter any issues.

https://developers.home.google.com/thread/release-notes#2023-02-27
https://firebase.google.com/support/release-notes/android#messaging_v23-1-2
https://github.com/google/ExoPlayer/releases/tag/r2.18.3
https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.4.3

## Screenshots

n/a

n/a

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -44,7 +44,7 @@ android {
     }
 
     composeOptions {
-        kotlinCompilerExtensionVersion = "1.4.2"
+        kotlinCompilerExtensionVersion = "1.4.3"
     }
 
     kotlinOptions {
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -168,8 +168,8 @@ dependencies {
 
     "fullImplementation"("com.google.android.gms:play-services-location:21.0.1")
     "fullImplementation"("com.google.android.gms:play-services-home:16.0.0")
-    "fullImplementation"("com.google.android.gms:play-services-threadnetwork:16.0.0-beta02")
-    "fullImplementation"(platform("com.google.firebase:firebase-bom:31.2.2"))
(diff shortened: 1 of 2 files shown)


app/build.gradle.kts
```diff
@@ -179,18 +179,18 @@ dependencies {
...
+    "fullImplementation"("com.google.android.exoplayer:extension-cronet:2.18.3")
+    "minimalImplementation"("com.google.android.exoplayer:extension-cronet:2.18.3") {
         exclude(group = "com.google.android.gms", module = "play-services-cronet")
     }
     "minimalImplementation"("org.chromium.net:cronet-embedded:108.5359.79")
 
     implementation(platform("androidx.compose:compose-bom:2023.01.00"))
     implementation("androidx.compose.animation:animation")
-    implementation("androidx.compose.compiler:compiler:1.4.2")
+    implementation("androidx.compose.compiler:compiler:1.4.3")
```
JBassett, line 193: Don't you also need to bump the compiler version here: https://github.com/home-assistant/android/blob/adf646b24d2483d7ddeb44946c25cc0537ddfe35/app/build.gradle.kts#L47

jpelgrom: I probably should - interesting it didn't complain during build

JBassett (approved):

app/build.gradle.kts
dshokouhi: > I probably should - interesting it didn't complain during build

it did for me previously but that was due to compose vs kotlin vesrion mismatch. It probably only cares if it cannot compile correctly due to minimum version requirements?

PR #3398, opened 2023-03-05
Fix listed beacons that are out of range after enable

Adi146:
## Summary

This MR fixes a bug, where the beacon monitor lists iBeacons that are out of range after enabling.
https://github.com/home-assistant/android/issues/3010 

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/bluetooth/ble/IBeaconMonitor.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/bluetooth/ble/IBeaconMonitor.kt
@@ -38,6 +38,10 @@ class IBeaconMonitor {
         }
     }
 
+    fun clearBeacons() {
+        beacons = listOf()
+    }
+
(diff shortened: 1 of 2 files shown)


jpelgrom:
How does this fix [the issue the previous PR to fix this encountered](https://github.com/home-assistant/android/issues/3010#issuecomment-1387437327), where the live data is caching/sending old values?

Adi146:
The Beacon Library finishes the current scan when we stop it and as we unbind the observer on stop we can use the check if any observer is bound. Otherwise we would empty the list and add the beacon which was found on the last scan to the empty list again. That's why the previous PR didn't work.

dshokouhi (approved):
Code looks good to me and we have a user who has verified the fix for us. Thanks for the fix! 🙏

leroyloren:
Yes, this is the only thing missing to make the Beacon Monitor work flawlessly ;-)

JBassett (approved):

Issue #3260, opened 2023-01-24
Add support for CarInfo


superm1:
**Is your feature request related to a problem? Please describe.**

Add support to the mobile application for Android to request [CarInfo](https://developer.android.com/reference/androidx/car/app/hardware/info/CarInfo) information when connected to Android Auto.

Of particular interest to me is what the current connected car is and what the EV battery level is, but I can see tons of opportunity with other things that can be reported as well.

**Describe the solution you'd like**

Add sensors to indicate Car Information.

**Describe alternatives you've considered, if any**

Using EVNotify and OBD2 dongle to get EV level.  This is functional but a bit clunky and insecure.  Since Android auto offers such an API, it makes sense to me that the HA AA application should use it!

**Additional context**

This was found as part of reading through the [announcement for HA android auto support](https://www.home-assistant.io/blog/2023/01/20/android-auto/) and looking at the API documentation for AA.

OzGav:
+1 this would be next level. I am also trying to pull info from the car using an OBD dongle but it is flaky at best. Access to EV battery level is also what I am after as well as total km (odometer)

PR #3399, opened 2023-03-05
Add CarInfo sensors

drosoCode:
## Summary
Add basic support for [CarInfo](https://developer.android.com/reference/androidx/car/app/hardware/info/CarInfo) sensors (as mentionned in #3260)

## Screenshots

Screenshot_20230305-234536_Home Assistant (2)

Documentation: [home-assistant/companion.home-assistant#936](https://github.com/home-assistant/companion.home-assistant/pull/936)

## Any other notes
Can be tested using the [DHU](https://developer.android.com/training/cars/testing?hl=fr#running-dhu) by running it with `-c config/default_sensors.ini` and typing commands like `fuel 40` in the cli.

For some reason this doesn't work on my oneplus (7T pro, android 12) but seems to works fine on samsung (tested on S23 (android 13) and S9+ (android 10)).

As using the [CarContext](https://developer.android.com/reference/androidx/car/app/CarContext) seems to be the only way to obtain the CarInfo class, the android auto app needs to be started to make the sensor work.


app/build.gradle.kts
```diff
@@ -22,7 +22,7 @@ android {
 
     defaultConfig {
         applicationId = "io.homeassistant.companion.android"
-        minSdk = 21
+        minSdk = 23
```
dshokouhi, line 25: This will be a very big breaking change for us as it means the app will no longer be available to these devices, is it absolutely necessary for this feature?

drosoCode: The [CarHardware](https://developer.android.com/training/cars/apps#car-hardware-requirements) APIs are provided by the library `androidx.car.app:app-projected` which declares a minSdk of 23 but I guess we could aslo use [overrideLibrary](https://developer.android.com/studio/build/manage-manifests?hl=fr#override_uses-sdk_for_imported_libraries) in the manifest.

dshokouhi: If we could do that we should as we can hide the sensors from unsupported devices using the `hasSensor` method

app/build.gradle.kts
```diff
@@ -210,6 +210,7 @@ dependencies {
     implementation("com.github.AppDevNext:ChangeLog:3.4")
 
     "fullImplementation"("androidx.car.app:app:1.3.0-rc01")
+    implementation("androidx.car.app:app-projected:1.2.0")
```
dshokouhi, line 213: this should also use the `fullImplementation` as these sensors are not offered in the minimal flavor and we need to make sure no google service dependency is added to that flavor.

app/src/full/java/io/homeassistant/companion/android/sensors/CarInfoSensorManager.kt
```diff
@@ -0,0 +1,116 @@
...
+        get() = R.string.sensor_name_car_info
+
+    override suspend fun getAvailableSensors(context: Context): List<SensorManager.BasicSensor> {
+        return listOf(
+            fuelLevel, batteryLevel
+        )
+    }
+
+    override fun requiredPermissions(sensorId: String): Array<String> {
+        return arrayOf("com.google.android.gms.permission.CAR_FUEL")
```
dshokouhi, line 59: Is this also required for the battery sensor or just the fuel sensor?

app/src/full/java/io/homeassistant/companion/android/sensors/CarInfoSensorManager.kt
```diff
@@ -0,0 +1,116 @@
...
+            isListenerRegistered = true
+            listenerLastRegistered = now.toInt()
+        } else {
+            Log.d(TAG, "CarInfo not available")
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    override fun onCarDataAvailable(data: EnergyLevel) {
+        if (data.fuelPercent.status == CarValue.STATUS_SUCCESS) {
```
dshokouhi, line 94: We should make sure to skip the updates for the sensors that are not enabled here. It does not seem like we can avoid any particular update here right? The listener looks to be all or nothing.

app/src/full/java/io/homeassistant/companion/android/sensors/CarInfoSensorManager.kt
drosoCode: Yes, both are provided by the same callback, [here](https://developer.android.com/training/cars/apps#carinfo) is the list of required permissions. We may want to filter on the sensorId if we add other listeners like mileage that requires different permissions.

app/src/full/java/io/homeassistant/companion/android/sensors/CarInfoSensorManager.kt
drosoCode: Yes, both values are returned in the same object.

Maybe we could add a check for each sensor like this:
```kotlin
if(data.fuelPercent.status == CarValue.STATUS_SUCCESS && isEnabled(latestContext, fuelLevel))
```

dshokouhi:
> As using the [CarContext](https://developer.android.com/reference/androidx/car/app/CarContext) seems to be the only way to obtain the CarInfo class, the android auto app needs to be started to make the sensor work.

Is there really no way to get this in the background? I wonder if there is maybe something we can do to make the situation better then for users to avoid cases where they may forget to launch the app. I don't think a notification command can open a car app right? I wonder if we can use the existing intent for the connection sensor? 🤔 

We also need to get a docs PR attached to this PR (once ready of course)

drosoCode:
I didn't found any way of getting the CarContext in the background. But I did find a way to show a notification on the car screen that is able to launch the android auto app when clicked (see the picture below). 

Just for testing purposes, I put the notification code with the existing connection sensor, but I am wondering if I should put everything in the existing AndroidAutoSensorManager (since it's also android auto related)

notif

jpelgrom:
IMO you shouldn't force a notification when connecting to Android Auto. If the user wants this, they can do it themselves using the Android Auto connected sensor.

dshokouhi:
> IMO you shouldn't force a notification when connecting to Android Auto. If the user wants this, they can do it themselves using the Android Auto connected sensor.

I was under the impression a standard HA notification will not show up, at least in my Android Auto setup they do not show up at all. I think they need to be a conversation or extend the car platform. Unless Android Automotive does something different than Android Auto when it comes to showing a notification.

jpelgrom:
> I was under the impression a standard HA notification will not show up, at least in my Android Auto setup they do not show up at all. I think they need to be a conversation or extend the car platform. Unless Android Automotive does something different than Android Auto when it comes to showing a notification.

Haven't tested it but if something special is required shouldn't that be added as a 'normal' notification option instead of forcing it in here?

dshokouhi:
> > I was under the impression a standard HA notification will not show up, at least in my Android Auto setup they do not show up at all. I think they need to be a conversation or extend the car platform. Unless Android Automotive does something different than Android Auto when it comes to showing a notification.
> 
> Haven't tested it but if something special is required shouldn't that be added as a 'normal' notification option instead of forcing it in here?

That may be a lil trickier than normal as the extended notification code is coming from the car app library which is only available to users on the full version of the app.

app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
```diff
@@ -25,29 +46,105 @@ class AndroidAutoSensorManager : SensorManager, Observer<Int> {
...
+            else -> emptyArray()
+        }
     }
 
     private lateinit var context: Context
     private var carConnection: CarConnection? = null
 
+    private fun allDisabled(): Boolean {
+        return !sensorsList.map { x -> isEnabled(context, x) }.contains(true)
+    }
```
jpelgrom, line 142: Using a few Kotlin features

```suggestion
    private fun allDisabled(): Boolean =
        sensorsList.none { isEnabled(context, it) }
```

app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
```diff
@@ -79,14 +178,189 @@ class AndroidAutoSensorManager : SensorManager, Observer<Int> {
...
+            listenerLastRegistered[l] = System.currentTimeMillis()
+        } else {
+            listenerLastRegistered[l] = -1L
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    private fun updateCarInfo() {
+        for (l in listenerSensors) {
+            if (l.value.map { x -> isEnabled(context, x) }.contains(true)) {
```
jpelgrom, line 282:
```suggestion
            if (l.value.any { isEnabled(context, it) }) {
```

app/src/full/java/io/homeassistant/companion/android/vehicle/HaCarAppService.kt
```diff
@@ -65,14 +70,21 @@ class HaCarAppService : CarAppService() {
                 1
             )
 
+            @RequiresApi(Build.VERSION_CODES.P)
```
jpelgrom, line 73: Android Auto is available on versions 8+ (Oreo). As a result, this annotation should be removed and car info should do a version check instead of removing Auto support entirely for Android 8..

dshokouhi:
What do we think about also updating the state of sensors that require `CarContext` to mention that the app needs to be started? It would be similar behavior to the Bluetooth Transmitter and Beacon Monitor sensors when Bluetooth is off. This would allow the sensors to have a better state instead of being stuck until the next time it is updated.

jpelgrom:
> > Haven't tested it but if something special is required shouldn't that be added as a 'normal' notification option instead of forcing it in here?
> 
> That may be a lil trickier than normal as the extended notification code is coming from the car app library which is only available to users on the full version of the app.

Good point. Still, I don't like this and would prefer giving the user control. I can't imagine the notification _always_ being useful. "CarInfo" is also quite technical.

Something with a extension function returning different information depending on the flavor, similar to how code is shared between the main app and Wear OS, could work to make a notification feature to get the required intent, but it is out of scope of the PR.

app/src/full/java/io/homeassistant/companion/android/sensors/CarInfoSensorManager.kt
```diff
@@ -0,0 +1,112 @@
...
+            onSensorUpdated(
+                latestContext,
+                batteryLevel,
+                data.batteryPercent.value!!,
+                batteryLevel.statelessIcon,
+                mapOf()
+            )
+        }
+        HaCarAppService.carInfo?.removeEnergyLevelListener(this)
+        Log.d(TAG, "CarInfo sensor listener unregistered")
```
nick2525: Please use some logger instead of Log.d. This allows you to have more control over the logging behavior and can be easier to configure than using Log directly.

dshokouhi: The project is not setup to use something other than `Log.*` at the moment

nick2525: ok, it is sad

app/src/full/java/io/homeassistant/companion/android/vehicle/HaCarAppService.kt
```diff
@@ -31,6 +34,8 @@ class HaCarAppService : CarAppService() {
 
     companion object {
         private const val TAG = "HaCarAppService"
+        var carInfo: CarInfo? = null
```
googlvalenzuela, line 37: This service should only be used by the Templates to display data and not for getting data; moreover this service will not be boun on Android Auto Project until the user taps on the application

app/src/full/java/io/homeassistant/companion/android/vehicle/HaCarAppService.kt
```diff
@@ -66,13 +71,19 @@ class HaCarAppService : CarAppService() {
...
                 return MainVehicleScreen(
                     carContext,
                     serverManager,
                     serverIdFlow,
                     entityFlow
                 ) { loadEntities(lifecycleScope, it) }
             }
+
+            override fun onNewIntent(intent: Intent) {
+                carContext.getCarService(ScreenManager::class.java).popToRoot()
```
googlvalenzuela, line 85: This will create a bug to always go back to the main screen, is this intentational?

app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
googlvalenzuela, line 1: This will only work when the user has the started the applicationo on Android Auto Projected (the concept of carContext is only true then).
For the case of Android Automotive this will require some refactor as you can get CarInfo directly.

app/src/full/java/io/homeassistant/companion/android/vehicle/HaCarAppService.kt
drosoCode: This is used to show the main screen when we receive the intent from the start notification. Is there any problem with this ?

app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
drosoCode: Yes for AA projected, I implemented a notification to ask the user to start the app. I don't really know for Android automotive, but that's great news if we can access directly the CarInfo

app/src/full/java/io/homeassistant/companion/android/vehicle/HaCarAppService.kt
dshokouhi: I think we may need to filter by the intent action that we expect to avoid changing the screen on accident if we get an unexpected action.

app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
```diff
@@ -79,14 +203,190 @@ class AndroidAutoSensorManager : SensorManager, Observer<Int> {
...
             )
-        )
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    private fun startAppNotification() {
+        val manager = CarNotificationManager.from(context)
+
+        val channelID = "HA_AA_OPEN"
```
dshokouhi, line 232: This needs to be added to the internal channel list that we maintain.

https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/util/AppNotifChannels.kt

app/src/full/java/io/homeassistant/companion/android/vehicle/HaCarAppService.kt
dshokouhi: Personally speaking I am not a fan of the notification route but I understand its one of the only ways to go about it. I think a better approach would be to put things in the hand of the user to chose to send the notification or not. Ideally we can use `command_activity` or a new command to launch the app from the background when a certain command is received. Putting it in the hands of the user allows them to add their own conditions for when they want things to auto start.

app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
```diff
@@ -25,29 +46,112 @@ class AndroidAutoSensorManager : SensorManager, Observer<Int> {
             "mdi:car",
             updateType = SensorManager.BasicSensor.UpdateType.INTENT
         )
+        private val fuelLevel = SensorManager.BasicSensor(
+            "android_auto_fuel_level",
```
dshokouhi, line 50: Given that these new sensors are also for Android Automotive I wonder if we should consider a different naming approach here? I understand we already have the connection sensor up but that does not really matter for Android Automotive as the phone may not be connected. Not really sure what makes the most sense here so thought I would spark the convo :)

app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
```diff
@@ -25,29 +46,112 @@ class AndroidAutoSensorManager : SensorManager, Observer<Int> {
...
+        )
+        private val batteryLevel = SensorManager.BasicSensor(
+            "android_auto_battery_level",
+            "sensor",
+            commonR.string.basic_sensor_name_android_auto_battery_level,
+            commonR.string.sensor_description_android_auto_battery_level,
+            "mdi:car-battery",
+            unitOfMeasurement = "%",
+            stateClass = SensorManager.STATE_CLASS_MEASUREMENT,
+            deviceClass = "battery"
```
dshokouhi, line 67: this should probably be set as diagnostic type for entity category to match the other battery level sensor we have

https://developers.home-assistant.io/blog/page/7?_highlight=diagnostic#entity-categories

May also want to look at the other sensors to see if they should be diagnostic or a primary sensor.

app/src/full/java/io/homeassistant/companion/android/vehicle/HaCarAppService.kt
drosoCode: I added an intent filter, but realised that in the end this wasn never called (even without the filter), so I just removed this part.

app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
drosoCode: I kinda agree but I'm not really sure of what to use as a prefix that would make sense for both

app/src/full/java/io/homeassistant/companion/android/vehicle/HaCarAppService.kt
drosoCode: I tried to launch the activity in background using the same intent from the notification with something like this: 
```kotlin
val intent = Intent(Intent.ACTION_VIEW).setComponent(ComponentName(context, HaCarAppService::class.java))
CarPendingIntent.getCarApp(context, intent.hashCode(), intent, 0).send()
```
but it doesn't seem to work, I guess that you need to include it in a notification to get it working.

dshokouhi: to start the intent it should be something more like `context.startActivity(intent)` 

pending intent is actually reserved for things like the notification or for something else to start it.

app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
dshokouhi: Yea we started to have a small discussion on this. This has the potential of creating a lot more sensor entities which we obviously don't want to say no to but we should make sure the sensor management organization is good. Once we merge with the IDs we won't be able to change that without causing a breaking change for users.

@jpelgrom / @JBassett  what do you guys think? Should we keep all the android auto/automotive sensors together? should we consider another naming convention so we dont have so many added in one area? a lot of things to consider here so we dont introduce a breaking change later.

jpelgrom: Split up because Automotive has a minimal flavor while Android Auto sensors will only ever be for the full flavor.

dshokouhi: I like `car` and `vehicle` prefixes actually. Good point about it being split up due to the automotive minimal flavor. So these should be added in either `main` or `common` then correct?

dshokouhi:
@drosoCode  now that number #3642 is merged please remove the code specific to posting a notification on the device. Users can now determine which notifications will show in the car and those notifications will take the user to HA car app service.

Please also fix the merge conflicts. 

I think at this point the last remaining item would be that the sensors probably need to be migrated to the `main` or `common` folder as these sensors should be available for the minimal automotive builds. Current location only applies to the play store builds.

dshokouhi:
@drosoCode just a note I updated my above comment, the new location should be in `main` so the sensors can be shared with the minimal automotive build.

drosoCode:
Ok, shoud I leave the android auto connection sensor here or also move it in main ?

dshokouhi:
> Ok, shoud I leave the android auto connection sensor here or also move it in main ?

It can stay here (`full`) as there is no such thing as Open Source Android Auto so it shouldn't be functional in non-play store builds.

common/src/main/res/values/strings.xml
```diff
@@ -56,6 +56,11 @@
     <string name="basic_sensor_name_bluetooth_ble_beacon_monitor">Beacon Monitor</string>
     <string name="basic_sensor_name_bluetooth_state">Bluetooth State</string>
     <string name="basic_sensor_name_bluetooth">Bluetooth Connection</string>
+    <string name="basic_sensor_name_car_battery_level">Battery Level</string>
```
dshokouhi, line 59: This string duplicates `basic_sensor_name_battery_level` at line 52 just above, we should either remove or rename it so its unique

drosoCode: I renamed it to `Car Battery Level` so that it isn't confused with the phone battery level

app/src/full/AndroidManifest.xml
```diff
@@ -2,7 +2,9 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
 
-    <uses-sdk tools:overrideLibrary="com.google.android.gms.threadnetwork" />
+    <uses-sdk tools:overrideLibrary="com.google.android.gms.threadnetwork,androidx.car.app.projected" />
+    <uses-permission android:name="com.google.android.gms.permission.CAR_FUEL" />
+    <uses-permission android:name="com.google.android.gms.permission.CAR_MILEAGE" />
```
dshokouhi, line 7: These should be moved over to the `main` manifest now :)

dshokouhi:
> For automotive support we need to ask for different permissions, is there a way to know if we are running the automotive version of the app ?

yes you can determine that using a check like below (minus the build config check which checks to see if its the play store version 😃 )

https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt#L86

drosoCode:
> > For automotive support we need to ask for different permissions, is there a way to know if we are running the automotive version of the app ?
> 
> yes you can determine that using a check like below (minus the build config check which checks to see if its the play store version smiley )
> 
> https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt#L86

I'm not sure on how to make it work in the `requiredPermissions` function, as we don't have access to the context (to get the packageManager)

app/src/full/AndroidManifest.xml
jpelgrom: Only the SDK override, permissions here are GMS permissions so only applicable to the full flavor

dshokouhi: > Only the SDK override, permissions here are GMS permissions so only applicable to the full flavor

they seem to be used for a few sensors though i thought

common/src/main/res/values/strings.xml
```diff
@@ -1098,4 +1109,5 @@
     <string name="ha_assist">HA: Assist</string>
     <string name="only_favorites">Only Show Favorites</string>
     <string name="beacon_scanning">Beacon Monitor Scanning</string>
+    <string name="car_data_unavailable">Open Home Assistant app to enable CarInfo sensors</string>
```
jpelgrom, line 1112: This shouldn't be used as a state, use `unavailable` like in other places (or maybe even don't change the state, just keep them as they were?)

dshokouhi:
> > > For automotive support we need to ask for different permissions, is there a way to know if we are running the automotive version of the app ?
> > 
> > 
> > yes you can determine that using a check like below (minus the build config check which checks to see if its the play store version smiley )
> > https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/launch/LaunchActivity.kt#L86
> 
> I'm not sure on how to make it work in the `requiredPermissions` function, as we don't have access to the context (to get the packageManager)

yea that might require a refactor of that method to add context

common/src/main/res/values/strings.xml
drosoCode: This was added to explicitly mention that the app needs to be started (https://github.com/home-assistant/android/pull/3399#issuecomment-1501993900), maybe we should keep it only for text sensors ?

jpelgrom: Missed that comment, sorry. The note could be more general/less technical, something like: Open Home Assistant app to activate this sensor.

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -0,0 +1,296 @@
...
+            Listener.STATUS to -1L,
+            Listener.MILEAGE to -1L
+        )
+    }
+
+    override val name: Int
+        get() = R.string.sensor_name_car
+
+    override suspend fun getAvailableSensors(context: Context): List<SensorManager.BasicSensor> {
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
dshokouhi, line 111: According to the [docs](https://developer.android.com/training/cars/apps#carinfo) you linked earlier, we need to hide the odometer sensor from automotive builds.

image

In fact if you don't want to refactor `requiredPermissions` in this PR (since its already getting pretty big), we can hide these sensors from the automotive build for now.  That may be better too for now as we have some sensors in that build that need to be hidden anyways.

dshokouhi:
Since the user needs to be on Android 8+ we should probably also use `hasSensor()` to hide the entire section in Manage Sensors screen from those devices. Just realized the android auto sensor does not do that but we can just do it here and fix that one in a separate PR :)

https://github.com/home-assistant/android/blob/a2f85ef47529dcdcb6aec72965c1236c3fe85b5f/common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorManager.kt#L137

app/src/full/AndroidManifest.xml
googlvalenzuela: permissions with GMS substring are related to the phone, for the car  you will have things like 
android.car.permission.CAR_SPEED are for cars.

app/src/main/AndroidManifest.xml
```diff
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
 
-    <uses-sdk tools:overrideLibrary="androidx.wear.remote.interactions" />
+    <uses-sdk tools:overrideLibrary="androidx.wear.remote.interactions,androidx.car.app.projected" />
```
googlvalenzuela, line 5: Does this breaks the compiling of the automotive build?

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -108,13 +109,23 @@ class CarSensorManager :
...
+                )
+        ) {
             sensorsList
         } else {
             emptyList()
         }
     }
 
+    override fun hasSensor(context: Context): Boolean {
+        return Build.VERSION.SDK_INT >= Build.VERSION_CODES.O
```
dshokouhi, line 126: We should add the automotive check here and I think we should also add a check to hide the sensors from the minimal version as well because those devices won't have the GMS permissions and thus wont be able to add any of the sensors.

drosoCode: The sensor for the car model doesn't require any special permissions, but I guess we can still limit this to full version since android auto probably doesn't work on minimal

dshokouhi: > The sensor for the car model doesn't require any special permissions,

my bad forgot about that one 😅 

> but I guess we can still limit this to full version since android auto probably doesn't work on minimal

I _think_ its safe to assume it wont work on the minimal build until someone tells us it would?

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -0,0 +1,306 @@
...
+        this.context = context.applicationContext
+        if (allDisabled()) {
+            return
+        }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            if (connected()) {
+                updateCarInfo()
+            } else {
+                connectList.forEach { (sensor, alreadySentMessage) ->
```
dshokouhi, line 157: I think the variable `alreadySentMessage` is safe to remove now?

dshokouhi:
@drosoCode appreciate your patience during this process 😃 Don't believe I have any more comments left 😅

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -0,0 +1,306 @@
...
+
+class CarSensorManager :
+    SensorManager,
+    DefaultLifecycleObserver {
+
+    companion object {
+        internal const val TAG = "CarSM"
+
+        private val fuelLevel = SensorManager.BasicSensor(
+            "car_fuel_level",
```
dshokouhi, line 28: Lets drop the `_level` as it does not add any value since there is already a unit and icon present. The phones battery level will remain as it has been there for a long time but for new sensors we are requested to remove it as it is not necessary. Please also update the strings to remove the word `Level` from the sensor name.

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -0,0 +1,306 @@
...
+            "sensor",
+            R.string.basic_sensor_name_car_fuel_level,
+            R.string.sensor_description_car_fuel_level,
+            "mdi:barrel",
+            unitOfMeasurement = "%",
+            stateClass = SensorManager.STATE_CLASS_MEASUREMENT,
+            deviceClass = "battery"
+        )
+        private val batteryLevel = SensorManager.BasicSensor(
+            "car_battery_level",
```
dshokouhi, line 38: Same as above please drop `_level` from the ID and `Level` from the sensor name in the string

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
drosoCode: I still use it to make sure that we call `onSensorUpdated` only once per sensor when the state is disconnected

drosoCode:
> @drosoCode appreciate your patience during this process smiley Don't believe I have any more comments left sweat_smile

I also appreciate the time taken to review this contribution :smiley:

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
jpelgrom: Other parts of the codebase will take care of not sending the sensor data again if you use a state that is the same as the previous one, so I think it can be removed?

drosoCode: Oh I didn't think that was the case, then yes we can remove this

dshokouhi (approved):
Looks good to me!

app/src/main/AndroidManifest.xml
jpelgrom: This library has a minSdk of 23, Automotive has a minSdk of 29 but the normal app 21 which is why this is required here but not for Automotive.

PR #3403, opened 2023-03-06
Update default entity icons

jpelgrom:
## Summary

We show entity icons in more and more places, so time to catch up with frontend changes for default entity icons :)

Notably, as we now have complications on Wear which can show any entity, I've also included:
 - Dynamic binary sensor icons
 - Dynamic sensor icons
 - Dynamic weather icons

## Screenshots

n/a, now better matches frontend

n/a


dshokouhi (approved):
Always fun expanding icons to be more dynamic like this :)

JBassett (approved):

PR #3410, opened 2023-03-08
Server trust feature

jpelgrom:
## Summary

This PR adds a feature that allows you to (dis)trust a server, or as it is named in the app choose whether the server can remotely control the app & device.

With multiserver you might add servers to the app for convenient access or push notifications, but you don't want them to receive and manage (all) your sensors and device like for your primary server. You can already for example disable sensors for a particular server, but what is the point if you can turn them back on remotely? This could expose data you don't want to.

Disabling this option (default is enabled to match current behavior) will stop the following features:
 - Use notification commands to manage device features and sensor settings
 - Enable/disable sensors from the server

As an intended side effect, when you don't trust a server and have disabled all sensors, the app will also stop fetching enabled sensors on the server in the worker (a request I've seen multiple times after it was added).

The goal is to have one toggle for all these 'remote control features'. A few users might want to fine tune the behavior but with the number of notification commands and other automatic integrations in the app this will quickly grow out of control.

## Screenshots

|Light|Dark|
|-----|-----|
|App setting with remote control icon titled 'Remotely control app & device', description 'Manage enabled sensors and use notification commands from this server', light mode|App setting with remote control icon titled 'Remotely control app & device', description 'Manage enabled sensors and use notification commands from this server', dark mode|

Documentation: home-assistant/companion.home-assistant#924

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -211,7 +211,6 @@ class MessagingManager @Inject constructor(
             COMMAND_LAUNCH_APP,
             COMMAND_APP_LOCK,
             COMMAND_PERSISTENT_CONNECTION,
-            TextToSpeechData.COMMAND_STOP_TTS,
             COMMAND_AUTO_SCREEN_BRIGHTNESS,
             COMMAND_SCREEN_BRIGHTNESS_LEVEL,
             COMMAND_SCREEN_OFF_TIMEOUT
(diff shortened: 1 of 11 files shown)


app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -294,271 +294,238 @@ class MessagingManager @Inject constructor(
...
-                            } else {
-                                mainScope.launch {
+        mainScope.launch {
+            val allowCommands = serverManager.integrationRepository(serverId).isTrusted()
+            when {
+                jsonData[NotificationData.MESSAGE] == REQUEST_LOCATION_UPDATE && allowCommands -> {
+                    Log.d(TAG, "Request location update")
+                    requestAccurateLocationUpdate()
+                }
+                jsonData[NotificationData.MESSAGE] == CLEAR_NOTIFICATION && !jsonData["tag"].isNullOrBlank() && allowCommands -> {
```
dshokouhi, line 304: Out of all the commands this one I feel should be left out as a lot of users make use of this to synchronize things like acknowledgement or to not notify if a certain situation is no longer occurring. 

If we were going to disallow this then we may consider also stopping replacing a notification.

wear/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -49,25 +50,29 @@ class MessagingManager @Inject constructor(
...
-            DeviceCommandData.COMMAND_BLE_TRANSMITTER -> {
-                if (!commandBleTransmitter(context, notificationData, sensorDao, mainScope)) {
-                    sendNotification(notificationData)
+                message == DeviceCommandData.COMMAND_BLE_TRANSMITTER && allowCommands -> {
+                    if (!commandBleTransmitter(context, notificationData, sensorDao, mainScope)) {
+                        sendNotification(notificationData)
+                    }
                 }
+                message == TextToSpeechData.TTS -> speakText(context, notificationData)
+                message == TextToSpeechData.COMMAND_STOP_TTS && allowCommands -> stopTTS()
```
dshokouhi, line 73: Looks like we are checking in Wear but not in the phone app. Not sure if its better to remove it or also stop TTS in these situations?

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
jpelgrom: I simply followed the list on https://companion.home-assistant.io/docs/notifications/notification-commands, but also agree that notification commands for notifications are probably fine. Where do we draw the line / how to communicate this clearly?

wear/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
jpelgrom: In the main app it's in the `DEVICE_COMMANDS` list which has the check.

dshokouhi: oops missed that :)

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
jpelgrom: So allow `clear_notification` (because you can update) and `remove_channel` (because you can create new channels with settings)?

dshokouhi: Yea `channel`s are not really removed either, more like a soft delete plus we don't block creation of them so why block clearing the list? Its not like they won't get recreated again 🤷‍♂️

jpelgrom: Excluded + opened docs PR to document this

JBassett (approved):

PR #3411, opened 2023-03-09
Properly handle multiple permissions requests

Suxsem:
## Summary

This PR starts while investigating the app crashing while trying to make a SIP call with the sip-hass-card (https://github.com/TECH7Fox/sip-hass-card/issues/96). After giving permissions to both microphone and camera, the app just crashes with the following log:
```
03-08 16:50:12.261  9201  9201 W System.err: java.lang.IllegalStateException: Either grant() or deny() has been already called.
03-08 16:50:12.262  9201  9201 W System.err:    at org.chromium.android_webview.permission.AwPermissionRequest.a(chromium-TrichromeWebViewGoogle.aab-stable-548115333:34)
03-08 16:50:12.262  9201  9201 W System.err:    at Wz0.grant(chromium-TrichromeWebViewGoogle.aab-stable-548115333:109)
03-08 16:50:12.262  9201  9201 W System.err:    at io.homeassistant.companion.android.webview.WebViewActivity$onCreate$2$4.onPermissionRequest(WebViewActivity.kt:491)
03-08 16:50:12.262  9201  9201 W System.err:    at org.chromium.android_webview.AwContents.onPermissionRequest(chromium-TrichromeWebViewGoogle.aab-stable-548115333:46)
03-08 16:50:12.262  9201  9201 W System.err:    at android.os.MessageQueue.nativePollOnce(Native Method)
03-08 16:50:12.262  9201  9201 W System.err:    at android.os.MessageQueue.next(MessageQueue.java:363)
03-08 16:50:12.262  9201  9201 W System.err:    at android.os.Looper.loop(Looper.java:176)
03-08 16:50:12.262  9201  9201 W System.err:    at android.app.ActivityThread.main(ActivityThread.java:8668)
03-08 16:50:12.262  9201  9201 W System.err:    at java.lang.reflect.Method.invoke(Native Method)
03-08 16:50:12.262  9201  9201 W System.err:    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:513)
03-08 16:50:12.262  9201  9201 W System.err:    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1109)
```
I soon discovered that there was a bug in the code: when multiple permissions are requested within the same request (as in my case camera and microphone) and at least two of them have been already granted by the user, the code calls `request.grant` twice (because it is inside the foreach loop). This is not allowed by the webview, thus the crash.

In order to fix it, I created two lists, the first one for the permissions already granted, the second one for the yet-ungranted ones. After the foreach loop populates the two lists, the `request.grant` is called once with all the granted permissions and then the `requestPermissions.launch` is called with all the yet-ungranted ones. For this last part, I have changed the `requestPermissions` member to use `RequestMultiplePermissions` instead of `RequestPermission`.

I thoroughly test the fix and all issues seems to be gone, without any regression on the existing code.

_No functionality has been impacted_

## Any other notes

While fixing the issue I also noticed that there were two _else if_ branches with the _same_ condition (`it == PermissionRequest.RESOURCE_AUDIO_CAPTURE`), so the second one was never executed; i deleted it

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -140,9 +140,11 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
     }
 
     private val ioScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
-    private val requestPermission =
-        registerForActivityResult(ActivityResultContracts.RequestPermission()) {
-            webView.reload()
+    private val requestPermissions =
+        registerForActivityResult(ActivityResultContracts.RequestMultiplePermissions()) {
+            if (it.any { it.value }) {
+                webView.reload()
+            }
         }
     private val requestStoragePermission =
         registerForActivityResult(ActivityResultContracts.RequestPermission()) { isGranted ->
@@ -499,16 +501,19 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
 
                 override fun onPermissionRequest(request: PermissionRequest?) {
                     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                        val alreadyGranted = ArrayList<String>()
+                        val toBeGranted = ArrayList<String>()
                         request?.resources?.forEach {
+
                             if (it == PermissionRequest.RESOURCE_VIDEO_CAPTURE) {
                                 if (ActivityCompat.checkSelfPermission(
                                         context,
                                         android.Manifest.permission.CAMERA
                                     ) == PackageManager.PERMISSION_GRANTED
                                 ) {
-                                    request.grant(arrayOf(it))
+                                    alreadyGranted.add(it)
                                 } else {
-                                    requestPermission.launch(android.Manifest.permission.CAMERA)
+                                    toBeGranted.add(android.Manifest.permission.CAMERA)
                                 }
                             } else if (it == PermissionRequest.RESOURCE_AUDIO_CAPTURE) {
                                 if (ActivityCompat.checkSelfPermission(
@@ -516,21 +521,19 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                                         android.Manifest.permission.RECORD_AUDIO
                                     ) == PackageManager.PERMISSION_GRANTED
                                 ) {
-                                    request.grant(arrayOf(it))
+                                    alreadyGranted.add(it)
                                 } else {
-                                    requestPermission.launch(
-                                        android.Manifest.permission.RECORD_AUDIO
-                                    )
-                                }
-                            } else if (it == PermissionRequest.RESOURCE_AUDIO_CAPTURE) {
-                                if (ActivityCompat.checkSelfPermission(
-                                        context,
-                                        android.Manifest.permission.MODIFY_AUDIO_SETTINGS
-                                    ) == PackageManager.PERMISSION_GRANTED
-                                ) {
-                                    request.grant(arrayOf(it))
+                                    toBeGranted.add(android.Manifest.permission.RECORD_AUDIO)
                                 }
                             }
+                        }
+                        if (alreadyGranted.size > 0) {
+                            request?.grant(alreadyGranted.toTypedArray())
+                        }
+                        if (toBeGranted.size > 0) {
+                            requestPermissions.launch(
+                                toBeGranted.toTypedArray()
+                            )
                         }
                     } else {
                         // If we are before M we already have permission, just grant it.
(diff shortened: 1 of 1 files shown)


Suxsem:
Ops, I made the commit with my other email address. I just added the other email address to my GitHub profile

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -499,39 +501,39 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
+                                    alreadyGranted.add(it)
                                 } else {
-                                    requestPermission.launch(
-                                        android.Manifest.permission.RECORD_AUDIO
-                                    )
-                                }
-                            } else if (it == PermissionRequest.RESOURCE_AUDIO_CAPTURE) {
-                                if (ActivityCompat.checkSelfPermission(
-                                        context,
-                                        android.Manifest.permission.MODIFY_AUDIO_SETTINGS
```
dshokouhi, line 528: This actual permission is separate from the one above, I think we still need to keep it as the actual permission check is different. All 3 permissions are required as they were previously tested by users in the past. Different devices have different purposes.

Suxsem: Dear @dshokouhi,
thank you for reviewing this PR.

As you can see from the original code, the _else if_ branch that you referenced was never reachable because the condition was exactly the same as the one before it:

line 513:
`} else if (it == PermissionRequest.RESOURCE_AUDIO_CAPTURE) {`

line 525:
`} else if (it == PermissionRequest.RESOURCE_AUDIO_CAPTURE) {`

Because these are _else if_ and not multiple _if_s, the second branch was completely unused.

What do you think?

dshokouhi: Ah that is a good catch indeed! In this case instead of removing that specific permission lets fix that bug here too if possible. Thank you for pointing that out.

Suxsem: @dshokouhi you are welcome!
I checked Android documentation about PermissionRequest (https://developer.android.com/reference/android/webkit/PermissionRequest#constants_1) and only possible values are:

- RESOURCE_AUDIO_CAPTURE (already handled)
- RESOURCE_VIDEO_CAPTURE (already handled)
- RESOURCE_MIDI_SYSEX (unrelated, sysex messages from MIDI devices)
- RESOURCE_PROTECTED_MEDIA_ID (unrelated, needed to generate licenses for protected media)

So I don't see any use case that could need the `android.Manifest.permission.MODIFY_AUDIO_SETTINGS` permission.
I could add it to the `RESOURCE_AUDIO_CAPTURE` case but I don't think this is needed and it could result in asking unnecessary permissions to the user.

As it was never used in the past I suggest to just remove it, maybe it was a leftover from tests during development, but if you prefer to add it to the `RESOURCE_AUDIO_CAPTURE` case I will edit the PR.

Thank you

dshokouhi: Thats strange. Looking at the original PR #1850 I can see that this code was always there but looking at the request the user pointed out the current version of the app (at the time of the request) did not work for them and the debug APK which contained that specific permission was there and that it was now working for their setup. I think we should still keep it there especially considering the error message in the request that calls out this specific permission. I don't believe this permission is auto granted once added to the manifest so I am confused as to how that PR fixed the users issue 🤔  Personally speaking I don't think it hurts to ask for more permissions especially if something is going to require it. The main thing to consider here is that while it may have been broken for some devices or users it wasn't broken for all.

Suxsem: @dshokouhi I think that what fixed the issue of the user was adding the `<uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />` to the AndroidManifest.xml.
Do you agree that this code was never reached due to the same else if condition?

dshokouhi: Yea that was probably it, that permission may have been granted alongside record audio.

Sorry for the confusion had to look up the original PR since I remembered the request 😅

Suxsem: Ok I can confirm this. Take a look at Android permission list (https://github.com/aosp-mirror/platform_frameworks_base/blob/master/core/res/AndroidManifest.xml).
`android.permission.MODIFY_AUDIO_SETTINGS` has `android:protectionLevel="normal"` so it doesn't need approval from the user, just the declaration in the manifest.
(as opposed to `android.permission.RECORD_AUDIO` that has `android:protectionLevel="dangerous|instant"`)
So what fixed the issue of the user was adding the permission to the manifest. It can be safely removed from the code

dshokouhi: Thank you for clarifying that :) I also see that Android Studio properly marks that condition to notify users of that mistake 😓

Suxsem: You are right, didn't noticed! 😆

dshokouhi (approved):
Code looks good to me, I don't have a way to test this as I do no have a device that meets the needs here but as the author found the crash and verified its fixed that is good enough for me.

Thank you for the PR 🙏

JBassett (approved):

PR #3423, opened 2023-03-16
Android Automotive Module

googlvalenzuela:
## Summary
Creation of Android Automotive Module in order to compile and deploy home assistant into Android Automotive following 
https://developer.android.com/training/cars/apps/automotive-os

Update Base Activity to be able to transition from Full activity to Car App library.

No real changes on the app looks / interactions, this will launch Android Automotive Car App Activity as the launcher and its following the same execution path that is today for Android Auto.

Android Automotive IOT documentation: https://developer.android.com/training/cars/apps/automotive-os

## Any other notes

Home_Assistant_Start

Home_Assistant_Not_Logged_In

Home_Assistant_Dashboard

Home_Assistnat_Cal

Home_Assistant_Car_App_Library_Covers

### DIFF
--- a/automotive/build.gradle.kts
+++ b/automotive/build.gradle.kts
@@ -180,15 +180,15 @@ dependencies {
     implementation("com.squareup.okhttp3:okhttp:4.10.0")
     implementation("com.squareup.picasso:picasso:2.8")
 
-    "fullImplementation"("com.google.android.gms:play-services-location:21.0.1")
-    "fullImplementation"("com.google.android.gms:play-services-home:16.0.0")
-    "fullImplementation"("com.google.android.gms:play-services-threadnetwork:16.0.0-beta02")
-    "fullImplementation"(platform("com.google.firebase:firebase-bom:31.2.2"))
-    "fullImplementation"("com.google.firebase:firebase-messaging")
-    "fullImplementation"("io.sentry:sentry-android:6.14.0")
-    "fullImplementation"("org.jetbrains.kotlinx:kotlinx-coroutines-play-services:1.6.4")
-    "fullImplementation"("com.google.android.gms:play-services-wearable:18.0.0")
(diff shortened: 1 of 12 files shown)


automotive/build.gradle.kts
```diff
@@ -0,0 +1,234 @@
...
+    useLibrary("android.car")
+
+    defaultConfig {
+        applicationId = "io.homeassistant.companion.android"
+        minSdk = 29
+        targetSdk = 33
+
+        versionName = System.getenv("VERSION") ?: "LOCAL"
+        // We add 2 because the app, wear (+1) and automotive versions need to have different version codes.
+        versionCode = (System.getenv("VERSION_CODE")?.toIntOrNull() ?: 1) + 3
```
JBassett, line 32: So we need the `VERSION_CODE` variable increased by 3 in the pipeline too.  Otherwise we are going to conflict.

automotive/build.gradle.kts
```diff
@@ -0,0 +1,234 @@
...
+        correctErrorTypes = true
+    }
+}
+
+play {
+    serviceAccountCredentials.set(file("playStorePublishServiceCredentialsFile.json"))
+    track.set("internal")
+    resolutionStrategy.set(ResolutionStrategy.IGNORE)
+    // We will depend on the wear commit.
+    commit.set(true)
```
JBassett, line 151: This should be false since the app release will do the actual commit.

JBassett (changes requested):

JBassett:
Another thing, in the `beta.yml` and `release.yml` you need to add the output of this module to be added to the pre-release that is generated.

automotive/build.gradle.kts
dvalenzuela: Done

dvalenzuela:
@JBassett let me know if there is anything else I can do to help

PR #3436, opened 2023-03-25
New intent extra types

pascal-t:
## Summary
Based on https://github.com/home-assistant/android/pull/2207 I added more types to the intent_extras in the `MessagingManager`.
I also added types for String Arrays and ArrayLists, that consist of urlencoded Strings. These are called `String[].urlencoded` and `ArrayList<String>.urlencoded`. The existing documentation for the urlencoded string reads:

> Special characters in extras are supported by urlencoding the extra value and appending :urlencoded to the end.

One could argue that for consistency, `String[]:urlencoded` and `ArrayList<String>:urlencoded` would be better names, hovever I decided against it, because my way was easyer to implement.

## Screenshots
n.a.

Documentation: home-assistant/companion.home-assistant#931


jpelgrom:
> One could argue that for consistency, `String[]:urlencoded` and `ArrayList<String>:urlencoded` would be better names, hovever I decided against it, because my way was easyer to implement.

It might be easier to implement but two notations to use url encoded strings is inconsistent and confusing for the user. There are more users than developers ;)

pascal-t:
I thought about it for a bit. If it was implemented that way it would make sense to also have  `String:urlencoded` as an option. That would make `EXTRA:String:urlencoded` unambiguous and more complex to parse.

Instead if we added `String.urlencoded` as an option we could keep `urlencoded` as a shorthand (and for backwards compatibility of course). The wording in the documentation could be changed accordingly.

jpelgrom:
Sorry for the late response. I think if something like `String.urlencoded` is added, and `urlencoded` only exists for backwards compatibility purposes, that would resolve the potential for confusion I mentioned.

pascal-t:
I also didn't get around to implement the changes last week. Thanks for working on my pull request. I actually tried to implement the `String:urlencoded` option, but I didn't like it.
I added the new types `String` and `String.urlencoded` and updated the documentation accordingly.

jpelgrom (approved):
Thanks for adding missing types and completing this feature

JBassett (approved):

PR #3449, opened 2023-04-02
Add Android OS sensors

frimtec:
## Summary
New sensors for Android OS build information.

## Screenshots
light
dark

Documentation: home-assistant/companion.home-assistant#934

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/AndroidOsSensorManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/AndroidOsSensorManager.kt
@@ -1,0 +1,80 @@
+package io.homeassistant.companion.android.common.sensors
+
+import android.content.Context
+import android.os.Build
+import io.homeassistant.companion.android.common.R as commonR
+
+class AndroidOsSensorManager : SensorManager {
+
+    companion object {
+        val osVersion = SensorManager.BasicSensor(
+            "android_os_version",
+            "sensor",
+            commonR.string.basic_sensor_name_android_os_version,
+            commonR.string.sensor_description_android_os_version,
+            "mdi:android",
+            entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC
+        )
+        val osSecurityPatch = SensorManager.BasicSensor(
+            "android_os_security_patch",
+            "sensor",
+            commonR.string.basic_sensor_name_android_os_security_patch,
+            commonR.string.sensor_description_android_os_security_patch,
+            "mdi:security",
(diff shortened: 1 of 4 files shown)


common/src/main/java/io/homeassistant/companion/android/common/sensors/AndroidOsSensorManager.kt
```diff
@@ -0,0 +1,80 @@
...
+
+    override fun docsLink(): String {
+        return "https://companion.home-assistant.io/docs/core/sensors#android-os-sensors"
+    }
+
+    override val name: Int
+        get() = commonR.string.sensor_name_android_os
+
+    override suspend fun getAvailableSensors(context: Context): List<SensorManager.BasicSensor> {
+        return listOf(osVersion, osSecurityPatch)
```
dshokouhi, line 36: Given that the security patch can only be read on devices running Android M or higher we should make sure to return the correct list to users based on their Android version. This will prevent a user on Android 5 trying to enable a sensor that will never return any data :)

common/src/main/java/io/homeassistant/companion/android/common/sensors/AndroidOsSensorManager.kt
```diff
@@ -0,0 +1,80 @@
...
+    ) {
+        checkState(context, osVersion)
+        checkState(context, osSecurityPatch)
+    }
+
+    private fun checkState(
+        context: Context,
+        sensor: SensorManager.BasicSensor
+    ) {
+        if (!isEnabled(context, sensor) || !hasSensor(context)) {
```
dshokouhi, line 58: We only need to check if the sensor is enabled here.

common/src/main/java/io/homeassistant/companion/android/common/sensors/AndroidOsSensorManager.kt
```diff
@@ -0,0 +1,80 @@
...
+        return listOf(osVersion, osSecurityPatch)
+    }
+
+    override fun requiredPermissions(sensorId: String): Array<String> {
+        return arrayOf()
+    }
+
+    override fun hasSensor(context: Context): Boolean {
+        return true
+    }
```
dshokouhi, line 45: this is actually the default value, we only need to override it if the value will be `false` :)

https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorManager.kt#L138

dshokouhi:
Given the general nature of these sensors we should probably also add them to Wear OS :)

frimtec:
Many thanks for your valuable review @dshokouhi 👍🏻 
I have adapted all the points you mentioned.

JBassett (approved):

PR #3454, opened 2023-04-03
Disable favorites only view on Wear if all favorites are deleted

jpelgrom:
## Summary

Fixes the 'favorites only' setting staying enabled on Wear watches after removing all favorites, which blocked the user from turning the setting off.

## Screenshots

n/a

n/a


dshokouhi (approved):
Good catch :)

JBassett (approved):

PR #3502, opened 2023-04-29
Set Loading State for Entities

googlvalenzuela:
## Summary
Declare executing state of an entity based on verb states.
This allows Android Auto scenarios to show that the entity is loading and provid
`
fun <T> Entity<T>.isExecuting()=  state.endsWith("ing")
`

## Screenshots
Screenshot 2023-04-29 at 11 07 35 AM
e feedback for the user about the state on which the entity is in.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
@@ -650,5 +650,6 @@ fun <T> Entity<T>.isExecuting() = when (state) {
     "unlocking" -> true
     "buffering" -> true
     "disarming" -> true
+    "arming" -> true
     else -> false
 }
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
@@ -643,4 +643,4 @@ fun <T> Entity<T>.friendlyState(context: Context): String {
     return friendlyState
 }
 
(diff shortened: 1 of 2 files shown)


marazmarci:
> States is an open system and can be any value. While I think this works for the currently supported domains, to prevent showing the 'loading' indicator unintentionally IMO the `isExecuting` function should check against a list of known executing states instead to be safe.

a list of known states can be found here I think: https://github.com/home-assistant/core/blob/ef9bcd9285980385eede726e15aa1e7d20238eee/homeassistant/const.py#L330

JBassett (changes requested):
I agree, just checking if the state ends in ing is too broad.

googlvalenzuela:
Thanks for the feedback, Ive updated the code.

common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -642,3 +642,13 @@ fun <T> Entity<T>.friendlyState(context: Context): String {
...
     return friendlyState
 }
+
+fun <T> Entity<T>.isExecuting() = when (state) {
+    "closing" -> true
+    "locking" -> true
+    "opening" -> true
+    "unlocking" -> true
+    "buffering" -> true
+    "disarming" -> true
```
jpelgrom, line 652: If you include `disarming`, also add `arming`

googlvalenzuela:
Added Armig.

jpelgrom (approved):

JBassett (approved):

PR #3560, opened 2023-05-30
Align dependency versions and bump Google dependencies

jpelgrom:
## Summary

This PR:
- aligns dependency versions between projects (for example, the app was using two different Kotlin patch releases) / modules (`automotive` was behind on dependencies already increased in `app`), and
- updates Google dependencies that are not picked up by dependabot.

Noteworthy changes:
https://developer.android.com/jetpack/androidx/releases/compose-compiler support for Kotlin 1.8.21
https://developer.android.com/jetpack/androidx/releases/compose-runtime#1.4.3 fixes a "potential deadlock in snapshot list and map" which may be what I was encountering with the ANR (see below)
https://firebase.google.com/support/release-notes/android no cloud messaging changes

These changes also seem to resolve a frequent ANR after load I'm facing with the latest beta for the Wear app and wasn't able to figure out. Not server related probably as it also happens with <10 supported entities and no state changes.

## Screenshots

n/a

n/a

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -46,7 +46,7 @@ android {
     }
 
     composeOptions {
-        kotlinCompilerExtensionVersion = "1.4.3"
+        kotlinCompilerExtensionVersion = "1.4.7"
     }
 
     kotlinOptions {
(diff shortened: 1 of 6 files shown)


app/build.gradle.kts
```diff
@@ -147,10 +147,10 @@ dependencies {
     implementation("com.maltaisn:iconpack-community-material:5.3.45")
 
     implementation("org.jetbrains.kotlin:kotlin-stdlib:1.8.21")
-    implementation("org.jetbrains.kotlin:kotlin-reflect:1.8.10")
+    implementation("org.jetbrains.kotlin:kotlin-reflect:1.8.21")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.1")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.7.1")
-    "fullImplementation"("org.jetbrains.kotlinx:kotlinx-coroutines-play-services:1.6.4")
+    "fullImplementation"("org.jetbrains.kotlinx:kotlinx-coroutines-play-services:1.7.1")
```
dshokouhi, line 153: isnt this duplicated by line 176?

jpelgrom: Yes but removing it breaks the automotive build because it doesn't do full/minimal properly, see the open PR for that 😅. I decided to just keep it for now as we'll likely pick it up when reconfiguring automotive for full/minimal?

dshokouhi: ooh I see what you mean now :)

dshokouhi (approved):
🤞 this resolves the ANR in release ready builds

JBassett (approved):

PR #3561, opened 2023-05-30
Speed up cold start by skipping Jackson ObjectMapper if possible

jpelgrom:
## Summary

While doing performance tracing on a Pixel 4a, I noticed that when launching the app it takes on average 75-100ms to read server data. Almost all of that time is spent getting a Jackson `ObjectMapper` to convert the database value for internal SSIDs to a list...
Performance trace showing 133 ms spent on fromStringToList function

... which if empty `"[]"` is a huge waste of time. So this PR adds a simple string comparison before getting Jackson to speed up startup. This performance improvement is even nicer on watches which are usually slower and will **never** have SSIDs.

It might be easier to compare the `setContent` function in the trace, reading from the database is now negligible:
Performance trace showing 4 ms spent on reading from the database. The setContent function is about 1/3rd of all time spent on launching the activity, whereas before it was closer to 1/10th of all time spent launching.

Jackson will be used later for network but at that point the app isn't waiting on it to continue launching.

Modern phones don't take quite as long to get Jackson but it still shaves ~20ms off launch (Pixel 7a).

## Screenshots

n/a

n/a

## Any other notes

Using Android Studio's "low overhead" profiling, averages of 5-6 launches.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
@@ -106,19 +106,27 @@ data class ServerConnectionInfo(
 class InternalSsidTypeConverter {
     @TypeConverter
     fun fromStringToList(value: String): List<String> {
-        return try {
-            jacksonObjectMapper().readValue(value)
-        } catch (e: JsonProcessingException) {
+        return if (value == "[]" || value.isBlank()) {
             emptyList()
+        } else {
+            try {
+                jacksonObjectMapper().readValue(value)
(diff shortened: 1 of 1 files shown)


JBassett (approved):
This is fine for a temp workaround, but we should look into creating a single instance of the object mapper and injecting it where needed.

jpelgrom:
No object mapper is still faster than a cached object mapper for empty lists? The function itself already does some caching, I'm specifically calling out the first use here on a critical path.

JBassett:
I'm not saying we remove this optimization, just that we'll need that mapper not long after that block gets executed.

PR #3570, opened 2023-06-06
Move vehicle service to main to allow for minimal Automotive build

jpelgrom:
## Summary

Untangling the Automotive build so we can do a build for devices without GAS as well...

 - Move the vehicle package from full to main so it is included in the minimal build as well, to allow for a minimal Automotive build
 - Adjust dependencies from full to any version (checked: the `androidx.car` dependencies do not have a dependency on any GMS libraries)
 - Move Automotive `CarAppActivity` + service + metadata to main manifest; not moved in app module because it is only used for Android Auto there
 - Adjust GitHub Actions to include minimal Automotive APK

Closes #3545: full only dependency now works, deduplication included in this PR.

## Screenshots

Minimal automotive build running on an Android Automotive emulator:
Settings screen on an Android Automotive emulator with "LOCAL-minimal" version visible

n/a


dshokouhi (approved):
Nice work!

JBassett (approved):

PR #3572, opened 2023-06-07
Set Wear theme background color to black

jpelgrom:
## Summary

I noticed that the background for Assist on watches was still dark grey instead of black. It's black in most other activities because of Compose's `SwipeToDismissBox` adding it for us, or because we manually added it.

This PR sets the theme default to black to match the [app quality guidelines](https://developer.android.com/docs/quality-guidelines/wear-app-quality#upcoming) + removes activity/view specific backgrounds that are black and no longer needed.

## Screenshots

|Before|After|
|-----|-----|
|Assist activity with a dark grey background|Assist activity with a black background|

n/a

## Any other notes

> Visual quality | Black background | `WO-V13` | Use a black background for all apps and tiles.


dshokouhi (approved):
Good catch!

JBassett (approved):

PR #3510, opened 2023-05-02
Use additional http headers to bypass Cloudflare SSO. (Or other SSO)

Meister1977:
## Summary

Added option to define 2 additional http headers to use Cloudflare Zero Trust with service token. (Or other SSO systems).
https://github.com/home-assistant/android/issues/2650

## Any other notes

Please feel free to decline the pull request, I am not a kotlin developer, just added this feature, because I needed. If you can use any part of this commit to add this feature in a better way, just do it.

Issue #2650, opened 2022-06-30
Support Cloudflared Zero Trust protected instance from App


kingamajick:
**Is your feature request related to a problem? Please describe.**
I use Cloudflared Zero Trust to protect my Home Assistance. To access my Home Assistant instance, I have to log in using oAuth. When I do this via the Home Assistant app, the process ends in Chrome rather than the Home Assistant App. 

Actual Results:
1. Open HA App
2. App opens Chrome to login to Zero Trust
3. After login, HA is shown in Chrome,

**Describe the solution you'd like**
1. Open HA App
2. Zero Trust login shown in HA App
3. After login, HA is shown in HA App

**Describe alternatives you've considered, if any**
Zero Trust also supports [Service Tokens](https://developers.cloudflare.com/cloudflare-one/identity/service-tokens), an alternative could be to allow custom headers to be attached to requests (this could potentially allow for a solution to other providers).

Another alternative is to use warp for login, buy this isn't feasible on my corporate phone.

**Additional context**

marazmarci:
I like this idea! 👍

However, a problem came to my mind:
A Cloudflare Zero Trust session can last 1 month at most, see the documentation [here](https://developers.cloudflare.com/cloudflare-one/identity/users/session-management/).
When the session expires, suddenly the app will start to silently fail, and you won't get push notifications. That could be unpleasant at times when you want to be notified about some important smart home event.

jpelgrom:
Isn't this the same as #1438, just with a different service? Or am I missing something?

> When the session expires, suddenly the app will start to silently fail, and you won't get push notifications.

This won't be a problem for most users because push notifications are delivered using Firebase.

kingamajick:
Seems I searched for `Cloud flare` rather than `Cloudflare` when looking for previous issues :facepalm:

momelod:
I have the same setup and use Cloudflare Zerotrust to protect my home-assistant.  I got this to work by only protecting the auth subfolder `auth/authorize`. I've been using it for some time and everything has been working flawlessly until today when I tried logging in from a new device. On the new device I see the same aforementioned useragent error, meanwhile my old device continues working with out issue.

momelod:
After disabling ZeroTrust and authenticating on the new device it works. After the device has been authenticated I can re-enable ZeroTrust and everything continues to work just like it has on my original device. 

Hope that helps

ported-pw:
[...]

PR #3593, opened 2023-06-17
Add custom header support

EricMedina024:
## Summary

Adds support for adding a list of headers to add to requests made to a HA server. I tried my best to find all the places the headers need to be added, but I'm sure I missed some. Please let me know any additional places I need to address.

This pull request is a work in progress. The main functionality works, but the UI for adding/removing headers is pretty wonky functionally still.

## Screenshots

image
image

image
image

None yet

## Any other notes
Thanks to @Meister1977 for his PR #3510 as it helped me figure out where to start digging. I saw his pull request was tagged help wanted, but our implementations ended up being so different I thought it made more sense to make my own PR.

I have little to no experience with Android dev and Kotlin, so please feel free to point out where I can improve the code!

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/server/ServerSettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/server/ServerSettingsFragment.kt
@@ -31,6 +31,7 @@ import io.homeassistant.companion.android.launch.LaunchActivity
 import io.homeassistant.companion.android.settings.SettingsActivity
 import io.homeassistant.companion.android.settings.ssid.SsidFragment
 import io.homeassistant.companion.android.settings.url.ExternalUrlFragment
+import io.homeassistant.companion.android.settings.url.HeadersFragment
 import io.homeassistant.companion.android.settings.websocket.WebsocketSettingFragment
 import io.homeassistant.companion.android.webview.WebViewActivity
 import kotlinx.coroutines.launch
@@ -149,6 +150,18 @@ class ServerSettingsFragment : ServerSettingsView, PreferenceFragmentCompat() {
                     Bundle().apply { putInt(ExternalUrlFragment.EXTRA_SERVER, serverId) }
                 )
                 addToBackStack(getString(commonR.string.input_url))
+            }
+            return@setOnPreferenceClickListener true
+        }
+
+        findPreference<Preference>("connection_headers")?.setOnPreferenceClickListener {
+            parentFragmentManager.commit {
+                replace(
+                    R.id.content,
+                    HeadersFragment::class.java,
+                    Bundle().apply { putInt(HeadersFragment.EXTRA_SERVER, serverId) }
+                )
+                addToBackStack(getString(commonR.string.pref_headers))
             }
             return@setOnPreferenceClickListener true
         }
--- a/app/src/main/java/io/homeassistant/companion/android/settings/url/HeadersFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/url/HeadersFragment.kt
@@ -1,0 +1,51 @@
+package io.homeassistant.companion.android.settings.url
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.compose.ui.platform.ComposeView
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.viewModels
+import com.google.accompanist.themeadapter.material.MdcTheme
+import dagger.hilt.android.AndroidEntryPoint
+import io.homeassistant.companion.android.common.R
+import io.homeassistant.companion.android.settings.url.views.HeadersView
+
+@AndroidEntryPoint
+class HeadersFragment : Fragment() {
+
+    companion object {
+        const val EXTRA_SERVER = "server"
+    }
+
+    val viewModel by viewModels<HeadersViewModel>()
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setHasOptionsMenu(false)
+    }
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View {
+        if (viewModel.headers.size == 0) {
+            viewModel.headers[""] = ""
+        }
+        return ComposeView(requireContext()).apply {
+            setContent {
+                MdcTheme {
+                    HeadersView( headersViewModel = viewModel )
+                }
+            }
+        }
+    }
+
+    override fun onResume() {
+        super.onResume()
+        activity?.title = getString(R.string.pref_headers)
+    }
+
+}
--- a/app/src/main/java/io/homeassistant/companion/android/settings/url/HeadersViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/url/HeadersViewModel.kt
@@ -1,0 +1,66 @@
+package io.homeassistant.companion.android.settings.url
+
+import android.app.Application
+import androidx.compose.runtime.mutableStateMapOf
+import androidx.compose.runtime.snapshots.SnapshotStateMap
+import androidx.lifecycle.AndroidViewModel
+import androidx.lifecycle.SavedStateHandle
+import androidx.lifecycle.viewModelScope
+import dagger.hilt.android.lifecycle.HiltViewModel
+import io.homeassistant.companion.android.common.data.servers.ServerManager
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+@HiltViewModel
+class HeadersViewModel @Inject constructor(
+    state: SavedStateHandle,
+    private val serverManager: ServerManager,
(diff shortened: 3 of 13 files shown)


marazmarci:
Please note that it's only possible to add custom headers to the `OkHttpClient`'s requests, but not to the `WebView`'s requests!

EricMedina024:
> I haven't tested yet, but I'm thinking (hoping) it's possible to just add the headers there.

It seems this is not possible. I modified my code to add the headers to the `WebResourceRequest`, then set up burp suite as my emulator's proxy and watched the traffic. Unfortunately, resource requests still did not have the custom headers. That said, they do still succeed as the cloudflare cookies have already been set at this point. Apparently, if you return null any modifications made to the `WebResourceRequest` are ignored. With this limitation in mind, adding full custom header support is likely either impossible or too hacky to be worth it.

What I can do is take this PR in a more Cloudflare Zero Trust specific direction. Basically, I'd remove the "custom header" wording and replace it with options for the service account id and secret. Since we can ensure that the headers are present the first time the WebView loads, this would work well for that. @jpelgrom I see you were active in this thread - is there interest in merging this if it's Cloudflare Zero Trust specific? I'm not sure what the HA stance might be on such a company specific feature.

marazmarci:
> HA already overrides `shouldOverrideUrlLoading` [here](https://github.com/home-assistant/android/blob/7ff1423c850b7bf71e800cf2e546e4bb7748ce43/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt#L399), and the `WebResourceRequest` arg has a `headers` field which is mutable. I haven't tested yet, but I'm thinking (hoping) it's possible to just add the headers there.

Here's the key part from my [comment](https://github.com/home-assistant/android/issues/2650#issuecomment-1533732560) on #2650 that I linked above:

> Technical details:
See [`WebViewClient::shouldInterceptRequest`](https://developer.android.com/reference/android/webkit/WebViewClient#shouldInterceptRequest(android.webkit.WebView,%20android.webkit.WebResourceRequest)). If you want to modify the `WebView`'s HTTP requests, you have to make the HTTP request yourself from the application code, with eg. using an `OkHttpClient`, **_but the `WebViewClient` API won't give you the request body, which makes it practically useless for `POST` requests, because those almost always have a body._** Not to mention WebSocket requests, as they're 100% not interceptable, and there's currently no Android API for them.

What this means, is that `WebView` practically doesn't support setting custom headers. However, you can set the `User-Agent` and the cookies, that's all.

And also, mTLS client certificates can be used, which is a better (more secure) alternative for Service Tokens.

EricMedina024:
> and Cloudflare's free plan includes support for client certificates

do you have a link to the docs on this? I was looking into it and it seems it is only available on the enterprise plan for zero trust. You can add a firewall rule for it, but that won't help if you want the website available with other access configurations as well.

EricMedina024:
Thanks for the reply. I agree that it is not worth moving forward with this feature.

> Like previously mentioned, client certificates are a good alternative and also available for free with Cloudflare :) [Documentation basic](https://developers.cloudflare.com/ssl/client-certificates/) or [documentation zero trust](https://developers.cloudflare.com/cloudflare-one/identity/devices/access-integrations/mutual-tls-authentication/) or for basic go to your domain > SSL/TLS > Client Certificates, there are two buttons there to create a certificate and create a rule which is all you need.

This is actually a different offering than the Zero Trust mTLS stuff. It allows most of the same behavior, but the Zero Trust version is different as you can configure it more flexibly as part of your access policies instead of using a WAF rule. With a WAF rule, you are forced to either deny or allow the traffic and can't offer an alternative auth scheme when the mTLS cert is missing. That said, you guys gave some great alternatives here so I ended up coming up with a scheme that works for me. Here is what I ended up configured in case it is useful to anyone else.

I have three subdomains (which are all just zero trust tunnels to my HA instance) set up as follows:
- ha.ericmedina024.com

This is what I would use if I wanted to access my HA instance from a device I don't own. It is configured via access policies to require access code auth for all requests.

- ha-mtls.ericmedina024.com

This is what I use for the app as I have the mTLS cert installed on my phone (under "VPN & app user cert" in settings). It is configured via a WAF rule to block any traffic that doesn't present a valid mTLS cert.

- ha-ga.ericmedina024.com

This is what I use to allow the Google Assistant integration to work. In my google action, I have the fulfillment URL set to `https://ha-ga.ericmedina024.com/api/google_assistant?key=xyz` where `xyz` is a secure "password". I then have a WAF rule set up to ensure that the only accessible path on this subdomain is `/api/google_assistant` and that the `key` query parameter is set correctly.

jeleniain:
>     * ha-mtls.ericmedina024.com
> 
> 
> This is what I use for the app as I have the mTLS cert installed on my phone (under "VPN & app user cert" in settings). It is configured via a WAF rule to block any traffic that doesn't present a valid mTLS cert.
> 
Thanks a bunch for the comment!! I was lurking in these threads for a while trying to find a solution. I am currently a bit stuck on the Cloudflare side and was hoping you could provide some more details on how to configure this. I tried around in Cloudflare and tried to use the Cloudflare documentation, chatgpt and youtube but am stuck :)

Within: SSL/ TLS > Client Certificates, here I have created an mTLS Certificate for a Host that is a subdomain but I am uncertain on what to do from here. I have looked at the "WAF > Custom rules" but not sure what the correct setting would be. Ideally I would like to block anything that does not come from e.g. Germany and can not provide the certificate. But I am also uncertain if I can than just create a cloudflared tunnel for that subdomain and how that goes together with the Zero Trust Applications and Access Groups that I have in place.

Any help would make my day.
Thanks!

EricMedina024:
> > ```
> > * ha-mtls.ericmedina024.com
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is what I use for the app as I have the mTLS cert installed on my phone (under "VPN & app user cert" in settings). It is configured via a WAF rule to block any traffic that doesn't present a valid mTLS cert.
> 
> Thanks a bunch for the comment!! I was lurking in these threads for a while trying to find a solution. I am currently a bit stuck on the Cloudflare side and was hoping you could provide some more details on how to configure this. I tried around in Cloudflare and tried to use the Cloudflare documentation, chatgpt and youtube but am stuck :)
> 
> Any help would make my day. Thanks!
> 
> Edit: I think I am bit closer to a solution, maybe it helps someone or if you can see if I made a mistake:
> 
> 1. Within Cloudflare > click on your domain > SSL/ TLS > Client Certificates > Create Certificate > Here I used the standard settings (RSA Key and 10 Years validity)
> 2. Create mTLS Rules which brings you to the WAF settings > Custom rules > Create rule
> 3. Choose a fancy name > Field: Hostname > Operator: is in > Value: mysubdomain.domain.com > add a 2nd line of rules > Field: Client Certificate Verified > equals not. -> The expression looks like this: `(http.host in {"mysubdomain.domain.com"} and not cf.tls_client_auth.cert_verified)` > Then take action: Block > Safe
> 
> So far this seems to block everything, now I just need to figure if it grants access once the certificate is deployed :)

Sounds like you got it. That is how I set it up as well. Here's a picture of the rule I set up in case it is helpful:
image

kingamajick:
> * ha-mtls.ericmedina024.com
> 
> This is what I use for the app as I have the mTLS cert installed on my phone (under "VPN & app user cert" in settings). It is configured via a WAF rule to block any traffic that doesn't present a valid mTLS cert.

Hi, thanks for putting this together, looks like it could be an acceptable work around for me at least. I had one question, I assume the ha-mtls domain. How did you disable authentication for this domain? The cloudflared zero trust doesn't have an option to run a endpoint without auth as far as I can see?

jeleniain:
@kingamajick you don't have to create an access application for the tunnel. I solely have the tunnel configured and than the firewall rule.

kingamajick:
@jeleniain thanks, removed that and it all works as expected :D

jeleniain:
Nice! I also had some troubles installing the certificate on my Android phone. Turned out to be a format issue of the certificate where it caused an endless loop when my phone tried to apply it.
Let me know if you have troubles with that. otherwise I have been using this solution for about a month now and it works flawless. Super nice solution and I feel so much better to have my HA behind 2fa and a certificate :)

przymul:
@jeleniain Hey, if could help with the certificate part on android. After generating the certificate in cloud flare you get pem and key files. From what i searched you combine them using openssl so they can be installed in phone. Could you write the command you used that generated file that was able to be installed by android

some-guy-23:
@jeleniain / @EricMedina024 / @kingamajick trying to walk through this now and running into issue getting the Cloudflare mTLS certificate into the correct format to install on Android. Any step-by-step instruction on how you achieved this is greatly appreciated!

jeleniain:
@some-guy-23 happy you resolved it. I cant remember how I converted it but also concerted it to a .pfx file to be able to install it. Only issue I ran into after is that for some reason Firefox could not handle the mtls cert so I had to use chrome for it instead. No clue why that is but seems to be a FF limitation.

moshoari:
image

image


I couldn’t find the header section. Can anyone help me navigate to that section? I want to add the service token information to my phone application and filter this out in Zero Trust.

PR #3606, opened 2023-06-23
Update change log week of 6/23

dshokouhi:


jpelgrom:
> native Assist

I wonder if this is the best way to phrase it, maybe it should focus on the advantages instead (use Assist anywhere, HTTPS requirement workaround)?

dshokouhi:
> > native Assist
> 
> I wonder if this is the best way to phrase it, maybe it should focus on the advantages instead (use Assist anywhere, HTTPS requirement workaround)?

Thats a good question, I figured to make it more generic since we didnt have any support for it in the phone side of things.

JBassett (approved):

PR #3610, opened 2023-06-27
Hide devices that are home from the android auto navigation screen

dshokouhi:
## Summary

Just a small clean up PR. I think if you have a lot of devices that are at home and have a device tracker it may seem a bit redundant to list them all. One example is if a user has wall mounted tablets with the app installed.

I decided not to touch `person` entity as you may not care if they are home or not, you just want to navigate to them.

sensors and zones are left untouched as sensors may not have an expected state and zones should always be shown like person.

Documentation: home-assistant/companion.home-assistant#949


jpelgrom:
This change also needs a docs update

dshokouhi:
> This change also needs a docs update

good call, updated the docs :)

jpelgrom (approved):
Sensible change in my opinion as most device trackers will be linked to a person

JBassett (approved):

PR #3623, opened 2023-06-30
Hide a few GMS features unavailable on Automotive

jpelgrom:
## Summary

Hides a few features that depend on GMS APIs currently unavailable on Automotive according to this list: https://developer.android.com/training/cars/google-services. Including the dependencies in the app doesn't cause the crash considering the current release so a simple check will be enough and keeps manifest/dependencies the same for `app` and `automotive`.

(activity sensors are in `com.google.android.gms.location` weirdly enough, so not hiding those for now)

With the native interface now hidden it doesn't really matter that much but still nice to be prepared.

## Screenshots

n/a

n/a


dshokouhi (approved):
Still got more things to hide but this is a great start!

JBassett (approved):

PR #3641, opened 2023-07-06
Only invalidate vehicle template UI when contents change

### LINKED ISSUE #3328: Navigation in Android Auto is jittery when using rotary input
2023.1.1.full

**Android version(s):** 
13

**Device model(s):** 
Samsung Z Fold 4

**Home Assistant version:**
2023.2.3

2018 Audi MMI only has rotary input. While using HA in AA, scrolling is unpredictable and the cursor doesn't always go where it should. 

**Companion App Logs:**

```

```

https://photos.app.goo.gl/BfE8Skn2Dm7oqCfa9
**Additional information:**

jpelgrom:
## Summary

This PR updates the vehicle service entity state changes collection to only cause an invalidation of the template UI if the contents change, instead of on every change (which is any state change, as a 'new' list is received).

Fixes #3328, which is easily reproduced when using the DHU + rotary input + a state change. Any invalidation would cause the position to be reset when scrolling, after this change it is far less likely to be invalidated while scrolling.

## Screenshots

No visual changes if the code is correct :)

n/a

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
@@ -49,8 +49,9 @@ class EntityGridVehicleScreen(
             lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
                 entitiesFlow.collect {
                     loading = false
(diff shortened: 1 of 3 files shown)


dshokouhi (approved):
This should make some users very happy!

GSzabados:
> very happy!

Indeed! I just tested it. And it works really well! Thanks for this fix!

PR #3642, opened 2023-07-07
Add Android Auto notification extension

### LINKED ISSUE #3520: Notification/Popup on Android Auto

It is not related to any problem but a suggestion on how to enhance some of the automations we have using home assistant.

When connecting my android to my car, I normamly get some popups on screen from different apps like maps and whatsapp. I would like to get popups from my android home assistant automation if possible so I can automate some popups to easy access like:
Do you wish to open your garage door ? (when getting close to your home )
Would you like to turn off your office ? (when leaving the house)
etc.....
At the moment I am already sending those notifications to my android but they are not showing up on my android auto screen. That would make it much more secure if I could interact with the android auto device instead of my mobile.

At the moment, I have to get my mobile and click on the automations I have set to send notifications/popups but that represents a security risk as I am driving while looking at my phone

**Additional context**

Thanks a lot for the amazing work you are putting into this 👏

jpelgrom:
## Summary

Add an option to have notifications show up on Android Auto. This will allow users to show any notification they want by adding a single line. This can be used for anything, like critical alerts like the docs example or getting sensors updated (#3399, instead of forcing a notification this can be handled with an automation).

Notification icons aren't processed/changed as much in Android Auto as the status bar, so also two related minor changes to make them look good:
 - Apply color filter to MDI notification icons as Android Auto doesn't make them white
 - Switch default notification icon for vector icon to prevent low res icon in Android Auto

The setting is named `car_ui` as these changes may also benefit Android Automotive and maybe even iOS/CarPlay can use it in the future? Automotive is unclear and could depend on the manufacturer though so not mentioning it in the docs.

Fixes #3520

## Screenshots

Android Auto head unit with a Home Assistant notification on top of the screen

Actions will show up with their [icon](https://github.com/home-assistant/android/blob/be022a5272c8198750686308b4935da6a894d7bd/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt#L1425), so I guess we need an option to specify the icon soon:
Android Auto head unit with a Home Assistant notification on top of the screen, and a button with a Home Assistant icon on the right

Documentation: home-assistant/companion.home-assistant#951


common/src/main/java/io/homeassistant/companion/android/common/notifications/NotificationFunctions.kt
```diff
@@ -211,7 +214,7 @@ fun handleSmallIcon(
         val iconDrawable =
             IconicsDrawable(context, "cmd-$iconName")
         if (iconDrawable.icon != null) {
-            builder.setSmallIcon(iconDrawable.toAndroidIconCompat())
+            builder.setSmallIcon(iconDrawable.colorFilter { PorterDuffColorFilter(Color.WHITE, PorterDuff.Mode.SRC_IN) }.toAndroidIconCompat())
```
dshokouhi, line 217: Are we forcing the color to be white for android auto? Should we add a check so this only applies to android auto notifications to not impede on others?

jpelgrom: Yes it's forced. This shouldn't impede as for the status bar Android normally forces it to be white, but Android Auto doesn't for some reason which is why I'm adding it (MDI defaults to black).

JBassett (approved):

Issue #3648, opened 2023-07-09
Media Session attributes not updating


Erelen-Laiquendi:
**Home Assistant Android app version(s):** 2023.6.0-full

**Android version(s):** 9, 10, 11

**Device model(s):** Samsung Note 9, S9, Lenovo P10, Sony X1

**Home Assistant version:** 2023.6.3

**Last working Home Assistant release (if known):** -

**Description of problem, include YAML if issue is related to notifications:**
`sensor.<>_media_session` updates only if state change (play -> pause, etc). There is no way to receive new attributes (`title_*`, `playback_position_*` etc) - neither in 15 minutes updates, nor using `command_update_sensors`.
So, after several hours non-stop playing music on phone, in HA I still see first track info.

**Companion App Logs:**

```

```

**Screenshot or video of problem:**

**Additional information:**

dshokouhi:
did you make sure to add your media player app to the allow list for the notification sensors?

its part of the recommendations in the docs

https://companion.home-assistant.io/docs/core/sensors#media-session-sensor

Erelen-Laiquendi:
> did you make sure to add your media player app to the allow list for the notification sensors?

I don't use "Last notification" sensor at all.
"Media session" sensor - is ok for my task, if I can get actual info when I need it by sending `command_update_sensors`.

> `forceUpdate = primaryPlaybackState == "Playing"`

Hm...
I pause player, when it playing "Song A". In next sensor update (every 15 minutes or by command) I receive in Media Session sensor. After that I switch player to "Song B" and pause it again. As I understand, in this case I never can receive "Song B" to HA sensor.

dshokouhi:
> "Media session" sensor - is ok for my task, if I can get actual info when I need it by sending `command_update_sensors`.

in the case of media players they must post a notification with media data, by using the last notification sensor you will get updates immediately upon changes. We mention that in the docs linked.

> I pause player, when it playing "Song A". In next sensor update (every 15 minutes or by command) I receive in Media Session sensor. After that I switch player to "Song B" and pause it again. As I understand, in this case I never can receive "Song B" to HA sensor.

When something is paused the likelihood of it changing is very low. We need to be mindful of the amount of updates we send out. Next time you go to play the item it will update. The likelihood of the same item being paused for hours sending the same update is a lot higher than something being paused with attributes changing.

jpbaril:
> in the case of media players they must post a notification with media data

[...]

PR #3652, opened 2023-07-10
Force update on media session sensor when a session is in progress

dshokouhi:
## Summary

Fixes: #3648  by forcing updates of the sensor when a media session is in progress. I chose to go by `Unavailable` state as that particular state should only get 1 update.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
@@ -283,7 +283,7 @@ class NotificationSensorManager : NotificationListenerService(), SensorManager {
             primaryPlaybackState,
             mediaSession.statelessIcon,
             attr,
-            forceUpdate = primaryPlaybackState != "Unavailable"
+            forceUpdate = primaryPlaybackState == "Playing"
         )
     }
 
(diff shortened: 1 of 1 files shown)


jpelgrom:
This works, but will cause unnecessary updates if you paused media but the app still has an active media session (which is quite likely if the app is still open in the background). Would it be possible to compare the attributes instead to decide if an update is needed?

dshokouhi:
> This works, but will cause unnecessary updates if you paused media but the app still has an active media session (which is quite likely if the app is still open in the background). Would it be possible to compare the attributes instead to decide if an update is needed?

thats a good point, I wonder which attribute would be best to go. Possibly `title` ? That seems to be the most unique I think.

dshokouhi:
actually maybe we need to make it more direct and only on `Playing` state? the other states I dont see updating too much. Maybe Fast forward and Rewind playback position but thats a very trivial attribute.

dshokouhi:
Switched the logic as I think it makes the most sense....all other states I would not expect the attributes to update that much if at all.

jpelgrom:
> Maybe Fast forward and Rewind playback position but thats a very trivial attribute.

The app's sensors also don't really update frequently enough for it to be useful, playing only should be good enough to support the issue's use case.

JBassett (approved):

PR #3660, opened 2023-07-14
Update change log week of 7/14

dshokouhi:

### DIFF
--- a/app/src/main/res/xml/changelog_master.xml
+++ b/app/src/main/res/xml/changelog_master.xml
@@ -1,14 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <changelog xmlns:tools="http://schemas.android.com/tools"
     tools:ignore="MissingDefaultResource">
-    <release version="2023.8 - Phone" versioncode="3">
+    <release version="2023.7.6 - Main" versioncode="3">
         <change>Tiles: Replace entity drop down selection with search based text field to easily find entities</change>
         <change>Bug fixes and dependency updates</change>
     </release>
-    <release version="2023.8 - Wear" versioncode="2">
+    <release version="2023.7.6 - Wear" versioncode="2">
         <change>Bug fixes and dependency updates</change>
     </release>
-    <release version="2023.8 - Automotive" versioncode="1">
+    <release version="2023.7.6 - Automotive" versioncode="1">
         <change>Bug fixes and dependency updates</change>
     </release>
 </changelog>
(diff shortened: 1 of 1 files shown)


JBassett:
This brings up a good question, when do we want to bump the changelog version from `2023.7.x` to `2023.8.x`?  Technically we could do another production release before the end of the month.

dshokouhi:
> This brings up a good question, when do we want to bump the changelog version from `2023.7.x` to `2023.8.x`? Technically we could do another production release before the end of the month.

great question, not really sure. I am just in the habit of bumping the date after a release lol

jpelgrom:
We could also drop the version number from the titles, as the change log is only shown when updated?

dshokouhi:
> We could also drop the version number from the titles, as the change log is only shown when updated?

I like that, one less thing to change lol. Should we replace `Phone` with `Device` to be more generic? I think `Wear` and `Automotive` are more self explanatory?

jpelgrom:
I would also remove the phone title if possible but I don't think it is. Device feels a bit too generic here 🤔

dshokouhi:
> I would also remove the phone title if possible but I don't think it is.

yea app crashes, its a required field.

Heres one problem, the word `Version` is still listed next to each row. Was trying to test if `Phone\Tablet` was too much (forgot to escape the slash lol) but it looks kinda odd now.

image

dshokouhi:
Device might feel a bit generic but we have to think of a term thats generic since the following device types can be used on that version.

- phone
- tablet
- chromebook
- windows
- anything else, quite literally lol

jpelgrom:
[Looks like 'Version' is hardcoded.](https://github.com/AppDevNext/ChangeLog/blob/master/ChangeLogLib/src/main/java/info/hannes/changelog/ChangeLog.kt#L270) To be fair this is a very simple library - it might be time to create something simple ourselves if we want to customize it.

In the mean time, maybe update the changelog each time with the tag that will be used, so 2023.7.6 now?

> Device might feel a bit generic but we have to think of a term thats generic since the following device types can be used on that version.

Simply 'Companion app', or 'main app'?

dshokouhi:
> In the mean time, maybe update the changelog each time with the tag that will be used, so 2023.7.6 now?

thats easier and a bit more precise :)



> Simply 'Companion app'

I think users may confuse that to mean anything mentioned in the first listing is available everywhere because the app is called Home Assistant Companion.



> or 'main app'?

This sounds more correct 🤷

JBassett (approved):

PR #3663, opened 2023-07-16
Add toggle for Assist as assistant app on Wear

jpelgrom:
## Summary

 Adds a toggle for enabling/disabling the option to use Assist as an assistant app (= intent filter on Assist activity), as users may not be able to change the app used after choosing 'Always' and Wear OS 2 doesn't ask the user which app to use, and users can't change the default assistant app, which can result in Assist instead of the Google Assistant being forced.

Also reported in [this forum post](https://community.home-assistant.io/t/wearos-companion-issues/592070) and [these Discord messages](https://discord.com/channels/330944238910963714/562408603345092636/1129914534732243085).

## Screenshots

Ironically, this is best visible when you can choose the assistant app, as you will see Home Assistant show or hide depending on the setting:
|Setting in app|On: HA available|Off: HA removed|
|-----|-----|-----|
|'Available as assistant app' setting in the Home Assistant app, on with an Assist icon|'Default digital assistant app' system setting showing 'None', 'Google Assistant', and 'Home Assistant' as options|'Default digital assistant app' system setting showing only 'None' and 'Google Assistant' as options|

n/a

## Any other notes

Tested on Wear OS 2 emulator with `adb shell am start -a android.intent.action.VOICE_ASSIST` and a Galaxy Watch 4.

### DIFF
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -35,6 +35,7 @@
     <string name="automation">Automation</string>
     <string name="autoplay_video_summary">Autoplay Videos when dashboard is active. Enabling this setting may increase data usage unexpectedly, proceed with caution.</string>
     <string name="autoplay_video">Autoplay Videos</string>
+    <string name="available_as_assistant_app">Available as assistant app</string>
     <string name="background_access_disabled">Home Assistant does not have access to run in the background. Without this permission the app will not be able to reliably send data back to your server. Click here to request permissions.</string>
     <string name="background_access_enabled">Home Assistant has access to run in the background.</string>
     <string name="background_access_title">Background Access</string>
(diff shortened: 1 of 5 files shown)


dshokouhi:
Based on the reported issues should we also drop a line in the docs? Seems like some users keep searching with no idea what to do. :(

dshokouhi (approved):
We may want to roll this one out sooner than later because the only real fix at the moment is to uninstall the app on those devices.

jpelgrom:
> Based on the reported issues should we also drop a line in the docs? Seems like some users keep searching with no idea what to do. :(

Might be good to update the [general Assist on Android docs](https://www.home-assistant.io/voice_control/android/) and mention this setting in there, I'll put something together soon.

JBassett (approved):

Issue #3219, opened 2023-01-12
Declutter Android Auto


GSzabados:
**Is your feature request related to a problem? Please describe.**

As #791 has been closed and Android Auto support is available. It would be nice to declutter it from possibly dozens or hundreds of entities.

**Describe the solution you'd like**

One option to solve the problem is to add a Favourites option where only selected entities will show up. Like what is needed from the car. 

**Describe alternatives you've considered, if any**

The other option is to provide an option from the app where only selected entities would get into the Android Auto display under each category. 

**Additional context**

In my case, I have only 1 real door lock, but all the Child Locks from Z2M shows up under the locks section. Selecting only the single lock which is needed in Android Auto would make it less cluttered.

I believe everyone who has an extensive setup of device will see the same issue as all switches, lights, etc shows up and some might not needed when driving and just distracts meanwhile searching for the right one.

dshokouhi:
I like the favorites suggestion as it falls in line with Wear OS where if a user select favorite entities they appear first and then just below all the other entities will be loaded.

JBassett:
It's on my list of things to tackle.

GSzabados:
Just to add as some additional information. 

The favourite way is good for easy access for selected entities, but it seems to be that the additional selection for entities would do a really good job as well.

As I wrote before, all the child locks for my Zigbee sockets ends up under the the lock category, and makes it really messy.
Also, I had to realise that from Zwave all the Ping push buttons ends up in the buttons category, and makes that as well messy.
The ESPHome switches which used for restarting ESPHome devices end up in the switch category as well.

Unfortunately some of these fall under Configuration but some are Controls in HA's device/entity categories.
So no easy solution just to disregard Configuration entities. 

Also, groups are not listed, neither if they are lights or switches. But I think that is different issue.

dshokouhi:
> Also, groups are not listed, neither if they are lights or switches. But I think that is different issue.

Yes `group` is not part of the initial supported domains

https://github.com/home-assistant/android/blob/master/app/src/full/java/io/homeassistant/companion/android/vehicle/MainVehicleScreen.kt#L51-L61

you can open a separate feature request for domains not currently shown :)

GSzabados:
> you can open a separate feature request for domains not currently shown :)

I will open one definitely.

I have to open a bug also, when I will have some time. The navigation in the menu is quite clunky. I need to talk with some other AA users first, because I cannot be sure that is it my head unit only or the way how it is working.

[...]

PR #3670, opened 2023-07-18
Add favorites selection for Android Auto

dshokouhi:
## Summary

Fixes: #3219 by adding a favorites category at the top to include a user selected list of entities on a per server basis.

Did some minor refactoring to make the favorite entity row shareable to Wear. Also makes use of the new simple entity picker!

## Screenshots

image

image

image

image

Documentation: home-assistant/companion.home-assistant#961

## Any other notes

I opted to place the category at the top of the list for now as this PR is already pretty big. I think in the future a tabbed template might be a good fit for a redesign but a bit out of scope for this PR.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -204,7 +204,7 @@ dependencies {
     implementation("com.mikepenz:iconics-compose:5.4.0")
     implementation("com.mikepenz:community-material-typeface:7.0.96.0-kotlin@aar")
 
-    "fullImplementation"("org.burnoutcrew.composereorderable:reorderable:0.9.6")
+    implementation("org.burnoutcrew.composereorderable:reorderable:0.9.6")
     implementation("com.github.AppDevNext:ChangeLog:3.4")
 
     implementation("androidx.car.app:app:1.3.0-rc01")
--- a/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
@@ -42,7 +42,7 @@ class ManageAndroidAutoViewModel @Inject constructor(
     val entities = mutableMapOf<Int, List<Entity<*>>>()
     init {
         viewModelScope.launch {
-            favoritesList.addAll(if (prefsRepository.getAutoFavorites().isNotEmpty()) prefsRepository.getAutoFavorites().split(", ") else emptyList())
+            favoritesList.addAll(prefsRepository.getAutoFavorites().ifEmpty { emptyList() })
             entitiesLoaded = true
             serverManager.defaultServers.map {
                 async {
--- a/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
@@ -1,0 +1,91 @@
+package io.homeassistant.companion.android.settings.vehicle
+
+import android.app.Application
+import android.os.Build
+import android.util.Log
+import androidx.annotation.RequiresApi
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateListOf
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
+import androidx.lifecycle.AndroidViewModel
+import androidx.lifecycle.viewModelScope
+import dagger.hilt.android.lifecycle.HiltViewModel
+import io.homeassistant.companion.android.common.data.integration.Entity
+import io.homeassistant.companion.android.common.data.integration.domain
+import io.homeassistant.companion.android.common.data.prefs.PrefsRepository
+import io.homeassistant.companion.android.common.data.servers.ServerManager
+import io.homeassistant.companion.android.vehicle.MainVehicleScreen
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.launch
+import org.burnoutcrew.reorderable.ItemPosition
+import javax.inject.Inject
+
+@RequiresApi(Build.VERSION_CODES.O)
+@HiltViewModel
+class ManageAndroidAutoViewModel @Inject constructor(
+    private val serverManager: ServerManager,
+    private val prefsRepository: PrefsRepository,
+    application: Application
+) : AndroidViewModel(application) {
+
+    companion object {
+        private const val TAG = "AAViewModel"
+    }
+    val favoritesList = mutableStateListOf<String>()
+
+    var entitiesLoaded by mutableStateOf(false)
+        private set
+    var sortedEntities by mutableStateOf<List<Entity<*>>>(emptyList())
+        private set
+    val entities = mutableMapOf<Int, List<Entity<*>>>()
+    init {
+        viewModelScope.launch {
+            favoritesList.addAll(prefsRepository.getAutoFavorites())
+            entitiesLoaded = true
+            serverManager.defaultServers.map {
+                async {
+                    entities[it.id] = try {
+                        serverManager.integrationRepository(it.id).getEntities().orEmpty()
+                            .filter { it.domain in MainVehicleScreen.SUPPORTED_DOMAINS }
+                    } catch (e: Exception) {
+                        Log.e(TAG, "Couldn't load entities for server", e)
+                        emptyList()
+                    }
+                }
+            }.awaitAll()
+            loadEntities(serverManager.getServer()?.id ?: 0)
+        }
+    }
+
+    fun onMove(fromItem: ItemPosition, toItem: ItemPosition) {
+        favoritesList.apply {
+            add(
+                favoritesList.indexOfFirst { it == toItem.key },
+                removeAt(favoritesList.indexOfFirst { it == fromItem.key })
+            )
+        }
+    }
+
+    fun canDragOver(position: ItemPosition) = favoritesList.any { it == position.key }
+
+    fun saveFavorites() {
+        viewModelScope.launch {
+            prefsRepository.setAutoFavorites(favoritesList)
+        }
+    }
+
+    fun loadEntities(serverId: Int) {
+        sortedEntities = entities[serverId] ?: emptyList()
+    }
+
+    fun onEntitySelected(checked: Boolean, entityId: String, serverId: Int) {
+        if (checked) {
+            favoritesList.add("$serverId-$entityId")
+        } else {
+            favoritesList.remove("$serverId-$entityId")
+        }
+        viewModelScope.launch { prefsRepository.setAutoFavorites(favoritesList) }
+    }
+}
--- a/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
@@ -35,15 +35,12 @@ class ManageAndroidAutoViewModel @Inject constructor(
     }
     val favoritesList = mutableStateListOf<String>()
 
-    var entitiesLoaded by mutableStateOf(false)
-        private set
     var sortedEntities by mutableStateOf<List<Entity<*>>>(emptyList())
         private set
     val entities = mutableMapOf<Int, List<Entity<*>>>()
     init {
         viewModelScope.launch {
-            favoritesList.addAll(prefsRepository.getAutoFavorites().ifEmpty { emptyList() })
-            entitiesLoaded = true
+            favoritesList.addAll(prefsRepository.getAutoFavorites())
             serverManager.defaultServers.map {
                 async {
                     entities[it.id] = try {
--- a/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
@@ -42,7 +42,7 @@ class ManageAndroidAutoViewModel @Inject constructor(
     val entities = mutableMapOf<Int, List<Entity<*>>>()
     init {
         viewModelScope.launch {
-            favoritesList.addAll(prefsRepository.getAutoFavorites())
+            favoritesList.addAll(if (prefsRepository.getAutoFavorites().isNotEmpty()) prefsRepository.getAutoFavorites().split(", ") else emptyList())
             entitiesLoaded = true
             serverManager.defaultServers.map {
                 async {
@@ -72,7 +72,7 @@ class ManageAndroidAutoViewModel @Inject constructor(
 
     fun saveFavorites() {
         viewModelScope.launch {
-            prefsRepository.setAutoFavorites(favoritesList)
+            prefsRepository.setAutoFavorites(favoritesList.toList())
         }
     }
 
@@ -86,6 +86,6 @@ class ManageAndroidAutoViewModel @Inject constructor(
         } else {
             favoritesList.remove("$serverId-$entityId")
         }
-        viewModelScope.launch { prefsRepository.setAutoFavorites(favoritesList) }
+        viewModelScope.launch { prefsRepository.setAutoFavorites(favoritesList.toList()) }
     }
 }
--- a/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
@@ -8,6 +8,7 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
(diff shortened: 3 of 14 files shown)


jpelgrom:
This also needs to be available for Automotive, perhaps with a different name (driving interface favorites?).

app/build.gradle.kts
```diff
@@ -204,7 +204,7 @@ dependencies {
     implementation("com.mikepenz:iconics-compose:5.4.0")
     implementation("com.mikepenz:community-material-typeface:7.0.96.0-kotlin@aar")
 
-    "fullImplementation"("org.burnoutcrew.composereorderable:reorderable:0.9.6")
+    implementation("org.burnoutcrew.composereorderable:reorderable:0.9.6")
```
jpelgrom, line 207: This should be added to Automotive as well

common/src/main/res/values/strings.xml
```diff
@@ -1114,4 +1114,9 @@
     <string name="only_favorites">Only Show Favorites</string>
     <string name="beacon_scanning">Beacon Monitor Scanning</string>
     <string name="car_data_unavailable">Open Home Assistant app to activate the sensor</string>
+    <string name="aa_set_favorites">Select your favorite entities to appear in the Favorites category in the Android Auto//Automotive app. You can also drag and drop to change the order in which they appear.</string>
+    <string name="aa_favorites">Android Auto Favorites</string>
+    <string name="aa_favorites_summary">Select your favorite entities to be shown in Android Auto</string>
+    <string name="android_automotive">Android Automotive</string>
+    <string name="android_automotive_favorites">Android Automotive Favorites</string>
```
jpelgrom, line 1121: Still not sure about this; you're already using Automotive. Do we call the template UI Automotive mode vs native mode? Isn't native mode a feature of Automotive?

app/build.gradle.kts
dshokouhi: looks like it already is :)

https://github.com/home-assistant/android/blob/2804e7c4a854c64792196374bb728b01d082900e/automotive/build.gradle.kts#L236

common/src/main/res/values/strings.xml
dshokouhi: Looking at the official docs the terms "Driving Optimized" is used a lot in both places. I don't think there is an exact name for it.

https://developer.android.com/training/cars

let me update the summary text to mention the screen shown while driving?

dshokouhi: ok that should hopefully be a bit more clear about what it is and where its used

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
```diff
@@ -0,0 +1,117 @@
...
+                onEntitySelected = {
+                    androidAutoViewModel.onEntitySelected(true, it, selectedServer)
+                    return@SingleEntityPicker false // Clear input
+                },
+                modifier = Modifier.padding(all = 16.dp),
+                label = { Text(stringResource(commonR.string.add_favorite)) }
+            )
+        }
+        if (favoriteEntities.isNotEmpty() && androidAutoViewModel.sortedEntities.isNotEmpty()) {
+            items(favoriteEntities.size, { favoriteEntities[it] }) { index ->
```
jpelgrom, line 89: The driving interface only shows one server's favorites at a time, you switch servers to see the other server favorites. Shouldn't the same apply to the settings for favorites?

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
```diff
@@ -0,0 +1,117 @@
...
+                val favoriteEntityID =
+                    favoriteEntities[index].replace("[", "").replace("]", "").split("-")[1]
+                androidAutoViewModel.sortedEntities.filter { it.entityId == favoriteEntityID }.let {
```
jpelgrom, line 92: Easier: `.firstOrNull { it.entityId == favroiteEntityID }?.let {`

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
```diff
@@ -0,0 +1,117 @@
...
+                            entityName = it.first().friendlyName,
+                            entityId = it.first().entityId,
+                            onClick = {
+                                androidAutoViewModel.onEntitySelected(
+                                    false,
+                                    it.first().entityId,
+                                    selectedServer
+                                )
+                            },
+                            checked = favoriteEntities.contains(favoriteEntities[index]),
```
jpelgrom, line 107: At this point you might as well make it `true`, this comparison adds nothing?

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
```diff
@@ -0,0 +1,117 @@
...
+                        androidAutoViewModel.loadEntities(it)
+                        selectedServer = it
+                    },
+                    modifier = Modifier.fillMaxWidth().padding(start = 16.dp, end = 16.dp, bottom = 16.dp)
+                )
+            }
+        }
+        item {
+            SingleEntityPicker(
+                entities = androidAutoViewModel.sortedEntities,
```
jpelgrom, line 77: The available entities should be updated to prevent you from selecting the same entity twice (it will also crash the app). See the Wear favorites selection for an example.

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
```diff
@@ -197,4 +198,12 @@ class PrefsRepositoryImpl @Inject constructor(
     override suspend fun setIgnoredSuggestions(ignored: List<String>) {
         localStorage.putStringSet(PREF_IGNORED_SUGGESTIONS, ignored.toSet())
     }
+
+    override suspend fun getAutoFavorites(): List<String> {
+        return localStorage.getStringSet(PREF_AUTO_FAVORITES)?.toList() ?: emptyList()
+    }
+
+    override suspend fun setAutoFavorites(favorites: List<String>) {
+        localStorage.putStringSet(PREF_AUTO_FAVORITES, favorites.toSet())
```
jpelgrom, line 207: Sets [do not preserve order](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-set/#:~:text=A%20generic-,unordered%20collection,-of%20elements%20that%20does%20not%20support%20duplicate%20elements.%20Methods), reordering doesn't work.

common/src/main/res/values/strings.xml
jpelgrom: I don't see "Driving Optimized" on android.com/auto anywhere, the consumer-facing site.

> Driving Favorites

I like that, that should allow us to stay away from developer vocabulary.

dshokouhi: went with Driving Favorites here :)

common/src/main/res/values/strings.xml
```diff
@@ -1114,4 +1114,9 @@
     <string name="only_favorites">Only Show Favorites</string>
     <string name="beacon_scanning">Beacon Monitor Scanning</string>
     <string name="car_data_unavailable">Open Home Assistant app to activate the sensor</string>
+    <string name="aa_set_favorites">Select your favorite entities to appear in the Favorites category in the Android Auto//Automotive driving optimized app. You can also drag and drop to change the order in which they appear.</string>
```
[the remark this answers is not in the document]
dshokouhi: for some reason I thought it needed to be escaped, oops lol

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
dshokouhi: this should be fixed now

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
dshokouhi: done, we now account for the server ID

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
jpelgrom: It isn't? If favorites start as ABC, and I change it to BAC, the next time I open the settings it's back to ABC.

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
```diff
@@ -0,0 +1,132 @@
...
+                    return@SingleEntityPicker false // Clear input
+                },
+                modifier = Modifier.padding(all = 16.dp),
+                label = { Text(stringResource(commonR.string.add_favorite)) }
+            )
+        }
+        if (favoriteEntities.isNotEmpty() && androidAutoViewModel.sortedEntities.isNotEmpty()) {
+            items(favoriteEntities.size, { favoriteEntities[it] }) { index ->
+                val favoriteEntity =
+                    favoriteEntities[index].replace("[", "").replace("]", "").split("-")
```
jpelgrom, line 106: Unless I'm missing something favorites never contain `[]`? Unless that is the fix for ordering favorites, just changing the string set into a string that holds the array like we do in the database (should actually work, entity IDs never contain `[],`).

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
dshokouhi: in my example I had Den Lights and Kitchen Lights.  I moved Kitchen Lights to be above Den Lights in the UI. In Android Auto it shows the order I used.

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
dshokouhi: Its converted to a string so if its the first or last value, or only value it has the potential to be wrapped in `[]`

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
jpelgrom: Try reordering without adding favorites, UI doesn't necessarily reflect how it is stored because it's converting set <> list.

jpelgrom: Nope not working

https://github.com/home-assistant/android/assets/8148535/cfeaa977-ccb7-4a69-900f-931e9c04b78d

> if it works in practice I think its ok as of now most users will only use 1 or 2 devices here

It should work for any number of favorites, for 1 or 2 we wouldn't need reordering

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
```diff
@@ -0,0 +1,91 @@
...
+    private val prefsRepository: PrefsRepository,
+    application: Application
+) : AndroidViewModel(application) {
+
+    companion object {
+        private const val TAG = "AAViewModel"
+    }
+    val favoritesList = mutableStateListOf<String>()
+
+    var entitiesLoaded by mutableStateOf(false)
```
jpelgrom, line 38: This isn't used and can be removed

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
```diff
@@ -0,0 +1,91 @@
...
+        private set
+    var sortedEntities by mutableStateOf<List<Entity<*>>>(emptyList())
+        private set
+    val entities = mutableMapOf<Int, List<Entity<*>>>()
+    init {
+        viewModelScope.launch {
+            favoritesList.addAll(prefsRepository.getAutoFavorites().ifEmpty { emptyList() })
```
jpelgrom, line 45: What does `ifEmpty { emptyList() }` add here?

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
dshokouhi: this should be fixed now in the latest commit :)

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
```diff
@@ -0,0 +1,132 @@
...
+        onMove = { from, to -> androidAutoViewModel.onMove(from, to) },
+        canDragOver = { draggedOver, _ -> androidAutoViewModel.canDragOver(draggedOver) },
+        onDragEnd = { _, _ -> androidAutoViewModel.saveFavorites() }
+    )
+
+    var selectedServer by remember { mutableStateOf(defaultServer) }
+
+    val favoriteEntities = androidAutoViewModel.favoritesList.toList()
+    var validEntities by remember { mutableStateOf<List<Entity<*>>>(emptyList()) }
+    LaunchedEffect(favoriteEntities.size, androidAutoViewModel.sortedEntities.size) {
```
jpelgrom, line 52: This also needs the server ID in case the user has two servers with identical numbers of entities.

```suggestion
    LaunchedEffect(favoriteEntities.size, androidAutoViewModel.sortedEntities.size, selectedServer) {
```

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
```diff
@@ -0,0 +1,132 @@
...
+        validEntities = withContext(Dispatchers.IO) {
+            androidAutoViewModel.sortedEntities
+                .filter {
+                    !favoriteEntities.contains("$selectedServer-${it.entityId}") &&
+                        (it.entityId.split(".")[0] in MainVehicleScreen.SUPPORTED_DOMAINS)
```
jpelgrom, line 57: Use the `domain` extension property here instead of splitting

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
```diff
@@ -0,0 +1,132 @@
...
+        if (serversList.size > 1) {
+            item {
+                ServerExposedDropdownMenu(
+                    servers = serversList,
+                    current = selectedServer,
+                    onSelected = {
+                        androidAutoViewModel.loadEntities(it)
+                        selectedServer = it
+                    },
+                    modifier = Modifier.fillMaxWidth().padding(all = 16.dp)
```
jpelgrom, line 86: When using more than one server there is too much padding between the 2 dropdowns (32dp instead of 16dp). Suggestion for Text above, this dropdown and the single entity picker: on each, set horizontal + bottom padding to fix this -> `.padding(horizontal = 16.dp).padding(bottom = 16.dp)`.

common/src/main/res/values/strings.xml
```diff
@@ -1121,4 +1121,9 @@
     <string name="sensor_description_car_fuel_type">List of available fuel types for the connected car</string>
     <string name="basic_sensor_name_car_ev_connector_type">Car EV Connector Type</string>
     <string name="sensor_description_car_ev_connector_type">List of available EV connectors for the connected car</string>
+    <string name="aa_set_favorites">Select your favorite entities to appear in the Favorites category in the Android Auto/Automotive driving optimized app. You can also drag and drop to change the order in which they appear.</string>
```
jpelgrom, line 1124: Should this include a note on that the number of favorites shown may be limit by the vehicle?

Still not a big fan of "Android Auto/Automotive driving optimized app". It is inconsistent with the `aa_favorites_summary` and I think "Home Assistant driving interface" would be simpler and good enough.

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
jpelgrom: Now that it is converting a string to a list in the prefs repository, which removes all `[],`, the replacing isn't needed.

jpelgrom (commented):
Almost there! Re-ordering now works, tested in both single and multi server mode.

common/src/main/res/values/strings.xml
dshokouhi: updated and I agree thats simpler and good enough, thanks for the suggestion

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
dshokouhi: done but I noticed that when the server picker is not added the padding is still not good so added a padding for `top` only when we are not adding the server above the entity picker

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
dshokouhi: thanks thats better

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
dshokouhi: ah yes thats good!

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
jpelgrom: That might be because you didn't update Text in line 74, now we're back to the same issue of [the server dropdown being too close to the text](https://github.com/home-assistant/android/pull/3670#discussion_r1268542404).

common/src/main/res/values/strings.xml
jpelgrom: Update the aa_favorites_summary string as well to use "Home Assistant driving interface" instead of "driving optimized screen" for consistency?

dshokouhi: ah agreed, updated

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/views/AndroidAutoFavoritesView.kt
dshokouhi: Ah yes indeed i forgot about that one, fixed :)

jpelgrom (approved):

Issue #3250, opened 2023-01-21
Android Auto Beta - Hidden Entities Appear in AA Home Assistant App UI


Nerwyn:
**Home Assistant Android version:**
beta-3074-d7ee3953-full

**Android version:**
13, Security Update December 5, 2022

**Phone model:**
Google Pixel 7 Pro

**Home Assistant version:**
Home Assistant 2023.1.6
Supervisor 2022.12.1
Operating System 9.4

**Last working Home Assistant release (if known):**
NA

**Description of problem:**
Hidden entities appear in the Android Auto Home Assistant UI.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**
NA
```

```

**Screenshot of problem:**
Here are the lights in Android Auto
![PXL_20230121_031434434](https://user-images.githubusercontent.com/43071698/213841323-5348cdfc-8a4e-4245-a96c-998ca921d59c.jpg)

And here are my light entities. Note how the bulbs and UI created group are hidden.
![image](https://user-images.githubusercontent.com/43071698/213841340-a7e5fabf-fc28-471e-ad7d-562af977c54d.png)

**Additional information:**
NA

jpelgrom:
This is correct as the Android Auto interface doesn't read any data from the entity registry yet, it is more like the large entities list without any filters. 

Changing the label to mark this as a feature request.

PR #3678, opened 2023-07-18
Hide hidden entities from Android Auto domain categories

dshokouhi:
## Summary

Fixes: #3250 by adding a quick entity registry check when loading the list of entities in a domain.

n/a

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/vehicle/MainVehicleScreen.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/vehicle/MainVehicleScreen.kt
@@ -153,35 +153,51 @@ class MainVehicleScreen(
                 null
             ).getIcon(carContext)
 
-            listBuilder.addItem(
-                Row.Builder().apply {
-                    if (icon != null) {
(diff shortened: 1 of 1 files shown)


jpelgrom:
It's not doing live updates but to be fair that's probably fine considering it is refreshed when opened 😉

Building/updating the domains list should also take entity registry in account. If you have hidden al buttons the buttons domain shouldn't be shown.

As a general note, this will conflict with #3670, when one is merged the other one should be updated before merging to reflect the changes in available entities.

dshokouhi:
> Building/updating the domains list should also take entity registry in account. If you have hidden al buttons the buttons domain shouldn't be shown.

I agree that should be done, good call



> As a general note, this will conflict with #3670, when one is merged the other one should be updated before merging to reflect the changes in available entities.

Yup for now I will mark as draft until favorites is done as that is more important

dshokouhi:
Closing for now, will be easier to implement in a new branch since code moved around since this PR was created. Focus for now will be on favorites anyways since thats the new default.

PR #3681, opened 2023-07-19
Hilt Dependency Update

JBassett:
## Summary

Because the Kotlin DSL doesn't support variables we need to up manually bump them at once, really annoying.

N/A


jpelgrom:
As Android Studio will [soon use Gradle version catalogs by default](https://developer.android.com/studio/preview/features#gradle-version-catalogs), I guess we should move towards using those to 'fix' this. Also nice to have with `app` and `automotive` sharing 99% of dependencies.

JBassett:
It looks like dependabot would support that: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#gradle

jpelgrom (approved):

PR #3685, opened 2023-07-20
Only show unit of measurement if we have it and if the state is not a string while managing sensors

dshokouhi:
## Summary

Noticed that sometimes we were showing unit of measurement when the state was not numeric, so now we will hide the units to show the string properly.


jpelgrom:
You don't need to add parentheses, looks good otherwise.

dshokouhi:
> You don't need to add parentheses, looks good otherwise.

android studio complained and initially the code did not work so I assumed it needed :)

jpelgrom:
Works fine without here, using latest stable Android Studio release?

jpelgrom (approved):

JBassett (approved):

PR #3691, opened 2023-07-21
Update change log week of 7/21

dshokouhi:


jpelgrom:
> Default to last used pipeline, support preferred pipeline

It already supported the preferred pipeline, the change is that it is no longer the default. Also in UI 'assistant' is used instead of 'pipeline'. Suggestion: 'default to last used assistant'.

jpelgrom:
Favorites also applies to Automotive

dshokouhi:
> ~Favorites also applies to Automotive~ Never mind as most currently can't access the interface to add them

ya that is going to take some extra consideration since automotive play store users cannot access settings 😞

JBassett (approved):

PR #3703, opened 2023-07-24
Show measurement unit in Wear OS complication

### LINKED ISSUE #3613: Show unit in Wear OS sensor complication

I have some complications for sensors with numeric data. Especially if the entity name is hidden it feels like something is missing because the unit of the value is missing, and depending on the type of unit it can sometimes be really hard to even understand what the value means. For example: I have a complication showing the current amount of power produced by my solar panels, and it just shows the ⚡ icon and the value (`175.3`), not the `W` unit (see the screenshot below).

Show the unit of the sensor value after the actual sensor value. Possibly opt-in in the same way as hiding the name is configurable.

One workaround is to create a template sensor in HA with the unit being part of the value, but that feels redundant and adds complexity.

**Additional context**

image

slovdahl:
## Summary

Show the measurement unit of entities in Wear OS watch face complications to make the value easier to understand. Fixes #3613.

## Screenshots

Bottom left/right complications are from the current Google Play store stable HA version. Top left/right is from a version built from this PR.

Screenshot_20230724_104305_sysui

Documentation: home-assistant/companion.home-assistant#965

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
@@ -604,7 +604,7 @@ suspend fun <T> Entity<T>.onPressed(
 val <T> Entity<T>.friendlyName: String
     get() = (attributes as? Map<*, *>)?.get("friendly_name")?.toString() ?: entityId
 
-fun <T> Entity<T>.friendlyState(context: Context, options: EntityRegistryOptions? = null): String {
+fun <T> Entity<T>.friendlyState(context: Context, options: EntityRegistryOptions? = null, appendUnitOfMeasurement: Boolean = false): String {
     var friendlyState = when (state) {
         "closed" -> context.getString(commonR.string.state_closed)
         "closing" -> context.getString(commonR.string.state_closing)
@@ -649,6 +649,15 @@ fun <T> Entity<T>.friendlyState(context: Context, options: EntityRegistryOptions
             }
         }
     }
+
+    if (appendUnitOfMeasurement) {
+        val unit = (attributes as? Map<*, *>)?.get("unit_of_measurement")?.toString()
+
+        if (unit?.isNotEmpty() == true) {
+            return "$friendlyState $unit"
+        }
+    }
+
     return friendlyState
 }
 
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
@@ -653,7 +653,7 @@ fun <T> Entity<T>.friendlyState(context: Context, options: EntityRegistryOptions
     if (appendUnitOfMeasurement) {
         val unit = (attributes as? Map<*, *>)?.get("unit_of_measurement")?.toString()
 
-        if (unit?.isNotEmpty() == true) {
+        if (unit?.isNotBlank() == true) {
             return "$friendlyState $unit"
         }
     }
--- a/wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
@@ -79,7 +79,14 @@ class EntityStateDataSourceService : SuspendingComplicationDataSourceService() {
         } else {
             null
         }
-        val text = PlainComplicationText.Builder(entity.friendlyState(this, entityOptions)).build()
+
+        var stateText = entity.friendlyState(this, entityOptions)
+        if (entity.attributes.containsKey("unit_of_measurement")) {
+            stateText += " " + entity.attributes["unit_of_measurement"]
+        }
+
+        val text = PlainComplicationText.Builder(stateText).build()
+
         val contentDescription = PlainComplicationText.Builder(getText(R.string.complication_entity_state_content_description)).build()
         val monochromaticImage = MonochromaticImage.Builder(Icon.createWithBitmap(iconBitmap)).build()
         val tapAction = ComplicationReceiver.getComplicationToggleIntent(this, request.complicationInstanceId)
--- a/wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
@@ -80,12 +80,13 @@ class EntityStateDataSourceService : SuspendingComplicationDataSourceService() {
             null
         }
 
-        var stateText = entity.friendlyState(this, entityOptions)
-        if (entity.attributes.containsKey("unit_of_measurement")) {
-            stateText += " " + entity.attributes["unit_of_measurement"]
-        }
-
-        val text = PlainComplicationText.Builder(stateText).build()
+        val text = PlainComplicationText.Builder(
+            entity.friendlyState(
+                this,
+                entityOptions,
+                appendUnitOfMeasurement = true
+            )
+        ).build()
 
         val contentDescription = PlainComplicationText.Builder(getText(R.string.complication_entity_state_content_description)).build()
         val monochromaticImage = MonochromaticImage.Builder(Icon.createWithBitmap(iconBitmap)).build()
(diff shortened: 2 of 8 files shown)


wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
```diff
@@ -79,7 +79,14 @@ class EntityStateDataSourceService : SuspendingComplicationDataSourceService() {
...
             null
         }
-        val text = PlainComplicationText.Builder(entity.friendlyState(this, entityOptions)).build()
+
+        var stateText = entity.friendlyState(this, entityOptions)
+        if (entity.attributes.containsKey("unit_of_measurement")) {
+            stateText += " " + entity.attributes["unit_of_measurement"]
+        }
+
+        val text = PlainComplicationText.Builder(stateText).build()
```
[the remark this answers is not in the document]
slovdahl: For 2 maybe something like :point_down: Happy to get help with writing idiomatic Kotlin, I'm writing Kotlin for the first time in my life (using Java professionally for the past 10+ years though) . :grin:

```diff
diff --git common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
index 2a1bbc00..5ec0fd7e 100644
--- common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
+++ common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
@@ -652,6 +652,21 @@ fun <T> Entity<T>.friendlyState(context: Context, options: EntityRegistryOptions
     return friendlyState
 }
 
+fun <T> Entity<T>.friendlyStateWithOptionalUnit(
+    context: Context,
+    options: EntityRegistryOptions? = null
+): String {
+    val friendlyState = friendlyState(context, options)
+
+    val unit = (attributes as? Map<*, *>)?.get("unit_of_measurement")?.toString()
+
+    return if (unit?.isNotEmpty() == true) {
+        "$friendlyState $unit"
+    } else {
+        friendlyState
+    }
+}
+
 fun <T> Entity<T>.canSupportPrecision() = domain == "sensor" && state.toDoubleOrNull() != null
 
 fun <T> Entity<T>.isExecuting() = when (state) {
diff --git wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
index c2757f76..58287714 100644
--- wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
+++ wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
@@ -19,7 +19,7 @@ import dagger.hilt.android.AndroidEntryPoint
 import io.homeassistant.companion.android.common.R
 import io.homeassistant.companion.android.common.data.integration.canSupportPrecision
 import io.homeassistant.companion.android.common.data.integration.friendlyName
-import io.homeassistant.companion.android.common.data.integration.friendlyState
+import io.homeassistant.companion.android.common.data.integration.friendlyStateWithOptionalUnit
 import io.homeassistant.companion.android.common.data.integration.getIcon
 import io.homeassistant.companion.android.common.data.servers.ServerManager
 import io.homeassistant.companion.android.database.wear.EntityStateComplicationsDao
@@ -80,12 +80,12 @@ class EntityStateDataSourceService : SuspendingComplicationDataSourceService() {
             null
         }
 
-        var stateText = entity.friendlyState(this, entityOptions)
-        if (entity.attributes.containsKey("unit_of_measurement")) {
-            stateText += " " + entity.attributes["unit_of_measurement"]
-        }
-
-        val text = PlainComplicationText.Builder(stateText).build()
+        val text = PlainComplicationText.Builder(
+            entity.friendlyStateWithOptionalUnit(
+                this,
+                entityOptions
+            )
+        ).build()
 
         val contentDescription = PlainComplicationText.Builder(getText(R.string.complication_entity_state_content_description)).build()
         val monochromaticImage = MonochromaticImage.Builder(Icon.createWithBitmap(iconBitmap)).build()
```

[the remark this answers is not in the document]
jpelgrom: 1. I agree, as some units can be quite long.
2. It would probably be nice to add it to a function instead of only here, so it can be used elsewhere. Combined with 1 an argument for the existing function (defaulting to false) is probably the easiest, allowing you to pass in the stored value instead of switching between two functions.

common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -649,6 +649,15 @@ fun <T> Entity<T>.friendlyState(context: Context, options: EntityRegistryOptions
             }
         }
     }
+
+    if (appendUnitOfMeasurement) {
+        val unit = (attributes as? Map<*, *>)?.get("unit_of_measurement")?.toString()
```
slovdahl, line 654: Is this safe? Inspired by other usages of it in the `Entity` class.

If not, how to do it?

wear/src/main/java/io/homeassistant/companion/android/complications/views/ComplicationConfigMainView.kt
```diff
@@ -172,9 +196,11 @@ fun PreviewMainConfigView() {
     MainConfigView(
         entity = simplifiedEntity,
         showTitle = true,
+        showUnit = true,
```
slovdahl, line 199: Not 100% sure if it's the right thing to do, but I tried to make it enabled by default whenever a new complication is created, but it doesn't seem like this did the trick. Am I doing it wrong? Do we need to make it `true` all over the place in the code and add a post-migration query that sets it to `false` for existing complications?

wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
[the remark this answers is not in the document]
slovdahl: Both 1 and 2 have been fixed now, please take a new look whenever you have time.

wear/src/main/java/io/homeassistant/companion/android/complications/views/ComplicationConfigMainView.kt
jpelgrom: To enable it by default for new complications, simply change [the default of the state to `true`](https://github.com/home-assistant/android/pull/3703/files#diff-d8b3ef0c1255a9de6e1024cf3f0aee55bfc3b64cbf1775ba9e36701f38b5c2abR61).

common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
jpelgrom: Yes this is fine, the app doesn't have a typesafe alternative and as long as you're using known values from the map it shouldn't cause any issues.

common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -649,6 +649,15 @@ fun <T> Entity<T>.friendlyState(context: Context, options: EntityRegistryOptions
...
+
+        if (unit?.isNotEmpty() == true) {
```
jpelgrom, line 656: Because this is for a String, use `isNotBlank()` instead.

wear/src/main/java/io/homeassistant/companion/android/complications/EntityStateDataSourceService.kt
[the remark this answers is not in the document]
jpelgrom: Looks good!

wear/src/main/java/io/homeassistant/companion/android/complications/views/ComplicationConfigMainView.kt
slovdahl: Ah, thanks. So is this one only used for the preview in Android Studio?

jpelgrom: Yes - that's why function name includes Preview and it has a `@Preview` annotation ;)

slovdahl: Changed default state to `true` for new complications,

common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
slovdahl: Makes sense, fixed :+1:

jpelgrom (approved):
Code looks good and works as expected. Thanks for picking this up!

JBassett (approved):

PR #3716, opened 2023-07-26
Fix version code for F-Droid builds

jpelgrom:
## Summary

The new release model renamed `version_code.txt` to `version.txt`, which [broke F-Droid's update detection](https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/io.homeassistant.companion.android.minimal.yml?ref_type=heads#L416). We could tell them to update but this is easier as this file was specifically added for them.

## Screenshots

n/a

n/a


jpelgrom:
Actually this is not right, I confused version name with version code. Let me close this and generate the version code in another PR.

PR #3725, opened 2023-07-27
Add navigation entities to favorites and convert nav screen to grid

dshokouhi:
## Summary

- Allow navigation entities to be added to favorites
- Convert navigation screen from list to grid items

## Screenshots

image

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
@@ -49,8 +49,8 @@ class ManageAndroidAutoViewModel @Inject constructor(
                                 it.domain in MainVehicleScreen.SUPPORTED_DOMAINS ||
                                     (
                                         it.domain in MainVehicleScreen.MAP_DOMAINS &&
-                                            ((it.attributes as Map<*, *>)["latitude"] as Double? != null) &&
-                                            ((it.attributes as Map<*, *>)["longitude"] as Double? != null)
+                                            ((it.attributes as Map<*, *>)["latitude"] as? Double != null) &&
+                                            ((it.attributes as Map<*, *>)["longitude"] as? Double != null)
                                         )
                             }
                     } catch (e: Exception) {
--- a/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
@@ -49,8 +49,8 @@ class ManageAndroidAutoViewModel @Inject constructor(
                                 it.domain in MainVehicleScreen.SUPPORTED_DOMAINS ||
                                     (
                                         it.domain in MainVehicleScreen.MAP_DOMAINS &&
-                                            ((it.attributes as Map<*, *>)["latitude"] as? Double != null) &&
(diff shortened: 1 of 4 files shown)


app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
```diff
@@ -45,7 +45,14 @@ class ManageAndroidAutoViewModel @Inject constructor(
                 async {
                     entities[it.id] = try {
                         serverManager.integrationRepository(it.id).getEntities().orEmpty()
-                            .filter { it.domain in MainVehicleScreen.SUPPORTED_DOMAINS }
+                            .filter {
+                                it.domain in MainVehicleScreen.SUPPORTED_DOMAINS ||
+                                    (
+                                        it.domain in MainVehicleScreen.MAP_DOMAINS &&
+                                            ((it.attributes as Map<*, *>)["latitude"] as Double? != null) &&
```
jpelgrom, line 52: These checks should use `as? Double` to be consistent with the car service implementation and prevent a crash in case `latitude` exists but it isn't a number.

app/src/main/java/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
```diff
@@ -160,8 +163,19 @@ class EntityGridVehicleScreen(
                 gridItem
                     .setOnClickListener {
                         Log.i(TAG, "${entity.entityId} clicked")
-                        lifecycleScope.launch {
-                            entity.onPressed(integrationRepository)
+                        if (entity.domain in MainVehicleScreen.MAP_DOMAINS) {
+                            val attrs = entity.attributes as Map<*, *>
+                            val lat = attrs["latitude"] as Double
```
jpelgrom, line 168: This attribute may no longer exist (removed after adding the favorite), add a check to handle null.

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
dshokouhi: done thanks for calling that out :)

app/src/main/java/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
dshokouhi: thanks for calling that out, made the check more consistent  here too

app/src/main/java/io/homeassistant/companion/android/settings/vehicle/ManageAndroidAutoViewModel.kt
```diff
@@ -45,7 +45,14 @@ class ManageAndroidAutoViewModel @Inject constructor(
                 async {
                     entities[it.id] = try {
                         serverManager.integrationRepository(it.id).getEntities().orEmpty()
-                            .filter { it.domain in MainVehicleScreen.SUPPORTED_DOMAINS }
+                            .filter {
+                                it.domain in MainVehicleScreen.SUPPORTED_DOMAINS ||
+                                    (
+                                        it.domain in MainVehicleScreen.MAP_DOMAINS &&
+                                            ((it.attributes as Map<*, *>)["latitude"] as? Double != null) &&
```
jpelgrom, line 52: Another consistency thing: attributes should also use `as? Map<*, *>`

jpelgrom (approved):

JBassett (approved):

PR #3726, opened 2023-07-27
Update sensors when driving interface is open or resumed

dshokouhi:
## Summary

Update sensors when the driving interface is open/resumed.

There really is no `onResume` method so need to use a separate launch method with the lifecycle set to resumed. Attempted to use this when the screen is created but that only gets called once and not on resume. Also attempting to move the method to `onGetTemplates` triggers the intent twice so this seemed the best fit.

This also matches existing behavior in the main/wear app

This should help users see updates from car sensors too 🤞 (for whoever they actually work for 😂 )


jpelgrom:
Shouldn't this use `SensorReceiver.updateAllSensors(context)` instead of `SensorUpdateReceiver.updateSensors(context)`?

dshokouhi:
> Shouldn't this use `SensorReceiver.updateAllSensors(context)` instead of `SensorUpdateReceiver.updateSensors(context)`?

too many update commands 🙈

jpelgrom:
I'm not completely sure what the difference is but the `updateAllSensors` one is used for main/wear at least

dshokouhi:
> I'm not completely sure what the difference is but the `updateAllSensors` one is used for main/wear at least

its because we couldnt use `SensorReceiver` inside of `common`

JBassett (approved):

PR #3728, opened 2023-07-28
Fix old server references in app-level preferences

jpelgrom:
## Summary

While testing the navigation entities in Android Auto PR I noticed the 'All entities' button showing up while also showing the domains list. It turns out this was because my preference still held some favorites, but they were for servers that had been deleted.

image

_(icon cut off at the bottom is 'All entities')_

When deleting a server most of the server-specific stuff is already deleted, but app-level preferences that reference a server aren't updated. This PR fixes that and adds a function to clean up any server references there might be in the app-level preferences.

## Screenshots

n/a / above screenshot will change but that is because it fixes the described bug

n/a


dshokouhi (approved):
thanks for the fix :)

JBassett (approved):

PR #3731, opened 2023-07-28
Hide hidden entities from android auto domains

### LINKED ISSUE #3250: Android Auto Beta - Hidden Entities Appear in AA Home Assistant App UI

dshokouhi:
## Summary

New attempt of #3678 to fix #3250


app/src/main/java/io/homeassistant/companion/android/util/vehicle/GridItems.kt
```diff
@@ -132,40 +138,57 @@ fun getDomainList(
...
-                            allEntities
-                        ) { }
-                    )
+        }
+        var domainIsEmpty = false
+        lifecycleScope.launch {
+            entityList.collect {
+                domainIsEmpty = it.isEmpty()
+            }
+        }
```
jpelgrom, line 154: Why collect here, as this doesn't invalidate the UI and cause an update?

dshokouhi: the entities are actually collected previously in the `init {}` here we are just building a small list to see if the domain should be hidden the last time `invalidate()` was called elsewhere.

jpelgrom: But this doesn't 'block' the function and it will continue with the initial `false` value / `launch` means the `domainIsEmpty` variable is probably only updated after `onGetTemplate` is already done?

dshokouhi: the code as is actually works. I did not have any `input_button` entities so I created one and saw the domain category, then I hid the entity and upon reloading the page the domain was hidden as well

jpelgrom: OK this works but I don't understand why. Even tried removing live state updates yet it still kept working.

jpelgrom:
Should this also apply to the list of entities in the navigation screen?

dshokouhi:
> Should this also apply to the list of entities in the navigation screen?

thats a good question. We do already hide device trackers that are home so I think the list should be sane. You dont normally hide a zone or person entity so maybe we can wait to see if a user complains?

jpelgrom:
> > Should this also apply to the list of entities in the navigation screen?
> 
> thats a good question. We do already hide device trackers that are home so I think the list should be sane. You dont normally hide a zone or person entity so maybe we can wait to see if a user complains?

The list also includes sensors with lat/long attributes which users might hide. Good to wait for feedback, as long as it's been thought about it's fine by me.

JBassett (approved):

PR #3732, opened 2023-07-28
Allow binary sensors and sensors to be added to driving favorites

### LINKED ISSUE #3248: Android Auto sensors display

I'm after my first tests of Android Auto support. I looks really promising.
Would it be possible to add there sensors display in addition to all the currently available domains?
Posiibility of favourites selection would be needed for that as well, because of the possible number of sensor entieties.

**Additional context**

dshokouhi:
## Summary

Fixes #3248 by allow `binary_sensor` and `sensor` to be added to the favorites list. Since there is a potential for majority of users to have more entities than their vehicle limit will allow I opted not to add a new domain category to the main supported list.

When a non actionable entity is added to the grid we are going to skip the click listener so they cannot be selected

Documentation: home-assistant/companion.home-assistant#967


app/src/main/java/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
```diff
@@ -160,28 +160,40 @@ class EntityGridVehicleScreen(
...
-                            if (attrs != null) {
-                                val lat = attrs["latitude"] as? Double
-                                val lon = attrs["longitude"] as? Double
-                                if (lat != null && lon != null) {
-                                    val intent = Intent(
-                                        CarContext.ACTION_NAVIGATE,
-                                        Uri.parse("geo:$lat,$lon")
-                                    )
-                                    carContext.startCarApp(intent)
+                if (entity.domain !in MainVehicleScreen.NOT_ACTIONABLE_DOMAINS) {
```
jpelgrom, line 163: Doesn't this break support for navigation to sensors with lat/lon attributes?

dshokouhi: good catch, added a check to not skip them

app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainCheck.kt
```diff
@@ -0,0 +1,19 @@
...
+import io.homeassistant.companion.android.vehicle.MainVehicleScreen
+
+@RequiresApi(Build.VERSION_CODES.O)
+fun isVehicleDomain(entity: Entity<*>): Boolean {
+    return entity.domain in MainVehicleScreen.SUPPORTED_DOMAINS ||
+        entity.domain in MainVehicleScreen.NOT_ACTIONABLE_DOMAINS ||
+        (entity.domain in MainVehicleScreen.MAP_DOMAINS && canNavigate(entity))
+}
+
+fun canNavigate(entity: Entity<*>): Boolean {
```
jpelgrom, line 16: Missing check: `domain == "sensor"`?

app/src/main/java/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
```diff
@@ -160,28 +161,40 @@ class EntityGridVehicleScreen(
...
+                                }
+
+                                in MainVehicleScreen.SUPPORTED_DOMAINS -> {
+                                    lifecycleScope.launch {
+                                        entity.onPressed(integrationRepository)
+                                    }
+                                }
+
+                                else -> {
+                                    // No op
```
jpelgrom, line 191: Should it be more obvious that there is no tap action for this item, for example by setting the item to be disabled?

app/src/main/java/io/homeassistant/companion/android/vehicle/MainVehicleScreen.kt
```diff
@@ -62,6 +62,11 @@ class MainVehicleScreen(
             "sensor",
             "zone"
         )
+
+        val NOT_ACTIONABLE_DOMAINS = listOf(
```
jpelgrom, line 66: As you created a `DomainCheck` file, maybe it is a good idea to make that a vehicle supported entities file and move this constant and the other existing ones above to it so everything is in one place?

app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainCheck.kt
```diff
@@ -0,0 +1,19 @@
+package io.homeassistant.companion.android.util.vehicle
+
+import android.os.Build
+import androidx.annotation.RequiresApi
+import io.homeassistant.companion.android.common.data.integration.Entity
+import io.homeassistant.companion.android.common.data.integration.domain
+import io.homeassistant.companion.android.vehicle.MainVehicleScreen
+
+@RequiresApi(Build.VERSION_CODES.O)
```
jpelgrom, line 9: This annotation is incorrect, nothing in here requires O.

jpelgrom (commented):
While I understand that users want to have sensors visible, I'm still not sure this is the right approach. If the question "is [sensor] [sensor state]?" is answered with a non-desirable answer, how would you act on it? IMO a helper/template entity based on the sensor that allows you to do said action is better. Google [seems to agree](https://developers.google.com/cars/design/create-apps/apps-for-drivers/templates/grid-template) that having something you can interact with is better:
> SHOULD | Have an action associated with each grid item (information-only items are not recommended).

app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainCheck.kt
dshokouhi: This check applies to any domain that needs to be navigable

app/src/main/java/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
dshokouhi: There is no disabled option for GridItem

app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainCheck.kt
dshokouhi: The domain variable required it

jpelgrom: Hmm I think it may be because MainVehicleScreen is annotated? `domain` definitely isn't. Also see [this suggestion](https://github.com/home-assistant/android/pull/3732/files/5da46d30710dbf14354cbfe2c058bab6f99be25d#r1278292919) which could resolve it.

app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainCheck.kt
jpelgrom: But is any domain other than sensor expected? Otherwise this also needs a docs update to reflect that it can be any entity with these attributes.

app/src/main/java/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
jpelgrom: Odd, but I guess that really shows Google doesn't intend you to have non-actionable grid items.

dshokouhi: yea to be fair I have seen an app that adds an information only GridItem which was the EV map for the car sensor testing.

app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainCheck.kt
dshokouhi: its not any domain with a sensor, only the map domains. Let me make the call more clear :)

jpelgrom: This function isn't checking for map domains though, only attributes

dshokouhi:
> While I understand that users want to have sensors visible, I'm still not sure this is the right approach. If the question "is [sensor] [sensor state]?" is answered with a non-desirable answer, how would you act on it? IMO a helper/template entity based on the sensor that allows you to do said action is better. Google [seems to agree](https://developers.google.com/cars/design/create-apps/apps-for-drivers/templates/grid-template) that having something you can interact with is better:
> 
> > SHOULD | Have an action associated with each grid item (information-only items are not recommended).

So there is one issue regarding helper/template entity. None of the supported domains today will allow for states seen by `sensor`, now `binary_sensor` can be put into a template/helper easily but you can't do that with a temperature sensor or any other type of different data. A sensor can for example let someone show the state of a `number` or `select` entity. So I do see some value there. The items are also in a grid so users can be a little unique in their designs and make rows with visual data.

app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainCheck.kt
dshokouhi: this check now considers the domain as well :)

app/src/main/java/io/homeassistant/companion/android/vehicle/MainVehicleScreen.kt
dshokouhi: done, I opted to keep the file the same as everything is still related to the `domain`

app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainCheck.kt
dshokouhi: Agreed, its now moved :)

jpelgrom:
> don't think its needed as these are non-actionable domains and we are currently not listing which domains favorites supports

It might be worth a comment in the documentation because sensors are _only_ supported using favorites.

dshokouhi:
> > don't think its needed as these are non-actionable domains and we are currently not listing which domains favorites supports
> 
> It might be worth a comment in the documentation because sensors are _only_ supported using favorites.

docs added :)

jpelgrom:
This seems to be working mostly fine except that state changes are not reflected until restarting Auto (not limited to sensors, any favorite entity) - do you want to address that here or in another PR?

app/src/main/java/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
jpelgrom: Looks like Auto already handles some of this for us: no highlight when tapped (touch) / greyed out highlight when selected (rotary input).

dshokouhi:
> This seems to be working mostly fine except that state changes are not reflected until restarting Auto (not limited to sensors, any favorite entity) - do you want to address that here or in another PR?

im actually trying to fix that in a different branch since that issue was introduced previously, admittedly I am stuck on a proper working solution 🤔

app/src/main/java/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
dshokouhi: yea they recommend not setting a click listener in these places which is why I opted to skip adding it for these entities

jpelgrom (approved):
Let's add this and see what the user feedback is (and Google feedback...), hard to deny it is a popular request and we should do something to make it easier.

JBassett (approved):

PR #3736, opened 2023-07-29
Invalidate Auto home on favorite state change

jpelgrom:
## Summary

Fixes favorites entity state not updating in the interface for Android Auto when their state changes, a bug [first mentioned here](https://github.com/home-assistant/android/pull/3732#issuecomment-1656804554). This is because the home screen only invalidated when domains change.

Discussed with @dshokouhi via DMs. `favoriteEntities` isn't used and also complicates things so merge/remove it with the other variable for entities of favorites.

## Screenshots

n/a

n/a


dshokouhi (approved):
thanks for the fix ❤️

JBassett (approved):

PR #3737, opened 2023-07-29
Initial work on migrating plugins to catalogs.

JBassett:
## Summary

Migrate most plugins to catalogs.  Still having issues with the stuff in `settings.gradle.kts` since it's run before the catalog is actually created.

## Screenshots

N/A

N/A


settings.gradle.kts
```diff
@@ -4,7 +4,15 @@ include(":common", ":app", ":wear", ":automotive")
...
 
+pluginManagement {
+    repositories {
+        gradlePluginPortal()
+        google()
+    }
+}
+
 plugins {
+    // So we can't reach the libs.plugins.* aliases from here so we need to declare them the old way...
```
dshokouhi, line 15: I cant remember the last time I saw these change anyways 😂

JBassett:
I tried a couple different ways to configure reckon but didn't find a way to make it work.  For now I guess we can leave it.  It's the only one currently like it now.

common/build.gradle.kts
```diff
@@ -18,7 +17,7 @@ android {
     compileSdk = 33
```
dshokouhi, line 17: Should this also use the same compile SDK variable?

`libs.versions.androidSdk.compile.get().toInt()`

wear/build.gradle.kts
```diff
@@ -16,7 +15,7 @@ android {
     defaultConfig {
         applicationId = "io.homeassistant.companion.android"
         minSdk = 26
-        targetSdk = 32
+        targetSdk = libs.versions.androidSdk.target.get().toInt()
```
dshokouhi, line 18: this is not correct for Wear OS as we do not handle notification permissions yet

jpelgrom: Another reason to have it be a different variable: we might not even be able to target the same version as the main app because it doesn't exist - the main app will probably target Android 14 later this year which doesn't exist for watches.

dshokouhi: yes thats a good point, Wear OS 4 comes out in October (or later this year) which is based on Android 13 so 1 version behind (at least)

jpelgrom: (Off-topic but the Galaxy Watch 6 with Wear OS 4 is actually releasing in August)

dshokouhi (approved):
looks good to me :) nice to have everything configured in one file now!

PR #3740, opened 2023-07-30
Hide some sensors from unsupported devices

dshokouhi:
## Summary

Hide some sensors that simply won't work in Automotive. Also hide battery manager if the device has no battery.

As these devices do not support the sensor they also wouldn't support for any registered intent so should be safe to continue hiding.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
@@ -4,6 +4,7 @@ import android.Manifest
 import android.app.UiModeManager
 import android.content.ComponentName
 import android.content.Context
+import android.content.pm.PackageManager
 import android.content.res.Configuration
 import android.media.MediaMetadata
 import android.media.session.MediaSessionManager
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
```diff
@@ -70,7 +71,11 @@ class NotificationSensorManager : NotificationListenerService(), SensorManager {
     }
     override fun hasSensor(context: Context): Boolean {
         val uiManager = context.getSystemService<UiModeManager>()
```
jpelgrom, line 73: You can move this down into the `if`, we don't need the UI manager if this is Automotive

jpelgrom (commented):
I don't know, always returning `in_vehicle` for the activity sensor when using Automotive also would be an option ;)

app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
dshokouhi: thanks :)

jpelgrom (approved):
Good to improve the Automotive experience

JBassett (approved):

PR #3749, opened 2023-08-01
Don't create release number/notes for PRs

jpelgrom:
## Summary

Don't create a release number and release notes for PRs
 - release number is unused
 - release notes requires a token with write access which should not be given to forks and without it the action will fail, so remove step

## Screenshots

n/a

n/a

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -94,21 +94,11 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-gradle-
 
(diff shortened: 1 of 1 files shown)


jpelgrom:
Yes this is ironic after a PR to update the release notes was just merged and wouldn't have been possible without the release notes step, but I realized this will break the CI for external PRs (which is probably close to 99%) only when it was merged.

JBassett (approved):

PR #3764, opened 2023-08-07
Set car sensors state appropriately if sensor registers with unit of measurement

### LINKED ISSUE #3762: HA error due car sensor not numeric

**Android version(s):**
13

**Device model(s):**
Pixel 4a

**Home Assistant version:**
2023.8.1

Using the 2023.7.8-full version of app and the car sensors. Noticed this in the HA log:

```
2023-08-05 10:32:49.611 ERROR (MainThread) [homeassistant.components.sensor] Error adding entities for domain sensor with platform mobile_app

Traceback (most recent call last):
File "/usr/src/homeassistant/homeassistant/components/sensor/__init__.py", line 593, in state
numerical_value = float(value) # type:ignore[arg-type]

^^^^^^^^^^^^

ValueError: could not convert string to float: 'Open Home Assistant app to activate the sensor'
The above exception was the direct cause of the following exception:

Traceback (most recent call last):

File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 510, in async_add_entities
await asyncio.gather(*tasks)
File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 750, in _async_add_entity
await entity.add_to_platform_finish()
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1004, in add_to_platform_finish
self.async_write_ha_state()
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 746, in async_write_ha_state
self._async_write_ha_state()
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 808, in _async_write_ha_state
state = self._stringify_state(available)

^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 752, in _stringify_state
if (state := self.state) is None:

^^^^^^^^^^

File "/usr/src/homeassistant/homeassistant/components/sensor/__init__.py", line 595, in state
raise ValueError(
ValueError: Sensor sensor.gavin_pixel_4a_car_battery has device class 'battery', state class 'None' unit '%' and suggested precision 'None' thus indicating it has a numeric value; however, it has the non-numeric value: 'Open Home Assistant app to activate the sensor' (<class 'str'>)
```

**Companion App Logs:**

```

```

**Additional information:**

dshokouhi:
## Summary

Fixes #3762 

When a sensor has a unit of measurement HA will expect certain states if non-numerical otherwise the sensor will not be added.

By default when we do not have `carInfo` the state will be `unavailable`. After the app is open user will either see the real state or `unknown` if the status is not `success`.  All sensors will show the `status` attribute which will show  the status of the sensor.

As the sensors get triggered by a listener we can safely call `forceUpdate = true` we either get a real state or we are getting a real status of the value to send to the user.

Removed the old string because we actually should not translate the states that we send, missed this in the original review 😬

Documentation: home-assistant/companion.home-assistant#969

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
@@ -25,7 +25,6 @@ class CarSensorManager :
 
     companion object {
         internal const val TAG = "CarSM"
-        private const val OPEN_APP = "Open Home Assistant app to activate the sensor"
 
         private val fuelLevel = SensorManager.BasicSensor(
             "car_fuel",
@@ -174,11 +173,9 @@ class CarSensorManager :
                         onSensorUpdated(
                             context,
                             it,
-                            if (it.unitOfMeasurement.isNullOrEmpty()) OPEN_APP else "unknown",
+                            "unavailable",
                             it.statelessIcon,
-                            mapOf(
-                                "status" to OPEN_APP
-                            )
+                            mapOf()
                         )
                     }
                 }
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
@@ -25,6 +25,7 @@ class CarSensorManager :
 
     companion object {
         internal const val TAG = "CarSM"
+        private const val OPEN_APP = "Open Home Assistant app to activate the sensor"
 
         private val fuelLevel = SensorManager.BasicSensor(
             "car_fuel",
@@ -173,9 +174,11 @@ class CarSensorManager :
                         onSensorUpdated(
                             context,
                             it,
-                            context.getString(R.string.car_data_unavailable),
+                            if (it.unitOfMeasurement.isNullOrEmpty()) OPEN_APP else "unknown",
                             it.statelessIcon,
-                            mapOf()
+                            mapOf(
+                                "status" to OPEN_APP
+                            )
                         )
                     }
                 }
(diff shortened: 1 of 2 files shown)


jpelgrom:
> So if the sensor has a unit of measurement we will default to `unknown` otherwise if the app was not open we will show the same state

Wouldn't `unavailable` be a better state when the car info isn't available? Different state when car info is not available depending on whether or not a unit of measurement is set is also (too) inconsistent (in my opinion).

> we actually should not translate the states that we send

Not sure I agree for a string like this, especially considering core/frontend focus recently on making sure everything can be translated.

dshokouhi:
> Not sure I agree for a string like this, especially considering core/frontend focus recently on making sure everything can be translated.

We can't translate state strings in the app they need to be translated in core for the integration. I think there might be an old issue where we discussed this before.

jpelgrom:
Yes so I would avoid strings that need translation like these :)

dshokouhi:
> Yes so I would avoid strings that need translation like these :)

Lol we don't need to avoid them. We are fine with using English. Someone who is skilled enough needs to add support so the states can be translated. iOS has the same issue.

 We are actually supposed to keep everything in English because the companion app docs are in English. It makes it easy to find sensors and other things we mention.

dshokouhi:
> > We are actually supposed to keep everything in English because the companion app docs are in English.
> 
> Hadn't heard that one before but considering the limitations it makes sense.
 
can't seem to find it now but I remember Paulus telling us what we needed to do if we wanted custom states translated which involved updating `mobile_app` integration then as we add more strings we update the integration.

> Note about unavailable / inconsistency is still relevant I think.

agreed, its fixed and should be more consistent now.

1. App not open - `unavailable`
2. App open but no status not success - `unknown`
3. App open and status success - state

dshokouhi:
> Now that it is easy to follow, should it maybe be mentioned in the docs that `unavailable` -> Auto not active?

Agreed, updated docs to mention both `unavailable` and `unknown` states so users can tell if a sensor is implemented or not :)

jpelgrom (approved):
Good to get this in before a release so a breaking change is prevented :)

JBassett (approved):

Issue #3710, opened 2023-07-26
Controls for alarm domain in Android Auto


sparkydave1981:
**Is your feature request related to a problem? Please describe.**

No problem, just no current ability to control the alarm domain in Android Auto
**Describe the solution you'd like**

Ability to control the `alarm_control_panel` domain from Android Auto HA app
**Describe alternatives you've considered, if any**

Currently planning to create a template sensor to display current status of alarm, then scripts for arming and disarming.
**Additional context**

https://community.home-assistant.io/t/home-assistant-coming-for-your-car/522288/189?u=sparkydave

dshokouhi:
This might be a domain where we can switch between `disarmed` and `armed_away` I think but I am not sure how that will work when there is a code for `alarm_control_panel` that will need some thought given to it. Personally speaking I don't use a code for my own `alarm_control_panel` but still need to consider it.

dshokouhi:
This request will require some extra thought as we do indeed need to provide the alarm code if a code is required. Unfortunately we cannot prompt the user to enter the code, this will need to be stored app side. This will be a larger change than originally thought.

sparkydave1981:
It would be nice if it can at least display the current state of `alarm_control_panel`, then we can use scripts to arm/disarm.

sparkydave1981:
Something I just tested but doesn't work: I created a template sensor to show the status of the `alarm_control_panel.house` entity however the Companion app won't let me add anything from the `sensor:` domain as  a n Android Auto favourite.

W would need to have `sensor:` allowed in order for the work-around to be feasible (using a sensor to display current status and scripts to arm/disarm)

jpelgrom:
> the Companion app won't let me add anything from the `sensor:` domain as a n Android Auto favourite.

Can you check your app version? This requires the latest beta release to work (>= 2023.8.0, may not yet be available from the Play Store).

sparkydave1981:
> an you check your app version? This requires the latest beta release to work

It seems my phone installed an update overnight and now I can select sensors.  Thanks!

dshokouhi:
> > an you check your app version? This requires the latest beta release to work
> 
> It seems my phone installed an update overnight and now I can select sensors. Thanks!

next beta will include support for alarm control panels without needing to create a sensor :) only alarms without a code are actionable

pergolafabio:
[...]

PR #3765, opened 2023-08-07
Add support for alarm control panels in android auto

dshokouhi:
## Summary

Fixes: #3710 

We will show Alarm Control Panels as a supported domain because they are important enough IMO.

Only if we detect a code is not required will a user be able to click on it, otherwise it will be non-actionable.

Given the use case we will either `disarm` or `arm_away` the alarm, the other arming options don't feel as if they would be used as much as `armed_away`

Used the same icons from the frontend: https://github.com/home-assistant/frontend/blob/858dd738684b604b7861ccb9e2c449f9280f8723/src/common/entity/alarm_panel_icon.ts#L15-L36

Documentation: home-assistant/companion.home-assistant#968

## Any other notes

After going on the first road trip this weekend the question of "did we turn on the alarm" came while in the car and it would've been easy to answer that 😂

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainChecks.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainChecks.kt
@@ -3,7 +3,7 @@ package io.homeassistant.companion.android.util.vehicle
 import io.homeassistant.companion.android.common.R
 import io.homeassistant.companion.android.common.data.integration.Entity
 import io.homeassistant.companion.android.common.data.integration.domain
-import io.homeassistant.companion.android.common.data.integration.supportsArmedAway
+import io.homeassistant.companion.android.common.data.integration.supportsAlarmControlPanelArmAway
 
 val SUPPORTED_DOMAINS_WITH_STRING = mapOf(
     "alarm_control_panel" to R.string.alarm_control_panels,
@@ -49,5 +49,5 @@ fun canNavigate(entity: Entity<*>): Boolean {
 fun alarmHasNoCode(entity: Entity<*>): Boolean {
     return entity.domain == "alarm_control_panel" &&
         (entity.attributes as? Map<*, *>)?.get("code_format") as? String == null &&
-        entity.supportsArmedAway()
+        entity.supportsAlarmControlPanelArmAway()
 }
--- a/app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainChecks.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainChecks.kt
@@ -3,6 +3,7 @@ package io.homeassistant.companion.android.util.vehicle
 import io.homeassistant.companion.android.common.R
 import io.homeassistant.companion.android.common.data.integration.Entity
 import io.homeassistant.companion.android.common.data.integration.domain
+import io.homeassistant.companion.android.common.data.integration.supportsArmedAway
 
 val SUPPORTED_DOMAINS_WITH_STRING = mapOf(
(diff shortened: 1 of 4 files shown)


app/src/main/java/io/homeassistant/companion/android/util/vehicle/DomainChecks.kt
```diff
@@ -42,3 +44,7 @@ fun canNavigate(entity: Entity<*>): Boolean {
             ((entity.attributes as? Map<*, *>)?.get("longitude") as? Double != null)
         )
 }
+
+fun alarmHasNoCode(entity: Entity<*>): Boolean {
+    return (entity.attributes as? Map<*, *>)?.get("code_format") as? String == null
```
jpelgrom, line 49: Shouldn't this include a check for the domain, in case there are any other entities with a `code_format` attribute out there?

common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -41,6 +41,7 @@ object EntityExt {
     val LIGHT_MODE_NO_BRIGHTNESS_SUPPORT = listOf("unknown", "onoff")
     const val LIGHT_SUPPORT_BRIGHTNESS_DEPR = 1
     const val LIGHT_SUPPORT_COLOR_TEMP_DEPR = 2
+    const val ALARM_CONTROL_PANEL_SUPPORT_ARMED_AWAY = 2
```
jpelgrom, line 44: The HA constant is for the feature/call, not the state: `ALARM_CONTROL_PANEL_SUPPORT_ARM_AWAY`

common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -114,6 +115,16 @@ fun <T> Entity<T>.getCoverPosition(): EntityPosition? {
     }
 }
 
+fun <T> Entity<T>.supportsArmedAway(): Boolean {
```
jpelgrom, line 118: For consistency with other feature checks:

```suggestion
fun <T> Entity<T>.supportsAlarmControlPanelArmAway(): Boolean {
```

jpelgrom:
One more thing: the `friendlyState` function should be updated to support the alarm specific states.

dshokouhi:
> One more thing: the `friendlyState` function should be updated to support the alarm specific states.

done :)

common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -631,15 +631,22 @@ val <T> Entity<T>.friendlyName: String
 
 fun <T> Entity<T>.friendlyState(context: Context, options: EntityRegistryOptions? = null, appendUnitOfMeasurement: Boolean = false): String {
     var friendlyState = when (state) {
+        "armed_away" -> context.getString(commonR.string.state_armed_away)
```
jpelgrom, line 634: Missing states: [`pending`, `arming`, `disarming`](https://github.com/home-assistant/core/blob/dev/homeassistant/components/alarm_control_panel/strings.json#L32-L44) (those last two should probably be added to the executing states as well)

dshokouhi: ah good catch!

jpelgrom (approved):
Code looks good to me, but I'm not able to test this live

dshokouhi:
> Code looks good to me, but I'm not able to test this live

for testing I had 2 panels up. 1 with a code and the other without a code. I could only test using the Manual alarm control panel

JBassett (approved):

PR #3767, opened 2023-08-07
Hide all entities when we dont have domains to show, show empty message just in case

dshokouhi:
## Summary

While testing another PR came across an issue where if user had multi server and a server was down we would show all entities and the domain list would never load as a result. So handle loading domains like the other screen, add some checks to only show all entities if we can and add an empty message just in case.


jpelgrom (approved):
Tested and working as expected now!

JBassett (approved):

PR #3697, opened 2023-07-22
Implement support for multiple Shortcut Tiles on Wear OS

marazmarci:
Closes #2475

## Summary
Implemented support for multiple Shortcut Tiles in the Wear OS app, so now it supports ~~3x~~ any number of separately configurable Tiles (only limited by Wear OS's current limit of 10 Tiles). ~~If we decide there should be more Tiles, it can be very easily changed.~~

There's a new submenu in the settings, where the user has to first select which Shortcut Tile they want to manage. The change is backward compatible, so if a user already had a Shortcut Tile set up, it will show up as the first Tile in the list.

### ToDo
- [x] update documentation

## Screenshots
image
image

## Link to pull request in Documentation repository

Documentation: https://github.com/home-assistant/companion.home-assistant/pull/966

## Any other notes
I marked it as a draft, because I'm not sure about the wording in `strings.xml`, the translations, and the documentation, but I won't have too much more time to work on this for some days, but I wanted to gather feedback in the meantime.

PR #3783, opened 2023-08-11
Support multiple Template tiles on Wear OS

slovdahl:
## Summary

Support any number of Template tiles in the Wear OS app.

## Screenshots

TODO


common/src/main/java/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepositoryImpl.kt
```diff
@@ -51,6 +53,26 @@ class WearPrefsRepositoryImpl @Inject constructor(
...
+                    )
+
+                    localStorage.putString(PREF_TILE_TEMPLATES, JSONObject(templates).toString())
+                }
+
+                localStorage.remove(legacyPrefTileTemplate)
+                localStorage.remove(legacyPrefTileTemplateRefreshInterval)
+
+                localStorage.putInt(MIGRATION_PREF, MIGRATION_VERSION)
+            }
```
slovdahl, line 75: Not fully convinced whether this is the right way forward yet. Having a single entry for all the templates and their respective refresh intervals feels somewhat logical though.

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearHomeView.kt
```diff
@@ -48,30 +50,42 @@ fun LoadSettingsHomeView(
...
+                    SettingsWearTemplateTile(
+                        template = it.first,
+                        // TODO: use rendered template
+                        renderedTemplate = "",
+                        refreshInterval = it.second,
+                        onContentChanged = {},
+                        onRefreshIntervalChanged = {}
+                    ) {}
+                }
+            }
```
slovdahl, line 88: Thoughts about this part?

My idea was to change the current "Template tile" row to "Templates tiles". When clicked, it opens a list of all current Template tiles and a button for adding a new one. When a Template tile is clicked on in the list, the current `SettingsWearTemplateTile` is opened.

app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearViewModel.kt
```diff
@@ -280,6 +282,7 @@ class SettingsWearViewModel @Inject constructor(
         favoriteEntityIdList.forEach { entityId ->
             favoriteEntityIds.add(entityId)
         }
+        // TODO: forward and backward compatibility?
         setTemplateContent(data.getString(WearDataMessages.CONFIG_TEMPLATE_TILE, ""))
         templateTileRefreshInterval.value = data.getInt(WearDataMessages.CONFIG_TEMPLATE_TILE_REFRESH_INTERVAL, 0)
```
slovdahl, line 287: Please help me out here, what do we need to support?
- Old mobile app + new Wear OS app?
- New mobile app + old Wear OS app?
- Or can we always assume that both apps are of the same version?

slovdahl (commented):
Very very early draft, the code is not even compiling right now. But happy to get any kind of feedback anyway :slightly_smiling_face:

app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearViewModel.kt
jpelgrom: We generally assume the versions match and don't try to keep compatibility with older app combinations.

dshokouhi: play store is updated with both required versions at the same time too so its a safe assumption there :)

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearHomeView.kt
dshokouhi: how would it work to add a template tile here? Dont we need a tile ID to apply the template to it?

app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearViewModel.kt
slovdahl: Ah, great :+1:

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepository.kt
```diff
@@ -7,10 +7,10 @@ interface WearPrefsRepository {
...
     suspend fun getShowShortcutText(): Boolean
     suspend fun setShowShortcutTextEnabled(enabled: Boolean)
-    suspend fun getTemplateTileContent(): String
-    suspend fun setTemplateTileContent(content: String)
-    suspend fun getTemplateTileRefreshInterval(): Int
-    suspend fun setTemplateTileRefreshInterval(interval: Int)
+    suspend fun getAllTemplateTiles(): Map<Int?, Pair<String, Int>>
+    suspend fun getTemplateTile(tileId: Int?): Pair<String, Int>?
+    suspend fun setTemplateTile(tileId: Int?, content: String, refreshInterval: Int)
+    suspend fun removeTemplateTile(tileId: Int?): Pair<String, Int>?
```
slovdahl, line 13: Initially I wanted to expose the `Pair<String, Int>` as a proper data class, but I'm not sure where to put it. It needs to be in `common` because this interface is in `common`, but I couldn't find a good place to put it in. Hence I went with an internal data class in `WearPrefsRepositoryImpl` and exposing the data as `Pair`.

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepositoryImpl.kt
jpelgrom: Why are you adding a migration here _and_ doing the runCatching/fold thing in the individual preference? Having it in one place should be enough (and I think a migration here is nicest, so you can avoid complicated code in the individual preference functions).

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepositoryImpl.kt
```diff
@@ -51,6 +53,26 @@ class WearPrefsRepositoryImpl @Inject constructor(
 
                 localStorage.putInt(MIGRATION_PREF, MIGRATION_VERSION)
             }
+
+            if (currentVersion == 1) {
```
jpelgrom, line 57: This should be `< 2` to catch anyone that wasn't already migrated to 1 or future changes introduce another new version.

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepository.kt
jpelgrom: You can clarify it by adding documentation to functions, for example:

```kotlin
/** @return [Pair] containing the template content and the refresh interval, or `null` if it doesn't exist */
suspend fun getTemplateTile(tileId: Int?): Pair<String, Int>?
```

Or you could create an `impl.entities` package like authentication/integration/websocket are doing and add it there.

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepositoryImpl.kt
slovdahl: Very good catch :+1: In my head, `currentVersion` was updated to `1` here if the previous block had been taken.. But that is of course not the case. Maybe the name of the variable mislead me :thinking:

Should we update the value of `currentVersion` at the end of each migration instead?

slovdahl: Alternatively rename the variable to something else (e.g. `versionBeforeMigrations`).

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepository.kt
slovdahl: > Or you could create an `impl.entities` package like authentication/integration/websocket are doing and add it there.

Ah, thanks, that could work :+1: Being able to use the real field names instead of `first` and `second` makes the code so much easier to read IMO, so I'll go with that.

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepositoryImpl.kt
slovdahl: Good question! The migration should indeed be enough. I started out with a similar approach as #3697 (hence  `runCatching` etc), but I cleaned it up now. Please take a look.

wear/src/main/java/io/homeassistant/companion/android/phone/PhoneSettingsListener.kt
```diff
@@ -216,7 +216,8 @@ class PhoneSettingsListener : WearableListenerService(), DataClient.OnDataChange
     private fun saveTileTemplate(dataMap: DataMap) = mainScope.launch {
         val content = dataMap.getString(WearDataMessages.CONFIG_TEMPLATE_TILE, "")
         val interval = dataMap.getInt(WearDataMessages.CONFIG_TEMPLATE_TILE_REFRESH_INTERVAL, 0)
-        wearPrefsRepository.setTemplateTileContent(content)
+        // TODO: combine to one call
+        wearPrefsRepository.setTemplateTile(content)
```
slovdahl, line 220: I've read the documentation on https://developer.android.com/training/wearables/data/events and on other pages in the `Handling data` section, but I'm not still not sure exactly when/why this method is called :thinking: It's called from `onDataChanged` in this class, but when is that method called?

slovdahl: Ah, I somehow assumed the `PhoneSettingsListener` class was run on the companion device, but it's in fact run on the wearable device. So this method is called whenever the settings have been changed on the companion device, right?

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearHomeView.kt
slovdahl: > how would it work to add a template tile here? Dont we need a tile ID to apply the template to it?

I had misunderstood how things work. The way it works is by the user adding a new Tile on the Wear OS device (or in the Wear device settings app on the phone), and only then it will become visible in the Home Assistant phone/wear app settings. So no button for adding a new Template tile needed here.

slovdahl: And big thanks to @marazmarci for figuring most of this out for us :grin: :+1:

slovdahl:
While testing this locally, I somehow ended up with stale data in the HA app :thinking: I removed some Template tiles from the Wear OS tile carousel, but some of them remained in the Home Assistant app data, and hence are visible in the list of Template tiles in the settings. @marazmarci did you see anything like this when implementing support for multiple shortcuts tiles?

marazmarci:
Yes, I encountered this when I was in the middle of development, and the app was crashing when `onTileRemoveEvent` was called :stuck_out_tongue:

Other than this, I didn't encounter such issues. However, I can imagine that Wear OS may give up calling `TileService::onTileRemoveEvent` in low-memory situations.

marazmarci:
> > While testing this locally, I somehow ended up with stale data in the HA app 🤔 I removed some Template tiles from the Wear OS tile carousel, but some of them remained in the Home Assistant app data, and hence are visible in the list of Template tiles in the settings. @marazmarci did you see anything like this when implementing support for multiple shortcuts tiles?
> 
> Yes, I encountered this when I was in the middle of development, and the app was crashing when `TileService::onTileRemoveEvent` was called, so it couldn't remove the tile from the DB :stuck_out_tongue:
> 
> Other than this, I didn't encounter such issues. However, I can imagine that Wear OS may give up calling `onTileRemoveEvent` in low-memory situations.

To make users able to fix a potential issue like this themselves, maybe we could provide an option to force remove a Template/Shortcut Tile 🤔

slovdahl:
> Yes, I encountered this when I was in the middle of development, and the app was crashing when TileService::onTileRemoveEvent was called, so it couldn't remove the tile from the DB 😛

I still don't understand how it happened for me. It worked for one of the tiles I removed but the other one was left in place :shrug: AFAICT the `onTileRemoveEvent` method was properly implemented at the time at least.

> To make users able to fix a potential issue like this themselves, maybe we could provide an option to force remove a Template/Shortcut Tile 🤔

Yeah, that could be an option :+1:

I suppose there is no way to query the currently active tiles? Or could we through some other side-channel keep track of whether a tile is in use or not? :thinking:

For Template tiles with a refresh interval set, we could actually keep track of it. But maybe it's overkill, and an option to manually force remove a tile would be better. But still not easy and obvious, how would a user know from the HA Template tile config whether a tile is in use or not? It would be quite easy to remove a tile config for an active tile, and that means that tile is in some kind of limbo state.

wear/src/main/java/io/homeassistant/companion/android/home/views/SelectTemplateTileView.kt
```diff
@@ -0,0 +1,86 @@
...
+                            modifier = Modifier
+                                .fillMaxWidth(),
+                            label = {
+                                Text(stringResource(commonR.string.template_tile_n, index + 1))
+                            },
+                            onClick = { onSelectTemplateTile(templateTileId) },
+                            colors = ChipDefaults.secondaryChipColors()
+                        )
+                    }
+                }
```
slovdahl, line 60: I'm still too new to this to understand how to properly solve this. Whenever a new Template tile is added in the Tile carousel, how do we make sure this view is refreshed with that info? I had to force close the HA Wear app and starting it again for a new Template tile to show up.

It looks like the same problem does not exist for the Shortcut tiles, but I couldn't figure out what I'm doing differently here.

dshokouhi: So with compose as you are passing them in as a variable `templateTileIds` in the method it should trigger a recomposition anytime the variable changes. I would check to make sure the variable is updating as you would expect. Should work similar to how we load entities on the home page or process state updates.

slovdahl: I changed `SelectTemplateTileView` to take `templateTiles: Map<Int?, TemplateTileConfig>` as a parameter instead, and `HomeView` passes in the mutable reference from `MainViewModel`. I thought that would do the trick, but it still did not update automatically 😱

wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt
```diff
@@ -224,21 +228,47 @@ fun LoadHomePage(
...
+                    }
+                )
+            ) { backStackEntry ->
+                val tileId = backStackEntry.arguments!!.getString(ARG_SCREEN_TEMPLATE_TILE_ID)!!.toIntOrNull()
+                val templateTileConfig = mainViewModel.templateTiles[tileId]
+
                 TemplateTileSettingsView(
-                    templateContent = mainViewModel.templateTileContent.value,
-                    refreshInterval = mainViewModel.templateTileRefreshInterval.value
+                    templateContent = templateTileConfig?.template ?: "",
```
dshokouhi, line 251: I believe we may need to access the view model variables directly here in the function instead of the variable defined at line 248 which may lead to static data. Not sure if your last comment suggested you had tried this already?

jpelgrom:
Good to know: a new camera tile with support for multiple instances has been merged into the app. As a result there is a merge conflict, but also things like [a single launcher activity for the tile settings](https://github.com/home-assistant/android/blob/master/wear/src/main/java/io/homeassistant/companion/android/tiles/OpenTileSettingsActivity.kt) that would be good to adopt.

jpelgrom:
Another good to know: the Wear app switched to Material 3 components. While your current build won't fail, please update UI composables to M3 components when updating 🙏

slovdahl:
Thanks for the heads up 👍 I haven't given up on this, I've just been very short on time and energy lately. If someone else feels like picking this up while I'm dragging my feet, I won't mind. Otherwise I'll try to get it forward in the coming weeks.

slovdahl:
> Regarding the stale tile discussion: I recently was debugging for issues with the camera tile and found that the current setup with `serviceScope` + `launch` in add/remove functions could result in the system destroying the service very quickly, before adding/removing was completed. As a workaround, and because the system plays an animation anyway, we've changed this to `runBlocking` to ensure the state stays correct. Probably best to implement it here too. See #3974 for an example.

Wow, nice, thank you for that! :+1: This was the thing that made getting this PR over the goal line feel hard or impossible. I'll pick it up again now :slightly_smiling_face:

slovdahl:
> Good to know: a new camera tile with support for multiple instances has been merged into the app. As a result there is a merge conflict, but also things like [a single launcher activity for the tile settings](https://github.com/home-assistant/android/blob/master/wear/src/main/java/io/homeassistant/companion/android/tiles/OpenTileSettingsActivity.kt) that would be good to adopt.

Fixed in bee0a798 :heavy_check_mark: 

> Another good to know: the Wear app switched to Material 3 components. While your current build won't fail, please update UI composables to M3 components when updating 🙏

Fixed in f39f1b47 :heavy_check_mark: 

> Regarding the stale tile discussion: I recently was debugging for issues with the camera tile and found that the current setup with serviceScope + launch in add/remove functions could result in the system destroying the service very quickly, before adding/removing was completed. As a workaround, and because the system plays an animation anyway, we've changed this to runBlocking to ensure the state stays correct. Probably best to implement it here too. See https://github.com/home-assistant/android/pull/3974 for an example.

Fixed in 7ef7a615 :heavy_check_mark:

wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt
slovdahl: Ah, 2 months later I think I understand what you meant here. No, IIRC I did not try what I think you're suggesting here. I'll do that now. :drum:

slovdahl: It didn't seem to help :unamused: Pushed the change now as 609ea01.

dshokouhi: did you confirm the variables had the expected data? as long as data is updating and the viewmodel variable is updating as things get added it should work just like the shortcut tile works

slovdahl: I think I eventually figured it out late last night. `PhoneSettingsListener` and `TemplateTile` only ever update `WearPrefsRepository`, and changes in it are never propagated to `MainViewModel`. I haven't fixed it yet, but I think I understand the problem now at least.

slovdahl: What would be the proper way to solve this? 🤔 I thought I followed prior art but maybe this is broken for shortcut tiles too.

slovdahl: I tested the following now, but I honestly don't understand how it can work:

- Opened the HA wear app on my watch and went to the settings and list of shortcut tiles (one previously added tile listed)
- Without killing the HA wear app, I went to the home screen and added a new shortcut tile
- Went back to the HA wear app and opened the shortcut tile settings again
- The new shortcut tile is listed :exploding_head: 

AFAICT, `ShortcutsTile` does not call anything else than `WearPrefsRepository` from `onTileAddEvent`, and still the change is picked up by `MainViewModel`. How on earth is that possible? What am I missing here?

dshokouhi: > I tested the following now, but I honestly don't understand how it can work:
> 
> * Opened the HA wear app on my watch and went to the settings and list of shortcut tiles (one previously added tile listed)
> * Without killing the HA wear app, I went to the home screen and added a new shortcut tile
> * Went back to the HA wear app and opened the shortcut tile settings again
> * The new shortcut tile is listed 🤯

Thats exactly how it works with Camera tile as well. When the user clicks on Shortcut tile settings they are loaded again.

https://github.com/home-assistant/android/blob/2c20baf0f54897c720099b34e57cf78eaa70fedb/wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt#L143

Camera tile is a state flow so anytime the database is updated the other values update as well. With Camera tiles you can keep the camera tile setting page open and you'll see new camera tiles being added when you return back to the screen.

https://github.com/home-assistant/android/blob/2c20baf0f54897c720099b34e57cf78eaa70fedb/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt#L94

slovdahl: > When the user clicks on Shortcut tile settings they are loaded again.
> https://github.com/home-assistant/android/blob/2c20baf0f54897c720099b34e57cf78eaa70fedb/wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt#L143

Oh.. That one I missed :facepalm: Thanks, with 23e50710 it seems to finally work :tada:

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearHomeView.kt
```diff
@@ -48,30 +50,51 @@ fun LoadSettingsHomeView(
...
+                        },
+                        onRefreshIntervalChanged = { refreshInterval ->
+                            settingsWearViewModel.setTemplateTileRefreshInterval(tileId!!, refreshInterval)
+                            settingsWearViewModel.sendTemplateTileInfo()
+                        },
+                        onBackClicked = {
+                            navController.navigateUp()
+                        }
+                    )
+                }
```
slovdahl, line 96: Any suggestions on how to handle errors here? I guess it's more of a theoretical problem now that no stale tile entries should be left in the app state.

dshokouhi: I am not sure if error handling is needed here as you already have a null check. Have you tested the behavior if a user  had 1 template tile before the upgrade and looked at phone settings before opening the app on the watch? I would imagine if there is any error it would possibly be from existing data

dshokouhi:
Had a chance to run a test on the debug APK and the feature is working as expected. Adding tiles updates the phone and watch and templates and options are reflected in teh appropriate location.

dshokouhi (commented):
Lets just make sure to clean up the TODO comment and also let us know if you had checked the one existing data case mentioned.

slovdahl:
> Had a chance to run a test on the debug APK and the feature is working as expected. Adding tiles updates the phone and watch and templates and options are reflected in teh appropriate location.

Thanks :+1: 

> Lets just make sure to clean up the TODO comment and also let us know if you had checked the one existing data case mentioned.

While testing this I think I found another problem. After upgrading both apps, if the phone settings are opened before an existing template tile on the watch has been refreshed at least once the phone settings will hang because of a crash on the wear side:

```
    2023-11-06 22:01:19.162  9210-9210  AndroidRuntime          io....stant.companion.android.debug  E  FATAL EXCEPTION: main
      Process: io.homeassistant.companion.android.debug, PID: 9210
      com.fasterxml.jackson.databind.JsonMappingException: Null key for a Map not allowed in JSON (use a converting NullKeySerializer?) (through reference chain: kotlin.collections.builders.MapBuilder["null"])
      	at com.fasterxml.jackson.databind.SerializerProvider.mappingException(SerializerProvider.java:1358)
      	at com.fasterxml.jackson.databind.SerializerProvider.reportMappingProblem(SerializerProvider.java:1254)
      	at com.fasterxml.jackson.databind.ser.impl.FailingSerializer.serialize(FailingSerializer.java:32)
      	at com.fasterxml.jackson.databind.ser.std.MapSerializer.serializeFields(MapSerializer.java:791)
      	at com.fasterxml.jackson.databind.ser.std.MapSerializer.serializeWithoutTypeInfo(MapSerializer.java:764)
      	at com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:720)
      	at com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:35)
      	at com.fasterxml.jackson.databind.ser.DefaultSerializerProvider._serialize(DefaultSerializerProvider.java:480)
      	at com.fasterxml.jackson.databind.ser.DefaultSerializerProvider.serializeValue(DefaultSerializerProvider.java:319)
      	at com.fasterxml.jackson.databind.ObjectMapper._writeValueAndClose(ObjectMapper.java:4568)
      	at com.fasterxml.jackson.databind.ObjectMapper.writeValueAsString(ObjectMapper.java:3821)
      	at io.homeassistant.companion.android.phone.PhoneSettingsListener$sendPhoneData$1.invokeSuspend(PhoneSettingsListener.kt:93)
      	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:108)
      	at android.os.Looper.loop(Looper.java:288)
      	at android.app.ActivityThread.main(ActivityThread.java:7898)
      	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
      	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
      	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@a9ddef7, Dispatchers.Main]
```

I guess it fails because we have not yet been able to persist the tile ID for the existing template tile, and we cannot serialize a map with a `null` key to JSON.

I managed to work around the initial problem by using `"null"` (the string) as key, similar to how `WearPrefsRepositoryImpl` does internally. But that just opened a new problem instead:


```
FATAL EXCEPTION: main
Process: io.homeassistant.companion.android.debug, PID: 12498
java.lang.IllegalArgumentException: Navigation destination that matches request NavDeepLinkRequest{ uri=android-app://androidx.navigation/Template/null } cannot be found in the navigation graph ComposeNavGraph(0x0) startDestination={Destination(0xf6edc9ca) route=Landing}
	at androidx.navigation.NavController.navigate(NavController.kt:1814)
	at androidx.navigation.NavController.navigate(NavController.kt:2220)
	at androidx.navigation.NavController.navigate$default(NavController.kt:2215)
	at io.homeassistant.companion.android.settings.wear.views.SettingsWearHomeViewKt$LoadSettingsHomeView$1$1$3$1.invoke(SettingsWearHomeView.kt:63)
	at io.homeassistant.companion.android.settings.wear.views.SettingsWearHomeViewKt$LoadSettingsHomeView$1$1$3$1.invoke(SettingsWearHomeView.kt:60)
	at io.homeassistant.companion.android.settings.wear.views.SettingsWearTemplateTileListKt$SettingsWearTemplateTileList$2$1$1.invoke(SettingsWearTemplateTileList.kt:47)
	at io.homeassistant.companion.android.settings.wear.views.SettingsWearTemplateTileListKt$SettingsWearTemplateTileList$2$1$1.invoke(SettingsWearTemplateTileList.kt:42)
	at io.homeassistant.companion.android.settings.views.SettingsRowKt$SettingsRow$1$1.invoke(SettingsRow.kt:38)
	at io.homeassistant.companion.android.settings.views.SettingsRowKt$SettingsRow$1$1.invoke(SettingsRow.kt:38)
	at androidx.compose.foundation.ClickablePointerInputNode$pointerInput$3.invoke-k-4lQ0M(Clickable.kt:895)
	at androidx.compose.foundation.ClickablePointerInputNode$pointerInput$3.invoke(Clickable.kt:889)
	at androidx.compose.foundation.gestures.TapGestureDetectorKt$detectTapAndPress$2$1.invokeSuspend(TapGestureDetector.kt:255)
	at kotlinx.coroutines.DispatchedTaskKt.resume(DispatchedTask.kt:179)
```

I guess we can handle that too by changing the argument type for `SettingsWearTemplateTile` from `NavType.IntType` to `NavType`.StringType`.. Though I didn't manage to get it working right away

```
FATAL EXCEPTION: main
Process: io.homeassistant.companion.android.debug, PID: 14435
java.lang.IllegalArgumentException: Wrong argument type for 'tileId' in argument bundle. string expected.
	at androidx.navigation.NavDestination.addInDefaultArgs(NavDestination.kt:602)
	at androidx.navigation.NavController.navigate(NavController.kt:1805)
	at androidx.navigation.NavController.navigate(NavController.kt:2220)
	at androidx.navigation.NavController.navigate$default(NavController.kt:2215)
	at io.homeassistant.companion.android.settings.wear.views.SettingsWearHomeViewKt$LoadSettingsHomeView$1$1$3$1.invoke(SettingsWearHomeView.kt:63)
	at io.homeassistant.companion.android.settings.wear.views.SettingsWearHomeViewKt$LoadSettingsHomeView$1$1$3$1.invoke(SettingsWearHomeView.kt:60)
	at io.homeassistant.companion.android.settings.wear.views.SettingsWearTemplateTileListKt$SettingsWearTemplateTileList$2$1$1.invoke(SettingsWearTemplateTileList.kt:47)
	at io.homeassistant.companion.android.settings.wear.views.SettingsWearTemplateTileListKt$SettingsWearTemplateTileList$2$1$1.invoke(SettingsWearTemplateTileList.kt:42)
	at io.homeassistant.companion.android.settings.views.SettingsRowKt$SettingsRow$1$1.invoke(SettingsRow.kt:38)
	at io.homeassistant.companion.android.settings.views.SettingsRowKt$SettingsRow$1$1.invoke(SettingsRow.kt:38)
	at androidx.compose.foundation.ClickablePointerInputNode$pointerInput$3.invoke-k-4lQ0M(Clickable.kt:895)
	at androidx.compose.foundation.ClickablePointerInputNode$pointerInput$3.invoke(Clickable.kt:889)
	at androidx.compose.foundation.gestures.TapGestureDetectorKt$detectTapAndPress$2$1.invokeSuspend(TapGestureDetector.kt:255)
```

But this does not feel like a sustainable path forward overall :thinking: Any ideas? Could we somehow trigger a refresh of the tiles after updating the app so we minimize the window for this to happen?

dshokouhi:
> But this does not feel like a sustainable path forward overall 🤔 Any ideas? Could we somehow trigger a refresh of the tiles after updating the app so we minimize the window for this to happen?

IIRC I think the OS handles this outside the freshness interval. Maybe its best not to send any data when the tile hasn't been refreshed yet and just provide a friendly message to the user? Either way though we shouldn't crash in any of these cases so would be good to add some try/catch blocks so we don't have a crash unexpectedly.

slovdahl:
Using `null` as key for the template tile that we don't yet know the ID didn't feel 100% clean from the start, but now I got rid of it (0ff5c0a) and replaced it with `-1` instead. Hopefully that doesn't clash with a real ID given by Android. All IDs I have seen have been positive integers at least.

That also nicely solved the problem I had with serializing a `null` key with Jackson.

AFAIK, everything should finally work now and the PR is ready for (final) review.

jpelgrom:
@slovdahl I don't mind reviewing but you literally pushed a `TODO` item for what looks like debugging after requesting a review. This does not seem finished? If you want feedback on something specific just ask.

slovdahl:
@jpelgrom I'm sorry for the confusion, it was not intentional. I didn't notice that there was a rebase conflict before I requested the review, so I rebased it and fixed the trivial import-related conflict locally, and tried to push all the commits except the last `REVERTME` one. But I obviously failed and managed to push that one as well. :see_no_evil:

It should all be fixed now, and the PR is ready for review whenever you have time. :bow:

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
```diff
@@ -0,0 +1,88 @@
...
+    Scaffold(
+        topBar = {
+            SettingsWearTopAppBar(
+                title = { Text(stringResource(commonR.string.template_tiles)) },
+                onBackClicked = onBackClicked,
+                docsLink = WEAR_DOCS_LINK
+            )
+        }
+    ) { padding ->
+        Column(
```
jpelgrom, line 32: This column should be scrollable (vertical) in case of a short screen/large text and multiple tiles to ensure they all can be edited. Make sure to apply scroll after setting padding, otherwise you end up with a white/black border around your scrollable list :)

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
```diff
@@ -0,0 +1,88 @@
...
+                .padding(all = 16.dp)
+        ) {
+            if (templateTiles.entries.isEmpty()) {
+                Text(stringResource(commonR.string.template_tile_no_tiles_yet))
+            } else {
+                var index = 1
+                for (templateTileEntry in templateTiles.entries) {
+                    SettingsRow(
+                        primaryText = stringResource(commonR.string.template_tile_n, index++),
+                        secondaryText = stringResource(commonR.string.template_tile_configure),
```
jpelgrom, line 44: If you have more than one tile, this is a bit repetitive for every row. Suggestion: add it as a header to the list (similar to the 'Manage [Wear device] header' in the main screen).

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/impl/entities/TemplateTileConfig.kt
```diff
@@ -0,0 +1,36 @@
...
+        )
+    }
+
+    fun withTemplate(template: String): TemplateTileConfig {
+        return TemplateTileConfig(template, refreshInterval)
+    }
+
+    fun withRefreshInterval(refreshInterval: Int): TemplateTileConfig {
+        return TemplateTileConfig(template, refreshInterval)
+    }
```
jpelgrom, line 35: Why create these functions when Kotlin data classes have the built-in [copy function for this purpose](https://kotlinlang.org/docs/data-classes.html#copying) (for example `templateTileConfig?.copy(template = newTemplate)`)?

jpelgrom (commented):
Sorry for the delay in getting back to the review. As a result I've noticed some new things to address... Thanks for the efforts so far, almost there 🙏

app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearViewModel.kt
```diff
@@ -144,17 +142,42 @@ class SettingsWearViewModel @Inject constructor(
...
-                        serverManager.integrationRepository(serverId).renderTemplate(template, mapOf()).toString()
+                    templateTilesRenderedTemplates[tileId] = serverManager
+                        .integrationRepository(serverId)
+                        .renderTemplate(template, mapOf()).toString()
                 } catch (e: Exception) {
-                    Log.e(TAG, "Exception while rendering template", e)
+                    Log.e(TAG, "Exception while rendering template for tile ID $tileId", e)
                     // JsonMappingException suggests that template is not a String (= error)
-                    templateTileContentRendered.value = getApplication<Application>().getString(
+                    getApplication<Application>().getString(
```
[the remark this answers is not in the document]
slovdahl: Oops, good catch there. Fixed!

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
slovdahl: Good idea - and fully agreed about the repetition :+1: Added it as a header.

`SettingsRow` requires a `secondaryText` to be provided though (as far as I understand it), and by giving it an empty string, the alignment looks a bit off. Any suggestions?

Making `secondaryText` optional and only outputting it if it's non-null seems to fix the alignment at least:

```diff
diff --git app/src/main/java/io/homeassistant/companion/android/settings/views/SettingsRow.kt app/src/main/java/io/homeassistant/companion/android/settings/views/SettingsRow.kt
index 01231d2a..190884e0 100644
--- app/src/main/java/io/homeassistant/companion/android/settings/views/SettingsRow.kt
+++ app/src/main/java/io/homeassistant/companion/android/settings/views/SettingsRow.kt
@@ -28,7 +28,7 @@ import com.mikepenz.iconics.typeface.IIcon
 @Composable
 fun SettingsRow(
     primaryText: String,
-    secondaryText: String,
+    secondaryText: String?,
     mdiIcon: IIcon?,
     enabled: Boolean,
     onClicked: () -> Unit
@@ -66,11 +66,13 @@ fun SettingsRow(
                 text = primaryText,
                 style = MaterialTheme.typography.body1
             )
-            CompositionLocalProvider(LocalContentAlpha provides ContentAlpha.medium) {
-                Text(
-                    text = secondaryText,
-                    style = MaterialTheme.typography.body2
-                )
+            secondaryText?.let {
+                CompositionLocalProvider(LocalContentAlpha provides ContentAlpha.medium) {
+                    Text(
+                        text = it,
+                        style = MaterialTheme.typography.body2
+                    )
+                }
             }
         }
     }
```

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
slovdahl: Makes sense :+1: Would this be the right way of doing that?

```diff
diff --git app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
index 24b927e1..b4d33273 100644
--- app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
+++ app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
@@ -2,6 +2,8 @@ package io.homeassistant.companion.android.settings.wear.views
 
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.Scaffold
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
@@ -33,6 +35,7 @@ fun SettingsWearTemplateTileList(
             Modifier
                 .padding(padding)
                 .padding(all = 16.dp)
+                .verticalScroll(rememberScrollState())
         ) {
             if (templateTiles.entries.isEmpty()) {
                 Text(stringResource(commonR.string.template_tile_no_tiles_yet))
```

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/impl/entities/TemplateTileConfig.kt
slovdahl: Probably because this is the first time I'm using Kotlin for real and I'm used to Java :rofl: Thanks, fixed (and removed the methods):

```diff
diff --git app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearViewModel.kt app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearViewModel.kt
index fd091489..2a26f85e 100644
--- app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearViewModel.kt
+++ app/src/full/java/io/homeassistant/companion/android/settings/wear/SettingsWearViewModel.kt
@@ -145,7 +145,7 @@ class SettingsWearViewModel @Inject constructor(
     fun setTemplateTileContent(tileId: Int, updatedTemplateTileContent: String) {
         val templateTileConfig = templateTiles[tileId]
         templateTileConfig?.let {
-            templateTiles[tileId] = it.withTemplate(updatedTemplateTileContent)
+            templateTiles[tileId] = it.copy(template = updatedTemplateTileContent)
             renderTemplate(tileId, updatedTemplateTileContent)
         }
     }
@@ -153,7 +153,7 @@ class SettingsWearViewModel @Inject constructor(
     fun setTemplateTileRefreshInterval(tileId: Int, refreshInterval: Int) {
         val templateTileConfig = templateTiles[tileId]
         templateTileConfig?.let {
-            templateTiles[tileId] = it.withRefreshInterval(refreshInterval)
+            templateTiles[tileId] = it.copy(refreshInterval = refreshInterval)
         }
     }

diff --git wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
index 976d76db..9d471af1 100644
--- wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
+++ wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
@@ -497,7 +497,7 @@ class MainViewModel @Inject constructor(
         viewModelScope.launch {
             homePresenter.setTemplateTileRefreshInterval(tileId, interval)
             templateTiles[tileId]?.let {
-                templateTiles[tileId] = it.withRefreshInterval(interval)
+                templateTiles[tileId] = it.copy(refreshInterval = interval)
             }
         }
     }
```

jpelgrom: I think you forgot to push this commit

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
jpelgrom: If you need to show something as the secondary label, maybe show the template (or the first 100 characters to prevent a very long template from filling the screen)? The current fix with an optional secondary label is also fine.

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
jpelgrom: Sorry, wrong way around :) Apply the scroll first, then the padding. If you scroll now, it'll have the borders I mentioned because you apply the scrolling modifier to what remains after applying the padding.

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
slovdahl: Hmm, that's a really good idea. Only the numbers make it hard to know which tile a specific one is. Showing the template would make it a little bit easier :ok_hand: Fixed like that instead.

common/src/main/java/io/homeassistant/companion/android/common/data/prefs/impl/entities/TemplateTileConfig.kt
slovdahl: Staged but not commited :see_no_evil: Pushed now.

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
slovdahl: Alright :+1: 

```diff
diff --git app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
index f6fb0fb3..423ea159 100644
--- app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
+++ app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
@@ -38,9 +38,9 @@ fun SettingsWearTemplateTileList(
     ) { padding ->
         Column(
             Modifier
+                .verticalScroll(rememberScrollState())
                 .padding(padding)
                 .padding(all = 16.dp)
-                .verticalScroll(rememberScrollState())
         ) {
             if (templateTiles.entries.isEmpty()) {
                 Text(stringResource(commonR.string.template_tile_no_tiles_yet))
```

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
```diff
@@ -0,0 +1,119 @@
...
+                        primaryText = stringResource(commonR.string.template_tile_n, index++),
+                        secondaryText = when {
+                            template.length < 100 -> {
+                                template
+                            }
+
+                            else -> {
+                                template.substring(0, template.length.coerceAtMost(100)) + "..."
+                            }
+                        },
```
jpelgrom, line 75: 100 wasn't handled, use the proper ellipsis character, make it a bit more Kotlin-y

```suggestion
                        secondaryText = when {
                            template.length <= 100 -> template
                            else -> "${template.take(100)}…"
                        },
```

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
```diff
@@ -0,0 +1,119 @@
...
+                onBackClicked = onBackClicked,
+                docsLink = WEAR_DOCS_LINK
+            )
+        }
+    ) { padding ->
+        Column(
+            Modifier
+                .verticalScroll(rememberScrollState())
+                .padding(padding)
+                .padding(all = 16.dp)
```
jpelgrom, line 43: The title / rows already have padding so you don't actually need this most of the time. 

You do need some padding for the 'no tiles yet' message (line 46), so I suggest setting the 16dp padding on that text instead.

wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
```diff
@@ -121,9 +122,9 @@ class MainViewModel @Inject constructor(
         private set
     var isShowShortcutTextEnabled = mutableStateOf(false)
         private set
-    var templateTileContent = mutableStateOf("")
+    var templateTiles = mutableStateMapOf<Int, TemplateTileConfig>()
         private set
-    var templateTileRefreshInterval = mutableStateOf(0)
+    var templateTilesRenderedTemplates = mutableStateMapOf<Int, String>()
```
jpelgrom, line 127: Unused

app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
slovdahl: Like this?

```diff
diff --git app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
index f77983e8..85329f96 100644
--- app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
+++ app/src/full/java/io/homeassistant/companion/android/settings/wear/views/SettingsWearTemplateTileList.kt
@@ -40,10 +40,13 @@ fun SettingsWearTemplateTileList(
             Modifier
                 .verticalScroll(rememberScrollState())
                 .padding(padding)
-                .padding(all = 16.dp)
         ) {
             if (templateTiles.entries.isEmpty()) {
-                Text(stringResource(commonR.string.template_tile_no_tiles_yet))
+                Text(
+                    text = stringResource(commonR.string.template_tile_no_tiles_yet),
+                    modifier = Modifier
+                        .padding(all = 16.dp)
+                )
             } else {
                 Row(
                     modifier = Modifier
```

wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt
```diff
@@ -317,21 +324,46 @@ fun LoadHomePage(
...
-            composable(SCREEN_SET_TILE_TEMPLATE) {
+            composable("$ROUTE_TEMPLATE_TILE/$SCREEN_SELECT_TEMPLATE_TILE") {
+                SelectTemplateTileView(
+                    templateTiles = mainViewModel.templateTiles,
+                    onSelectTemplateTile = { tileId ->
+                        swipeDismissableNavController.navigate("$ROUTE_TEMPLATE_TILE/$tileId/$SCREEN_SET_TILE_TEMPLATE")
+                    }
+                )
+            }
+            composable(
```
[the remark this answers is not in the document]
slovdahl: Like this?

```diff
diff --git wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt
index 219f99ba..1ffd48b7 100644
--- wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt
+++ wear/src/main/java/io/homeassistant/companion/android/home/views/HomeView.kt
@@ -338,6 +338,9 @@ fun LoadHomePage(
                     navArgument(name = ARG_SCREEN_TEMPLATE_TILE_ID) {
                         type = NavType.StringType
                     }
+                ),
+                deepLinks = listOf(
+                    navDeepLink { uriPattern = "$DEEPLINK_PREFIX_SET_TEMPLATE_TILE/{$ARG_SCREEN_TEMPLATE_TILE_ID}" }
                 )
             ) { backStackEntry ->
                 val tileId = backStackEntry.arguments!!.getString(ARG_SCREEN_TEMPLATE_TILE_ID)!!.toIntOrNull()
```

jpelgrom (approved):
🎉

dshokouhi (approved):

Issue #2369, opened 2022-03-14
App doesn't always connect using external address - works fine in browser


surge919:
**Home Assistant Android version:**
beta-2162-2de85151-full

**Android version:**
12

**Phone model:**
Pixel 5a

**Home Assistant version:**
2022.3.4

**Last working Home Assistant release (if known):**
don't remember

**Description of problem:**
I’m seeing intermittent weirdness.

If I connect locally and turn off wifi , it works and I get this in the log

```
2022-03-13 18:51:29 ERROR (MainThread) [frontend.js.latest.202203011] https://<external_ip>/lovelace/home?external_auth=1:0:0 Uncaught
```

When I’m away for a period of time, I’m unable to connect externally and this is in the log
I can also replicate this by restarting HASS and then trying to connect using the app

```
2022-03-13 19:04:36 WARNING (MainThread) [homeassistant.components.http.ban] Login attempt or request with invalid authentication from 172.xx.xx.xx (172.xx.xx.xx). (HomeAssistant/Android Pixel 5a beta-2162-2de58151-2162)
```

I searched all folders and can’t find any ip_bans.yaml so it doesn’t seem to actually be banning the IP.

**Traceback (if applicable, to get the logs you may refer to: https://companion.home-assistant.io/docs/troubleshooting/faqs/#android-crash-logs):**

This is from the app log and seems to be the only mention of not being able to connect to the external URL

```
03-13 19:06:18.704 24736 24736 E WebViewPresenterImpl: Unable to retrieve external auth
03-13 19:06:18.704 24736 24736 E WebViewPresenterImpl: io.homeassistant.companion.android.common.data.authentication.AuthorizationException
03-13 19:06:18.704 24736 24736 E WebViewPresenterImpl: 	at io.homeassistant.companion.android.common.data.authentication.impl.AuthenticationRepositoryImpl.ensureValidSession(AuthenticationRepositoryImpl.kt:231)
03-13 19:06:18.704 24736 24736 E WebViewPresenterImpl: 	at io.homeassistant.companion.android.common.data.authentication.impl.AuthenticationRepositoryImpl.access$ensureValidSession(AuthenticationRepositoryImpl.kt:27)
03-13 19:06:18.704 24736 24736 E WebViewPresenterImpl: 	at io.homeassistant.companion.android.common.data.authentication.impl.AuthenticationRepositoryImpl$ensureValidSession$1.invokeSuspend(Unknown Source:15)
[...]

PR #3787, opened 2023-08-12
Bugfix/crash web view

dmiales:
This PR fixes issue https://github.com/home-assistant/android/issues/3500. What was the issue:
I have a tablet hanging as a home control. The tablet is running home assistant. And about once a day the application crashes. And when I pass by it, I see the desktop. The reason is that the WebView is leaking memory. And in the event of the death of the WebView process (com.google.android.webview:sandboxed_process0:org.chromium.content.app.SandboxedProcessService), our application also died. It is unacceptable.
I immediately provide a reproduction of the problem:

1) use adb: `adb shell`
2) without permissions, we won't be able to kill the process: `su`
3) find and kill the process: `ps -A | grep sandboxed | awk '{print $2}' | xargs kill`

For clarity, I will add two videos (before / after)

[![before]](https://github.com/home-assistant/android/assets/24318525/e043fc7f-27cd-40c6-b626-b886e2328cbb)

[![after]](https://github.com/home-assistant/android/assets/24318525/16f01bac-0661-4606-b4d6-d5c115f925a2)

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -225,6 +225,7 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
     private var downloadFileUrl = ""
     private var downloadFileContentDisposition = ""
     private var downloadFileMimetype = ""
+    private var javascriptInterface = "externalApp"
 
     @SuppressLint("SetJavaScriptEnabled")
     override fun onCreate(savedInstanceState: Bundle?) {
@@ -402,7 +403,10 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                     handler: RenderProcessGoneDetail?
                 ): Boolean {
                     Log.e(TAG, "onRenderProcessGone: webView crashed")
-                    view?.let { reload() }
+                    view?.let {
+                        reload()
+                        webViewAddJavascriptInterface()
+                    }
 
                     return true
                 }
@@ -591,6 +595,63 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                 }
             }
 
+            webViewAddJavascriptInterface()
+        }
+
+        // Set WebView background color to transparent, so that the theme of the android activity has control over it.
+        // This enables the ability to have the launch screen behind the WebView until the web frontend gets rendered
+        binding.webview.setBackgroundColor(Color.TRANSPARENT)
+
+        themesManager.setThemeForWebView(this, webView.settings)
+
+        val cookieManager = CookieManager.getInstance()
+        cookieManager.setAcceptCookie(true)
+        cookieManager.setAcceptThirdPartyCookies(webView, true)
+
+        window.decorView.setOnSystemUiVisibilityChangeListener { visibility ->
+            if (visibility and View.SYSTEM_UI_FLAG_FULLSCREEN == 0) {
+                if (presenter.isFullScreen()) {
+                    hideSystemUI()
+                }
+            }
+        }
+
+        if (presenter.isKeepScreenOnEnabled()) {
+            window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+        }
+
+        currentAutoplay = presenter.isAutoPlayVideoEnabled()
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val webviewPackage = WebViewCompat.getCurrentWebViewPackage(this)
+            Log.d(TAG, "Current webview package ${webviewPackage?.packageName} and version ${webviewPackage?.versionName}")
+        }
+
+        lifecycleScope.launch {
+            lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
+                presenter.getMatterCommissioningStatusFlow().collect {
+                    Log.d(TAG, "Matter commissioning status changed to $it")
+                    when (it) {
+                        MatterFrontendCommissioningStatus.THREAD_EXPORT_TO_SERVER,
+                        MatterFrontendCommissioningStatus.IN_PROGRESS -> {
+                            presenter.getMatterCommissioningIntent()?.let { intentSender ->
+                                commissionMatterDevice.launch(IntentSenderRequest.Builder(intentSender).build())
+                            }
+                        }
+                        MatterFrontendCommissioningStatus.ERROR -> {
+                            Toast.makeText(this@WebViewActivity, commonR.string.matter_commissioning_unavailable, Toast.LENGTH_SHORT).show()
+                            presenter.confirmMatterCommissioningError()
+                        }
+                        else -> { } // Do nothing
+                    }
+                }
+            }
+        }
+    }
+
+    private fun webViewAddJavascriptInterface() {
+        webView.removeJavascriptInterface(javascriptInterface)
+        webView.apply {
             addJavascriptInterface(
                 object : Any() {
                     // TODO This feature is deprecated and should be removed after 2022.6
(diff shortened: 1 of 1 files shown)


jpelgrom:
The [documentation](https://developer.android.com/reference/android/webkit/WebViewClient#onRenderProcessGone(android.webkit.WebView,%20android.webkit.RenderProcessGoneDetail)) specifically calls out that the actual view shouldn't be reused:

> The given WebView can't be used, and should be removed from the view hierarchy, all references to it should be cleaned up, e.g any references in the Activity or other classes saved using `View.findViewById(int)` and similar calls, etc.

but this is reusing the view?

Clearing cache also should be removed as it isn't related to crash handling.

dmiales:
Removed unnecessary calls.

Because we're using a trick.
The documentation says: 

> Notify host application that the given WebView's render process has exited. Our goal is not to let it complete, we process the situation and reboot (a new sandbox process is created).

> if the host application handled the situation that process has exited, otherwise, application will crash if render process crashed, or be killed if render process was killed by the system.

And in this case, a running application is clearly preferable to a falling one.

jpelgrom:
Now we've quoted all documentation for that function :)

I still don't understand your position. You're reusing the view. That is not a 'trick'? As I understand it, you can use this callback to tear out the view and handle it (which could be creating a new one).

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -386,6 +387,15 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                     )
                 }
 
+                override fun onRenderProcessGone(
+                    view: WebView?,
+                    handler: RenderProcessGoneDetail?
+                ): Boolean {
+                    webView.reload()
```
dshokouhi, line 394: Should we add a logging statement here? not sure if any details will be helpful but could be useful for troubleshooting purposes?

dshokouhi: Also it looks like this method gets called for every open webview so maybe we need to match `view` to make sure its what we expected? `view == webView`  before doing any recreation?

> Multiple WebView instances may be associated with a single render process; onRenderProcessGone will be called for each WebView that was affected. 

Not sure if we need to worry about an actual crash or not too?

https://developer.android.com/reference/android/webkit/RenderProcessGoneDetail#didCrash()

this is indeed a good find though, nice to know we can handle this.

dmiales: @dshokouhi 
I added logging. It's important, I agree.

didCrash() is a cool check. But there is a problem:

> Added in [API level 26](https://developer.android.com/guide/topics/manifest/uses-sdk-element#ApiLevels)

we have min 21

> view == webView 

wouldn't that be overkill? As far as I could check in webView is always the same as in the view parameter.


##
I keep you informed, the application on the tablet has been working for almost two days without failures

dshokouhi: > didCrash() is a cool check. But there is a problem:\n\nAdded in API level 26

So is the entire method you are overriding

https://developer.android.com/reference/android/webkit/WebViewClient#onRenderProcessGone(android.webkit.WebView,%20android.webkit.RenderProcessGoneDetail)

Just add a conditional check we have plenty of code like that

dmiales: this is strange behavior
изображение

Maybe then like this? Reload view from parameter

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -386,6 +387,16 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                     )
                 }
 
+                override fun onRenderProcessGone(
+                    view: WebView?,
+                    handler: RenderProcessGoneDetail?
+                ): Boolean {
+                    Log.e(TAG, "onRenderProcessGone: webView crashed")
+                    view!!.reload()
```
dshokouhi, line 395: This will crash if `view` is `null` we should avoid crashing when possible

dmiales: confirm

dshokouhi: Reading the description again I dont think WebView can load in certain cases which is why the description says to remove the view from the hierarchy

`The application's implementation of this callback should only attempt to clean up the specific WebView given as a parameter, and should not assume that other WebView instances are affected. The given WebView can't be used, and should be removed from the view hierarchy, all references to it should be cleaned up, e.g any references in the Activity or other classes saved using [View.findViewById(int)](https://developer.android.com/reference/android/view/View#findViewById(int)) and similar calls, etc.`

have you tested this fix against the site Google recommended to test against? 

` To cause an render process crash for test purpose, the application can call loadUrl("chrome://crash") on the WebView.`

dmiales:
@dshokouhi 

I was on vacation. But before that, I committed the changes and tested the hypotheses. Replying to the first comment:

> have you tested this fix against the site Google recommended to test against?

Yes, I checked for chrome://crash, it works identical to killing the process.
Replying to the last comment:

> We need to make sure we follow Google's recommendation here and to clean up the WebView instance that is no longer available

https://github.com/home-assistant/android/pull/3787/files#diff-76bbedb2918a5db79c7070e9bb2e01d9241388745f60b0719c0d1f74123d89bdR396 I corrected it here, we take exactly the instance that comes into the method.

While I was testing my solution, I found an unusual error that broke the application
https://github.com/home-assistant/core/issues/71205 (the same error exists in neighboring projects.
https://github.com/home-assistant/android/issues/2076#issuecomment-1002206381
https://github.com/home-assistant/android/issues/2369#issuecomment-1067247865
The essence is the same: Disconnected: "Did not receive auth message within 10 seconds") I debugged and realized that when we recreate the WebView, addJavascriptInterface is not created again. Because of this, we lose communication with the internal API. As a result, I put it in a separate method; we use it during initial initialization and again when re-creating the WebView.

dshokouhi:
> https://github.com/home-assistant/android/pull/3787/files#diff-76bbedb2918a5db79c7070e9bb2e01d9241388745f60b0719c0d1f74123d89bdR396 I corrected it here, we take exactly the instance that comes into the method.

so looking at the commit we are attempting to `reload()` the webview if the `view` is still there. However that is not following Google's recommendation of removing the view from the hierarchy.

Reloading does not do anything to the view hierarchy according to the docs.

https://developer.android.com/reference/android/webkit/WebView#reload()

dmiales:
I understand what you are talking about. Probably the ideal solution is to completely recreate the webview? But it’s extremely difficult to do this with this hook. It’s unlikely that I have enough competence to implement this.

The question here is why this is not recommended, and whether we should adhere to this recommendation in this case. After all, we are rebooting in a super-critical situation, the other way out of which is a complete crash of the application.
And if you look at which is worse, the choice is obvious.

At least my smart home is on this commit, which makes me happy

dmiales:
@dshokouhi check my comment

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -224,6 +225,7 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
     private var downloadFileUrl = ""
     private var downloadFileContentDisposition = ""
     private var downloadFileMimetype = ""
+    private var javascriptInterface = "externalApp"
```
dshokouhi, line 228: this can be a `val` as its not going to be changed

dshokouhi (commented):
So I ran some testing here and using Googles recommended way of testing the issue I can see that it is working and the page is refreshed accordingly. Lets just make sure to clean up any new introduced warnings and I think this should be ok.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
dmiales: done

dshokouhi (approved):
So I tested the PR using the recommendation from Google (loading `chrome://crash`). Actually could only test using remote debugging. Sure enough webview crashed and the code reloaded and the page was fine so the fix looks good here.

dmiales:
I also tested using this method. And the main thing is that the tablet has been crash-free for almost two weeks now

JBassett (approved):

Issue #3488, opened 2023-04-23
Add click event to entity widgets 


torwag:
**Is your feature request related to a problem? Please describe.**

An entity widget depicts the current value of an entity. I somehow expect clicking on the widget to be guided to a more detailed page of the home assistant companion app. E. g. to see a graph of said entity with historical values. 
**Describe the solution you'd like**

As people expect different actions, a generic solution, within the setup to define am action might be the best way to go. A on-click parameter could be set by default to "trigger update" as it is right now. However, one could also select "open dashboard", "open entity graph of the last x hours" and potentially other actions. 
That would give people a lot of customisation options. 

**Describe alternatives you've considered, if any**

Open the app, select the dashboard resp. the entity graph. 

**Additional context**

dshokouhi:
The problem with this request is that the current click action is designed to update the widget on tap. This will be a behavior change that will impact all users. Have you considered using the shortcuts feature that will open the entity and/or specific dashboard. I think that will make the most sense here to avoid a breaking change. It's best to avoid adding too many features that need to be maintained when there are existing features that already do what is requested.

torwag:
Thanks for looking into this.
I think there is no need for a breaking change, as the default / none configured function could be exactly what it is now, update the widget.

Only if a user decides to trigger another action on click, that behavior would change.

I think this feature also originates from the actual behavior in the HA web frontend, where one often will be able to open a history graph clicking on a dashboard widget. 

I will look into shortcuts.

PR #3798, opened 2023-08-18
Add 'toggle' tap action to entity state widget

### LINKED ISSUE #3078: Widget with entity state and call service combined into one
At present, the button widgets can only be used to call services, and the entity state widgets can only be used to view entity status. I hope to control entities and display entity status on one widget.

Add a widget with entity state and call service widgets combined into one.

jpelgrom:
## Summary

This PR adds a tap action option to the entity state widget, allowing users to choose between 'refresh' (current behavior) and 'toggle' (approximately, like QS/Wear) on supported entities. This should solve the most common complaint people have with the app widgets: it requires two widgets to see a state and change it, they want something like a switch.

I'm going to say this fixes #3078 - all examples given there relate to toggling entities. The tap action is set up to allow for extending it to do other actions in the future though, which is covered by #3488.

New widgets for entities that are tappable now default to this action, to match how the frontend button card also toggles/turns on/presses those entities by default.

To prevent duplicate code, the app tap/toggle action code is shared between the main app widgets, Wear shortcut tiles, and the quick settings panel tiles. The implementation is based on the pressed function used in the car interface, but without requiring an `Entity` object if possible (which would need to be fetched and slows down the request).

## Screenshots

Tapping shows a loading indicator. It is hidden by the state changing which triggers a regular widget update.

The widget configuration has a 'tap action' spinner on supported entities:
|Light|Dark|
|-----|-----|
|Entity state widget configuration with a 'Tap action: Toggle / Refresh' option, light mode|Entity state widget configuration with a 'Tap action: Toggle / Refresh' option, dark mode|

Documentation: home-assistant/companion.home-assistant#974


dshokouhi:
should we consider updating the widget description with this change?

jpelgrom:
> should we consider updating the widget description with this change?

Updated. Android limits string length to 2 lines in the widget chooser so I've simplified the description to make it fit. German translators might have a difficult time ;)

Home Assistants widgets panel with entity state widget visible with description "View any entity's current state and toggle it"

JBassett (approved):

JBassett:
Just need to fix merge conflicts then good to go.

PR #3805, opened 2023-08-19
Android Auto: Set icon color when entity is considered in an active state

dshokouhi:
## Summary

Implements: https://community.home-assistant.io/t/add-accent-color-to-the-android-auto-ui/604634

Sets the icon color tint to the frontend yellow color if the entity is considered to be active according to the [frontend](https://github.com/home-assistant/frontend/blob/dev/src/common/entity/state_active.ts)

Also updates the Wear OS yellow color to more closely match the color from the HA frontend.

## Screenshots

image

## Any other notes

I had thought about mimicking the frontend using blue for off and yellow for on but the screen looked a bit busy so decided to  only change teh default color if the entity is considered active.

image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
@@ -29,6 +29,7 @@ import io.homeassistant.companion.android.common.data.integration.domain
 import io.homeassistant.companion.android.common.data.integration.friendlyName
 import io.homeassistant.companion.android.common.data.integration.friendlyState
 import io.homeassistant.companion.android.common.data.integration.getIcon
+import io.homeassistant.companion.android.common.data.integration.isActive
 import io.homeassistant.companion.android.common.data.integration.isExecuting
 import io.homeassistant.companion.android.common.data.integration.onPressed
 import io.homeassistant.companion.android.common.data.prefs.PrefsRepository
@@ -212,7 +213,16 @@ class EntityGridVehicleScreen(
                                 sizeDp = 64
                             }.toAndroidIconCompat()
                         )
-                            .setTint(CarColor.DEFAULT)
+                            .setTint(
+                                if (entity.isActive()) {
+                                    CarColor.createCustom(
+                                        carContext.getColor(R.color.colorYellow),
+                                        carContext.getColor(R.color.colorYellow)
+                                    )
+                                } else {
+                                    CarColor.DEFAULT
+                                }
+                            )
                             .build()
                     )
             }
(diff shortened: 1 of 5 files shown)


jpelgrom:
(also implements [this review comment by a frontend member on the CarPlay PR](https://github.com/home-assistant/iOS/pull/2320#issuecomment-1615938547), so design approved I"d say!)

Can you add a link to the frontend code for determining active state in a comment for that function, like the icon function has? If it is ever changed that makes it easier to reference and update.

dshokouhi:
> (also implements [this review comment by a frontend member on the CarPlay PR](https://github.com/home-assistant/iOS/pull/2320#issuecomment-1615938547), so design approved I"d say!)
> 
> Can you add a link to the frontend code for determining active state in a comment for that function, like the icon function has? If it is ever changed that makes it easier to reference and update.

Sure once my home Internet is back up, storm took it down lol

dshokouhi:
> Can you add a link to the frontend code for determining active state in a comment for that function, like the icon function has? If it is ever changed that makes it easier to reference and update.

finally back online, this is done

jpelgrom:
Looks like a `lawn_mower` domain was added yesterday which requires a custom case in the active states check...

Code looks good, but I'm seeing some differences that I'm not yet following. For example an entities card with 'Color icons based on state' with `button`s is all blue in the frontend (= inactive), but all yellow in the Auto interface (= active).

dshokouhi:
> Looks like a `lawn_mower` domain was added yesterday which requires a custom case in the active states check...

lol were gonna have to be on top of newly added domains then with updating icons and availability in features. Probably a good idea to come up with a process for adding support for new domains



> Code looks good, but I'm seeing some differences that I'm not yet following. For example an entities card with 'Color icons based on state' with `button`s is all blue in the frontend (= inactive), but all yellow in the Auto interface (= active).
> 
> Edit: think I found it right after I spent 15 minutes looking ;) https://github.com/home-assistant/frontend/blob/dev/src/common/entity/state_color.ts appears to be based on a list of domains as well whether or not the icon gets a color for active states. I would say add the check from lines 109-111, that seems easy enough but prevents making this function too complicated.

Ah good find, went ahead and added this. Should hopefully take care of any inconsistencies :)

jpelgrom:
I think the domains that can be colored list should be separate from `isActive`. Only used for coloring icons now but who knows what will be added in the future.

The yellow color that is used does appear to be the most common so let's stick to that for now. Not sure adding more colors later is a good idea - that's more logic to keep track of, and we never had complaints about Wear.

dshokouhi:
> I think the domains that can be colored list should be separate from `isActive`. Only used for coloring icons now but who knows what will be added in the future.

Agreed, good call :)  In fact we can probably apply `isActive` to tiles as we do something similar there



> The yellow color that is used does appear to be the most common so let's stick to that for now. Not sure adding more colors later is a good idea - that's more logic to keep track of, and we never had complaints about Wear.

Thats a good point!

jpelgrom:
You forgot to update the `MapVehicleScreen` to check for state colored domain in addition to active :)

> In fact we can probably apply `isActive` to tiles as we do something similar there

Good idea, nice to have it all in one place.

dshokouhi:
> You forgot to update the `MapVehicleScreen` to check for state colored domain in addition to active :)

🙈

jpelgrom (approved):

JBassett (approved):

PR #3811, opened 2023-08-22
Update deeplink support with multiserver: ask which to use

jpelgrom:
## Summary

Ask the user which server to use, instead of assuming the most recently used, when using deeplinks with an app that has multiple servers configured. This matches iOS behavior.

## Screenshots

The existing server chooser that is, for example, also visible when swiping up with 3 fingers.

Video showing the usage with my and homeassistant://navigate links:

https://github.com/home-assistant/android/assets/8148535/d01feb52-a5e7-440f-89e3-0563feea23ad

Coming soon


dshokouhi (approved):
Always good to match features :)

JBassett (approved):

PR #3816, opened 2023-08-25
Various changes to entity usage

dshokouhi:
## Summary

Possible breaking change for Tiles and Device Controls as active state may change slightly

- Use `isActive()` in Device controls, Quick setting tiles and Wear OS home screen
- Use `friendlyState()` in device controls
- Add more friendly states and link to some known HA core ones
- Add lawn mower icon
- Fix a few warnings
- Print an exception


common/src/main/res/values/strings.xml
```diff
@@ -1147,4 +1147,29 @@
...
+    <string name="state_partlycloudy">Partly cloudy</string>
+    <string name="state_playing">Playing</string>
+    <string name="state_problem">Problem</string>
+    <string name="state_pouring">Pouring</string>
+    <string name="state_rainy">Rainy</string>
+    <string name="state_snowy_rainy">Snowy rainy</string>
+    <string name="state_snowy">Snowy</string>
+    <string name="state_standby">Standby</string>
+    <string name="state_sunny">Sunny</string>
+    <string name="state_windy_variant">Windy variant</string>
```
jpelgrom, line 1173: This should simply be 'windy'

https://github.com/home-assistant/core/blob/57144a6064db0ff97cdc536580a329381bc3b1d3/homeassistant/components/weather/strings.json#L21

common/src/main/res/values/strings.xml
```diff
@@ -1147,4 +1147,29 @@
...
+    <string name="state_home">Home</string>
+    <string name="state_lightning_raining">Lightning raining</string>
+    <string name="state_lightning">Lightning</string>
+    <string name="state_not_home">Not home</string>
+    <string name="state_partlycloudy">Partly cloudy</string>
+    <string name="state_playing">Playing</string>
+    <string name="state_problem">Problem</string>
+    <string name="state_pouring">Pouring</string>
+    <string name="state_rainy">Rainy</string>
+    <string name="state_snowy_rainy">Snowy rainy</string>
```
jpelgrom, line 1169: This should include a comma

https://github.com/home-assistant/core/blob/57144a6064db0ff97cdc536580a329381bc3b1d3/homeassistant/components/weather/strings.json#L18

common/src/main/res/values/strings.xml
```diff
@@ -1147,4 +1147,29 @@
...
+    <string name="state_active">Active</string>
+    <string name="state_below_horizon">Below horizon</string>
+    <string name="state_buffering">Buffering</string>
+    <string name="state_cloudy">Cloudy</string>
+    <string name="state_exceptional">Exceptional</string>
+    <string name="state_clear_night">Clear night</string>
+    <string name="state_fog">Fog</string>
+    <string name="state_hail">Hail</string>
+    <string name="state_home">Home</string>
+    <string name="state_lightning_raining">Lightning raining</string>
```
jpelgrom, line 1161: This should include a comma

https://github.com/home-assistant/core/blob/57144a6064db0ff97cdc536580a329381bc3b1d3/homeassistant/components/weather/strings.json#L13

common/src/main/res/values/strings.xml
```diff
@@ -1147,4 +1147,29 @@
...
     <string name="state_disarming">Disarming</string>
     <string name="state_pending">Pending</string>
+    <string name="state_mowing">Mowing</string>
+    <string name="state_above_horizon">Above horizon</string>
+    <string name="state_active">Active</string>
+    <string name="state_below_horizon">Below horizon</string>
+    <string name="state_buffering">Buffering</string>
+    <string name="state_cloudy">Cloudy</string>
+    <string name="state_exceptional">Exceptional</string>
+    <string name="state_clear_night">Clear night</string>
```
jpelgrom, line 1157: This should include a comma

https://github.com/home-assistant/core/blob/57144a6064db0ff97cdc536580a329381bc3b1d3/homeassistant/components/weather/strings.json#L7

jpelgrom (commented):
Should this be listed as a breaking change as the active state of tiles/controls can change, even if it is more consistent?

dshokouhi:
> Should this be listed as a breaking change as the active state of tiles/controls can change, even if it is more consistent?

Agreed, called out in summary :)

common/src/main/res/values/strings.xml
jpelgrom: Actually we don't need to duplicate the string as it isn't using reflection, so this can be removed?

common/src/main/res/values/strings.xml
jpelgrom: Actually we don't need to duplicate the string as it isn't using reflection, so this can be removed?

dshokouhi: good call out! indeed it should be

dshokouhi: ok removed it and updated the state to use the correct string

jpelgrom (approved):
Good to share more code for similar features :)

JBassett (approved):

PR #3831, opened 2023-08-29
Shortcut path input improvements

jpelgrom:
## Summary

I noticed that when editing a shortcut, it was relatively easy to enter incorrect data. This PR updates the keyboard + handling of shortcut to hopefully prevent input errors.

## Screenshots

No app UI changes but the on screen keyboard should now be optimized for URI entry.

For example Gboard shows a `/` key for me instead of `,`:
On screen keyboard with a / key in the bottom left where there would normally be a , key

n/a


dshokouhi (approved):
Good improvement! I always forget about the keyboard options that help us in these cases :)

JBassett (approved):

Issue #2222, opened 2022-02-01
biometric prompt sometimes appears twice


regevbr:
**Home Assistant Android version:**
beta-2000-2e0700f7-full
**Android version:**
11
**Phone model:**
OnePlus 8T
**Home Assistant version:**
core-2022.2.0b5
**Last working Home Assistant release (if known):**
unknown
**Description of problem:**

I have a fingerprint lock setting set. Occasionally (but noticeable) when I open the application, I get the blurred screen and the biometric prompt created at https://github.com/home-assistant/android/blob/be3b414fa97e6b83eb5e5f40e3fd9ef701b63363/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt#L877
After I perform a successful login with my fingerprint, the blur is disabled, and then the prompt appears again immediately (with the same title, which means it has to originate from the same line of code above). At this point, I can click outside of the prompt to make it go away, or I can authenticate again, in any case, I gain access to the app (as I mentioned, the blur is already gone)

<details>
  <summary>**Traceback:**</summary>

```
01-31 23:33:01.798 28623 28623 E mpanion.androi: Invalid ID 0x00000000.
01-31 23:33:01.798 28623 28623 E mpanion.androi: Invalid ID 0x00000000.
01-31 23:33:01.806 28623 28623 V ViewRootImpl: The specified message queue synchronization  barrier token has not been posted or has already been removed
01-31 23:33:01.806 28623 28050 E sqlite3_android: ONEPLUS_NAME_PARTS_MATCH SQLITE_OK
01-31 23:33:01.836 28623 28623 W Choreographer: Already have a pending vsync event.  There should only be one at a time.
01-31 23:33:01.838 28623 28623 D UrlRepository: localUrl is: false and usesInternalSsid is: true
01-31 23:33:01.838 28623 28623 D UrlRepository: Using external URL
01-31 23:33:01.838 28623 28623 D UrlRepository: localUrl is: false and usesInternalSsid is: true
01-31 23:33:01.840 28623 31162 D UrlRepository: localUrl is: false and usesInternalSsid is: true
01-31 23:33:01.848 28623 28623 D DecorView: onWindowFocusChangedFromViewRoot hasFocus: true, DecorView@e11493f[LaunchActivity]
01-31 23:33:01.849 28623 28623 D ViewRootImpl[LaunchActivity]: windowFocusChanged hasFocus=true inTouchMode=true
[...]

PR #3842, opened 2023-09-04
Fix Duplicate Biometric prompts

RoboMagus:
## Summary

Check if there's a pending unlock action before triggering a (new) biometric prompt to unlock the app. 

Fix for #2222.

## Any other notes

This is definately a workaround, but as we have no influence on the `onWindowFocusChanged` calls being produced it seems to be a necessary one.


jpelgrom:
Shouldn't this also be added to the settings activity?

RoboMagus:
I've only been able to reproduce the issue when cold starting the app. So it should not be possible for this to be an issue in the settings.

jpelgrom (approved):
Thanks for looking into this!

Personally I'm unable to reproduce the original issue and thus verify the fix, but the changes seem to be the right ones for the issue and app lock still works as expected.

Issue #3566, opened 2023-06-01
Support Android 14's new Device Control feature to embed a custom activity 


Hupro:
**Is your feature request related to a problem? Please describe.**
Home Assistant currently supports the Android's Device Control feature which works great. Android 14 (and possibly 13 QPR3) adds a new feature to allow an app to add a custom embedded activity as a part of Device Controls called [META_DATA_PANEL_ACTIVITY](https://developer.android.com/reference/android/service/controls/ControlsProviderService#META_DATA_PANEL_ACTIVITY).  

**Describe the solution you'd like**
Allow for a users lovelace dashboard to be an embedded activity when opening Android' Device Control feature to provide a more consistent experience.    

**Describe alternatives you've considered, if any**
Leaving the Device Control implementation as is. The current implementation works good but I believe supporting this new feature when it's ready would make the experience nicer.   

**Additional context**
This feature has been shown with the Google Home app on the upcoming Pixel tablet. According to this [9to5Google article](https://9to5google.com/2023/05/10/google-pixel-home-panel-controls/), Pixel devices should be getting this feature this month so it seems like the feature may start working with Android 13 QPR3 and may not require Android 14.  [This article also shows](https://9to5google.com/2023/03/28/google-home-device-controls-redesign/) the Google Home app initially testing the change and it working on the 13 QPR3 beta.

jpelgrom:
I think this should be presented as an option controlled by the user to prevent a breaking change, if possible. 

(Personally I also like the more focused, super fast native experience instead of having to build a dashboard.)

dnestico:
+1 please support this!

jpelgrom:
aka ['Home Panel' for Google Pixels with QPR3](https://blog.google/products/pixel/feature-drop-june-2023/#:~:text=Quick%20access%20to%20smart%20home%20controls)

jpelgrom:
Working on this right now.

Looks like you actually need to remove all existing controls before the new panel activity is shown. Still, the app settings will allow you to choose between the existing controls with multiserver/lock screen integration, or a dashboard panel as controls.

https://github.com/home-assistant/android/assets/8148535/d5440098-8053-4fcf-aadb-7889a074e5d2

dnestico:
Awesome man, it looks like it's working great! 👍

I might create a subview just for this so that way the top nav bar disappears, but that's imo.

PR #3855, opened 2023-09-10
Add Android 14 device controls panel

jpelgrom:
## Summary

Implements #3566

## Screenshots

|Device controls panel|Settings*, light mode|Settings*, dark mode|
|-----|-----|-----|
|Device controls panel showing the Home Assistant app with a dashboard, instead of the previous built-in style|Home Assistant app 'Manage device controls' screen, with two radio buttons: Built-in and Dashboard, allowing the user to choose between them, light mode|Home Assistant app 'Manage device controls' screen, with two radio buttons: Built-in and Dashboard, allowing the user to choose between them, dark mode|

\* The warning about removing previous controls only shows up when changing the setting from built-in to dashboard

Documentation: home-assistant/companion.home-assistant#978

## Any other notes

Testing note: transitioning activities like done here works but clearly it wasn't the intended use case and there may be some flickering when the device is locked. Finishing in `onPause()` instead of immediately after launching seems to reduce it but on quick, repeated launches I could still get some odd behavior (tested on Android 14 beta 5.3).

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -243,7 +243,7 @@
             android:exported="true"
             android:resizeableActivity="true"
             android:taskAffinity="io.homeassistant.companion.android.controls"
-            android:enabled="true" />
+            android:enabled="false" />
 
         <receiver
             android:name=".sensors.LocationSensorManager"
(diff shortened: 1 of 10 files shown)


dshokouhi (approved):
Looks good to me!

JBassett (approved):

PR #3862, opened 2023-09-14
Fix isActive for locks

aronsky:
## Summary
Lock entity should be active when locked, inactive otherwise

## Screenshots
*irrelevant*

*irrelevant*

## Any other notes
A recent PR (#3816) made changes to how different entities and their states are represented. This change caused the locks to be represented by active (on) tiles when *unlocked*, and inactive (off) tiles when locked. This new representation is counter-intuitive, and contradicts the lock representation in tiles provided by other home automation software (e.g. Google Home).

This PR reverts this change at its core (the logic of `isActive` for lock entities), while keeping the rest of the changes in #3816 intact.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
@@ -829,6 +829,7 @@ fun <T> Entity<T>.isActive() = when {
     (domain == "cover") -> state != "closed"
     (domain in listOf("device_tracker", "person")) -> state != "not_home"
     (domain == "lawn_mower") -> state in listOf("mowing", "error")
+    // on Android, contrary to HA Frontend, a lock is considered active when locked
     (domain == "lock") -> state == "locked"
     (domain == "media_player") -> state != "standby"
     (domain == "vacuum") -> state !in listOf("idle", "docked", "paused")
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
@@ -829,7 +829,7 @@ fun <T> Entity<T>.isActive() = when {
     (domain == "cover") -> state != "closed"
     (domain in listOf("device_tracker", "person")) -> state != "not_home"
     (domain == "lawn_mower") -> state in listOf("mowing", "error")
-    (domain == "lock") -> state != "locked"
+    (domain == "lock") -> state == "locked"
     (domain == "media_player") -> state != "standby"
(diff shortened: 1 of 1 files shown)


dshokouhi:
If you read the actual PR #3805 we listed it as a breaking change in #3836 because this makes the app more in line with what the frontend considers as active and we should actually match the frontend here.

https://github.com/home-assistant/frontend/blob/dev/src/common/entity/state_active.ts#L40

aronsky:
I see.

And I also understand the desire to simplify code, and to be in line with what the frontend is doing. But specifically for locks, and the tiles on Android, the resulting behavior is counter-productive. As mentioned, other home automation products display lock tiles as active when they are *locked*. Maybe the real fix would be to change the frontend logic that you linked to, as well? I would argue that a lock's purpose is to, well, lock things - so it's active when locked, and inactive when unlocked.

dshokouhi:
> Maybe the real fix would be to change the frontend logic that you linked to, as well?

Yes this would be the correct approach before we should accept this PR because we do base our logic on what HA users expect and not what the expectation is from other apps. That is also the reason why we link to frontend logic in several places so its easier for us to update and keep things in sync. This is also why we highlighted it as a breaking change as we are aware some users may not like it.

common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -829,7 +829,7 @@ fun <T> Entity<T>.isActive() = when {
     (domain == "cover") -> state != "closed"
     (domain in listOf("device_tracker", "person")) -> state != "not_home"
     (domain == "lawn_mower") -> state in listOf("mowing", "error")
-    (domain == "lock") -> state != "locked"
+    (domain == "lock") -> state == "locked"
```
dshokouhi, line 832: After having some discussion and another user mentioning this in Discord we have noticed that the lock icon which is controlled by Android actually matches your suggested behavior.

Can you please add a comment to this line mentioning that we are deviating from the frontend to better match Android native behavior? This way we can remember that we need to keep this logic.

image

https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/controls/LockControl.kt#L40

dshokouhi (changes requested):
Just need to add a comment so we can remember why we explicitly changed the logic from the fronted.

You can also close your frontend PR too

jpelgrom:
Linked another issue to this PR, as the action will no longer be inverted if this is merged

dshokouhi (approved):
Thanks for the adding the comment. As the requested changes are now completed I will put this PR ready for review so we can merge it.

JBassett (approved):

PR #3875, opened 2023-09-18
Automatically fix broken registrations/webhooks

jpelgrom:
## Summary

After helping a few users recently with broken webhooks causing issues, as the app appears to be working but anything depending on the webhook doesn't, I decided to implement an automatic fix for this. It's easy to recognise in the log files so it also should be easy to handle with code:
- HTTP 200 with empty body: direct connection broken webhook
- HTTP 404: broken cloudhook
- HTTP 410: no config entry (and as a result, also a broken webhook)

By doing a new registration we get a new webhook. Removes the need for the user to log out and back in again.

[The iOS app does something very similar](https://github.com/home-assistant/iOS/blob/e065d9bfe05bf1a179a729c3f63ec7745f5b3446/Sources/Shared/API/HAAPI.swift#L184-L197).

## Screenshots

n/a

n/a


dshokouhi (approved):
This is great, good to know iOS did something similar too :)

JBassett (approved):

PR #3876, opened 2023-09-18
Add smooth scrolling for rotary input

Dubzer:
## Summary
This pull request adds smooth scrolling for rotary input. That's especially important on Galaxy Watch Classic models, since it has "low-res" input.
Unfortunately, unlike views, Compose for Wear OS still doesn't have smooth scrolling for lists by default, but there's a [horologist](https://github.com/google/horologist) library by Google, which has an implementation of it. That's what I used.

## Screenshots

https://github.com/home-assistant/android/assets/18244287/97b950a1-d886-4e2d-9b32-49e86e1dee3f

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -64,6 +64,7 @@ wearInput = "1.2.0-alpha02"
 webkit = "1.8.0"
 wear-remote-interactions = "1.0.0"
 workRuntimeKtx = "2.8.1"
+horologist = "0.5.5"
 
 [plugins]
 android-application = { id = "com.android.application", version.ref = "androidPlugin" }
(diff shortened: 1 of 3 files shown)


dshokouhi:
This is great! I wonder if we should also apply it to other areas like the refresh interval picker? Noticed the other day it felt a bit extra jumpy using rotary input :)

https://github.com/home-assistant/android/blob/2c20baf0f54897c720099b34e57cf78eaa70fedb/wear/src/main/java/io/homeassistant/companion/android/home/views/RefreshIntervalPickerView.kt#L63-L71

Dubzer:
> I wonder if we should also apply it to other areas like the refresh interval picker?

This behavior is also implemented in Horologist. There's a modifier called [rotaryWithSnap](https://github.com/google/horologist/blob/aa6151a8b04f4da0249becc0d6b9eaa9955b47e6/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt#L181C21-L181C35), which is [used](https://github.com/google/horologist/blob/aa6151a8b04f4da0249becc0d6b9eaa9955b47e6/composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt#L574C24-L574C24) by similarly designed component - [TimePicker](https://google.github.io/horologist/composables/#time-pickers).

But I'd like to keep this in a separate PR as it looks like a loosely related changes

dshokouhi:
> But I'd like to keep this in a separate PR as it looks like a loosely related changes

absolutely, I wasn't sure when I looked at the code recently :)

dshokouhi (approved):

JBassett (approved):

PR #3883, opened 2023-09-21
Add helper class for ordered maps

NotWoods:
## Summary
Simplifying the compose state in wear by creating a helper class for the common pattern of associating a map and a list of keys.

This change introduces a new class, `OrderedMap`, that is just a map with an `orderedKeys` property. It also includes some helper functions for creating ordered maps and iterating through them.

## Screenshots
N/A, not user facing change


jpelgrom:
This change appears to break Compose's state change detection, for example tapping on a light will toggle it but the chip doesn't update. Did you test the code? It is quite easy to notice.

Also for future contributions: there are more than a few changes unrelated to the ordered maps in code (filter variables, item keys, ...) which make it harder to review. Not against them, they are welcome, but if you do include improvements like these try to put them in a separate commit / at least mention them 😊

jpelgrom:
Closing due to no response / conflicting changes

PR #3886, opened 2023-09-23
Add troubleshooting tool: location history

jpelgrom:
## Summary

This PR adds a troubleshooting tool for location tracking: a persistent location history (up to 48 hours, enabled by default). 

Location updates on Android are hard to get your head around and the current troubleshooting steps with logs aren't very user friendly. This hopefully makes it easier for users to review and for us as developers to find things to improve. The location history is specifically intended as a troubleshooting tool, which is why there is limited UI (no maps for now!) and the updates are only stored 48 hours. If someone wants to do more, it should probably be handled on the server.

UI inspiration is Android's built-in notification history, trying to get more M3 elements in the app :)

## Screenshots

|     |Light|Dark|
|-----|-----|-----|
|Setting in the troubleshooting menu|Troubleshooting menu with entry 'Location tracking: View a history of location tracking updates to troubleshoot the device tracker(s)', light mode|Troubleshooting menu with entry 'Location tracking: View a history of location tracking updates to troubleshoot the device tracker(s)', light mode|
|History, tap on a row to expand|A list of rows with various times and subtitles like 'Background location', 'Sent/Old location/Duplicate', one item is expanded and shows a latitude/longitude, accuracy, and buttons to open a map and share, light mode|A list of rows with various times and subtitles like 'Background location', 'Sent/Old location/Duplicate', one item is expanded and shows a latitude/longitude, accuracy, and buttons to open a map and share, dark mode|
|Empty state, off|'Use location history' toggle is off, text 'Location history is off: Turn on location history to see and review all location updates received by the app', light mode|'Use location history' toggle is off, text 'Location history is off: Turn on location history to see and review all location updates received by the app', dark mode
|Empty state, no updates|Screen with 'No recent locations: Location updates received by the app will appear here', light mode|Screen with 'No recent locations: Location updates received by the app will appear here', dark mode|

<details>
<summary>Sharing converts the update to text with details untranslated, good for issues</summary>

```
Created: 2023-09-23 11:18:36.512
Trigger: FLP_BACKGROUND
Result: SENT
Data: UpdateLocation(gps=[5lat, lng], gpsAccuracy=100, locationName=null, speed=0, altitude=46, course=0, verticalAccuracy=100)

Location: 5lat, lng
Accuracy: 100
Server: Home
```
</details>

Documentation: home-assistant/companion.home-assistant#987

## Any other notes

The iOS app also does some [location history logging](https://github.com/home-assistant/iOS/blob/e065d9bfe05bf1a179a729c3f63ec7745f5b3446/Sources/Shared/API/HAAPI.swift#L516-L530) but seemingly doesn't have the need to filter updates.

For the future I can imagine a status check at the top of the screen that tells you "everything is OK" or "setting X isn't optimal", which is why this screen is called location tracking and not location history.


app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -859,23 +895,27 @@ class LocationSensorManager : LocationSensorManagerBase() {
...
                 }
             }
         } else {
             Log.d(TAG, "Skipping location update due to old timestamp ${location.time} compared to $now")
+            logLocationUpdate(location, updateLocation, serverId, trigger, LocationHistoryItemResult.SKIPPED_OLD)
             return
         }
         lastLocationSend[serverId] = now
         lastUpdateLocation[serverId] = updateLocationString
+        logLocationUpdate(location, updateLocation, serverId, trigger, LocationHistoryItemResult.SENT)
```
dshokouhi, line 918: I wonder should we consider logging it here as sent or should we consider also logging when we fail to send the location update down below in the `try/catch`?

jpelgrom: Could be, but would it be very useful to know that the location was OK but it was unable to send it? For example depending on the device and server config the server being offline can produce different exceptions, so it's harder to make the error more specific without storing the raw exception (which I'd avoid).

dshokouhi: Well I think Sent or Failed to Send would indicate that its a valid report to be sent to the server and its either successful or not. I dont think we need to store the exception just the frequency of failures should be enough to go by.

jpelgrom: Updated, it also better matches log output.
Screenshot showing a row with 'Failed to send' and a small icon of an alert sign

dshokouhi (approved):
Looks good to me, been testing it for a few days. This should be very helpful to users who have issues with location tracking.

JBassett (approved):

PR #3888, opened 2023-09-23
Update icon

jpelgrom:
## Summary

 - Replace old icon with [the new icon](https://www.home-assistant.io/blog/2023/09/17/a-refreshed-logo-for-home-assistant/)
 - Deduplicate icon resources

## Screenshots

Everywhere the Home Assistant icon was, or the primary color was used, it should now be changed. For example in the launcher, old vs new:
Old Home Assistant icon on the left, new Home Assistant icon on the right

And as a notification icon:
Status bar with various notification icons, including one from Home Assistant

n/a


jpelgrom:
Changing to draft to receive clarification on what the frontend will do re. primary color

LasseRosenow:
I think now that we have a new logo, maybe we should discuss the "centering in a circle" problem.

On the old icon it was much more obvious, but still the new icon still suffers this problem I think.
Even though it is `technically` centered, it does not look quite centered.

There is a very good post on medium about this:
https://medium.com/@erqiudao/the-play-button-is-not-optical-alignment-4cea11bda175

So in general we should try to center the center of the house by drawing a circle around it and using the circles center point as the center and not by drawing a rectangle and centering this :)

How do you think about this?

LasseRosenow:
I created an example of how the icon would look using a circle to center it:

| Current | "Fixed" |
|---|---|
|  launcher_new_broken | launcher_new_fixed  |
| launcher_new_broken_red_border  |  launcher_new_fixed_red_border |

jpelgrom:
Thanks. Turns out they did create a circle-specific design I wasn't aware of.

dshokouhi:
Should we also update our play store listing icon with this PR?

jpelgrom:
It is also updating the listing icon, isn't it?

dshokouhi:
yes it is, shouldve searched before commenting 🤦

jpelgrom:
Frontend isn't changing primary color, decoupled.

Madelena:
Looking good 👍

dshokouhi (approved):

JBassett (approved):

PR #3895, opened 2023-09-27
Add missing scroll bar to remaining screens

dshokouhi:
## Summary

We had a play store review failure. One of the failing items was missing scroll bars when a user is scrolling.

To fix this I am adding `Scaffold` to `ThemeLazyColumn` to ensure we always show the scroll bars and time

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/conversation/views/ConversationView.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/conversation/views/ConversationView.kt
@@ -271,17 +271,29 @@ fun ConversationPipelinesView(
     onSelectPipeline: (String) -> Unit
 ) {
     WearAppTheme {
-        ThemeLazyColumn {
-            item {
-                ListHeader(stringResource(R.string.assist_change_pipeline))
-            }
-            items(items = pipelines, key = { it.id }) {
-                Chip(
-                    modifier = Modifier.fillMaxWidth(),
-                    label = { Text(it.name) },
-                    onClick = { onSelectPipeline(it.id) },
-                    colors = ChipDefaults.secondaryChipColors()
-                )
+        val scrollState = rememberScalingLazyListState()
+        Scaffold(
+            positionIndicator = {
+                if (scrollState.isScrollInProgress) {
+                    PositionIndicator(scalingLazyListState = scrollState)
+                }
+            },
+            timeText = { TimeText(scalingLazyListState = scrollState) }
+        ) {
+            ThemeLazyColumn(
+                state = scrollState
+            ) {
+                item {
+                    ListHeader(stringResource(R.string.assist_change_pipeline))
+                }
+                items(items = pipelines, key = { it.id }) {
+                    Chip(
+                        modifier = Modifier.fillMaxWidth(),
+                        label = { Text(it.name) },
+                        onClick = { onSelectPipeline(it.id) },
+                        colors = ChipDefaults.secondaryChipColors()
+                    )
+                }
             }
         }
     }
--- a/wear/src/main/java/io/homeassistant/companion/android/conversation/views/ConversationView.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/conversation/views/ConversationView.kt
@@ -43,7 +43,6 @@ import androidx.wear.compose.material.ButtonDefaults
 import androidx.wear.compose.material.Chip
 import androidx.wear.compose.material.ChipDefaults
 import androidx.wear.compose.material.LocalContentColor
-import androidx.wear.compose.material.Scaffold
 import androidx.wear.compose.material.Text
 import androidx.wear.compose.navigation.SwipeDismissableNavHost
 import androidx.wear.compose.navigation.composable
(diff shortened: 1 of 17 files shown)


jpelgrom:
When adding a `Scaffold` anywhere we use a `ThemeLazyColumn`, doesn't it make more sense to create a new Composable with Scaffold + ThemeLazyColumn?

dshokouhi:
> When adding a `Scaffold` anywhere we use a `ThemeLazyColumn`, doesn't it make more sense to create a new Composable with Scaffold + ThemeLazyColumn?

Honestly I thought about adding the Scaffold to the Lazy Column as we should always use the 2 together no matter what. In reality WearAppTheme + Scaffold + ThemeLazyColumn should always be used together right?

Not sure if we should combine them or just continue to keep it separate but IMO we use it as a theme to show the scroll bar and time in the lazy column while scrolling so the 3 feel very much related.

jpelgrom:
WearAppTheme is not always followed by a scaffold (eg navigation), but it's a function so you can create multiple and use them as needed :) Combining Scaffold + ThemeLazyColumn would make sense because they both use the same state for the scroll indicator.

dshokouhi:
> Combining Scaffold + ThemeLazyColumn would make sense because they both use the same state for the scroll indicator.

completed in the latest commit

wear/src/main/java/io/homeassistant/companion/android/conversation/views/ConversationView.kt
```diff
@@ -118,14 +116,7 @@ fun ConversationResultView(
...
 
-    Scaffold(
-        positionIndicator = {
-            if (scrollState.isScrollInProgress) {
-                PositionIndicator(scalingLazyListState = scrollState)
-            }
-        },
-        timeText = { TimeText(scalingLazyListState = scrollState) }
-    ) {
+    Scaffold {
```
jpelgrom, line 119: Because the `ThemeLazyColumn` is updated to include a `Scaffold` here you have nested Scaffolds, don't think that is correct.

wear/src/main/java/io/homeassistant/companion/android/conversation/views/ConversationView.kt
```diff
@@ -271,7 +262,10 @@ fun ConversationPipelinesView(
     onSelectPipeline: (String) -> Unit
 ) {
     WearAppTheme {
-        ThemeLazyColumn {
+        val scrollState = rememberScalingLazyListState()
```
jpelgrom, line 265: Not used anywhere else, can be removed (default parameter of `ThemeLazyColumn`)

jpelgrom (commented):
A general pattern for composables is:

```kotlin
val scalingLazyListState = rememberScalingLazyListState()
WearAppTheme {
    ThemeLazyColumn(state = scalingLazyListState) {
        ...
    }
}
```

But as the `ThemeLazyColumn` now includes a `Scaffold`, which was almost always the only reason to get the state in a variable, can't it be removed in a lot of places?

wear/src/main/java/io/homeassistant/companion/android/conversation/views/ConversationView.kt
dshokouhi: I think we need this code in the `Scaffold` but it did load so I thought it was a loading based on a hierarchy

dshokouhi:
> A general pattern for composables is:
> 
> ```kotlin
> val scalingLazyListState = rememberScalingLazyListState()
> WearAppTheme {
>     ThemeLazyColumn(state = scalingLazyListState) {
>         ...
>     }
> }
> ```
> 
> But as the `ThemeLazyColumn` now includes a `Scaffold`, which was almost always the only reason to get the state in a variable, can't it be removed in a lot of places?

Do you mean to stick with the default which also gets the latest state? I think so but was used to always grabbing the state and sending it in :)

jpelgrom:
> Do you mean to stick with the default which also gets the latest state? I think so but was used to always grabbing the state and sending it in :)

I think simply putting `ThemeLazyColumn { ... }` (using the default parameter) would work.

wear/src/main/java/io/homeassistant/companion/android/conversation/views/ConversationView.kt
jpelgrom: Which `Scaffold` are you talking about? This is getting confusing

dshokouhi: I am talking about the specific `Scaffold` code here, the `LaunchedEffect` cannot be moved to the `ThemeLazyColumn` so while there is a nested `Scaffold` I think thats the only way to get it to work?

jpelgrom: But the `LaunchedEffect` can still be outside the `ThemeLazyColumn`?

```kotlin
val scrollState = rememberScalingLazyListState()
LaunchedEffect(key) { ... }
if (hapticFeedback) { ... }
ThemeLazyColumn(state = scrollState) { ... }
```

dshokouhi: oh my bad thought it had to be in the `Scaffold` but looks like it just needed a composable scope :)

dshokouhi: done, thank for the suggestion :)

jpelgrom (approved):
Installed and tested, I think you got all screens now. Love how simple something like the sensor list is with these changes.

JBassett (approved):

PR #3896, opened 2023-09-27
Tweak design to avoid cutting off text when largest font size is used

dshokouhi:
## Summary

Our other play store review failure is due to text getting cut off, this PR attempts to resolve that.

We had some text getting cut off when the largest font size was set in Wear OS settings. We need to make sure none of the text is getting cut off on the screen otherwise we will fail play store review.

- Details pane has the toggle moved to its own `item` and a spacer added to the top to avoid text getting cut off and the button going off screen
- Main view error message pushed down a bit with padding like loading message to avoid text getting cut off
- Settings view added spacers to first element to avoid text getting cut off

## Screenshots

BEFORE (hard to tell but images taken from a rounded screen:

In the below screen the toggle button is not on screen and the beginning of the entity name gets cut off in the rounded screen
image

In the below screen the list header is getting cut off in the rounded screen
image

AFTER
image

image

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/home/views/DetailsPanelView.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/views/DetailsPanelView.kt
@@ -2,13 +2,12 @@ package io.homeassistant.companion.android.home.views
 
 import android.content.Context
 import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
-import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.ColorFilter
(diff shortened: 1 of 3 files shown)


jpelgrom:
Your fix is... more padding? How would this help on a larger watch, or for languages with longer strings? Now there are also a few headers at random that look different.

dshokouhi:
The idea is to fix the starting position so its not at the top of the page where the cut off issue is more prevalent. Its hard to tell exactly what is being cut off based on the review email. I don't see how random headers can be impacted as the changes are only done to the top of the Details view and Settings view and when entities fail to load on the main view.

jpelgrom:
But by that logic shouldn't every screen's first header be adjusted? German probably has longer headers, or area names can be long, or ...

dshokouhi:
So I didn't notice it getting cut off when the font size was larger on all screens just the ones I submitted fixes for.

dshokouhi:
I think at this point the only remaining screen would be the home page 🤔 the other screens seem to start more in the middle of the screen instead of at the top where the cut off happens.  Really wish some screenshots were shared to know what exactly needs to be fixed.

jpelgrom:
Still doesn't feel like a permanent solution and error prone...

Eg for favorite settings: why isn't the text centered and wouldn't that potentially prevent the issue there?

dshokouhi:
> Still doesn't feel like a permanent solution and error prone...

I am not really sure whats more appropriate here, would be a lot easier to know what the issue is instead of hoping this is actual the issue :(

> Eg for favorite settings: why isn't the text centered and wouldn't that potentially prevent the issue there?

it is centered but when using the largest font its no longer centered and appears like the screenshot

dshokouhi:
So I do think we should add some padding to the top in order to prevent the beginning/top of of the page to be cut off. I just can't think of what else is referred to when they mention text is cut off. Even our app version which can be long does not get cut off. Entity details does not get cut off and wraps properly. I only see the beginning of text at the top of the page getting cut off.

JBassett:
FWIW I ended up using the alpha/beta material 3 components for work because they handled the font size changes better.

dshokouhi:
> FWIW I ended up using the alpha/beta material 3 components for work because they handled the font size changes better.

Is Material 3 supported on Wear OS yet? Seems like a pretty large change for us to switch to Material 3 possibly if we can?

dshokouhi:
oops disregard my last comment, its here:

https://developer.android.com/jetpack/androidx/releases/wear-compose#wear_compose_material3_version_10_2

dshokouhi:
going to close this PR as the attempt does not seem to be acceptable and moving to material 3 is more involved than what this PR tries to accomplish.

PR #3899, opened 2023-09-28
Update Wear more details screen to prevent cut off

jpelgrom:
## Summary

Updates the more details screen on a watch to prevent content from being cut off. The entity title and toggle were forced onto one line, so with longer entity names and/or larger text sizes it would get cut off.

New design prominently shows the entity icon, which can be tapped to toggle, and the entity title on another line (which can wrap / limit length). Design is very similar to the frontend tile card and default Wear round buttons with icons, so hopefully familiar to users. Colors are not dynamic - on is always yellow, matching the vehicle interface.

## Screenshots

|Before|After, off|After, on|After, no toggle*|
|-----|-----|-----|-----|
|More details screen with text and toggle immediately at the top cut off by the round screen|More details screen with a lightbulb icon in a circle that can be tapped, grey for off|More details screen with a lightbulb icon in a circle that can be tapped, yellow for on|More details screen with a palette icon for a scene, which is not in a circle indicating it cannot be tapped|

\* and default text size

n/a


dshokouhi (approved):
Great improvement!

JBassett (approved):

PR #3904, opened 2023-10-01
Begin migrating Wear OS components to Material 3

dshokouhi:
## Summary

Begin migration from Material 2 to Material 3 in Wear OS.

I have used the Material theme builder using the starting recommendation https://developer.android.com/jetpack/compose/designsystems/material2-material3#color

Not everything can be migrated just yet as Iconics does not support the new material 3 changes, I have opted to switch to material icons in places I was able to. Others I opted to keep the `Button` the same until the library is updated.

For this update I have migrated any components that could be migrated without additional changes. I have also migrated material 2 buttons to material 3.

Additional corrections:

- in `SettingsView` several variables were incorrectly used so I have updated
- in `SettingsView` switched from checkboxes to switches for consistency
- Cleaned up some warnings on pages that have been updated

We still need to migrate `Chip` to `Button` and `ToggleChip` to `ToggleButton`. I am not sure if we will be able to migrate `WearToggleChip` just yet.
## Screenshots

Toggle chip colors when selected are different
image

WearToggleChip colors are different
image

Details pane colors are different, new sliders
image

Example of how text is no longer bold by default
image

## Any other notes

Placing screen shots of fixes  here:

Wear toggle chip fixed (for the most part)
image

Settings toggle chip
image

details pane
image


dshokouhi:
Started reading the release notes more closely now. Going to do another update and migrate from `Chip` to `Button` based on the following.

`We have added the Material 3 Button component - this is our stadium-shaped button and was formerly named Chip in the Wear Compose Material library (it has been renamed to Button for consistency with the Compose Material 3 library). `

https://developer.android.com/jetpack/androidx/releases/wear-compose#1.0.0-alpha03

I may also look into replacing `ToggleChip` with `ToggleButton` if possible too.

dshokouhi:
after spending some time on the above comment thats going to take a lot of time and probably the review will be impacted. I will focus on cleaning up the current state unless we find other areas that should be addressed. We should migrate all features to material 3 but we dont have to rush to do them all.

JBassett (approved):

PR #3906, opened 2023-10-02
Switch to Wear Device specific preview

dshokouhi:
## Summary

Noticed that our preview device was deprecated so switching to the actual preview device we should be using.


jpelgrom (approved):
Deprecated but the alternative is something in its first alpha release... Good thing it's copy and paste and only used for development.

JBassett (approved):

JBassett:
> Deprecated but the alternative is something in its first alpha release... Good thing it's copy and paste and only used for development.

It's the Google way, leap before you look!

PR #3909, opened 2023-10-02
Update manual onboarding screens to prevent the button from overlapping elements

dshokouhi:
## Summary

Noticed when largest font size is used and user attempts to manually enter the URL using Advanced option when app is not installed the next button overlaps the text field.

Moved to `ScrollView` to allow the screen to be scrollable.

Increased min height so the tap area is larger per recommendation in android studio

Used the same manual URL string as the phone app as the current string was odd

Add padding if the device is round based on: https://stackoverflow.com/a/49338298

## Screenshots

BEFORE
image

image

AFTER

image

image

image

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationActivity.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationActivity.kt
@@ -10,6 +10,7 @@ import androidx.wear.activity.ConfirmationActivity
 import dagger.hilt.android.AndroidEntryPoint
 import io.homeassistant.companion.android.databinding.ActivityIntegrationBinding
 import io.homeassistant.companion.android.home.HomeActivity
+import io.homeassistant.companion.android.util.adjustInset
 import javax.inject.Inject
 import io.homeassistant.companion.android.common.R as commonR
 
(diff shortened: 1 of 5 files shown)


jpelgrom:
Adjusting padding to fit what is currently the smallest screen isn't a fix. This does not work for longer strings (for example, if the language requires more words). The screen should scroll.
image

dshokouhi:
I opted to limit the text 2 lines. These screens are very basic. Server URL and Device Name. It really does not need to scroll to be descriptive and in fact the provided hints and supplied device name really make it easy to understand.

dshokouhi:
temporarily convert to draft while I test a different approach

JBassett:
IMO scrolling should almost always be an option depending on text and screen size.

dshokouhi:
I am working on moving away from `BoxInsetLayout` which does not allow scrolling to a `ScrollView`

dshokouhi:
ok think this PR now addresses all concerns and issues found on the page, really hope this was the cut off that was referenced.

jpelgrom:
+ you need to hide the main view with input now while the loading view is shown because it'll put it at the bottom instead of on top in a `ScrollView` vs the previous box/frame setup.

jpelgrom (approved):

JBassett (approved):

PR #3917, opened 2023-10-05
Update `Chip` and `ToggleChip` to `Button` and `ToggleButton` except for `WearToggleChip`

dshokouhi:
## Summary

Updates  `Chip` to `Button` and `ToggleChip` to `ToggleButton` respectively

Also bumps material3 library to latest version

Adds a preview for a missing button

## Screenshots

ignore background colors as the screenshot is from the compose preview whcih doesnt get it quite right

image


wear/src/main/java/io/homeassistant/companion/android/theme/Color.kt
```diff
@@ -44,3 +46,11 @@ internal val wearColorScheme: ColorScheme = ColorScheme(
     outline = md_theme_dark_outline,
     outlineVariant = md_theme_dark_outlineVariant
 )
+
+@Composable
```
dshokouhi, line 50: I opted to add this here as its a theme color for the switch and might be easier to reference if it was all in the same location? I can move it if we feel thats better.

jpelgrom: Makes sense as a location for now

wear/src/main/java/io/homeassistant/companion/android/util/ContentDescription.kt
```diff
@@ -0,0 +1,14 @@
+package io.homeassistant.companion.android.util
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.res.stringResource
+import io.homeassistant.companion.android.common.R
+
+@Composable
+fun getDescription(isChecked: Boolean): String {
```
jpelgrom, line 8: This should be more descriptive, eg `getSwitchDescription`

wear/src/main/java/io/homeassistant/companion/android/complications/views/ComplicationConfigMainView.kt
```diff
@@ -133,54 +131,44 @@ fun MainConfigView(
...
-                    toggleControl = {
-                        Icon(
-                            imageVector = ToggleChipDefaults.switchIcon(isChecked),
-                            contentDescription = if (isChecked) {
-                                stringResource(R.string.enabled)
-                            } else {
-                                stringResource(R.string.disabled)
+                    label = { Text(stringResource(R.string.show_unit_title)) },
+                    selectionControl = {
+                        Switch(
```
jpelgrom, line 161: This seems quite standard, so maybe it's worth putting the entire `Switch` with checked, modifier and colors in a Composable?

wear/src/main/java/io/homeassistant/companion/android/complications/views/ComplicationConfigMainView.kt
```diff
@@ -133,54 +131,44 @@ fun MainConfigView(
...
+                                this.contentDescription = description
                             },
-                            tint = if (isChecked) wearColorScheme.tertiary else wearColorScheme.onSurface
+                            colors = switchColors()
                         )
                     },
                     modifier = Modifier.fillMaxWidth(),
-                    enabled = loaded && entity != null
+                    enabled = loaded && entity != null,
+                    colors = ToggleButtonDefaults.toggleButtonColors(checkedContainerColor = wearColorScheme.outlineVariant)
```
jpelgrom, line 150: This seems like it will be used a lot, I'd also put this in a function.

wear/src/main/java/io/homeassistant/companion/android/complications/views/ComplicationConfigMainView.kt
dshokouhi: ah good point :)

wear/src/main/java/io/homeassistant/companion/android/complications/views/ComplicationConfigMainView.kt
dshokouhi: was considering that too, will add

wear/src/main/java/io/homeassistant/companion/android/util/ToggleSwitch.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
+import androidx.wear.compose.material3.Switch
+import io.homeassistant.companion.android.theme.getSwitchColors
+
+@Composable
+fun getToggleSwitch(isChecked: Boolean, description: String) = Switch(
```
jpelgrom, line 11: This should be called `ToggleSwitch` or something like `WearSwitch` (it is a variation on `Switch` just like they do internally). If this is the only place 'our' switch colors are used it could be specified here again instead of breaking it out in a function.

dshokouhi: file has been renamed as we now also have checkboxes in here

wear/src/main/java/io/homeassistant/companion/android/util/ToggleButtonControls.kt
```diff
@@ -0,0 +1,28 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
+import androidx.wear.compose.material3.Checkbox
+import androidx.wear.compose.material3.Switch
+import io.homeassistant.companion.android.theme.getCheckboxColors
+import io.homeassistant.companion.android.theme.getSwitchColors
+
+@Composable
+fun GetToggleSwitch(isChecked: Boolean, description: String) = Switch(
```
jpelgrom, line 13: The `description` is always based on `isChecked`, so you could remove it from the constructor + put the if/else directly in here as it isn't used elsewhere.

wear/src/main/java/io/homeassistant/companion/android/home/views/MainView.kt
```diff
@@ -321,7 +318,7 @@ fun MainView(
                                             }.sorted()
                                         ) { true }
                                     },
-                                    colors = ChipDefaults.secondaryChipColors()
+                                    colors = ButtonDefaults.filledTonalButtonColors()
```
jpelgrom, line 321: These buttons have no visible background? Wouldn't it make more sense to have the same buttons as the settings items, like we did before? Same applies for sensor managers.
image

dshokouhi: ah yes will fix in next commit :)

wear/src/main/java/io/homeassistant/companion/android/util/ToggleButtonControls.kt
dshokouhi: thanks this is now done

jpelgrom:
This is starting to grow on me ❤️

Before merging the difference in disabled states should be resolved in my opinion.
image

wear/src/main/java/io/homeassistant/companion/android/home/views/SetShortcutsTileView.kt
```diff
@@ -69,7 +68,7 @@ fun SetShortcutsTileView(
                         )
                     },
                     onClick = { onShortcutEntitySelectionChange(index) },
-                    colors = ChipDefaults.secondaryChipColors()
+                    colors = ButtonDefaults.filledTonalButtonColors(containerColor = wearColorScheme.outlineVariant)
```
jpelgrom, line 71: This should also use `getFilledTonalButtonColors()`

jpelgrom (approved):

JBassett (approved):

PR #3919, opened 2023-10-05
Fix Picker rotary input on some devices and update design

jpelgrom:
## Summary

On some devices with imprecise rotary input, such as the Galaxy Watch 4 I personally use, the picker items were too small resulting in skipping every other item which is impractical (you could only select odd-indexed items using touch). This PR updates the picker design to [the design guidelines](https://developer.android.com/design/ui/wear/guides/components/pickers), which fixes this issue as the items are now larger. Code for style based on Horologist's DatePicker.

## Screenshots

|Before|After|
|-----|----|
|A Picker with smaller text, normal font weight and different color depending on position|A Picker with larger text, bold and all the same color|

n/a


dshokouhi (approved):
This was annoying on my pixel watch as well 😅

JBassett (approved):

PR #3930, opened 2023-10-11
Prevent notification attachments failing due to spaces ( ) in url

### LINKED ISSUE #3926: Phone Notification Local Image fails to download
Hi i have Samsuing and Android 13, last few months idk why but Local image notifcations not work i dont see in phone, logs are clear in phone everytinhg is good, but when i will replace it with a url everything is works good, where is the problem?, thx,

```
service: notify.mobile_app_milan_phone
data:
  message: test
  data:
    image: /media/local/Camera Snapshots/Doorbell Camera.jpg
```

image

jpelgrom:
## Summary

Fix #3926 by automatically replacing ` ` in URLs with the encoded character `%20`. Not replacing any other characters as that might break URLs including already encoded URLs. (This feels like a 'safe' change to automatically handle common user errors.)

Without this change, a notification attachment containing a space will fail the URL validation. Because of that failure it returns the base URL and downloads it (= the login page), which then cannot be processed as an image/video.

## Screenshots

n/a

n/a

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -1210,9 +1210,10 @@ class MessagingManager @Inject constructor(
         data: Map<String, String>
     ) {
         data[ICON_URL]?.let {
+            val dataIcon = it.trim().replace(" ", "%20")
             val serverId = data[THIS_SERVER_ID]!!.toInt()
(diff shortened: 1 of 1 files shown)


dshokouhi (approved):
Looks good to me and the user confirmed the actual issue too. I do agree as well this should solve the most common case.

https://github.com/home-assistant/android/issues/3926#issuecomment-1758520928

JBassett (approved):

PR #3954, opened 2023-10-19
Improve camera tile add/remove event handling

jpelgrom:
## Summary

Following reports by @dshokouhi in chats and [on Discord](https://discord.com/channels/330944238910963714/562408603345092636/1162095891210063883) of camera tiles resetting unexpectedly or not being removed from the app, I tested adding/removing camera tiles again.

For camera tiles resetting, this can be caused by `onTileAddEvent` or `onTileRemoveEvent`. This PR updates `onTileAddEvent` to not overwrite tile data if it already exists, hopefully fixing this (remove+add is annoying, the app would have to add soft delete).

For camera tiles sticking around, during testing I noticed that the system was often destroying the service almost immediately after the add/remove events were sent. As a result, code started in the service's coroutine was cancelled before completing, and a camera tile wasn't added to the database or not removed from the database. I could semi-reliably get it to destroy the service within several milliseconds of adding/removing.
Logcat example showing 10ms between event and destroy
To prevent this, making sure the database work is completed before completing the callback function seems to be the only reliable way (`runBlocking`). This PR changes it for the camera tiles, during testing it still finishes in <100ms and the app isn't visible at the time (system plays animation) so blocking isn't an issue.

## Screenshots

n/a

n/a

## Any other notes

If accepted I intend to update other tile services to do the same later (there haven't been any complaints yet, but they use other storage  which may behave differently on cancellation).

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/tiles/CameraTile.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/tiles/CameraTile.kt
@@ -28,7 +28,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.guava.future
-import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
 import okhttp3.OkHttpClient
 import okhttp3.Request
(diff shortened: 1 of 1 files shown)


dshokouhi (approved):
Thanks for the fix 🙏 ! Tough to actually test my issue since theres no direct path to reproduce and it can be days in between but I think the approach is correct and should solve some issues we've seen reported around this.

JBassett (approved):

PR #3959, opened 2023-10-20
Show sensor state in secondary label in Wear OS sensor UI

dshokouhi:
## Summary

Show sensor states in the UI to help with troubleshooting states in app compared to what HA core has. Adds more preview data to view different combination of states.

## Screenshots

image

## Any other notes

I am also considering in the future to move to `SplitToggleButton` so we can have a sensor detail screen to show attributes and sensor settings.

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
@@ -21,6 +21,7 @@ import io.homeassistant.companion.android.database.sensor.Sensor
 import io.homeassistant.companion.android.theme.getToggleButtonColors
 import io.homeassistant.companion.android.util.ToggleSwitch
 import io.homeassistant.companion.android.util.batterySensorManager
+import io.homeassistant.companion.android.views.ThemeLazyColumn
 import kotlinx.coroutines.runBlocking
 
 @SuppressLint("InlinedApi")
@@ -85,6 +86,11 @@ fun SensorUi(
                 overflow = TextOverflow.Ellipsis
             )
         },
+        secondaryLabel = {
+            sensor?.state?.let {
+                Text(if (sensor.unitOfMeasurement != null) "$it ${sensor.unitOfMeasurement}" else it)
+            }
+        },
         selectionControl = { ToggleSwitch(isChecked) },
         colors = getToggleButtonColors()
     )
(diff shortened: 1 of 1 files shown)


wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
```diff
@@ -95,10 +101,48 @@ fun SensorUi(
...
+                    sensor = Sensor(
+                        "battery_level",
+                        0,
+                        true,
+                        state = "80",
+                        unitOfMeasurement = "%"
+                    ),
+                    manager = batterySensorManager,
+                    basicSensor = runBlocking {
+                        batterySensorManager.getAvailableSensors(context).first()
```
jpelgrom, line 116: For the previews, maybe always get the sensor by id instead of relying on `first()`/`last()`? Otherwise it might break if the sensor order ever changes. You could also move the `batterySensorManager.getAvailableSensors(context)` call out of the inidividual items so it can be reused.

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
```diff
@@ -85,6 +86,11 @@ fun SensorUi(
                 overflow = TextOverflow.Ellipsis
             )
         },
+        secondaryLabel = {
+            sensor?.state?.let {
```
jpelgrom, line 90: To better match the phone sensors list
 - this should only be added if the sensor is enabled
 - this should skip the unit of measurement if the state is a string

(just [copy this](https://github.com/home-assistant/android/blob/6118933e4ef2eb8a20bc42f21f1dc25dd1371536/app/src/main/java/io/homeassistant/companion/android/settings/sensor/views/SensorListView.kt#L99-L111) except the enabled/disabled string which would be a bit much on a small screen)

dshokouhi: thanks added :)

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
dshokouhi: ok done :)

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
jpelgrom: You inverted the check which does not work - if the state is a number it will show the unit of measurement, which is exactly what should be avoided.

wear/src/main/java/io/homeassistant/companion/android/home/views/SensorUi.kt
jpelgrom: Now `batterySensors.first { ... }` is not using suspending functions so `runBlocking` can be removed ;)

jpelgrom (approved):
Looks good while testing, nice to make the menu more complete!

JBassett (approved):

Issue #3969, opened 2023-10-23
4 WiFi sensors broken on Android 11 (OnePlus 8T, OxygenOS) since 2023.7.8


kulmegil:
**Home Assistant Android app version(s):**
2023.10.4-full (11586) broken
2023.8.2-full (10992) broken
2023.7.8-full (10770) broken
2023.7.7-full (10650) working
2023.7.5-full (10506) working

**Android version(s):**
Android 11 (OxygenOS 11.0.13.13)

**Device model(s):**
OnePlus 8T

**Home Assistant version:**
2023.10.5

**Last working Home Assistant release (if known):**
Companion App 2023.7.7-full (10650)

**Description of problem, include YAML if issue is related to notifications:**
Following sensors shows incorrect values:
* WiFi IP address
* WiFi Frequency
* WiFi Link Speed
* WiFi Signal Strenght

**Companion App Logs:**

```
--------- beginning of main
10-23 15:16:04.236 23473 23473 E mpanion.androi: Unknown bits set in runtime_flags: 0x800000
10-23 15:16:04.346 23473 23473 W pool-9-thread-1: type=1400 audit(0.0:1865): avc: denied { read } for name="version" dev="proc" ino=4026532253 scontext=u:r:untrusted_app:s0:c90,c258,c512,c768 tcontext=u:object_r:proc_version:s0 tclass=file permissive=0 app=io.homeassistant.companion.android
10-23 15:16:04.242 23473 23473 E mpanion.androi: Not starting debugger since process cannot load the jdwp agent.
10-23 15:16:04.279 23473 23473 I Perf    : Connecting to perf service.
10-23 15:16:04.281 23473 23473 D NetworkSecurityConfig: Using Network Security Config from resource network_security_config debugBuild: false
10-23 15:16:04.282 23473 23473 D NetworkSecurityConfig: Using Network Security Config from resource network_security_config debugBuild: false
10-23 15:16:04.290 23473 23473 I FirebaseApp: Device unlocked: initializing all Firebase APIs for app [DEFAULT]
10-23 15:16:04.296 23473 23473 I FirebaseInitProvider: FirebaseApp initialization successful
10-23 15:16:04.299 23473 29907 E sqlite3_android: ONEPLUS_NAME_PARTS_MATCH SQLITE_OK
10-23 15:16:04.301 23473 23473 D WM-WrkMgrInitializer: Initializing WorkManager with default configuration.
[...]

PR #3970, opened 2023-10-23
Use deprecated wifi method in case WifiInfo is null

dshokouhi:
## Summary

Fixes: #3969 by defaulting to the old deprecated method in case `WifiInfo` returns as null. The 4 sensors mentioned use the new method but Wifi SSID and BSSID use the old method which in this case still works.


dshokouhi:
> Maybe add a short comment that this is a workaround for device-specific issues, as connectionInfo from the WifiManager can still be null?

added in latest commit.

Also have received confirmation from a user that the fix is working: https://github.com/home-assistant/android/issues/3969#issuecomment-1776034630

jpelgrom (approved):

JBassett (approved):

PR #3977, opened 2023-10-27
Always send sensor registration device class + entity category to prevent errors due to old registration

### LINKED ISSUE #3202: is using native unit of measurement 'None' which is not a valid unit for the device class ('battery')
### The problem

After upgrading to Home Assistant core 2023.1 I am seeing new WARNING logs like:

2023-01-05 11:27:47.301 WARNING (MainThread) [homeassistant.components.sensor] Entity sensor.firetablet_battery_state (<class 'homeassistant.components.mobile_app.sensor.MobileAppSensor'>) is using native unit of measurement 'None' which is not a valid unit for the device class ('battery') it is using; Please update your configuration if your entity is manually configured, otherwise create a bug report at https://github.com/home-assistant/core/issues?q=is%3Aopen+is%3Aissue+label%3A%22integration%3A+mobile_app%22

This is an automatic integration and there are not manual settings I can see.

### What version of Home Assistant Core has the issue?

2023.1

### What was the last working version of Home Assistant Core?

 2022.12.8

### What type of installation are you running?

Home Assistant Core

### Integration causing the issue

mobile_app

### Link to integration documentation on our website

https://www.home-assistant.io/integrations/mobile_app/

### Diagnostics information

_No response_

### Example YAML snippet

_No response_

### Anything in the logs that might be useful for us?

```txt
2023-01-05 11:27:47.301 WARNING (MainThread) [homeassistant.components.sensor] Entity sensor.firetablet_battery_state (<class 'homeassistant.components.mobile_app.sensor.MobileAppSensor'>) is using native unit of measurement 'None' which is not a valid unit for the device class ('battery') it is using; Please update your configuration if your entity is manually configured, otherwise create a bug report at https://github.com/home-assistant/core/issues?q=is%3Aopen+is%3Aissue+label%3A%22integration%3A+mobile_app%22
```

### Additional information

_No response_

jpelgrom:
## Summary

Fixes #3202, fixes home-assistant/core#102853

Always send the values for `device_class` and `entity_category` when registering sensors to make sure they match the app value. Some sensors were changed in the past based on feedback, but users with existing registrations wouldn't have seen this updated in core as the app didn't send `null` values instead of sending that it should be `null`. To make this possible, registrations and updates need to be split as the app currently depends on using `null` as "don't send".

Example before/after payloads:
```
{"type":"register_sensor","data":{"unique_id":"battery_state","state":"discharging","type":"sensor","icon":"mdi:battery-minus","attributes":{},"name":"Battery state","entity_category":"diagnostic","disabled":false}}
{"type":"register_sensor","data":{"unique_id":"battery_state","state":"discharging","type":"sensor","icon":"mdi:battery-minus","attributes":{},"name":"Battery state","device_class":null,"entity_category":"diagnostic","disabled":false}}

{"type":"register_sensor","data":{"unique_id":"audio_sensor","state":"normal","type":"sensor","icon":"mdi:volume-high","attributes":{},"name":"Ringer mode","disabled":false}}
{"type":"register_sensor","data":{"unique_id":"audio_sensor","state":"normal","type":"sensor","icon":"mdi:volume-high","attributes":{},"name":"Ringer mode","device_class":null,"entity_category":null,"disabled":false}}
```

## Screenshots

n/a

n/a

## Any other notes

Requires home-assistant/core#86965, which was first included in 2023.2.0.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
@@ -701,30 +701,33 @@ class IntegrationRepositoryImpl @AssistedInject constructor(
         val canRegisterCategoryStateClass = server.version?.isAtLeast(2021, 11, 0) == true
         val canRegisterEntityDisabledState = server.version?.isAtLeast(2022, 6, 0) == true
         val canRegisterDeviceClassDistance = server.version?.isAtLeast(2022, 10, 0) == true
+        val canRegisterNullProperties = server.version?.isAtLeast(2023, 2, 0) == true
+
+        val registrationData = SensorRegistrationRequest(
+            sensorRegistration.uniqueId,
+            if (canRegisterEntityDisabledState && sensorRegistration.disabled) {
+                null
+            } else if (sensorRegistration.state is String) {
+                sensorRegistration.state.ifBlank { null }
+            } else {
+                sensorRegistration.state
+            },
+            sensorRegistration.type,
+            sensorRegistration.icon,
+            sensorRegistration.attributes,
+            sensorRegistration.name,
+            when (sensorRegistration.deviceClass) {
+                "distance" -> if (canRegisterDeviceClassDistance) sensorRegistration.deviceClass else null
+                else -> sensorRegistration.deviceClass
+            },
+            sensorRegistration.unitOfMeasurement,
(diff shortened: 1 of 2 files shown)


dshokouhi (approved):
Thanks for the fix 🙏 ! 

I think we may want to opt to cut a release alongside HA core 2023.11 otherwise we may get more noise on this issue. I think this was one of the original sensors enabled by default for all users before we broke up attributes so could have a fair amount of users impacted by it depending on when they first setup the app.

JBassett (approved):

PR #3980, opened 2023-10-29
Official Gradle Wrapper Validation Action

IsakTheHacker:
To protect against malicious gradle-wrapper.jar binaries, I have added an action that verifies its checksum.

See: https://github.com/gradle/wrapper-validation-action


jpelgrom:
Thanks for the suggestion!

To mitigate the risk described, I think the correct approach would be [adding](https://github.com/gradle/wrapper-validation-action#add-to-an-existing-workflow) it to the existing `pr.yml` and `onPush.yml` workflows as [a job that is required to complete before](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#defining-prerequisite-jobs) the other jobs (which use Gradle) run. You currently add a new workflow, which would not prevent execution of the jar file as the other workflow also/still runs.

IsakTheHacker:
That is a great idea, didn't think about that. I have implemented your requested changes in `pr.yml` and `onPush.yml` respectively

jpelgrom (approved):
Sorry for the delay - looks good to me now

JBassett (approved):

PR #4012, opened 2023-11-13
Added a sensor to detect if the hotspot is enabled.

RubenNL:
## Summary

This adds a new binary_sensor displaying if the WiFi hotspot is enabled.

Documentation: home-assistant/companion.home-assistant#1005

## Any other notes

This feature had to use reflection, which usually isn't recommended. But this is the only method to check if a hotspot is enabled.

The sensor will only display if the private function exists, to prevent issues in the future.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
@@ -203,8 +203,8 @@ class NetworkSensorManager : SensorManager {
 
     @SuppressLint("PrivateApi")
     private fun hasHotspot(context: Context): Boolean {
-        //Watch doesn't have hotspot.
-        if(context.packageManager.hasSystemFeature(PackageManager.FEATURE_WATCH)) {
+        // Watch doesn't have hotspot.
+        if (context.packageManager.hasSystemFeature(PackageManager.FEATURE_WATCH)) {
             return false
         }
         val wifiManager: WifiManager = context.applicationContext.getSystemService()!!
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
@@ -3,6 +3,7 @@ package io.homeassistant.companion.android.common.sensors
 import android.Manifest
 import android.annotation.SuppressLint
 import android.content.Context
+import android.content.pm.PackageManager
 import android.net.ConnectivityManager
 import android.net.NetworkCapabilities
 import android.net.wifi.WifiInfo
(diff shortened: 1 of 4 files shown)


dshokouhi:
I don't think hotspot functionality works on Wear OS devices right? We may need to hide this sensor from them if thats the case.

common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
```diff
@@ -184,6 +200,35 @@ class NetworkSensorManager : SensorManager {
...
+        val wifiManager: WifiManager = context.applicationContext.getSystemService()!!
+        return try {
+            wifiManager.javaClass.getDeclaredMethod("isWifiApEnabled")
+            true
+        } catch (e: NoSuchMethodException) {
+            false
+        }
+    }
+    private fun updateHotspotEnabledSensor(context: Context) {
+        if (!isEnabled(context, hotspotState) || !hasWifi(context) || !hasHotspot(context)) {
```
dshokouhi, line 214: This should be an `and` condition as all 3 need to be met in order to be used

RubenNL: How should this be? I think this is the correct method, because if any of those are `false`, it should return.

dshokouhi: Oh sorry you are correct. In fact we dont need to add `hasWifi` and `hasHotspot` methods in this check as its already done up above in `getAvailableSensors`

RubenNL:
> I don't think hotspot functionality works on Wear OS devices right? We may need to hide this sensor from them if thats the case.

After a lot of work to get watchOS in an emulator running, I can confirm this. Fixed by checking if we aren't running a watch.

common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
```diff
@@ -168,6 +184,7 @@ class NetworkSensorManager : SensorManager {
     override fun requestSensorUpdate(
         context: Context
     ) {
+        updateHotspotEnabledSensor(context)
```
dshokouhi, line 187: None of the checks done for this sensor is required for the location permission, you can skip that for this sensor in the `requiredPermissions()` method

dshokouhi (changes requested):
Just a couple of changes to help the sensor update faster and also to remove the unnecessary permission check for this sensor

RubenNL:
I have also updated the icon to [`mdi:access-point`](https://icon-sets.iconify.design/mdi/access-point/). Sadly MDI doesn't have a "hotspot" icon like Android has.

dshokouhi (approved):
Looks good to me, tested the sensor and works as expected!

JBassett (approved):

teek541:
Is there a command/switch to turn the hotspot on/off?

PR #4027, opened 2023-11-23
Set the default BLE transmitter minor value to 40004

chatziko:
## Summary
Currently beacons sent by the BLE transmitter are ignored by the ibeacon integration, since they have an empty/default name. A workaround proposed [here](https://github.com/home-assistant/core/pull/104419) uses a whitelisted minor value (`40004`) to force processing a beacon with no name.

This tiny PR sets the default minor value to `40004`. Most likely the user doesn't care about it, so it will transparently solve the issue for them.

Documentation: home-assistant/companion.home-assistant#1007


dshokouhi (approved):
Looks good to me, will make sure to call this out in the change log once merged so existing users can change the minor value. Thanks for the PR!

JBassett (approved):

chatziko:
The initial solution of a hard-coded allowlisted minor was replaced by a manual UUID allow list ([new PR](https://github.com/home-assistant/core/pull/104790)). So this PR should be probably reverted.

PR #4035, opened 2023-11-28
mobile_app_notification_cleared for wearos

JosephAbbey:
## Summary

Moved the `NotificationDeleteRecieve` to common and added it to the WearOS app.

Documentation: home-assistant/companion.home-assistant#1009

## Any other notes

I have not been able to run the WearOS app at all on the simulator (even without any of my changes) so I haven't been able to test it but it compiles.


dshokouhi:
I think for this PR we also need to make a manifest change to add the receiver to the`wear manifest like it is added for the main app manifest

what issue are you running into with the emulator?

JosephAbbey:
Ok, I have added the receiver to the manifest.

The entire simulator just would crash as soon as the app started loading, I wasn't too worried because I have only just installed the wearos virtual device and I hoped someone else could test this change.

JosephAbbey:
I can't get logs because the whole simulator crashes and the logcat logs disappear to quickly. I haven't setup the notification service yet but the app won't even start anyway.

wear/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -133,4 +138,25 @@ class MessagingManager @Inject constructor(
             }
         }
     }
+
+    private fun handleDeleteIntent(
```
dshokouhi, line 142: We should move this and the same method from phone to `common` as well as the function is nearly identical here.

dshokouhi (changes requested):
Looks good to me and tested it on my watch locally. Still needs docs update before we can merge. Please submit a PR to add a mention in the Wear OS docs.

We need to update this page: https://companion.home-assistant.io/docs/wear-os/#notifications

dshokouhi (approved):

JBassett (approved):

PR #4073, opened 2023-12-26
Add option to make tap on entity state complication toggle the corresponding device.

clju:
## Summary

Currently, tapping a complication only refreshes the entity state. This PR adds a "toggle entity when tapped" option to the complication configuration page. When this option is enabled, tapping on the complication triggers a call to the #onPressed method of the corresponding device entity, which in turns executes a RPC to the corresponding service. This lets users quickly toggle home assistant devices with a single click on their watch face.

## Screenshots

screen

Documentation: home-assistant/companion.home-assistant#1015

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
@@ -97,7 +97,7 @@ import io.homeassistant.companion.android.common.R as commonR
         Server::class,
         Setting::class
     ],
-    version = 45,
+    version = 46,
     autoMigrations = [
         AutoMigration(from = 24, to = 25),
         AutoMigration(from = 25, to = 26),
@@ -118,7 +118,8 @@ import io.homeassistant.companion.android.common.R as commonR
         AutoMigration(from = 41, to = 42),
         AutoMigration(from = 42, to = 43),
         AutoMigration(from = 43, to = 44),
-        AutoMigration(from = 44, to = 45)
+        AutoMigration(from = 44, to = 45),
+        AutoMigration(from = 45, to = 46)
     ]
 )
 @TypeConverters(
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -795,6 +795,7 @@
     <string name="show_share_logs">Show and share logs</string>
     <string name="show_entity_title">Show entity name</string>
     <string name="show_unit_title">Show entity unit</string>
+    <string name="forward_taps_title">Toggle entity when tapped</string>
     <string name="sign_in_on_phone">Sign in on phone</string>
     <string name="slider_decreased">%1$s decreased</string>
     <string name="slider_increased">%1$s increased</string>
--- a/common/schemas/io.homeassistant.companion.android.database.AppDatabase/46.json
+++ b/common/schemas/io.homeassistant.companion.android.database.AppDatabase/46.json
@@ -1,0 +1,1121 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 46,
+    "identityHash": "6e80c503b23cc2d48e240d72a1d9e08d",
+    "entities": [
+      {
+        "tableName": "sensor_attributes",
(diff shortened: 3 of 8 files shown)


common/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
```diff
@@ -193,7 +194,8 @@ abstract class AppDatabase : RoomDatabase() {
                     MIGRATION_21_22,
                     MIGRATION_22_23,
                     MIGRATION_23_24,
-                    Migration40to41(context.assets)
+                    Migration40to41(context.assets),
+                    MIGRATION_45_46
```
jpelgrom, line 198: You only need the `AutoMigration` and version increase in the line at the top. This line and your changes below are a manual migration, please remove them.

common/src/main/res/values/strings.xml
```diff
@@ -795,6 +795,7 @@
     <string name="show_share_logs">Show and share logs</string>
     <string name="show_entity_title">Show entity name</string>
     <string name="show_unit_title">Show entity unit</string>
+    <string name="forward_taps_title">Forward taps to entity</string>
```
jpelgrom, line 798: Reading this, I was very confused about what the option will actually do. While it may be because I'm not a native English speaker, that will mean some other people probably also won't understand it.

Why not mimic what the entity state widget on the phone does (which is much more obvious in my opinion) and make it tap action: toggle or tap action: refresh?

clju: Hello,

Thanks for the feedback! I did think quite a bit about the wording and was left unsatisfied too :D

1) the word "toggle" seemed a bit too restrictive to me: indeed, depending on the entity domain, it can execute the following services:
- toggle
- lock  / unlock
- arm / disarm alarm
- press (for a `button` / `input_button`)
- turn on (for a `scene`)

Even if we take "toggle" as a loose term used to convey flipping between two states, it does not include the last two cases where the action is unary.

That being said, maybe given the precedent set by the widget (that I was unaware of), and the fact that it might be easier for users to understand (even though technically less accurate), we should go for it?

2) it might seem to the user that they are choosing between two exclusive options: refresh OR toggle. In actuality, both would refresh the complication - with one additionally triggering the entity prior to the refresh.

3) material3 / compose for wear do not appear to have a way to create drop-down dialogs easily.

Proposal: how about keeping this toggle, and changing the text to "Toggle entity" to make it more user-friendly and consistent with the widget?

common/src/main/java/io/homeassistant/companion/android/database/AppDatabase.kt
clju: Good catch! Done.

common/src/main/res/values/strings.xml
jpelgrom: Naming things is often harder than building them :D

> the word "toggle" seemed a bit too restrictive to me

While I can understand your points, your two "That being said" points feel more important to me than being technically correct.

 - consistency with the widget/other parts of the app for what is the same feature, instead of giving different names for the same feature
 - when you don't know Home Assistant inside out, it is easier to understand what will happen when you "toggle a scene/toggle a button" vs "forward tap to a scene/forward tap a button", even if it is technically not toggling

> it might seem to the user that they are choosing between two exclusive options: refresh OR toggle

I'd still expect feedback when I tap something which in addition to doing the thing, would be refreshing to show the updated state, so it's implied?

> keeping this toggle, and changing the text to "Toggle entity when tapped"

The issue with keeping this as a toggle is that it is unclear what happens when you uncheck it - there is no clear opposite. The native pattern for dropdowns on a watch would be 2 chips/buttons that are radio buttons.

clju: There is a clear opposite to "toggle entity when tapped": "don't toggle entity when tapped" :) IMO that's quite intuitive given the usual semantics of toggles.

I defer to your opinion in this matter - but my 2 cents is that "toggle" is not ideal. On top of the cases I listed above, consider the case of a sensor entity - how are users expected to understand "toggle" in this context? It sounds like it would enabling / disabling the sensor, doesn't it? In any case, this is outside the scope of this PR since it would require refactoring the other uses of the "toggle" word in similar context (eg. the widget). Just wanted to give my opinion :)

Will work on the radio button later. Thanks for the discussion!

jpelgrom:
Going to mark this as draft while waiting for radio buttons (either the 'official' one or manually mimicking the behavior). Assuming there are no big issues you can update the dependency once it reaches alpha02/03 - we've done so before to get access to components that were needed on Wear.

frenck:

../Frenck

PR #4117, opened 2024-01-05
Update change log week of 1/5

dshokouhi:


JBassett:
@jpelgrom do we want to mention the Matter fix and plan to release it next week?

jpelgrom:
If we're going to do a small release there's another Thread change I would like to get in, but that code is not yet ready for a PR 😅

Yes something generic like "More reliable Matter device commissioning when using Home Assistant 2024.1" is probably good to add

JBassett (approved):

PR #4119, opened 2024-01-06
Add padding to refresh interval to avoid text getting cut off

dshokouhi:
## Summary

We got a review failure from Google with a screenshot this time!

image

Adding padding to this screen to prevent the text from getting cut off like that. I tried to move to `ThemeLazyColumn` however the rotary input does not play nice so the best option seemed to be adding some sane padding

## Screenshots

After fix using largest font size in Pixel Watch (one side effect of this change is that not many elements are visible in the picker however we do need to push things down a bit so this might be expected in larger screen sizes):

image

Still behaves fine in default text size:

image


jpelgrom:
You only needed horizontal padding here, now it's quite hard to see the text can scroll on the largest size...

PR #4122, opened 2024-01-06
Integrate car sensors with automotive, and added permissions

ivancea:
## Summary

Integrate car sensors with Android Automotive:
- Added required permissions to manifest
- Updated the car sensors manager logic with the automotive permissions, separated from the auto ones

To better separate Auto from Automotive, I decided to add the core differences (Sensor availability and permissions) to the sensors themselves. This way, we have all the sensor logic and requirements in the same location, which is interesting when we start adding more sensors.

For that I made a CarSensor wrapper class of BasicSensor (I considered inheritance, but data classes are final, and I'm not sure changing them would have been worth it). Everything else is quite the same, but using the sensor objects instead.

## Screenshots

Example entities on an Automotive emulator:
imagen

The permission request on enabling some sensors:
imagen

> Documentation: home-assistant/companion.home-assistant#

We could change the "Currently unavailabe for AAOS users" to "Currently unavailabe for AAOS users from Google Play" (Or similar). But I'd say it's not worth it yet, at least until the full flow is ready (?)

## Any other notes

Only available in Automotive Full flavor (Not in the Google Play version yet). Sensors must be manually enabled, which can't be done in the minimal flavor. Enabling the sensors from the server won't ask for permissions from what I tested, so it may only work for some.


app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -123,10 +166,23 @@ class CarSensorManager :
...
+
+    private fun allDisabled(): Boolean = sensorsList.none { isEnabled(context, it) }
+
+    private fun connected(): Boolean = HaCarAppService.carInfo != null
+
     override val name: Int
         get() = R.string.sensor_name_car
 
     override suspend fun getAvailableSensors(context: Context): List<SensorManager.BasicSensor> {
+        this.context = context.applicationContext
```
ivancea, line 184: Those lateinit context set are a bit sketchy. I had to set it from more methods, as it wasn't set in some calls. I added it to every method accepting a context, just in case it happens again in the future.

I'm considering 2 other options: Adding it on construction (Not sure if possible), or simply drilling it through method calls, so we're sure we actually have a context when calling something that needs it.

I'd like to see more opinions about this

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -135,33 +191,34 @@ class CarSensorManager :
...
 
     override fun hasSensor(context: Context): Boolean {
-        // TODO: show sensors for automotive (except odometer) once
-        //  we can ask for special automotive permissions in requiredPermissions
-        return Build.VERSION.SDK_INT >= Build.VERSION_CODES.O &&
-            !context.packageManager.hasSystemFeature(PackageManager.FEATURE_AUTOMOTIVE) &&
-            BuildConfig.FLAVOR == "full"
+        this.context = context.applicationContext
+
+        return Build.VERSION.SDK_INT >= Build.VERSION_CODES.O
```
jpelgrom, line 196: This will also make these sensors available on minimal builds when not running Automotive, which is undesirable as they will never get data. To summarise, they should be available if:

|     |`full` flavor|`minimal` flavor|
|-----|-----|-----|
|`hasSystemFeature(PackageManager.FEATURE_AUTOMOTIVE) == false` (phones, tablets, watches, ...)|yes (Android Auto)|no|
|`hasSystemFeature(PackageManager.FEATURE_AUTOMOTIVE) == true` (cars)|no (for now, lacking UI access)|yes (this PR)|

> Only available in Automotive Full flavor (Not in the Google Play version yet)

That's the wrong way around, full = Play Services

ivancea: Oh, I see. I thought the semantics were about the screens, not about play services. Fixed!

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
dshokouhi: So we do actually use a `latestContext` method of storing the latest context received in the app for some sensors, very similar to what you are doing here.

https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/sensors/LightSensorManager.kt#L57

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -135,33 +192,39 @@ class CarSensorManager :
...
     override fun hasSensor(context: Context): Boolean {
-        // TODO: show sensors for automotive (except odometer) once
-        //  we can ask for special automotive permissions in requiredPermissions
-        return Build.VERSION.SDK_INT >= Build.VERSION_CODES.O &&
-            !context.packageManager.hasSystemFeature(PackageManager.FEATURE_AUTOMOTIVE) &&
-            BuildConfig.FLAVOR == "full"
+        this.context = context.applicationContext
+
+        return if (isAutomotive) {
+            BuildConfig.FLAVOR == "minimal"
```
dshokouhi, line 198: I vote we dont hide this from the automotive play store build personally, its up to google to allow users to have access to settings at this point. The code should work as is once settings access is granted.

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -135,33 +192,39 @@ class CarSensorManager :
...
-        }
+        } ?: emptyArray()
     }
 
-    private lateinit var context: Context
+    fun isEnabled(context: Context, carSensor: CarSensor): Boolean {
+        this.context = context.applicationContext
 
-    private fun allDisabled(): Boolean = sensorsList.none { isEnabled(context, it) }
+        if (isAutomotive && !carSensor.automotiveEnabled || !isAutomotive && !carSensor.autoEnabled) {
```
dshokouhi, line 218: Do we need to add some parenthesis here so the conditions are properly evaluated?

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
jpelgrom: If we do that, it should be checked that the sync triggers the notification missing permission correctly (it may need a vehicle extension?)

dshokouhi: ok in that case it may be out of scope for this PR then :)

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
ivancea: Adding them to improve readability

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
ivancea: I see. Renaming to latestContext, to keep the same naming as the other sensors

chrismelba:
Looks really good, will be following with interest

dshokouhi (approved):
Ran a local test with the changes and don't see any issues. Thank you for taking this on!

JBassett (approved):

domeng83:
Sorry if this is a poor phrased question but how does one access "Automotive Full flavor " version?  I have a vehicle Google Automotive Services(GAS) w/AAOS and have installed Home Assistant from Play Store.  Is there an APK I can download and sideload on this vehicle IVI of my own vehicle to get this functionality?

dshokouhi:
> Sorry if this is a poor phrased question but how does one access "Automotive Full flavor " version? I have a vehicle Google Automotive Services(GAS) w/AAOS and have installed Home Assistant from Play Store. Is there an APK I can download and sideload on this vehicle IVI of my own vehicle to get this functionality?

currently you cannot access them if you have installed the app via the play store, but you can follow this PR which seeks to enable them https://github.com/home-assistant/android/pull/4652 you just need to have some patience and wait.

PR #4142, opened 2024-01-17
Add ability for ClimateControl Card to cycle through entity states

StopMotionCuber:
## Summary

The ClimateControl right now doesn't cycle through modes (toggle in case it's a normal heating) when clicking on it within "Device controls" on Android which means more hassle when trying to change modes, as I would have to go into the app and in general it requires more clicks.

## Screenshots/Screencast

image

[Screencast from 2024-01-17 20-23-47.webm](https://github.com/home-assistant/android/assets/16007727/bc6be879-4084-4d85-8475-78842b3c55b5)

None so far. If that is needed, please come back to me.

## Any other notes
Didn't know whether I should create an issue for that just to have the implementation directly around and decided against that. In case I missed some workflow feel free to reach out to me. Would be happy to see this change merged

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
@@ -76,7 +76,8 @@ object ClimateControl : HaControl {
             val toggleRangeTemplate = ToggleRangeTemplate(
                 entity.entityId + "_range",
                 entity.state != "off",
-                "Toggle climate", rangeTemplate
+                "Toggle climate",
+                rangeTemplate
             )
             var modesFlag = 0
             this.currentMode = entity.state
--- a/app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
@@ -75,7 +75,7 @@ object ClimateControl : HaControl {
             supportedModes.clear()
             val toggleRangeTemplate = ToggleRangeTemplate(
                 entity.entityId + "_range",
-                entity.state != "off",
+                true,
                 "Toggle climate",
                 rangeTemplate
             )
--- a/app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
@@ -4,11 +4,13 @@ import android.content.Context
 import android.os.Build
 import android.service.controls.Control
 import android.service.controls.DeviceTypes
+import android.service.controls.actions.BooleanAction
 import android.service.controls.actions.ControlAction
 import android.service.controls.actions.FloatAction
 import android.service.controls.actions.ModeAction
 import android.service.controls.templates.RangeTemplate
 import android.service.controls.templates.TemperatureControlTemplate
+import android.service.controls.templates.ToggleRangeTemplate
 import androidx.annotation.RequiresApi
 import io.homeassistant.companion.android.common.R as commonR
 import io.homeassistant.companion.android.common.data.integration.Entity
@@ -31,6 +33,8 @@ object ClimateControl : HaControl {
         "heat_cool" to TemperatureControlTemplate.FLAG_MODE_HEAT_COOL,
         "off" to TemperatureControlTemplate.FLAG_MODE_OFF
     )
+    private var currentMode = "off"
+    private val supportedModes = ArrayList<String>()
 
     override fun provideControlFeatures(
         context: Context,
@@ -68,14 +72,22 @@ object ClimateControl : HaControl {
             "%.${temperatureFormatSize}f $temperatureUnit"
         )
         if (entityShouldBePresentedAsThermostat(entity)) {
+            supportedModes.clear()
+            val toggleRangeTemplate = ToggleRangeTemplate(
+                entity.entityId + "_range",
+                entity.state != "off",
+                "Toggle climate", rangeTemplate
+            )
             var modesFlag = 0
+            this.currentMode = entity.state
             (entity.attributes["hvac_modes"] as? List<String>)?.forEach {
                 modesFlag = modesFlag or temperatureControlModeFlags[it]!!
+                this.supportedModes.add(it)
             }
             control.setControlTemplate(
                 TemperatureControlTemplate(
                     entity.entityId,
-                    rangeTemplate,
+                    toggleRangeTemplate,
                     temperatureControlModes[entity.state]!!,
                     temperatureControlModes[entity.state]!!,
                     modesFlag
@@ -125,6 +137,18 @@ object ClimateControl : HaControl {
                                 it.value == ((action as? ModeAction)?.newMode ?: -1)
                             }?.key ?: ""
                             )
+                    )
+                )
+                true
+            }
+            is BooleanAction -> {
+                val nextMode = (this.supportedModes.indexOf(currentMode) + 1) % this.supportedModes.count()
+                integrationRepository.callService(
+                    action.templateId.split(".")[0],
+                    "set_hvac_mode",
+                    hashMapOf(
+                        "entity_id" to action.templateId,
+                        "hvac_mode" to this.supportedModes[nextMode]
                     )
                 )
                 true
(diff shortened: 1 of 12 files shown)


jpelgrom:
Do you actually have an example of this working? I've seen this option in the documentation before but couldn't get it to cycle/do anything with multiple modes, and your video also doesn't show anything except off/heat. If anything, it shows it not working (0:07 tap and nothing happens, needs another tap at 0:10).

jpelgrom:
> pushed an additional statement that modifies the internal state directly before sending stuff to home assistant to allow for more consistent cycling

This does result in weird looking updates with 2 transitions where state and temperature update independently. Why do this manually when all other controls depend entirely on the state to change? This way the state will appear updated even if it failed.

The "Toggle climate" string in line 76 should also be translated, or use a more generic "Toggle" like already exists in the strings file.

StopMotionCuber:
> > pushed an additional statement that modifies the internal state directly before sending stuff to home assistant to allow for more consistent cycling
> 
> This does result in weird looking updates with 2 transitions where state and temperature update independently. Why do this manually when all other controls depend entirely on the state to change on the server? This way the state will appear updated even if it failed.

The intention was that a "double-press" would be more responsive on that, as the new state is already internal and another press would then go to the next state _before_ the update has propagated to the server. Not sure if it actually works like that.
Regarding the weird looking transition, I've just tested to remove that instruction and the transition still looks as weird.
When using BetterThermostat this behavior doesn't happen, as it doesn't change the target temperature based on the mode.

> The "Toggle climate" string in line 76 should also be translated, or use a more generic "Toggle" like already exists in the strings file.

Thanks for the hint, adapted that.
As far as I see this string is not used within the UI though (the `ToggleRangeTemplate` is embedded in the `TemperatureControlTemplate`, so all strings are set based on the latter), so the string value shouldn't matter.

StopMotionCuber:
`pr_build` is failing, but showing `java.lang.OutOfMemoryError`. I guess that's a GitHub Actions failure and unrelated to the PR itself. Builds fine on my workstation. Not sure how I can rerun the action though :/

app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
```diff
@@ -31,6 +33,8 @@ object ClimateControl : HaControl {
         "heat_cool" to TemperatureControlTemplate.FLAG_MODE_HEAT_COOL,
         "off" to TemperatureControlTemplate.FLAG_MODE_OFF
     )
+    private var currentMode = "off"
+    private val supportedModes = ArrayList<String>()
```
jpelgrom, line 37: This does not work as you expect if you have more than one control with different values. Consider the following example: https://pl.kotl.in/SM3ioCSag

Because an `object` is a singleton, it will keep the latest value you set. This means that, for example, if you have 2 climate entities which are in `off` and `cool` mode, `currentMode` will always be `cool`. When you do not interact with the latest added entity, this breaks your "finding the next mode" logic as it'll use the state of the second entity to update the first one.

jpelgrom:
> `pr_build` is failing, but showing `java.lang.OutOfMemoryError`. I guess that's a GitHub Actions failure and unrelated to the PR itself. Builds fine on my workstation. Not sure how I can rerun the action though :/

GitHub Actions can be flaky sometimes, we can/will restart it for you in that case :)

StopMotionCuber:
@jpelgrom In general I also wanted to ask whether this PR has a good chance of being merged, once stuff is implemented properly? I feel like it's kind of a hack using a Toggle for an action that's cycling through modes, but one that enhances usability with the way android is implementing their controls right now.
You haven't given a green light on the idea of this enhancement yet, but jumped right into a review.

app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
StopMotionCuber: Explored the options that I had here:

  1. Use HashMap to save values in `ClimateControl.kt` locally, based on their `entityId`, basically the same idea as before, just with a fix to the singleton behavior. This has the advantage that the internal state would be the same as the one presented on the screen, as when the visible state is updated the internal one is too.
  2. Change `performAction` to supply the `entity`, including it's state. This is in my opinion from an interface perspective the cleanest design. However, in the caller at `HaControlsProviderService::performControlAction()` this entity is not directly available. So either:
    1. This needs to be cached within the `HaControlsProviderService`
    2. This is fetched from the server right before sending the request
  Fetching for all entities is a really bad idea, as that would double the response time and adding an unnecessary request that would be thrown away most of the times anyways. Caching within `HaControlsProviderService` also seems like overkill to me.
  
For those reasons I went with the first option. Tested it with 2 devices and it worked as expected, making toggling independent of the state of the other devices.

One final note I have is that the logic would break with multi-server environments where device IDs are available in multiple servers

jpelgrom:
> @jpelgrom In general I also wanted to ask whether this PR has a good chance of being merged

Sorry for the confusion, that was implied by reviewing the code - as far as I'm concerned this would be accepted when properly implemented and I currently don't see anything that would block it. You've described a good use case, it doesn't conflict with existing functionality and while you may consider this a hack it also doesn't feel unexpected/Android isn't exposing the modes otherwise.

app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
jpelgrom: I agree that the first option seems the best solution here. However, now that you've mentioned it this can't be ignored as the app supports multiple servers and specifically mixing controls from different servers:

> the logic would break with multi-server environments where device IDs are available in multiple servers

Suggestion for a fix: the `area` and `baseUrl` arguments for the `provideControlFeatures` function come from a class that holds more information about the control, including the `systemId` which is unique. In the interface, change [these last two arguments](https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/controls/HaControl.kt#L87-L88) to a single `info` argument that passes the entire class?

StopMotionCuber: Ah shit, shouldn't have mentioned it :smile: 

Fix suggestion makes sense, had to only touch 12 source files to make that work :grimacing: 
However for most controls both arguments were not used and therefore only a signature change.

As I'm now also using the systemId as ID for the templates I had to adapt the call service part. My assumption was that `systemId` would be of the format `[SYSTEM.]AREA.ENTITYID` where at max 2 `.` are within the string and if there are 2 that means the `system` is present, otherwise there is only 1 `.` and the system is not present.

Therefore I'm getting the `AREA.ENTITYID` by splitting the first part off, if >= 2 `.` are there

app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
```diff
@@ -60,22 +63,31 @@ object ClimateControl : HaControl {
...
             maxValue,
             currentValue,
             temperatureStepSize,
             "%.${temperatureFormatSize}f $temperatureUnit"
         )
         if (entityShouldBePresentedAsThermostat(entity)) {
+            val state = ClimateState(entity.state, ArrayList())
+            val toggleRangeTemplate = ToggleRangeTemplate(
+                info.systemId + "_range",
+                true,
```
jpelgrom, line 77: Could you add a comment at the end of this line explaining this is `true` so that the temperature will always be shown?

jpelgrom:
Tested and left one minor comment, otherwise it's looking good!

Do you also want to submit a documentation PR to reflect that it is not just a temperature slider, but also cycles through modes when tapped now?

app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt
StopMotionCuber: Have added that. I've put it in a separate line as `ktlint` doesn't like comments in parameter lists at the end of the line

StopMotionCuber:
> Tested and left one minor comment, otherwise it's looking good!
> 
> Do you also want to submit a documentation PR to reflect that it is not just a temperature slider, but also cycles through modes when tapped now?

Surely can do so within the next days. I assume you mean here: https://github.com/home-assistant/companion.home-assistant/blob/master/docs/integrations/android-device-controls.md

jpelgrom (approved):
Good idea, thanks for implementing it!

jpelgrom:
> Surely can do so within the next days. I assume you mean here: https://github.com/home-assistant/companion.home-assistant/blob/master/docs/integrations/android-device-controls.md

Yes, that'd be the place to add it. Don't forget to add a beta label (see any recent documentation PR as an example).

JBassett (approved):

PR #4144, opened 2024-01-18
Replace 'Refresh interval' title with icon

jpelgrom:
## Summary

To (hopefully) finally address review failures about text being cut off on round screens replace it with an icon. Same icon as used in the previous screen so should be pretty clear.

## Screenshots

Picker with a clock with cog icon above it

As a reminder, the preference chip:
Refresh interval: Manual, with the same icon

n/a


dshokouhi (approved):
Much better fix to use a icon here!

JBassett (approved):

Issue #4146, opened 2024-01-20
Update home-assistant network Thread credentials not possible


agners:
**Home Assistant Android app version(s):**
2024.1.2-full

**Android version(s):**
14

**Device model(s):**
Pixel Pro 8

**Home Assistant version:**
2024.1.3

**Last working Home Assistant release (if known):**

**Description of problem, include YAML if issue is related to notifications:**
When creating a new `home-assistant` Thread network in Home Assistant (e.g. via "Reset border router" feature in the Home Assistant Thread integration configuration page) updating the Thread credentials seems not to work. 

To reproduce:

1. "Clear all data" of "Google Play services" method to make sure to start from a clear slate
2. Have a `home-assistant` network as preferred set in Core (say `home-assistant` network 1)
3. Select "Sync Thread credentials" -> "✅ Added network from Home Assistant to this device"
4. Select "Sync Thread credentials" -> "✅ Home Assistant and this device use the same network" :+1: 
5. Create a `home-assistant` Thread network e.g. by using "Reset border router"
6. Select "Make preferred network" to mark this new `home-assistant` network as preferred
7. Select "Sync Thread credentials" -> "✅ Updated network from Home Assistant to this device"
8. Select "Sync Thread credentials" -> "✅ Updated network from Home Assistant to this device"

Up to 7 everything is as expected. However, when trying to commission a device after step 7, it won't work because the credentials didn't really got updated. Step 8 also proves that. It seems the app is not able to update the dataset.

<details>

**Screenshot and log of step 3:**

![image](https://github.com/home-assistant/android/assets/34061/31e4fcad-c02f-4747-8206-dcd79be0a47d)

```
01-20 14:36:30.259  7110  7110 W WindowOnBackDispatcher: sendCancelIfRunning: isInProgress=falsecallback=com.google.android.material.motion.MaterialBackOrchestrator$Api34BackCallbackDelegate$1@288a0ce
01-20 14:36:30.268  7110  7110 W InputEventReceiver: Attempted to finish an input event but the input event receiver has already been disposed.
01-20 14:36:30.282  7110  7110 D ServerConnectionInfo: localUrl is: false, usesInternalSsid is: false, usesWifi is: true
01-20 14:36:30.282  7110  7110 D WebSocketRepository: Sending message 49: {type=get_config, id=49}
01-20 14:36:30.282  7110  7110 D WebSocketRepository: Message number 49 sent
[...]

PR #4150, opened 2024-01-22
Remove Thread credential sync from start of Matter commissioning flow

jpelgrom:
## Summary

After discussion with @agners, remove Thread credential syncing from the start of the Matter commissioning flow. There are 2 major reasons for this:

* it significantly slows down the flow while only being useful some of the time (initially, and on Thread network/BR changes)
* the results have limited usefulness because the API limits how much Home Assistant can influence the device data
 
The alternative for the user is to use the manual 'import credentials' button in the frontend (matching iOS), or access the sync option in the app settings > troubleshooting menu.

This should result in a more reliable user experience when using Matter and prevents unstable/unfinished Thread work on the core and app side 'polluting a real device'. Adding a comment in the code in case/when we revisit this in the future.

## Screenshots

n/a / the Thread permission dialog will no longer show up when doing Matter commissioning

n/a, this behavior wasn't documented, general Thread companion app import/export documentation to be done independently

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/matter/MatterCommissioningViewModel.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/matter/MatterCommissioningViewModel.kt
@@ -2,7 +2,6 @@ package io.homeassistant.companion.android.matter
 
 import android.app.Application
 import android.content.IntentSender
-import android.util.Log
 import androidx.activity.result.ActivityResult
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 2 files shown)


agners (approved):
As it is now, this only really would be helpful for the very first time, if the user has indeed a Home Assistant controlled OTBR, and if he was indeed triggering the flow from within the Home Assistant Android App and has no Google Thread border router yet. For all other use-cases this currently just adds lot of waiting time.

jpelgrom:
I'd argue that there is still some value if you do not have OTBR but do have Google border routers, as this would get their dataset into Home Assistant (where you can't really use it but maybe for futureproofing?).

agners:
> I'd argue that there is still some value if you do not have OTBR but do have Google border routers, as this would get their dataset into Home Assistant (where you can't really use it but maybe for futureproofing?).

I think both direction can be useful, it is just that very few use cases are actually benefit from it. Even when there is no Thread network on a phone, importing only seems to work once (at least in my test case, see #4146).

When a user has no TBR and uses a HA OTBR, then he *probably* wants to import, but even that is not a 100% given, especially since the limitations on Android side means he will be stuck with that setup (and we decided for him :sweat_smile: ).

I think in the current situation, and since the sync is rather slow, it is just better to only do this manually. We can revisit an automatic mechanism later.

dshokouhi (approved):

JBassett (approved):

Scutua (approved):

PR #4160, opened 2024-01-25
Support Cloudflare Access cookie expiration flow

mcuelenaere:
## Summary

If the user is hosting their HA instance behind Cloudflare Access, then at some point in the future their session will expire and they will need to reauthenticate with Cloudflare.

Currently, this results in a web browser being opened (because of how the URL intercepting is implemented). This results in the user authenticating in the browser and thus the browser containing these new cookies, not the HA companion app.

This PR changes this so that the Cloudflare authentication happens in the webview (and thus the cookies are accessible by the companion app and can be used at any further point in the flow).

## Any other notes

This should mostly resolve issues like https://github.com/home-assistant/android/issues/2650 and others. The solution is still not foolproof, as the browser is still being opened after the auth process, due to how the current `shouldLaunchWebBrowser` logic is implemented (eg app wanted to navigate to `https://yourhainstance.com`, got redirected to `https://yourhainstance.com/cdn-cgi/access/xxx` and after login it goes back to `https://yourhainstance.com`, which no longer matches `https://yourhainstance.com/cdn-cgi/access/xxx`).

A better solution would be to check if the domain of the to-be-navigated-to-page matches a domain of `internalUrl` or `externalUrl` of any of the registered servers (in `serverManager.defaultServers`). I did not implement this change as I wanted to make this PR as small as possible, to increase the chances of it getting accepted.


mathix420:
There is maybe an issue with this PR, the whole time between cookie expiration time and user refreshing it via the web ui none of the HA sensors will be updated, and the HA client could not receive any notifications.
I found [this solution (#3510)](https://github.com/home-assistant/android/pull/3510) more practical, but yours is still a good step toward allowing the use of cf tunnels on HA android!

mcuelenaere:
> There is maybe an issue with this PR, the whole time between cookie expiration time and user refreshing it via the web ui none of the HA sensors will be updated, and the HA client could not receive any notifications. I found [this solution (#3510)](https://github.com/home-assistant/android/pull/3510) more practical, but yours is still a good step toward allowing the use of cf tunnels on HA android!

That is indeed a limitation (which is fine for my particular usecase). This approach should be minimal and I believe upstream might be more likely to accept this than https://github.com/home-assistant/android/pull/3510, which has user-facing changes.

Meister1977:
As the owner of the PR https://github.com/home-assistant/android/pull/3510 , my solution can use service cookies, which is an accepted solution for services running without user interaction instead of login session cookies.
I don't really know why they do not accept my PR. But I gave it up, I made a fork, and if I have time, I rebase, and release.

Meister1977:
As the owner of the PR https://github.com/home-assistant/android/pull/3510 , my solution can use service cookies, which is an accepted solution for services running without user interaction instead of login session cookies.
I don't really know why they do not accept my PR. But I gave it up, I made a fork, and if I have time, I rebase, and release.

Caligo82:
Not trying to hate here, but it seems this "feature" is purposely reserved for the premium plan of homeassistant. I mean not exactly the same feature, but the logic of having an encrypted tunnel to a cloud shield, which serves as login platform for homeassistant for people using the app. It sounds granular, but it seems everyone and his mum is literally using this. I see where nabucasa is protecting its business model of sorts, but since the setup requires some technical expertise I would not assume it having a relevant impact there... who knows.

ericmedina024:
> Not trying to hate here, but it seems this "feature" is purposely reserved for the premium plan of homeassistant. I mean not exactly the same feature, but the logic of having an encrypted tunnel to a cloud shield, which serves as login platform for homeassistant for people using the app. It sounds granular, but it seems everyone and his mum is literally using this. I see where nabucasa is protecting its business model of sorts, but since the setup requires some technical expertise I would not assume it having a relevant impact there... who knows.

Balloob is right - I tried to implement a PR for this myself (#3593) and it just ended up not making sense for a variety of technical reasons, including some which were limitations imposed by the Android API. It's still possible to use zero trust with the app (see https://github.com/home-assistant/android/pull/3593#issuecomment-1597693594), and honestly the solution described there is even better than what my and this PR added.

PR #4162, opened 2024-01-26
Make volume sensors update instantly by subscribing to a broadcast

marazmarci:
## Summary
Audio volume sensors weren't updating instantly, so a new broadcast Intent subscription has been introduced in this PR to improve this.

How I tested these changes:
Enabled all audio-related sensors, and tried modifying the volume on the device several times, and the new values were reflected nearly instantly on the server. I did this testing for both the Android and Wear OS apps.

Documentation: home-assistant/companion.home-assistant#1025

## Any other notes
The broadcast is `android.media.VOLUME_CHANGED_ACTION`, and it is undocumented. No official Android documentation mentions it.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
@@ -115,6 +115,7 @@ class SensorReceiver : SensorReceiverBase() {
         AudioManager.ACTION_MICROPHONE_MUTE_CHANGED to AudioSensorManager.micMuted.id,
         AudioManager.ACTION_SPEAKERPHONE_STATE_CHANGED to AudioSensorManager.speakerphoneState.id,
         AudioManager.RINGER_MODE_CHANGED_ACTION to AudioSensorManager.audioSensor.id,
+        AudioSensorManager.VOLUME_CHANGED_ACTION to AudioSensorManager.volMusic.id,
         Intent.ACTION_MANAGED_PROFILE_UNAVAILABLE to DevicePolicyManager.isWorkProfile.id,
         Intent.ACTION_MANAGED_PROFILE_AVAILABLE to DevicePolicyManager.isWorkProfile.id,
         WifiManager.WIFI_STATE_CHANGED_ACTION to NetworkSensorManager.wifiState.id,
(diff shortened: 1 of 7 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/SensorReceiver.kt
```diff
@@ -115,6 +115,7 @@ class SensorReceiver : SensorReceiverBase() {
         AudioManager.ACTION_MICROPHONE_MUTE_CHANGED to AudioSensorManager.micMuted.id,
         AudioManager.ACTION_SPEAKERPHONE_STATE_CHANGED to AudioSensorManager.speakerphoneState.id,
         AudioManager.RINGER_MODE_CHANGED_ACTION to AudioSensorManager.audioSensor.id,
+        AudioSensorManager.VOLUME_CHANGED_ACTION to AudioSensorManager.volMusic.id,
```
dshokouhi, line 118: I am not sure if this is correct. According to the docs this intent gets fired for each and every volume stream and not just the media stream. The way this is now if the user does not have the volume music sensor but has any of the other volume sensors enabled then they will not update immediately.

https://android.googlesource.com/platform/frameworks/base/+/8bd69610aafc6995126965d1d23b771fe02a9084/media/java/android/media/AudioManager.java#102

marazmarci: I've just tested it and you're right, other volume sensors won't update immediately if the music volume sensor is disabled.

marazmarci: Maybe you have a solution in mind that you'd prefer?

dshokouhi: currently its a 1 to 1 mapping here, not all intents are added here like network changes which is more similar to the intent you have added as multiple sensors can be updated by it.

https://github.com/home-assistant/android/blob/f73a9dddec423a3ef66f02f70175168733c501d6/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt#L129

the best place to add something else unique would be in `SensorReceiverBase` where we do additional processing like the skippable actions.

marazmarci: I came up with a bit different solution. Please check it in (26d496aa802af3bc81619df6ec94aaf24716379a) and let me know if you like it or not. If not, I'm open to dropping this and implementing something based on your previous hints.

dshokouhi: great solution :)

dshokouhi:
we should do a docs change for this PR as well to mention the sensors will update instantly in the beta. Currently we mention the volume sensors follow the normal update schedule.

https://companion.home-assistant.io/docs/core/sensors#audio-sensors

other than that functionality looks good to me :)

after this PR we should probably look at the other intents that impact more than 1 sensor and add them to the list :)

dshokouhi (approved):
Looks good to me

JBassett (approved):

Issue #4198, opened 2024-02-10
Add IPv6 to WifiInfoSensor


h3mmy:
**Is your feature request related to a problem? Please describe.**

My core personal issue is that my phone regularly drops it's IPv6 address. This drives me nuts. I want to be able to detect when this happens and automatically cycle the wifi to regain an IPv6 assignment. I thought, well Home Assistant should have access to this data and is always running on my phone anyway. After checking the sensor for WifiInfo wifi_ip_address, I noted it was only the IPv4 address. This struck me as odd. I went to check the code in case there was a hidden option that I maybe missed.

This was not the case. It turns out that the companion app only retrieves the IPv4 address and additionally relies on a deprecated (as of version 31) way to get the address.

https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt#L345

**Describe the solution you'd like**

Update the NetworkSensorManager to use [LinkAddress](https://developer.android.com/reference/android/net/LinkAddress) to retrieve the inet addresses and display both types of addresses as part of the wifiInfo sensor. I am happy to open a PR for this change as well.

**Describe alternatives you've considered, if any**

Doing it manually instead of using home-assistant. I feel like the simple check and cycle is exactly the kind of automation that fits the goals of the home-assistant project.

**Additional context**

I am happy to make the code changes, which should be rather minimal. Just want a go ahead from someone before putting in the work.

dshokouhi:
The current sensor is designed to show the IPv4 address. If you would like to add one for IPv6 please make sure its a new sensor 😃 . The other sensor needs to continue to use the deprecated method to support older devices but a new sensor can be setup to show the IPv6 address as its state.

h3mmy:
That seems reasonable to me. It'll probably take me a little extra fiddling to get it right, but I can get working on that then. Thanks for the quick response!

h3mmy:
This might be a silly question, but is there somewhere specific I can add unit tests to? I'm not able to find the testcase files in the repository.

jpelgrom:
> This might be a silly question, but is there somewhere specific I can add unit tests to? I'm not able to find the testcase files in the repository. 

There are currently no unit tests in the project.

PR #4202, opened 2024-02-13
add new sensor for ip6 addresses

h3mmy:
## Summary

This PR is intended to address #4198: Adds a new network sensor to detect ipv6 addresses and list them as attributes

Documentation: home-assistant/companion.home-assistant#1031

## Any other notes

I'm still trying to figure out how best to test this. As such I'm leaving this as a draft for now

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
@@ -126,7 +126,7 @@ class NetworkSensorManager : SensorManager {
             commonR.string.basic_sensor_name_ip6_addresses,
             commonR.string.sensor_description_ip6_addresses,
             "mdi:ip",
-            unitOfMeasurement = "connection(s)",
+            unitOfMeasurement = "address(es)",
             stateClass = SensorManager.STATE_CLASS_MEASUREMENT,
             entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC,
             updateType = SensorManager.BasicSensor.UpdateType.INTENT
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
@@ -384,7 +384,7 @@ class NetworkSensorManager : SensorManager {
             if (!ipAddresses.isNullOrEmpty()) {
                 val ip6Addresses = ipAddresses.filter { linkAddress -> linkAddress.address is Inet6Address }
                 if (ip6Addresses.isNotEmpty()) {
-                    ipAddressList = ipAddressList.plus(elements = ipAddresses.map { linkAddress -> linkAddress.toString() })
+                    ipAddressList = ipAddressList.plus(elements = ip6Addresses.map { linkAddress -> linkAddress.toString() })
                     totalAddresses += ip6Addresses.size
                 }
             }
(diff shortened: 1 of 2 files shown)


common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
```diff
@@ -148,7 +161,7 @@ class NetworkSensorManager : SensorManager {
             wifiSignalStrength
         )
         val list = if (hasWifi(context)) {
-            val withPublicIp = wifiSensors.plus(publicIp)
+            val withPublicIp = wifiSensors.plus(publicIp).plus(ip6Addresses)
```
dshokouhi, line 164: this sensor requires the user to have Android 6 or higher so we need to make sure we do not make the sensor available to all user as the app supports Android 5+

common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
```diff
@@ -356,6 +369,40 @@ class NetworkSensorManager : SensorManager {
...
+        }
+        val ipAddressList: List<String> = ArrayList()
+        var totalAddresses = 0
+
+        if (checkPermission(context, ip6Addresses.id)) {
+            val connectivityManager = context.applicationContext.getSystemService<ConnectivityManager>()
+            val activeNetwork = connectivityManager?.activeNetwork
+            val ipAddresses = connectivityManager?.getLinkProperties(activeNetwork)?.linkAddresses
+            if (ipAddresses.isNullOrEmpty()) {
+                return
```
dshokouhi, line 385: `return` does not seem correct here as we need to make sure to actually update the sensor to send an appropriate state if the sensor is enabled with proper permissions

common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
```diff
@@ -119,6 +120,18 @@ class NetworkSensorManager : SensorManager {
...
         )
+        val ip6Addresses = SensorManager.BasicSensor(
+            "ip6_addresses",
+            "sensor",
+            commonR.string.basic_sensor_name_ip6_addresses,
+            commonR.string.sensor_description_ip6_addresses,
+            "mdi:ip",
+            unitOfMeasurement = "connection(s)",
+            stateClass = SensorManager.STATE_CLASS_MEASUREMENT,
+            docsLink = "https://companion.home-assistant.io/docs/core/sensors#connection-type-sensor",
```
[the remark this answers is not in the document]
h3mmy: Thanks for pointing that out. Fixed

common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
h3mmy: Yes, I fixed that in a later commit. The change should be in here now

common/src/main/java/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
h3mmy: copy-paste error. The only return now is if it is disabled

h3mmy:
I'm marking this as ready for review, but I'd like to do some more manual testing before it is considered merge-able.

h3mmy:
The PR is ready. I've verified that existing functionality is not impacted with some cursory testing. 

[x] Sensor does not update when disabled
[x] Sensor displays `0 address(es)` when no ipv6
[x] Sensor displays correct count of ipv6 addresses bound to active interface
[x] Sensor contains list of addresses in attribute
[x] IPv4 addresses are excluded from the list and also the count

Screenshot_20240213-200826
Screenshot_20240213-200819

dshokouhi (approved):
Looks good to me, tested the new sensor and its working as expected

JBassett (approved):

Issue #4221, opened 2024-02-23
command_screen_off_timeout not working and not prompting permission


ministryofsillywalks:
**Home Assistant Android app version(s):**
2024.2.3-full
**Android version(s):**
11 and 14
**Device model(s):**
Lenovo M10 Tab Gen 3 and Pixel 7a
**Home Assistant version:**
2024.2.2
**Last working Home Assistant release (if known):**
-
**Description of problem, include YAML if issue is related to notifications:**
Sending command_screen_on to the tablet works great. Sending command_screen_off_timeout with any value doesn't change the screen off timeout on the device.
The app gives me a notification:
Please open the Home Assistant app and send the command again in order to grant the proper permissions. You will be taken to a page to grant the Home Assistant app the permission....

Doing that does nothing. No page to grant the permissions. The app already has all permissions granted that I can grant it, but still no changing the screen off timeout.

Please let me know if I can provide any logs or further information

**Companion App Logs:**

```

```

**Screenshot or video of problem:**

**Additional information:**

ministryofsillywalks:
Here is the command

```
service: notify.mobile_app_m10
data:
  message: "command_screen_off_timeout"
  data:
    command: 10000
```

Here is the log

```
--------- beginning of main
02-23 16:30:36.707 18943 27898 W FirebaseMessaging: Unable to log event: analytics library is missing
02-23 16:30:36.709 18943 27898 D FCMService: From: 331041709873
02-23 16:30:36.746 18943 18943 D MessagingService: Processing device command
02-23 16:30:36.804 18943 18943 D MessagingService: Show notification with tag "null" and id "-694747049"
02-23 16:32:21.947 18943 18943 D WebviewActivity: Matter/Thread step changed to NOT_STARTED
02-23 16:32:21.958 18943 18943 D ServerConnectionInfo: localUrl is: false, usesInternalSsid is: false, usesWifi is: true
02-23 16:32:21.960 18943 18943 D ServerConnectionInfo: localUrl is: false, usesInternalSsid is: false, usesWifi is: true
02-23 16:32:21.961 18943 18943 D IntegrationRepository: isAppLocked(): false. (LockEnabled: false, appActive: false, expireMillis: 1708691290488, currentMillis: 1708702341961)
02-23 16:32:22.144 18943 21195 I gralloc4: register: id=28000003a71, handle:0xb4000076b579c7f0, vaddr=0x75f078a000
[...]

PR #4223, opened 2024-02-23
Add command screen off to write permissions method

dshokouhi:
## Summary

Fixes: #4221 as the command needs write system settings permission but was never actually directing the user to the correct location for this specific command


jpelgrom (approved):
Tested, and in a quick check I didn't see any other missing commands

dshokouhi:
> Tested, and in a quick check I didn't see any other missing commands

yea I think this was easy to miss because the other screen commands require the same permission and most likely it was already granted during testing

JBassett (approved):

PR #4243, opened 2024-03-03
Add option to set different status for Tile

BlastyCZ:
## Summary

I'm adding a new functionality (modification of the Android Tile) to the Home Assistant Android application. I want to control the garage door as quickly as possible. However, the switch that controls it does not correspond to the state but only opens or closes the garage door (it turns on for 1 second and then turns off). I have another entity for the state, but its state cannot be displayed. So, I added an option that allows the user to display (overwrite) the state, and I have only allowed binary_sensor.

## Screenshots

ha-dark
ha-light
tile

Documentation: home-assistant/companion.home-assistant#1039

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/qs/TileExtensions.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/qs/TileExtensions.kt
@@ -106,13 +106,13 @@ abstract class TileExtensions : TileService() {
             stateUpdateJob = mainScope.launch {
                 val tileData = tileDao.get(getTileId())
                 if (tileData != null &&
-                    tileData.isSetup &&
-                    tileData.actionEntityId.split('.')[0] in toggleDomainsWithLock &&
-                    serverManager.getServer(tileData.serverId) != null
+                    tileData.isSetup && (
+                        tileData.actionEntityId != "" || tileData.statusEntityId.split('.')[0] in domainWithIconChange
+                        ) && serverManager.getServer(tileData.serverId) != null
                 ) {
-                    val entity = if (tileData.statusEntityId != ""){
+                    val entity = if (tileData.statusEntityId != "") {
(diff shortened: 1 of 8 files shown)


jpelgrom:
This feels needlessly complex to me, even confusing if you don't want to override things (why is it only showing binary sensors? what do I enter if I don't want to override?), and now I'm afraid people will see this as a 'solution' everywhere - eg widgets, device controls, in the car interface, ...

For your use case, why not use a [template cover](https://www.home-assistant.io/integrations/cover.template/) to decouple the state and actions?

BlastyCZ:
To be honest, it doesn't seem like an overly complex solution, more like a simple way to change it in the UI. Maybe I'm just describing it wrong, I haven't needed a template cover yet, because I use https://www.home-assistant.io/dashboards/tile, which also uses exactly this solution - with the difference that you can change the action here.

Why is it only showing binary sensors?
Glad you asked, to be honest I wasn't sure what to choose, I only know HA as a user. I haven't contributed anything and I developed android apps many years back. I thought all entities with on/off state were included.

What do I enter if I don't want to override?
Simply nothing. Mandatory items are marked, this one is not. You think it's not clear at a glance?

For your use case, why not use a template cover to decouple the state and actions?
The answer is simple - I wasn't familiar with it, I try to click most things, however, I was familiar with Tile cards, which is clickable, so I thought it would be good to do the same in the app.

dshokouhi:
Personally speaking I don't think this is logic we should keep in the app but logic that should be handled by HA core. Template Covers are the intended use case for when you want a custom state and custom actions. A tile card is just a UI element and not quite the same thing here even though they are both called tiles and are clickable.

jpelgrom:
Going to close this as template cover - or any of the other template entities/helpers - already cover the use case.

The app shouldn't recreate functionality that already exists without workarounds. Trying to replicate all dashboard functionality in QS/controls/widgets is not a goal and won't be fully possible.

PR #4273, opened 2024-03-15
Notification: "when_relative" for "chronometer"

Roffild:
"timeout" is relative, but "when" is absolute by default.

It is now possible to synchronize timers in seconds without time difference:
```yaml
timeout: 120
chronometer: "true"
when: 120
when_relative: "true"
```


dshokouhi:
This PR requires the new parameter to be added to the server side code as well.

https://github.com/home-assistant/mobile-apps-fcm-push

dshokouhi (changes requested):
Please attach a PR to the FCM repo adding the new parameter and link it here, otherwise Firebase notifications will be missing this parameter completely. You may refer to a [past PR](https://github.com/home-assistant/mobile-apps-fcm-push/pull/103) in that repo to see how to add it.

dshokouhi (approved):
Looks good to me, nice to sync up the chronometer countdown with the timeout feature :)

Issue #2867, opened 2022-09-08
BLE transmitter: auto enable/disable depending on home wifi SSID


sibbl:
**Is your feature request related to a problem? Please describe.**
I want to use the BLE transmitting feature for presence detection at home. However, I want to turn of the BLE transmitter when leaving home to avoid being tracked.

**Describe the solution you'd like**
Just like in the location options, I'd love to have an enable/disable BLE transmitter automation in the app, depending on whether I'm in one of the selected HA zones. So when I'm in the zone "home", I want the BLE transmitter to be turned on. When I leave the zone, I want it to be turned off. The companion app should already have geofencing from GPS high accuracy features, so I think this should behave similarly.

**Describe alternatives you've considered, if any**
* I've considered turning off the BLE transmitter by sending a push notification via HA automations. However, this doesn't work reliably sometimes as the HA location doesn't update quick enough or I do have internet connection issues a few minutes after leaving home. 
* I've tried to regularly rotate my BLE IDs. However, this isn't possible with out of the box HA capabilities and messes up a lot of automations as I need to reference the current values, which contain the BLE IDs of the current hour, in all parts of HA (automations and such).

**Additional context**
Especially thanks to the new shiny ESP32 bluetooth proxies, I assume a lot of users will implement presence tracking using BLE. They might enable their BLE transmitter for these reasons without knowing that they are identifiable everywhere by exposing the IDs outside their homes as well.

dshokouhi:
> * I've considered turning off the BLE transmitter by sending a push notification via HA automations. However, this doesn't work reliably sometimes as the HA location doesn't update quick enough

Just a FYI but the in-app automation you are requesting will also be subject to the same delay here, you may want to look at the zone events to see if they are even occurring or if you need to adjust your accuracy. The best data to go by is what gets updated with HA because zone events have no guarantee as to when they will trigger and can vary from user to user. 

A better approach would be to use the Wifi connection sensor as that updates as soon as a state change is detected.  Personally I would prefer to leave these types of automations in the hands of the user instead of the app as the amount of requests for things like this can pick up.

pjv:
@dshokouhi When you say “leave these types of automations in the hands of the user” do you mean having the ability to create in-app automations, or automations written on the HA instance?

[...]

PR #4277, opened 2024-03-17
Add option to transmit BLE on Home SSID only

leccelecce:
## Summary
Sensors -> Bluetooth Sensors -> BLE Transmitter

Added a toggle called "Transmit on Home SSID only". When enabled, beacon transmission will only occur if the device is connected to a Home Network WiFi SSID. 

Closes https://github.com/home-assistant/android/issues/2867

## Screenshots
n/a

Documentation: home-assistant/companion.home-assistant#1046

## Any other notes
This is my first commit to an Android app, please be gentle!
I have only been able to test this on an AVD as I can't get USB or Wifi debugging working on my Samsung S23 (I believe because it has corporate apps installed). However the AVD allowed me to toggle Bluetooth and Wifi and looked to behave correctly.


leccelecce:
I've now managed to test this on my Samsung S23 Ultra by installing the debug APK created and can confirm that it's working for me. This includes also testing that switching from one "home" SSID to a "non-home" SSID disables transmission, as well as turning WiFi off entirely.

dshokouhi:
@leccelecce have you tested this against multiple servers?

leccelecce:
No, I don't have that setup at home currently. I've not looked into it before but can do so if required?

My reading of server Manager.getServer() is that it always returns the currently active server. Is your point that someone could change the currently active server, and that would need to retrigger the home network check?

dshokouhi:
> Is your point that someone could change the currently active server, and that would need to retrigger the home network check?

No the point is that a user may  have multiple servers setup with different BSSID and SSID configured so we need to test that it properly turns on for all servers. Sensor settings impact all servers the user can have added. BSSID and SSID is configured on a per server basis so by looking at just the currently active server ID is not correct as a  user may not have their server active and the check will skip it.

leccelecce:
So I'm still trying to get my head around the multiple server concept. Say a user has:

Server A: Home 1 (using SSID: Wifi X)
Server B: Home 2 (using SSID: Wifi Y)

Are you saying that even if Server A is active, we should accept Wifi Y at Home 2 as valid for transmitting the beacon?

I had assumed that if you connected to Wifi Y, the app would automatically make Server B the active server, but presumably not?

jpelgrom:
The whole 'active' concept only exists for what you're viewing in the app - the dashboard. Everywhere else (sensors, widgets, ...) *all* servers are supposed to be 'active' and stay updated.

leccelecce:
OK, looked around the codebase and updated as follows which hopefully addresses this

```
private fun isPermittedOnThisWifiNetwork(context: Context): Boolean {
        return serverManager(context).defaultServers.any { it.connection.isHomeWifiSsid() }
    }
```

leccelecce:
Oh my god, the more I look at Kotlin the more I love it - just discovered the expression body syntax. Just going to simplify this a little more...there

dshokouhi:
We should also add a line about this setting in the [sensor description](https://github.com/home-assistant/android/blob/master/common/src/main/res/values/strings.xml#L576) as we mention all other settings there.  

Also not sure but should we make the setting title match the setting the user defines the SSIDs under to make it more apparent?

leccelecce:
Done and done

dshokouhi (approved):
Tested with 2 servers each connected to their own network and it respects the setting, even connecting to a 3rd non-home network leaves the transmitter disabled. 

Thank you very much for the contribution :)

leccelecce:
My pleasure, the code changes have turned out to be very simple but needed some more understanding, so thank you for the help in getting to the right solution.

I can see the last Companion App release was 2024.1.5 on Jan 30th; I can't find any information on the release cadence but looks like it is approximately every 2 months? Are the "pre-release" releases on Github all to the beta channel?

dshokouhi:
> I can see the last Companion App release was 2024.1.5 on Jan 30th; I can't find any information on the release cadence but looks like it is approximately every 2 months? Are the "pre-release" releases on Github all to the beta channel?

We currently do not have a set schedule for releasing to production. Every saturday all merged changes get pushed to the beta channel in the play store, after that is when we decide to cut a release. We usually like to let features and fixes get tested by beta testers before pushing changes to production.

leccelecce:
Noted, many thanks! Will enroll in the beta programme next week :-)

PR #4284, opened 2024-03-20
Support showing all frames of animated gif image in Android 14+

NickM-27:
## Summary

This adds support for saving a gif that is sent as a notification image locally so that it can be loaded using a content uri which enables the notification to show the gif as animated including all frames.

Should be noted that this only is only supported on Android 14+, for previous versions the behavior and handling will remain the same where only the first frame of the gif is shown.

## Screenshots

https://github.com/home-assistant/android/assets/14866235/7e3c579a-4a03-440d-b961-9bb05184687c

Documentation: https://github.com/home-assistant/companion.home-assistant/pull/1047

## Any other notes

closes https://github.com/home-assistant/android/issues/1270

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -1255,9 +1255,9 @@ class MessagingManager @Inject constructor(
             }
 
             // delete previous images that are no longer needed
-            val imageCutoff = LocalDateTime.now().minusHours(1)
+            val imageCutoff = LocalDateTime.now().minusDays(2)
             context.externalCacheDir?.listFiles()?.filter { file ->
-                file.endsWith("_animated_notification.gif") &&
+                file.absolutePath.endsWith("_animated_notification.gif") &&
                     imageCutoff.isAfter(LocalDateTime.ofInstant(Instant.ofEpochMilli(file.lastModified()), ZoneId.systemDefault()))
             }?.forEach { expired -> expired.delete() }
 
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -87,6 +87,9 @@ import java.io.File
 import java.io.FileOutputStream
 import java.net.URL
 import java.net.URLDecoder
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.ZoneId
 import javax.inject.Inject
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
@@ -100,10 +103,6 @@ import kotlinx.coroutines.withContext
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import org.json.JSONObject
-import java.time.Instant
-import java.time.LocalDateTime
-import java.time.ZoneId
-import java.util.Calendar
 
 class MessagingManager @Inject constructor(
     @ApplicationContext val context: Context,
@@ -1258,8 +1257,8 @@ class MessagingManager @Inject constructor(
             // delete previous images that are no longer needed
             val imageCutoff = LocalDateTime.now().minusHours(1)
             context.externalCacheDir?.listFiles()?.filter { file ->
-                file.endsWith("_animated_notification.gif")
-                    && imageCutoff.isAfter(LocalDateTime.ofInstant(Instant.ofEpochMilli(file.lastModified()), ZoneId.systemDefault()))
+                file.endsWith("_animated_notification.gif") &&
+                    imageCutoff.isAfter(LocalDateTime.ofInstant(Instant.ofEpochMilli(file.lastModified()), ZoneId.systemDefault()))
             }?.forEach { expired -> expired.delete() }
 
             val file = File(context.externalCacheDir, "${System.currentTimeMillis()}_animated_notification.gif")
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -100,6 +100,10 @@ import kotlinx.coroutines.withContext
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import org.json.JSONObject
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.ZoneId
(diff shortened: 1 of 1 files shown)


dshokouhi:
I think we will need a minor docs change for this as we currently mention GIFs are not animated :)

https://companion.home-assistant.io/docs/notifications/notification-attachments

`GIF will not be animated in notification shade`

`GIF filetype will not be animated in notification shade`

yup we have 2 on the same page 😅

NickM-27:
Yeah I was wondering about that, sounds good I will create that and link it up

dshokouhi (approved):
Tested and works well! Thank you for your contribution!

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -1233,6 +1242,34 @@ class MessagingManager @Inject constructor(
...
 
+    private suspend fun saveTempAnimatedImage(serverId: Int, url: URL?, requiresAuth: Boolean = false): Uri? =
+        withContext(
+            Dispatchers.IO
+        ) {
+            if (url == null || url.path.endsWith("gif").not()) {
+                return@withContext null
+            }
+
+            val file = File(context.externalCacheDir, "animated_notification_image.gif")
```
jpelgrom, line 1253: Using the same file name every time will result in overwriting any previous image. When someone sends two notifications with different gifs, won't this result in it showing the second gif for both notifications?

NickM-27: I was doing that on purpose to alleviate concerns of using too much storage. While implementing I thought that notifications will be handled synchronously so the notification itself will have loaded the image data by the time the next notification comes through. 

Looking at it now and we are using coroutines so there could be a race condition. What would the preferred way of handling images be so that we do not use too much storage on these images? Perhaps remove any images over 1 hour old?

jpelgrom: Another thought: does this need to be external instead of the internal cache dir? Per the documentation internal will be automatically cleared by the system and is not visible to other apps, unlike external.

jpelgrom: > Perhaps remove any images over 1 hour old?

Wouldn't that result in non-working images if you view the notification >1 hour after receiving it?

NickM-27: It does need to be external because otherwise the system is unable to load the file. I tried that first actually but it was failing with file not found.

To be clear,`FileProvider.getUriForFile(context, "${context.packageName}.provider", file)` came at the end and is definitely required. If this works with internal directories then that could be used, but I was assuming it would not

NickM-27: > Wouldn't that result in non-working images if you view the notification >1 hour after receiving it?

my assumption is when the notification is delivered to the device and displayed for the user is when the system reads the image from storage, so in that case I don't think it would be an issue.

If the system doesn't load it until the user actually expands the notification then yeah that would be a problem.

jpelgrom: I guess we should test it before continuing this discussion, otherwise changes will be made where none are required and your code works perfectly fine

NickM-27: I am pulling things up now to test if we can use internal storage. While doing that I will also test multiple notifications

NickM-27: Okay, so my findings are that we must use an external directory, otherwise the `getUriForFile` fails. 

Also, there is a race condition here when multiple notifications are sent simultaneously, however event sent a second apart works as expected. I have pushed a commit to keep the files but delete previous images older than an hour.

NickM-27: extra note to clarify, I did test the case where the screen is off and the screen is turned on after notifications are delivered

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -1233,6 +1245,42 @@ class MessagingManager @Inject constructor(
...
+            Dispatchers.IO
+        ) {
+            if (url == null || url.path.endsWith("gif").not()) {
+                return@withContext null
+            }
+
+            // delete previous images that are no longer needed
+            val imageCutoff = LocalDateTime.now().minusHours(1)
+            context.externalCacheDir?.listFiles()?.filter { file ->
+                file.endsWith("_animated_notification.gif") &&
```
jpelgrom, line 1260: This always returns `false`, probably because of the confusing behavior where it tries comparing components? Suggestion: `file.path.endsWith`.

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
jpelgrom: So while this seems to work initiallly, after fixing deleting (see other comment) at some point the system seems to no longer have the gif in memory and if it was deleted from the directory it's gone, even though you still have the notification. Note the thumbnail still shows up (because that is a Bitmap attached to the notification itself), expanding doesn't show the image anymore:
Closed notification with thumbnail
Opened notification with no image

Quick steps to reproduce: set the cutoff time to 1 minute, send a notification, wait 1 minute, send another notification, and toggle dark mode on/off to force the system to reload stuff without waiting.

NickM-27:
@jpelgrom I see what you mean, what is your suggestion to handle this? We could extend the cutoff to 24 hours assuming no notifications will be kept that long

jpelgrom:
Extending the cutoff to 24-48 hours seems acceptable enough to me while still keeping storage use to a minimum.

jpelgrom (approved):

JBassett (approved):

NickM-27:
@dshokouhi just curious, is there a way to access the apks that are automatically generated or just need to wait for weekly release on github?

dshokouhi:
> @dshokouhi just curious, is there a way to access the apks that are automatically generated or just need to wait for weekly release on github?

any action taht pushes to master will contain the latest APK now like https://github.com/home-assistant/android/actions/runs/8402417849 in the artifact

NickM-27:
awesome, thank you!

PR #4289, opened 2024-03-25
Fix normal images not working after gif changes

NickM-27:
## Summary

This fixes the case where a user is on Android 14+ and sends a normal image that is not a gif.


jpelgrom (approved):
How did we focus so much on the GIFs that we missed this... Thanks for the quick fix!

Issue #3343, opened 2023-02-17
command_webview notification does not change current view on the companion app


osnwt:
**Home Assistant Android app version(s):**
2023.1.1-full

**Android version(s):**
10

**Device model(s):**
Teclast P20HD

**Home Assistant version:**
Home Assistant 2023.2.2
Supervisor 2023.01.1
Operating System 9.5
Frontend 20230202.0

**Last working Home Assistant release (if known):**
N/A

**Description of problem, include YAML if issue is related to notifications:**
According to documentation, the script below should navigate to the specified by path view on a remote Android companion application. In fact, this works only if the companion app is not yet running. So running the script below, it launches and navigates to the requested view. Next, after the current view is changed manually, or another path is specified in the script, and script started again, navigation does not happen, and view is not changed.

```
- service: notify.mobile_app_android_tablet
    data:
      message: command_webview
      data:
        command: /dashboard-main/resources

```

Steps to reproduce:
- close the companion app
- start the script above (using existing view path)
- see the app is open and the view is set as requested
- switch to a different view by hand on the Android app OR change the view path in the script
- start the script again

Expected result: view changes accordingly.
Actual result: view is not changed anymore.

**Companion App Logs:**
In logs below:
19:46:xx - no app running, script launched, requested view opened
19:47:xx - view changed manually on the android app
19:48:xx - scipt launched again, no view changed (expected: change back to requested one)

```
02-17 19:46:25.993 23130 23584 W FirebaseMessaging: Unable to log event: analytics library is missing
02-17 19:46:25.997 23130 23584 D FCMService: From: 331041709873
02-17 19:46:25.999 23130 23584 D MessagingService: Processing device command
02-17 19:46:26.281 23130 23584 W FirebaseMessaging: Unable to log event: analytics library is missing
02-17 19:46:26.283 23130 23584 D FCMService: From: 331041709873
02-17 19:46:26.285 23130 23584 D MessagingService: Processing device command
02-17 19:46:26.514 23130 23584 W FirebaseMessaging: Unable to log event: analytics library is missing
02-17 19:46:26.515 23130 23584 D FCMService: From: 331041709873
02-17 19:46:26.517 23130 23584 D MessagingService: Processing device command
02-17 19:46:26.551 23130 23130 W ActivityThread: handleWindowVisibility: no activity for token android.os.BinderProxy@5df5a59
[...]

PR #4307, opened 2024-03-28
Allow multiple tasks for `command_webview`

jpelgrom:
## Summary

'Fix' #3343 by allowing `command_webview` to create multiple, new tasks. This allows launching the webview to work reliably irrespective of how the app was originally started.

While this behavior may not be perfect in my opinion (you end up with hidden tasks which can be unpredictable), all app features that allow you to navigate to a specific page the webview such as shortcuts, tapping on notifications, device controls long press, etc. use this to reliably load the frontend with a specific page. Different behavior for this specific feature (which as a result didn't always work) doesn't make a lot of sense.

## Screenshots

n/a

Docs don't mention anything about tasks/recents so let's not complicate things

## Any other notes

Based on [these specific reproduction steps](https://github.com/home-assistant/android/issues/3343#issuecomment-2016581587), without this change:
- If the initial task was from the launcher, it seems to create a new activity (which will load the page)
- If the initial task was from `command_webview`, it seems to bring the existing activity to the front and that's it

There must be something here with task stacks/affinity different for launch vs webview as the 'root' that allowed it to work without requiring the multiple tasks flag. I've also tried checking the intent extras in `onNewIntent` for the second point, but it didn't call the function.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -1789,7 +1789,8 @@ class MessagingManager @Inject constructor(
             } else {
                 WebViewActivity.newInstance(context, title, serverId)
             }
-            intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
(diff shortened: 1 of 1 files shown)


osnwt:
Wow, thanks! I hope it will be in release soon.

dshokouhi:
> Wow, thanks! I hope it will be in release soon.

it will be part of tomorrows beta push

PR #4310, opened 2024-03-29
Update change log week of 3/29

dshokouhi:


jpelgrom:
Do the battery sensors work on Automotive? I thought that car battery required the specific car sensor, so adding a note about the voltage divisor may be misleading.

dshokouhi:
> Do the battery sensors work on Automotive? I thought that car battery required the specific car sensor, so adding a note about the voltage divisor may be misleading.

no they do not and they [should be hidden](https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/sensors/BatterySensorManager.kt#L121-L124) good point

jpelgrom (approved):

PR #4326, opened 2024-04-06
Adds http header check after a GIF File

BitWuehler:
Adds a http header check after a GIF File

## Summary
Today I wanted to show a GIF in a notification an Android 14. I saw that this was added befor two weeks, Sadly this only works with URLs that ends with ".gif". Some URL don't have the filename in the url so i added a http header check for that.

The Documentation should be fine. The function is already decribed there.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -1256,7 +1256,7 @@ class MessagingManager @Inject constructor(
         withContext(
             Dispatchers.IO
         ) {
-            if (url == null || url.path.endsWith("gif").not()) {
+            if (url == null || (url.path.endsWith("gif").not() && !isGif(url))) {
                 return@withContext null
             }
 
@@ -1286,6 +1286,20 @@ class MessagingManager @Inject constructor(
             }
             return@withContext FileProvider.getUriForFile(context, "${context.packageName}.provider", file)
         }
+
+    private fun isGif(url: URL): Boolean {
+        val connection = url.openConnection() as? HttpURLConnection ?: return false
+        try {
+            connection.setRequestProperty("User-Agent", "Mozilla/5.0")
+            val contentType = connection.contentType
+            return contentType != null && contentType.startsWith("image/gif")
+        } catch (e: IOException) {
+            Log.e(TAG, "Error checking content type", e)
+            return false
+        } finally {
+            connection.disconnect()
+        }
+    }
 
     private suspend fun handleVideo(
         builder: NotificationCompat.Builder,


app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -1287,6 +1287,18 @@ class MessagingManager @Inject constructor(
             return@withContext FileProvider.getUriForFile(context, "${context.packageName}.provider", file)
         }
 
+    private suspend fun isGif(url: URL): Boolean {
+        try {
+            val connection = url.openConnection()
+            connection.setRequestProperty("User-Agent", "Mozilla/5.0")
```
jpelgrom, line 1293: Any reason for setting the user agent, and if so why not use the [app's user agent] (https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/data/HomeAssistantApis.kt#L27)?

BitWuehler: Not really. Just thought, that this is a good general one. But you are probably right. Il will commit it!

BitWuehler: Do you have an idea how to import the BuildConfig.VERSION_NAME? Seems not imported.

jpelgrom: Better to import the user agent variable I linked than to copy the contents. Add:
```kotlin
import io.homeassistant.companion.android.common.data.HomeAssistantApis
```
and use:
```kotlin
connection.setRequestProperty("User-Agent", HomeAssistantApis.USER_AGENT_STRING)
```

BitWuehler: done

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -1287,6 +1288,18 @@ class MessagingManager @Inject constructor(
             return@withContext FileProvider.getUriForFile(context, "${context.packageName}.provider", file)
         }
 
+    private suspend fun isGif(url: URL): Boolean {
+        try {
+            val connection = url.openConnection()
+            connection.setRequestProperty("User-Agent", HomeAssistantApis.USER_AGENT_STRING)
+            val contentType = connection.contentType
```
jpelgrom, line 1295: It looks like this results in the entire image (or whatever is behind the URL) being downloaded. As the image is also fetched later, that is work that should be avoided as it duplicates data used. Try only getting the headers which should contain the content type. This also doesn't handle the authentication required images (hosted on the HA server).

BitWuehler: You are right. I commit a "connection.requestMethod = "HEAD" for that. 
But what du you mean with "authentication required images (hosted on the HA server)"? I only know direct image paths but these ones should be handled with the filename ending.
Do you maybe mean the use of Image entities? For that this really does not work. But i'm not shure how to resolve this problem. Maybe we could check the file header here?

BitWuehler: I now added a byte checkup for the Entity urls. Im not sure if this works good like this, but i will check it with the generated apk!

BitWuehler: Because I dont have a image entitiy and i dont now how to add one manually i can't test the byte GIF-check with the image entity. I tryed it with the camera entity, because i used a gif in my camera entity but that doesn't work. Do you now, if the camera entity converts the input somehow? If yes i will delete the camera gif-check.

jpelgrom:
I think you've misunderstood what I was saying... There is a `requiresAuth` parameter that should be handled. At this point, paths are never relative because it is transformed [here](https://github.com/home-assistant/android/blob/3849bb8c1a96e1208394962e9ad545fe37fe55ea/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt#L1209). If you're saying there is no way to serve a GIF from one of HA's built in proxies, that's also fine! No need for these complicated checks I think.

BitWuehler:
I now eddited the ByteCheck so it shoukd work now. Maybe it's a bit complicated but I think it would be a god addition. Otherwise you couldn't use gifs from the internat folders.

I would leave out the http authentication for now or do you have a good suggestion for implementation? In any case, I have added the parameter. @jpelgrom

jpelgrom:
As you are clearly still developing I'm going to change this to draft, mark it as easy once you're ready :)

TimoPtr:
@BitWuehler what is the status of this PR? Should we close it

BitWuehler:
@TimoPtr
Not taught about it for a while but it's still on my list.
I'll take care of it soon!

PR #4331, opened 2024-04-07
Add support for voice command action

joseph39:
## Summary

Adding the voice command support so that Assist can be triggered via a headset. Notably the change includes adding the voice command intent for Assist, being able to wake up from it even when the screen is off, and stopping the current playing audio only when the activity is destroyed rather than paused, so that it doesn't interfere with screen turning off half way through the playback.

## Screenshots

N/A

I don't think a doc update is needed?

## Any other notes

This change was tested with/without the voice command and, other than the change in behavior where the playback will continue even with the screen turning off halfway, I've verified that no other UX regression is introduced.


app/src/main/AndroidManifest.xml
```diff
@@ -34,6 +34,7 @@
         tools:ignore="QueryAllPackagesPermission" />
     <uses-permission android:name="com.android.alarm.permission.SET_ALARM" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
+    <uses-permission android:name="android.permission.REQUEST_COMPANION_START_FOREGROUND_SERVICES_FROM_BACKGROUND"/>
```
jpelgrom, line 37: I believe this is only for using the [Companion Device Manager](https://developer.android.com/develop/background-work/services/foreground-services#background-start-restriction-exemptions) (for example smartwatch apps on the phone), which HA does not use or qualify for

app/src/main/java/io/homeassistant/companion/android/assist/AssistActivity.kt
```diff
@@ -129,9 +132,9 @@ class AssistActivity : BaseActivity() {
         viewModel.setPermissionInfo(hasRecordingPermission()) { requestPermission.launch(Manifest.permission.RECORD_AUDIO) }
     }
 
-    override fun onPause() {
-        super.onPause()
-        viewModel.onPause()
+    override fun onDestroy() {
```
jpelgrom, line 135: I can understand changing this to onDestroy for audio playback, but at least for recording it was intentionally on onPause.

A lot of people press home instead of back/finish when exiting apps, and we don't want to have to wait for the timeout for the mic (which can take a while, on server signal and it is a bit slow with it) to stop recording.

jpelgrom (commented):
The automotive manifest should also be updated with these changes.

app/src/main/AndroidManifest.xml
joseph39: Yeah I think it was added when I was trying to get the service to work.

app/src/main/java/io/homeassistant/companion/android/assist/AssistActivity.kt
joseph39: Makes sense; adding `viewModel.onPause` back with just `stopRecording()`

jpelgrom (approved):
Looks good to me

Issue #3934, opened 2023-10-12
Get sensor : time left before full charge


Gus1994:
Following a question on home assistant [here](https://community.home-assistant.io/t/android-phone-get-time-left-before-full-charge/625177)
I would like to have a sensor that gives the time left to charge my phone.
It seems this is somehow accessible [here](https://developer.android.com/reference/android/os/BatteryManager#computeChargeTimeRemaining())

PR #4346, opened 2024-04-17
Add a remaining charge time sensor for supported devices

dshokouhi:
## Summary

Fixes: #3934 by adding the sensor for devices on Android P +

Sensor reports in minutes so have used the `duration` class.  This sensor will only update during the normal sensor update interval and may work better if using the fast while charging sensor update frequency setting.

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#1058

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/BatterySensorManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/BatterySensorManager.kt
@@ -10,7 +10,7 @@ import io.homeassistant.companion.android.common.R as commonR
 import io.homeassistant.companion.android.common.util.STATE_UNAVAILABLE
 import io.homeassistant.companion.android.common.util.STATE_UNKNOWN
 import java.math.RoundingMode
-import kotlin.math.ceil
+import kotlin.math.floor
 
 class BatterySensorManager : SensorManager {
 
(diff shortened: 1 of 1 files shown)


jpelgrom:
This will probably change on *every* sensor update, lots of updates in the state machine.

Would it make sense to round this to minutes (as all devices seem to do this in settings etc.), and/or maybe report is a timestamp?

dshokouhi:
> Would it make sense to round this to minutes (as all devices seem to do this in settings etc.)

This sounds reasonable and good point about too many updates, every minute will also line up with fast while charging update frequency :)



> and/or maybe report is a timestamp?

I think this may still be subject to a lot of updates right? Using the phone while charging may also impact this number as I can see battery temp and voltage being impacted too

jpelgrom:
This does not appear to work correctly / produces results that are different from the settings app and lock screen and seem impossible. (Pixel 7a with Android 14)

The sensor does correctly change from unavailable when unplugged > 0 min at first (-1 internally) > 4 min after a while > stays at 4 min despite charging and the estimate in the settings app changing.

dshokouhi:
Thank you! Both should be corrected in the next push :)

jpelgrom:
Awesome, this seems to match the system's estimate almost perfectly during the ~15 min I observed.

Almost, because the system seems to always show 1 minute less remaining, meaning it rounds down whereas the app rounds up. Can we change that for consistency?

dshokouhi:
> Almost, because the system seems to always show 1 minute less remaining, meaning it rounds down whereas the app rounds up. Can we change that for consistency?

So always round down or round down in certain cases?

jpelgrom:
Right now it looks like always round down is the way to do this.

jpelgrom (approved):

PR #4347, opened 2024-04-18
Use a foreground service for reliable background transmitting

dshokouhi:
## Summary

Hopefully fixes: #3159 by implementing a similar foreground service like in #3369 based on the [docs](https://altbeacon.github.io/android-beacon-library/beacon-transmitter.html) to create a "fake" scanner that looks for nothing to help keep the transmitter alive.  

Needs end user testing. 🤞 

## Screenshots

image

image

## Any other notes

Code may not look pretty for now but wanted to get end user feedback


jpelgrom:
Maybe it's good to split the Bluetooth scan permission into its own PR which we can quickly approve+merge as that fixes an obvious oversight/bug, whereas the foreground service is an effort to improve things but also introduces new behavior?

dshokouhi:
> Maybe it's good to split the Bluetooth scan permission into its own PR which we can quickly approve+merge as that fixes an obvious oversight/bug, whereas the foreground service is an effort to improve things but also introduces new behavior?

actually the scan permission is required in this PR because we are now using the scanner in this sensor where we were not before

jpelgrom:
I missed that's new, sorry. In that case never mind and let's wait for positive results.

dshokouhi:
Agreed this requires end users to confirm its working for them as I cannot [reproduce](https://github.com/home-assistant/android/issues/3159#issuecomment-2064669513)

dshokouhi:
May have taken some time but we got [end user confirmation](https://github.com/home-assistant/android/issues/3159#issuecomment-2161505725) :) probably want to keep this in beta for a lil bit to let it soak

this is now ready for review!

common/src/main/java/io/homeassistant/companion/android/common/bluetooth/ble/TransmitterManager.kt
```diff
@@ -58,9 +63,20 @@ object TransmitterManager {
...
             if (!physicalTransmitter.isStarted) {
+                val builder = beaconNotification(true, context)
+                beaconManager?.enableForegroundServiceScanning(builder.build(), 445)
+                beaconManager?.setEnableScheduledScanJobs(false)
+                beaconManager?.beaconParsers?.clear()
+                beaconManager?.backgroundBetweenScanPeriod = Long.MAX_VALUE
+                beaconManager?.backgroundScanPeriod = 0
+                beaconManager?.foregroundBetweenScanPeriod = Long.MAX_VALUE
+                beaconManager?.foregroundScanPeriod = 0
+                beaconManager?.startMonitoring(region)
```
jpelgrom, line 79: Might be good to add a comment here explaining this enables a foreground service to improve reliability, so that one doesn't need to blame the code to find out why we are scanning in a monitoring service.

Minor cleanup so you don't need `beaconManager?.` everywhere:

```kotlin
beaconManager?.apply {
    enableForegroundServiceScanning(builder.build(), 445)
    setEnableScheduledScanJobs(false)
    // etc
}
```

jpelgrom (commented):
Not really able to test this myself but user comment sounds promising, and I don't see it causing any harm at the moment.

Edit: actually, have you tested this when scanning _and_ transmitting is enabled? Do these changes interfere with regular scanning?

dshokouhi:
> Edit: actually, have you tested this when scanning _and_ transmitting is enabled? Do these changes interfere with regular scanning?

I thought I did that test before but now it seems that only one foreground service can be up at a time 🤔 not sure what to do here. I dont think we can share them either.

dshokouhi:
putting this back in draft while considering new changes based on the [latest test comment](https://github.com/home-assistant/android/issues/3159#issuecomment-2171270366)

dshokouhi:
Based on the most recent [comment](https://github.com/home-assistant/android/issues/3159#issuecomment-2344767595) I am thinking of maybe updating our docs and sensor string description to mention the beacon monitor being enabled.

dshokouhi:
Closing this for now unless changes are actually needed, the issue can probably be closed by a docs update suggesting the user to use the beacon monitor alongside the transmitter as the library itself suggests.

PR #4350, opened 2024-04-19
Add some missing states for audio sensor and correct mic muted intent condition

dshokouhi:
## Summary

Noticed we got new Audio Manager modes for our sensor so adding in the new possible states

https://developer.android.com/reference/android/media/AudioManager#MODE_COMMUNICATION_REDIRECT
https://developer.android.com/reference/android/media/AudioManager#MODE_CALL_REDIRECT

I also noticed that our mic muted SDK condition was incorrect so split it up to be correct.

Documentation: home-assistant/companion.home-assistant#1059

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -163,14 +163,19 @@ open class HomeAssistantApplication : Application() {
             }
         )
 
-        // Add extra audio receiver for devices that support it
+        // Listen for microphone mute changes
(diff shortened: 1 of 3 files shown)


jpelgrom:
> I also noticed that our mic muted SDK condition was incorrect so split it up to be correct.

Don't forget to also update the `updateType` for that sensor then ;)

dshokouhi:
> > I also noticed that our mic muted SDK condition was incorrect so split it up to be correct.
> 
> Don't forget to also update the `updateType` for that sensor then ;)

ah good catch!

jpelgrom (approved):
Looks good to me, and I can't think of a more specific icon for redirected audio.

PR #4358, opened 2024-04-21
Fix popup for reconnecting to the server

### LINKED ISSUE #4272: AutoReConnect
The server is already up! How can I disable this window?

bug

The local server may be temporarily unavailable. Why is there no normal server wait? You need to add another server without accidentally clicking.

I want to see a normal page (not a pop-up!) with the inscription:
```
Waiting for server...

Please return to the server's coverage area or check your connection settings.
```

Roffild:
After receiving an ERROR, the page must be reloaded.

WebView cannot report an error from the server.

Closed #4272


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -470,7 +494,7 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                                 startActivity(browserIntent)
                                 return true
                             } else {
-                                Log.d(TAG, "No unique cases found to override")
+                                // Log.d(TAG, "No unique cases found to override url=${it}") // Unstoppable flood.
```
jpelgrom, line 497: Please revert. We don't leave commented code like this, and if you're seeing this statement a lot that is probably an issue with your setup (it is only called for main page navigation/redirects, so with a normal setup you don't see it unless trying to navigate to outside). Some of the log statements you added will generate more lines than this one.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1511,6 +1544,7 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
     }
 
     private fun waitForConnection() {
+        // Called inside an this.loadUrl() [not webView.loadUrl()]
```
jpelgrom, line 1547: This is called from other functions as well? I don't really understand the comment and there seems to be broken Markdown syntax as well.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -371,7 +394,8 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                     errorResponse: WebResourceResponse?
                 ) {
                     Log.e(TAG, "onReceivedHttpError: ${errorResponse?.statusCode} : ${errorResponse?.reasonPhrase} for: ${request?.url}")
-                    if (request?.url.toString() == loadedUrl) {
+                    if (request?.isForMainFrame == true && request.url.toString() == loadedUrl) {
+                        pageError = errorResponse?.statusCode ?: 404
```
jpelgrom, line 398: Why fallback to 404? HTTP errors can be for various codes; if unknown is a valid response the `pageError` variable should be nullable.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -208,6 +209,8 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
     private var mFilePathCallback: ValueCallback<Array<Uri>>? = null
     private var isConnected = false
     private var isShowingError = false
+    private var isPageFinished = false
+    private var pageError = 0 // 0 - OK, <0 - Connection Error, >=400 - HTTP Error
```
jpelgrom, line 213: What purpose does this have? It is not used in the UI? I can only see it in a log statement when reloading at which point it is no longer relevant because there is reloading 🙃

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -903,6 +927,10 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
             checkAndWarnForDisabledLocation()
         }
         changeLog.showChangeLog(this, false)
+
+        if (::loadedUrl.isInitialized) {
+            reloadPage()
+        }
```
jpelgrom, line 933: Please revert. This will cause a reload with state loss anytime the app is resumed, eg. when opening it from the background.

jpelgrom (changes requested):
Demanding merges via Discord is very much not appreciated.

The relation to the issue (which in itself is hard to follow) is unclear. I'm seeing some attempts to handle errors in another callback here and trying to review those, but a clear issue description would be appreciated.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -338,11 +350,22 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
                     Log.e(TAG, "onReceivedError: errorCode: $errorCode url:$failingUrl")
                     if (failingUrl == loadedUrl) {
+                        pageError = errorCode
+                        showError()
+                    }
+                }
+
+                @RequiresApi(23)
+                override fun onReceivedError(view: WebView?, request: WebResourceRequest?, error: WebResourceError?) {
+                    Log.e(TAG, "onReceivedError: errorCode: ${error?.errorCode} ${error?.description} url:${request?.url}")
```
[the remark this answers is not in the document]
Roffild: This is a copy-paste as in line 351. There is no log flooding here.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
Roffild: Without a URL, the information content of this log is poor.
This log shows the correct URL, which has no additional action.

For example, 3 logs per second:  https://community.home-assistant.io/t/apps-cant-access-ha/534210/3

I was receiving more than 10 logs per second.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1394,27 +1425,29 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
                         "external"
                     }
+                } else {
+                    waitForConnection()
                 }
-                waitForConnection()
             }
             alert.setNeutralButton(commonR.string.wait) { _, _ ->
-                waitForConnection()
+                reloadPage()
```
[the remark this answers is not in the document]
Roffild: `reloadPage` reloads when there is an ERROR + run `waitForConnection`. Only `waitForConnection` does not work on error.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
Roffild: `reloadPage` reloads when there is an ERROR + run `waitForConnection`. Only `waitForConnection` does not work on error.

If the page is normal, then there is no reload. This is needed when returning from Settings, etc.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
Roffild: This is not Markdown. This is a pointer to line 1263.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -329,6 +332,15 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
             settings.mediaPlaybackRequiresUserGesture = !presenter.isAutoPlayVideoEnabled()
             settings.userAgentString = settings.userAgentString + " ${HomeAssistantApis.USER_AGENT_STRING}"
             webViewClient = object : TLSWebViewClient(keyChainRepository) {
+                override fun onPageStarted(view: WebView?, url: String?, favicon: Bitmap?) {
+                    // onReceivedHttpError is getting called before onPageStarted.
+                    // https://issuetracker.google.com/issues/210920403
+                    // onPageStarted is always called before onPageFinished.
+                    // For localhost it is possible to catch onReceived* after onPageFinished !
+                    isPageFinished = false
+                    // Don't pageError=0 here!
```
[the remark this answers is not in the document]
Roffild: The page can load without calling `loadUrl`. Javascript for example.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
Roffild: Used in `reloadPage`. It is possible to show more information about the error to the user.

Roffild:
Full fix for:
https://github.com/Roffild/roffild.github.io/assets/38705255/2e049aea-bc77-482c-8355-d03e22345464

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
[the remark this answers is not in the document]
jpelgrom: The wait button is not for errors. In case of an error, you are supposed to fix it and refresh :)

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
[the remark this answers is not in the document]
jpelgrom: The existing code isn't perfect. I think my comment about this being an unnecessary log statement when nothing is done with it still stands.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: Adding an url could be useful but you're commenting it out so in the end adding no value. And as mentioned, we don't leave in commented code like this.

> I was receiving more than 10 logs per second.

Did you manage to find out what caused it? If so, that could be discussed in an issue and/or changed/fixed in a different PR.

jpelgrom:
> Full fix for: https://github.com/Roffild/roffild.github.io/assets/38705255/2e049aea-bc77-482c-8355-d03e22345464

This still doesn't describe the issue and also doesn't match the code changes which are all over the place. Trying hard to find something here but if we cannot agree on an issue and the value of changes, maybe this should be rejected.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: But you're not showing anything. For debugging with http status codes etc. you may as well use [Chrome remote debugging](http://companion.home-assistant.io/docs/integrations/android-webview/#remote-debugging), the app is never going to match that level of detail.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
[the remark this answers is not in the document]
Roffild: If the WebView has an ERROR, then WAIT doesn't work.
Details below...

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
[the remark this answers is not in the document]
Roffild: Yes, I agree. But this is not my code:
https://github.com/home-assistant/android/blob/2024.5.2/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt#L339-L342
https://github.com/home-assistant/android/blob/2024.5.2/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt#L373-L376

Replace?

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
Roffild: The reason is that shouldOverrideUrlLoading() is constantly being called on another thread.
shouldOverrideUrlLoading() is used to filter special URLs.
If the URL is normal, then the flood will be in the log.

It's better to delete this line.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
Roffild: This is a note:
```kotlin
// BAD:
this.loadUrl()
this.waitForConnection()

// Good:
this.loadUrl()

// Good:
webView.loadUrl()
this.waitForConnection()
```

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
Roffild: Currently used for status in WebView.
It is now possible to show the tip.
`pageError < 0`:
> Check your network connection.

`pageError >= 400` (Android 6.0):
> Check the server configuration.

Roffild:
`showError()` gives a choice to the user. But this choice doesn't know about the WebView's status.

Open http://hass.example.com in your browser. You will get an error. WebView doesn't autoReload. Will you WAIT endlessly (do nothing)? Or do you use page reload (F5)?

`webView.reload()` always reloads the page.

`this.reloadPage()` waits for the page to load or for an error.

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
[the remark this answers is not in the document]
jpelgrom: Sorry, but this still doesn't address the contradiction you've created by making the wait button refresh. I've submitted an alternative in #4412.

jpelgrom:
Closing in favor of #4412

PR #4366, opened 2024-04-26
fastalne i18n ru

yurtpage:
I tried to find if it's possible to translate the app metadata on ht e 
https://app.lokalise.com/
But it looks like it doesn't support it.

So here I added manually.

The Weblate allows to do the App Metadata translation.
Here is a sample of Conversations app metadata translation via Weblate https://translate.codeberg.org/projects/conversations/app-store-metadata-conversations/

Here is some description how to do that:
https://github.com/WeblateOrg/weblate/issues/1525#issuecomment-464086454

Anyway, please merge the PR


dshokouhi:
if this is to be translated then we should probably also add this to lokalise if possible

dshokouhi:
> So now it can be merged.

we dont know if we are going to keep translations in this repo or not so we will not be merging yet

dshokouhi:
> Maybe this can be changed somehow, but given that meta description is not updated that often as in-app translation it shouldn't be a big burden to have them in the fastlane.

Lokalise actually helps by allowing others to do easier language reviews and we dont have to clutter the PR area with translation PRs. It makes things easier to move forward. 

You will need to have some patience while we figure out a way forward here.

jpelgrom:
Closing this as the text has changed, and we don't have a good way to have this be community-managed and community-checked like the other strings at the moment.

PR #4382, opened 2024-05-03
Update change log week of 5/3

dshokouhi:


jpelgrom:
Is the BatterySensorManager and remaining charge time sensor available on Automotive? Reading your description it sounds a bit like remaining charge time for an electric car which I'm not sure this is.

dshokouhi:
> Is the BatterySensorManager and remaining charge time sensor available on Automotive? Reading your description it sounds a bit like remaining charge time for an electric car which I'm not sure this is.

yup good point, just realized that....one of these days ill get it right 😂

jpelgrom (approved):
Dependency updates has never been more true

PR #4391, opened 2024-05-06
Update FCM debug log to match Wear OS for better debugging data

dshokouhi:
## Summary

Noticed our debug log for phone FCM messages did not contain all the important debugging data like notification content. So matching Wear OS output here for easier debugging. Websocket messages is already covered.

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/notifications/FirebaseCloudMessagingService.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/notifications/FirebaseCloudMessagingService.kt
@@ -28,7 +28,7 @@ class FirebaseCloudMessagingService : FirebaseMessagingService() {
     private val mainScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
 
     override fun onMessageReceived(remoteMessage: RemoteMessage) {
-        Log.d(TAG, "From: ${remoteMessage.from}")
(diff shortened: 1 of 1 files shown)


jpelgrom:
I'm not really a big fan of logging so much, even if we do it on Wear. The received notification is immediately stored:

https://github.com/home-assistant/android/blob/d8fc2221418eb9191c66f75788da2074d7a7e638/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt#L287

and unlike Wear, it can be viewed from the app settings (as a JSON object so almost raw!). What are you trying to cover with the log statement that isn't available here? Maybe it's better to expose that in settings.

> Websocket messages is already covered.

Unless I'm missing something, WS doesn't do this. Message contents are only logged on debug builds, and [WebsocketManager](https://github.com/home-assistant/android/blob/d8fc2221418eb9191c66f75788da2074d7a7e638/app/src/main/java/io/homeassistant/companion/android/websocket/WebsocketManager.kt#L172) doesn't log contents for received messages either.

(This is why I'm late to comment on this PR - wanted to take the time to consider and see what we already are doing.)

dshokouhi:
was helping someone troubleshoot an issue and wanted to check the output it was receiving, agreed we dont need this as the history on the phone will be good enough.

PR #4394, opened 2024-05-07
Use device class enum where possible and supply possible states for those sensors

### LINKED ISSUE #4393: [FR] Use `device_class: enum` for string sensors with known enumerable values.

When writing automations with mobile app sensors, like the battery state, the UI does not offer any suggestions for possible states of these sensors:

image

Since these sensors have fixed possible states which can be enumerated, set the sensor device_class to `enum` and enumerate the states. This will improve the UI experience. E.g.:

image

image

**Additional context**

This was primarily raised in relation to battery_state sensor, but probably there are others as well that could also have the same treatment, such as battery health, charger type, etc. 

Thank you!

dshokouhi:
## Summary

Fixes #4393 

the device class is only set for when the new device class was added.

 i decided to keep the options for all users as I can personally see value in it for user on older versions.

i reviewed all our sensors and these ones have a defined list in our code.

## Screenshots

image

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/ActivitySensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/ActivitySensorManager.kt
@@ -40,7 +40,8 @@ class ActivitySensorManager : BroadcastReceiver(), SensorManager {
             "sensor",
             commonR.string.basic_sensor_name_activity,
             commonR.string.sensor_description_detected_activity,
-            "mdi:walk"
+            "mdi:walk",
+            deviceClass = "enum"
         )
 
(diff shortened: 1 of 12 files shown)


jpelgrom:
For `CarSensorManager`: these values seem like they would be localized by the framework? Do we also know if this is a limited list or if Google can add new ones?

dshokouhi:
> I see those are strings we defined now. Can we maybe do a small refactor so the values aren't duplicated?

hmm not really sure I see how we can share a list of strings with the `when` statement like that 🤔

jpelgrom (approved):
Nice QoL improvement!

The 'options' attribute also shows up in-app in the attributes list but I think that's fine as the frontend also shows 'Possible states' as an attribute.

PR #4395, opened 2024-05-08
Use device class for binary sensor friendly state to match frontend and core

dshokouhi:
## Summary

Use the device class for the friendly state translation when the domain is a binary sensor as those have unique states for on/off behavior.

Also updated our link to the proper HA frontend translation

## Any other notes

ref: https://discord.com/channels/330944238910963714/562408603345092636/1237526500715266048


common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -700,70 +700,100 @@ val <T> Entity<T>.friendlyName: String
...
-        "unlocked" -> context.getString(commonR.string.state_unlocked)
-        "unlocking" -> context.getString(commonR.string.state_unlocking)
-        "unknown" -> context.getString(commonR.string.state_unknown)
-        "windy", "windy-variant" -> context.getString(commonR.string.state_windy)
-        else -> state
+    val attributes = this.attributes as Map<String, Any?>
+
+    var friendlyState = when (domain) {
+        // https://github.com/home-assistant/core/blob/dev/homeassistant/components/binary_sensor/strings.json#L113
+        "binary_sensor" -> {
```
jpelgrom, line 707: Switch these two around as the comment only applies to binary sensor (and else has the comment inside the block) 😇

```suggestion
        "binary_sensor" -> {
            // https://github.com/home-assistant/core/blob/dev/homeassistant/components/binary_sensor/strings.json#L113
```

jpelgrom (approved):

PR #4412, opened 2024-05-17
Hide 'Wait' if WebViewActivity error is not due to external bus timeout

jpelgrom:
## Summary

Hide the 'Wait' button in WebViewActivity errors if the error is not due to an external bus timeout, but a more general loading error where there's nothing to wait for. The PR also slightly updates what the refresh button does to ensure loading errors on refresh trigger immediately (instead of waiting for a timeout) + uses the URL it says on the button text.

This PR was inspired by the [(previously unheard) confusion here](https://github.com/home-assistant/android/pull/4358#discussion_r1592796669), and looking back perhaps also by [this suggested change](https://github.com/home-assistant/android/pull/4358/files#diff-76bbedb2918a5db79c7070e9bb2e01d9241388745f60b0719c0d1f74123d89bdR1439) in there (although that change wasn't explained, and incomplete).

## Screenshots

You get a smaller dialog without a wait button:
|Light|Dark|
|-----|-----|
|Companion app with 'Unable to connect to Home Assistant' error, and only two buttons, which can now be displayed on one line instead of three, light mode|Companion app with 'Unable to connect to Home Assistant' error, and only two buttons, which can now be displayed on one line instead of three, dark mode|

n/a

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebView.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebView.kt
@@ -8,7 +8,12 @@ interface WebView {
         AUTHENTICATION,
         SSL,
         SECURITY_WARNING,
-        TIMEOUT
+
+        /** Timeout or general loading error */
+        TIMEOUT_GENERAL,
+
+        /** Timeout due to no 'connection-status: connected' event on the external bus */
+        TIMEOUT_EXTERNAL_BUS
     }
 
     fun loadUrl(url: String, keepHistory: Boolean, openInApp: Boolean)
@@ -23,5 +28,5 @@ interface WebView {
 
     fun unlockAppIfNeeded()
 
-    fun showError(errorType: ErrorType = ErrorType.TIMEOUT, error: SslError? = null, description: String? = null)
+    fun showError(errorType: ErrorType = ErrorType.TIMEOUT_GENERAL, error: SslError? = null, description: String? = null)
 }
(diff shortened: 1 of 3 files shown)


Roffild:
I also started with this simple fix. It doesn't fix ALL problems. The SETTINGS button needs to be fixed too.

I hid WAIT. It's horrible. Android moves buttons. You see 3 buttons. After 10 seconds you see 2 buttons in a different order. A missed click is guaranteed.

Accept my PR. I've already walked this path.

Roffild:
After 10 seconds again:

bug

dshokouhi:
> I also started with this simple fix. It doesn't fix ALL problems.

in order to fix things iteratively we need to go step by step, we won't be able to fix all issues in a single PR but we can begin to walk through the steps and determine what's appropriate

> The SETTINGS button needs to be fixed too.

we should not attempt to fix multiple things here and also users do indeed need to access Settings so not sure why you say it needs to be fixed, its doing what its intended to do, take users to settings to adjust things like their URL.

> Accept my PR.

As mentioned in the previous PR please do not demand that PRs get merged and accepted, there is a process to follow and your PR still has lots of open comments that you did not address.

dshokouhi (approved):
Works as described in my testing, the wait button gets hidden if server is down while attempting to be accessed

Roffild:
I am a professional programmer. I really need a working application.
You don't want or can't reproduce all the error variants from the server.

I tested all the variants. I looked at the Android source code. I would write UnitTests, but there are no tests. (Why?!)

Maybe you can trust me?

JBassett (approved):

jpelgrom:
Sorry, but the only one that seems to be confused by the app's behavior or having issues with it to this extent is you.

Merging as this change was approved by the other 2 main contributors.

Roffild:
@JBassett, have you actually reproduced the problem? Or just approved the patch?

This patch does not work when WebView has cache. You also cannot reproduce bug #4439 without a cache.

This patch fixes external connection checking. My #4358 fixes the internal status of the WebView.

Roffild:
**Conflict in UI/UX:**
Most users do not understand the difference between UPDATE and WAIT. I don't understand the difference either (without source code).

If you want to reduce the size of the Pop-up, then combine the buttons. I want to see a fixed size Pop-up!

As users, do you really like: After start 3 buttons, after 180 seconds 2 buttons in a different order, after 10 again 3 buttons... Really?

PR #4439, opened 2024-06-05
Remove "No unique cases found to override" log (not specific enough)

Roffild:
bug

Problem-Error:
```
'if' must have both main and 'else' branches if used as an expression :467
```

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -470,7 +470,7 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                                 startActivity(browserIntent)
                                 return true
                             } else {
-                                Log.d(TAG, "No unique cases found to override")
+                                // Do nothing.
                             }
                         } catch (e: Exception) {
                             Log.e(TAG, "Unable to override the URL", e)


dshokouhi:
please share the YAML you used to reproduce the flooded log message as i do not see that flooded message at all on my end

Roffild:
https://github.com/home-assistant/android/pull/4358#discussion_r1592809922

Server off-line.

> please share the YAML you used

YAML from the server?

dshokouhi:
> [#4358 (comment)](https://github.com/home-assistant/android/pull/4358#discussion_r1592809922)
> 
> Server off-line.

is this repeated log message caused by using the code in that PR? Because the method will not continuously load as the page does not refresh repeatedly.



> > please share the YAML you used
> 
> YAML from the server?

the method in question gets called when a user clicks on a link in their dashboard, we have specific code in place to intercept links and if you are seeing those repeated log messages then you either have some erroneous dashboard setup or submitting a PR against code changes that are not in the code base.

While I agree the log line is not informative here its important to understand how your seeing repeated messages as us devs do not see that.

Roffild:
Screenshot from testing https://github.com/home-assistant/android/pull/4412 But this bug is not related to this PR.

This is a WebView feature:
After the first HA load, a script for reconnection gets into the cache.
This script requests the URL.
WebView filters the URL in `shouldOverrideUrlLoading`.

[WebViewClient.shouldOverrideUrlLoading()](https://developer.android.com/reference/android/webkit/WebViewClient#shouldOverrideUrlLoading(android.webkit.WebView,%20android.webkit.WebResourceRequest))
>If a WebViewClient is provided, returning true causes the current WebView to abort loading the URL, while returning false causes the WebView to continue loading the URL as usual.

bug

jpelgrom:
The fact that you're seeing the countdown go negative and refresh every second shouldn't happen. If you can reliably reproduce this, [please file a frontend bug](https://github.com/home-assistant/frontend/issues/new?assignees=&labels=bug&projects=&template=bug_report.yml).

Editing PR title to better reflect the change.

Issue #1438, opened 2021-03-19
mobile app to support authelia / other authentication layers


repomanz:
**Is your feature request related to a problem? Please describe.**
My homeassistant server is external and behind reverse proxy (nginx) and authenticating / protecting URL with Authelia.  When I configure home assistant url in mobile application with the has.mydomain.com what happens is that the mobile app opens chrome.

I authenticate through chrome and I'm now logged into home assistant (within chrome).  However the mobile app doesn't work.

**Describe the solution you'd like**
I would like the mobile app to see there is an authentication layer in front of the homeassistant server. So all within the application I would be presented with the authelia login screen and if successful authentication then handed back to the mobile app as an authenticated user.

**Describe alternatives you've considered, if any**
Using google chrome works 100% of the time

**Additional context**
I wanted an additional layer of security in front of home assistant instead of users hitting the home assistant login page directly.

dshokouhi:
It looks like this may still be a reverse proxy issue and not an app issue.  You are seeing a new site open because the HA app is not receiving the HA authentication call back that we are expecting and instead because its a URL to a different domain the browser loads.  This is the same behavior we do in the HA frontend when a user has a link to an external site.  I think you need to make sure that your reverse proxy is properly sending data to where it needs to go.

https://github.com/home-assistant/android/blob/52fec8cffd5101b62083470904008690a9e1e2d8/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationPresenterImpl.kt#L39

repomanz:
Hello - thanks for responding. Here is what I have in nginx proxy manager under advanced settings. Do you see anything wrong here?

`location /authelia {
internal;
set $upstream_authelia http://autheliaserver:9091/api/verify;
proxy_pass_request_body off;
proxy_pass $upstream_authelia;    
proxy_set_header Content-Length "";

# Timeout if the real server is dead
proxy_next_upstream error timeout invalid_header http_500 http_502 http_503;
client_body_buffer_size 128k;
proxy_set_header Host $host;
proxy_set_header X-Original-URL $scheme://$http_host$request_uri;
proxy_set_header X-Real-IP $remote_addr;
proxy_set_header X-Forwarded-For $remote_addr; 
proxy_set_header X-Forwarded-Proto $scheme;
proxy_set_header X-Forwarded-Host $http_host;
proxy_set_header X-Forwarded-Uri $request_uri;
proxy_set_header X-Forwarded-Ssl on;
[...]

PR #4466, opened 2024-06-16
Improve usability when external url is behind a reverse proxy

TechGuard:
## Summary
Just a few quality of life bug fixes that make it easier to use the app when home assistant is hosted behind a reverse proxy.

- Store cookies during the onboarding phase. This is already common behavior for the remainder of the app. Usually a reverse proxy stores a cookie, but this is lost after the onboarding which causes the app to fail on next startup.
- Don't launch the browser in the WebViewActivity when the url changes while the app is not connected yet. Sometimes a reverse proxy wants to show a different page before accepting the connection. This gives the proxy the opportunity to do so. Otherwise the app can launch the browser over 15 times (the app is confused and gets stuck in a loop)

## Screenshots
N/A

Is this necessary for bug fixes?

## Any other notes
Relevant issue: #1438

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
@@ -9,6 +9,7 @@ import android.util.Log
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.webkit.CookieManager
 import android.webkit.SslErrorHandler
 import android.webkit.WebResourceError
 import android.webkit.WebResourceRequest
@@ -158,6 +159,11 @@ class AuthenticationFragment : Fragment() {
                                     showError(requireContext().getString(commonR.string.error_ssl), error, null)
                                 }
                             }
+
+                            val cookieManager = CookieManager.getInstance()
+                            cookieManager.setAcceptCookie(true)
+                            cookieManager.setAcceptThirdPartyCookies(this, true)
+
                             authUrl = buildAuthUrl(viewModel.manualUrl.value)
                             loadUrl(authUrl!!)
                         }
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -463,7 +463,7 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                                     startActivity(intent)
                                 }
                                 return true
-                            } else if (!webView.url.toString().contains(it.toString())) {
+                            } else if (isConnected && !webView.url.toString().contains(it.toString())) {
                                 Log.d(TAG, "Launching browser")
                                 val browserIntent = Intent(Intent.ACTION_VIEW, it)
                                 startActivity(browserIntent)


jpelgrom:
I'm assuming you're referring to a reverse proxy *that adds its own auth/session handling on top of Home Assistant*? Reverse proxies in itself (sending data between host <> HA) shouldn't require any changes.

> Don't launch the browser in the WebViewActivity when the url changes while the app is not connected yet.

We cannot accept this. The `WebViewActivity` is a 'trusted' context and exposes interfaces that should not be accessible to other sites, such as a way to get authentication data. Allowing redirecting to any site breaks this trust.

Accepting this would basically make the attacks mentioned in https://github.com/home-assistant/core/security/advisories/GHSA-jvpm-q3hq-86rg a feature if you can redirect the app loading a site (MITM).

> Otherwise the app can launch the browser over 15 times (the app is confused and gets stuck in a loop)

This sounds like something that should be able to be fixed (it is supposed to be one redirect, so only opening the browser once).

TechGuard:
> I'm assuming you're referring to a reverse proxy _that adds its own auth/session handling on top of Home Assistant_? Reverse proxies in itself (sending data between host <> HA) shouldn't require any changes.
Yes, I should have specified. In my example I'm running Authelia in front of all my services and it requires 2FA.

> We cannot accept this. The `WebViewActivity` is a 'trusted' context and exposes interfaces that should not be accessible to other sites, such as a way to get authentication data. Allowing redirecting to any site breaks this trust.
> 
> Accepting this would basically make the attacks mentioned in [GHSA-jvpm-q3hq-86rg](https://github.com/home-assistant/core/security/advisories/GHSA-jvpm-q3hq-86rg) a feature if you can redirect the app loading a site (MITM).
My bad, I didn't realize. Is that the also the main difference (exposed interfaces) between the onboarding webview vs the `WebViewActivity`? Because the onboarding page works quite well with a reverse proxy, I was trying to replicate the same behavior. What would be an acceptable change to support this kind of functionality?

And how do feel about my change of allowing the onboarding page to store the cookies? That actually makes it work quite well. If we can't change the behavior of the `WebViewActivity`, it would be enough to remove and re-add the server again as a workaround.

jpelgrom:
> Is that the also the main difference ...

Yes/onboarding is not HA app specific and shouldn't include app specific changes to the webview as a result. This discussion is not new, I don't see a way to add more support to the main app without compromises for normal setups. The solution require HA core to [properly support 3rd party auth providers](https://community.home-assistant.io/t/open-letter-for-improving-home-assistants-authentication-system-oidc-sso/494223?u=jpelgrom) like you're using - or for now use TLS client certificates if you want something on top of HA, which are fully supported by the app.

> And how do feel about my change of allowing the onboarding page to store the cookies?

That's fine and would be accepted as it matches what is set in the WebViewActivity. Cookies are allowed by default but I'm guessing you want the 3rd party cookies for your setup.

TechGuard:
Thanks for all the info @jpelgrom, I realize how big of a change this would be so I've changed this back to draft. I will get back to the drawing board to see if there's anything else I could do or if it will just be the cookie change.

jjcf89:
Sad to see this got rejected.  I'm also using an oauth proxy in front of HA, and the redirect to chrome breaks the authentication workflow.  

I can auth for the proxy, then get redirected to HA, and auth for HA. But that only manages to auth HA in chrome and the app is left out in the cold.

PR #4477, opened 2024-06-22
[CameraWidget] Add picker for action type on CameraWidgetConfigureAct…

michael-oknov:
## Summary
The CameraWidgetEntity class has been extended to include WidgetTapAction as a new field (column), allowing the application to handle different actions on widget taps

## Screenshots
| Light Mode | Dark Mode |
|------------|-----------|
| image | image |

Documentation: home-assistant/companion.home-assistant#1071

## Any other notes
N/A


jpelgrom:
Hey, thanks for the suggestion. I think the option to view instead of refresh, and switching to Compose for configuration, are both good additions. We prefer to [keep PRs as small as possible](https://developers.home-assistant.io/docs/review-process#creating-the-perfect-pr) however, and considering reviewers' time is currently limited, could you split this up?

For the tap action, something very similar already exists with the entity state widget and the class + strings were intended for general use so I also would recommend building on that unless there's a good reason not to.

michael-oknov:
@jpelgrom Thanks for your feedback. You`re right, I should have done the Compose migration as a separate PR. My fault, will fix. 

Regarding the reuse of WidgetTapAction instead of creating a new one, CameraWidgetTapAction, I considered it but encountered some issues. Different widgets require different action lists. We could add all possible actions into one enum. Still, different widget types might require different sets of actions, and we would need to ensure proper filtering on each configuration screen. Additionally, based on the data, we can`t be certain that the correct action from the enum is preserved, so we`d also need to implement extra handling during the tap action.

What do you think? Can I leave it like this, or would it be better to use the same enum for all actions?

jpelgrom:
> Regarding the reuse of WidgetTapAction instead of creating a new one, CameraWidgetTapAction, I considered it but encountered some issues. Different widgets require different action lists. We could add all possible actions into one enum. Still, different widget types might require different sets of actions, and we would need to ensure proper filtering on each configuration screen. Additionally, based on the data, we can't be certain that the correct action from the enum is preserved, so we'd also need to implement extra handling during the tap action.
> 
> What do you think? Can I leave it like this, or would it be better to use the same enum for all actions?

I think what matters most is that you considered it and didn't miss it. The current implementation with one enum seems to have all the right checks in place, and considering we can add/change/remove tap actions in the future I don't really see how those would've been avoided in a separate enum.  Not sure filtering is a big concern right now either, and as the database stores text it can always be split in the future.

Reusing the strings will be appreciated by non-English users :)

jpelgrom (approved):
No comments, looks good and works as expected!

PR #4499, opened 2024-07-08
Add device id to Assist pipeline runs

### LINKED ISSUE #4158: Assist area awareness support
2024.1.2-full
**Android version(s):**
12
**Device model(s):**
Samsung galaxy S10e
**Home Assistant version:**
2024.1.2
NA

Assist should be aware of the area of a voice assistant. If I set the area of my android device within HASS, sentences that require this implicit area do not work (for example, 'turn on the lights'). This would be useful for me (and maybe others) as I have an android device stuck to a wall as a control interface, and I'd like to be able to say 'turn on the lights' to it and have that make sense.

I've traced this through in the hass codebase, and it appears that the websocket pipeline for voice assistant takes `device_id` in the websocket message when it starts off here: https://github.com/home-assistant/core/blob/1fa7ceede390df037fe9bdb1ba4f7939c9acc138/homeassistant/components/assist_pipeline/websocket_api.py#L144

If this is supplied then somewhere else (intent handling) the device ID is translated to area, which is then supplied in the context for interpreting the intent. 

I raised this against hass core here https://github.com/home-assistant/core/issues/108839 and the understanding on that side is that it's the responsibility of the phone app to tell the API what it needs to know.

I'm happy to try and work on this if someone can give me a pointer. I'm guessing if the app's own `device_id` is lying around somewhere then it is a matter of poking it into here: https://github.com/home-assistant/android/blob/d9e65f8e5daf9bfb94808f545d29624af67f5ded/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt#L285
or somewhere nearby.

**Companion App Logs:**
NA
NA
**Additional information:**
NA

jpelgrom:
## Summary

When running Assist pipelines, include the device ID if we have it so users can do location based Assist requests. Requires home-assistant/core#121496 (core 2024.7.2+). Fixes #4158.

## Screenshots

n/a

n/a

## Any other notes

I basically copied @bgoncal's iOS work here

### DIFF
--- a/common/schemas/io.homeassistant.companion.android.database.AppDatabase/47.json
+++ b/common/schemas/io.homeassistant.companion.android.database.AppDatabase/47.json
@@ -1,0 +1,1127 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 47,
+    "identityHash": "f02f3b0062cfc7559bbd23f8148fb3b1",
+    "entities": [
+      {
+        "tableName": "sensor_attributes",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`sensor_id` TEXT NOT NULL, `name` TEXT NOT NULL, `value` TEXT NOT NULL, `value_type` TEXT NOT NULL, PRIMARY KEY(`sensor_id`, `name`))",
+        "fields": [
+          {
+            "fieldPath": "sensorId",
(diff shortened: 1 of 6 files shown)


bgoncal (approved):

dshokouhi:
Going to merge now that 2024.7.2 has been released

PR #4507, opened 2024-07-12
Update DND sensor icon based on state

mikosoft83:
## Summary
Adding icons based on DND sensor state. Makes for easier visual aid.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/DNDSensorManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/DNDSensorManager.kt
@@ -81,7 +81,7 @@ class DNDSensorManager : SensorManager {
             context,
             dndSensor,
             state,
-            icon,
+            if (state != "off") dndSensor.statelessIcon else "mdi:minus-circle-off",
             mapOf(
                 "options" to listOf("alarms_only", "off", "priority_only", "total_silence")
             )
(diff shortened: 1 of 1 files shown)


jpelgrom:
Not sure I agree with this change as modern Android versions use DND as on/off (switching between priority_only and off). The icons you're using sort of resemble ringer mode.

mikosoft83:
Would it make sense then to clump all DND modes together using single icon for "on" and another for "off"?

jpelgrom:
If there must be an icon change yes, but as said I don't think the icon should be changed as the system doesn't change icons depending on on/off.

mikosoft83:
At least in my version of android, no, the icon itself doesn't change but the background changes as a visual indication.
And in HA itself the binary sensors often have different icons for on and off state.

common/src/main/java/io/homeassistant/companion/android/common/sensors/DNDSensorManager.kt
```diff
@@ -72,12 +72,16 @@ class DNDSensorManager : SensorManager {
...
+            "off" -> "mdi:minus-circle-off"
+            else -> "mdi:minus-circle"
+        }
 
         onSensorUpdated(
             context,
             dndSensor,
             state,
-            dndSensor.statelessIcon,
+            icon,
```
dshokouhi, line 84: You can simplify this and remove the need for the `icon` variable defined above

```suggestion
            if (state == "off") dndSensor.statelessIcon else "mdi:minus-circle",
```

dshokouhi (approved):

PR #4516, opened 2024-07-15
Add setting to control inclusion of all notification contents as attributes in the Active Notification Count sensor

marazmarci:
## Summary

Currently, the Active Notification Sensor attaches the content of all notifications in the sensor attributes. I think this has two problems:

1. The fact that, this sensor also includes all notification content in addition to the count is not intuitive judging only based on the sensor's name, and can surprise users (including myself).
2. It can increase power and data consumption.

And these could surprise the user, maybe in a negative way.

So here comes a new toggle setting for the Active Notification Count sensor to the rescue: `Include content from all notifications as sensor attributes`

Documentation: https://github.com/home-assistant/companion.home-assistant/pull/1077

## Any other notes
Please check [this comment](https://github.com/home-assistant/android/pull/4516/files#r1677876942).

## Screenshots
Screenshot_20240715_190818_Home Assistant

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
@@ -27,6 +27,7 @@ class NotificationSensorManager : NotificationListenerService(), SensorManager {
         private const val TAG = "NotificationManager"
         private const val SETTING_ALLOW_LIST = "notification_allow_list"
         private const val SETTING_DISABLE_ALLOW_LIST = "notification_disable_allow_list"
+        private const val SETTING_INCLUDE_CONTENTS_AS_ATTRS = "active_notification_count_content_attrs"
 
         private var listenerConnected = false
         val lastNotification = SensorManager.BasicSensor(
@@ -98,6 +99,7 @@ class NotificationSensorManager : NotificationListenerService(), SensorManager {
 
     override fun requestSensorUpdate(context: Context) {
         updateMediaSession(context)
+        updateActiveNotificationCount()
     }
 
     override fun onListenerConnected() {
@@ -237,16 +239,19 @@ class NotificationSensorManager : NotificationListenerService(), SensorManager {
 
         try {
             val attr: MutableMap<String, Any?> = mutableMapOf()
-            for (item in activeNotifications) {
-                attr += mappedBundle(item.notification.extras, "_${item.packageName}_${item.id}").orEmpty()
-                    .plus("${item.packageName}_${item.id}_post_time" to item.postTime)
-                    .plus("${item.packageName}_${item.id}_is_ongoing" to item.isOngoing)
-                    .plus("${item.packageName}_${item.id}_is_clearable" to item.isClearable)
-                    .plus("${item.packageName}_${item.id}_group_id" to item.notification.group)
-                    .plus("${item.packageName}_${item.id}_category" to item.notification.category)
+            val includeContentsAsAttrsSetting = getToggleSetting(applicationContext, activeNotificationCount, SETTING_INCLUDE_CONTENTS_AS_ATTRS, default = true)
+            if (includeContentsAsAttrsSetting) {
+                for (item in activeNotifications) {
+                    attr += mappedBundle(item.notification.extras, "_${item.packageName}_${item.id}").orEmpty()
+                        .plus("${item.packageName}_${item.id}_post_time" to item.postTime)
+                        .plus("${item.packageName}_${item.id}_is_ongoing" to item.isOngoing)
+                        .plus("${item.packageName}_${item.id}_is_clearable" to item.isClearable)
+                        .plus("${item.packageName}_${item.id}_group_id" to item.notification.group)
+                        .plus("${item.packageName}_${item.id}_category" to item.notification.category)
 
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                    attr["${item.packageName}_${item.id}_channel_id"] = item.notification.channelId
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                        attr["${item.packageName}_${item.id}_channel_id"] = item.notification.channelId
+                    }
                 }
             }
             onSensorUpdated(
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -734,7 +734,7 @@
     <string name="sensor_setting_nextalarm_allow_list_title">Allow list</string>
     <string name="sensor_setting_notification_allow_list_title">Allow list</string>
     <string name="sensor_setting_notification_disable_allow_list_title">Disable allow list requirement</string>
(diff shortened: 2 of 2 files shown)


app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
```diff
@@ -98,6 +99,7 @@ class NotificationSensorManager : NotificationListenerService(), SensorManager {
 
     override fun requestSensorUpdate(context: Context) {
         updateMediaSession(context)
+        updateActiveNotificationCount()
```
marazmarci, line 102: I noticed that this call was missing here. Maybe it was intentional, but I couldn't come up with an explanation why it wouldn't belong here.

If you think this should be a part of a separate PR, or if it was intentionally not called and I should remove it, please let me know!

dshokouhi:
Is this changing default behavior? We should avoid a breaking change here and keep existing behavior as default.

app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
dshokouhi: Did you test this change? Last I checked is only accessible in certain places and may cause an error if not initialized.

dshokouhi:
Maybe it makes more sense to just add 1 line to the sensor description mentioning the attributes if its that confusing?

marazmarci:
> Is this changing default behavior? We should avoid a breaking change here and keep existing behavior as default.

No, it's not a breaking change, since I made `true` the setting's default value 🙂

app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
dshokouhi: I remember now `requestSensorUpdate` is not going to make a difference here because we update the sensor when items are added and removed from the status bar. Also we only have access to the status bar when the listener is connected which happens as items get added/removed. I think there might be an old bug on this change.

marazmarci: One thing I forgot to mention: without this call, the setting didn't appear until the first notification got posted/removed.

dshokouhi: yea that also happens for last notification sensors as well, allow list setting wont appear until the code is reachable.

marazmarci: `SensorManager::requestSensorUpdate` is called [here](https://github.com/home-assistant/android/blob/15b1518a4466e9c58d26efce7e4605858b88a689/app/src/main/java/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt#L217) in `SensorDetailViewModel::setEnabled`.

marazmarci: > Did you test this change? Last I checked is only accessible in certain places and may cause an error if not initialized.

Thanks for pointing this out. I actually tested it, but only lightly. I tried enabling/disabling the sensor, and triggering sensor updates.

marazmarci: I just wanted to point out the `|| !listenerConnected` part in the early return condition 😄

> there may still be times where the listener is not connected and a setting value wont be created

Do you think it is a blocker problem? Eventually the setting will be created, right?

Or I can move the creation of the setting somewhere else where it is guaranteed that it will be called when enabling the sensor. (But I think a proper solution would be to handle the sensor settings more centrally and generally across all sensors.)

dshokouhi: Its really only these sensors because the code only becomes active when things are posted or removed. The other sensors all provide settings upon being enabled as the code is always reachable. Maybe we should consider a setting initialization step?

we have had that issue with last notification since the setting was introduced so its not a blocker by any means given the amount of time its been like that

marazmarci:
> Maybe it makes more sense to just add 1 line to the sensor description mentioning the attributes if its that confusing?

Would you support if we both added 1 line to the description and also the new sensor setting?

dshokouhi:
> > Maybe it makes more sense to just add 1 line to the sensor description mentioning the attributes if its that confusing?
> 
> Would you support if we both added 1 line to the description and also the new sensor setting?

sure thats fine with me

app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
marazmarci: > Maybe we should consider a setting initialization step?

Yes, that would be a good direction, assuming that we want to keep the current architecture which lists sensor settings from the local DB, and not some statically defined sensor definition. (And yes, I know that the sensor settings list can contain dynamic number of items, like the Last Notification sensor, that's why it would be complicated to refactor)

dshokouhi: Yea I think if we wanted to do a refactor that would be way out of scope for this PR. Getting the setting to update and appear faster leads to a better user experience IMO. Unless we want to change how it currently is I think it should be ok as we are already checking if the listener is connected for us from the past PR.

marazmarci:
I extended the in-app sensor description. Also, I converted the PR to a draft, as I realized that it won't/shouldn't be merged until the documentation PR is ready.

app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
marazmarci: > Yea I think if we wanted to do a refactor that would be way out of scope for this PR.

I agree!

> Getting the setting to update and appear faster leads to a better user experience IMO.

I agree!

> Unless we want to change how it currently is I think it should be ok as we are already checking if the listener is connected for us from the past PR.

OK, I wouldn't change it now.

common/src/main/res/values/strings.xml
```diff
@@ -734,6 +734,7 @@
     <string name="sensor_setting_nextalarm_allow_list_title">Allow list</string>
     <string name="sensor_setting_notification_allow_list_title">Allow list</string>
     <string name="sensor_setting_notification_disable_allow_list_title">Disable allow list requirement</string>
+    <string name="sensor_setting_active_notification_count_content_attrs_title">Include content from all notifications as sensor attributes</string>
```
marazmarci, line 737: Now rereading it, I think the word `sensor` is unnecessary here. WDYT?

```suggestion
    <string name="sensor_setting_active_notification_count_content_attrs_title">Include content from all notifications as attributes</string>
```

dshokouhi: I think that makes sense :)

dshokouhi (approved):
Thanks for the improvement!

Issue #4544, opened 2024-08-01
Location update will be marked as duplicate even when the previous one was not sent due an error


edenhaus:
**Home Assistant Android app version(s):**
2024.7.5-full

**Android version(s):**
Android 14

**Device model(s):**
Oneplus 8T

**Home Assistant version:**
2024.7.4

**Last working Home Assistant release (if known):**

**Description of problem, include YAML if issue is related to notifications:**
Due to an internet outage, the app couldn't send any location updates and reported them as error, which is fine.
When the internet came back, I was able to access HA within the app successfully, but all further location updates were marked as duplicates, and so HA assumed I was still in a different place.

Proposal:
The app should only mark a location update as duplicate when the last one was successfully sent. Otherwise, the app should send it to the HA instance.

Workaround:
Go for a walk with Google Maps open until a new location update comes in, which is not marked as duplicate...

If you need more information, you can also write me on discord

**Companion App Logs:**

```

```

**Screenshot or video of problem:**

![Screenshot](https://github.com/user-attachments/assets/811a4c6e-5a50-4c4c-9eca-55b3c3e3f16f)

**Additional information:**

dshokouhi:
Looks like we may need to move the positioning of when the variables are updated as we are indeed setting them before understanding if the call was successful or not, good catch!

https://github.com/home-assistant/android/blob/master/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt#L920-L921

PR #4545, opened 2024-08-01
Only update location sent variables when successful

dshokouhi:
## Summary

Fixes: #4544 by only setting the variables IF the call was successful

## Any other notes

Brief testing looks to be ok, I dont recall why we had these variables before understanding if the call was successful or not.


jpelgrom (approved):

jpelgrom:
> I dont recall why we had these variables before understanding if the call was successful or not.

Blaming the code gives me #888 as the original source and I guess it just wasn't considered.

Issue #3290, opened 2023-01-31
Additional notification type to change the reference power transmitted with iBeacon advertisements


rpatel3001:
**Is your feature request related to a problem? Please describe.**
The Wear OS app currently does not allow adjusting BLE beacon parameters except via notification. I would like to change the reference power to improve distance measurement accuracy.

**Describe the solution you'd like**
An additional command for message type command_ble_transmitter called ble_set_reference_power taking a float parameter ble_reference

**Describe alternatives you've considered, if any**
Ignoring the included power and using a custom value on the backend. Not ideal.

**Additional context**
The reference power is called the measured power in the phone app settings so s/reference/measured/ above if you want.

agarbato:
I have the same exact need, change rssi1m as default -59 is completely wrong.
Even with station at 10cm espresence report 3 meters which won't work for all stations max distance settings.
Is it possible at least to have a default value as -80,-85 which is definitely more realistic for 1m distance (at least for all my other devices).
Thanks!

dshokouhi:
> Is it possible at least to have a default value as -80,-85 which is definitely more realistic for 1m distance (at least for all my other devices).

a lot of research went into the current default value here: https://github.com/home-assistant/android/issues/1715

Either way thats not the same as this request so please file a separate request to discuss things, I will say be prepared to have some data to backup your findings compared to what was discovered in the issue

agarbato:
Thanks @dshokouhi .
I found this already [#3359 ](https://github.com/home-assistant/android/issues/3359) which brought me here.
Smart watches have less power antennas (mine is a Ticwatch pro5). Still we have 3 android phones and we had to change setting to -80 -85 to make distance realistic.   Anyway I guess it depends on the device but for smart watches -59 1m does not really works so being able to change it via notification would be great :-)

OptimusGREEN:
Any movement on this feature? Desperate to use our watches for room presense. 
It would be a great feature.

PR #4546, opened 2024-08-01
Allow measured power to be updated via notification command

dshokouhi:
## Summary

Implements: #3290

example service call:

```
service: notify.mobile_app_pixel_8_pro
data:
  message: command_ble_transmitter
  data:
    command: ble_set_measured_power
    ble_measured_power: "-10"
```

The field is validated that its not blank and that it is a negative number

Documentation: home-assistant/companion.home-assistant#1086

## Any other notes

FCM PR: https://github.com/home-assistant/mobile-apps-fcm-push/pull/150

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/notifications/DeviceCommands.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/notifications/DeviceCommands.kt
@@ -2,7 +2,6 @@ package io.homeassistant.companion.android.common.notifications
 
 import android.content.Context
 import android.util.Log
-import androidx.core.text.isDigitsOnly
 import io.homeassistant.companion.android.common.sensors.BluetoothSensorManager
 import io.homeassistant.companion.android.common.sensors.SensorUpdateReceiver
 import io.homeassistant.companion.android.database.sensor.SensorDao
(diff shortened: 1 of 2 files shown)


jpelgrom:
> in some cases the notification will post on the device, in other cases the setting may revert to default

Why have it behave differently, instead of always ignoring it (which is the default behavior with notification commands for invalid values)?

dshokouhi:
> > in some cases the notification will post on the device, in other cases the setting may revert to default
> 
> Why have it behave differently, instead of always ignoring it (which is the default behavior with notification commands for invalid values)?

Added a check to make sure that doesn't happen, should be good now :)

dshokouhi:
That was why i added the check before to capture any other bad value,  added `toIntOrNull` because wanted to avoid a crash. I tested with teh following

- negative numbers - works
- positive numbers - posts notification
- alphanumeric - posts notification
- blank - posts notification

jpelgrom:
But why not do `toIntOrNull() != null` before instead of this convoluted check? The `orNull` is literally there for bad values.

And even if there is a bad (positive) value, it is still set as you don't change the value in line 160?

```kotlin
data[NotificationData.COMMAND] == DeviceCommandData.BLE_SET_MEASURED_POWER &&
    (
        data[DeviceCommandData.BLE_MEASURED_POWER].toIntOrNull() != null && data[DeviceCommandData.BLE_MEASURED_POWER]?.toInt() < 0
        )
```

jpelgrom:
I see my comment is inconsistent with current behavior - invalid changed to something valid for UUID/major/minor.

https://github.com/home-assistant/android/blob/485a84271887478f4715d767b21422d6d9587048/common/src/main/java/io/homeassistant/companion/android/common/notifications/DeviceCommands.kt#L144-L145

But that is the exception to the default which ignores the command if invalid (`in` check):

https://github.com/home-assistant/android/blob/485a84271887478f4715d767b21422d6d9587048/common/src/main/java/io/homeassistant/companion/android/common/notifications/DeviceCommands.kt#L68

dshokouhi:
Your check is simpler and for some reason I did not consider it, lets go with that as its easier to read :)

jpelgrom (approved):

Issue #4500, opened 2024-07-09
Wear OS: Update app to target API 33 by August 30 for play store policy


dshokouhi:
Skipping the template here. 

We need to update Wear OS to API 33 by August 30 or the app wont be updated in the play store after that date.

![image](https://github.com/home-assistant/android/assets/1634145/73fb21b3-4d09-4c10-a93d-9de3ee358ed0)

This API level had the requirement for requesting notification permissions which needs to be done during onboarding, amongst other things.

jpelgrom:
Looking at the behavior changes I only see 2 (documented) items that are relevant for us:

 - https://developer.android.com/develop/ui/views/notifications/notification-permission: This won't be too hard, we should probably ask for it when the app is opened (≈ current behavior) and maybe add a quick option for it in settings to make it easier to find. On some devices (Samsung) this may be linked to the ability to receive phone notifications on the watch as well so probably best to keep the description general.
 - https://developer.android.com/about/versions/13/behavior-changes-13#body-sensors-background-permission: Looks like we may need to disable the heart rate sensor on devices running API 33+, as HA won't be allowlisted for background use anywhere and only having it work while the app is open is pretty useless.

dshokouhi:
ouch thats not going to be fun regarding teh heart sensor. I remember a feature request for adding the update sensors command to the watch so users can get the heart rate sensor updating more frequently.

adding heart rate sensor to health services may work but i fear it may lead to more battery drain because it updates frequently, going to require some testing there.

dshokouhi:
Regarding the heart rate sensor I just found something specific to Wear OS. Looks like users will be able to grant the app background permission :)

https://developer.android.com/training/wearables/versions/4/changes#background-body-sensor-permission

https://developer.android.com/health-and-fitness/guides/health-services/background-body-sensors

`Note: BODY_SENSORS_BACKGROUND is a restricted permission which cannot be held by your app until the installer adds your app to an allowlist, or until the user lets your app have the permission.`

**or until the user lets your app have the permission.** So this must correspond to the "All the time" setting I imagine.

jpelgrom:
I cannot choose all the time after targeting API 30. Is there a manufacturer difference or am I making an error?

dshokouhi:
yea looks like the option is hidden a bit

[...]

PR #4551, opened 2024-08-03
Handle notification permission on Wear

jpelgrom:
## Summary

To prepare for the API 33 targeting of Wear (#4500), add handling around the notification permission to the Wear app. This should only apply to watch notifications, not bridged notifications*, but it is nice to have if possible so:

 - Ask for notification permission immediately after completing onboarding (Android 13+ only)
 - Add a settings item to open the watch settings to grant permission (any Android version)

## Screenshots

|Permission prompt after logging in|Settings item if notifications are disabled for the app|Tapping the setting open the system notifications settings**|
|----|----|----|
|A screenshot of a dialog with the question 'Allow Home Assistant to send you notifications?', a bell icon, an obscured button near the bottom.|Settings titled 'Notifications', one option labeled 'Enable notifications'|System notification settings for Home Assistant, with a enable everything toggle at the top and individual notification channels below|

n/a

## Any other notes

\* At least [that's what Google says](https://developer.android.com/training/wearables/versions/4/changes#:~:text=The%20notification%20permission%20doesn%27t%20apply%20to%20bridged%20notifications), but then you have Samsung making it impossible to change one without the other in their companion app 🤷
\*\* Fun fact: while I can find this in settings app on the emulator, I actually can't find it on my Samsung watch

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/HomeActivity.kt
@@ -1,11 +1,15 @@
 package io.homeassistant.companion.android.home
 
+import android.Manifest
 import android.content.Context
 import android.content.Intent
+import android.os.Build
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
+import androidx.activity.result.contract.ActivityResultContracts
 import androidx.activity.viewModels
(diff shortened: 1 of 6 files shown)


dshokouhi (approved):
Hopefully we dont have to make it clear this is for watch notifications only and not bridged :) Not sure why Samsung decided to hide the option, seems silly but maybe they did it for simplicity?

PR #4552, opened 2024-08-03
Breaking Change: Update heart rate sensor for Wear OS 4+ target

jpelgrom:
## Summary

Update the permission requirements for the heart rate sensor on Wear OS 4+ to handle the new behavior when targeting it. These permissions need to be requested in two stages like background location. Any user with the sensor currently enabled on Wear OS 4+, will need to re-enable it after updating.

## Screenshots

n/a

~~Documentation: home-assistant/companion.home-assistant#1087~~ n/a

### DIFF
--- a/wear/src/main/AndroidManifest.xml
+++ b/wear/src/main/AndroidManifest.xml
@@ -11,6 +11,7 @@
     <uses-permission android:name="android.permission.ACTIVITY_RECOGNITION" />
     <uses-permission android:name="android.permission.READ_PHONE_STATE" />
     <uses-permission android:name="android.permission.BODY_SENSORS" />
+    <uses-permission android:name="android.permission.BODY_SENSORS_BACKGROUND" />
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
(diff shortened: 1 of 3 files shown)


dshokouhi (approved):
Well this sucks, hopefully Google will change that some day or hopefully health connect will allow it to be pulled. I agree lets merge alongside the API bump.

dshokouhi:
Given what we know in https://github.com/home-assistant/android/issues/4500#issuecomment-2303848622 should we hold off on this?

jpelgrom:
Oh I completely forgot about your comment, I still wanted to test this because I couldn't find it the first time. I'll do that in a couple of hours and comment.

jpelgrom:
You're right, I must have missed something during previous tests (or an update happened). Handling this like background location with requesting the main permission first, then the background permission, seems to work. Updating the PR to change it.

(Still a breaking change in a way: because a permission is added, users need to toggle the sensor on again.)

dshokouhi (approved):
Thank you for testing and updating the PR so we don't lose a feature :)

Issue #4550, opened 2024-08-02
WearOS: Option for adding an 'Assist' shortcut to Shortcut-Tiles


RoboMagus:
**Is your feature request related to a problem? Please describe.**

Currently the Assist feature on WearOS can be launched through complications, through the app menu, or by its own Assist-Tile. However as I'd like to have all my home assistant related entries on a single page, it'd be nice to have a shortcut for Assist added to the Shortcuts-Tile.

**Describe the solution you'd like**

Add the option to not just add entities to the shortcuts tile, but also a shortcut to the Assist intent (and perhaps also a shortcut to the main HA wear app itself). This way a user does not have to swipe to a different page to get access to this feature.

**Describe alternatives you've considered, if any**

**Additional context**

Happy to take a look at implementing this myself, if you're open to merging a PR for such a feature.

dshokouhi:
Another Assist entry point sounds great!

PR #4556, opened 2024-08-05
WearOS: Add app activities to ShortcutsTile

RoboMagus:
## Summary

This adds the option to add buttons to launch either the Companion app, or the Assist dialog to the ShortcutsTile. 
See #4550

## Screenshots

Shortcut tile with newly added buttons;
ShortcutTile

Config for this example;
ShortcutTileConfig

New button options are added at the bottom of the entity selection list under the `shortcuts` section;
ShortcutActivityOptions

## Any other notes

1. This implementation has been tested only on an AndroidStudio emulator.
2. It seems that the HA Icon included in the `CommunityMaterial.Icon` library has not yet been updated.
3. I'm not fluent in Kotlin / Android app development, so if there's any improvements to be made I'm happy to learn :)

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/tiles/TileActionReceiver.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/tiles/TileActionReceiver.kt
@@ -39,8 +39,7 @@ class TileActionReceiver : BroadcastReceiver() {
                     it.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
                     context.startActivity(it)
                 }
-            }
-            else {
+            } else {
                 runBlocking {
                     if (wearPrefsRepository.getWearHapticFeedback() && context != null) hapticClick(context)
 
--- a/wear/src/main/java/io/homeassistant/companion/android/tiles/TileActionReceiver.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/tiles/TileActionReceiver.kt
@@ -8,6 +8,8 @@ import dagger.hilt.android.AndroidEntryPoint
 import io.homeassistant.companion.android.common.data.integration.onEntityPressedWithoutState
 import io.homeassistant.companion.android.common.data.prefs.WearPrefsRepository
 import io.homeassistant.companion.android.common.data.servers.ServerManager
+import io.homeassistant.companion.android.conversation.ConversationActivity
+import io.homeassistant.companion.android.home.HomeActivity
 import javax.inject.Inject
 import kotlinx.coroutines.runBlocking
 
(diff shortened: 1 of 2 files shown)


dshokouhi:
> * It seems that the HA Icon included in the `CommunityMaterial.Icon` library has not yet been updated.

Yea I am not sure it will be acceptable to include the old branding there 🤔

wear/src/main/java/io/homeassistant/companion/android/tiles/TileActionReceiver.kt
```diff
@@ -28,16 +30,27 @@ class TileActionReceiver : BroadcastReceiver() {
...
+                    "assist" to ConversationActivity.newInstance(context),
+                    "home_assistant" to HomeActivity.newInstance(context)
+                )
+                m[entityId.split(".")[1]]?.let {
+                    it.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                    context.startActivity(it)
+                }
+            } else {
+                runBlocking {
+                    if (wearPrefsRepository.getWearHapticFeedback() && context != null) hapticClick(context)
```
dshokouhi, line 44: I think users may expect to continue to get feedback on clicks in the tile, I think we have added it in other places as well

jpelgrom:
> > * It seems that the HA Icon included in the `CommunityMaterial.Icon` library has not yet been updated.
> 
> Yea I am not sure it will be acceptable to include the old branding there 🤔

MDI is dropping brand icons in the next major release so probably not open to updating, so I think the way forward is to load the icon from the app's resources in this specific case.

RoboMagus:
> Yea I am not sure it will be acceptable to include the old branding there 🤔

Well, any entities inside HA that have its icon set to `mdi:home-assistant` will still show the old icon too. 
So this would best be addressed at the icon library side.

wear/src/main/java/io/homeassistant/companion/android/tiles/TileActionReceiver.kt
RoboMagus: Adding the haptic feedback when the icon is clicked?

Sure that's easily added. However for clicking on entity shortcuts there is no inherent feedback other than the added haptics. These 2 additions would open an app on screen, which is likely feedback enough by itself.

dshokouhi:
> > Yea I am not sure it will be acceptable to include the old branding there 🤔
> 
> Well, any entities inside HA that have its icon set to `mdi:home-assistant` will still show the old icon too. So this would best be addressed at the icon library side.

Yea for any entities a user has defined is fine, but since we are creating this entry it should use the appropriate icon

wear/src/main/java/io/homeassistant/companion/android/tiles/TileActionReceiver.kt
dshokouhi: yea in the past we have gotten requests from users for things not related to entity action we can leave it out for now and add it back upon request

RoboMagus:
If you agree with this approach I'd prefer looking into creating a separate PR which would patch any use of `mdi:home-assistant` to be in line with what would be shown on the web version of the HA instance for consistency across platforms. That way any use of the HA icon would be using the new icon instead of the old one.

jpelgrom:
Changing this to draft while waiting for the other PR

Issue #4572, opened 2024-08-11
Mobile Network Strength Sensor


camsaway:
**Is your feature request related to a problem? Please describe.**

Not a problem. But more information I want access to. I want to be able to track mobile network availability and signal quality.

**Describe the solution you'd like**

The Phone Sensor -SIM1 and SIM2 to have extra attribute(s) that show mobile network Strength and connection mode, etc.

**Describe alternatives you've considered, if any**

I don't have any other ideas/alternatives.

**Additional context**

Below screenshot from network cell info shows the things I'd like to track RSRP (signal strength) and DataNetwork(e.g. LTE, 5G, etc)

![Screenshot_20240811_144732_Network Cell Info Lite](https://github.com/user-attachments/assets/83cf4fd2-f043-48ed-ac16-ce89a884f7b7)

igorsantos07:
If the attributes are already there, you could simply create a template sensor exposing those values directly as a sensor, and on new HA versions, even link it to the original device.

dshokouhi:
a sensor like this would never update immediately, it would update during the normal sensor update interval. Would that even be useful here?

camsaway:
Unfortunately the attributes aren't there yet. That's what this request is
for. (Screenshot is from another app to show that the data does exist)

In terms of frequency, it's a fair point. I hadn't considered that.  15min
default update and 1 min fast update should do for my use case though I
think. I'm not trying to create some sort of coverage mapping tool.

So in short, I still think it would be useful to add the attribute.

Regards


Cameron Gray

079 528 4802

***@***.***


https://www.linkedin.com/in/camerongray76

dshokouhi:
> The Phone Sensor -SIM1 and SIM2 to have extra attribute(s) that show mobile network Strength and connection mode, etc.

this should not be an attribute but a separate sensor for signal strength

camsaway:
Happy either way but all the other info is as attributes under the SIM card
sensor (with multiple simcards you will get multiple values) so I was just
proposing to continue this pattern.

Regards


Cameron Gray

079 528 4802

***@***.***


https://www.linkedin.com/in/camerongray76

PR #4575, opened 2024-08-12
Add signal strength and data network type sensors

dshokouhi:
## Summary

Fixes: #4572 by adding 4 new sensors for signal strength and data network type for each SIM slot

## Screenshots

image

image

image

Documentation: home-assistant/companion.home-assistant#1094

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -26,7 +26,6 @@
     <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
     <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
     <uses-permission android:name="android.permission.NFC" />
-    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
     <uses-permission android:name="android.permission.ACTIVITY_RECOGNITION" />
     <uses-permission android:name="com.google.android.gms.permission.ACTIVITY_RECOGNITION" />
     <uses-permission android:name="android.permission.ACCESS_NOTIFICATION_POLICY" />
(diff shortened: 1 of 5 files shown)


jpelgrom:
Should/can these by split up for SIM 1 and 2?

dshokouhi:
may not be simple for me to test as I only use one. FWIW my phone reports its using SIM2 but still reports data. I assume the first option in the list is probably the active SIM card?

jpelgrom:
I think only one can be active for data at a time, but both can have signal strength and be active in standby (you can receive a call on either SIM).

dshokouhi:
thats a good point, unfortunately the API seems to only give a list of entries but none of them have anything to help identify which operator it belongs to. The API even states teh order of the list items may not always be the same so we cant assume SIM indexes either. 🤔 My own personal assumption is whichever was most recently actively used is probably first?

jpelgrom:
So after testing this we should definitely split this up, it seems to handle two sims but it's confusing from which one the data is coming. @dshokouhi as discussed in DMs I've pushed changes to your branch because you're unable to test this yourself and describing here while I could push is pointless, so please review 🤝

Additionally, for the data network type sensor branding is an issue 😅 as 4G+/LTE Advanced and some 5G networks still show as LTE because they technically are ([display info overrides](https://developer.android.com/about/versions/11/features/5g#detection)). We should add a note to the docs.

jpelgrom:
Updated sensor list: 
List of app sensors, titled 'Phone sensors' with new options for 'Data network type (SIM 1)', 'Data network type (SIM 2)', 'Signal strength (SIM 1)', 'Signal strength (SIM 2)'

dshokouhi:
this works great on my end! Thank you for fixing it!!!

image

jpelgrom (approved):
No more comments from me, feel free to merge if you're happy as well

PR #4587, opened 2024-08-18
Patch `mdi:home-assistant` icon

RoboMagus:
## Summary

Replace the old `mdi:home-assistant` icon with the new one, wherever the MDI icon is used. (e.g. entity icons in WearOS app, and custom notification icons)

As the used _mikepenz.iconics_ icon library uses Fonts as the underlying mechanism that is used throughout this codebase, a 1 glyph font is added and a wrapper class that provides that font with the interfaces used to render it through the aforementioned library.

It does feel like a bit of a workaround, but the alternative would likely require much more changes throughout the codebase and this change is as contained as can be.

## Screenshots

Based on work done in #4556, as I'm not sure what other places the `mdi:home-assistant` icon could appear.
image

## Any other notes

This PR would be a prerequisite for #4556.


jpelgrom:
Can you share how you generated the font?

> I'm not sure what other places the mdi:home-assistant icon could appear.

In other places where the `IconicsDrawable(context, name)` constructor is used. Just searching for the constructor you can find multiple usages in the code base so I'm honestly not sure what you're doing now is that effective. Is there any reason not to replace all of those with `CommunityMaterial.getIconByMdiName`?

RoboMagus:
> Can you share how you generated the font?

That was honestly a lot of just messy trial and error using some online font generators to get the HA icon in the [mikepenz CommunityMaterial Fonts](
https://github.com/mikepenz/Android-Iconics/blob/develop/community-material-typeface-library/src/main/res/font/community_material_design_icons_font_v7_0_96.ttf) replaced with the new SVG logo from the [HA assets](https://github.com/home-assistant/assets) repo, and getting it lined up with the font borders so it would render at the same size and offset as all other icons in that set. After I figured out we'd only need the one glyph in the font I used `pyftsubset` from [fonttools](https://github.com/fonttools/fonttools) to remove anything but the one glyph required.

> ... I'm honestly not sure what you're doing now is that effective. Is there any reason not to replace all of those with CommunityMaterial.getIconByMdiName?

What this PR covers thus far is the direct use of the old icon (`CommunityMaterial.Icon2.cmd_home_assistant`), what is directly queried by `CommunityMaterial.getIconByMdiName()` and the calls to `entity.getIcon()`. 
It does look like more places need to be covered and this could be done as you suggest by using `CommunityMaterial.getIconByMdiName`, but it'd be a shame to invest effort into tackling all the edge cases if the concept of this change would be shot down anyways.

If we can reach consensus about the best way of globally updating the old `mdi:home-assistant` icon with the new one I'll update the rest of the codebase where these icons could be queried. 

Broad use of the `IconicsDrawable` that is provided by the mikepenz library throughout the codebase makes adding a custom font that can be used by this library the least invasive change from what I can tell. But I'm open to alternative solutions for this issue.

jpelgrom:
> It does look like more places need to be covered and this could be done as you suggest by using `CommunityMaterial.getIconByMdiName`, but it'd be a shame to invest effort into tackling all the edge cases if the concept of this change would be shot down anyways.
> 
> If we can reach consensus about the best way of globally updating the old `mdi:home-assistant` icon with the new one I'll update the rest of the codebase where these icons could be queried.
> 
> Broad use of the `IconicsDrawable` that is provided by the mikepenz library throughout the codebase makes adding a custom font that can be used by this library the least invasive change from what I can tell. But I'm open to alternative solutions for this issue.

I agree with this solution and would accept it, but it needs to be implemented consistently/everywhere which is why I commented. The `mdi` prefix is used to determine the font so patching needs to use 'our' function to override the icon returned.

@dshokouhi any objections?

dshokouhi:
> any objections?

none from me :)

common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -303,6 +304,9 @@ fun <T> Entity<T>.getIcon(context: Context): IIcon {
     val icon = attributes["icon"] as? String
     return if (icon?.startsWith("mdi") == true) {
         val mdiIcon = icon.split(":")[1]
+        if (mdiIcon == "home-assistant") {
+            return HaIconTypeface.Icon.mdi_home_assistant
+        }
         return IconicsDrawable(context, "cmd-$mdiIcon").icon ?: CommunityMaterial.Icon.cmd_bookmark
```
jpelgrom, line 310: Any particular reason for not using `CommunityMaterial.getIconByMdiName` here?

frenck:

../Frenck

Issue #4549, opened 2024-08-02
Update Android widgets to match new IOS widget style


dnestico:
**Is your feature request related to a problem? Please describe.**
I personally think the current Android Home Assistant widgets look outdated and are much worse then the new IOS widgets that came out last week. 

Last week the IOS HA app got a new updated that brought all new Tile card style looking widgets which look much better then the old widgets and there is also a button/action group that allows you to have multiple actions/buttons in one widget, below is an attached image of the new IOS widgets.

**Describe the solution you'd like**
The solution would be to copy the design of the new IOS HA widgets and bring them to the Android HA app, specifically the widget that holds multiple actions/buttons in one widget would be the best one. 

Below is a picture of what the IOS HA widget looks like. 
<img width="357" alt="Screenshot 2024-08-01 at 10 44 39 PM" src="https://github.com/user-attachments/assets/7d68a0ce-5790-409c-9d47-5437e8e8162d">

**Describe alternatives you've considered, if any**
None, other than using the old widgets.	

**Additional context**
Attached above is a picture of the new IOS widgets that would look great on Android devices too.

PR #4594, opened 2024-08-26
UI updates for ButtonWidget

PizzaMarinara:
## Summary
Implementation of a fresher look for Button widget as suggested in #4549, adding a parameter to make the icon's color customizable.

## Screenshots
Config:

Day:

Night:

Dynamic:

Transparent:

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/button/ButtonWidget.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/button/ButtonWidget.kt
@@ -67,6 +67,7 @@ class ButtonWidget : AppWidgetProvider() {
 
         // Vector icon rendering resolution fallback (if we can't infer via AppWidgetManager for some reason)
         private const val DEFAULT_MAX_ICON_SIZE = 512
+        private const val SMALL_MAX_ICON_SIZE = 80
     }
 
     @Inject
@@ -202,6 +203,9 @@ class ButtonWidget : AppWidgetProvider() {
             val icon = DrawableCompat.wrap(iconDrawable)
             if (widget?.backgroundType == WidgetBackgroundType.TRANSPARENT) {
                 setInt(R.id.widgetImageButton, "setColorFilter", textColor)
+                setInt(R.id.widgetImageButton, "setBackgroundColor", Color.TRANSPARENT)
+            } else {
+                setInt(R.id.widgetImageButton, "setColorFilter", Color.TRANSPARENT)
             }
 
             // Determine reasonable dimensions for drawing vector icon as a bitmap
@@ -210,11 +214,11 @@ class ButtonWidget : AppWidgetProvider() {
             val maxWidth = (
                 awo?.getInt(AppWidgetManager.OPTION_APPWIDGET_MAX_WIDTH, DEFAULT_MAX_ICON_SIZE)
                     ?: DEFAULT_MAX_ICON_SIZE
-                ).coerceAtLeast(16)
+                ).coerceAtLeast(16).coerceAtMost(SMALL_MAX_ICON_SIZE)
             val maxHeight = (
                 awo?.getInt(AppWidgetManager.OPTION_APPWIDGET_MAX_HEIGHT, DEFAULT_MAX_ICON_SIZE)
                     ?: DEFAULT_MAX_ICON_SIZE
-                ).coerceAtLeast(16)
+                ).coerceAtLeast(16).coerceAtMost(SMALL_MAX_ICON_SIZE)
             val width: Int
             val height: Int
             if (maxWidth > maxHeight) {
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/button/ButtonWidget.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/button/ButtonWidget.kt
@@ -61,6 +61,7 @@ class ButtonWidget : AppWidgetProvider() {
         internal const val EXTRA_ACTION_DATA = "EXTRA_SERVICE_DATA"
         internal const val EXTRA_LABEL = "EXTRA_LABEL"
         internal const val EXTRA_ICON_NAME = "EXTRA_ICON_NAME"
+        internal const val EXTRA_ICON_COLOR = "EXTRA_ICON_COLOR"
         internal const val EXTRA_BACKGROUND_TYPE = "EXTRA_BACKGROUND_TYPE"
         internal const val EXTRA_TEXT_COLOR = "EXTRA_TEXT_COLOR"
         internal const val EXTRA_REQUIRE_AUTHENTICATION = "EXTRA_REQUIRE_AUTHENTICATION"
@@ -232,6 +233,13 @@ class ButtonWidget : AppWidgetProvider() {
             // Render the icon into the Button's ImageView
             setImageViewBitmap(R.id.widgetImageButton, icon.toBitmap(width, height))
 
+            widget?.iconColor?.let {
+                if(it.isNotBlank()) {
+                    setInt(R.id.widgetImageButton, "setColorFilter", Color.parseColor(it))
+                    setInt(R.id.widgetImageButtonBackground, "setColorFilter", Color.parseColor(it))
+                }
+            }
+
             setOnClickPendingIntent(
                 R.id.widgetImageButtonLayout,
                 PendingIntent.getBroadcast(
@@ -367,6 +375,7 @@ class ButtonWidget : AppWidgetProvider() {
         val label: String? = extras.getString(EXTRA_LABEL)
         val requireAuthentication: Boolean = extras.getBoolean(EXTRA_REQUIRE_AUTHENTICATION)
         val icon: String = extras.getString(EXTRA_ICON_NAME) ?: "mdi:flash"
+        val iconColor: String? = extras.getString(EXTRA_ICON_COLOR)
         val backgroundType = BundleCompat.getSerializable(extras, EXTRA_BACKGROUND_TYPE, WidgetBackgroundType::class.java)
             ?: WidgetBackgroundType.DAYNIGHT
         val textColor: String? = extras.getString(EXTRA_TEXT_COLOR)
@@ -387,7 +396,7 @@ class ButtonWidget : AppWidgetProvider() {
                     "label: " + label
             )
 
-            val widget = ButtonWidgetEntity(appWidgetId, serverId, icon, domain, action, actionData, label, backgroundType, textColor, requireAuthentication)
+            val widget = ButtonWidgetEntity(appWidgetId, serverId, icon, domain, action, actionData, label, iconColor, backgroundType, textColor, requireAuthentication)
             buttonWidgetDao.add(widget)
 
             // It is the responsibility of the configuration activity to update the app widget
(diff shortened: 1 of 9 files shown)


bgoncal:
Thats looks nice! I see some different top and leading padding, can you check that? For reference, this is the iOS view https://github.com/home-assistant/iOS/blob/master/Sources/Extensions/Widgets/Common/WidgetBasicView.swift
Also, how big of a change it would be to allow multiple items inside the same widget like iOS has?

jpelgrom:
> Also, how big of a change it would be to allow multiple items inside the same widget like iOS has?

Quite big, let's not increase the scope of the PR here. (And personally, I think a combined widget with entity state, which has an easy toggle/run <entity> instead of requiring you to configure each action, would make more sense for large amounts of items.)

PizzaMarinara:
> Thats looks nice! I see some different top and leading padding, can you check that? For reference, this is the iOS view https://github.com/home-assistant/iOS/blob/master/Sources/Extensions/Widgets/Common/WidgetBasicView.swift Also, how big of a change it would be to allow multiple items inside the same widget like iOS has?

Hi @bgoncal, the different padding is totally a personal preference given the default appearance of the widget is more rectangular shaped I tried leaving less padding horizontally than vertically to allow a little bit more space for the label. I can change it to an uniform padding (given the widget can be also resized). Also as @jpelgrom mentioned, I assume a multi-item widget would be a completely different kind of widget, with its own configuration so possibly a bit too outside the scope of the PR.

app/src/main/res/layout/widget_button.xml
```diff
@@ -10,44 +10,51 @@
...
+        android:layout_height="wrap_content"
+        android:paddingVertical="8dp"
+        android:paddingHorizontal="4dp"
         android:orientation="vertical">
 
         <ImageView
+            android:id="@+id/widgetImageButtonBackground"
+            android:layout_width="40dp"
+            android:layout_height="40dp"
+            android:translationZ="-2dp"
```
jpelgrom, line 21: Why use `translationZ` which is relative, instead of `elevation`?

app/src/main/res/layout/widget_button.xml
```diff
@@ -10,44 +10,51 @@
...
+            android:alpha="0.4"
+            android:contentDescription="@string/widget_button_image_description"
```
jpelgrom, line 23: We don't need the same accessibility descriptions on both the background and the icon. I think you can just leave it out here / mark it as no description needed.

bgoncal:
Agree on having a separate PR for multiple items widget ✅
About the UI, it's not really personal preference, you are proposing a change to align it more with iOS widget style, so, while for the text I may agree it is useful to have more space, for the icon it looks misaligned when it's top marging is different from it's leading margin, unless this is something common in Material Design and I am unaware

app/src/main/java/io/homeassistant/companion/android/widgets/button/ButtonWidget.kt
```diff
@@ -228,6 +233,13 @@ class ButtonWidget : AppWidgetProvider() {
             // Render the icon into the Button's ImageView
             setImageViewBitmap(R.id.widgetImageButton, icon.toBitmap(width, height))
 
+            widget?.iconColor?.let {
+                if (it.isNotBlank()) {
+                    setInt(R.id.widgetImageButton, "setColorFilter", Color.parseColor(it))
+                    setInt(R.id.widgetImageButtonBackground, "setColorFilter", Color.parseColor(it))
```
jpelgrom, line 239: This will throw an exception if the input is not valid and crash:
```
2024-08-27 12:12:31.708   747-747   AndroidRuntime          pid-747                              E  FATAL EXCEPTION: main
                                                                                                    Process: io.homeassistant.companion.android.debug, PID: 747
                                                                                                    java.lang.IllegalArgumentException: Unknown color
                                                                                                    	at android.graphics.Color.parseColor(Color.java:1399)
                                                                                                    	at io.homeassistant.companion.android.widgets.button.ButtonWidget.getWidgetRemoteViews(ButtonWidget.kt:238)
                                                                                                    	at io.homeassistant.companion.android.widgets.button.ButtonWidget.access$getWidgetRemoteViews(ButtonWidget.kt:47)
                                                                                                    	at io.homeassistant.companion.android.widgets.button.ButtonWidget$updateAllWidgets$1.invokeSuspend(ButtonWidget.kt:114)
                                                                                                    	at android.os.Handler.handleCallback(Handler.java:959)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:100)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:232)
                                                                                                    	at android.os.Looper.loop(Looper.java:317)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:8592)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:580)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:878)
                                                                                                    	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@61859a4, Dispatchers.Main]
```

app/src/main/java/io/homeassistant/companion/android/widgets/button/ButtonWidgetConfigureActivity.kt
```diff
@@ -454,6 +455,10 @@ class ButtonWidgetConfigureActivity : BaseWidgetConfigureActivity() {
                 ButtonWidget.EXTRA_ICON_NAME,
                 binding.widgetConfigIconSelector.tag as String
             )
+            intent.putExtra(
+                ButtonWidget.EXTRA_ICON_COLOR,
+                binding.iconColor.text.toString()
+            )
```
jpelgrom, line 461: Please add validation to prevent invalid input (see crash mentioned). If blank, the value should not be sent to avoid saving unnecessary blank text in the database.

app/src/main/res/layout/widget_button.xml
```diff
@@ -10,44 +10,51 @@
     android:background="@drawable/widget_button_background"
     android:padding="4dp">
 
-    <LinearLayout
+    <RelativeLayout
         android:id="@+id/widgetImageButtonLayout"
         android:layout_width="match_parent"
-        android:layout_height="match_parent"
-        android:paddingVertical="4dp"
+        android:layout_height="wrap_content"
```
jpelgrom, line 14: The size change will create visual problems with longer labels and/or smaller widget sizes:
Widget with text overlapping icon
(default size, and this label isn't even that long)

app/src/main/res/layout/widget_button_configure.xml
```diff
@@ -117,6 +117,28 @@
...
+                android:layout_gravity="center_vertical"
+                android:padding="5dp"
+                android:text="@string/label_icon_color" />
+
+            <androidx.appcompat.widget.AppCompatEditText
+                android:id="@+id/icon_color"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:hint="@string/widget_text_hint_icon_color"
+                android:imeOptions="actionDone"
```
jpelgrom, line 131: This should not be set, as this isn't the last text field (so the keyboard action should continue to the next field).

app/src/main/res/layout/widget_button_configure.xml
```diff
@@ -117,6 +117,28 @@
...
+                android:inputType="text" />
+        </LinearLayout>
+
+        <LinearLayout
+            android:layout_width="match_parent"
+            android:layout_height="50dp"
+            android:gravity="center"
+            android:orientation="horizontal">
```
jpelgrom, line 139: This appears to be unused / not closed

PizzaMarinara:
Added a commit that fixes most of @jpelgrom's remarks.

> Why use translationZ which is relative, instead of elevation?
I believe they're equivalent? I'm not sure if there's any specific reason to use either.

app/src/main/res/drawable/rounded_image_background.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android">
+    <solid android:color="@color/colorDynamicWidgetPrimary" />
```
jpelgrom, line 3: There are two issues here:
 - this value is for the dynamic color theme and should not be used in other themes
 - this value does not exist on older Android versions (note how it is declared in the `values-v31` directory)

app/src/main/res/layout/widget_button.xml
jpelgrom: This has not been resolved. It may be less obvious now, but text is cut off (note the bottom of the g/p), and especially for users with larger font sizes configured this will happen quite easily.
image

PizzaMarinara: Yeah I figured the issue is now caused by the text's size rather than its layout overlapping the above icon. I can reproduce it only if I manually set my font size in the phone's setting to the maximum, otherwise it looks good to me. Is that the case here as well?

jpelgrom: No, this is on the default text and display size for me (Pixel 7a; text size 2/7, display size 2/5).

app/src/main/res/layout/widget_button.xml
jpelgrom: > I believe they're equivalent? I'm not sure if there's any specific reason to use either.

They both work, but for static values [`elevation` is suggested while `translation` is suggested for animation](https://developer.android.com/develop/ui/views/theming/shadows-clipping#Elevation). And here, it actually seems to be useless as it is already the lowest item.

```suggestion
```

app/src/main/java/io/homeassistant/companion/android/widgets/button/ButtonWidget.kt
```diff
@@ -228,6 +233,10 @@ class ButtonWidget : AppWidgetProvider() {
             // Render the icon into the Button's ImageView
             setImageViewBitmap(R.id.widgetImageButton, icon.toBitmap(width, height))
 
+            widget?.iconColor?.let {
+                setCustomColorToIcon(it, this)
```
jpelgrom, line 237: This will overwrite the color set to the icon in line 204/206, which is fine if you have set a custom color but when using the transparent theme it might create a mismatch (black icon on a colored circle with white text, while the option says it'll color text + icon, and it just looks odd).
Transparent widget: blue circle with black icon, and white text below

app/src/main/res/drawable/rounded_image_background.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android">
+    <solid android:color="@color/colorPrimary" />
```
jpelgrom, line 3: Use the widget's theme color as a default instead of always blue, or pick something which doesn't look out of place in a dynamic color theme (which is very hard and why theme colors exist, see main comment). The [widget's background](https://github.com/home-assistant/android/blob/2279d5d7dfb6151a0b4c3ac4ccdb60f251bc2056/app/src/main/res/drawable/widget_button_background.xml#L4) is an example of how they are used in another file.

```suggestion
    <solid android:color="?colorWidgetButton" />
```

jpelgrom (commented):
The layout changes in the latest commit completely change the design, no longer being close to the requested tile design, with small text and with (too) narrow padding. If I'm being honest, it looks broken even without the tile design reference.

Updated tile design: blue circle with a black icon in the top left, on a yellow dynamic color background, with small text right next to the curved corner

app/src/main/res/layout/widget_button.xml
jpelgrom: Please see [this review comment](https://github.com/home-assistant/android/pull/4594#pullrequestreview-2266977219).

dshokouhi:
Hi @PizzaMarinara just wanted to check in with you to see if you wanted to pick up this PR again? its about 4 months old with some conflicts as well. I know a lot of users will be very happy with this if you decide to pick it back up!

dnestico:
Should've just merged it this whole time. Welp.

Issue #1529, opened 2021-04-27
Widget for sensor history time series


edgimar:
It does not appear to be possible to view historical data via a widget.  It would be nice to be able to get a visualization similar to what you see when pressing on a sensor entity in the Android app (a time series plot over some period of time up until the current time).

edgimar:
There still isn't any such widget for the Android Home Assistant app.

sunnyd24:
I really would like to see this android widget which will show entity history over a customisable specific time frame, say 1 day or 2 hours.

ctml91:
Also looking forward to this. It would be especially helpful if you could group entities using the same unit measurement on the same graph, like on the history page directly in HomeAssistant.

ivorsmorenburg:
![Screenshot 2024-09-01 105938](https://github.com/user-attachments/assets/14082a8b-bd5d-4c2c-bbc4-1e6c8a5de86c)
![803F70C8-55DF-40D0-B052-29B44E8D80A0](https://github.com/user-attachments/assets/ee683d08-9de1-4b18-9bb0-ffeb0577d605)

NunoMCMoreira:
Somene to help me intalling this widget ?

jpelgrom:
> Somene to help me intalling this widget ?

This is an open feature request, not something that's currently available in the app.

NunoMCMoreira:
> > Somene to help me intalling this widget ?
> 
> This is an open feature request, not something that's currently available in the app.

Oh, that's even better if this will became available on next compassion app update. Thanks

krauthaus:
> This is how it looks, waiting for some comments, probably we could improve the whole Widget side of the app. ![image](https://github.com/user-attachments/assets/4d89ff89-3989-4e61-9bcd-5c11559e86fd)

That is pretty much the one killer feature I am missing in the app! (Next to buttons who view on/off states.)

dgudim:
Was the idea abandoned of is it still in the works? Because having a graph widget would be amazing

whcrg:
Also want this! Unfortunately no suitable skills to contribute meaningfully

TimoPtr:
Anyone can pick this up and continue the word done.

Mannshoch:
I would also like it to see this. If possible I would like to have it also possible on a small  1x2 or 2x1 size. 
in the widget settings I would like to set somthing like the last 5h on x. and have a black background with a white transparent overlay text for axis lable.

PR #4601, opened 2024-09-01
Graph widget to show historical data

ivorsmorenburg:
## Summary
Adds Graph Widget to make possible see historical data

## Screenshots
https://github.com/home-assistant/android/issues/1529#issuecomment-2323396196
 
## Any other notes
https://github.com/home-assistant/android/issues/1529


bgoncal:
Hi Ivor, can you add a small video demo the functionality?

common/src/main/java/io/homeassistant/companion/android/database/sensor/SensorDao.kt
```diff
@@ -14,10 +14,10 @@ import kotlinx.coroutines.flow.Flow
 @Dao
 interface SensorDao {
 
-    @Query("SELECT * FROM Sensors WHERE id = :id")
+    @Query("SELECT * FROM sensors WHERE id = :id")
```
bgoncal, line 17: Was this change intentional?

app/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="historical_graph">Historical Graph</string>
+    <string name="widget_graph_we_are_retrieving_data">We are retrieving some data, please came back later</string>
```
bgoncal, line 4: Is this something that take several minutes? Otherwise we can change the text to something that does not give the impression it will take long.

bgoncal (changes requested):
Any tests we could add for it?

common/src/main/java/io/homeassistant/companion/android/database/sensor/SensorDao.kt
ivorsmorenburg: Yes to keep everything lowercase consistent as I saw other where rising the right name of the table lowercase

app/src/main/res/values/strings.xml
ivorsmorenburg: I'll look into this usually depends on how the sensor get refreshed. For example if is electric sensor the W will change very often but it is a for example temperature of a room it might not change that often.

But I can change it for something more generic

ivorsmorenburg:
> Any tests we could add for it?

Yes we could but I didn't see that as an requirement for the 4 hours test. I want to optimise this piece of code before it gets merged.

ivorsmorenburg:
> Hi Ivor, can you add a small video demo the functionality?

This is the video example of an MVP for the GraphWidget

https://github.com/user-attachments/assets/91439c27-5f59-46f4-b635-a9dd50177bb5

app/src/main/java/io/homeassistant/companion/android/settings/widgets/ManageWidgetsViewModel.kt
```diff
@@ -1,72 +1,75 @@
-package io.homeassistant.companion.android.settings.widgets
```
dshokouhi, line 1: This file has become hard to review as it looks like the entire file was changed but that does not seem to be the case. Can you update this so the diff shows the actual changes done to the file please?

app/src/main/java/io/homeassistant/companion/android/settings/widgets/views/ManageWidgetsView.kt
```diff
@@ -110,7 +114,7 @@ fun ManageWidgetsView(
         ) {
             if (viewModel.buttonWidgetList.value.isEmpty() && viewModel.staticWidgetList.value.isEmpty() &&
                 viewModel.mediaWidgetList.value.isEmpty() && viewModel.templateWidgetList.value.isEmpty() &&
-                viewModel.cameraWidgetList.value.isEmpty()
+                viewModel.templateWidgetList.value.isEmpty() && viewModel.graphWidgetList.value.isEmpty()
```
dshokouhi, line 117: Looks like you may have mixed up `cameraWidgetList` with `templateWidgetList` here :)

app/src/main/java/io/homeassistant/companion/android/settings/widgets/ManageWidgetsViewModel.kt
ivorsmorenburg: Yes I agree with you, what I have done is runned KtLint on the project and it change the format but not the content I'll have a look into that 🤓

jpelgrom: (this is usually caused by changing the line endings on the file)

ivorsmorenburg: That's correct, seems that the KtLint decided to add an ending line. I'll try to revert it

dshokouhi:
@ivorsmorenburg can you please add some screenshots of the configuration page so we can see what options the user needs to use? For example I can see the code contains stuff related to entity attributes but I am not sure if those are even used in the graph?

ivorsmorenburg:
> @ivorsmorenburg can you please add some screenshots of the configuration page so we can see what options the user needs to use? For example I can see the code contains stuff related to entity attributes but I am not sure if those are even used in the graph?

Yes you are correct, right now is just an **MVP**, the UI Config Screen is the Entity UI Screen.

I'll work on it no worries : ) I have big plans for the UI

dshokouhi:
ok thanks for confirming :) sometimes when a PR is in draft its tough to tell how much to look and comment 🙈  I will go ahead and look at the generated APK for now until you feel the code is ready to look at :)

ivorsmorenburg:
> Any tests we could add for it?

I have been trying to add some tests.. but the project structure is not making easy task... and the effort is too big for just adding the first test to the project without a structure already in place.

I believe this is also a bigger feature, as we will need to have Repositories for example, to be able to mock and test logic, unit test suppose to test the logic not the Android Views.

ivorsmorenburg:
I made few improvements this is my 'final' version, comments are welcome @dshokouhi  :)
https://github.com/user-attachments/assets/9dc0d826-3343-4c0e-826e-27e02077ca30

dshokouhi:
@ivorsmorenburg i am encountering a few issues in teh debug APK.

the widget example on my device is not correct

image

the debug APK is missing the proper title , maybe caused by strings not all being in common?

image

also the widget never updates for me seems to be stuck here for me

image

also when I edit the widget the changes are not saved

for testing I am using the battery level sensor with either 24 hours, 5 hours or even 1 hour time sampling. Seems to always be stuck loading.

ivorsmorenburg:
> the widget example on my device is not correct
- That is very strange as it uses an image like all the others widgets :S, but I think the actual issue might be that on your device it isn't using the png image, but an emulated version of the graph. I was using Android 13 OnePlus 8 Google Stock ROM. Could you provide me with your device specs? so that I can emulate it and test the weird behaviour on the widget.

> the debug APK is missing the proper title , maybe caused by strings not all being in common?
- Yes, that's on me, as I have added a debug library to test and inspect the database, one of the side effects is that it changes the name of the app for debug. We could remove the library and only use it when local builds to prevent this.

> also the widget never updates for me seems to be stuck here for me
- Ok I feel that is a normal behaviour as it needs at least 2 values to show a graph, probably due to the slow discharging of the battery the second value might take a while to appear. 

> also when I edit the widget the changes are not saved
- I'll test this, but I have only fixed updating the widget for the graph as other widgets doesn't seem to work

> for testing I am using the battery level sensor with either 24 hours, 5 hours or even 1 hour time sampling. Seems to always be stuck loading.
- About sampling that has been removed, and it uses a time range to delete old entries and only keep the ones within that range. so for 1 hour you will see only historic as old as 1 hour. When it gets "stuck" is because doesn't have enough data.. the sensor might need to send more at least 2 events to show the graph, if you like you can test it using the dev tools in home assistant to fire an event for that sensor.

ivorsmorenburg:
In resume, I'll try to have a look tonight or tomorrow probably evening into this :) @dshokouhi Thanks for the review! Really appreciate the testing efforts!

dshokouhi:
> * I was using Android 13 OnePlus 8 Google Stock ROM. Could you provide me with your device specs? so that I can emulate it and test the weird behaviour on the widget.

Pixel 9 Pro XL



> * Yes, that's on me, as I have added a debug library to test and inspect the database, one of the side effects is that it changes the name of the app for debug. We could remove the library and only use it when local builds to prevent this.

Is htat library necessary for the scope of this PR? If not we should probably remove it and add it in another PR to keep this one easier to review.

 

> * Ok I feel that is a normal behaviour as it needs at least 2 values to show a graph, probably due to the slow discharging of the battery the second value might take a while to appear.

is the graph waiting for live data or does it pull up past data just like HA graphs do? We may need to consider user expectations here as I thought this graph would match the frontend graph personally.




> * I'll test this, but I have only fixed updating the widget for the graph as other widgets doesn't seem to work

last I checked all of our widgets update just fine, maybe the debug library is causing an issue here?

dshokouhi:
Thanks your latest changes have addressed my previous comments.

image

 It looks like the config screen is not showing the correct saved value for the amount of time to display. The slider seems to default to 24h when I come back to edit, even though the graph updates with the correct value.

Couple more comments on the design: 

- we should restrict which domains qualify for this feature, things like `automation` domain may not qualify
- will there be theming added to match the other widgets?
- should we inform the user about using graphs that dont update quickly? the slow loading may lead to unwanted bugs I feel user may expect data to be live just like their dashboards and other widgets

jpelgrom:
There's a lot in here so don't consider this a full review, but just jumping in the current discussion to add my thoughts:

> Ok I feel that is a normal behaviour as it needs at least 2 values to show a graph, probably due to the slow discharging of the battery the second value might take a while to appear.

I'm not sure this is a good approach when the history API and long term statistics exist. Any reason those weren't used?

Now your device needs to download all the data, and time scale depends on the data received. If the device is not receiving updates, the widget doesn't update with the values while offline and the points are not set along the x axis correctly.

> Removed preview layout as we are using preview image

I don't know if this is the best experience. Android 12+ can render the preview live, better matching the real widget including device theming. A preview image exists for older versions where this isn't supported. You just need to make sure the default values in the widget xml make sense.

ivorsmorenburg:
@jpelgrom @dshokouhi 

At the beginning I tried to keep the idea under 4 hours coding to solve the linked issue, to enable graph feature with minimum changes, but I got carried away haha. If you could please point me in the right directions using history API and the structure you like, I'm happy to refactor and make some optimisations. 

Due to my limited time this might take a bit longer and Ill close this PR and make another PR in the future, but with a plan we could make something shine for sure.

we should restrict which domains qualify for this feature, things like automation domain may not qualify
- yes I'll add a filter for it, that was on a bigger scope, so I will add it

will there be theming added to match the other widgets?
- At the moment I didn't have that in the scope for this widget as I felt that we should refactor everything a bit and can be done in a different pr

should we inform the user about using graphs that dont update quickly? the slow loading may lead to unwanted bugs I feel user may expect data to be live just like their dashboards and other widgets
- for this I think the best way will be to use history API. The reason for not doing those is it want thought for a widget and I want it to work with the events already available.

ivorsmorenburg:
I have fixed few things, thanks to @PizzaMarinara we can support multiple entities in the graph in the near future. Tomorrow ill finish the implementation with historical data fetched from the API. I feel is starting to look pretty cool.
Something I think it needs a bigger thought is the layout preview rather than image preview, but because we render and paint the Graph seems something can be done in the future, @jpelgrom. @dshokouhi  I hope it works now on your Pixel I have tested on the emulator but I don't trust them haha

jpelgrom:
This PR seems to be endlessly growing (there are now _two new widgets_ in it). I know this is still marked as draft, but please try to not increase the scope all the time, and keep the PR description updated. Doing that will make it easier for us to review. If accepted and merged your changes will immediately be included in app releases, so prefer contained, tested and polished changes to something big which isn't ready for prime-time. 

([point 1, 2 and 6 here](https://developers.home-assistant.io/docs/review-process#creating-the-perfect-pr))

jpelgrom:
> This PR seems to be endlessly growing (there are now _two new widgets_ in it). I know this is still marked as draft, but you are requesting reviews so please try to not increase the scope all the time, and keep the PR description updated. Doing that will make it easier for us to review. If accepted and merged your changes will immediately be included in app releases, so prefer contained, tested and polished changes to something big which isn't ready for prime-time. 
> 
> ([point 1, 2 and 6 here](https://developers.home-assistant.io/docs/review-process#creating-the-perfect-pr))

ivorsmorenburg:
I closed the PR today because the request features don't match the scope of a graph widget. I'll try to make s cleaner smaller one. But I won't be able to include historical data from API as it will be out of scope, Preview layout updating (fixing) the "update widget", some unit testing and others. I feel for this requirements is something that should be already in the project so I'll have to create few different PRs before implementing the ideal changes on my widget PR and keep it small and clear which was the goal at the begining. 😊

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidget.kt
```diff
@@ -0,0 +1,484 @@
...
+
+    companion object {
+
+        private const val TAG = "GraphWidget"
+        internal const val TOGGLE_ENTITY =
+            "io.homeassistant.companion.android.widgets.entity.GraphWidget.TOGGLE_ENTITY"
+
+        internal const val EXTRA_SERVER_ID = "EXTRA_SERVER_ID"
+        internal const val EXTRA_ENTITY_ID = "EXTRA_ENTITY_ID"
+        internal const val EXTRA_ATTRIBUTE_IDS = "EXTRA_ATTRIBUTE_IDS"
```
dshokouhi, line 59: As attributes are not used we should remove all references to them

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidget.kt
```diff
@@ -0,0 +1,484 @@
...
+        internal const val EXTRA_LABEL = "EXTRA_LABEL"
+        internal const val EXTRA_ATTRIBUTE_SEPARATOR = "EXTRA_ATTRIBUTE_SEPARATOR"
+        internal const val EXTRA_TAP_ACTION = "EXTRA_TAP_ACTION"
```
dshokouhi, line 62: As tap action is not used we should remove references to it

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidget.kt
```diff
@@ -0,0 +1,484 @@
...
+        internal const val EXTRA_BACKGROUND_TYPE = "EXTRA_BACKGROUND_TYPE"
```
dshokouhi, line 63: As theming is not used yet please remove this and references to it

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidget.kt
```diff
@@ -0,0 +1,484 @@
...
+        val attributeSeparatorSelection: String? = extras.getString(EXTRA_ATTRIBUTE_SEPARATOR)
+        val tapActionSelection = BundleCompat.getSerializable(extras, EXTRA_TAP_ACTION, WidgetTapAction::class.java)
+            ?: WidgetTapAction.REFRESH
+        val backgroundTypeSelection = BundleCompat.getSerializable(extras, EXTRA_BACKGROUND_TYPE, WidgetBackgroundType::class.java)
+            ?: WidgetBackgroundType.DAYNIGHT
+        val textColorSelection: String? = extras.getString(EXTRA_TEXT_COLOR)
+        val timeRange = extras.getInt(EXTRA_TIME_RANGE)
+
+        if (entitySelection == null) {
+            Log.e(TAG, "Did not receive complete service call data")
```
dshokouhi, line 378: This log should be updated for the proper widget

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidget.kt
```diff
@@ -0,0 +1,484 @@
...
+                )
+            }
+
+            // Get views and update widget
+            val views = getWidgetRemoteViews(context, appWidgetId, entity as Entity<Map<String, Any>>)
+            AppWidgetManager.getInstance(context).updateAppWidget(appWidgetId, views)
+        }
+    }
+
+    private fun toggleEntity(context: Context, appWidgetId: Int) {
```
dshokouhi, line 439: I dont think this applies to entities in the graph widget

wear/build.gradle.kts
```diff
@@ -67,6 +67,7 @@ android {
 
 dependencies {
     implementation(project(":common"))
+    implementation(libs.mpgraph)
```
dshokouhi, line 70: why do we need to include the module in Wear OS if its not used there?

app/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dshokouhi, line 1: these strings should be merged with the ones in `common` we dont use strings in `main`

dshokouhi (changes requested):
Looks like there is still a bit of leftover code that should be removed. Anything the user cannot configure we should remove to help cut down on the scope. 

Also I am not sure if the current implementation makes sense, the graph can take a very long time to generate and that won't lead to a good user experience. I think the history APIs may be required here as part of MVP otherwise it may be minutes or longer without a graph showing.

image

wear/build.gradle.kts
ivorsmorenburg: Otherwise it wont compile

ivorsmorenburg:
https://github.com/user-attachments/assets/92d4baa8-4a20-479c-b580-7e002217ee50

dshokouhi (changes requested):
Ok things are coming along well. The graph now loads very fast and the experience is much better. Few things that came up during testing. What value should be on teh X axis? Currently it does not make sense and does not seem to represent time

image

also the graphs dont necessarily match what the frontend shows in the same 24 hour period for the same entity

image

I also noticed that when I edit a widget and change teh entity the graph gets mangled

In the first graph screenshot I changed the entity from phone to watch battery and it looks like the values are mixed?

image

if i create a new widget just for the watch it loads properly

image

common/src/main/res/values/strings.xml
```diff
@@ -1274,4 +1275,10 @@
     <string name="state_update_available">Update available</string>
     <string name="state_up_to_date">Up-to-date</string>
     <string name="sensor_name_health_connect">Health Connect sensors</string>
+    <string name="graph_state_widgets">Graph historic Widget</string>
```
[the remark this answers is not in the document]
jpelgrom: Could we perhaps remove widget entirely and make it more like the other widgets? "History graph" sounds nice, but may be confusing because of the frontend card of the same name, or "Sensor graph".

(To me as a non-native speaker, "graph historic widget" doesn't sound right.)

automotive/build.gradle.kts
```diff
@@ -147,6 +147,7 @@ android {
 
 dependencies {
     implementation(project(":common"))
+    implementation(libs.mpgraph)
```
jpelgrom, line 150: Not sure there is a way to use widgets on Automotive, but if you're declaring the dependency then we should also add it to manifest.

common/src/main/java/io/homeassistant/companion/android/database/widget/graph/GraphWidgetEntity.kt
```diff
@@ -0,0 +1,31 @@
...
+    val entityId: String,
+    @ColumnInfo(name = "label")
+    val label: String?,
+    @ColumnInfo(name = "graph_time_range")
+    val timeRange: Int = 24,
+    @ColumnInfo(name = "tap_action", defaultValue = "REFRESH")
+    val tapAction: WidgetTapAction,
+    @ColumnInfo(name = "last_update")
+    val lastUpdate: Long,
+    @ColumnInfo(name = "background_type", defaultValue = "DAYNIGHT")
```
jpelgrom, line 27: This should not include a `defaultValue`, as it is always set by the widget. It exists on other widget data classes for data migration purposes.

app/src/main/java/io/homeassistant/companion/android/settings/widgets/ManageWidgetsViewModel.kt
ivorsmorenburg: Yes that's because the ending line. When I added the graphwidget I had to apply KTLint and that triggers the format of the whole time due not having the ending line.

app/src/main/res/drawable-night/widget_example_graph.png
jpelgrom, line 1: Why are day/night previews different in content, but also size in pixels? For night, use a tool like tinypng.com to reduce the binary size to something more reasonable.

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidget.kt
```diff
@@ -0,0 +1,388 @@
...
+        }
+
+        return ResolvedText(entity?.friendlyName, entityCaughtException)
+    }
+
+    private suspend fun fetchHistory(appWidgetId: Int, serverId: Int, entityId: String, fromMillis: Long, toMillis: Long) {
+        val entitiesStatesList: List<List<Entity<Map<String, Any>>>>?
+        try {
+            entitiesStatesList = serverManager.integrationRepository(serverId)
+                .getHistory(
```
jpelgrom, line 282: Unless I'm missing something, this is the only place you're getting history from the API. This function (`fetchHistory`) is only called when the widget is added. That means that the issue we previously pointed out (the device being responsible for getting the history - device not connected means no history updates) still exists.

Shouldn't this be called any time after the widget has lost connection?

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidget.kt
```diff
@@ -0,0 +1,388 @@
...
+                    listOf(entityId),
+                    fromMillis,
+                    toMillis
+                )
```
jpelgrom, line 286: Any particular reason for not using `minimalResponse` in the request here? That should significantly reduce the data needed, and you're not using any of the other data.

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidget.kt
```diff
@@ -0,0 +1,388 @@
...
+            }?.forEachIndexed { _, historyEntity ->
+                repository.insertGraphWidgetHistory(
+                    GraphWidgetHistoryEntity(
+                        entityId = historyEntity.entityId,
+                        graphWidgetId = appWidgetId,
+                        state = historyEntity.state,
+                        lastChanged = historyEntity.lastChanged.timeInMillis
+                    )
+                )
+            }
```
jpelgrom, line 299: Can we put this in a [database transaction](https://developer.android.com/reference/androidx/room/Transaction) to improve performance?

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidgetConfigureActivity.kt
```diff
@@ -0,0 +1,308 @@
...
+            } else {
+                sliderTimeRange.toInt()
+            }
+
+            intent.putExtra(GraphWidget.EXTRA_TIME_RANGE, graphTimeRange)
+
+            context.sendBroadcast(intent)
+
+            repository.updateWidgetLastLabel(appWidgetId, binding.label.text.toString())
+            repository.updateWidgetTimeRange(appWidgetId, binding.timeRange.value.toInt())
```
jpelgrom, line 285: Why save these here? They are already saved by the widget receiving the broadcast.

common/src/main/res/values/strings.xml
```diff
@@ -1274,4 +1275,10 @@
     <string name="state_update_available">Update available</string>
     <string name="state_up_to_date">Up-to-date</string>
     <string name="sensor_name_health_connect">Health Connect sensors</string>
+    <string name="graph_state_widgets">Graph historic widget</string>
+    <string name="widget_graph_we_are_retrieving_data">We are collecting some data…</string>
+    <string name="graph_widget_desc">To see historical data on graph</string>
+    <string name="interval_sampling">Interval sampling:</string>
+    <string name="historical_graph">Historical Graph</string>
+    <string formatted="true" name="graph_time_range">Graph Range (%1$d hours)</string>
```
jpelgrom, line 1283: We follow [Material guidelines and use sentence case](https://m3.material.io/foundations/content-design/style-guide/ux-writing-best-practices#fc5c2a78-f4bf-4d42-bdac-42ff80391129) so the second words Graph/Range should be graph/range

common/src/main/java/io/homeassistant/companion/android/database/widget/graph/GraphWidgetEntity.kt
ivorsmorenburg: I had to override it , so I want it to keep it consistant

app/src/main/res/layout/widget_graph.xml
```diff
@@ -0,0 +1,83 @@
...
+        <ProgressBar
+            android:layout_width="48dp"
+            android:layout_height="48dp"
+            android:indeterminate="true"
+            android:indeterminateTint="@color/colorAccent" />
+
+        <TextView
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:text="@string/widget_graph_we_are_retrieving_data" />
```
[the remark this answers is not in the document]
ivorsmorenburg: We could remove the text potentially, but sometimes API can take few seconds to come back depending how much data you have on the sensor and the hardware used. wdyt? @jpelgrom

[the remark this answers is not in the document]
jpelgrom: For a few seconds a loading indicator only seems good enough to me, that seems like a pretty obvious indicator and the frontend also doesn't show text while loading graphs.

(you don't need to mention me when replying directly)

common/src/main/java/io/homeassistant/companion/android/database/widget/graph/GraphWidgetEntity.kt
jpelgrom: It is an override so I'd say that explicitly means you're allowed to change it 😅 My suggestion was only for the annotation (which is used by Room to set the value on the table, which isn't needed here!), not for the Kotlin variable.

app/src/main/java/io/homeassistant/companion/android/settings/widgets/ManageWidgetsViewModel.kt
jpelgrom: No, it is because you're changing between CRLF and LF. You should be able to change it back / manually add a line at the end without changing everything else.

app/src/main/res/layout/widget_graph.xml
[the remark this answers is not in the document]
ivorsmorenburg: I know @jpelgrom  :) just want it to ay your "name". Yes I will remove the text then :)

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidget.kt
ivorsmorenburg: We could :)

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidget.kt
ivorsmorenburg: Hello! @jpelgrom I have been testing this behaviour and, I could not reproduce your scenario by having the widget almost 24h. When I unlock and check the widget it updates, maybe not as often but the idea is also to not hammer by downloading everytime soo much data.

app/src/main/java/io/homeassistant/companion/android/settings/widgets/ManageWidgetsViewModel.kt
ivorsmorenburg: Nope doesn't seem to work

common/src/main/res/values/strings.xml
```diff
@@ -1274,4 +1275,7 @@
     <string name="state_update_available">Update available</string>
     <string name="state_up_to_date">Up-to-date</string>
     <string name="sensor_name_health_connect">Health Connect sensors</string>
+    <string name="graph_state_widgets">Statistics widget</string>
```
dshokouhi, line 1278: any reason why we have 2 different names for the widget?

dshokouhi (changes requested):
Thing are coming along well! The updated graph looks well and the axis now shows time as expected. Nice touch to add the steps like the HA graphs!

image

Looks like there are still a few open comments and also there are now 2 different names for the widget? 🙈

common/src/main/res/values/strings.xml
ivorsmorenburg: Just tryed to keep consistent from Home Assistant. The graph look pretty much like it and is called statistics.

dshokouhi: If you want to change the name to make it more correct thats fine but we should keep it consistent. For example the manage widget screen uses both labels now.

image

ivorsmorenburg: You are right my friend! I'll then refactor all "graph" -> "stats"

app/src/main/java/io/homeassistant/companion/android/settings/widgets/ManageWidgetsViewModel.kt
ivorsmorenburg: I found that it was not the ending line or the CRLF.. I try that multiple times.. seems somehow I change the permissions of the file. I was tracking permissions with git, so I have revert it and stop tracking this changes. Please have another look

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidget.kt
ivorsmorenburg: Do you think is a good idea to add a 'cron' ?

common/src/main/res/values/strings.xml
```diff
@@ -1025,10 +1025,11 @@
     <string name="widget_spinner_server">Server:</string>
     <string name="widget_state_separator_label">State and attribute separator:</string>
     <string name="widget_static_image_description">Entity state</string>
+    <string name="widget_graph_image_description">Statistics widget</string>
     <string name="widget_tap_action_label">Tap action:</string>
     <string name="widget_tap_action_toggle">Toggle</string>
     <string name="widget_template_error">Unable to render the template</string>
-    <string name="widget_text_hint_label">Label</string>
+    <string name="widget_text_hint_label">Entity name</string>
```
dshokouhi, line 1032: This change may make sense for widgets but its also used in the Assist shortcut where this won't make sense

https://github.com/home-assistant/android/blob/15b1518a4466e9c58d26efce7e4605858b88a689/app/src/main/java/io/homeassistant/companion/android/widgets/assist/AssistShortcutView.kt#L64

common/src/main/res/values/strings.xml
```diff
@@ -1025,10 +1025,11 @@
     <string name="widget_spinner_server">Server:</string>
     <string name="widget_state_separator_label">State and attribute separator:</string>
     <string name="widget_static_image_description">Entity state</string>
+    <string name="widget_graph_image_description">Statistics widget</string>
```
dshokouhi, line 1028: This string is now duplicated in line 1281 :)

app/src/main/java/io/homeassistant/companion/android/widgets/graph/GraphWidget.kt
jpelgrom: That doesn't seem necessary for a widget. The history API allows you to fetch any missing data, why not use that to get data from the last update (or start of time range, if later) until now on screen on? If there is nothing, the data usage is minimal.

common/src/main/java/io/homeassistant/companion/android/database/widget/graph/GraphWidgetEntity.kt
```diff
@@ -0,0 +1,35 @@
...
+    val unitOfMeasurement: String,
+    @ColumnInfo(name = "graph_time_range")
+    val timeRange: Int = 24,
+    @ColumnInfo(name = "tap_action", defaultValue = "REFRESH")
+    val tapAction: WidgetTapAction,
+    @ColumnInfo(name = "last_update")
+    val lastUpdate: Long,
+    @ColumnInfo(name = "significantChangesOnly")
+    val significantChangesOnly: Boolean,
+    @ColumnInfo(name = "smoothGraphs")
```
jpelgrom, line 31: For consistency with other column names, please use lowercase + underscores instead of camelcase

common/src/main/java/io/homeassistant/companion/android/database/widget/graph/GraphWidgetEntity.kt
```diff
@@ -0,0 +1,35 @@
...
+    @ColumnInfo(name = "entity_id")
+    val entityId: String,
+    @ColumnInfo(name = "label")
+    val label: String?,
+    @ColumnInfo(name = "unit_of_measurement")
+    val unitOfMeasurement: String,
+    @ColumnInfo(name = "graph_time_range")
+    val timeRange: Int = 24,
+    @ColumnInfo(name = "tap_action", defaultValue = "REFRESH")
+    val tapAction: WidgetTapAction,
```
jpelgrom, line 26: Repeating https://github.com/home-assistant/android/pull/4601#discussion_r1749332313: unused so this should be removed for now

dshokouhi:
@ivorsmorenburg this PR has a lot going on with just the logic for Graph Widget alone and I dont think a lot of that we can escape. There are other things done in this PR that are touching other widgets thats making this a bit difficult to not only review but also test and make sure the other widgets are still fine. Can you please split up the refactoring of the other widgets so we can review it separately and make the review of this PR easier? That should help cut down on the amount of files touched and make this PR easier to review. We can review that PR before this one as I think you may have some dependencies on it.

NunoMCMoreira:
This is exactly what I was looking for more that one year. 

I do have a portfolio tracker sensor on my home assistant and I would like to have it represented by a widget. 

is there anyone to help me with this ? I am not an expert on coding so all this looks chinese to me :)

Issue #3338, opened 2023-02-15
Crash: Websocket Illegal State Exception


dshokouhi:
**Home Assistant Android app version(s):**

First appeared beta-2762-93844ffe-full, still impacting 2023.1.1

**Android version(s):**

All supported versions 5-13

**Device model(s):**

Numerous - almost 7.5k users impacted

**Home Assistant version:**

unknown

**Last working Home Assistant release (if known):**

unknown

**Description of problem, include YAML if issue is related to notifications:**

I am unsure what causes this issue but it is second from the top in our crash list

**Companion App Logs:**

Log from 2023.1.1
```
Exception java.lang.IllegalStateException:
  at kotlinx.coroutines.CancellableContinuationImpl.alreadyResumedError (CancellableContinuationImpl.kt:482)
  at kotlinx.coroutines.CancellableContinuationImpl.resumeImpl (CancellableContinuationImpl.kt:447)
  at kotlinx.coroutines.CancellableContinuationImpl.resumeImpl$default (CancellableContinuationImpl.kt:420)
  at kotlinx.coroutines.CancellableContinuationImpl.resumeWith (CancellableContinuationImpl.kt:328)
  at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.handleMessage (WebSocketRepositoryImpl.kt:510)
  at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl.access$handleMessage (WebSocketRepositoryImpl.kt:70)
  at io.homeassistant.companion.android.common.data.websocket.impl.WebSocketRepositoryImpl$onMessage$1$1.invokeSuspend (WebSocketRepositoryImpl.kt:653)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:33)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:106)
  at kotlinx.coroutines.internal.LimitedDispatcher.run (LimitedDispatcher.kt:42)
  at kotlinx.coroutines.scheduling.TaskImpl.run (Tasks.kt:95)
  at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely (CoroutineScheduler.kt:570)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask (CoroutineScheduler.kt:750)
[...]

PR #4616, opened 2024-09-08
Prevent socket response continuation being called multiple times (#3338)

ShaunPlummer:
## Summary
Prevent the crash outlined in #3338. Coroutines continuations can only be invoke once, see CancellableContinuation. Use atomic boolean to ensure the continuation is called only once in a thread safe way. While this does not fix the underlying cause it will prevent any more crashes. 

## Any other notes
We could log the stack trace whenever the continuation is invoked a second time in order to help identify where the second call comes from.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/WebSocketRequest.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/WebSocketRequest.kt
@@ -1,10 +1,10 @@
 package io.homeassistant.companion.android.common.data.websocket
 
 import io.homeassistant.companion.android.common.data.websocket.impl.entities.SocketResponse
+import java.util.concurrent.atomic.AtomicBoolean
 import kotlinx.coroutines.CancellableContinuation
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.SharedFlow
-import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * A class that holds information about messages that are currently active (sent and no response
--- a/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/WebSocketRequest.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/WebSocketRequest.kt
@@ -4,6 +4,7 @@ import io.homeassistant.companion.android.common.data.websocket.impl.entities.So
 import kotlinx.coroutines.CancellableContinuation
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.SharedFlow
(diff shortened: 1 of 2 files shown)


jpelgrom:
> We could log the stack trace whenever the continuation is invoked a second time in order to help identify where the second call comes from

Isn't that the stack trace we already have as a crash?

ShaunPlummer:
Yes, with these changes, it won't be logged out any more so it will be less obvious when the problem occurs. A better idea would probably be to log a warning when the second invocation is attempted.

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/WebSocketRequest.kt
```diff
@@ -24,4 +25,6 @@ data class WebSocketRequest(
     val eventTimeout: Long = 0L,
     val onEvent: Channel<Any>? = null,
     var onResponse: CancellableContinuation<SocketResponse>? = null
-)
+) {
+    val hasContinuationBeenInvoked = AtomicBoolean(onResponse == null)
```
jpelgrom, line 29: Setting this here seems to break the websocket because `onResponse` is null by default (as mentioned in the doc above the class definition), and as a result most of the native app functionality. It is only set [when the message is sent](https://github.com/home-assistant/android/blob/ece89d45c7177386fac34f6403b4d151c89a4cd5/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt#L652).

(Everything will time out waiting for a response which never arrives.)

jpelgrom (changes requested):

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/WebSocketRequest.kt
ShaunPlummer: I've set the initial value to false, which I believe resolves the issue. 

To test this I checked the list of available entities loads when adding a widget and that already added widgets load correctly. 

The property could be made immutable and the copy constructor used add the continuation. But since that would mean 2 objects are created for every message that's probably not the way to go.

jpelgrom: > I've set the initial value to false, which I believe resolves the issue.

Yes, looks like it. Feel free to mark the PR as ready to review so we can approve + merge.

> The property could be made immutable and the copy constructor used add the continuation. But since that would mean 2 objects are created for every message that's probably not the way to go.

That seems a bit wasteful and complicated when setting it to `false` by default also works.

jpelgrom (approved):
Thanks for the workaround!

Issue #2440, opened 2022-04-08
Add Orientation Sensor support


MrFr33man123:
**Describe the solution you'd like**
An extra Sensor that is made avaliable. 
It should output an orientation like face down, face up, horizontal, and so on like a dice. 

**Describe alternatives you've considered, if any**
Alternative i think one could give the raw degree angles for each axis. XYZ
But it would be a pain to use. 

**Additional context**
Example: You want to Automate something that is only Triggered if you wake up at night. if state was face up and charging and between 10pm and 5am(you where probably asleep) then turn on some light in nightlight mode. 

Or in the Car if connected to specific Bluetooth device and upright orientation (phone holder) + leaving home zone
=> send notification to partner phone with text "I'am leaving"

dshokouhi:
So looking more into this request. I think we can get by with reporting the rotation angle of the device that can translate to portrait/landscape based on the 4 data points (0, 90, 180, 270). The other part of this request should be doable already. The proximity sensor should be able to tell the user if something is blocking the sensor or not so on a flat surface that can determine if its laying face down or not. Might be a bit of a stretch, if that is not good enough here then yes we will be needing to create a sensor with raw values from the accelerometer so the user can translate them for their own use case.

tschaedi:
Any update on this?
I'm interested in this new sensor, too. This would me allow to make the Tasker App obsolete, which I use for updating the device orientation. My use case is similar to the above use case. I bring my phone(s) and the lighting in house to "sleep mode" if after 10pm and phone is charging and face down.

Hish15:
I would use this to give more power to NFC tags. Here is an example: 
To control a motorized blind, scanning the NFC tag with the phone upward make it go up, and downward make it go down.

dshokouhi:
Hey guys I have a submitted a PR adding 3 sensors containing the requested data. It would be great to get some end user testing on this to make sure it works as one would expect. In particular if anyone has a foldable to test on that would be great.

You can grab the debug APK from the artifact in https://github.com/home-assistant/android/actions/runs/10854606133

Extract the file and install the full or minimal debug APK. This will install side by side production, the app will have an easy identifiable red icon. I recommend using a unique device name to easily delete after the test.

The 3 sensors:

- Orientation of the device screen (should update immediately when the screen is on and orientation updates)
[...]

PR #4634, opened 2024-09-13
Add 2 different display sensors around various screen states

dshokouhi:
## Summary

Fixes: #2440 by adding 2 different sensors to capture the various states.

- Orientation sensor - Displays orientation typically portrait vs landscape
- Rotation sensor - Displays device rotation degree angle based on the natural orientation (i.e. a phone will be portrait 0 while a tablet will be landscape 0)

2 different sensors because the states can all be independent of each other.

Originally I tried to be fancy with reverse portrait states however I realized the angles are dependent on the device in question and theres too much for account for when you consider all the different types of screens like on a watch, car etc...

The intent that is added only fires while the screen is on and if the screen has to redraw teh UI so battery impact is very minimal  here.

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#1102

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -279,7 +279,7 @@ open class HomeAssistantApplication : Application() {
             this,
             sensorReceiver,
             IntentFilter(Intent.ACTION_CONFIGURATION_CHANGED),
-            ContextCompat.RECEIVER_NOT_EXPORTED
+            ContextCompat.RECEIVER_EXPORTED
         )
 
         // Update widgets when the screen turns on, updates are skipped if widgets were not added
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -274,6 +274,14 @@ open class HomeAssistantApplication : Application() {
             )
         }
 
+        // Register for changes to the configuration
+        ContextCompat.registerReceiver(
+            this,
+            sensorReceiver,
+            IntentFilter(Intent.ACTION_CONFIGURATION_CHANGED),
+            ContextCompat.RECEIVER_NOT_EXPORTED
+        )
+
         // Update widgets when the screen turns on, updates are skipped if widgets were not added
         val buttonWidget = ButtonWidget()
         val entityWidget = EntityWidget()
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
@@ -201,15 +201,19 @@ class DisplaySensorManager : SensorManager, SensorEventListener {
 
         val display = getRotationString(dm.getDisplay(Display.DEFAULT_DISPLAY).rotation)
 
-        val hasMultipleDisplays = dm.displays.size > 1
         val multiple = dm.displays.associate { it.name to getRotationString(it.rotation) }
-
+        val possibleStates = listOf("0", "90", "180", "270")
+        val attrs = if (dm.displays.size > 1) {
+            multiple.plus("options" to possibleStates)
+        } else {
+            mapOf("options" to possibleStates)
+        }
         onSensorUpdated(
             context,
             screenRotation,
             display,
             screenRotation.statelessIcon,
-            if (hasMultipleDisplays) multiple else mapOf()
+            attrs
         )
     }
 
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
@@ -61,13 +61,13 @@ class DisplaySensorManager : SensorManager, SensorEventListener {
             unitOfMeasurement = "°"
         )
 
-        val isFaceDown = SensorManager.BasicSensor(
-            "is_face_down",
-            "binary_sensor",
-            commonR.string.sensor_name_is_face_down,
-            commonR.string.sensor_description_is_face_down,
+        val isFaceDownOrUp = SensorManager.BasicSensor(
+            "is_face_down_or_up",
+            "sensor",
+            commonR.string.sensor_name_is_face_down_or_up,
+            commonR.string.sensor_description_is_face_down_or_up,
             "mdi:hand-pointing-down",
-            docsLink = "https://companion.home-assistant.io/docs/core/sensors#is-face-down-sensor"
+            docsLink = "https://companion.home-assistant.io/docs/core/sensors#is-face-down-or-up-sensor"
         )
     }
     override val name: Int
@@ -76,7 +76,7 @@ class DisplaySensorManager : SensorManager, SensorEventListener {
     override suspend fun getAvailableSensors(context: Context): List<SensorManager.BasicSensor> {
         hasAccelerometer = context.packageManager.hasSystemFeature(FEATURE_SENSOR_ACCELEROMETER)
         return if (hasAccelerometer) {
-            listOf(screenBrightness, screenOffTimeout, screenOrientation, screenRotation, isFaceDown)
+            listOf(screenBrightness, screenOffTimeout, screenOrientation, screenRotation, isFaceDownOrUp)
         } else {
             listOf(screenBrightness, screenOffTimeout, screenOrientation, screenRotation)
         }
@@ -102,7 +102,7 @@ class DisplaySensorManager : SensorManager, SensorEventListener {
         updateScreenOrientation(context)
         updateScreenRotation(context)
         if (hasAccelerometer) {
-            updateIsFaceDown(context)
+            updateIsFaceDownOrUp(context)
         }
     }
 
@@ -217,8 +217,8 @@ class DisplaySensorManager : SensorManager, SensorEventListener {
         )
     }
 
-    private fun updateIsFaceDown(context: Context) {
-        if (!isEnabled(context, isFaceDown)) {
+    private fun updateIsFaceDownOrUp(context: Context) {
+        if (!isEnabled(context, isFaceDownOrUp)) {
             return
         }
         val now = System.currentTimeMillis()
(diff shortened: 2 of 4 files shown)


common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
```diff
@@ -98,4 +149,104 @@ class DisplaySensorManager : SensorManager {
...
+            listenerLastRegistered = now.toInt()
+        }
+    }
+
+    override fun onSensorChanged(event: SensorEvent?) {
+        if (event?.sensor?.type == Sensor.TYPE_ACCELEROMETER) {
+            onSensorUpdated(
+                latestContext,
+                isFaceDown,
+                event.values[2] < -9,
```
jpelgrom, line 238: So I think this is from [here](https://developer.android.com/reference/android/hardware/SensorEvent#values) and inverted for the flat on the table example? I think it might be helpful to add a note here why this value is set 😅

common/src/main/res/values/strings.xml
```diff
@@ -1283,4 +1283,10 @@
     <string name="basic_sensor_name_sim_1_data_network_type">Data network type (SIM 1)</string>
     <string name="basic_sensor_name_sim_2_data_network_type">Data network type (SIM 2)</string>
     <string name="sensor_description_data_network_type">The radio technology (network type) currently in use on the device for data transmission</string>
+    <string name="sensor_name_screen_orientation">Screen Orientation</string>
```
jpelgrom, line 1286: Sensor names should use sentence case ;)

(also for the next two)

common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
```diff
@@ -26,23 +37,63 @@ class DisplaySensorManager : SensorManager {
...
+        )
+
+        val screenRotation = SensorManager.BasicSensor(
+            "screen_rotation",
+            "sensor",
+            commonR.string.sensor_name_screen_rotation,
+            commonR.string.sensor_description_screen_rotation,
+            "mdi:screen-rotation",
+            docsLink = "https://companion.home-assistant.io/docs/core/sensors#screen-rotation-sensor",
+            updateType = SensorManager.BasicSensor.UpdateType.INTENT
```
jpelgrom, line 58: This feels misleading to me, as it only updates when the rotation changes so much that the orientation changes. Slightly rotating the device like a compass or level doesn't result in sensor updates for every degree (which would be a bad idea, as it would be updating nonstop).

common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
```diff
@@ -98,4 +149,104 @@ class DisplaySensorManager : SensorManager {
...
+
+    private fun updateScreenRotation(context: Context) {
+        if (!isEnabled(context, screenRotation)) {
+            return
+        }
+
+        val wm = context.getSystemService(Context.WINDOW_SERVICE) as WindowManager
+
+        @Suppress("DEPRECATION")
+        val display = when (wm.defaultDisplay.rotation) {
```
jpelgrom, line 191: This is deprecated which makes me wonder: what is the default display on a foldable? When folded open is it landscape or portrait and if it is landscape would you expect to still see portrait because the cover display hasn't rotated?

jpelgrom (commented):
Haven't tested this yet, but how do the values hold up over time when they are updated in the background?

Is it a good idea to ask the users of the original request to test whether it solves their issue?

dshokouhi:
Placing in draft while working out some more specifics in the PR, this also needs testing on other devices as mentioned and also by end users who requested the sensor(s)

common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
dshokouhi: Latest commit will no longer use this deprecated method, instead the state of the sensor will always be of the default display. There will be an attribute for rotation angle of additional displays on those devices.

dshokouhi:
> Haven't tested this yet, but how do the values hold up over time when they are updated in the background?

in my extensive testing they work well but like other sensors require the app to run in the background to get timely updates

this still needs testing especially from a foldable device

dshokouhi:
Testing on a foldable works as expected! All states update as we expect. Marking as ready for review, have not heard back from end user but I dont think it should hold this up.

common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
```diff
@@ -98,4 +159,110 @@ class DisplaySensorManager : SensorManager {
...
+        }
+
+        @Suppress("DEPRECATION")
+        val orientation = when (context.resources.configuration.orientation) {
+            Configuration.ORIENTATION_PORTRAIT -> "portrait"
+            Configuration.ORIENTATION_LANDSCAPE -> "landscape"
+            Configuration.ORIENTATION_SQUARE -> "square"
+            Configuration.ORIENTATION_UNDEFINED -> STATE_UNKNOWN
+            else -> STATE_UNKNOWN
+        }
```
jpelgrom, line 175: I believe we should add these possible states as an `options` attribute to help with automation editing

common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
```diff
@@ -98,4 +159,110 @@ class DisplaySensorManager : SensorManager {
             mapOf()
         )
     }
+
+    private fun updateScreenOrientation(context: Context) {
+        if (!isEnabled(context, screenOrientation)) {
+            return
+        }
+
+        @Suppress("DEPRECATION")
```
jpelgrom, line 168: I hadn't looked at the docs yet, but now I did and it [literally says](https://developer.android.com/reference/kotlin/android/view/Display.html?hl=en#getorientation): "Deprecated: use getRotation" since forever (API 15). Is it really useful to have both then? Yes they are technically different but orientation seems mostly like a convenient way to know that 0 = portrait/landscape which the user will almost certainly already know.

common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
```diff
@@ -98,4 +159,110 @@ class DisplaySensorManager : SensorManager {
...
+        val orientation = when (context.resources.configuration.orientation) {
+            Configuration.ORIENTATION_PORTRAIT -> "portrait"
+            Configuration.ORIENTATION_LANDSCAPE -> "landscape"
+            Configuration.ORIENTATION_SQUARE -> "square"
+            Configuration.ORIENTATION_UNDEFINED -> STATE_UNKNOWN
+            else -> STATE_UNKNOWN
+        }
+
+        val icon = when (orientation) {
+            "portrait" -> "mdi:phone-rotate-portrait"
```
jpelgrom, line 178: This is the icon to rotate from landscape to portrait, not sure about the fit 🤔

jpelgrom (commented):
Positively surprised by how well the face down sensor is working (even if it did get stuck once)!

common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
dshokouhi: no this deprecation is for the square display type lol https://developer.android.com/reference/android/content/res/Configuration.html#ORIENTATION_SQUARE

I had removed teh rotation one already in a previous commit :)

the actual API used is not deprecated :)
https://developer.android.com/reference/android/content/res/Configuration.html#orientation

> Is it really useful to have both then?

its to help differentiate between reverse landscape/portrait which are not real states but still a different orientation, the original request wanted to know if the device is horizontal and this should help with that and additional states.

common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
dshokouhi: yes the icons were not easy to pick, originally I had `crop-portrait` but this looked more fun lol

common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
jpelgrom: Oops my bad for mixing it up, sorry.

> its to help differentiate between reverse landscape/portrait which are not real states but still a different orientation, the original request wanted to know if the device is horizontal and this should help with that and additional states.

So a convenience, but not for the original request :)

Horizontal is not a state we provide so you still have to determine if that is landscape/portrait/0/90/180/270 yourself. The only state you cannot determine from rotation is 'square' but the docs say that isn't used anymore on APIs the app supports.

common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
jpelgrom: When quickly looking at it the icons seem identical but I recognize how hard it is. I was thinking about `tablet` for landscape, `cellphone` for portrait for a while but the screen aspect ratios don't match unfortunately.

common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
dshokouhi: yea the angles are device dependent so they will vary from tablet to phone in this case so I figured best to leave it in their hands to determine what to do. The square one is weird but doesnt surprise me if a device like NS panel pro has it lol

common/src/main/java/io/homeassistant/companion/android/common/sensors/DisplaySensorManager.kt
dshokouhi: yea this one wasnt easy based on the current available options lol

dshokouhi: Maybe this could be good to leave as is in case another contributor has a better idea? Picking icons and naming things are usually the most thought driven part of PRs like this :)

dshokouhi:
added possible states to rotation sensor based on [user feedback](https://github.com/home-assistant/android/issues/2440#issuecomment-2358235615) to help make it more clear

dshokouhi:
I have updated the face down sensor logic ad name, I am honestly not sure what to name or even if it should belong under the display sensors. I originally had it under display sensors because the state is relative to the screen. I am not sure if "position" is an appropriate sensor name but I do see value in having various states like up or down to determine the display position. I opted to add attributes as well in case we find additional states to use. In that case the current name makes no sense. I am also ok with removing for now but the coolness factor of the data is holding me back 🤓

dshokouhi:
i am going to remove teh face up/down sensor for now as it needs more thought about how to best handle it and the different states that may be useful

app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
```diff
@@ -274,6 +274,14 @@ open class HomeAssistantApplication : Application() {
             )
         }
 
+        // Register for changes to the configuration
+        ContextCompat.registerReceiver(
+            this,
+            sensorReceiver,
+            IntentFilter(Intent.ACTION_CONFIGURATION_CHANGED),
+            ContextCompat.RECEIVER_NOT_EXPORTED
```
jpelgrom, line 282: Do you want to update this with the changes in #4661?

dshokouhi: hah good catch! yes indeed :)

jpelgrom (approved):

Issue #4315, opened 2024-04-02
Android Auto - Allow Automations to be triggered when selecting somewhere to navigate


theverex:
**Is your feature request related to a problem? Please describe.**
No - this is a new feature request.

**Describe the solution you'd like**
The Android Auto App is quite handy already with the ability to toggle entities and use favourites, however I notice that there is a navigation feature that works great to be able to navigate to certain zones.

It would be great if we could trigger an automation based on the zone / location that is selected to navigate to. At the moment, we can only trigger when you enter / leave a zone, not when you are navigating to a zone.

I would love to be able to jump in the car, use the HA Android Auto App, select Navigation -> Home and and have automations run when selecting that zone.

EG: Turn on lights, send vacuum to home base, open blinds etc.

Alternatively, when selecting 'Work' zone, turn on vacuum.

**Describe alternatives you've considered, if any**
The only current way to do the above is to use location based automations (enter / leave), however there might be times I don't want the automation to run when I enter / leave a zone, only when I navigate to that zone.

**Additional context**
No additional context to add.

Thanks!

dshokouhi:
we can probably send back an event when an item is clicked on from the car, not sure if that should only be in navigate entities or extended to actionable entities as well.

theverex:
@dshokouhi it would be great to have it for navigation entities, but all entities would also be powerful because we could then build automations based on if 

I assume that the 'Android Auto' entity / attribute could work here as well, but if you can understand where the entity was clicked from, that could open up all sort of great automations!

dshokouhi:
well the thing is that we dont do this behavior anywhere else and the expectation may be that if we implement this, things like widgets and device controls are very similar. I wonder if there is a better way to get the data you want. Have you tried using the Last Notification sensor to see if Google Maps gives you that information to go by? 

If that doesnt work you can also try adding an item to your favorites so you can execute the script or whatever you need before you start navigating. 

Some specifics need to be worked out here if we plan on adding this as we need to consider all other app features.

theverex:
@dshokouhi understood about all entities - if a change could be made just for the navigation entity only, that would work.

I tried the last notification sensor, but two main issues in this use case with navigation:

1. Battery usage is high for immediate notification syncing with HomeAssistant
2. Google Maps app doesn't give useful information in the notification itself - it will show the turn by turn text but it won't provide the data needed in the notification to understand the destination location.

DarkWolf7:
[...]

PR #4637, opened 2024-09-13
Fire an event when a navigation entity is selected while driving

dshokouhi:
## Summary

Fixes: #4315 by adding a new event `android.navigation_started` which only occurs for navigation entities in the driving interface.

which provides the following example event data:

```
event_type: android.navigation_started
data:
  entity_id: zone.test
  device_id: DEVICE_ID
origin: REMOTE
time_fired: "2024-09-13T23:10:24.304583+00:00"
context:
  id: ID
  parent_id: null
  user_id: USER_ID

```

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/vehicle/MapVehicleScreen.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/vehicle/MapVehicleScreen.kt
@@ -118,6 +118,14 @@ class MapVehicleScreen(
                         )
                         .setOnClickListener {
                             Log.i(TAG, "${pair.first.entityId} clicked")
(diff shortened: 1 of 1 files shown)


jpelgrom:
I'm still not sold on this because of the slippery slope of adding this to all clicks with >1 requester as it might be useful to someone, somehow - as discussed in the original issue. What made you change your mind?

dshokouhi:
Personally speaking I dont think we should add this to all other entities but i do see value in navigating to an entity which I dont think is a frequent action users will take.

jpelgrom:
This function will directly throw exceptions in case the call wasn't successful which should be caught :)

jpelgrom (approved):

PR #4640, opened 2024-09-15
First refactor for widgets

ivorsmorenburg:
## Summary

Refactor of Widgets to improve the code and behaviour, abstracts widget logic and added repositories to add later on unit testing making the code more testable
- Refactor BaseWdigetProvider
- Improved offline widget interaction and error handling
- Added repositories
- Fix wrong widget layout of static widget using button widget layout
- Recover from reboots and reinstalls using latest data and not showing "Problem Loading Widget"
- Camera improvements when offline will keep the latest image fetched
- Camera has a Icon to avoid seen an empty space on the screen for better UX until it gets the camera image
- Very Small XML UI optimisations
- Adds Missing CameraWidget to listen SCREEN Broadcast
## Screenshots

Not Required

https://github.com/home-assistant/android/pull/4601#issuecomment-2346755447

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -24,6 +24,7 @@ import io.homeassistant.companion.android.settings.language.LanguagesManager
 import io.homeassistant.companion.android.util.LifecycleHandler
 import io.homeassistant.companion.android.websocket.WebsocketBroadcastReceiver
 import io.homeassistant.companion.android.widgets.button.ButtonWidget
+import io.homeassistant.companion.android.widgets.camera.CameraWidget
 import io.homeassistant.companion.android.widgets.entity.EntityWidget
 import io.homeassistant.companion.android.widgets.mediaplayer.MediaPlayerControlsWidget
 import io.homeassistant.companion.android.widgets.template.TemplateWidget
@@ -277,6 +278,7 @@ open class HomeAssistantApplication : Application() {
         // Update widgets when the screen turns on, updates are skipped if widgets were not added
         val buttonWidget = ButtonWidget()
         val entityWidget = EntityWidget()
+        val cameraWidget = CameraWidget()
         val mediaPlayerWidget = MediaPlayerControlsWidget()
         val templateWidget = TemplateWidget()
 
@@ -286,6 +288,7 @@ open class HomeAssistantApplication : Application() {
 
         ContextCompat.registerReceiver(this, buttonWidget, screenIntentFilter, ContextCompat.RECEIVER_NOT_EXPORTED)
         ContextCompat.registerReceiver(this, entityWidget, screenIntentFilter, ContextCompat.RECEIVER_NOT_EXPORTED)
+        ContextCompat.registerReceiver(this, cameraWidget, screenIntentFilter, ContextCompat.RECEIVER_NOT_EXPORTED)
         ContextCompat.registerReceiver(this, mediaPlayerWidget, screenIntentFilter, ContextCompat.RECEIVER_NOT_EXPORTED)
         ContextCompat.registerReceiver(this, templateWidget, screenIntentFilter, ContextCompat.RECEIVER_NOT_EXPORTED)
     }
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/camera/CameraWidget.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/camera/CameraWidget.kt
@@ -2,14 +2,11 @@ package io.homeassistant.companion.android.widgets.camera
 
 import android.app.PendingIntent
 import android.appwidget.AppWidgetManager
-import android.appwidget.AppWidgetProvider
 import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
-import android.content.res.Resources
+import android.graphics.Bitmap
 import android.os.Bundle
-import android.os.Handler
-import android.os.Looper
 import android.util.Log
 import android.view.View
 import android.widget.RemoteViews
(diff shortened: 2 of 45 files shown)


jpelgrom:
Why are the widget repositories in `common`? There is absolutely no reason to have this around for watches. 

You seem to be moving quite a lot of code around without making any changes to it (most notably: xml view attributes), which makes it harder to follow where there are 'real' changes. Is this really necessary, and if so could you at least split it up in a separate commit or something?

> other loading improvements

Can you specify these 'improvements'? It is a refactor so you should try to limit functional changes.

ivorsmorenburg:
Repositories can be move to a new location

dshokouhi:
> * Added Frame to Camera Widget for better looking

if this is changing the design can you please provide a screenshot of the new look?

ivorsmorenburg:
> > * Added Frame to Camera Widget for better looking
> 
> if this is changing the design can you please provide a screenshot of the new look?
camera widget placeholder

dshokouhi:
> > > * Added Frame to Camera Widget for better looking
> > 
> > 
> > if this is changing the design can you please provide a screenshot of the new look?
> > Also imaging the background in white for light theme

the transparent background was added a while ago for these widgets, I personally feel that it should become an option if we are going to add the background back. Tough to tell if the background is there when an actual image loads based on the provided picture. If we are going to do that we should really move it out of this PR. 

This PR is starting to get a bit larger like the other one and that makes it more difficult to review. We appreciate the desire to fix everything but doing that in one PR is not easy to review. Smaller changes get reviewed easier and faster.

ivorsmorenburg:
> > > > * Added Frame to Camera Widget for better looking
> > > 
> > > 
> > > if this is changing the design can you please provide a screenshot of the new look?
> > > Also imaging the background in white for light theme
> 
> the transparent background was added a while ago for these widgets, I personally feel that it should become an option if we are going to add the background back. Tough to tell if the background is there when an actual image loads based on the provided picture. If we are going to do that we should really move it out of this PR.
> 
> This PR is starting to get a bit larger like the other one and that makes it more difficult to review. We appreciate the desire to fix everything but doing that in one PR is not easy to review. Smaller changes get reviewed easier and faster.

Just updated the description message as you referenced in that issue, the background frame for the camera has been removed.

app/src/main/java/io/homeassistant/companion/android/widgets/camera/CameraWidget.kt
```diff
@@ -129,18 +67,18 @@ class CameraWidget : AppWidgetProvider() {
...
-                    setViewVisibility(
-                        R.id.widgetCameraPlaceholder,
-                        View.VISIBLE
-                    )
-                    setViewVisibility(
-                        R.id.widgetCameraImage,
-                        View.GONE
-                    )
+//                    setImageViewResource(
+//                        R.id.widgetCameraImage,
```
dshokouhi, line 71: We should not leave commented code, this looks like it should be used though right?

app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
```diff
@@ -286,6 +288,7 @@ open class HomeAssistantApplication : Application() {
 
         ContextCompat.registerReceiver(this, buttonWidget, screenIntentFilter, ContextCompat.RECEIVER_NOT_EXPORTED)
         ContextCompat.registerReceiver(this, entityWidget, screenIntentFilter, ContextCompat.RECEIVER_NOT_EXPORTED)
+        ContextCompat.registerReceiver(this, cameraWidget, screenIntentFilter, ContextCompat.RECEIVER_NOT_EXPORTED)
```
dshokouhi, line 291: Why add this here? the widget should update once per hour. Additionally I am not seeing it update when the screen turns on but if I tap on the widget I see 2 updates after the screen turns on.

app/src/main/java/io/homeassistant/companion/android/widgets/camera/CameraWidget.kt
```diff
@@ -228,41 +148,60 @@ class CameraWidget : AppWidgetProvider() {
...
-    override fun onEnabled(context: Context) {
-        // Enter relevant functionality for when the first widget is created
-    }
-
-    override fun onDisabled(context: Context) {
-        // Enter relevant functionality for when the last widget is disabled
-    }
+//    override suspend fun onEntityStateChanged(context: Context, appWidgetId: Int, entity: Entity<*>) {
+//        super.onEntityStateChanged(context, appWidgetId, entity)
+//        widgetScope?.launch {
```
dshokouhi, line 201: Please dont leave commented code in the PR

app/src/main/java/io/homeassistant/companion/android/widgets/camera/CameraWidget.kt
```diff
@@ -228,41 +148,60 @@ class CameraWidget : AppWidgetProvider() {
...
                     "entity id: " + entitySelection + System.lineSeparator()
             )
-            cameraWidgetDao.add(
+            repository.add(
                 CameraWidgetEntity(
-                    appWidgetId,
-                    serverSelection,
-                    entitySelection,
-                    tapActionSelection
+                    id = appWidgetId,
```
dshokouhi, line 159: This change feels unnecessary adds more to the amount of lines changed, we should avoid reformatting the page as it makes the review more difficult.

common/src/main/java/io/homeassistant/companion/android/common/data/repositories/BaseDaoWidgetRepository.kt
```diff
@@ -0,0 +1,25 @@
+package io.homeassistant.companion.android.common.data.repositories
```
dshokouhi, line 1: Please move this and all other files under the newly created `repositories` folder and place it under `main` to avoid adding this to Wear OS. The database code is fine to be here as both Wear OS and the main app share database code.

dshokouhi (changes requested):
There is a lot of changes in this PR so this review is really only a first pass after looking at the code briefly and running the debug APK for testing.

First it looks like the default widget theme was changed when I look at adding widgets in the launcher. The default theme for dynamic colors is now changed to the device theme which is odd and makes it less personable in my opinion.

image

image

Widget updates when teh screen turns on is not correct either. I am seeing production and the debug app falling out of sync when changes to the entity occur off screen. Production does stay up to as expected. I am also not seeing the subscription restarting to keep it up to date like production app.

image

Another request please go through the changes and any that show line numbers being changed without any real changes please revert them. 

image

There is a lot and it makes teh review very difficult to follow and see if anything got lost as a result. I understand android studio may offer formatting but thats ok to use on new files, existing files its too much especially when other bits of code are touched it makes the review a lot harder.

ivorsmorenburg:
This will be splitted into different PR's linked to the issues

app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
ivorsmorenburg: If the camera is not registered stays like that, I feel getting the latest camera snapshot.
https://github.com/user-attachments/assets/f9964768-a61a-46d6-9d86-7c06c14cc57f

common/src/main/java/io/homeassistant/companion/android/common/data/repositories/BaseDaoWidgetRepository.kt
ivorsmorenburg: When adding repositories, will add a layer where we can do the logic to interact with the database (no Android objects). So wear should also use this methods rather than the database Dao directly, as this makes easier to make the logic testable.

app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
dshokouhi: i dont know what the video is supposed to shw but on the surface it looks like its taking over a minute to update?

ivorsmorenburg: Sorry no, is using the version without the broadcast and then compiling and relaunching the app

dshokouhi: ok, as this code was moved to #4648 we should move this convo there :)

PR #4647, opened 2024-09-16
Updates .gitignore to not include .kotlin/ changes into git

ivorsmorenburg:
## Summary

.gitignore update to not include .kotlin/ generated folder into git changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ google-services.json
 *.iml
 .idea/
 .gradle/
+.kotlin/
 build/
 *.keystore
 


dshokouhi:
Personally I am a little indifferent on this change. The only time I see this is if the kotlin build fails, other times android studio will clean it up. Otherwise I think we may have added this to `.gitignore` a long time ago. These files are specific to changes done on a local machine. Personally I think we should delete these and just not commit them.

ivorsmorenburg:
Yes I agree with that, but it makes life a lot easier when switching branches and Android Studio crashes

jpelgrom (approved):
Thanks. This seems to follow [Kotlin plugin changes](https://youtrack.jetbrains.com/issue/KTIJ-30076).

Issue #4040, opened 2023-12-07
phone state sensor stopped working after android 14 update


ea5522:
**Home Assistant Android app version(s):**
2023.11.2-full
**Android version(s):**
14
**Device model(s):**
sm-g998u
**Home Assistant version:**
latest
**Last working Home Assistant release (if known):**
issue happend after update to android 14
**Description of problem, include YAML if issue is related to notifications:**
phone state sensor stopped working after update to android 14

**Companion App Logs:**
not needed

**Screenshot or video of problem:**

**Additional information:**

ea5522:
Hi Daniel, I isolated the issue to gsm calls and some apps. Whatsapp and my voip app works but gsm stopped working. Telegram calls never worked. Where do I find companion app logs?

dshokouhi:
> Where do I find companion app logs?

Settings > companion app > troubleshooting > show and share logs

make sure to open the app, reproduce the issue and then get the logs

ea5522:
Ok, seems like it works when companion app is opened. When minimized, it works 50/50. Possibly an android issue. I did check the battery optimization settings, etc..

dshokouhi:
Samsung devices are known to add multiple different battery optimization settings under different names like Power Saving, Data Saving etc... check all your device settings to see if a new one was added in Android 14

ea5522:
Yeah, I checked them all already. Strange, i'll report back if i find something

ccristal:
Exactly same issue here on a Galaxy S21 after Android14 update. Phone State sensor gets updated immediately as soon as the HA companion app is brought to foreground. If it's not, it's a hit and miss.

dshokouhi:
> I have disabled all the optimizations and enabbled every permission I could find under the Settings->Apps->Home Assistant, to no avail.

have you checked things like power saving, data saving etc...? Those are Samsung specific settings and they will not be found under the HA app.

your device is indeed getting intents as expected so looks like the OS is not sending them when screen is off so most likely there is another new setting added that needs to be disabled.

`12-31 13:19:01.884  5320  5320 D SensorReceiver: Received intent: android.intent.action.PHONE_STATE`

jpelgrom:
> your device is indeed getting intents as expected so looks like the OS is not sending them when screen is off 

Which [could be explained by an Android 14 behavior change] (https://developer.android.com/about/versions/14/behavior-changes-all#pending-broadcasts-queued), nothing you can do about that.

dshokouhi:
[...]

PR #4661, opened 2024-09-24
Use receiver exported for all sensor receiver intents

dshokouhi:
## Summary

After #4559 was added I was doing more research into the [missing intent issue](https://github.com/home-assistant/android/issues/4040). I know that issue predates the android 14 SDK requirement however I noticed the following google docs page that was updated this month 🙃 

https://developer.android.com/develop/background-work/background-tasks/broadcasts#context-registered-receivers

specifically this part in step 4 stood out the most to me.

` If this receiver is listening for broadcasts sent from the system or from other apps—even other apps that you own—use the RECEIVER_EXPORTED flag.`

and then followed by

`If instead this receiver is listening only for broadcasts sent by your app, use the RECEIVER_NOT_EXPORTED flag.`

based on the above alone we should always export all intents because they are not sent by our own app.

The note underneath step 4 also confirms we need to use the same exported flag because we cannot intermix in the same receiver.

## Any other notes

important to note that in my own setup I have not noticed an issue in the internal track, testing this PR also shows intents working the same.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -94,7 +94,7 @@ open class HomeAssistantApplication : Application() {
                 addAction(Intent.ACTION_POWER_CONNECTED)
                 addAction(Intent.ACTION_POWER_DISCONNECTED)
             },
-            ContextCompat.RECEIVER_NOT_EXPORTED
+            ContextCompat.RECEIVER_EXPORTED
         )
 
         // This will cause interactive and power save to update upon a state change
@@ -106,7 +106,7 @@ open class HomeAssistantApplication : Application() {
                 addAction(Intent.ACTION_SCREEN_ON)
                 addAction(PowerManager.ACTION_POWER_SAVE_MODE_CHANGED)
             },
-            ContextCompat.RECEIVER_NOT_EXPORTED
+            ContextCompat.RECEIVER_EXPORTED
         )
 
         // Update Quest only sensors when the device is a Quest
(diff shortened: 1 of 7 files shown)


jpelgrom:
I can sort of follow your logic for the sensor receiver here, but you're also updating others where there is no mixing? The page also includes a note between the two parts you mentioned that non-privileged system broadcasts will work with `NOT_EXPORTED` which is the list in the original PR and I think these are working fine right now (like widgets).

dshokouhi:
ok let me rollback the changes for widgets and complications, when reading the change they really make it seem like you need to use `RECEIVER_EXPORTED` so you dont miss any intents. Even when it talks about when its ok to use `RECEIVER_NOT_EXPORTED` it still makes it sound like you can miss intents so I figured for any system intents we should always use exported instead of not.

jpelgrom:
For battery intents: the receiver is `null` here because we only care about the sticky intent, so do we need to change this at all? It seems to be working as it is right now and there is no mixing of receivers like in the other situations.

dshokouhi:
> For battery intents: the receiver is `null` here because we only care about the sticky intent, so do we need to change this at all? It seems to be working as it is right now and there is no mixing of receivers like in the other situations.

tough to tell the missing intent issue had some mentions of the charger intents 🤔 

https://github.com/home-assistant/android/issues/4040#issuecomment-2091881265 and https://github.com/home-assistant/android/issues/4040#issuecomment-2257295879

however you have a good point on the `null` receiver so it may not be necessary as you mentioned :)

jpelgrom (approved):
OK, let's do this to be safe and not mix exported/not exported in the same receiver

PR #4662, opened 2024-09-24
Use correct location batching call

dshokouhi:
## Summary

This shouldve been done in https://github.com/home-assistant/android/pull/2999 however I just now noticed the mistake while reviewing the code to see why sometimes updates are just randomly delayed. I am not sure if this is the case but I wanted to use the correct API here. I am also ok with removing this line as well and not having any batching if that is what we would like to do. Apparently it was not working for a while 😬 

## Any other notes

Ref of old API used: https://developers.google.com/android/reference/com/google/android/gms/location/LocationRequest#setMaxWaitTime(long)


jpelgrom:
> I am also ok with removing this line as well and not having any batching if that is what we would like to do.

The system is already limiting background location updates so I think we can keep it in, makes it nicer for people on old devices while there should be no/minimal differences on modern devices.

dshokouhi:
ok will leave as is then :)

Issue #3354, opened 2023-02-21
Cannot use a whitespace-only SSID for "Home Network WiFi SSID"


bburky:
**Home Assistant Android app version(s):** 2023.1.1-full

**Android version(s):** 13 (TQ1A.230105.002)

**Device model(s):** Pixel 7 Pro

**Home Assistant version:** 2023.2.5

**Last working Home Assistant release (if known):** 2023.2.5

**Description of problem, include YAML if issue is related to notifications:**

An SSID of only whitespace is actually a valid SSID. I currently have my network name set to `  ` (two space characters).

When configuring the "Home Network WiFi SSID" in the Home Assistant Companion for Android app, it rejects a whitespace-only SSID. After tapping "Add", the "New WiFi SSID" text input is outlined red and an "!" icon is shown. There is no error message shown.

(I won't ask you to support newlines (which... kind of actually work in SSIDs too... but break absolutely everything) but I've generally not actually had compatibility issues with spaces, so I thought I'd file a bug.)

**Screenshot or video of problem:**

**Additional information:**

PR #4674, opened 2024-09-27
Allow only spaces SSID when configuring home network wifi (#3354)

frenk91:
As for issue #3354 this changes allow the use of just spaces in wifi ssid


jpelgrom:
This will still trigger the `isBlank()` check; 

https://pl.kotl.in/rMnoaxQ-4

Trim was originally added to remove the space added at the end after tapping a suggestion on the keyboard. But we should still make sure you cannot enter a SSID with no length.

ivorsmorenburg:
I feel that the function should just set the WiFi name without any kind of logic. To add an SSID is something that should be done from a list of scanned SSID, so the validation of the name is not based in rules but in the actual value.

jpelgrom (approved):

jpelgrom:
> should be done from a list of scanned SSID

You may not be near the network to pick it from a list. The currently connected network is already suggested and can be added in one tap if not added.

Issue #3325, opened 2023-02-09
Wear OS home screen occasionally crashes with concurrent modification exception


dshokouhi:
**Home Assistant Android app version(s):**
latest beta for both

**Android version(s):**

Phone: 13
Watch: 11

**Device model(s):**

Pixel 7 pro
Pixel watch

**Home Assistant version:**

2023.1.x

**Last working Home Assistant release (if known):**

n/a

**Description of problem, include YAML if issue is related to notifications:**

Sometimes the wear OS app will load and then crash within a few seconds after showing the areas.

**Companion App Logs:**

```
2023-02-09 09:40:54.755 16190-16190 AndroidRuntime          io....stant.companion.android.debug  E  FATAL EXCEPTION: main
                                                                                                    Process: io.homeassistant.companion.android.debug, PID: 16190
                                                                                                    java.util.ConcurrentModificationException
                                                                                                    	at androidx.compose.runtime.snapshots.StateListIterator.validateModification(SnapshotStateList.kt:278)
                                                                                                    	at androidx.compose.runtime.snapshots.StateListIterator.next(SnapshotStateList.kt:257)
                                                                                                    	at io.homeassistant.companion.android.home.views.MainViewKt$MainView$1$3$1.invoke(MainView.kt:259)
                                                                                                    	at io.homeassistant.companion.android.home.views.MainViewKt$MainView$1$3$1.invoke(MainView.kt:76)
                                                                                                    	at androidx.wear.compose.material.ScalingLazyColumnKt$ScalingLazyColumn$1$1$2$1.invoke(ScalingLazyColumn.kt:425)
                                                                                                    	at androidx.wear.compose.material.ScalingLazyColumnKt$ScalingLazyColumn$1$1$2$1.invoke(ScalingLazyColumn.kt:410)
                                                                                                    	at androidx.compose.foundation.lazy.LazyListItemProviderKt$rememberLazyListItemProvider$1$itemProviderState$1.invoke(LazyListItemProvider.kt:53)
                                                                                                    	at androidx.compose.foundation.lazy.LazyListItemProviderKt$rememberLazyListItemProvider$1$itemProviderState$1.invoke(LazyListItemProvider.kt:52)
                                                                                                    	at androidx.compose.runtime.snapshots.Snapshot$Companion.observe(Snapshot.kt:2139)
                                                                                                    	at androidx.compose.runtime.DerivedSnapshotState.currentRecord(DerivedState.kt:161)
[...]

PR #4680, opened 2024-10-02
Attempt to fix Wear OS concurrent modification bug

dshokouhi:
## Summary

This is an attempt to fix: #3325 by avoiding clearing a list and just reassigning it which should hopefully help fix the issue. At least in my testing where I am able to reproduce the crash by just keeping the app open, with the updated code I am unable to replicate it. The issue may be caused by having a very large list and once cleared and updated may cause the modification crash. This should get some more testing but I am happy with the behavior compared to master branch with my setup.

### DIFF
--- a/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
+++ b/wear/src/main/java/io/homeassistant/companion/android/home/MainViewModel.kt
@@ -229,8 +229,7 @@ class MainViewModel @Inject constructor(
 
         if (!isFavoritesOnly) {
             areaRegistry = getAreaRegistry.await()?.also {
-                areas.clear()
-                areas.addAll(it)
+                areas = it.toMutableList()
             }
             deviceRegistry = getDeviceRegistry.await()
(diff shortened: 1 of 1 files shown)


jpelgrom:
I'm unable to reliably reproduce this issue, and expected Compose to handle this intelligently (remove all + add new = it figures out what changed between the two without considering the empty list), but if you're saying it is now better I'm willing to trust that.

With that said, this does seem to break something for Compose's update magic on sub-screens like areas or all entities. You can tap an entity and it will correctly toggle, but the state isn't reflected until you exit out and open it again. With favorites it does still update the UI correctly.

dshokouhi:
Thanks for testing I'll see about fixing the area issue, that should def be fixed otherwise this fix is no good :(

dshokouhi:
hmm may not be an easy fix here as the variables responsible for showing the state updates are the same variables touched by this PR :( I am surprised its not updating if all thats really changed is clearing and adding to reassignment.

PR #4686, opened 2024-10-02
Add native flow/support for Improv Wi-Fi devices with Bluetooth

jpelgrom:
## Summary

Adds a native flow to the app to add [Improv Wi-Fi devices](https://www.improv-wifi.com/ble/) using the Android device's Bluetooth. This mostly follows the iOS companion app implementation but with a more Android-like UI.

Two minor changes are in this PR which aren't obviously connected to the flow:
 - the dim amount for the bottom sheet dialog is changed everywhere to match what it looks like on the Assist 'dialog'
 - the root view of `WebViewActivity` was changed to give the added Snackbar nice animations and make it possible to swipe to dismiss, which required a few references to be updated

This PR is submitted as a draft because the app feature is basically complete but I'm waiting for the PRs on the [improv-wifi/sdk-android repo](https://github.com/improv-wifi/sdk-android) to be accepted, and hopefully we can publish the SDK instead of including an aar file. The aar file added right now is a local build of the SDK with all my PR changes included.

## Screenshots

The following video shows the flow:

https://github.com/user-attachments/assets/1459b61f-45ad-42d2-97dd-16e51ee3862b

When tapping the frontend option to add a device, permission dialogs are shown if required (depending on Android version and what permissions the app already has), and if granted the app starts scanning in the background. If any devices are found a Snackbar will be shown. Tapping on the configure action, like frontend found devices, will trigger a sheet/dialog to select the device, enter your Wi-Fi credentials, and connect and configure. When done tapping Continue will redirect to setup that integration in the frontend.

n/a


dshokouhi (approved):
Looks good to me

JBassett (approved):

PR #4696, opened 2024-10-04
Breaking change: Improve sensor mobile_data

frimtec:
## Summary
Improve gathering data for sensor mobile_data. 
The currently used API for the mobile_data sensor is marked as  ```@UnsupportedAppUsage``` and is therefore not supported on newer devices (e.g. Pixel 9).
As of SDK 26 there is an API ```TelephonyManager#isDataEnabled``` that is supported on all devices.

## Screenshots

Frontend not affected due to this change.

Documentation: home-assistant/companion.home-assistant#435
The existing documentation introduced with above pull request is still fully valid.

### DIFF
--- a/common/src/main/java/io/homeassistant/companion/android/common/sensors/MobileDataManager.kt
+++ b/common/src/main/java/io/homeassistant/companion/android/common/sensors/MobileDataManager.kt
@@ -34,6 +34,7 @@ class MobileDataManager : SensorManager {
     override fun docsLink(): String {
         return "https://companion.home-assistant.io/docs/core/sensors#mobile-data-sensors"
     }
+
     override val name: Int
         get() = commonR.string.sensor_name_mobile_data
 
@@ -42,7 +43,7 @@ class MobileDataManager : SensorManager {
     }
 
     override fun requiredPermissions(sensorId: String): Array<String> {
-        return if (sensorId == mobileDataRoaming.id) {
+        return if (sensorId == mobileDataRoaming.id || Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
             arrayOf(Manifest.permission.READ_PHONE_STATE)
         } else {
             arrayOf()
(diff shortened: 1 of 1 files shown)


dshokouhi:
I see that permissions are changing here for a sensor, that will be a breaking change for users who have that sensor enabled but not the permission granted. Is the sensor not updating properly? on my pixel 9 pro xl its showing the correct state.

frimtec:
The state for the sensor "mobile_data" on my pixel 9 Pro is always "on" even when mobile data is disabled. I could also reproduce that behavior on a pixel 7 of a friend.
With this fix it works again on my pixel 9, as  I was used it with my old pixel 4.

I have currently no idea why it still works on your pixel 9 pro xl.

The declaration of the old Android API Settings.Global.MOBILE_DATA is as follows :
```
        @UnsupportedAppUsage
        @Readable
        public static final String MOBILE_DATA = "mobile_data";
````
```@UnsupportedAppUsage``` means: 
___Marks elements as internal and not intended for external use.
Avoid using unsupported APIs to ensure app stability and compatibility.___

Therefore I think it makes more sense to use the public API if available (as of SDK 26).

As the new API requires permission READ_PHONE_STATE, you are right this could be a breaking change.
I updated the ```requiredPermissions()``` method to make the change as smal as possible.

dshokouhi:
My bad i did not realize that the sensor state had stopped working, I had incorrectly assumed as I had mobile data enabled that the sensor had the correct state 🙈  Now I wonder how long it was broken for 😂

dshokouhi (approved):

dshokouhi:
> Now I wonder how long it was broken for 😂

Looks like my Pixel watch had the correct state with the old method. Surprisingly it's on the same version of Android as my phone. Guess it shows the platform differences there lol. Thanks again for the PR 🙏

frimtec:
> Looks like my Pixel watch had the correct state with the old method. Surprisingly it's on the same version of Android as my phone. 

It is not the android version that is relevant whether it works or not, but the device itself. E.g. the old API still works up to android 15 on the android emulator. But more and more devices do not support the old API because it violates the security model. With the old API you can access the phone state without the required permission READ_PHONE_STATE.

Thx for merging the pull request 👍

PR #4721, opened 2024-10-11
Fix leading zero handling in toJsonType method

blackhead1981:

## Any other notes
- Added a check in the `toJsonType` method to preserve leading zeros in strings.
- This ensures that strings starting with a leading zero are treated as strings rather than numbers, preventing unwanted data conversion.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/widgets/common/WidgetDynamicFieldAdapter.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/widgets/common/WidgetDynamicFieldAdapter.kt
@@ -196,16 +196,22 @@ class WidgetDynamicFieldAdapter(
             return jsonArray.toList()
         }
 
-        // Check if the string starts with a leading zero
-        if (this.trim().startsWith("0") && this.trim().length > 1) {
-            return this.trim() // Treat it as a string to preserve leading zeros
-        }
-
-        // Parse the base types
         this.trim().let { trimmedStr ->
+            // Check if the string is a double (e.g., 0.5)
+            trimmedStr.toDoubleOrNull()?.let {
+                return it
+            }
+
+            // Check if the string starts with a leading zero (but not a decimal number)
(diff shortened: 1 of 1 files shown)


blackhead1981:
After a brief check in the debug APK, it seems to be working so far. At least a PIN code with a leading zero works without any issues. I will conduct further tests tomorrow. I kindly ask for your additional testing as well.

blackhead1981:
After extensive testing, I was unable to identify any issues. Everything seems to be working as expected without any problems.

jpelgrom:
I wonder if it's worth making this more specific? For example `0.5` will now be sent as a string while it would work fine as a number. This change is really only necessary for input starting with 0 that would otherwise be treated as a whole number (and therefore the 0 would be dropped).

blackhead1981:
Thank you for your feedback!

You made a good point about handling inputs like "0.5" and "0." I’ve updated the logic to parse valid decimal numbers starting with '0' as Double, except for '0' itself (which should be an Int), while keeping strings for cases like "02."

As a beginner, I encountered some difficulties, so I appreciate your understanding. I've also verified the logic in the playground: [Kotlin Playground Link](https://pl.kotl.in/M8d5N2QRF).

Thanks for your input!

blackhead1981:
I'm actually uncertain whether "02.5" should be treated as a Double or left as a String. My concern is that for cases like passwords or other data starting with "0", treating it as a Double would drop the leading zero, which might be intentional. Shouldn't we assume a leading zero is meant to be preserved when explicitly written, except for cases like "0.***"?

As for "0.231231", that's a mistake in my comment. Of course, it should be treated as a Double.

dshokouhi:
> Shouldn't we assume a leading zero is meant to be preserved when explicitly written

IMO we should preserve and send exactly whatever a user saves. If they want to send "00000.00010" then let them. I think that is the general expectation.

blackhead1981:
Yes, I think so too.
The code should work accordingly then.

Of course, one option could be to ensure it's treated as a string by wrapping the input in quotes. However, I don't think that would be intuitive for most users.

jpelgrom (approved):

PR #4722, opened 2024-10-11
Add car speed sensor

dshokouhi:
## Summary

Adds a car speed sensor whose state will contain display speed in meters per second, Attributes for raw speed and also the display units are supported (untested due to head unit emulator limitations). The values can be either positive, negative or 0 depending on vehicle movement (driving, reverse, stopped)

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#1119

### DIFF
--- a/app/src/full/AndroidManifest.xml
+++ b/app/src/full/AndroidManifest.xml
@@ -5,6 +5,7 @@
     <uses-sdk tools:overrideLibrary="com.google.android.gms.threadnetwork" />
     <uses-permission android:name="com.google.android.gms.permission.CAR_FUEL" />
     <uses-permission android:name="com.google.android.gms.permission.CAR_MILEAGE" />
+    <uses-permission android:name="com.google.android.gms.permission.CAR_SPEED" />
 <!--    <uses-permission android:name="android.permission.health.READ_ACTIVE_CALORIES_BURNED" />-->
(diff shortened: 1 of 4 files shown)


jpelgrom:
I'm not sure I agree with this sensor having attributes. They seem unnecessary and don't all describe the sensor state.

With the default sensor update speed, this isn't very relevant or up-to-date. Are there any intents to listen for (like completely stopped/started moving, thresholds passed) or can you add a recommendation for a combination that does this?

dshokouhi:
> I'm not sure I agree with this sensor having attributes. They seem unnecessary and don't all describe the sensor state.
> 
> With the default sensor update speed, this isn't very relevant or up-to-date. Are there any intents to listen for (like completely stopped/started moving, thresholds passed) or can you add a recommendation for a combination that does this?

I'll remove the attributes but the unit is interesting admittedly I wonder if US cars and EU have a difference in what reports? 

I don't believe there is an intent however this sensor won't update anymore than the other more important sensors like battery and fuel level. This more closes the loop on the sensors that can be added on the list. 

Also I'm curious to see if my car supports this sensor 🙈

jpelgrom (approved):

jpelgrom:
There is a merge conflict but other than that it looks good to me!

PR #4732, opened 2024-10-16
Add arrow to bottom sheet list items to indicate tappable

jpelgrom:
## Summary

 - Based on feedback, add an arrow to bottom sheet list items to indicate that they can be tapped. This isn't really an Android UI pattern, but I agree it should be more obvious (especially with 1 item lists), don't have a good alternative and we already use the same pattern in onboarding.
 - Make onboarding arrow RTL aware and clean up old resources

## Screenshots

|Light|Dark|
|----|----|
|Bottom sheet 'Devices ready to set up', with one item 'esp-web-tools-e34304' now showing an arrow to the right of it, light mode|Bottom sheet 'Devices ready to set up', with one item 'esp-web-tools-e34304' now showing an arrow to the right of it, light mode|

n/a


dshokouhi (approved):
Nice improvement! Hopefully its more noticeable to users now!

PR #3668, opened 2023-07-17
Send state based on reported status, apply car prefix to car sensors, add some logging around new car sensors

dshokouhi:
## Summary

- Send state based on reported status so users know if a sensor is not implemented  
- Apply `Car` prefix to sensor names for consistency and easy identification.
- Add some more logging for troubleshooting
- Minor clean-up

image

Add some debug logging around car data as I am only seeing Car Name being pulled in, even after going for a drive increasing odometer and decreasing fuel.

## Link to pull request in Documentation repository

Documentation: home-assistant/companion.home-assistant#

## Any other notes

CC: @drosoCode in case you have any ideas on why some data is not retrieved.

PR #4733, opened 2024-10-16
Add a car range remaining sensor

dshokouhi:
## Summary

Add a car range sensor

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#1121

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
@@ -78,6 +78,25 @@ class CarSensorManager :
                 "android.car.permission.READ_CAR_DISPLAY_UNITS"
             )
         )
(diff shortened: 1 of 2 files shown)


jpelgrom:
Maybe my brain is fried after reviewing a lot of strings just now, but isn't "Remaining car range" a more logical sensor name? Same for the description, why not use the simpler description from the documentation?

dshokouhi:
> but isn't "Remaining car range" a more logical sensor name?

I think this sensor may be a good fit for that but after #3668 we have some consistency to consider here 🤔 



> Same for the description, why not use the simpler description from the documentation?

sure ill update that to match our docs rather than Google 😄

jpelgrom:
> > but isn't "Remaining car range" a more logical sensor name?
> 
> I think this sensor may be a good fit for that but after #3668 we have some consistency to consider here 🤔

I do like consistency and "car range remaining" is still better than "car remaining range". Thanks for the reminder :)

PR #4746, opened 2024-10-18
Correct sensor identification for car speed sensor

dshokouhi:
## Summary

This should be a sensor and not diagnostic


jpelgrom:
What is your reasoning here? As I understand it we should use diagnostic for sensors that you cannot directly influence. I can understand car speed as primary but range remaining feels like it should be classified diagnostic similar to the battery remaining charge time sensor.

dshokouhi:
> What is your reasoning here? As I understand it we should use diagnostic for sensors that you cannot directly influence. I can understand car speed as primary but range remaining feels like it should be classified diagnostic similar to the battery remaining charge time sensor.

Agreed reverted the change for car range remaining. Received a comment from an internal tester about their identification but I probably shouldve given it a lil more thought :)

jpelgrom (approved):

PR #4750, opened 2024-10-22
Add package name to health connect sensors, additional sensor tweaks

dshokouhi:
## Summary

- update weight icon to be more correct
- Add state class total increasing for total calories as the sensor resets at midnight and accumulates until then
- Add package as an attribute so users can understand from where the data came from

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
@@ -52,6 +52,7 @@ class HealthConnectSensorManager : SensorManager {
             "mdi:fire",
             "energy",
             unitOfMeasurement = "kcal",
+            stateClass = SensorManager.STATE_CLASS_TOTAL_INCREASING,
             entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC
         )
 
@@ -60,7 +61,7 @@ class HealthConnectSensorManager : SensorManager {
             type = "sensor",
             commonR.string.basic_sensor_name_weight,
             commonR.string.sensor_description_weight,
-            "mdi:weight",
+            "mdi:scale-bathroom",
             unitOfMeasurement = "g",
             entityCategory = SensorManager.ENTITY_CATEGORY_DIAGNOSTIC,
             deviceClass = "weight"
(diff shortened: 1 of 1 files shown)


app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
```diff
@@ -104,13 +105,16 @@ class HealthConnectSensorManager : SensorManager {
...
             onSensorUpdated(
                 context,
                 totalCaloriesBurned,
-                BigDecimal(it.inKilocalories).setScale(2, RoundingMode.HALF_EVEN),
+                BigDecimal(energy.inKilocalories).setScale(2, RoundingMode.HALF_EVEN),
                 totalCaloriesBurned.statelessIcon,
-                attributes = mapOf("endTime" to LocalDateTime.of(LocalDate.now(), LocalTime.now()).toInstant(ZoneOffset.UTC))
+                attributes = mapOf(
+                    "endTime" to LocalDateTime.of(LocalDate.now(), LocalTime.now()).toInstant(ZoneOffset.UTC),
+                    "package" to totalCaloriesBurnedRequest.dataOrigins.map { it.packageName }
```
jpelgrom, line 116: This returns a `Set` so you're creating a list here, is that intentional? If so maybe it is worth adding a `s` to the attribute name to differentiate it.

jpelgrom (commented):
I'm not sure the attribute should be `package` or if `source` makes more sense for less technical users 🤔

Other changes look good to me

jpelgrom:
Also, was Health Connect approved? Or is this just cleaning it up in case it is like the other cleanups we've done?

app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
dshokouhi: I made it into a list as I was unsure a `Set` would translate properly in teh attribute so wanted to use something I knew worked :) and I agree adding a `s` here makes sense.

dshokouhi:
> I'm not sure the attribute should be `package` or if `source` makes more sense for less technical users 🤔

hmm `source` feels more correct I think I will update

jpelgrom (approved):

PR #4751, opened 2024-10-22
Hide widgets from automotive build and prevent intent registration too

dshokouhi:
## Summary

As normal app widgets dont exist in automotive we should not register the screen on intent nor should we show the widget option in settings.

Should prevent below crash from unsupported device:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'int[] android.appwidget.AppWidgetManager.getAppWidgetIds(android.content.ComponentName)' on a null object reference
    at io.homeassistant.companion.android.widgets.button.ButtonWidget$updateAllWidgets$1.invokeSuspend(ButtonWidget.kt:97)
    at android.os.Looper.loop(Looper.java:288
```

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
@@ -176,7 +176,7 @@ class SettingsFragment(
         val isAutomotive =
             Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && requireContext().packageManager.hasSystemFeature(PackageManager.FEATURE_AUTOMOTIVE)
 
-        findPreference<PreferenceCategory>("widgets")?.isVisible = Build.MODEL != "Quest" || isAutomotive
+        findPreference<PreferenceCategory>("widgets")?.isVisible = Build.MODEL != "Quest" || !isAutomotive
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -173,7 +173,10 @@ class SettingsFragment(
             it.entryValues = percentages.map { pct -> pct.toString() }.toTypedArray()
         }
 
-        findPreference<PreferenceCategory>("widgets")?.isVisible = Build.MODEL != "Quest"
+        val isAutomotive =
+            Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && requireContext().packageManager.hasSystemFeature(PackageManager.FEATURE_AUTOMOTIVE)
+
+        findPreference<PreferenceCategory>("widgets")?.isVisible = Build.MODEL != "Quest" || !isAutomotive
```
jpelgrom, line 179: This will make the preference visible because you're using or instead of and. Visible if [not Quest] or [not Automotive] which can't both be true.

dshokouhi: 🙈

jpelgrom (approved):

PR #4760, opened 2024-10-27
Add feature flag concept and screen

TimoPtr:
## Summary

This PR introduces a feature management system to streamline future development, laying a flexible foundation designed to adapt as needs evolve. A new UI for managing feature toggles is available in the troubleshooting menu, allowing features to be controlled directly from the app.

Additionally, this update includes unit tests for feature-related functionality and a CI job to execute them automatically on each PR.

This design takes inspiration from [Baracoda’s blog on trunk-based development](https://baracoda.com/blog/trunk-based-development).

## Screenshots

https://github.com/user-attachments/assets/86318da3-ae5c-4479-9d4f-fcffcc1fcb0b

## Any other notes

This is my first contribution to Home Assistant, and I tried to keep the existing structure and conventions.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -67,6 +67,28 @@ jobs:
       - name: Validate Lint
         run: ./gradlew lint
 
+  unit_tests:
+    runs-on: ubuntu-latest
+    steps:
+        - uses: actions/checkout@v4
+        - name: Set up JDK 17
+          uses: actions/setup-java@v4.5.0
+          with:
+              distribution: 'temurin'
+              java-version: '17'
+
+        - name: Mock google-services.json
+          run: |
+              cp .github/mock-google-services.json app/google-services.json
+              cp .github/mock-google-services.json wear/google-services.json
+              cp .github/mock-google-services.json automotive/google-services.json
+
+        - name: Setup Gradle
+          uses: gradle/actions/setup-gradle@v4
+
+        - name: Run test
+          run: ./gradlew test
+
   pr_build:
     runs-on: ubuntu-latest
     permissions:
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -1299,4 +1299,8 @@
     <string name="sensor_description_car_speed">The current speed of the car</string>
     <string name="basic_sensor_name_car_range_remaining">Car range remaining</string>
     <string name="sensor_description_car_range_remaining">The range remaining for the car in meters</string>
+    <string name="feature_flag">Feature flag</string>
+    <string name="feature_flag_summary">Manage feature flags for the app (advanced users only). Proceed with caution, as this area contains experimental features that may affect stability.</string>
+    <string name="string_feature_flag_update">Update</string>
+    <string name="string_feature_flag_cancel">Cancel</string>
 </resources>
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -104,10 +104,8 @@ android {
         unitTests.isReturnDefaultValues = true
     }
 
-    tasks.withType<Test>().configureEach {
-        useJUnitPlatform {
-            includeEngines("spek2")
-        }
+    tasks.withType<Test> {
+        useJUnitPlatform()
     }
 
     lint {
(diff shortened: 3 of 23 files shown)


common/src/main/res/values/strings.xml
```diff
@@ -1299,4 +1299,8 @@
     <string name="sensor_description_car_speed">The current speed of the car</string>
     <string name="basic_sensor_name_car_range_remaining">Car range remaining</string>
     <string name="sensor_description_car_range_remaining">The range remaining for the car in meters</string>
+    <string name="feature_flag">Feature flag</string>
+    <string name="feature_flag_summary">Manage feature flags for the app (advanced users only). Proceed with caution, as this area contains experimental features that may affect stability.</string>
+    <string name="string_feature_flag_update">Update</string>
+    <string name="string_feature_flag_cancel">Cancel</string>
```
TimoPtr, line 1305: I don't think I can add this new keys myself in Lokalise.

.github/workflows/pr.yml
```diff
@@ -67,6 +67,28 @@ jobs:
       - name: Validate Lint
         run: ./gradlew lint
 
+  unit_tests:
```
TimoPtr, line 70: This is a very basic job, that most probably should evolve over time.

common/src/main/res/values/strings.xml
jpelgrom: If your PR is accepted and merged, Lokalise will pick up the new strings automatically

bgoncal:
Could you bring some examples where you see this feature flag engine relevant for future HA features? Also have in mind the open source contributors and their interact with it.

On the iOS side I haven't felt the need for feature flags yet (even though I worked with a similar approach in past companies I worked on), so I am curious about your perspective.

bgoncal:
I see the value on it and I apretiate you sharing your past experience, I'd say this becomes interesting as soon as we have a real need for it, for now we focus on small PRs, with single purpose, that can be merged without breaking the rest of the code, for a big refacture, perhaps a feature flag engine could be helpful, on the other hand it's usually possible to breakdown refactors as well.

About A/B testing, we don't collect user data, as one of our privacy principles from the Open Home Foundation/Home Assistant.

We have a call today and we can discuss this more in depth, meanwhile @jpelgrom @dshokouhi are the right people to decide if this can be helpful for android right now or perhaps hold for future use cases.

jpelgrom:
While I do see the value of such a feature in general and for the developer while working on PR, I still don't fully see how it would benefit Home Assistant.
- The examples given for (partial) refactoring suggest to me you want to merge code that would otherwise break the app in some way and hide it behind a flag, instead of splitting it up into smaller chunks that don't break?
- As Bruno mentioned, Home Assistant has a strong stance on privacy. Using it for A/B testing as you suggested or random bits of AI are not really compatible as I see it.
- We use pull requests for discussions where opinions differ, not merging changes immediately and discussing later.
- There already are not a lot of contributors to the companion app. I feel like introducing a framework like this and mandating it would increase the barrier to entry, with very little gain (the app is practically already able to release from the main branch at any time, and there are limited conflicts.)

At a quick glance, the code looks fine. Did you also write/contribute to the blog post you've linked as it is from your current company?

TimoPtr:
I was more thinking about the case where the refactoring is done partially. Imagine that we redo a big Activity with multiple fragments to Compose with the navigation library. In that case, I would expect to have a first PR with a feature flag defined, and then subsequent PRs for each fragment transition to the new system. Finally, one last PR would remove the flag and the old code.

With this approach, everything can be merged quickly to main without losing any features. I don't know the code base well enough to say that this case will happen.

When it comes to the A/B testing, I'm fully aligned with the privacy concern, and I wouldn't want my data collected. I was more thinking about getting feedback on GitHub or Discord from the users about, let's say, a new UI that we want to test. I guess that there is inside the community users that would be ok testing new things and give us feedbacks.

> random bits of AI are not really compatible as I see it

I'm not sure I follow. Are you referring to my point on Gemini? If so, I was thinking about the local usage of LLM models. Today, I'm experimenting with running "small" LLMs locally, as the latest flagship devices come with significant power and dedicated hardware. It could be nice to leverage that. Again, I'm aligned with the privacy concerns, which is why I would prefer to keep the LLM running locally on open source libraries that is not collecting data.

I fully agree that forcing people to use this framework can be too complex. It probably shouldn't be forced. Maybe sometimes the reviewers could suggest splitting the PR if it's too big using feature flags or if the feature needs testing before broader deployment.

When it comes to the blog post, in my current position, I'm supervising the writing in the blog, contribute through reviews and approve the publications.

TimoPtr:
Closing for since we don't plan to use it.

PR #4763, opened 2024-10-29
Support behavior: textInput key on notification action for reply actions

### LINKED ISSUE #3503: A way to differentiate multiple REPLY actions
It is possible to send an actionable notification with multiple `REPLY` actions, that allow a custom text reply. When the `mobile_app_notification_action` response comes back, it cannot be differentiated which one was pressed.

Some kind of metadata field that can be included under the action key that is populated in the response.

If extra metadata cannot be added, maybe create more `REPLY` actions that do the same thing, like `REPLY_1`, `REPLY_2`, ,... `REPLY_9`, so that the button that was pressed can be differentiated by their action name. 

**Additional context**

jpelgrom:
## Summary

Fix #3503 by supporting the use of `behavior: textInput` on a notification action to trigger the reply action. This allows you to use the `action` key to differentiate the different actions, which is already possible on iOS.

## Screenshots

n/a

Documentation: home-assistant/companion.home-assistant#1130

## Any other notes

FCM changes: home-assistant/mobile-apps-fcm-push#170

cc @bgoncal

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -139,7 +139,7 @@ class MessagingManager @Inject constructor(
         const val INTENT_CLASS_NAME = "intent_class_name"
         const val URI = "URI"
         const val REPLY = "REPLY"
-        const val TEXT_INPUT = "textInput"
+        const val TEXT_INPUT = "textinput"
         const val HIGH_ACCURACY_UPDATE_INTERVAL = "high_accuracy_update_interval"
         const val PACKAGE_NAME = "package_name"
         const val CONFIRMATION = "confirmation"
@@ -1462,7 +1462,7 @@ class MessagingManager @Inject constructor(
                             )
                         }
                     }
-                    notificationAction.key == REPLY || notificationAction.behavior == TEXT_INPUT -> {
+                    notificationAction.key == REPLY || notificationAction.behavior?.lowercase() == TEXT_INPUT -> {
                         val remoteInput: RemoteInput = RemoteInput.Builder(KEY_TEXT_REPLY).run {
                             setLabel(context.getString(commonR.string.action_reply))
                             build()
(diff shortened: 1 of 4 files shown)


dshokouhi (approved):

bgoncal:
Will "reply_title" be the title of the action that the user tapped in the notification? I think we could improve the name of the variable, it's not very intuitive, perhaps keep as you suggested `reply_text_for` or `reply_text_for_action`

bgoncal (approved):

bgoncal:
I dropped you a message in discord about this, maybe we should use "identifier" like:
```
action: notify.mobile_app_iphone_11
data:
  data:
    actions:
      - action: REPLY
        identifier: action-1
        title: Action 1
      - action: REPLY
        identifier: action-2
        title: Action 2
  message: Testing action
```

In iOS I can't access the action title easily, I would need to cache it somewhere else and using identifier to me makes more sense, then the user can even change the title of the action freely without having to change their automations trigger for example (since the ID will not change if they dont want to)

jpelgrom:
I'll see if it is perhaps easier for Android to also support `behavior: textInput`. That way it stays aligned between apps with no changes required on the iOS side and users can use the `action` key as the identifier as suggested in the docs.

jpelgrom:
Changed it so you can use `behavior: textInput` like iOS for this and use the `action` as the identifier:

```yaml
action: notify.mobile_app_pixel_7a_dev
data:
  message: Test UnifiedPush 3a
  data:
    actions:
      - action: ONE
        title: Reply 1
        behavior: textInput
      - action: TWO
        title: Reply 2
        behavior: textInput
```

will generate the event:

```yaml
event_type: mobile_app_notification_action
data:
  action_1_key: ONE
  action_1_title: Reply 1
  action_1_behavior: textInput
  action_2_key: TWO
  action_2_title: Reply 2
  action_2_behavior: textInput
  message: Test UnifiedPush 3a
  webhook_id: ...
  server_id: "9"
  reply_text: Which one is it?
  action: TWO
  ...
origin: REMOTE
time_fired: "2024-10-30T19:05:20.917447+00:00"
context:
  ...
```

Please approve again if you agree with this as a better solution or reply with a comment otherwise :)

dshokouhi (approved):

bgoncal (approved):
LGTM!

Issue #1606, opened 2021-07-07
Add phone flash light/torch as light entity


michamichamicha:
**Is your feature request related to a problem? Please describe.**
Not related to a problem

**Describe the solution you'd like**
I would like the Companion App to facilitate a light entity that controls the torch of mobile phones.

**Describe alternatives you've considered, if any**
-

**Additional context**
Could be used for notifications, automations (arriving at home in the middle of the night, scanning an NFC tag and automatically turn on just the torch) and a lot more in my opinion.

dshokouhi:
> I would like the Companion App to facilitate a light entity that controls the torch of mobile phones.



so currently the `mobile_app` integration only provides sensors for the companion apps to use.  We would need to make controlling the flashlight a notification command and then represent its state as a binary sensor in the meantime.  A developer with python experience would need to add support for the light/switch entity.



> Maybe add a service to make the phone vibrate as well?

This could probably be its own feature request, probably can make some options configurable like pause and length of vibration

virtualJonesie:
In general, I would love to see the phone controllable via the app. For instance, when I run my bed time routine, Home Assistant turns on Do Not Disturb mode.

Other interesting integration points could be:
Ringer/Vibrate/Silent
Wifi On/Off
Bluetooth on/off
Airplane mode on/off
etc

dshokouhi:
> In general, I would love to see the phone controllable via the app. For instance, when I run my bed time routine, Home Assistant turns on Do Not Disturb mode.
> 
> Other interesting integration points could be:
> Ringer/Vibrate/Silent
> Wifi On/Off
> Bluetooth on/off
> Airplane mode on/off
> etc

Runner, DND, Bluetooth already done. Airplane mode impossible lol, think about it for a sec. Wifi no longer possible via Android API, google put a stop to it. You should read the docs so you know what you can already control.

virtualJonesie:
Fair point on Airplane Mode. Didn't think that one through. :)

I see that there are sensors for the items you listed, but I don't see any switches to control them. 

For example, I would love to turn on WiFi automatically when I arrive to my home location, to ensure I don't run on cell data unnecessarily. Or turn off the ringer at night when I place it on the charger. Controlling some basic functions would be really nice. 

If it's already possible, I may just be missing it. But when I search in the Developer Tools for my phone, all I see are sensors, binary_sensors, and device_tracker entities, no switch entities to control these settings.

dshokouhi:
[...]

PR #4766, opened 2024-10-30
Add notification command for turning the flashlight on/off

marazmarci:
## Summary
I added the `command_flashlight` notification command. It accepts two values in the `command` field: `turn_on` or `turn_off`.

Currently, my implementation only supports API 23 and above, while the app has `minSdkVersion = 21`. The main reason is my ~~laziness~~ lack of time, and because the Camera2 API is available starting from API 23, and I haven't found a trivial `...Compat` solution. If someone mentions needing this on API 21-22, it can be implemented as a separate feature request. Please let me know if this approach is OK for you.

The PR is marked as a draft because I still have to implement requesting the Camera permission, but I first wanted to ask for your advice on it. The problem is that in `MessagingManager` only so-called "special permissions" are being requested, and there are no runtime permissions, and it would require an `Activity`. I thought of adding another `registerForActivityResult(ActivityResultContracts.RequestPermission())` in `BaseActivity` and making `BaseActivity` subscribe to an event-like `Flow` from a singleton class, which would trigger a permission request, and `MessagingManager` could trigger this event. WDYT?

Related feature request: https://github.com/home-assistant/android/issues/1606

## Screenshots
N/A

Documentation: https://github.com/home-assistant/companion.home-assistant/pull/1131

## Any other notes
Future improvement possibilities:
- Support API 21 and 22
- Add `command: toggle`
- Add a `binary_sensor` representing the current state of the torch (maybe even a `light` entity?)
- Add [torch strength control](https://source.android.com/docs/core/camera/torch-strength-control)

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -540,7 +540,7 @@ class MessagingManager @Inject constructor(
                         }
                         COMMAND_FLASHLIGHT -> {
                             val command = jsonData[NotificationData.COMMAND]
-                            if (command in DeviceCommandData.ENABLE_COMMANDS) {
+                            if (command in DeviceCommandData.ENABLE_COMMANDS && Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
                                 handleDeviceCommands(jsonData)
                             } else {
                                 Log.d(TAG, "Invalid flashlight command received, posting notification to device")
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -78,6 +78,7 @@ import io.homeassistant.companion.android.sensors.LocationSensorManager
 import io.homeassistant.companion.android.sensors.NotificationSensorManager
 import io.homeassistant.companion.android.sensors.SensorReceiver
 import io.homeassistant.companion.android.settings.SettingsActivity
+import io.homeassistant.companion.android.util.FlashlightHelper
 import io.homeassistant.companion.android.util.UrlUtil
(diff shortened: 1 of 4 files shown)


dshokouhi:
> The problem is that in `MessagingManager` there are only so-called "special permissions" being requested, and there are no runtime permissions,

we actually request the phone permission for one specific intent I think you can use the same method. Take user to webview, then they grant the permission and try again.

 https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt#L689-L697

marazmarci:
I created a PR in the documentation repository: https://github.com/home-assistant/companion.home-assistant/pull/1131

app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -532,6 +538,15 @@ class MessagingManager @Inject constructor(
                                 sendNotification(jsonData)
                             }
                         }
+                        COMMAND_FLASHLIGHT -> {
+                            val command = jsonData[NotificationData.COMMAND]
+                            if (command in DeviceCommandData.ENABLE_COMMANDS) {
```
dshokouhi, line 543: if this command requires a certain android version we should check for it here like we do with `command_bluetooth`

marazmarci: Done! ✅

I also [updated the docs PR](https://github.com/home-assistant/companion.home-assistant/pull/1131/commits/0b1e9999597c4f77da2d9cd0d4cc542e84e3940a).

dshokouhi (changes requested):
Can you please submit a PR to the mobile app FCM repo adding the new command to the list commands so it can be excluded from the rate limit?

https://github.com/home-assistant/mobile-apps-fcm-push/blob/main/functions/android.js#L70-L74

dshokouhi (approved):
Tested debug APK and works as expected

PR #4783, opened 2024-11-03
Multi-action grid widget implementation

### LINKED ISSUE #1193: Grid-based multi-action scene control widget.

Not a problem but a frustration, currently I have to create an individual action widget (which is somewhat time consuming) for each scene that I would like to be accessible from the homepage, and there is no way to save the configuration of these widgets as a group.

Allow the user to create a grid with a predefined number of rows and columns, then assign each button to a scene that is already defined within Home Assistant.

Hue Pro and Hue apps already have similar widgets, but this obviously doesn't allow you to control any of your lights outside of Hue. Currently I have individual action widgets from HA that are set to turn on a scene for each of the scenes I use. The limitation of this is that it isn't easily duplicable, meaning that I have to manually create/configure each widget from HA live, and adding to a second device would take the full amount of time again.
**Additional context**

https://imgur.com/a/INjQxiF | ignore the center widget with the up/down power control as that is very different to make.

mrdanielps:
## Summary

Adds a new widget that allows the user to configure multiple actions in a grid. It's somewhat based on the existing ButtonWidget.

The widget is responsive, adjusting the number of columns based on size, and becomes scrollable when it doesn't fit all the actions.

Fixes #1193, #4549

## Screenshots

Dynamic colors - light
Screenshot_20241102_185606

Dynamic colors - dark
Screenshot_20241102_185620

API <31
Screenshot_20241102_185635

Configuration and resizing

https://github.com/user-attachments/assets/483986b6-3774-4f0a-bb5f-01fb1fa2ab6f

## Any other notes

I've limited the scope of this PR for easier review, hence the draft status, but I will update it based on feedback.
The main features I think are necessary before merging are auto-completion for icons and action data, feedback when the button is pressed, and maybe entity states.

Additionally, the existing button widget could be updated to reflect the UI changes and share some of their code. However, I didn't want to change too much considering a proposal exists to refactor this code (#4640).


bgoncal:
Looking great! Will the state of each item reflect in realtime? for example if the light is ON

mrdanielps:
> Looking great! Will the state of each item reflect in realtime? for example if the light is ON

Not yet. As mentioned in the additional notes, I limited the scope of this PR for the initial review, just to get some feedback. Although looking at the entity widget implementation, I see it should be pretty straightforward to add, so I'll look into as soon as I can.

bgoncal:
How often can a widget update in Android? In iOS the minimum auto-refresh is 15 minutes (besides other techniques like updating through push notification trigger)

mrdanielps:
It's actually quite similar on Android. The fastest auto-refresh is 30 minutes, but AFAIK anything that wakes up the app can manually trigger an update, like broadcasts or push notifications. Widget interactions can also update it, so pressing the button will allow for the state to be updated immediately.

dshokouhi:
The template widget updates instantly for me, there is no delay there. Our Template widget and entity state widget use a websocket subscription to update instantly.

bgoncal:
@dshokouhi so you keep a websocket connection all the time for the widget? 🪫😭
(We do that in iOS for local push too hahaha)

dshokouhi:
> so you keep a websocket connection all the time for the widget? 🪫😭

For widgets only when the screen turns on, then we stop the connection when the screen turns off. Its our best effort to keep them as up to as possible while not draining too much battery



> (We do that in iOS for local push too hahaha)

yup same for local push we never stop the connection and also send a ping to the server every 30 seconds to ensure its active and current.

mrdanielps:
> Looking great! Will the state of each item reflect in realtime? for example if the light is ON

Done!

[Screen_recording_20241111_005218.webm](https://github.com/user-attachments/assets/a98be49c-32d5-4c79-bc9e-69562f72454b)

common/src/main/res/values/strings.xml
```diff
@@ -1008,6 +1008,20 @@
...
     <string name="widget_creation_error">Unable to create widget.</string>
     <string name="widget_entity_fetch_error">Unable to fetch data for configured entity.</string>
+    <string name="grid_widget_actions">Actions</string>
+    <string name="widget_grid_add_action">Add action</string>
+    <string name="widget_grid_desc">A configurable grid of actions</string>
+    <string name="widget_grid_domain">Domain</string>
+    <string name="widget_grid_edit_action">Edit action</string>
+    <string name="widget_grid_entity">Entity</string>
+    <string name="widget_grid_entity_state_unknown">Unknown</string>
+    <string name="widget_grid_icon">Icon</string>
```
dshokouhi, line 1018: Some of these strings appear to be duplicated, we should re-use the same string if we have them to prevent duplicate translations.

common/build.gradle.kts
```diff
@@ -58,6 +58,7 @@ dependencies {
     implementation(libs.androidx.lifecycle.runtime.ktx)
     implementation(libs.androidx.lifecycle.viewmodel.ktx)
     implementation(libs.androidx.media)
+    api(libs.core.remoteviews)
```
dshokouhi, line 61: does this need to be included in the common module? I dont seem to see it used in any common module files

dshokouhi:
Gave this PR a quick test run. I think you have captured the obvious missing features like auto complete, ideally the behavior for each action should mimic the existing button widget with how fields are populated. I think that will match general user expectation. For example that widget handles when a custom integration provides incomplete action data and also when a user provides invalid action data. Looking forward to seeing this PR progress!

common/build.gradle.kts
mrdanielps: You're right, it's only actually needed in :app and :automotive. I've now moved the declaration to each module.

mrdanielps:
> ideally the behavior for each action should mimic the existing button widget with how fields are populated. I think that will match general user expectation. For example that widget handles when a custom integration provides incomplete action data and also when a user provides invalid action data.

I see, I was thinking of implementing it like the entity widget, which only takes an ID, but I'll certainly take that into account.

vlaraort:
There is no progress in this? 🫠

It's a really interesting improvement

dnestico:
Wish this could be merged....

vlaraort:
This is awesome! Also like the idea of the reload button to prevent excessive connections to check status. Can't wait!

dnestico:
Any update on this? Would love to see it merged.

mrdanielps:
I've rebased the branch and added tests, so the PR should now be ready for review. I've also updated the lint baseline to ignore some warnings relating to stability of lists in composables which should be safe. However, I can correct them if needed.

app/src/main/AndroidManifest.xml
```diff
@@ -254,6 +254,20 @@
                 android:resource="@xml/template_widget_info" />
         </receiver>
 
+        <receiver android:name=".widgets.grid.GridWidget" android:label="@string/widget_grid_label"
+            android:exported="false">
+            <intent-filter>
+                <action android:name="android.appwidget.action.APPWIDGET_UPDATE" />
+                <action android:name="io.homeassistant.companion.android.widgets.GridWidget.CALL_SERVICE" />
+                <action android:name="io.homeassistant.companion.android.widgets.GridWidget.CALL_SERVICE_AUTH" />
+                <action android:name="io.homeassistant.companion.android.widgets.GridWidget.RECEIVE_DATA" />
```
TimoPtr: I'm not sure you need this

app/lint-baseline.xml
```diff
@@ -2434,6 +2434,61 @@
             column="21"/>
     </issue>
 
+    <issue
```
TimoPtr: Let's address the issue not ignoring them.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+        verticalAlignment = Alignment.CenterVertically,
+        horizontalAlignment = Alignment.CenterHorizontally,
+        modifier = GlanceModifier.gridWidgetBackground().semantics { testTag = "LoadingScreen" },
+    ) {
+        CircularProgressIndicator(
+            color = GlanceTheme.colors.primary,
+            modifier = GlanceModifier.size(HomeAssistantGlanceTheme.dimensions.iconSize),
+        )
+    }
+}
```
TimoPtr, line 145: Let's avoid duplication this since it's a duplicate of the TODOWidget and create a shared GlanceLoadingScreen.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+            TitleBar(
+                startIcon = ImageProvider(R.drawable.ic_home_variant_outline),
+                title = label,
+                actions = {
+                    CircleIconButton(
+                        imageProvider = ImageProvider(R.drawable.ic_refresh),
+                        contentDescription = glanceStringResource(commonR.string.refresh),
+                        onClick = actionRefreshGrid(),
+                        backgroundColor = null,
+                    )
```
TimoPtr, line 160: Same as above you could reuse the same as the one in TODO widget, it would also be helpful to add the `sync_problem` state that is quite important, since today the WebSocket sometimes is not working well, it helps to know when the state is not up to date.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+import io.homeassistant.companion.android.util.compose.HomeAssistantGlanceTypography
+import io.homeassistant.companion.android.util.compose.glanceStringResource
+import io.homeassistant.companion.android.util.icondialog.getIconByMdiName
+import io.homeassistant.companion.android.widgets.grid.GridWidgetLayoutSize.Companion.gridCells
+import io.homeassistant.companion.android.widgets.grid.GridWidgetLayoutSize.Companion.showTitleBar
+import io.homeassistant.companion.android.widgets.grid.GridWidgetState.Companion.getButtonColors
+import io.homeassistant.companion.android.widgets.grid.GridWidgetState.Companion.getColors
+import kotlin.math.ceil
+
+class GridGlanceAppWidget : GlanceAppWidget() {
```
TimoPtr, line 89: Missing KDocs check the TodoGlanceAppWidget for inspiration.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+            label = it.label,
+            icon = it.icon,
+            state = it.state,
+            isActive = it.isActive,
+            buttonColors = state.getButtonColors(),
+            onClick = actionPressEntity(it.id),
+        )
+    }
+
+    val scaffoldTopPadding = if (showTitleBar()) 0.dp else 12.dp
```
TimoPtr:
```suggestion
    val scaffoldTopPadding = if (showTitleBar()) HADimens.SPACE0 else HADimens.SPACE3
```

TimoPtr: Keep `showTitleBar` into a variable to avoid computing this multiple times. You should also wrap it into a remember IMO to avoid computing this on each recomposition.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+                items = state.items,
+                itemContentProvider = buttonProvider(),
+            )
+        } else {
+            LazyButtonGrid(
+                modifier = GlanceModifier.fillMaxSize(),
+                gridCells = gridCells(),
+                items = state.items,
+                itemContentProvider = buttonProvider(),
+            )
```
TimoPtr:
```suggestion
        if (gridCells == 1) {
            LazyButtonColumn(
                modifier = GlanceModifier.fillMaxSize(),
                items = state.items,
                itemContentProvider = buttonProvider(),
            )
        } else {
            LazyButtonGrid(
                modifier = GlanceModifier.fillMaxSize(),
                gridCells = gridCells,
                items = state.items,
                itemContentProvider = buttonProvider(),
            )
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+
+    val scaffoldTopPadding = if (showTitleBar()) 0.dp else 12.dp
+
+    Scaffold(
+        modifier = modifier
+            .gridWidgetBackground()
+            .padding(top = scaffoldTopPadding),
+        titleBar = if (showTitleBar() && state.label != null) titleBar() else null,
+    ) {
+        val gridCells = gridCells()
```
TimoPtr: Move this outside the scaffold and remember it like `showTitleBar`.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+ */
+private enum class GridWidgetLayoutSize(val maxWidth: Dp) {
+    // Compact buttons, no title bar
+    Small(maxWidth = 180.dp),
+
+    // Icon and text
+    Normal(maxWidth = 439.dp),
+    ;
+
+    companion object {
```
TimoPtr, line 227: I don't think we need a companion object for this functions.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+
+        @Composable
+        fun showTitleBar(): Boolean {
+            return fromLocalSize() == Normal && LocalSize.current.height >= 180.dp
+        }
+
+        @Composable
+        fun gridCells(): Int {
+            val width = LocalSize.current.width
+            return when (width) {
```
TimoPtr:
```suggestion
            return when (val width = LocalSize.current.width) {
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+         */
+        @Composable
+        fun fromLocalSize(): GridWidgetLayoutSize {
+            val width = LocalSize.current.width
+            return if (width >= Small.maxWidth) Normal else Small
+        }
+
+        @Composable
+        fun showTitleBar(): Boolean {
+            return fromLocalSize() == Normal && LocalSize.current.height >= 180.dp
```
TimoPtr:
```suggestion
            return fromLocalSize() == Normal && LocalSize.current.height >= Small.maxWidth
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+        }
+
+        @Composable
+        fun gridCells(): Int {
+            val width = LocalSize.current.width
+            return when (width) {
+                in 0.dp..Small.maxWidth -> ceil(width / 96.dp).toInt()
+                else -> ceil(width / 260.dp).toInt()
```
TimoPtr, line 247: A lot of magic numbers in the file in general. We should extract this into const with proper name. Also if you have a reason to pick this specific number it would be nice to disclose it into a comment.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+private fun LazyButtonColumn(
+    items: List<GridButtonData>,
+    itemContentProvider: @Composable (item: GridButtonData) -> Unit,
+    modifier: GlanceModifier = GlanceModifier,
+) {
+    LazyColumn(
+        modifier = modifier,
+        horizontalAlignment = Alignment.CenterHorizontally,
+    ) {
+        items(items) {
```
TimoPtr, line 263: You should set the itemId lambda to help compose while recomposing.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+            Box(
+                modifier = GlanceModifier
+                    .fillMaxSize()
+                    .padding(bottom = 8.dp),
```
TimoPtr:
```suggestion
                    .padding(bottom = HADimens.SPACE2),
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+}
+
+@Composable
+private fun LazyButtonGrid(
+    gridCells: Int,
+    items: List<GridButtonData>,
+    itemContentProvider: @Composable (item: GridButtonData) -> Unit,
+    modifier: GlanceModifier = GlanceModifier,
+    horizontalAlignment: Alignment.Horizontal = Alignment.Start,
+    cellSpacing: Dp = 12.dp,
```
TimoPtr, line 282: Doesn't seem to be ever changed it should be in the function itself and use `HADimens.SPACE3`

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+    }
+}
+
+@Composable
+private fun LazyButtonGrid(
+    gridCells: Int,
+    items: List<GridButtonData>,
+    itemContentProvider: @Composable (item: GridButtonData) -> Unit,
+    modifier: GlanceModifier = GlanceModifier,
+    horizontalAlignment: Alignment.Horizontal = Alignment.Start,
```
TimoPtr, line 281: It doesn't seem to be ever changed if so remove it from the parameters.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+    cellSpacing: Dp = 12.dp,
+) {
+    val numRows = ceil(items.size.toDouble() / gridCells).toInt()
+
+    // Cell spacing is achieved by allocating equal amount of padding to each cell. Cells on edge
+    // apply it completely to inner sides, while cells not on edge apply it evenly on sides.
+    val perCellHorizontalPadding = (cellSpacing * (gridCells - 1)) / gridCells
+    val perCellVerticalPadding = (cellSpacing * (numRows - 1)) / numRows
```
TimoPtr, line 289: You should protect from potential division by zero the 3 operations. Especially because if one day gridcells return 0 for any reason it would break.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+
+    LazyVerticalGrid(
+        modifier = modifier,
+        gridCells = GridCells.Fixed(gridCells),
+        horizontalAlignment = horizontalAlignment,
+    ) {
+        itemsIndexed(items) { index, item ->
```
TimoPtr: Set the lambda `itemId`

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+                0 -> 0.dp
+                gridCells - 1 -> perCellHorizontalPadding
+                else -> perCellHorizontalPadding / 2
+            }
+
+            val cellEndPadding = when (column) {
+                0 -> perCellHorizontalPadding
+                gridCells - 1 -> 0.dp
+                else -> perCellHorizontalPadding / 2
+            }
```
TimoPtr, line 322: Mutualize the when to only have 2 and set the variables in it. Or do something like this

```
  private fun calculateEdgePadding(
      index: Int,
      totalCount: Int,
      totalPadding: Dp
  ): Pair<Dp, Dp> {
      return when (index) {
          0 -> 0.dp to totalPadding
          totalCount - 1 -> totalPadding to totalPadding
          else -> totalPadding / 2 to totalPadding / 2
      }
  }
...
      val (cellTopPadding, cellBottomPadding) = calculateEdgePadding(
          row,
          numRows,
          perCellVerticalPadding
      )

```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+
+@Composable
+private fun GridButton(
+    label: String,
+    onClick: Action,
+    modifier: GlanceModifier = GlanceModifier,
+    icon: String? = null,
+    state: String? = null,
+    isActive: Boolean = false,
+    buttonColors: GridButtonColors = GridButtonColors.Default,
```
TimoPtr: You probably don't need the default value for these parameters.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+) {
+    val size = GridWidgetLayoutSize.fromLocalSize()
+
+    val isCompact = size == GridWidgetLayoutSize.Small
+    val horizontalAlignment = if (isCompact) Alignment.CenterHorizontally else Alignment.Start
+    val baseModifier = if (!isCompact) modifier.height(64.dp) else modifier
+    val innerPadding = if (isCompact) 8.dp else 12.dp
```
TimoPtr:
```suggestion
    val baseModifier = if (!isCompact) modifier.height(HADimens.SPACE16) else modifier
    val innerPadding = if (isCompact) HADimens.SPACE2 else HADimens.SPACE3
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+    val backgroundColor = if (isActive) buttonColors.activeBackgroundColor else buttonColors.backgroundColor
+    val contentColor = if (isActive) buttonColors.activeContentColor else buttonColors.contentColor
+
+    Row(
+        modifier = baseModifier
+            .clickable(onClick)
+            .cornerRadius(16.dp)
```
TimoPtr:
```suggestion
            .cornerRadius(HARadius.XL)
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+}
+
+@Composable
+private fun rememberMdiIconProvider(name: String): ImageProvider {
+    val context = LocalContext.current
+    return remember(name) {
+        val bitmap = CommunityMaterial.getIconByMdiName(name)?.let {
+            IconicsDrawable(context, it).toBitmap()
+        }
+        bitmap?.let { ImageProvider(bitmap) } ?: ImageProvider(commonR.drawable.ic_priority)
```
TimoPtr:
```suggestion
        bitmap?.let(::ImageProvider) ?: ImageProvider(commonR.drawable.ic_priority)
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+
+@Composable
+private fun GridButtonIconFallback(name: String, modifier: GlanceModifier = GlanceModifier) {
+    Box(modifier, contentAlignment = Alignment.Center) {
+        Text(name.first().toString(), style = HomeAssistantGlanceTypography.titleLarge)
+    }
+}
+
+@Composable
+private fun rememberMdiIconProvider(name: String): ImageProvider {
```
TimoPtr, line 400: This is nice and should be extracted into a shared location so that future widget can use it.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+        colorFilter = colorFilter,
+    )
+}
+
+@Composable
+private fun GridButtonIconFallback(name: String, modifier: GlanceModifier = GlanceModifier) {
+    Box(modifier, contentAlignment = Alignment.Center) {
+        Text(name.first().toString(), style = HomeAssistantGlanceTypography.titleLarge)
+    }
+}
```
TimoPtr: Why use the first letter? and not a generic icon like `mdi:bookmark` that is used on the frontend?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+            .background(backgroundColor),
+        verticalAlignment = Alignment.CenterVertically,
+        horizontalAlignment = horizontalAlignment,
+    ) {
+        icon?.let {
+            GridButtonIcon(icon, colorFilter = ColorFilter.tint(contentColor))
+        } ?: GridButtonIconFallback(label)
+
+        if (!isCompact) {
+            Spacer(GlanceModifier.width(8.dp))
```
TimoPtr:
```suggestion
            Spacer(GlanceModifier.width(HADimens.SPACE2))
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+            .clickable(onClick)
+            .cornerRadius(16.dp)
+            .padding(innerPadding)
+            .background(backgroundColor),
+        verticalAlignment = Alignment.CenterVertically,
+        horizontalAlignment = horizontalAlignment,
+    ) {
+        icon?.let {
+            GridButtonIcon(icon, colorFilter = ColorFilter.tint(contentColor))
+        } ?: GridButtonIconFallback(label)
```
TimoPtr: We should make sure that whatever happens the size is always going to be 24dp (HADimens.SPACE6). It should probably be set here and not in each individual Composable, for instance the Fallback today is not 24.dp.

Also I would like of the preview to display the fallback.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+                Text(
+                    text = label,
+                    style = HomeAssistantGlanceTypography.bodyLarge.copy(color = contentColor),
+                    maxLines = 1,
+                )
+                Text(
+                    text = state ?: "—",
+                    style = HomeAssistantGlanceTypography.bodySmall.copy(color = contentColor),
+                    maxLines = 1,
+                )
```
TimoPtr, line 382: Subtitle are cut 
image

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+            }
+        }
+    }
+}
+
+@Composable
+private fun GridButtonIcon(icon: String, modifier: GlanceModifier = GlanceModifier, colorFilter: ColorFilter? = null) {
```
TimoPtr:
```suggestion
private fun GridButtonIcon(icon: String, colorFilter: ColorFilter, modifier: GlanceModifier = GlanceModifier) {
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
```diff
@@ -0,0 +1,419 @@
...
+@Preview(widthDp = 259, heightDp = 200)
+@Preview(widthDp = 259, heightDp = 120)
+@Preview(widthDp = 438, heightDp = 200)
+@Preview(widthDp = 644, heightDp = 200)
+private annotation class GridWidgetBreakpointPreviews
+
+@OptIn(ExperimentalGlancePreviewApi::class)
+@Composable
+@GridWidgetBreakpointPreviews
+private fun DynamicGridWidgetPreview() {
```
TimoPtr, line 421: Add previews for other states like error, empty and loading. And data with no mdi icon

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidget.kt
```diff
@@ -0,0 +1,20 @@
...
+import android.content.Context
+import androidx.glance.appwidget.GlanceAppWidget
+import dagger.hilt.android.AndroidEntryPoint
+import io.homeassistant.companion.android.database.widget.GridWidgetDao
+import io.homeassistant.companion.android.database.widget.GridWidgetEntity
+import io.homeassistant.companion.android.widgets.BaseGlanceEntityWidgetReceiver
+import io.homeassistant.companion.android.widgets.EntitiesPerServer
+
+@AndroidEntryPoint
+class GridWidget : BaseGlanceEntityWidgetReceiver<GridWidgetEntity, GridWidgetDao>() {
```
TimoPtr, line 21: Missing KDocs

app/src/test/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidgetTest.kt
```diff
@@ -0,0 +1,55 @@
...
+    @Test
+    fun `Given entries in DAO when invoking getWidgetEntitiesByServer then returns properly mapped items`() = runTest {
+        val context = mockk<Context>()
+
+        coEvery { dao.getAll() } returns listOf(
+            GridWidgetEntity(
+                id = 1,
+                serverId = 42,
+                label = null,
+                items = listOf(GridWidgetEntity.Item(1, "entity1", "Label", "Icon")),
```
TimoPtr, line 38: You should add multiple items to test that the IDs are properly read from the DAO. You could also add an Entity with an emptyList to see what is the output.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidgetActions.kt
```diff
@@ -0,0 +1,112 @@
...
+
+@VisibleForTesting
+internal val ENTITY_ID_KEY = ActionParameters.Key<String>("ENTITY_ID_KEY")
+
+/**
+ * Action that will toggle the given Entity through a given parameters with the key [ENTITY_ID_KEY]. Use [actionPressEntity] to get the
+ * Action for Glance.
+ *
+ * The action call the server to toggle the item. On Success the widget is updated and the state will be updated
+ * through the [GridWidgetStateUpdater]. On Failure it will show a toast with the failure message.
```
TimoPtr: No there are no toast currently. Double check the doc.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidgetActions.kt
```diff
@@ -0,0 +1,112 @@
...
+    override suspend fun onAction(context: Context, glanceId: GlanceId, parameters: ActionParameters) {
+        GridGlanceAppWidget().update(context, glanceId)
+    }
+}
+
+@VisibleForTesting
+internal val ENTITY_ID_KEY = ActionParameters.Key<String>("ENTITY_ID_KEY")
+
+/**
+ * Action that will toggle the given Entity through a given parameters with the key [ENTITY_ID_KEY]. Use [actionPressEntity] to get the
```
TimoPtr: It's not a toggle it's a press event.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidgetState.kt
```diff
@@ -0,0 +1,94 @@
...
+        } else {
+            WidgetBackgroundType.DAYNIGHT
+        }
+    val textColor: String?
+        get() = null
+
+    companion object {
+        @Composable
+        fun GridWidgetState.getColors(): ColorProviders {
+            return when (backgroundType) {
```
TimoPtr, line 36: This is a duplicate of the TODOWidget we probably can do something to avoid it. I can do it later if you don't want to do it here. In that case please create an issue.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidgetState.kt
```diff
@@ -0,0 +1,94 @@
...
+                )
+                else -> GridButtonColors.Default
+            }
+        }
+    }
+}
+
+internal object LoadingGridState : GridWidgetState
+
+internal data class GridStateWithData(val label: String? = null, val items: List<GridButtonData> = emptyList()) :
```
TimoPtr:
```suggestion
internal data class GridStateWithData(val label: String?, val items: List<GridButtonData>) :
```

This is something that is always set.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidgetStateUpdater.kt
```diff
@@ -0,0 +1,79 @@
...
+import kotlinx.coroutines.flow.onStart
+import kotlinx.coroutines.flow.transformLatest
+import timber.log.Timber
+
+internal class GridWidgetStateUpdater @Inject constructor(
+    private val gridWidgetDao: GridWidgetDao,
+    private val serverManager: ServerManager,
+    @param:ApplicationContext private val context: Context,
+) {
+    fun stateFlow(widgetId: Int): Flow<GridWidgetState> = gridWidgetDao.observe(widgetId).withEntityState()
```
TimoPtr:
```suggestion
    fun stateFlow(widgetId: Int): Flow<GridWidgetState> = gridWidgetDao.observe(widgetId).withEntityState().distinctUntilChanged(
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidgetStateUpdater.kt
```diff
@@ -0,0 +1,79 @@
...
+    private fun Flow<GridWidgetEntity>.withEntityState(): Flow<GridStateWithData> = transformLatest { widgetEntity ->
+        // Use transformLatest to cancel previous entity fetches if widgetEntity changes quickly
+        var currentUiState = widgetEntity.toUiState()
+        emit(currentUiState)
+
+        val entityIds = widgetEntity.items.map(GridWidgetEntity.Item::entityId)
+        val entityUpdates = getAndObserveEntities(widgetEntity.serverId, entityIds)
+
+        entityUpdates?.collect { updatedEntity ->
+            currentUiState = widgetEntity.toUiState().copy(
```
TimoPtr: widgetEntity.toUiState() is duplicated here simply use the currentUiState.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidgetStateUpdater.kt
```diff
@@ -0,0 +1,79 @@
...
+                items = currentUiState.items.map { item ->
```
TimoPtr: Instead of mapping all the item. Just find the item that has change in the currentUiState.items and modify the list (.toMutable) with the new value.

Something like that should do it with an immutable collection

```kotlin
val itemIndex = currentUiState.items.indexOfFirst { it.id == updatedEntity.entityId }

          if (itemIndex != -1) {
              currentUiState = currentUiState.copy(
                  items = currentUiState.items.mutate { list ->
                      list[itemIndex] = list[itemIndex].copy(
                          state = updatedEntity.friendlyState(context, appendUnitOfMeasurement = true),
                          isActive = updatedEntity.isActive(),
                      )
                  }
              )
              emit(currentUiState)
          }
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidgetUtils.kt
```diff
@@ -0,0 +1,31 @@
...
+    serverId = serverId,
+    label = label,
+    items = items.map(GridWidgetEntity.Item::asGridItem),
+)
+
+fun GridWidgetEntity.Item.asGridItem() = GridItem(
+    label = label,
+    icon = iconName,
+    entityId = entityId,
+)
```
TimoPtr:
```suggestion
internal fun GridConfiguration.asDbEntity(widgetId: Int) = GridWidgetEntity(
    id = widgetId,
    serverId = serverId,
    label = label,
    items = items.map { it.asDbEntity(widgetId) },
)

internal fun GridItem.asDbEntity(widgetId: Int) = GridWidgetEntity.Item(
    gridId = widgetId,
    entityId = entityId,
    label = label,
    iconName = icon,
)

internal fun GridWidgetEntity.asGridConfiguration() = GridConfiguration(
    serverId = serverId,
    label = label,
    items = items.map(GridWidgetEntity.Item::asGridItem),
)

private fun GridWidgetEntity.Item.asGridItem() = GridItem(
    label = label,
    icon = iconName,
    entityId = entityId,
)
```

Even if for now it is useless since we are at the :app level this would help in a future refactor to properly set the visibilities.

TimoPtr: It's also a good idea to add some dumb tests to verify the mapping is working as intended.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+import io.homeassistant.companion.android.util.icondialog.getIconByMdiName
+import io.homeassistant.companion.android.util.previewEntity1
+import io.homeassistant.companion.android.util.previewEntity2
+import io.homeassistant.companion.android.util.previewServer1
+import io.homeassistant.companion.android.util.previewServer2
+import io.homeassistant.companion.android.util.safeBottomWindowInsets
+import io.homeassistant.companion.android.util.safeTopWindowInsets
+
+@Composable
+fun GridWidgetConfigurationScreen(viewModel: GridConfigurationViewModel, onAddWidget: (GridConfiguration) -> Unit) {
```
TimoPtr: Check `AndroidAutoFavoritesSettings` We should also offer the reordering capably here. It shouldn't be hard to reuse the `ReorderableItem`

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+    val servers by viewModel.servers.collectAsStateWithLifecycle()
+    val entities by viewModel.entities.collectAsStateWithLifecycle()
+    val config by viewModel.gridConfig.collectAsStateWithLifecycle()
+
+    GridWidgetConfigurationContent(
+        servers = servers,
+        selectedServerId = config.serverId,
+        onServerSelected = viewModel::setServer,
+        entities = entities,
+        onConfigure = onAddWidget,
```
TimoPtr: onConfigure or onAddWidget? also it could be update

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+
+@Composable
+fun GridWidgetConfigurationScreen(viewModel: GridConfigurationViewModel, onAddWidget: (GridConfiguration) -> Unit) {
+    val servers by viewModel.servers.collectAsStateWithLifecycle()
+    val entities by viewModel.entities.collectAsStateWithLifecycle()
+    val config by viewModel.gridConfig.collectAsStateWithLifecycle()
+
+    GridWidgetConfigurationContent(
+        servers = servers,
+        selectedServerId = config.serverId,
```
TimoPtr: You are passing the whole config down already, you probably can remove this param

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+    Scaffold(
+        topBar = {
+            TopAppBar(
+                title = { Text(stringResource(commonR.string.widget_grid_title)) },
+                windowInsets = safeTopWindowInsets(),
+                backgroundColor = colorResource(commonR.color.colorBackground),
+                contentColor = colorResource(commonR.color.colorOnBackground),
+            )
+        },
+    ) { padding ->
```
TimoPtr, line 98: Let's exrtract the content into a dedicated function to make it shorter. A rule I try to follow is that it should fix on my screen without scrolling.

(I know that I'm not always respecting this rule, so if you have trouble with naming just skip this comment)

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+        Column(
+            modifier = Modifier.padding(16.dp),
+            verticalArrangement = Arrangement.spacedBy(8.dp),
+        ) {
+            title?.let { SectionHeader(title) }
+            content()
+        }
+
+        Divider(Modifier.fillMaxWidth())
+    }
```
TimoPtr, line 302: You should be able to remove one Column here

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+                }
+            }
+
+            PreferenceSection(
+                modifier = Modifier
+                    .verticalScroll(rememberScrollState())
+                    .weight(1f),
+                title = stringResource(commonR.string.widget_grid_add_entities),
+            ) {
+                config.items.forEachIndexed { i, item ->
```
TimoPtr: You could use a LazyColumnList here

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+                    val iconPainter = remember(item.icon) {
+                        CommunityMaterial.getIconByMdiName(item.icon)?.let {
+                            IconicsPainter(it)
+                        }
+                    }
+                    EntityEditorRow(
```
TimoPtr: FYI I'm redoing the Entity picker so it looks nicer and handle the icon directly within it.

here a sneak peek of how it's going to look. The first version is going to be for single entity but it could be updated to support multiple.

https://github.com/user-attachments/assets/49b85c99-f25c-414a-935a-137d2f8132a0

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+                        entityId = item.entityId,
+                        entities = entities,
+                        icon = iconPainter?.let { { Icon(it, null) } },
+                        onSelect = { selected ->
+                            val entity = entities.first { it.entityId == selected }
```
TimoPtr: Let's be a bit safer here and use firstOrNull (check the rest of the file too), and add a warning log when it happens even if it shouldn't.

You could also use `FailFast` instead of a simply log.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
TimoPtr:
```suggestion
                config.items.forEachIndexed { index, item ->
```

It's clearer and we are not limited in chars so let's use a meaningful name for variables.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+                        false
+                    },
+                    onDelete = {},
+                )
+            }
+
+            Button(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(16.dp),
```
TimoPtr:
```suggestion
                    .padding(HADimens.SPACE4),
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+@Composable
+private fun PreferenceSection(
+    modifier: Modifier = Modifier,
+    title: String? = null,
+    content: @Composable ColumnScope.() -> Unit,
+) {
+    Column(modifier) {
+        Column(
+            modifier = Modifier.padding(16.dp),
+            verticalArrangement = Arrangement.spacedBy(8.dp),
```
TimoPtr:
```suggestion
            modifier = Modifier.padding(HADimens.SPACE4),
            verticalArrangement = Arrangement.spacedBy(HADimens.SPACE2),
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        Divider(Modifier.fillMaxWidth())
+    }
+}
+
+@Composable
+private fun SectionHeader(title: String, modifier: Modifier = Modifier) {
+    Text(
+        modifier = modifier.padding(start = 48.dp),
```
TimoPtr:
```suggestion
        modifier = modifier.padding(start = HADimens.SPACE12),
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+@Composable
+private fun GridConfigurationPreview() {
+    HomeAssistantAppTheme {
+        GridWidgetConfigurationContent(
+            config = GridConfiguration(
+                label = "Home lights",
+                items = listOf(
+                    GridItem("Bedroom", "mdi:lightbulb"),
+                    GridItem("Bedroom", "mdi:lightbulb"),
+                    GridItem("Living room", "mdi:lightbulb"),
```
TimoPtr, line 326: Add an item without icon

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
```diff
@@ -0,0 +1,256 @@
...
+                    },
+                )
+
+                if (servers.size > 1) {
+                    ServerExposedDropdownMenu(
+                        servers = servers,
+                        current = selectedServerId,
+                        onSelected = { onServerSelected(it) },
+                    )
+                }
```
TimoPtr: You should let the user pick the theme they want to use.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationViewModel.kt
```diff
@@ -0,0 +1,164 @@
...
+    @OptIn(ExperimentalCoroutinesApi::class)
+    val entities: StateFlow<List<Entity>> = gridConfig
+        .map { it.serverId }
+        .distinctUntilChanged()
+        .filterNotNull()
+        .mapLatest { serverId ->
+            if (serverManager.isRegistered()) {
+                serverManager.integrationRepository(serverId)
+                    .getEntities()
+                    .orEmpty()
```
TimoPtr, line 69: Do we accept all kind of entities?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationViewModel.kt
```diff
@@ -0,0 +1,164 @@
...
+            } else {
+                Timber.w("No server registered")
+                emptyList()
+            }
+        }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(500.milliseconds), emptyList())
+
+    fun onSetup(widgetId: Int) {
+        if (this.widgetId == AppWidgetManager.INVALID_APPWIDGET_ID) {
+            loadPreviousState(widgetId)
```
TimoPtr, line 79: If you look at the TodoConfigureViewModel we do use a Mutex to ensure that the thread safety, we probably should apply some here too. I didn't look into all the details of the VM here, but most probably you have multiple sources touching the same thing, you want to avoid race condition as much as possible.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationViewModel.kt
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    private fun loadPreviousState(widgetId: Int) = viewModelScope.launch {
+        gridWidgetDao.get(widgetId)?.let { dao ->
+            _gridConfig.update { dao.asGridConfiguration() }
+        }
+    }
+
+    fun setServer(serverId: Int) = _gridConfig.update { currentConfig ->
+        currentConfig.copy(
```
TimoPtr: You could simply check if the sever is the same in that case return the currentConfig instead of making a new copy?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationViewModel.kt
```diff
@@ -0,0 +1,164 @@
...
+
+    fun addItem(entity: Entity) = _gridConfig.update { it.copy(items = it.items + entity.asGridItem()) }
+
+    fun editItem(i: Int, entity: Entity) = _gridConfig.update {
+        it.copy(items = it.items.mapIndexed { index, it -> if (index == i) entity.asGridItem() else it })
+    }
+
+    fun deleteItem(i: Int) = _gridConfig.update {
+        it.copy(items = it.items.filterIndexed { index, _ -> index != i })
+    }
```
TimoPtr: Avoid using `i` 🙏🏻 use something like `selectedIndex`

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationViewModel.kt
```diff
@@ -0,0 +1,164 @@
...
+
+    suspend fun updateWidgetConfiguration(config: GridConfiguration) {
```
TimoPtr: Let's return false instead of an error since you are not using the error message.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridWidgetConfigureActivity.kt
```diff
@@ -0,0 +1,94 @@
...
+import io.homeassistant.companion.android.settings.widgets.ManageWidgetsViewModel
+import io.homeassistant.companion.android.util.compose.HomeAssistantAppTheme
+import kotlinx.coroutines.launch
+
+@AndroidEntryPoint
+class GridWidgetConfigureActivity : BaseActivity() {
+    private val viewModel: GridConfigurationViewModel by viewModels()
+
+    public override fun onCreate(savedInstanceState: Bundle?) {
+        enableEdgeToEdge()
```
TimoPtr:
```suggestion
```

Not needed it's in BaseActivity now.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridWidgetConfigureActivity.kt
```diff
@@ -0,0 +1,94 @@
...
+import androidx.lifecycle.lifecycleScope
+import dagger.hilt.android.AndroidEntryPoint
+import io.homeassistant.companion.android.BaseActivity
+import io.homeassistant.companion.android.common.R
+import io.homeassistant.companion.android.settings.widgets.ManageWidgetsViewModel
+import io.homeassistant.companion.android.util.compose.HomeAssistantAppTheme
+import kotlinx.coroutines.launch
+
+@AndroidEntryPoint
+class GridWidgetConfigureActivity : BaseActivity() {
```
TimoPtr, line 20: You should be able to start this activity from the Settings, like all the other widgets.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridWidgetConfigureActivity.kt
```diff
@@ -0,0 +1,94 @@
...
+            AppWidgetManager.INVALID_APPWIDGET_ID,
+        ) ?: AppWidgetManager.INVALID_APPWIDGET_ID
+
+        viewModel.onSetup(widgetId)
+
+        setContent {
+            HomeAssistantAppTheme {
+                GridWidgetConfigurationScreen(
+                    viewModel = viewModel,
+                    onAddWidget = { onSetupWidget(it) },
```
TimoPtr: Match the name of the lambda

app/src/main/res/drawable-nodpi/widget_example_grid.png
TimoPtr, line 1: Try to match the color of existing widgets like TODO

image

app/src/main/res/xml/grid_widget_info.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<appwidget-provider xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:minWidth="40dp"
+    android:minHeight="40dp"
```
TimoPtr, line 5: I've played a bit with the minimal version it's IMO too small, but maybe it's fine for some users
image

I would personally in the smallest version only allow one item to be displayed and make it scrollable maybe.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidgetPressActionTest.kt
```diff
@@ -0,0 +1,77 @@
...
+
+        // GridGlanceAppWidget.update() will throw IllegalArgumentException when called, so use it to verify onAction
+        val exception = assertThrows(IllegalArgumentException::class.java) {
+            runTest {
+                action.onAction(mockk(), FakeGlanceId(widgetId), parameters)
+            }
+        }
+        assertEquals("Invalid Glance ID", exception.message)
+
+        coVerify { integrationRepository.callAction(any(), any(), any()) }
```
TimoPtr: You want to assert that it calls the right action with the right param.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidgetStateTest.kt
```diff
@@ -0,0 +1,29 @@
+package io.homeassistant.companion.android.widgets.grid
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+
+class GridWidgetStateTest {
+
+    @Test
+    fun `Given GridStateWithData with items when created then properties are correct`() {
```
TimoPtr: You are testing the constructor which does nothing I don't think this test is relevant

app/src/test/kotlin/io/homeassistant/companion/android/widgets/grid/GridWidgetStateUpdaterTest.kt
```diff
@@ -0,0 +1,81 @@
...
+import io.mockk.mockk
+import java.time.LocalDateTime
+import kotlinx.coroutines.flow.emptyFlow
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+
+class GridWidgetStateUpdaterTest {
```
TimoPtr, line 20: I would like some test 
- to test the error scenarios.
- test that verify that we are not updating more than we need when things doesn't change and verify the number of interaction we make with the integrationRepo to avoid making call that we shouldn't.
- Making multiple update of the same entity is making the right amount of call on the integrationRepo (to verify that the collect within the transformLatest is properly canceled it's important if one day we change the operator to have this test).

automotive/lint-baseline.xml
```diff
@@ -2566,6 +2573,17 @@
...
+        id="UnusedResources"
+        message="The resource `R.xml.grid_widget_info` appears to be unused"
+        errorLine1="&lt;appwidget-provider xmlns:android=&quot;http://schemas.android.com/apk/res/android&quot;"
+        errorLine2="^">
+        <location
+            file="${:automotive*fullDebug*MAIN*sourceProvider*0*resDir*1}/xml/grid_widget_info.xml"
+            line="2"
+            column="1"/>
+    </issue>
+
```
TimoPtr: You shouldn't have this here. Could you revert the changes it should be used, like the other widgets.

common/src/main/kotlin/io/homeassistant/companion/android/database/AppDatabase.kt
```diff
@@ -103,7 +106,7 @@ import timber.log.Timber
         Server::class,
         Setting::class,
     ],
-    version = 51,
+    version = 52,
```
TimoPtr, line 80: I'm missing the JSON of the new version ⚠️ it needs to be commit.

common/src/main/kotlin/io/homeassistant/companion/android/database/widget/GridWidgetDao.kt
```diff
@@ -0,0 +1,38 @@
...
+
+@Dao
+@TypeConverters(GridWidgetItemConverter::class)
+interface GridWidgetDao : WidgetDao<GridWidgetEntity> {
+
+    @Query("SELECT * FROM grid_widgets WHERE id = :id")
+    suspend fun get(id: Int): GridWidgetEntity?
+
+    @Query("SELECT * FROM grid_widgets WHERE id = :id")
+    fun observe(id: Int): Flow<GridWidgetEntity>
```
TimoPtr:
```suggestion
    fun getFlow(id: Int): Flow<GridWidgetEntity>
```
observe is wrong, when you invoke this function until you collect we are not observing anything.

common/src/main/kotlin/io/homeassistant/companion/android/database/widget/GridWidgetDao.kt
```diff
@@ -0,0 +1,38 @@
...
+    override suspend fun delete(id: Int)
+
+    @Query("DELETE FROM grid_widgets WHERE id IN (:ids)")
+    override suspend fun deleteAll(ids: IntArray)
+
+    @Query("SELECT * FROM grid_widgets")
+    suspend fun getAll(): List<GridWidgetEntity>
+
+    @Query("SELECT * FROM grid_widgets")
+    fun observeAll(): Flow<List<GridWidgetEntity>>
```
TimoPtr:
```suggestion
```
Not used

common/src/main/kotlin/io/homeassistant/companion/android/database/widget/GridWidgetEntity.kt
```diff
@@ -0,0 +1,28 @@
...
+data class GridWidgetEntity(
+    @PrimaryKey
+    override val id: Int,
+    @ColumnInfo(name = "server_id", defaultValue = "0")
+    override val serverId: Int,
+    @ColumnInfo(name = "label")
+    val label: String?,
+    @ColumnInfo(name = "items")
+    val items: List<Item>,
+) : WidgetEntity<GridWidgetEntity> {
```
TimoPtr:
```suggestion
) : WidgetEntity<GridWidgetEntity>, ThemeableWidgetEntity {
```

common/src/main/res/values/strings.xml
```diff
@@ -983,6 +983,14 @@
     <string name="widget_config_action_error">A custom component is preventing action data from loading.</string>
     <string name="widget_creation_error">Unable to create widget.</string>
     <string name="widget_entity_fetch_error">Unable to fetch data for configured entity.</string>
+    <string name="grid_widget_actions">Actions</string>
```
TimoPtr:
```suggestion
```
I think it's not used

common/src/main/res/values/strings.xml
```diff
@@ -983,6 +983,14 @@
...
+    <string name="widget_grid_add_entities">Add entities</string>
+    <string name="widget_grid_desc">A configurable grid of actions</string>
+    <string name="widget_grid_edit_action">Edit action</string>
```
TimoPtr:
```suggestion
```
same

common/src/main/res/values/strings.xml
```diff
@@ -983,6 +983,14 @@
...
+    <string name="widget_grid_label">Action grid</string>
+    <string name="widget_grid_name">Name</string>
+    <string name="widget_grid_title">Grid widget</string>
+    <string name="widget_image_description">Home Assistant widget</string>
```
TimoPtr:
```suggestion
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
mrdanielps: Neither ;)
I've renamed it to onSubmit, which I think describes the composable action better. Update could be confusing because of similar function names in the activity and the parent class.

automotive/lint-baseline.xml
mrdanielps: It's not referenced in the automotive manifest, so I'm guessing that's why it complains about the unused resource.
I can add it to the grid_widget_info.xml like in the TODO widget:
```xml
tools:ignore="UnusedAttribute,UnusedResources"
```
I wonder if auto can benefit from this widget though.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationViewModel.kt
mrdanielps: Should we filter by pressable or toggleable? I can see the state entities being used in the grid, although it is a bit strange to see them represented as buttons.
image

TimoPtr: Until we make it possible to attach an action to an element (another PR 🙏🏻 ) we could simply display them a bit differently that makes it clear that it is not clickable.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/config/GridConfigurationScreen.kt
TimoPtr: If you want to take a look https://github.com/home-assistant/android/pull/6253

app/src/main/kotlin/io/homeassistant/companion/android/widgets/grid/GridGlanceAppWidget.kt
mrdanielps: showTitleBar and and gridCells are composable, so they can't be `remember`ed (they actually need to be recomposed when the composition changes). I'll move it to a variable so it's only calculated once on recomposition.

TimoPtr:
@mrdanielps the new entity picker has been implemented and we've replaced the old one with the new one in https://github.com/home-assistant/android/pull/6293 You can inspire yourself from the AndroidAutoFavorites for selecting the entities especially because it has the capability of reordering items which IMO would make your widget configuration even better.

Let's make sure your screen uses Material3 and the reusable components from https://github.com/home-assistant/android/tree/6b497e66400ed9f0f69466aa4d4ffb078962ebe4/common/src/main/kotlin/io/homeassistant/companion/android/common/compose

PR #4784, opened 2024-11-04
Add heart rate sensor for Health Connect

dshokouhi:
## Summary

Adds heart rate sensor from Health Connect

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#1133

## Any other notes

Adding 1 sensor for now as this was tested earlier and probably a good idea to check the new flow after adding 1 sensor. I assume the action will fail and the declaration needs to be updated and submitted with the APK uploaded again. If that is the case then we know larger PRs are necessary to ease the approval process.


jpelgrom (approved):
Code looks like other sensors (good) to me and in a quick test it seemed to work.

We may want to introduce a helper function for these attributes in the future to make sure they are set consistently.

Let's see what Google does when we add a new permission :)

PR #4789, opened 2024-11-07
Fix some sensor configurations to be more precise

dshokouhi:
## Summary

- Add duration device class to Sleep segment sensor
- Add unit and device class to screen timeout as it is duration
- Add total increasing state class to all health services sensors that reset at midnight 


jpelgrom (approved):
Nice to align these with the upcoming similar Health Connect sensors :)

PR #4790, opened 2024-11-08
Make changes to allow health connect to work on Android 9 and higher

dshokouhi:
## Summary

Makes `requestSensorUpdate` and `checkPermissions` suspend functions to allow health connect to work on Android 9 and higher devices instead of just Android 14+. Some other methods were also converted to make the changes here easier to follow (I hope)

Documentation: home-assistant/companion.home-assistant#1135

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
@@ -71,31 +71,32 @@ class AndroidAutoSensorManager : SensorManager, Observer<Int> {
         sensorWorkerScope.launch {
             if (!isEnabled(context, androidAutoConnected)) {
                 carConnection?.type?.removeObserver(this@AndroidAutoSensorManager)
+                return@launch
             }
-            return@launch
-        }
-        val (connected, typeString) = when (value) {
-            CarConnection.CONNECTION_TYPE_NOT_CONNECTED -> {
-                false to "Disconnected"
+
+            val (connected, typeString) = when (value) {
+                CarConnection.CONNECTION_TYPE_NOT_CONNECTED -> {
+                    false to "Disconnected"
+                }
+                CarConnection.CONNECTION_TYPE_PROJECTION -> {
+                    true to "Projection"
+                }
+                CarConnection.CONNECTION_TYPE_NATIVE -> {
+                    true to "Native"
+                }
+                else -> {
+                    false to "Unknown($value)"
+                }
             }
-            CarConnection.CONNECTION_TYPE_PROJECTION -> {
-                true to "Projection"
-            }
-            CarConnection.CONNECTION_TYPE_NATIVE -> {
-                true to "Native"
-            }
-            else -> {
-                false to "Unknown($value)"
-            }
-        }
-        onSensorUpdated(
-            context,
-            androidAutoConnected,
-            connected,
-            if (connected) androidAutoConnected.statelessIcon else "mdi:car-off",
-            mapOf(
-                "connection_type" to typeString
+            onSensorUpdated(
+                context,
+                androidAutoConnected,
+                connected,
+                if (connected) androidAutoConnected.statelessIcon else "mdi:car-off",
+                mapOf(
+                    "connection_type" to typeString
+                )
             )
-        )
+        }
     }
 }
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
@@ -17,6 +17,7 @@ class AndroidAutoSensorManager : SensorManager, Observer<Int> {
     companion object {
 
         internal const val TAG = "AndroidAutoSM"
+        private var enabled = false
 
         private val androidAutoConnected = SensorManager.BasicSensor(
             "android_auto",
@@ -51,7 +52,7 @@ class AndroidAutoSensorManager : SensorManager, Observer<Int> {
     private lateinit var context: Context
     private var carConnection: CarConnection? = null
 
-    override fun requestSensorUpdate(context: Context) {
+    override suspend fun requestSensorUpdate(context: Context) {
         this.context = context.applicationContext
         if (!isEnabled(context, androidAutoConnected)) {
             return
@@ -70,11 +71,11 @@ class AndroidAutoSensorManager : SensorManager, Observer<Int> {
     }
 
     override fun onChanged(value: Int) {
-        if (!isEnabled(context, androidAutoConnected)) {
-            CoroutineScope(Dispatchers.Main + Job()).launch {
+        CoroutineScope(Dispatchers.Main + Job()).launch {
+            if (!isEnabled(context, androidAutoConnected)) {
                 carConnection?.type?.removeObserver(this@AndroidAutoSensorManager)
             }
-            return
+            return@launch
         }
         val (connected, typeString) = when (value) {
             CarConnection.CONNECTION_TYPE_NOT_CONNECTED -> {
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
@@ -7,9 +7,6 @@ import androidx.car.app.connection.CarConnection
 import androidx.lifecycle.Observer
 import io.homeassistant.companion.android.common.R as commonR
 import io.homeassistant.companion.android.common.sensors.SensorManager
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 
 class AndroidAutoSensorManager : SensorManager, Observer<Int> {
@@ -57,7 +54,7 @@ class AndroidAutoSensorManager : SensorManager, Observer<Int> {
         if (!isEnabled(context, androidAutoConnected)) {
             return
         }
-        CoroutineScope(Dispatchers.Main + Job()).launch {
+        sensorWorkerScope.launch {
             if (carConnection == null) {
                 carConnection = try {
                     CarConnection(context.applicationContext)
@@ -71,7 +68,7 @@ class AndroidAutoSensorManager : SensorManager, Observer<Int> {
     }
 
     override fun onChanged(value: Int) {
-        CoroutineScope(Dispatchers.Main + Job()).launch {
+        sensorWorkerScope.launch {
             if (!isEnabled(context, androidAutoConnected)) {
                 carConnection?.type?.removeObserver(this@AndroidAutoSensorManager)
             }
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
@@ -39,7 +39,6 @@ import java.time.temporal.ChronoUnit
 import kotlin.reflect.KClass
 import kotlin.time.DurationUnit
 import kotlin.time.toDuration
-import kotlinx.coroutines.runBlocking
 
 class HealthConnectSensorManager : SensorManager {
     companion object {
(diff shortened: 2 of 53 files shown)


jpelgrom:
Broadcast receivers are short lived and as I understand it may be killed even earlier if you launch (run work asynchronously) without indicating it to the system. Have you tested/encountered this with the app running in the background? Some nice discussions and a Kotlin extension function for it here: https://stackoverflow.com/questions/22741202/how-to-use-goasync-for-broadcastreceiver

dshokouhi:
> Have you tested/encountered this with the app running in the background?

So have not enabled all sensors but so far working well about 7 hours in

image

image

image

health connect

image

dshokouhi:
device tracker also updating regularly during this time

image

image

dshokouhi:
24 hours later all enabled sensors including location tracking still working as expected

jpelgrom (commented):
You're using `MainScope().launch { }` in a couple of places, but also a regular `CoroutineScope` in another one or two. Would it be worth it to introduce a helper function for it, or a custom scope like `sensorWorkerScope` (which is defined in the base class as `= MainScope()`) so it is easier to track and update later?

dshokouhi:
> You're using `MainScope().launch { }` in a couple of places, but also a regular `CoroutineScope` in another one or two.

The reason for that is because we had it split up. Specifically the android auto connection sensors requires the CoroutineScope or it fails to update.



> Would it be worth it to introduce a helper function for it, or a custom scope like `sensorWorkerScope` (which is defined in the base class as `= MainScope()`) so it is easier to track and update later?

if we were to do this it would need to be teh CoroutineScope

jpelgrom:
> Specifically the android auto connection sensors requires the CoroutineScope or it fails to update

That's odd because you're still using the main/UI thread, only difference is the supervisor behavior which shouldn't matter.

dshokouhi:
> > Specifically the android auto connection sensors requires the CoroutineScope or it fails to update
> 
> That's odd because you're still using the main/UI thread, only difference is the supervisor behavior which shouldn't matter.

was failing due to the thread not calling `Looper.prepare()` dont have teh exact traceback now but that was the root cause

app/src/main/java/io/homeassistant/companion/android/onboarding/integration/MobileAppIntegrationFragment.kt
```diff
@@ -70,10 +71,12 @@ class MobileAppIntegrationFragment : Fragment() {
         var checked = isChecked
         if (isChecked) {
             val locationEnabled = DisabledLocationHandler.isLocationEnabled(requireContext())
-            val permissionOk = LocationSensorManager().checkPermission(
-                requireContext(),
-                LocationSensorManager.backgroundLocation.id
-            )
+            val permissionOk = runBlocking {
```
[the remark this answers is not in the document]
dshokouhi: not sure how i missed the `lifecyclescope` there lol

common/src/main/java/io/homeassistant/companion/android/common/sensors/BluetoothSensorManager.kt
```diff
@@ -133,8 +131,6 @@ class BluetoothSensorManager : SensorManager {
         }
     }
 
-    private val ioScope: CoroutineScope = CoroutineScope(Dispatchers.IO)
```
jpelgrom, line 136: You don't need to remove this one? Good to do it on the IO thread instead of main

dshokouhi: done

app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
```diff
@@ -70,11 +68,11 @@ class AndroidAutoSensorManager : SensorManager, Observer<Int> {
     }
 
     override fun onChanged(value: Int) {
-        if (!isEnabled(context, androidAutoConnected)) {
-            CoroutineScope(Dispatchers.Main + Job()).launch {
+        sensorWorkerScope.launch {
```
jpelgrom, line 71: This check will now run async and the code below will continue even if the observer is removed inside this job. I think you should move everything inside the `launch`.

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -409,154 +410,144 @@ class CarSensorManager :
...
+            latestContext,
+            rangeRemaining.sensor,
+            if (rangeRemainingStatus == "success") data.rangeRemainingMeters.value!! else STATE_UNKNOWN,
+            rangeRemaining.sensor.statelessIcon,
+            mapOf(
+                "status" to rangeRemainingStatus
+            ),
+            forceUpdate = true
+        )
+        sensorWorkerScope.launch {
```
jpelgrom, line 445: Why are you launching something here? `setListener` is not suspending.

app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
```diff
@@ -438,15 +427,13 @@ class HealthConnectSensorManager : SensorManager {
     }
 
     override fun hasSensor(context: Context): Boolean {
-        return SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE &&
+        return SDK_INT >= Build.VERSION_CODES.P &&
             HealthConnectClient.getSdkStatus(context) == HealthConnectClient.SDK_AVAILABLE
```
jpelgrom, line 431: The client will handle SDK checks for us

```suggestion
        return HealthConnectClient.getSdkStatus(context) == HealthConnectClient.SDK_AVAILABLE
```

app/src/main/java/io/homeassistant/companion/android/sensors/CarSensorManager.kt
dshokouhi: ah sorry old code

app/src/full/java/io/homeassistant/companion/android/sensors/AndroidAutoSensorManager.kt
dshokouhi: done

jpelgrom:
Testing this on Android 13, requesting permissions seems to be broken (immediately denied) and tapping on settings in the snackbar crashes the app:

```
2024-11-09 22:23:42.768 23522-23522 AndroidRuntime          io....stant.companion.android.debug  E  FATAL EXCEPTION: main
                                                                                                    Process: io.homeassistant.companion.android.debug, PID: 23522
                                                                                                    android.content.ActivityNotFoundException: No Activity found to handle Intent { act=androidx.health.ACTION_REQUEST_PERMISSIONS pkg=io.homeassistant.companion.android.debug (has extras) }
                                                                                                    	at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:2200)
                                                                                                    	at android.app.Instrumentation.execStartActivity(Instrumentation.java:1839)
                                                                                                    	at android.app.Activity.startActivityForResult(Activity.java:5507)
                                                                                                    	at androidx.activity.ComponentActivity.startActivityForResult(ComponentActivity.kt:704)
                                                                                                    	at androidx.core.app.ActivityCompat.startActivityForResult(ActivityCompat.java:244)
                                                                                                    	at androidx.activity.ComponentActivity$activityResultRegistry$1.onLaunch(ComponentActivity.kt:225)
                                                                                                    	at androidx.activity.result.ActivityResultRegistry$register$3.launch(ActivityResultRegistry.kt:199)
                                                                                                    	at androidx.activity.compose.ActivityResultLauncherHolder.launch(ActivityResultRegistry.kt:154)
                                                                                                    	at androidx.activity.compose.ManagedActivityResultLauncher.launch(ActivityResultRegistry.kt:143)
                                                                                                    	at androidx.activity.result.ActivityResultLauncher.launch(ActivityResultLauncher.kt:37)
                                                                                                    	at io.homeassistant.companion.android.settings.sensor.views.SensorDetailViewKt$SensorDetailView$2$1$1.invokeSuspend(SensorDetailView.kt:119)
                                                                                                    	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:101)
                                                                                                    	at androidx.compose.ui.platform.AndroidUiDispatcher.performTrampolineDispatch(AndroidUiDispatcher.android.kt:81)
                                                                                                    	at androidx.compose.ui.platform.AndroidUiDispatcher.access$performTrampolineDispatch(AndroidUiDispatcher.android.kt:41)
                                                                                                    	at androidx.compose.ui.platform.AndroidUiDispatcher$dispatchCallback$1.run(AndroidUiDispatcher.android.kt:57)
                                                                                                    	at android.os.Looper.loop(Looper.java:288)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:7918)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
                                                                                                    	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [androidx.compose.ui.platform.MotionDurationScaleImpl@1008cfc, androidx.compose.runtime.BroadcastFrameClock@f18ff85, StandaloneCoroutine{Cancelling}@15d81da, AndroidUiDispatcher@afce60b]
```

dshokouhi:
> Testing this on Android 13, requesting Health Connect permissions seems to be broken (immediately denied) and tapping on settings in the snackbar crashes the app:

thanks I was able to hunt down an older device will use that for testing and update this PR

dshokouhi:
Fixed, also the snackbar has been fixed too. Just an FYI these permissions are also subject to not being prompted multiple times if they are continuously denied.

Verified on my old Pixel 3 XL running android 12

jpelgrom:
> also the snackbar has been fixed too

I'm not completely sure what you mean. For me, if I deny the permission twice tapping Settings on the snackbar will completely stop working (nothing happens/the system auto-denies). 

Shouldn't the settings button link to an intent with [ACTION_MANAGE_HEALTH_PERMISSIONS](https://developer.android.com/reference/android/health/connect/HealthConnectManager#ACTION_MANAGE_HEALTH_PERMISSIONS)? That seems like it would always work.

dshokouhi:
> I'm not completely sure what you mean. For me, if I deny the permission twice tapping Settings on the snackbar will completely stop working (nothing happens/the system auto-denies).

that should behave like the other sensors right?

> Shouldn't the settings button link to an intent with [ACTION_MANAGE_HEALTH_PERMISSIONS](https://developer.android.com/reference/android/health/connect/HealthConnectManager#ACTION_MANAGE_HEALTH_PERMISSIONS)? That seems like it would always work.

thats only for android 14+, we can launch the HC app though

https://developer.android.com/reference/kotlin/androidx/health/connect/client/HealthConnectClient#ACTION_HEALTH_CONNECT_SETTINGS()

jpelgrom:
> thats only for android 14+, we can launch the HC app though
> 
> https://developer.android.com/reference/kotlin/androidx/health/connect/client/HealthConnectClient#ACTION_HEALTH_CONNECT_SETTINGS()

Oops / that seems like a good option :) (need to put it in the SM to avoid the dependency in minimal)

app/src/main/java/io/homeassistant/companion/android/settings/sensor/views/SensorDetailView.kt
```diff
@@ -116,7 +116,7 @@ fun SensorDetailView(
                 if (result == SnackbarResult.ActionPerformed) {
                     if (it.actionOpensSettings) {
                         if (viewModel.sensorId.startsWith("health_connect")) {
-                            healthConnectPermission?.launch(HealthConnectSensorManager().requiredPermissions(viewModel.sensorId).toSet())
+                            context.startActivity(HealthConnectSensorManager.getPermissionIntent())
```
jpelgrom, line 119: Just to be safe...

```suggestion
                            HealthConnectSensorManager.getPermissionIntent()?.let { intent ->
                                context.startActivity(intent)
                            }
```

dshokouhi: good call, thank you :)

jpelgrom (approved):
Looks good and seems to work as before during testing, but this should be in beta for a while to make sure there is no unexpected impact.

dshokouhi:
> Looks good and seems to work as before during testing, but this should be in beta for a while to make sure there is no unexpected impact.

Agreed this needs beta time

PR #4791, opened 2024-11-08
Bump health services for security fix

dshokouhi:
## Summary

https://developer.android.com/jetpack/androidx/releases/health#health-services-client-1.1.0-alpha04

tested briefly and did not see any issues with health services sensors

does not seem like the RC builds wll get an update so updating to alpha.


jpelgrom:
Looking a previous change logs for other changes included by switching to 1.1 I think there is nothing breaking. Have you tested it?

> does not seem like the RC builds wll get an update so updating to alpha.

Google's updates for this library are quite annoying, commit to it!!

dshokouhi:
> Looking a previous change logs for other changes included by switching to 1.1 I think there is nothing breaking. Have you tested it?

yup tested a few sensors (activity state and daily calories) and got data returned as expected

jpelgrom (approved):

PR #4818, opened 2024-11-14
Clean up location manager after making updates suspend

dshokouhi:
## Summary

Noticed that we had a few places to clean up and remove overlapping suspend calls

also added in a flush location call to only be called when we need to restart location services to clear out the queue, its also recommended we call it as we use `setMaxWaitTime()` so this felt like the most appropriate place to call it.

geofences are also limited to 100 per app so adding in 2 condition checks there. Probably needs a docs update for this one and combined with high accuracy expanded mode will be fun to write 🙃 

https://developer.android.com/develop/sensors-and-location/location/geofencing

Documentation: home-assistant/companion.home-assistant#1141

## Any other notes

Tested these changes overnight to ensure it all still works as expected

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
@@ -231,18 +231,14 @@ class LocationSensorManager : BroadcastReceiver(), SensorManager {
                             forceHighAccuracyModeOn = turnOn
                             forceHighAccuracyModeOff = false
                             setHighAccuracyModeSetting(latestContext, turnOn)
-                            ioScope.launch {
-                                setupBackgroundLocation()
-                            }
+                            setupBackgroundLocation()
                         }
 
                         MessagingManager.FORCE_OFF -> {
                             Log.d(TAG, "High accuracy mode forced off")
                             forceHighAccuracyModeOn = false
                             forceHighAccuracyModeOff = true
-                            ioScope.launch {
-                                setupBackgroundLocation()
-                            }
+                            setupBackgroundLocation()
                         }
 
                         MessagingManager.HIGH_ACCURACY_SET_UPDATE_INTERVAL -> {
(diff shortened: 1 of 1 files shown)


app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -996,14 +983,20 @@ class LocationSensorManager : BroadcastReceiver(), SensorManager {
 
         var geofenceCount = 0
         getEnabledServers(latestContext, zoneLocation).map { serverId ->
-            ioScope.async {
+            sensorWorkerScope.async {
```
dshokouhi, line 986: I am not sure if this shouldve been updated, removed or reverted. For consistency sake I kept it as its a `async` call and just applied our worker to it.

jpelgrom:
By removing `launch` in a lot of places the code now waits/runs in sequence instead of doing this async. I feel like launch was done on purpose in a couple of places, especially for multiserver setups where we want to send to all and don't want to wait for server A to timeout before trying server B. Was this change intentional on your end?

app/src/full/java/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
jpelgrom: Why would you change this to run this on the UI thread instead of a worker thread?

jpelgrom:
OK I think this is good now but am always a bit hesitant with location tracking changes.

Do you want to update the docs to mention the geofence limit? (And also specifically how it impacts with the extended zones for high accuracy)

dshokouhi:
> OK I think this is good now but am always a bit hesitant with location tracking changes.

agreed they are not fun to review nor make changes on



> Do you want to update the docs to mention the geofence limit? (And also specifically how it impacts with the extended zones for high accuracy)

yes we should do that, let me write something up when i get a free moment and link it here

dshokouhi:
lets put this in the beta cycle now

splichy:
Hi, it seems extended zone high accuracy mode no longer works after update to 2025.2.1
I was comming into geofence location two times today and high accuracy mode wasn't triggered as usual - phone was awake running waze, so GPS was active. So I have tried to remove and add high accuracy zone and will try tomorrow again.

jpelgrom:
@splichy If you have an issue, please open a new issue and add the provided information. PR comments are not a good place to discuss potential issues.

Issue #4585, opened 2024-08-16
Last notification sensor doesn't handle grouped notifications well


cryptk:
**Home Assistant Android app version(s):**

**Android version(s):**
14

**Device model(s):**
Pixel 7 Pro

**Home Assistant version:**
2024.8.1

**Last working Home Assistant release (if known):**
unknoen

**Description of problem, include YAML if issue is related to notifications:**
The Last Notification sensor seems to work great for individual notifications, but if an app sends a second notification and Android groups them together, then the sensor appears to report data about the notification group rather than the newly added notification. This makes it difficult to automate things based on incoming notifications as you need to ensure to always dismiss a notification before another arrives.

Here is a mildly redacted (removing real names, etc) payload of the sensor as seen within NodeRed (so it's JSON rather than Yaml)

```json
{
    "payload": "net.smarttag.parentapp",
    "data": {
        "entity_id": "sensor.<REDACTED>_pixel_7_pro_last_notification",
        "old_state": {
            "entity_id": "sensor.<REDACTED>_pixel_7_pro_last_notification",
            "state": "<REDACTED> loaded bus 060",
            "attributes": {
                "android.appInfo": "ApplicationInfo{7822416 net.smarttag.parentapp}",
                "android.bigText": "<REDACTED> loaded bus <REDACTED>",
                "android.infoText": "null",
                "android.largeIcon": "null",
                "android.progress": 0,
                "android.progressIndeterminate": false,
                "android.progressMax": 0,
                "android.reduced.images": true,
                "android.remoteInputHistory": "null",
                "android.showChronometer": false,
                "android.showWhen": true,
                "android.subText": "null",
                "android.template": "android.app.Notification$BigTextStyle",
                "android.text": "<REDACTED> loaded bus <REDACTED>",
                "android.title": "LOADED: <REDACTED> was picked up",
                "androidx.core.app.extra.COMPAT_TEMPLATE": "androidx.core.app.NotificationCompat$BigTextStyle",
                "category": "null",
                "channel_id": "fcm_fallback_notification_channel",
                "group_id": "null",
                "is_clearable": true,
                "is_ongoing": false,
                "package": "net.smarttag.parentapp",
                "post_time": 1723837062699,
                "icon": "mdi:bell-ring",
                "friendly_name": "<REDACTED>' Pixel 7 Pro Last notification"
            },
            "last_changed": "2024-08-16T19:37:50.998323+00:00",
[...]

PR #4822, opened 2024-11-17
Skip ranker_group notifications as they are meaningless and system generated and prevent valid notification data from sending

dshokouhi:
## Summary

Skip the system generated `ranker_group` from sending a meaningless update to HA as it prevents the actual notification data needed.

In my own data I can see that this "special" group is quite meaningless as it has no text or anything other than the app posted a notification

image

we can see the previous notification had no group however the notification that was desired got skipped due to the `ranker_group` superseding the update

image

testing locally with the HA prod and debug app side by side with each app listening for the others notification. Sending a message with "test" followed by "test2" results in "test" being sent and then the ranker_group for prod. The debug build with this change shows the expected sequence of "test" followed by "test2" instead of the ranker_group.

## Any other notes

this behavior was also called out in https://github.com/home-assistant/android/issues/4585 where we can see the system generated group is just unwanted.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
@@ -144,6 +144,10 @@ class NotificationSensorManager : NotificationListenerService(), SensorManager {
                 return@launch
             }
 
+            if (sbn.notification.group == "ranker_group") {
+                return@launch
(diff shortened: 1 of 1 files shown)


jpelgrom:
> however the notification that was desired got skipped due to the ranker_group superseding the update

I think this just highlights how this data is a bad fit for a sensor. Nevertheless, nice workaround.

Issue #4829, opened 2024-11-19
Crash caused by Health Connect rate limiting


marazmarci:
**Home Assistant Android app version(s):**
`2024.11.5-full`

**Android version(s):**
Android 13

**Device model(s):**
Samsung S20 FE 5G

**Home Assistant version:**
`2024.8.1`

**Last working Home Assistant release (if known):**
I first observed this crash on this version, but it may have crashed earlier in the background.

**Description of problem, include YAML if issue is related to notifications:**

**Companion App Logs:**

```
FATAL EXCEPTION: main
Process: [io.homeassistant.companion.android](http://io.homeassistant.companion.android/)
android.os.RemoteException: Request rejected. Rate limited request quota has been exceeded. Please wait until quota has replenished before making further requests.
    at androidx.health.connect.client.impl.HealthConnectClientImpl.getGrantedPermissions(HealthConnectClientImpl.kt:301)
    at androidx.health.connect.client.impl.HealthConnectClientImpl$getGrantedPermissions$1.invokeSuspend(Unknown Source:14)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:99)
    at android.os.Handler.handleCallback(Handler.java:942)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at android.os.Looper.loopOnce(Looper.java:226)
    at android.os.Looper.loop(Looper.java:313)
    at android.app.ActivityThread.main(ActivityThread.java:8762)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:604)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1067)
    Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@64cf3d, Dispatchers.Main.immediate]
Caused by: android.os.RemoteException: Request rejected. Rate limited request quota has been exceeded. Please wait until quota has replenished before making further requests.
    at androidx.health.platform.client.impl.error.ErrorStatusConverterKt.toException(ErrorStatusConverter.kt:49)
    at androidx.health.platform.client.impl.FilterGrantedPermissionsCallback.onError(FilterGrantedPermissionsCallback.kt:34)
[...]

PR #4830, opened 2024-11-19
Catch all health connect exceptions so we dont crash

dshokouhi:
## Summary

Fixes: #4829 by catching the exceptions so the app does not crash

## Screenshots

data is still good 

image

### DIFF
--- a/app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
+++ b/app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
@@ -175,19 +175,24 @@ class HealthConnectSensorManager : SensorManager {
         get() = commonR.string.sensor_name_health_connect
 
     override fun requiredPermissions(sensorId: String): Array<String> {
-        return when {
-            (sensorId == activeCaloriesBurned.id) -> arrayOf(HealthPermission.getReadPermission(ActiveCaloriesBurnedRecord::class))
-            (sensorId == bloodGlucose.id) -> arrayOf(HealthPermission.getReadPermission(BloodGlucoseRecord::class))
-            (sensorId == bodyFat.id) -> arrayOf(HealthPermission.getReadPermission(BodyFatRecord::class))
-            (sensorId == distance.id) -> arrayOf(HealthPermission.getReadPermission(DistanceRecord::class))
-            (sensorId == elevationGained.id) -> arrayOf(HealthPermission.getReadPermission(ElevationGainedRecord::class))
-            (sensorId == floorsClimbed.id) -> arrayOf(HealthPermission.getReadPermission(FloorsClimbedRecord::class))
-            (sensorId == heartRate.id) -> arrayOf(HealthPermission.getReadPermission(HeartRateRecord::class))
-            (sensorId == sleepDuration.id) -> arrayOf(HealthPermission.getReadPermission(SleepSessionRecord::class))
-            (sensorId == steps.id) -> arrayOf(HealthPermission.getReadPermission(StepsRecord::class))
-            (sensorId == totalCaloriesBurned.id) -> arrayOf(HealthPermission.getReadPermission(TotalCaloriesBurnedRecord::class))
-            (sensorId == weight.id) -> arrayOf(HealthPermission.getReadPermission(WeightRecord::class))
-            else -> arrayOf()
+        return try {
+            when {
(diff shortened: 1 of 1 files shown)


app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
```diff
@@ -229,188 +234,232 @@ class HealthConnectSensorManager : SensorManager {
...
-        }
-        onSensorUpdated(
-            context,
-            weight,
-            BigDecimal(response.records.last().weight.inGrams).setScale(2, RoundingMode.HALF_EVEN),
-            weight.statelessIcon,
-            attributes = mapOf(
-                "date" to response.records.last().time,
-                "source" to response.records.last().metadata.dataOrigin.packageName
+        try {
```
jpelgrom, line 254: That's a lot of try/catch if you remember that there are tons of additional sensors we could add. What about an extension function:
```kotlin
    private suspend fun <T : Record> HealthConnectClient.readRecordsOrNull(request: ReadRecordsRequest<T>): ReadRecordsResponse<T>? = try {
        readRecords(request)
    } catch (e: Exception) {
        Log.e(TAG, "Could not read records", e)
        null
    }
```
In sensor code then you only need to update 2 lines:
```kotlin
    val response = healthConnectClient.readRecordsOrNull(weightRequest)
    if (response == null || response.records.isEmpty()) {
        return
    }
```

app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
```diff
@@ -434,8 +483,13 @@ class HealthConnectSensorManager : SensorManager {
 
     override suspend fun checkPermission(context: Context, sensorId: String): Boolean {
         val healthConnectClient = getOrCreateHealthConnectClient(context) ?: return false
-        val result = healthConnectClient.permissionController.getGrantedPermissions().containsAll(requiredPermissions(sensorId).toSet())
-        return result
+        return try {
+            val result = healthConnectClient.permissionController.getGrantedPermissions().containsAll(requiredPermissions(sensorId).toSet())
+            result
```
jpelgrom, line 488: You can now immediately return here without assigning it to result

jpelgrom (commented):
Rate limiting does pose an interesting question as we add more sensors: should this somehow be exposed / allow the user to limit update frequency for HC independently?

app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
dshokouhi: this wont work each record has unique data which is why we have all those unchecked casts

dshokouhi:
> should this somehow be exposed / allow the user to limit update frequency for HC independently?

feels like lots of others will request independent updates, ideally all users will have foreground reads here i think

app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt
```diff
@@ -434,8 +440,12 @@ class HealthConnectSensorManager : SensorManager {
 
     override suspend fun checkPermission(context: Context, sensorId: String): Boolean {
         val healthConnectClient = getOrCreateHealthConnectClient(context) ?: return false
-        val result = healthConnectClient.permissionController.getGrantedPermissions().containsAll(requiredPermissions(sensorId).toSet())
-        return result
+        return try {
+            healthConnectClient.permissionController.getGrantedPermissions().containsAll(requiredPermissions(sensorId).toSet())
+        } catch (e: Exception) {
+            Log.e(TAG, "Unable to check permissions", e)
+            false
```
jpelgrom, line 447: This works, but will actively turn off the HC sensors once we reach the rate limit. Is that really desirable?

Update [checks `isEnabled`](https://github.com/home-assistant/android/blob/master/app/src/full/java/io/homeassistant/companion/android/sensors/HealthConnectSensorManager.kt#L195) -> `isEnabled` will check the permission and [pass the false value to the database](https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/sensors/SensorManager.kt#L113) -> the database will [disable any enabled sensors found](https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/database/sensor/SensorDao.kt#L165) while there is no permission.

Issue #4826, opened 2024-11-19
Most notification channels do not update their names when app languages changes


NoRi2909:
**Home Assistant Android app version:**
2024.11.5-full (beta)
**Android version:**
15
**Device model:**
Pixel 6
**Home Assistant version:**
2024.11.2

**Description of problem, include YAML if issue is related to notifications:**

When changing the app language between English and German the name of the 'Beacon monitor scanning' notification channel does update after changing the language (and re-enabling it). So that is working as expected. 👍 

For the 'Persistent connection' notification channel this is not the case , it stays at the name that was given to it when first enabled.
The old German translation was "Dauerhafte Verbindung" ("_permanent_ connection") which I recently changed to "Anhaltende Verbindung" ("_persistent_ connection") all over Home Assistant. So we now have that consistency everywhere, but the notification channel is stuck at its old German name, too.

Same seems to apply to the 'High accuracy mode' and 'Sensor worker' channels (note that we have camel-cased that in German to become a name), They are also stuck at their German names when switching the app language back to English:

![Screenshot](https://github.com/user-attachments/assets/19fef988-ab37-42ff-a7cd-655a33164b89)

Finally the 'General' channel needs a localizable string as it shows up with the English name in German, too. Now in German "General" is the military rank only, while the correct translation here is "Allgemein". Probably quite similar in other, future languages.

dshokouhi:
General does indeed need to be updated but the others need to be triggered again meaning you need to turn on persistent connection, trigger a sensor update and turn on high accuracy mode. With the desired language after you confirm the strings are in the app version

NoRi2909:
The updated German strings are already in that beta, I see them in the UI. But regardless it should have changed between English and German names when I switched the app language.

I will try again this evening by enabling and disabling the different functions to see if that triggers the language change.

I did try for quite a while yesterday and did not see any effect, only with the Beacon Monitor Scanning that changed on every languages switch without a hitch.

dshokouhi:
> But regardless it should have changed between English and German names when I switched the app language.

channels can only have 1 name at a time, they need to be updated each and every time when you switch languages for it to take place. Sounds like you may have had the notification up and switched languages which caused the update to send.

those values you see are not from the app in this screen but what was provided to android at the time the channel was created and/or updated

dshokouhi:
you need to trigger the notifications in question to appear everytime you switch the language when the strings are actually provided

NoRi2909:
[...]

PR #4833, opened 2024-11-20
Use translated channel name strings where possible to update them when language changes

dshokouhi:
## Summary

Fixes: #4826 by ensuring we always set the latest channel name

### DIFF
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -1314,5 +1314,5 @@
     <string name="basic_sensor_name_blood_glucose">Blood glucose</string>
     <string name="sensor_description_blood_glucose">Last recorded blood glucose reading in milligrams per deciliter from Health Connect</string>
     <string name="general">General</string>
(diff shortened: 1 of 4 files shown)


common/src/main/res/values/strings.xml
```diff
@@ -1313,4 +1313,6 @@
     <string name="sensor_description_steps">Total steps taken since midnight from Health Connect</string>
     <string name="basic_sensor_name_blood_glucose">Blood glucose</string>
     <string name="sensor_description_blood_glucose">Last recorded blood glucose reading in milligrams per deciliter from Health Connect</string>
+    <string name="general">General</string>
+    <string name="sensor_worker">Sensor Worker</string>
```
jpelgrom, line 1317: Maybe now is the time to rename this to something that makes more sense to users like "Sensor updates"?

dshokouhi: I am glad you said something :)  think that sounds like a good expectation

PR #4835, opened 2024-11-21
Add a test to introduce compose screenshot testing

dshokouhi:
## Summary

Stumbled upon the following page and thought it might be good to add these tests to ensure we dont break the UI, for now a very simple test to compare the entity list just to provide a concept. I like the ease of use of the compose screenshot tool as it plays very very nice with our existing previews. Wear OS first as we have seen the alpha compose builds break things. I think it might be nice to add tests for all of our previews so we can use them more. One thing to consider is that we may need to duplicate the preview code so we can continue to see the previews wehn editing the file. Maybe there is a better way I am not seeing but for now I copied a very simple preview for this.

https://developer.android.com/studio/preview/compose-screenshot-testing

open to suggestions also wanted to submit the action to see how it flows.

## Screenshots

Example report:
image

## Any other notes

If this is approach is acceptable then the readme should be updated here to help contributors understand how to update reference images if they need to.


dshokouhi:
ok updated and passes locally:

image

however we may be a victim of https://issuetracker.google.com/issues/364331849

TimoPtr:
@dshokouhi I fixed the issue with ktlint in https://github.com/home-assistant/android/pull/5156 you can see it working on your code on this closed PR https://github.com/home-assistant/android/pull/5155

.github/workflows/pr.yml
```diff
@@ -93,6 +93,43 @@ jobs:
...
+            ${{ runner.os }}-gradle-
+
+      - name: Mock google-services.json
+        run: |
+          cp .github/mock-google-services.json app/google-services.json
+          cp .github/mock-google-services.json wear/google-services.json
+          cp .github/mock-google-services.json automotive/google-services.json
+
+      - uses: gradle/actions/wrapper-validation@v4
+        name: Validate Gradle Wrapper
```
TimoPtr, line 122:
```suggestion
      - uses: ./.github/actions/setup-build-env
        with:
          mock-google-services: "true"
```

.github/workflows/pr.yml
```diff
@@ -93,6 +93,43 @@ jobs:
...
+
+      - name: Validate Screenshot Previews
+        run: ./gradlew validateDebugScreenshotTest
+
+      - name: Upload screenshot results
+        uses: actions/upload-artifact@v4
+        with:
+          name: screenshot-results
+          path: wear/build/reports/screenshotTest
+
```
TimoPtr, line 132: Since the task `validateDebugScreenshotTest` is generating a Junit report you should store it with
```
      - name: Upload test results
        if: always()
        uses: actions/upload-artifact@v4
        with:
          name: Screenshot test results
          path: |
            **/build/test-results/**/TEST-*.xml
```

and then modify the job `publish_test_results` to depends on `screenshot_tests`

```
  publish_test_results:
    name: "Publish Tests Results"
    needs: [instrumentation_test, unit_tests, screenshot_test]
```

with this Github will know exactly which test failed and it's going to be integrated to the report.

TimoPtr:
I ran the test on my macbook and it failed to validate the image
image

dshokouhi:
> I ran the test on my macbook and it failed to validate the image

added the threshold example based on the issue tracker that discusses this exact issue

https://issuetracker.google.com/issues/348590914

we probably need to tweak the number but hopefully the example is good enough 🤞

TimoPtr:
image
That's the number I need to have it working on macOS.

dshokouhi:
should we update the readme here or wait until we use it more? the other tests that were added recently are also not mentioned so not sure if it should be mentioned or if there is a better place? maybe better to define the process once more tests have been added?

TimoPtr:
If that's fine for you I would like to add a mobile section in the developer website of Home assistant to explain the general architecture and all the things like the CI and testing.

dshokouhi:
> If that's fine for you I would like to add a mobile section in the developer website of Home assistant to explain the general architecture and all the things like the CI and testing.

I think that would be great and help clean up the read me more and keep things in line with the rest of HA development. May even help encourage more contributors to come!

.github/workflows/pr.yml
```diff
@@ -97,6 +97,26 @@ jobs:
         with:
           category: "Android Lint"
 
+  screenshot_test:
+    runs-on: ubuntu-latest
+    steps:
```
jpelgrom, line 102: Looking at the flow diagram on #5158: wouldn't it make more sense to run this after lint/basic build has completed? So `    needs: [lint, lockfiles, ktlint]` like the other tests?

dshokouhi: yea grouped next to unit tests maybe?

jpelgrom: next to = running at the same time? That's what I was thinking, so basically as 'stage 3'.

dshokouhi: yup let me submit a PR to add that, thanks for the suggestion :)

PR #4837, opened 2024-11-21
Add dialog before prompting for Improv permissions

jpelgrom:
## Summary

Based on feedback: add a dialog explaining why the app is about to request permissions for scanning for Improv devices. This should prevent users from denying it because it is not understood. Note that the 2nd prompt may not actually result in a system prompt if it is auto-denied.

Moves some code from onboarding views and bottom sheet dialogs to utility files where it is now used in more places.

## Screenshots

|Light|Dark|
|-----|-----|
|Add integration with a bottom dialog sheet titled 'Search devices', light mode|Add integration with a bottom dialog sheet titled 'Search devices', dark mode|

n/a

## Any other notes

related to home-assistant/iOS#3104

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/improv/ui/ImprovPermissionDialog.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/improv/ui/ImprovPermissionDialog.kt
@@ -1,0 +1,93 @@
+package io.homeassistant.companion.android.improv.ui
+
+import android.Manifest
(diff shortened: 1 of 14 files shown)


dshokouhi (approved):
Nice improvement and always good to inform the user about things like this! Tested the APK and it works well. I really dig these bottom style sheets!

bgoncal (approved):

PR #4845, opened 2024-11-24
AAOS/Android Auto: Add a button to refresh the main screen entity list

anstick:
## Summary
If the HA app starts without internet access (e.g., when the car is in an underground garage, as in my case), the list of domains/entities remains empty. Currently, there's no way to refresh it without killing the app. For example, Volvo AAOS doesn't have an option to force-stop apps, so I end up having to restart the car.

This PR adds a refresh button to the main screen header to solve this issue.

## Screenshots
image

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/util/vehicle/NativeModeActionStrip.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/util/vehicle/NativeModeActionStrip.kt
@@ -5,20 +5,17 @@ import android.content.pm.PackageManager
 import android.util.Log
 import androidx.car.app.CarContext
 import androidx.car.app.model.Action
-import androidx.car.app.model.ActionStrip
 import io.homeassistant.companion.android.common.R
 import io.homeassistant.companion.android.launch.LaunchActivity
 
 private const val TAG = "NativeActionStrip"
 
-fun nativeModeActionStrip(carContext: CarContext): ActionStrip {
-    return ActionStrip.Builder().addAction(
-        Action.Builder()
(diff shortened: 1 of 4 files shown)


dshokouhi:
The native mode button only shows up when the car is parked, for this maybe we should only show the refresh button if we have 0 entities? Otherwise the icon looks like its required to refresh state data. Also not sure if we should have text instead of an icon to make it more clear what it actually does?

anstick:
> The native mode button only shows up when the car is parked, for this maybe we should only show the refresh button if we have 0 entities?

We can implement this, however there are use cases where users may still want to refresh the list (though these scenarios are relatively uncommon)

> Also not sure if we should have text instead of an icon to make it more clear what it actually does?

We can't have 2 text action buttons in a header (AAOS templates limitation).

app/src/main/java/io/homeassistant/companion/android/util/vehicle/NativeModeActionStrip.kt
```diff
@@ -1,39 +1,36 @@
-package io.homeassistant.companion.android.util.vehicle
```
dshokouhi, line 1: The diff in this file has become hard to review, can you please update this so only the changes are visible? I think this might have to do with the line endings on the file being changed?

anstick: Sorry, it's `./gradlew ktlintFormat`. Fixed it

dshokouhi (approved):
Tested and works well when I bring my dev environment on and offline! Thank you for the contribution! I dont see any negative effects of always leaving the refresh button there.

Issue #4866, opened 2024-12-02
Add an alternative method for setting up server endpoints in the companion app without requiring location services


JosephM101:
**Is your feature request related to a problem? Please describe.**

I have remote access for my Home Assistant server set up through Tailscale. Love the service, love the concept, but their mobile app for Android is a bit buggy. For example, if my phone restarts, the Tailscale app isn't automatically enabled and stuff like that, or it turns itself off randomly.

My companion app is set up with the Tailscale address as `Home Assistant URL`, which is used even when I'm at home. The reason I do this is because the `Internal connection URL` requires `Home network WiFi SSID` to be set, and that detection method requires location to be enabled all of the time. I keep location turned off on my phone because I view it as both a privacy risk and a battery drain, so turning it on for this just isn't viable for me.

**Describe the solution you'd like**

I would like an alternative method that doesn't require me to specify a `Home network WiFi SSID`. Basically, the way this alternative configuration option would work is I could specify the local/home address and the Tailscale address in a list, where the address at the top of the list is the one with the highest priority. In the background, either when a network change is detected or on a set interval, the companion app would ping each URL to see which are reachable, and connect to the first one available, or the one with the highest priority. So if I'm at home, and Tailscale deactivates itself, the companion app would detect that Tailscale was unreachable and instead try to use one of the other address(es) in the list.

**Describe alternatives you've considered, if any**

Can't think of any.

**Additional context**

dshokouhi:
is this a request to add something for tailscale? because we will not be doing anything specific for that.

the reason location is required is so the app can read the connected SSID otherwise we have no idea what network we are on or any other network related data.

dshokouhi:
> It's not specific to Tailscale. It would just be an alternate way of determining which URL to use. Instead of determining it on a network basis, you ping every URL in a list and pick the accessible one, thus negating the requirement for Location services to be enabled.

Settings > companion app > select your server > home wifi > change bottom option to prefer internal and you will need to be ok with any pop-ups that may need to be refreshed.

this is how it is intended to work.

JosephM101:
Then I'm afraid that's a problem. I stated in my original post that I'm not okay with having location services always enabled on my Android phone because I view it as a privacy risk, plus it is a battery drain.

It appears that we have reached an impasse. My apologies for wasting your time. Have a good day.

dshokouhi:
turn off location once you enable that setting and it should work as you expect

PR #4872, opened 2024-12-03
Allow treating ethernet and VPN as home network/for internal URL

### LINKED ISSUE #2532: Recognize ethernet as Home Network

### LINKED ISSUE #4852: The HomeAssistant app does not connect to the server when WireGuard VPN is active on my phone.
2024.10.3
**Android version(s):**
12
**Device model(s):**
S61Pro
**Home Assistant version:**
Core
2024.11.3
Supervisor
2024.11.4
Operating System
13.2
Kezelőfelület
20241106.2

I have a WireGuard server running on my router, and the WireGuard app is also installed on my phone. When I activate the connection, I can access all my home devices. However, when I launch the Home Assistant app, it loads for a while and then displays an error saying it cannot connect to the server. I assume this is because it is not trying to connect using the internal IP but rather the external URL. I couldn’t find any settings in the app to specify which address it should use when a VPN connection is active.

**Additional information:**

jpelgrom:
## Summary

Add options to also treat ethernet or VPN as home network (for internal URL, app lock, websocket push, BLE monitor). It reads the value from the system, not a specific app, so it matches when the status bar icon shows a <...> for ethernet or key for VPN (no additional details are available).

This should help with users who want internal URL + VPN, which I've seen multiple times on forums/Discord etc., and does not require location access for the app or location services to be on. Addresses:
 - Fixes #2532 
 - Fixes #4852 - "I assume this is because it is not trying to connect using the internal IP but rather the external URL" can now be fixed by enabling VPN
 - Sort of #4866 - it is not pinging but doesn't require location
 - Split tunnel VPN setups

The new values are nullable so we can change defaults and possibly introduce an inverse option in the future.

## Screenshots

The 'home network WiFi SSID' screen is extended with the new options. 
|Light|Dark|
|----|----|
|image|image|

The check for location access is moved to this screen, instead of before the screen, and will display inline next to the WiFi networks setting.
|Light|Dark|
|----|----|
|image|image|

I don't see how adding this advanced information to the basic setup page helps, probably fine to leave it as it is

## Any other notes

For review purposes: the first two commits include the actual logic. The third commit includes UI.

### DIFF
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -394,7 +394,7 @@
     <string name="location_history_skipped_not_latest">Old location</string>
     <string name="location_history_skipped_old">Very old location</string>
     <string name="location_history_use">Use location history</string>
-    <string name="location_perm_info_message">Android set up restrictions for apps which want to use your WiFi, because WiFi can be theoretically used to determine your location.\n\nAlso to ensure the app can access WiFi in background (URL decision making, sensors) you need to allow location access all the time.\n\nTherefore, to use this option location access all the time is needed.\n\nContinue granting location access?</string>
+    <string name="location_perm_info_message">Android set up restrictions for apps which want to use your WiFi, because WiFi can be theoretically used to determine your location.\n\nATo ensure the app can access WiFi in background (URL decision making, sensors), you must allow location access all the time.\n\nTherefore, to use this option location access all the time is needed.\n\nContinue granting location access?</string>
     <string name="location_perm_info_title">Location access</string>
     <string name="location_tracking">Location tracking</string>
     <string name="location_tracking_summary">View a history of location tracking updates to troubleshoot the device tracker</string>
@@ -402,8 +402,8 @@
     <string name="location">Location</string>
     <string name="lock_summary">Use biometric or screen lock credentials to unlock app</string>
     <string name="lock_title">Lock app</string>
-    <string name="lock_home_bypass_summary">Disable app lock when connected to the home WiFi network</string>
-    <string name="lock_home_bypass_title">Unlock on home WiFi</string>
+    <string name="lock_home_bypass_summary">Disable app lock when connected to the home network</string>
+    <string name="lock_home_bypass_title">Unlock on home network</string>
     <string name="locks">Locks</string>
     <string name="lockscreen_title">App locked!</string>
     <string name="log_file_not_created">Log file could not be created.</string>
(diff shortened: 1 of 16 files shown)


common/src/main/res/values/strings.xml
```diff
@@ -394,16 +394,16 @@
     <string name="location_history_skipped_not_latest">Old location</string>
     <string name="location_history_skipped_old">Very old location</string>
     <string name="location_history_use">Use location history</string>
-    <string name="location_perm_info_message">Android set up restrictions for apps which want to use your WiFi, because WiFi can be theoretically used to determine your location.\n\nAlso to ensure the app can access WiFi in background (URL decision making, sensors) you need to allow location access all the time.\n\nTherefore, to use this option location access all the time is needed.\n\nContinue granting location access?</string>
+    <string name="location_perm_info_message">Android set up restrictions for apps which want to use your Wi-Fi, because Wi-Fi can be theoretically used to determine your location.\n\nATo ensure the app can access Wi-Fi in background (URL decision making, sensors), you must allow location access all the time.\n\nTherefore, to use this option location access all the time is needed.\n\nContinue granting location access?</string>
```
dshokouhi, line 397:
```
     location.\n\nATo 
```

looks like a typo?

jpelgrom: Yes, thanks

dshokouhi (approved):
Tested the APK and works well over here with Google VPN and also my home wifi configured. Great PR! Should make a lot of people happy I think :)

marazmarci:
FYI there's a minor [security concern](https://github.com/home-assistant/android/issues/2532#issuecomment-2518400511) associated to this feature. I don't really have a feasible idea on how to address it, other than showing some kind of disclaimer / warning text for the users before allowing turning this on.

((But here's an unfeasible idea, which still isn't 100% secure: get the IP of the current default gateway's IP and based on that, query its MAC address, which is a unique (but spoofable) identifier of the router. This MAC address could act as the identifier for the network, similarly to the SSID in case of WiFi networks.))

marazmarci:
If only there would be an unauthenticated* endpoint on the HA server, through which it could verify its identity towards the app, before the app making an authenticated connection, then it could be solved pretty easily, I think. But I'm pretty sure there's no such endpoint, so we'd need it to be developed in the [core](https://github.com/home-assistant/core) repository...

*: so the app won't "leak" the session token too early

jpelgrom:
I was aware of that as I read the issues, but believe this feature is limited enough in scope (users will probably not use different ethernets/vpns) and 'advanced' enough for normal users to leave the settings at the default (off) that they are/should be aware.

> But here's an unfeasible idea, which still isn't 100% secure

As I see it we could explore that for a v1.1, but that may increase the permissions required for the app.

> If only there would be an unauthenticated* endpoint on the HA server, through which it could prove its identity towards the app

This would also apply to SSIDs. MITM is always a concern, so let's not introduce that discussion here.

marazmarci:
And what do you think of my suggestion about adding a warning text to the UI?

Something along these lines:

_Warning: this could pose a security risk when using the device on multiple Ethernet networks / VPNs._

jpelgrom:
Your warning is very specific for http + internal connection, and "a security risk" doesn't tell users a lot, while this feature is more general. At the same time I want to avoid writing a whole paragraph of text. What do you think about:

> Make sure to set up your home network correctly. Adding public Wi-Fi networks or using multiple ethernet/VPN connections may unintentionally expose information about or access to your app or server.

marazmarci:
> Your warning is very specific for http + internal connection

Oh, yes indeed 🤦‍♂️

> What do you think about:
> > Make sure to set up your home network correctly. Adding public Wi-Fi networks or using multiple ethernet/VPN connections may unintentionally expose information about or access to your app or server.

This is to the point, and concise. Well done, thanks!

jpelgrom:
Warning alert added as discussed:
image

PR #4917, opened 2024-12-19
Add support for media players in device controls

dshokouhi:
## Summary

Adds support for media player entities in device controls.

Toggle control will play/pause, active state is only if `playing`

Slider control for volume if media player supports setting the volume

Behavior designed to match frontend as close as possible due to conversions for volume

## Screenshots

image

Documentation: home-assistant/companion.home-assistant#1149

## Any other notes

Ref: https://community.home-assistant.io/t/wth-why-cant-media-player-entities-be-added-to-android-device-controls/813242

The actions accept any value, we don't particularly need to round things but good to keep things clean i think. Tested with a few of my media players.

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
@@ -17,6 +17,7 @@ import io.homeassistant.companion.android.common.data.integration.Entity
 import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
 import io.homeassistant.companion.android.common.data.integration.getVolumeLevel
 import io.homeassistant.companion.android.common.data.integration.getVolumeStep
+import io.homeassistant.companion.android.common.data.integration.isActive
 import io.homeassistant.companion.android.common.data.integration.supportsVolumeSet
 import java.math.BigDecimal
 import java.math.RoundingMode
@@ -29,13 +30,12 @@ object MediaPlayerControl : HaControl {
         entity: Entity<Map<String, Any>>,
         info: HaControlInfo
     ): Control.StatefulBuilder {
-        val isPlaying = entity.state == "playing"
         if (entity.supportsVolumeSet()) {
             val volumeLevel = entity.getVolumeLevel()
             control.setControlTemplate(
                 ToggleRangeTemplate(
                     entity.entityId,
-                    isPlaying,
+                    entity.isActive(),
                     "",
                     RangeTemplate(
                         entity.entityId,
@@ -52,7 +52,7 @@ object MediaPlayerControl : HaControl {
                 ToggleTemplate(
                     entity.entityId,
                     ControlButton(
-                        isPlaying,
+                        entity.isActive(),
                         ""
                     )
                 )
--- a/app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
@@ -1,0 +1,98 @@
+package io.homeassistant.companion.android.controls
+
+import android.content.Context
+import android.os.Build
+import android.service.controls.Control
+import android.service.controls.DeviceTypes
+import android.service.controls.actions.BooleanAction
+import android.service.controls.actions.ControlAction
+import android.service.controls.actions.FloatAction
+import android.service.controls.templates.ControlButton
+import android.service.controls.templates.RangeTemplate
+import android.service.controls.templates.ToggleRangeTemplate
+import android.service.controls.templates.ToggleTemplate
+import androidx.annotation.RequiresApi
+import io.homeassistant.companion.android.common.R as commonR
+import io.homeassistant.companion.android.common.data.integration.Entity
+import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
+import io.homeassistant.companion.android.common.data.integration.getVolumeLevel
+import io.homeassistant.companion.android.common.data.integration.getVolumeStep
+import io.homeassistant.companion.android.common.data.integration.supportsVolumeSet
+import java.math.BigDecimal
+import java.math.RoundingMode
+
+@RequiresApi(Build.VERSION_CODES.R)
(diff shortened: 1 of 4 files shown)


app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
```diff
@@ -0,0 +1,98 @@
...
+                        ""
+                    )
+                )
+            )
+        }
+        return control
+    }
+
+    override fun getDeviceType(entity: Entity<Map<String, Any>>): Int =
+        DeviceTypes.TYPE_TV
```
jpelgrom, line 65: This doesn't make a lot of sense for anything other than a tv like speakers, also because it sets the icon to a tv, but I don't see a good alternative 🤔

common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -298,6 +299,50 @@ fun <T> Entity<T>.getLightColor(): Int? {
...
+fun <T> Entity<T>.getVolumeLevel(): EntityPosition? {
+    return try {
+        if (!supportsVolumeSet()) return null
+
+        val minValue = 0f
+        val maxValue = 100f
+
+        // Convert to percentage to match frontend behavior:
+        // https://github.com/home-assistant/frontend/blob/dev/src/dialogs/more-info/controls/more-info-media_player.ts#L137
+        val currentValue = ((attributes as Map<*, *>)["volume_level"] as? Number)?.toFloat()?.times(100) ?: 0f
```
jpelgrom, line 321: Why do `times(100)` if you're going to divide it again when submitting the action and can set the step size?

app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
dshokouhi: yea was meaning to comment its the closest to anything media related except for `display`

common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt
dshokouhi: in reality to really read volume level we don't need this method we could just read the attribute and call it a day but to do all the logic needed for device controls display we have this. Even for Wear OS we would want to show the controls as the frontend does and leave the attributes unconverted.

dshokouhi: The current logic  here is similar to `LightControl` as well where we do some conversion in the helper and the rest in appropriate places (lights for example impacts device controls and wear OS details pane).

app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
jpelgrom: Checked Google Home: looks like they are showing a cast icon. Doesn't seem like it's built in so must be set manually (similar to how we override the icon if the user has set a different icon for the entity). I wonder if it is maybe worth always setting an icon here to better match the HA media player entity, or if I'm overreacting.

app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
```diff
@@ -0,0 +1,98 @@
...
+
+@RequiresApi(Build.VERSION_CODES.R)
+object MediaPlayerControl : HaControl {
+    override fun provideControlFeatures(
+        context: Context,
+        control: Control.StatefulBuilder,
+        entity: Entity<Map<String, Any>>,
+        info: HaControlInfo
+    ): Control.StatefulBuilder {
+        val isPlaying = entity.state == "playing"
```
jpelgrom, line 32: Side effect I noticed while testing: now it visually turns off when buffering. Should we match the frontend here instead, even if the toggle action makes less sense? <https://github.com/home-assistant/frontend/blob/e58bef7795f610b40820f5b8df5e62d733fe4642/src/common/entity/state_active.ts#L42> active = not `standby`.

jpelgrom:
Seems to work well, just not completely sure about the presentation yet (added two more comments).

app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
dshokouhi: > I wonder if it is maybe worth always setting an icon here to better match the HA media player entity, or if I'm overreacting.

my media players have different type of icons, we could go based off `device_class`? if we do that we have extra considerations as we are only overriding if icon is provided. Do we just first apply device_class then icon? Only do a special case for media player?

app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
dshokouhi: i originally had this but changed it based on internal discussion :)

app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
jpelgrom: I would only consider it for media player because it is hard to fit into the native UI (every manufacturer doing something different)

app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
jpelgrom: Seeing it live cam change your mind :)

app/src/main/java/io/homeassistant/companion/android/controls/MediaPlayerControl.kt
dshokouhi: ok will make sure to still respect the icon if it was overridden

dshokouhi: done and left a comment about the special case :)

jpelgrom (approved):

Issue #3142, opened 2022-12-07
Improve splashscreen by hiding webview until fully loaded.


LasseRosenow:
**Is your feature request related to a problem? Please describe.**

I have been working on improving the splash screen experience of this app in the past: See #1938
The idea was to have the android launch-screen match the in-app launch screen.

But this does not work anymore since android 12 or 13? Has this new automatic splash screen, that has constraints on how large the icon could be etc. And the html icon is larger.

**Describe the solution you'd like**

I think having the html version in sync with the android version is to complicated and could always break in the future.
This is why I want to separate this and let the android app draw the whole splash screen, skipping the html splash screen in total.

The procedure should look like this:
1. Native android splash screen drawn by the os
2. in-app splash-screen that exactly matches the os splash-screen (we can set icon and color for os-splash screen, to make this more solid)
3. the in-app splash-screen is behind the webview, which is transparent
4. We wait for some event like `theme-update` from the webview maybe call it `app-initialized` and then show the webview.

This way the app startup would be massively clean and not consist of so many blinking etc. anymore :)

**Describe alternatives you've considered, if any**
I have considered making the icon in the html smaller for android devices, but I think this is not a good solution.

LasseRosenow:
If you think this is a good idea, I will make a PR for this when I have spare time, probably in the next few weeks (at most month)

LasseRosenow:
Maybe the `homeassistant_started` event could be the correct one to listen for :thinking:

jpelgrom:
While it would be nice to improve this 'handoff' from app to frontend, relying on events makes it easier for this to break/prevent the webview from showing at all if the event isn't emitted or received for whatever reason. Examples: the frontend's "can't connect" message and #2684 already trigger the app to show a dialog it can't connect while it can (and the webview shows useful stuff).

LasseRosenow:
What about showing the webview after for example 5 seconds, if we don't get the event?

This would probably make it work how it should be in 99% of the cases, but still show the webview, if something does not work as expected. The user in this case would just have a bit longer loading time, which should be fine since this is not the usual case, but rather some error case, that would not happen all the time.

dshokouhi:
> I think having the html version in sync with the android version is too complicated and could always break in the future.
> This is why I want to separate this and let the android app draw the whole splash screen, skipping the html splash screen in total.

[...]

PR #4937, opened 2024-12-27
Update splash screen icon to match style/size of frontend

jpelgrom:
## Summary

Use the same style and size of icon on the splash screen as the frontend to make the transition smoother. This doesn't include the positioning - it may (almost) line up but no guarantees as that requires more device testing and seems to require us going edge to edge on modern devices which is currently blocked by webview / frontend support (and edge to edge shouldn't be combined with this PR). Related: #3142

Also update onboarding screen with the same icon, and don't set splash screen icon as placeholder in camera/media player widgets as they are not visible anyway.

## Screenshots

https://github.com/user-attachments/assets/f6871aee-5e28-4998-8983-a618bf71b344

|    |Light|Dark|
|----|----|----|
|Updated onboarding screen|Home Assistant Companion app onboarding screen with a blue Home Assistant logo like the frontend instead of a white icon like the app logo, light mode|Home Assistant Companion app onboarding screen with a blue Home Assistant logo like the frontend instead of a white icon like the app logo, dark mode|
|Version setting row uses notification icon|Application version settings row with a small notification style icon, light mode|Application version settings row with a small notification style icon, dark mode|

n/a


dshokouhi (approved):
wow that transition is very smooth!

PR #4952, opened 2025-01-04
Always request Bluetooth permission for Improv if 1/2 is granted

### LINKED ISSUE #4936: Voice PE / Improv Wi-Fi device Not Discovered
2024.12.1-full
**Android version(s):**
Android 15
**Device model(s):**
Pixel 6 Pro
**Home Assistant version:**
2024.12.4

Companion App will not discover the Voice PE device.  Voice PE displaying warm twinkle animation.  App precise location permission is set to allow all the time. 

**Companion App Logs:**

```

```

**Additional information:**
Android pair new Bluetooth device does show "home-assistan-xxxxxx" as an available device but manual pairing does not complete.

jpelgrom:
## Summary

Fixes #4936. The issue is caused by the following combination:
1. from the Nearby devices permission group _a_ Bluetooth permission is granted, so the system settings show the permission as on, but not _all_ permissions are granted
2. the in-app permission prompt was shown at least twice so it is not shown again

It results in the user thinking everything is good, while the app doesn't continue and doesn't request additional permissions -> confusion why nothing is discovered.

This PR tries to fix it by checking when 1) not all permissions are granted and 2) the prompt was shown at least twice, if we need just _one_ Bluetooth permission, and if so always requesting that permission. Because Improv needs _two_ Bluetooth permissions that means something from the Nearby devices group was granted so when requesting it Android automatically grants it without asking the user.

For testing this you probably want to use some adb commands:
- `adb shell dumpsys package io.homeassistant.companion.android.debug | grep permission` to verify what exactly is granted
- `adb shell pm revoke io.homeassistant.companion.android.debug android.permission.BLUETOOTH_CONNECT` to revoke one but not the other Bluetooth permission (or the `grant` equivalent)
adb output with Improv related permissions highlighted

## Screenshots

n/a

n/a

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -157,6 +157,12 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                 downloadFile(downloadFileUrl, downloadFileContentDisposition, downloadFileMimetype)
             }
         }
+    private val requestImprovPermission =
+        registerForActivityResult(ActivityResultContracts.RequestPermission()) { isGranted ->
+            if (isGranted) {
+                presenter.startScanningForImprov()
+            }
+        }
     private val writeNfcTag = registerForActivityResult(WriteNfcTag()) { messageId ->
         sendExternalBusMessage(
             ExternalBusMessage(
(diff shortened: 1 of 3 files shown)


jpelgrom:
FYI some additional details:

```kotlin
packageManager.getPlatformPermissionsForGroup(
    android.Manifest.permission_group.NEARBY_DEVICES,
    mainExecutor
)
```

returns the following list: android.permission.BLUETOOTH_ADVERTISE, android.permission.BLUETOOTH_CONNECT, android.permission.BLUETOOTH_SCAN, android.permission.UWB_RANGING, android.permission.NEARBY_WIFI_DEVICES.

The app does not use UWB_RANGING or NEARBY_WIFI_DEVICES anywhere. Improv requires BLUETOOTH_SCAN and BLUETOOTH_CONNECT. Other Bluetooth permissions are used by some parts of the app in varying combinations.

dshokouhi (approved):

Issue #2555, opened 2022-05-30
WearOS Thermostat Controls/Entity


eric10k93:
**Is your feature request related to a problem? Please describe.**
I'm currently unable to control a climate entity via WearOS.  I would like to have thermostat controls.  I have a Samsung Galaxy Watch 4 Classic, running WearOS 3.0.

**Describe the solution you'd like**
Years ago, Nest had a WearOS application which allowed me to control my thermostat via my watch, however, they deleted their app from the Play Store and it's not been available for some time.  I'd like similar functionality from Home Assistant to replace that.

Ideally, I would like to be able to select a climate entity from my watch, and then have a display pop up with the current temperature in the house, as well as the temperature set points (for heat and cool).  Then I would like to be able to select the heat or cool set point and use a plus and minus button (or up and down arrows) to adjust that set point.

**Describe alternatives you've considered, if any**
Currently, my only option would be to create multiple button entities, where pressing a single button would adjust the temperature a degree at a time in a given direction, but this gets much more complicated when accounting for multiple set points and multiple temperature adjustment directions.

**Additional context**
A minimized/minimal version of the existing Lovelace thermostat card would be perfect.

dnestico:
+1

Weax:
No updates on this for 1.5 year

DaFern:
+1

PR #4959, opened 2025-01-05
New thermostat tile for Wear OS

Martreides:
## Summary
This pull request is intended to add a thermostat tile to the Wear OS app. It shows the target and current temperature, and allows a user to change the target temperature. The high level set up of the implementation is based on the existing tiles, mostly the camera tile.

## Screenshots
Screenshot_20250105_174129_One UI Watch Home

Documentation: home-assistant/companion.home-assistant#1155

## Any other notes

This is my first contribution to an Android project. A good review of the PR is appreciated.


jpelgrom:
Here's a tip: running `./gradlew ktlintFormat` locally will fix most of the errors the linter is giving you automatically (see project README file).

Don't forget to mark the PR as ready to review when you're done.

Martreides:
Thanks for the tip! Before my last commit I had one issue left, which should be fixed in my latest commit. Good to know for future commits though!

dshokouhi:
So just a thought but maybe we should call this a Climate Tile instead of Thermostat tile to be more precise? Users can add any climate entity but does not specifically need to be a thermostat?

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,236 @@
...
+        if (targetTemp == "null") targetTemp = entity?.attributes?.get("temperature").toString()
+
+        if (lastId == "Up" || lastId == "Down") {
+            val entityStr = entity?.entityId.toString()
+            val stepSize = entity?.attributes?.get("target_temp_step").toString().toFloat()
+            val updatedTargetTemp = targetTemp.toFloat() + if (lastId == "Up") +stepSize else -stepSize
+
+            serverManager.integrationRepository().callAction(
+                entityStr.split(".")[0],
+                "set_temperature",
```
dshokouhi, line 76: Looks like we need to do a state check to prevent a possible error?

https://www.home-assistant.io/integrations/climate/#action-climateset_temperature

`New target temperature for climate device (commonly referred to as a setpoint). Do not use if hvac_mode is heat_cool.`

dshokouhi:
Thanks for clarifying it does make sense in the current use case however one thing to consider is that later on changing the name may lead to a breaking change as the tile path is in the manifest. Changing that later may have the undesired effect of removing the tile as the path can no longer be referenced.

With that said I can see that multiple tiles are allowed to be added for this tile. Maybe we should consider showing the entity friendly name or icon to help distinguish between multiple tiles? Also did you consider showing the state? Might be helpful to know if the thermostat is on for example?

Thanks for the PR as well 🙏 , looks like this will let us close #2555 based on current functionality.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: Good point, I wasn't aware of that. I will add a state check later today to handle that. Any best practice on what to do if that hvac_mode is heat_cool? From the docs I understand that in that case both target_temp_high and target_temp_low should be provided, but that is not supported by the tile currently.

dshokouhi: Hmm might be best to skip that state for now 🤔 and  maybe just hide controls? Dont think it makes sense at all to set both low and high to the same value right?

dshokouhi:
Looks like we also need to check if the climate entity supports changing the temp

https://github.com/home-assistant/core/blob/dev/homeassistant/components/climate/__init__.py#L643

 https://github.com/home-assistant/core/blob/dev/homeassistant/components/climate/const.py#L144

We do already have some of that logic in our device controls implementation you may refer to

https://github.com/home-assistant/android/blob/master/app/src/main/java/io/homeassistant/companion/android/controls/ClimateControl.kt

Martreides:
>Looks like we also need to check if the climate entity supports changing the temp

I guess we can combine your earlier comment into this as well right? We do one check to determine whether the entity can be controlled by this tile and use that throughout the code. I will have to dive into this for a bit. I see my own climate entity reports as 385, which is target_temperature, turn_off and turn_on. Since the ClimateControl code that implements this is part of the phone app I guess we cannot use any part of it directly right?

>Maybe we should consider showing the entity friendly name or icon to help distinguish between multiple tiles?

I don´t entirely follow what you mean here. Do you mean to show the name or icon on the tile?

>Also did you consider showing the state?

Ideally I would like to show it the same as in the browser, it shows the state and glows if heating. The glowing is a bit too complicated for me at this point, but I think the state should be quite easy to add.

>looks like this will let us close https://github.com/home-assistant/android/issues/2555

Nice, I wasn't aware that issue existed!

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: I don't have any device myself that has a high and low. But, I can imagine that it might not be good to set those to the same value. I guess it would always be either cooling or heating because it's never between the min and max :sweat_smile:.

dshokouhi:
> I guess we can combine your earlier comment into this as well right? We do one check to determine whether the entity can be controlled by this tile and use that throughout the code.

Yup that sounds like a good approach



> I will have to dive into this for a bit. I see my own climate entity reports as 385, which is target_temperature, turn_off and turn_on. Since the ClimateControl code that implements this is part of the phone app I guess we cannot use any part of it directly right?

So not directly but you can move some of that logic to our [Entity extensions](https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt) and then reuse that elsewhere. Helps keep all that business logic in one place to easily reference :)



> I don´t entirely follow what you mean here. Do you mean to show the name or icon on the tile?

Yes something to help differentiate between multiple tiles being added otherwise its a guessing game :) Imagine a user adding a tile for upstairs and downstairs thermostats how do they tell them apart quickly


> Ideally I would like to show it the same as in the browser, it shows the state and glows if heating. The glowing is a bit too complicated for me at this point, but I think the state should be quite easy to add.

Hmm maybe instead of glow we match the frontend color? They seem to use orange for heat, blue for cool and white for idle/off. We typically try to match frontend behavior to help meet user expectations

image

image

image

Martreides:
The last commits contains changes to the layout of the tile. It now shows the friendly name of the entity at the top. I'm not completely sure on whether each entity has a friendly name though. It also shows the state of the entity (Idle, Heating or Cooling), based on the state the color of the state and the actual temperature changes (this reflects the behaviour of the card).
Screenshot_20250106_224748
Screenshot_20250106_224736

dshokouhi:
Latest screenshot looks great! Does make you wonder how _closely_ the tile should try to match the frontend here. Hard to consider things here too because we cant scroll so we have limited space

> I think it would be nice as well if we add the unit (C or F) to the tile. However, I am struggling with figuring out how to determine whether to use C or F. In the docs I do see a required field temperature_unit, but at least my climate entity does not have it.

hmm mine does not either, maybe for the unit it could be best to go by the config response because on my end i have no control over the unit like we can with sensors

https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/entities/GetConfigResponse.kt#L10

Martreides:
>Does make you wonder how closely the tile should try to match the frontend here. Hard to consider things here too because we cant scroll so we have limited space

Indeed! I think this looks good, but we should not make it more crowded.

>We actually handle that by defaulting to entity ID if not provided

Great! I am thinking about adding a setting so users can choose whether to show the name. Myself I would prefer it without as I only have a single climate entity. That should be quite easy to add though.

>hmm mine does not either, maybe for the unit it could be best to go by the config response because on my end i have no control over the unit like we can with sensors

Thanks! Will have a look at that and the other changes tomorrow evening.

dshokouhi:
> but we should not make it more crowded.

I agree however I cannot escape the thought of users asking about controlling the state too 🙈 . For this PR not necessary as the MVP solves a good use case, but in the future we can consider a button to toggle through the states just like we do with device controls. Mentioning this here so others can see we discussed it. Given the state of the app its not required for this tile as users can still get by using scripts and such on the watch for state control

Should we consider updating the UI to match the frontend when the device is off? State control seems to only be allowed when it is `off` so not sure if its ok to send the set temp action when it is off.

Side note: we can probably adapt the changes here for other domains like media player, number etc... for additional tiles in the future :) Great work!!

Martreides:
>Should we consider updating the UI to match the frontend when the device is off? State control seems to only be allowed when it is off so not sure if its ok to send the set temp action when it is off.

I'm thinking about showing "--.-" instead of the target temperature, keep the current temperature and hide the "+" and "-" button. That should align quite well to what we have in the card. Currently I do this by checking whether the hvacAction == "Off". However, my climate entity does not really have an off state so I can't test that. Do you have an idea whether this should work?

> but in the future we can consider a button to toggle through the states just like we do with device controls

It would be nice to show the action (cooling or heating) by coloring the edge of the tile. That should be doable and will free up some room on the tile. I will leave that as nice to have for now though.

>hmm mine does not either, maybe for the unit it could be best to go by the config response because on my end i have no control over the unit like we can with sensors

Btw, this works!

Martreides:
It ended up being a bit more difficult than I expected, but I have the toggle for showing the friendly name of the entity in😄.
Screenshot_20250107_223758

wear/src/main/java/io/homeassistant/companion/android/home/views/SetThermostatTileView.kt
```diff
@@ -85,6 +88,23 @@ fun SetThermostatTileView(
                     onClick = onSelectRefreshInterval
                 )
             }
+            item {
+                SwitchButton(
+                    modifier = Modifier.fillMaxWidth(),
+                    checked = tile?.showEntityName == true,
+                    onCheckedChange = {
```
Martreides, line 95: @dshokouhi, could you have a look at how I ended up handling the call back? I had some issues here because the callback only accepts functions with just a Boolean parameter. To work around that I ended up doing it this wat.

Martreides:
How do you like these? The top one is cooling, bottom one is heating. If idle it would not show a colored ring but just say "Idle", as in the previous version.
Screenshot_20250108_193952_One UI Watch Home
Screenshot_20250108_193753_One UI Watch Home

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    private suspend fun timeline(tileConfig: ThermostatTile?, targetTemperature: String): Timeline = Timeline.fromLayoutElement(
+        LayoutElementBuilders.Box.Builder().apply {
+            val entity = tileConfig?.entityId?.let {
+                serverManager.integrationRepository().getEntity(it)
+            }
+
+            val currentTemperature = entity?.attributes?.get("current_temperature").toString()
+            val hvacAction = entity?.attributes?.get("hvac_action").toString().replaceFirstChar { if (it.isLowerCase()) it.titlecase() else it.toString() }
```
jpelgrom, line 163: This is not translatable. Try adding it to the Entity.kt functions (in common) returning a string instead - like a very simple [`friendlyState`](https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/data/integration/Entity.kt#L794).

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,286 @@
...
+            val friendlyName = entity?.attributes?.get("friendly_name").toString()
```
jpelgrom, line 164: Use the extension function `friendlyName` instead

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,286 @@
...
+            val config = serverManager.webSocketRepository().getConfig()
+            val temperatureUnit = config?.unitSystem?.getValue("temperature").toString()
+
+            val hvacActionColor = when (hvacAction) {
+                "Idle" -> getColor(R.color.colorWidgetButtonLabelWhite)
+                "Heating" -> getColor(R.color.colorDeviceControlsThermostatHeat)
+                "Cooling" -> getColor(R.color.colorDeviceControlsDefaultOn)
+                else -> getColor(R.color.colorWidgetButtonLabelWhite)
+            }
```
jpelgrom, line 173: This shouldn't use strings the user is shown, but the raw state attribute like `idle`.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,286 @@
...
+        serviceScope.cancel()
+    }
+
+    private suspend fun timeline(tileConfig: ThermostatTile?, targetTemperature: String): Timeline = Timeline.fromLayoutElement(
+        LayoutElementBuilders.Box.Builder().apply {
+            val entity = tileConfig?.entityId?.let {
+                serverManager.integrationRepository().getEntity(it)
+            }
+
+            val currentTemperature = entity?.attributes?.get("current_temperature").toString()
```
jpelgrom, line 162: L162-L173 are not necessary if no entity is configured. Move that code down so the empty tile is returned earlier, saving resources.

jpelgrom (commented):
> thermostat_tile_example.png

This should be a realistic preview of the tile like in your last comment (but maybe without the title so we don't have to deal with localizing the image).

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,286 @@
...
+                    LayoutElementBuilders.Column.Builder()
+                        .addContent(
+                            LayoutElementBuilders.Text.Builder()
+                                .setText(hvacAction)
+                                .setFontStyle(LayoutElementBuilders.FontStyle.Builder().setColor(ColorBuilders.argb(hvacActionColor)).build())
+                                .build()
+                        )
+                        .addContent(
+                            LayoutElementBuilders.Text.Builder()
+                                .setText(if (hvacAction == "Off") "--.- $temperatureUnit" else "$targetTemperature $temperatureUnit")
```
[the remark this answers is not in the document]
dshokouhi: haha no decimals here :)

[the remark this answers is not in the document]
Martreides: Is it acceptable to show `--` for both Celsius and Fahrenheit? Or would it ideally dependent on the unit?

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,286 @@
...
+            // Refresh button
+            addContent(getRefreshButton())
+            setModifiers(getRefreshModifiers())
+        }.build()
+    )
+
+    private fun getTempUpButton(): LayoutElement {
+        val clickable = Clickable.Builder()
+            .setOnClick(ActionBuilders.LoadAction.Builder().build())
+            .setId("Up")
```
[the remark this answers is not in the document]
Martreides: Is it okay to define the constant at the top of ThermostatTile.kt or should that be done somewhere else? Also, is it okay to hardcode it to "Up" or should I refer to strings.xml?

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,286 @@
...
+                                        .build()
+                                )
+                                .build()
+                        )
+                        .addContent(
+                            LayoutElementBuilders.Spacer.Builder()
+                                .setHeight(DimensionBuilders.dp(10f)).build()
+                        )
+                        .addContent(
+                            if (hvacAction != "Off") {
```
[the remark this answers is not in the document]
Martreides: If I understand correctly `entity?.attributes?.get("hvac_action")` returns the state attritube `off` when the entity is off. How can I use that in an if statement? I have to compare it against the off state right, but not sure where to get that from.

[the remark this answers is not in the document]
jpelgrom: Just use `entity?.attributes?.get("hvac_action")` again, or assign `val hvacActionValue = entity?.attributes?.get("hvac_action")`, en then `if (hvacActionValue != "off")`.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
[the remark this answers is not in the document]
jpelgrom: You add it to the companion object at the top, alongside the other `const val`. The actual value of the string doesn't matter, but it should be in the file. Example `private const val TAP_ACTION_UP = "up"`.

wear/src/main/java/io/homeassistant/companion/android/home/views/SetThermostatTileView.kt
```diff
@@ -0,0 +1,110 @@
...
+                        onNameEnabled(tile?.id.toString().toInt(), it)
+                    },
+                    label = { Text(stringResource(commonR.string.setting_entity_name_on_tile)) },
+                    icon = {
+                        Image(
+                            asset = CommunityMaterial.Icon3.cmd_page_layout_header,
```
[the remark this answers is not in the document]
Martreides: Good idea, I will also change the button to checkboxButton to align closer to the shortcuts tile.

Martreides:
@jpelgrom, what is the best practice on resolving your review comments? I have some other changes pending as well on the layout of the tile. Should I make specific commits per review comments or can I combine them together? First time working on a project like this :) .

wear/src/main/java/io/homeassistant/companion/android/home/views/SetThermostatTileView.kt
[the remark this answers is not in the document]
jpelgrom: No checkbox is actually wrong there because it is a binary choice, not a list, this should remain a toggle :)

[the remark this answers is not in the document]
Martreides: Ah, got to change it back a bit then haha!

jpelgrom:
You can use individual commits or one big commit for all review comments, personal preference. Just don't push other unrelated changes in the same commit, or rewrite commit history, so it is easy to check changes for review.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,286 @@
...
+        }
+
+        if (requestParams.currentState.lastClickableId.isNotEmpty()) {
+            if (wearPrefsRepository.getWearHapticFeedback()) hapticClick(applicationContext)
+        }
+
+        val lastId = requestParams.currentState.lastClickableId
+
+        var targetTemp = tileConfig?.targetTemperature.toString()
+        if (targetTemp == "null") targetTemp = entity?.attributes?.get("temperature").toString()
```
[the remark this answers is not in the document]
Martreides: [52493ce](https://github.com/home-assistant/android/pull/4959/commits/52493ce74912597957743468dd189df587fb435d)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [728e6c5](https://github.com/home-assistant/android/pull/4959/commits/728e6c5f076a0291f90b7044e0149fb2f6e79c82)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [728e6c5](https://github.com/home-assistant/android/pull/4959/commits/728e6c5f076a0291f90b7044e0149fb2f6e79c82)

Martreides: Changed, it does still require to add .toString() though.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [a4264da](https://github.com/home-assistant/android/pull/4959/commits/a4264daaa5bb3da8597e3ff6a8450065fead85d2)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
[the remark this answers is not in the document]
Martreides: [31467e1](https://github.com/home-assistant/android/pull/4959/commits/31467e173eec4ec4f4eaeb18badbc50dfa4c9091)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,284 @@
...
+                val currentTemperature = entity?.attributes?.get("current_temperature").toString()
+                val hvacAction = entity?.attributes?.get("hvac_action").toString().replaceFirstChar { if (it.isLowerCase()) it.titlecase() else it.toString() }
+                val config = serverManager.webSocketRepository().getConfig()
+                val temperatureUnit = config?.unitSystem?.getValue("temperature").toString()
+                Log.d(TAG, hvacAction)
+
+                val hvacActionColor = when (hvacAction) {
+                    "Heating" -> getColor(R.color.colorDeviceControlsThermostatHeat)
+                    "Cooling" -> getColor(R.color.colorDeviceControlsDefaultOn)
+                    else -> 0x00000000
```
jpelgrom, line 178: Also use a color here (`android.R.color.black`)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [4cdc938](https://github.com/home-assistant/android/pull/4959/commits/4cdc93844c3c843355665d77d9ab6ba4469b14a0)

For now I went with the easy solution and actually use friendlyState for the text shown on the tile. This does mean that it now shows "Heat" instead of "Heating" and "Cool" instead of "Cooling". For me that is acceptable for the current stage. The other logic now relies on the unadjusted state (heating, cooling, idle, off).

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,286 @@
...
+                    ColorBuilders.argb(getColor(R.color.colorWidgetButtonLabelBlack))
+                )
+            )
+            .build()
+    }
+
+    private fun getTempDownButton(): LayoutElement {
+        val clickable = Clickable.Builder()
+            .setOnClick(ActionBuilders.LoadAction.Builder().build())
+            .setId("Down")
```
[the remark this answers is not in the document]
Martreides: [d84ce37](https://github.com/home-assistant/android/pull/4959/commits/d84ce377a5031e5d463488576ddac91c339e7e18)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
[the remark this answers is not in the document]
Martreides: [d84ce37](https://github.com/home-assistant/android/pull/4959/commits/d84ce377a5031e5d463488576ddac91c339e7e18)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [4cdc938](https://github.com/home-assistant/android/pull/4959/commits/4cdc93844c3c843355665d77d9ab6ba4469b14a0)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: See [9039b70](https://github.com/home-assistant/android/pull/4959/commits/9039b7017973848eab00bd808dd32e56d153807e).
I ran into a bit of trouble implementing this in Entity.kt since this would only be applicable to climate entities. I couldn't find a decent solution to making that work. I ended up doing a when statement in ThermostatTile.kt mapping the hvac_action to strings from strings.xml.

Martreides:
Thanks a lot for the thorough review! I have responded to all of the individual comments. Let me know if you have any follow-up or additional feedback!

At the current stage I am quite happy with how the tile looks and functions. If I have time tomorrow I would like to do some test on emulated devices with different screens (sizes and shapes).

jpelgrom:
Could you put the example tile image in tinypng.com and update it? Should be able to get it smaller, not a lot of color so probably compresses nicely without too much artifacts.

Martreides:
>Could you put the example tile image in tinypng.com and update it? Should be able to get it smaller, not a lot of color so probably compresses nicely without too much artifacts.

63% saved 😄

Martreides:
I have added a specific layout for when the tile fails to load. Had this happen myself due to the entity being unavailable. It now shows an error message and a refresh button.

With the above included, the tile is good to go from my perspective.

Martreides:
I had to reshuffle the code a bit. If no entity was selected the code would fail before reaching the point at which it would output the "select entity" content.

common/src/main/res/values/strings.xml
```diff
@@ -1355,4 +1357,16 @@
...
+    <string name="thermostat_tile_desc">See and change the thermostat temperature.</string>
+    <string name="thermostat_tile_no_tiles_yet">There are no thermostat tiles added yet - add one from the watch face to set it up</string>
+    <string name="thermostat_tile_no_entity_yet">Edit the tile settings and select a thermostat to show</string>
+    <string name="thermostat_tile_n">Thermostat tile #%d</string>
+    <string name="thermostat_tile_log_in">Log in to add a thermostat tile</string>
+    <string name="thermostat_tile">Thermostat tile</string>
+    <string name="climate_heating">Heating</string>
+    <string name="climate_cooling">Cooling</string>
+    <string name="climate_idle">Idle</string>
+    <string name="climate_off">Off</string>
```
jpelgrom, line 1371: Try using the existing `state_*` strings for these as they are generic/are used in the same context/have the same meaning. Duplicating strings means more work for translating.

wear/src/main/java/io/homeassistant/companion/android/home/views/SelectThermostatTileView.kt
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+}
+
+@Preview(device = WearDevices.LARGE_ROUND)
+@Composable
+private fun PreviewSelectThermostatTileViewOne() {
+    SelectThermostatTileView(
+        tiles = listOf(
+            ThermostatTile(id = 1, entityId = "camera.buienradar", refreshInterval = 300)
```
jpelgrom, line 61: This entity ID should be updated to a more realistic `climate.living_room` or something like that ;)

wear/src/main/java/io/homeassistant/companion/android/home/views/SetThermostatTileView.kt
```diff
@@ -0,0 +1,115 @@
...
+    WearAppTheme {
+        ThemeLazyColumn {
+            item {
+                ListHeader(commonR.string.thermostat_tile)
+            }
+            item {
+                val entity = tile?.entityId?.let { tileEntityId ->
+                    entities?.firstOrNull { it.entityId == tileEntityId }
+                }
+                val icon = entity?.getIcon(LocalContext.current) ?: CommunityMaterial.Icon3.cmd_thermometer
```
jpelgrom, line 46: Why use `thermometer` as the fallback instead of the domain default `thermostat`?

wear/src/main/java/io/homeassistant/companion/android/home/views/SettingsView.kt
```diff
@@ -189,6 +190,13 @@ fun SettingsView(
                     onClick = onClickTemplateTiles
                 )
             }
+            item {
+                SecondarySettingsChip(
+                    icon = CommunityMaterial.Icon3.cmd_thermometer,
```
jpelgrom, line 195: Why use `thermometer` as the fallback instead of the domain default `thermostat`?

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,303 @@
...
+            (tileConfig?.refreshInterval != null && tileConfig.refreshInterval!! <= 1) -> 0
+            tileConfig?.refreshInterval != null -> tileConfig.refreshInterval!!
+            else -> DEFAULT_REFRESH_INTERVAL
+        }
+
+        val tile = Tile.Builder()
+            .setResourcesVersion("$TAG$tileId.${System.currentTimeMillis()}")
+            .setFreshnessIntervalMillis(TimeUnit.SECONDS.toMillis(freshness))
+
+        if (tileConfig?.entityId.isNullOrBlank()) {
```
jpelgrom, line 76: The first check you do, should be the logged out check. Right now, it wouldn't show the screen that you need to log in after adding a tile because the default value for entity ID is `null` so you show the select an entity screen. (Not showing login required is a failure in Google's review process.)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,303 @@
...
+                            LayoutElementBuilders.Text.Builder()
+                                .setText(getString(R.string.thermostat_tile_no_entity_yet))
+                                .setMaxLines(10)
+                                .build()
+                        ).build()
+                )
+            ).build()
+        } else {
+            try {
+                val entity = tileConfig?.entityId?.let {
```
jpelgrom, line 90: Because of your if/else above, you know that `tileConfig != null` so you can write `tileConfig.entityId` instead of including `?` every time (unnecessary safe call). Android Studio will also highlight this for you with a squiggly line:

image

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,303 @@
...
+                    serverManager.integrationRepository().getEntity(it)
+                }
+
+                val lastId = requestParams.currentState.lastClickableId
+                var targetTemp = tileConfig?.targetTemperature ?: entity?.attributes?.get("temperature").toString().toFloat()
+
+                if (lastId == TAP_ACTION_UP || lastId == TAP_ACTION_DOWN) {
+                    val entityStr = entity?.entityId.toString()
+                    val stepSize = entity?.attributes?.get("target_temp_step").toString().toFloat()
+                    val updatedTargetTemp = targetTemp + if (lastId == "Up") +stepSize else -stepSize
```
jpelgrom, line 100: `if (lastId == "Up")` should use the constant instead of `"Up"`

common/src/main/res/values/strings.xml
```diff
@@ -1082,6 +1083,7 @@
     <string name="tile_icon">Tile icon</string>
     <string name="tile_icon_original">Use entity icon</string>
     <string name="tile_select">Select a tile to edit</string>
+    <string name="tile_fetch_entity_error">Cannot fetch entity</string>
```
jpelgrom, line 1086: This isn't very helpful if all your tiles failed to refresh, maybe include the name?

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,303 @@
...
+                "heating" -> getColor(R.color.colorDeviceControlsThermostatHeat)
+                "cooling" -> getColor(R.color.colorDeviceControlsDefaultOn)
+                else -> 0x00000000
+            }
+            val friendlyHvacAction = when (hvacAction) {
+                "heating" -> getString(R.string.climate_heating)
+                "cooling" -> getString(R.string.climate_cooling)
+                "idle" -> getString(R.string.climate_idle)
+                "off" -> getString(R.string.climate_off)
+                else -> hvacAction
```
jpelgrom, line 208:
```suggestion
                else -> hvacAction.replaceFirstChar { it.uppercase() }
```

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,303 @@
...
+                            .setColor(ColorBuilders.argb(0x00000000)) // Fully transparent
+                            .build()
+                    )
+                    .addContent(
+                        LayoutElementBuilders.ArcText.Builder()
+                            .setText(if (tileConfig?.showEntityName == true) entity?.friendlyName.toString() else "")
+                            .build()
+                    )
+                    .build()
+            )
```
jpelgrom, line 279: Why not put all of this inside an `if (tileConfig.showEntityName == true)` statement? That way you avoid having the content if it is off, maybe saving resources.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,303 @@
...
+                .delete(requestParams.tileId)
+        }
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        serviceScope.cancel()
+    }
+
+    private suspend fun timeline(tileConfig: ThermostatTile?, targetTemperature: Float): Timeline = Timeline.fromLayoutElement(
```
jpelgrom, line 187: Isn't `tileConfig` always non-null at this point (so `tileConfig: ThermostatTile` could be used)?

jpelgrom (commented):
> If I have time tomorrow I would like to do some test on emulated devices with different screens (sizes and shapes).

Did you end up testing on other screen sizes? Mainly interested in small screens because you are hardcoding quite a lot of units for spacing and the screenshot looks like it just fits and is not the smallest screen.

common/src/main/res/values/strings.xml
Martreides: [db9bfe0](https://github.com/home-assistant/android/pull/4959/commits/db9bfe07905b22c1fdf22c7a32af960d10122928)

jpelgrom: As cooling/heating are very specific climate states, I don't see a need to make it more generic.

wear/src/main/java/io/homeassistant/companion/android/home/views/SelectThermostatTileView.kt
Martreides: living_room it is!
[7d65b20](https://github.com/home-assistant/android/pull/4959/commits/7d65b205ed327a924dd622d7e3a2c84d49711aa0)

wear/src/main/java/io/homeassistant/companion/android/home/views/SettingsView.kt
Martreides: [09d9905](https://github.com/home-assistant/android/pull/4959/commits/09d9905c78d1c0bdfef997645061ffe54616cd67)

wear/src/main/java/io/homeassistant/companion/android/home/views/SetThermostatTileView.kt
Martreides: [09d9905](https://github.com/home-assistant/android/pull/4959/commits/09d9905c78d1c0bdfef997645061ffe54616cd67)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: I will have a look!
It does seem my configuration is different though:
image

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [273e42b](https://github.com/home-assistant/android/pull/4959/commits/273e42b799d2e0dd49bbe070c299fec4e060ba42)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [4b20946](https://github.com/home-assistant/android/pull/4959/commits/4b209463ca9e90933ec2e8139e387f6e872ed893)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [7e8a7aa](https://github.com/home-assistant/android/pull/4959/commits/7e8a7aa05d2053ee1149d989b4652130e6526aad)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [7d0f284](https://github.com/home-assistant/android/pull/4959/commits/7d0f284076c9e55bca6c4340f08c802033f85dba)
[1469042](https://github.com/home-assistant/android/pull/4959/commits/14690425b305204eca47b76d9bf64c19d3c1db43)

Martreides:
Large round:
large round
Small round:
small round
Square:
Screenshot_20250110_172048

Do we have any requirements for square screens? 😅

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: I'm not sure why but I actually get errors when making this change.
image

jpelgrom: Well right now you would get an error because it is no longer true because of the changes you've made :) You can still make similar changes if you address https://github.com/home-assistant/android/pull/4959/files#r1912244604.

jpelgrom:
> Do we have any requirements for square screens? 😅

Realistically no one is going to have a square Wear OS watch right now so the current layout looks fine to me.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [64ca70e](https://github.com/home-assistant/android/pull/4959/commits/64ca70e2b2a660117b9f6c70b9a9d5f1f450dc84)
I have made some changes, but I'm not really happy with it. Regardless of what I do I get no hint from Android Studio for removing safe calls. In the code after the last commit I actually get a bunch of warnings, but it does still build. Not sure whether to keep it like this.

common/src/main/res/values/strings.xml
Martreides: It's a bit tricky to solve this. The function I use to create the tile when it fails to load the entity is a shared function that expect a string resource (primaryLayoutTimeline in TileViews.kt). I don't see an easy possibility to include the entity name in there without rewriting the function to accept both strings and string resources.

jpelgrom: I didn't check that, sorry. Let's keep it as it is now then, to avoid the PR getting even bigger.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: @jpelgrom, what is your opinion on my last commit regarding the safe calls?

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,307 @@
...
+                    }
+
+                    tile.setTileTimeline(
+                        timeline(
+                            tileConfig,
+                            targetTemp
+                        )
+                    ).build()
+                } catch (e: Exception) {
+                    Log.e(TAG, "Unable to fetch entity ${tileConfig?.entityId}", e)
```
jpelgrom, line 135:
```suggestion
                    Log.e(TAG, "Unable to fetch entity ${tileConfig.entityId}", e)
```

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
jpelgrom: That looks a lot better, but now I found a couple of other points to comment on :) Let's mark this one as resolved as it is was 99% addressed and the remaining one has a comment.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,307 @@
...
+    override fun onDestroy() {
+        super.onDestroy()
+        serviceScope.cancel()
+    }
+
+    private suspend fun timeline(tileConfig: ThermostatTile, targetTemperature: Float): Timeline = Timeline.fromLayoutElement(
+        LayoutElementBuilders.Box.Builder().apply {
+            val entity = tileConfig.entityId?.let {
+                serverManager.integrationRepository().getEntity(it)
+            }
```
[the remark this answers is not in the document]
Martreides: How could I pass the entity to the timeline function? I'm struggling a bit with the type of the parameter. The following seems to not work:
`private suspend fun timeline(tileConfig: ThermostatTile, entity: Entity`

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: I still don't really see at what point it becomes okay to stop using safe calls. Android Studio keep giving me a lot of warning with the current code.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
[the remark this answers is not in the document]
jpelgrom: If you want to add it all to the current line 189 you can do:

```kotlin
private suspend fun timeline(tileConfig: ThermostatTile, entity: Entity<Map<String, Any>>, targetTemperature: Float): Timeline = Timeline.fromLayoutElement(
```

(but given the length adding a newline after = is probably nice/ktlint might reformat)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
jpelgrom: It is ok once you're sure that the variable is not `null`. In this case: because a `null` tileConfig would be handled by the if statement in line 86.

Not sure why your Android Studio isn't showing you these and is actively showing warnings, I can see how it might be unintuitive then. Search results online suggest power save mode should be off and/or invalidating caches might help.

Martreides: I might be misunderstanding this, but isn't it so that if tileConfig is null the if statement would immediately skip to the else part? If so, it could still be null in the else clause of the if/else.

jpelgrom: No, `isNullOrBlank()` also applies to `tileConfig`. Here's a simplified example that shows it: https://pl.kotl.in/zyrFZF9_c

Martreides: Ah, got it! I was under the impression `isNullOrBlank()` only applied to `entityId`. Thanks!

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,307 @@
...
+                    )
+                ).build()
+            } else {
+                try {
+                    val entity = tileConfig.entityId?.let {
+                        serverManager.integrationRepository().getEntity(it)
+                    }
+
+                    val lastId = requestParams.currentState.lastClickableId
+                    var targetTemp = tileConfig.targetTemperature ?: entity?.attributes?.get("temperature").toString().toFloat()
```
dshokouhi, line 105: When i run this locally i get an error here and my entity fails to load when my thermostat is off, when it is on it is fine

image


```
2025-01-21 20:52:32.560 27854-27921 ThermostatTile          io....stant.companion.android.debug  E  Unable to fetch entity climate.thermostat (Ask Gemini)
                                                                                                    java.lang.NumberFormatException: For input string: "null"
                                                                                                    	at jdk.internal.math.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2054)
                                                                                                    	at jdk.internal.math.FloatingDecimal.parseFloat(FloatingDecimal.java:122)
                                                                                                    	at java.lang.Float.parseFloat(Float.java:485)
                                                                                                    	at io.homeassistant.companion.android.tiles.ThermostatTile$onTileRequest$1.invokeSuspend(ThermostatTile.kt:105)
                                                                                                    	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:101)
                                                                                                    	at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:113)
                                                                                                    	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:89)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:589)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:823)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:720)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:707)
```

attributes when thermostat is off:

```
hvac_modes: off, heat, cool
min_temp: 32
max_temp: 122
fan_modes: Auto low, Low, Circulation
preset_modes: none, Resume (on), Energy heat, Energy cool
current_temperature: 71
temperature: null
fan_mode: Auto low
hvac_action: idle
preset_mode: none
fan_state: Idle / off
friendly_name: Thermostat 
supported_features: 409
```

wear/src/main/java/io/homeassistant/companion/android/home/views/SetThermostatTileView.kt
dshokouhi: Sorry for the delay, the callback looks good and works as expected when i test locally

wear/src/main/res/drawable/thermostat_tile_example.png
dshokouhi, line 1: Can we update the title of the thermostat to be more generic? this will be displayed to user in the tile carousel so the more generic the better :)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,307 @@
...
+
+                    if (lastId == TAP_ACTION_UP || lastId == TAP_ACTION_DOWN) {
+                        val entityStr = entity?.entityId.toString()
+                        val stepSize = entity?.attributes?.get("target_temp_step").toString().toFloat()
```
dshokouhi, line 109: When i attempt to click on a button i get a similar error from up above but for here:

```
2025-01-21 21:06:31.498 27854-28647 ThermostatTile          io....stant.companion.android.debug  E  Unable to fetch entity climate.thermostat (Ask Gemini)
                                                                                                    java.lang.NumberFormatException: For input string: "null"
                                                                                                    	at jdk.internal.math.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2054)
                                                                                                    	at jdk.internal.math.FloatingDecimal.parseFloat(FloatingDecimal.java:122)
                                                                                                    	at java.lang.Float.parseFloat(Float.java:485)
                                                                                                    	at io.homeassistant.companion.android.tiles.ThermostatTile$onTileRequest$1.invokeSuspend(ThermostatTile.kt:109)
                                                                                                    	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:101)
                                                                                                    	at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:113)
                                                                                                    	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:89)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:589)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:823)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:720)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:707)
```

current attributes:

```
hvac_modes: off, heat, cool
min_temp: 32
max_temp: 122
fan_modes: Auto low, Low, Circulation
preset_modes: none, Resume (on), Energy heat, Energy cool
current_temperature: 71
temperature: 71
fan_mode: Auto low
hvac_action: idle
preset_mode: none
fan_state: Idle / off
friendly_name: Thermostat 
supported_features: 409
```

Martreides: Ah, I should have checked whether `target_temperature_step` is required, which turns out is not the case. I will add some logic to use `target_temperature_step` if not null, otherwise use 0.5 for C and 1.0 for F. Does that make sense?

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: Thanks for checking!
Is you thermostat idle or off in this example? Based on the attributes this would currently be interpreted as being idle (`hvac_mod = idle`) in which case the tile still shows the current and target temperature. However, your thermostat does not seem to provide the target temperature when the `hvac_action` is idle.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
dshokouhi: that sounds right but having a tough time finding this mentioned anywhere in the docs or code 🤔

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
dshokouhi: in this scenario my thermostat state was `off`, on my thermostat `idle` can also be present if the thermostat is on but not actively `heating` or `cooling` for the `hvac_action`

example of state of and attributes when heat set but also idle

image

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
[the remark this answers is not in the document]
Martreides: [35d6fd5](https://github.com/home-assistant/android/pull/4959/commits/35d6fd50020e08b9985d455f9e822abcba53ebba)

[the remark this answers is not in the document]
Martreides: I now also use `check(entity != null)` to be able to say entity is never null in the timeline function.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [cead805](https://github.com/home-assistant/android/pull/4959/commits/cead80580024d94b97d8fae8c8a598c297055c03)

Martreides: image
It does specify it's default is None.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [6199198](https://github.com/home-assistant/android/pull/4959/commits/6199198930711c801b2a6ea568fb6e0d366e9105)

Martreides: I think this should be handled now, the approach has been changed quite a bit.

- Whether a temperature is shown on the tile is no longer based on the state of the entity, but only whether a temperature is available. As long as a (target) temperature is available it is shown.
- Initially I didn't take the state into account, only the hvac_mode. I have added a check, if the state is off or unavailable this is treated the same as hvac_mode off.

wear/src/main/res/drawable/thermostat_tile_example.png
Martreides: [c249c08](https://github.com/home-assistant/android/pull/4959/commits/c249c08782211d114f531d2b66bcfb4a22278be5)

dshokouhi:
Latest changes looking good here, no crashes on various states of my thermostat and the controls work as expected.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,325 @@
...
+                        val entityStr = entity.entityId
+                        var stepSize = entity.attributes["target_temp_step"]
+
+                        stepSize = if (stepSize == null && temperatureUnit == "°F") {
+                            1.0f
+                        } else if (stepSize == null) {
+                            0.5f
+                        } else {
+                            stepSize.toString().toFloat()
+                        }
```
jpelgrom, line 122: Try to avoid a `var` if you immediately reassign it

```suggestion
                        val attrStepSize = (entity.attributes["target_temp_step"] as? Number)?.toFloat()
                        val stepSize = attrStepSize ?: if (temperatureUnit ==  "°F") 1.0f else 0.5f
```

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,325 @@
...
+
+                        val updatedTargetTemp = targetTemp + if (lastId == TAP_ACTION_UP) +stepSize.toFloat() else -stepSize.toFloat()
+
+                        serverManager.integrationRepository().callAction(
+                            entityStr.split(".")[0],
```
jpelgrom, line 127: Use `entity.domain` (+ `import io.homeassistant.companion.android.common.data.integration.domain`) instead of splitting the string manually

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,325 @@
...
+                            "set_temperature",
+                            hashMapOf(
+                                "entity_id" to entityStr,
```
jpelgrom, line 130: Use `entity.entityId` directly, the extra variable isn't needed

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,325 @@
...
+                    "heating" -> getColor(R.color.colorDeviceControlsThermostatHeat)
+                    "cooling" -> getColor(R.color.colorDeviceControlsDefaultOn)
+                    else -> 0x00000000
+                }
+
+                hvacAction = when (state) {
+                    "off" -> "off"
+                    "unavailable" -> "off"
+                    else -> hvacAction
+                }
```
jpelgrom, line 221: Changing `hvacAction` to `off` when `unavailable`, also makes it _look_ like unavailable climate entities are off (see: `friendlyHvacAction`), that seems unwanted to me? Instead you should just edit line 259/264 to also disable the buttons when `unavailable`. (The code suggestion assumes you'll edit those lines.)

Try to avoid a `var` if you immediately reassign it.

```suggestion
                val hvacAction = entity.attributes["hvac_action"].toString()
                val hvacActionColor = when (hvacAction) {
                    "heating" -> getColor(R.color.colorDeviceControlsThermostatHeat)
                    "cooling" -> getColor(R.color.colorDeviceControlsDefaultOn)
                    else -> 0x00000000
                }
```

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,325 @@
...
+
+                        stepSize = if (stepSize == null && temperatureUnit == "°F") {
+                            1.0f
+                        } else if (stepSize == null) {
+                            0.5f
+                        } else {
+                            stepSize.toString().toFloat()
+                        }
+
+                        val updatedTargetTemp = targetTemp + if (lastId == TAP_ACTION_UP) +stepSize.toFloat() else -stepSize.toFloat()
```
jpelgrom, line 124: Try to avoid a `var` if you immediately reassign it, and don't convert floats to floats again ;)

```suggestion
                        val attrStepSize = (entity.attributes["target_temp_step"] as? Number)?.toFloat()
                        val stepSize = attrStepSize ?: if (temperatureUnit == "°F") 1.0f else 0.5f
                        val updatedTargetTemp = targetTemp + if (lastId == TAP_ACTION_UP) +stepSize else -stepSize
```

Martreides:
I'm away for the week, will have a look at the last comments as soon as I'm back.

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [f1a3ed7](https://github.com/home-assistant/android/pull/4959/commits/f1a3ed70eea8afc1c845e9f2ebe3448afc66e7c2)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [313f878](https://github.com/home-assistant/android/pull/4959/commits/313f8787921676d4ae7fb75523403835e218f032)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: Well, that's definitely a lot cleaner!

Martreides: [7803458](https://github.com/home-assistant/android/pull/4959/commits/780345872ec83a5ec206012289215ff631b69f41)

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
Martreides: [3d0e3e9](https://github.com/home-assistant/android/pull/4959/commits/3d0e3e9c7be365e03d6f5c4b6f80333c1f51c386)

Martreides:
I think I have covered all of the raised comments. Let me know if there are any other changes to be made!

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,311 @@
...
+                .delete(requestParams.tileId)
+        }
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        serviceScope.cancel()
+    }
+
+    private suspend fun timeline(tileConfig: ThermostatTile, entity: Entity<Map<String, Any>>, targetTemperature: Float?, temperatureUnit: String): Timeline =
```
jpelgrom, line 197: No suspending functions used

```suggestion
    private fun timeline(tileConfig: ThermostatTile, entity: Entity<Map<String, Any>>, targetTemperature: Float?, temperatureUnit: String): Timeline =
```

wear/src/main/java/io/homeassistant/companion/android/tiles/ThermostatTile.kt
```diff
@@ -0,0 +1,311 @@
...
+import androidx.wear.protolayout.material.Button
+import androidx.wear.protolayout.material.ButtonColors
+import androidx.wear.tiles.EventBuilders
+import androidx.wear.tiles.RequestBuilders
+import androidx.wear.tiles.RequestBuilders.ResourcesRequest
+import androidx.wear.tiles.TileBuilders.Tile
+import androidx.wear.tiles.TileService
+import com.google.common.util.concurrent.ListenableFuture
+import dagger.hilt.android.AndroidEntryPoint
+import io.homeassistant.companion.android.common.R
```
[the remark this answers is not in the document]
Martreides: [f6ab32d](https://github.com/home-assistant/android/pull/4959/commits/f6ab32da862d3a1a4e4849d7ad113b297b003310)

Martreides:
All done! Thanks a lot for the help @jpelgrom and @dshokouhi!

jpelgrom:
This tip still seems relevant: run `gradlew ktlintFormat` locally to quickly fix your import order errors

Martreides:
Fair point haha, I actually just tried that. But as before I get this error:
image

jpelgrom (approved):

dshokouhi (approved):

kevinleequinn:
hi all, I have tried refreshing, reloading, rebooting, reinstalling and waiting days but I still can't get the thermostat tile to show in my list of available tiles on my watch or watch app.  I have the shortcuts tile for favorites but no thermostat tile.

truly sorry to bother, but what am I missing?

dshokouhi:
@kevinleequinn make sure to update the watch app to 2025.4.3 and the tile should be available to add, if you continue ot have an issue please log a bug as merged PRs are not a place for troubleshooting

PR #4973, opened 2025-01-13
Use CustomTabs instead of WebView for login

johnvan7:
## Summary

This PR implements [Custom Tabs](https://developer.android.com/develop/ui/views/layout/webapps/overview-of-android-custom-tabs) instead of WebView for login.

Using Custom Tabs instead of WebView offers several advantages:

- Performance and Speed: Custom Tabs leverage the user's preferred browser, which is optimized for performance and speed. This results in faster loading times compared to WebView.

- Security: Custom Tabs inherit the security features of the browser, including safe browsing, sandboxing, and other security measures. WebView, on the other hand, requires additional handling to ensure security.

- User Experience: Custom Tabs provide a seamless user experience by allowing users to stay within the app while browsing. They support features like saved passwords, autofill, and payment methods, which are not available in WebView.

- Maintenance: Using Custom Tabs reduces the maintenance overhead as they automatically update with the browser. WebView requires manual updates and handling of web content rendering.

By integrating Custom Tabs, we can enhance the app's performance, security, and user experience while reducing maintenance efforts.

Consider also [this article](https://developers.googleblog.com/en/upcoming-security-changes-to-googles-oauth-20-authorization-endpoint-in-embedded-webviews/)


johnvan7:
Thanks for the reply. 
Regarding the fact that WebView is updated in the same way as the browser, I found myself in this situation with my old tablet that I use for Home Assistant. 
It has an outdated version of Android WebView that, unfortunately, cannot be updated. However, it has Google Chrome installed, which is more up-to-date. When I log in, the WebView cannot display the page correctly because it is too complex, preventing access. By using CustomTabs instead, the page opens with the Chrome engine, allowing access. 
The redirect is intercepted by receiving an intent with the URL 'homeassistant://auth-callback'.

So, I understand your reasoning, and it is appropriate to proceed as you deem most appropriate.

jpelgrom:
> When I log in, the WebView cannot display the page correctly because it is too complex, preventing access. By using CustomTabs instead, the page opens with the Chrome engine, allowing access. 

How would switching to custom tabs for login help here, as you can now login but the main UI is still using WebView and won't be able to load? The main UI won't switch because of customizations/native integration?

johnvan7:
Main UI seems to work, I think login page is not working for background animations.
I didn't check if the whole app is using WebView, my bad.

webknjaz:
@johnvan7 hey, could you confirm this would fix #5360?

johnvan7:
Yes

TimoPtr:
I'm going to close this PR since it's quite outdated. We do have a plan to redo the onboarding but I don't think it makes sense to use the customTabs since like @jpelgrom said we won't be able to do so for the main screen of the app so it won't fix the main issue you are facing.

PR #4985, opened 2025-01-17
Update Gradle wrapper files

runningcode:
## Summary

Update Gradle wrapper files by running `./gradlew wrapper`.


jpelgrom:
Looks like your changing the line endings which is creating an unnecessarily large diff.

runningcode:
@jpelgrom thanks for pointing that out. For background information, the file is `.bat` which is a Windows/DOS batch file. The standard line endings for Windows/DOS are CRLF. This changes them to be CRLF.
I did this by simply running `./gradlew wrapper` which is a command that ensures the integrity of the wrapper scripts. I didn’t manually change the line endings. You will get the same diff by running `./gradlew wrapper`.
I can change them to reduce the diff if you like. Let me know how to proceed. :)

JBassett:
You likely have the wrong GIT settings for line endings: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

jpelgrom:
> This changes them to be CRLF.

Ah, I see now it is stored as LF in the repo (can't check that on GitHub) and with the file being Windows-specific it makes sense to change it. But please mention changes like that - most often people don't pay attention when this happens. Simply updating the file and committing doesn't give this result for everyone because of Git line endings settings.

gradle.properties
```diff
@@ -27,4 +27,4 @@ org.gradle.vfs.watch=true
 
 android.defaults.buildfeatures.resvalues=false
 android.defaults.buildfeatures.shaders=false
-android.nonFinalResIds=false
\ No newline at end of file
+android.nonFinalResIds=false
```
jpelgrom, line 30: Please revert this change as it is unrelated

runningcode:
I can add a `.gitattributes` file to the repo which will keep `.bat` files as CLRF. This is the standard practice in a lot of gradle repos. [Example](https://github.com/gradle/gradle/blob/master/.gitattributes#L9)

jpelgrom:
OK, that sounds good, but now you've changed gradlew.bat to LF again? (as evidenced by the changed files no longer including gradlew.bat)

runningcode:
I’m not sure why it isn’t showing in the PR but locally on my computer it is still CRLF. I didn’t make any explicit changes to it.
```
$ file gradlew.bat
gradlew.bat: DOS batch file text, ASCII text, with CRLF line terminators
```

runningcode:
Thanks for that. I see the same output as you for `git ls-files 'gradlew.bat' —eol` so it seems the index is still incorrect here.

I tried a couple things including deleting the file with `git rm` and also `git add —renormalize gradlew.bat` and nothing seems to change here. Do you have an idea of what I should do?

runningcode:
Good idea @jpelgrom . Done. I’ll create a separate PR to add the .gitattributes.

jpelgrom (approved):

PR #4986, opened 2025-01-17
Use Gradle Build Action

runningcode:
## Summary
This adds the Gradle build action to all Gradle workflows to speed up
builds using caching.
This replaces the Github action cache as well as the Gradle Wrapper
Validation action since the Gradle Build action [performs the same
validation](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#gradle-wrapper-validation)

As an example: ktlint on this PR took just 56 seconds (compared to 2:30 on other PRs) due to partial caching since the Gradle Build Action is already enabled for some workflows.


jpelgrom (approved):
Thanks! Good to be consistent and use the default caching instead of our own which is probably outdated.

> As an example: ktlint on this PR took just 56 seconds (compared to 2:30 on other PRs) due to partial caching since the Gradle Build Action is already enabled for some workflows.

Comparing numbers across PR runs is difficult because GitHub Actions performance can vary a lot.

PR #5000, opened 2025-01-21
Add gitattributes file

runningcode:
## Summary
Note: this does not run any normalization on the existing files, it
simply adds the gitattributes. I can add the existing file normalization
either in this PR or a future one.

This is based on this gitattributes file from the Gradle repo: https://github.com/gradle/gradle/blob/master/.gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,28 @@
+* text eol=lf
+
+#
+# The above will handle all files NOT found below
+# https://help.github.com/articles/dealing-with-line-endings/
+# https://github.com/Danimoth/gitattributes
+
+# These are explicitly windows files and should use crlf
+*.bat           text eol=crlf
+
+# Java/kotlin sources
+*.java          text diff=java
+*.kt            text diff=kotlin
+*.gradle        text diff=java
+*.gradle.kts    text diff=kotlin
+
+# These files are text and should be normalized
+*.properties    text
+*.xml           text
+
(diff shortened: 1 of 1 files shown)


jpelgrom:
> This is based on this gitattributes file from the Gradle repo: https://github.com/gradle/gradle/blob/master/.gitattributes

Any particular reason for using this one? It seems reasonable to me but would've expected a more Android-specific one.

>  I can add the existing file normalization either in this PR or a future one.

I don't see why we'd do it in a different PR. If you rebase first, the wrapper changes should be included, and renormalizing then should only change lines for files, no content changes.

.gitattributes
```diff
@@ -0,0 +1,28 @@
...
+#
+# The above will handle all files NOT found below
+# https://help.github.com/articles/dealing-with-line-endings/
+# https://github.com/Danimoth/gitattributes
+
+# These are explicitly windows files and should use crlf
+*.bat           text eol=crlf
+
+# Java/kotlin sources
+*.java          text diff=java
```
jpelgrom, line 12: We don't have Java files and won't add them, so this line isn't needed

TimoPtr:
@runningcode do you mind if I take over this PR?

runningcode:
Yes please.

TimoPtr:
Closing in favor of https://github.com/home-assistant/android/pull/5115

PR #5001, opened 2025-01-21
Run debug builds on master branch pushes

runningcode:
## Summary
This runs debug builds on master branch pushes in order to populate the
github actions cache via the Gradle build action for faster PR builds. Future no-op builds should be significantly faster!
By default the Gradle build action writes to the caches on `master` branch builds. Other branches read from the `master` branch cache or their own branch cache. [Here is more information on branch caching](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#select-which-branches-should-write-to-the-cache).
I'm assuming that since this is an OSS project the github action runners
are free so there is no cost to this.

I also renamed `pr_build` to `build` to reflect this change but that means that the required status check has been renamed.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -3,6 +3,8 @@ name: Pull Request
 on:
   pull_request:
     branches: [ master ]
+  push:
+    branches: [ master ]
 
 jobs:
   ktlint:
@@ -49,7 +51,7 @@ jobs:
       - name: Validate Lint
         run: ./gradlew lint
 
-  pr_build:
+  build:
     runs-on: ubuntu-latest
     permissions:
       contents: write


jpelgrom:
This action also generates debug apk files and uploads them as an artifact, which we don't need. The entire workflow is named pull request and also does lint checks which are unnecessary after merging.

Maybe some resources could be saved by just creating a new, simple job in the master push workflow?

> I'm assuming that since this is an OSS project the github action runners are free so there is no cost to this.

I believe so but that doesn't mean we should do more work that isn't used

runningcode:
I spent some time debugging the slow builds and lack of caching by running some tests in my fork. The lack of caching is due to the timestamped versionnames and version codes used in every module. Nothing can be retrieved from the cach because of that and the reckon gradle plugin unfortunately so it doesn’t make sense to make this change.

Issue #4955, opened 2025-01-05
[FR] Long press action for quick settings tiles


Jatoxo:
**Is your feature request related to a problem? Please describe.**
The quick settings tiles only support toggling entities. When you long press them, they just open the home assistant app on the default landing screen. 

**Describe the solution you'd like**
Long pressing the quick settings tiles should be a configurable action, ideally similar to the way actionable notifications can be configured. 

One example use case would be a tile for a light. You could press it to toggle the light, or long press to  open the light brightness and color controls.

**Describe alternatives you've considered, if any**
N/A

**Additional context**
N/A

dshokouhi:
The only use case I see here is to open the entity more info panel. No need for configuration on that. Keep it simple.

Jatoxo:
I was wrong in saying that long pressing a quick tile opens the home assistant app, it actually opens the home assistant app info screen



> The only use case I see here is to open the entity more info panel. No need for configuration on that. Keep it simple.

That would certainly be better than the current capability, but I can definitely think of use cases where you do not want want to open the entities "more info" screen

If it is just a boolean toggle, then the quick settings tile already covers all of its functionality, you could use the long press to instead open a screen with additional controls or settings pertaining to the same device / category

dshokouhi:
> I was wrong in saying that long pressing a quick tile opens the home assistant app, it actually opens the home assistant app info screen

that matches what Google says: https://developer.android.com/develop/ui/views/quicksettings-tiles#launch_an_activity

`Long-tapping a tile prompts the App Info screen for the user.`

However the intention is that overriding still takes user to a setting screen so the entity more info popup makes the most sense given that

https://developer.android.com/reference/android/service/quicksettings/TileService.html?utm_campaign=adp_series_quicksettingstiles_092916&utm_source=medium&utm_medium=blog#ACTION_QS_TILE_PREFERENCES

`An activity that provides a user interface for adjusting TileService preferences.` 

Long press historically always went to settings, i am not sure we should do anything outside of that personally speaking. The long press action should behave just like long press does in device controls.

Jatoxo:
The issue I have with opening the "more info" panel is that it can only be helpful if there is additional functionility on the same entity that you are toggling

If I look at the quick settings provided by Android, long pressing them brings up many other related settings. For example if you long press on the WiFi toggle, it opens the WiFi overview where you can manage multiple other related WiFi and network settings. Same with the bluetooth, Location or mobile data, the settings page they open includes many other related settings

[...]

PR #5008, opened 2025-01-24
Show more info on QS tile long press

jpelgrom:
## Summary

Shows the more info dialog when long pressing on a quick settings tile (if setup; otherwise redirects to settings or onboarding). This implements #4955 partially - but this is the minimum to start, and personally I'm not sure we should allow more configuration here.

## Screenshots

Short clip showing long press opening frontend, more info (Android doesn't really have a smooth transition here):

https://github.com/user-attachments/assets/5133cdc6-4988-41b8-a392-7ddfebe6a037

I don't think this is necessary right now

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -445,6 +445,15 @@
                 android:resource="@xml/provider_paths"/>
         </provider>
 
+        <activity
+            android:name=".qs.TilePreferenceActivity"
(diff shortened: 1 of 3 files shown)


dshokouhi:
> and personally I'm not sure we should allow more configuration here.

i agree here, makes the most sense and aligns with other features like device controls

dshokouhi (approved):

Issue #4547, opened 2024-08-01
Make the navigation bar and status bar transparent (edge-to-edge)


MathisP75:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered, if any**

**Additional context**

jpelgrom:
Unfortunately this isn't as easy as you might think because of the WebView. Last time I tried the problem was that information about insets wasn't set correctly on the WebView, making it impossible for the frontend to adapt. That means content will end up behind the system navigation, which is especially problematic when using button navigation or when you have a tall status bar due to a camera/cutout. It looks like the bug is still unresolved: https://issuetracker.google.com/issues/40699457.

(I thought we already had an issue for this, but can't find it.)

MathisP75:
Ah well that sucks, it would have been neat. Let's hope Google does something about it.

Gregman-js:
What do you think about passing this inset values from android activity to webview on page load?
I created a quick demo in Android Compose:
```kotlin
Scaffold (modifier = Modifier.fillMaxSize()) { padding ->
                AndroidView(
                    factory = { context ->
                        WebView(context).apply {
                            settings.javaScriptEnabled = true
                            webViewClient = object : WebViewClient() {
                                override fun onPageFinished(view: WebView?, url: String?) {
                                    super.onPageFinished(view, url)
                                    val safeInsetTop = (padding.calculateTopPadding().value)
                                    val safeInsetBottom = (padding.calculateBottomPadding().value)
                                    val safeAreaJs = """
                                    document.documentElement.style.setProperty('--android-safe-area-inset-top', '${safeInsetTop}px');
                                    document.documentElement.style.setProperty('--android-safe-area-inset-bottom', '${safeInsetBottom}px');
                                """.trimIndent()
                                    evaluateJavascript(safeAreaJs, null)
                                }
                            }
                        }
                    },
                    update = { webView ->
                        webView.loadUrl("http://10.0.2.2:8000")
                    },
                )
            }
```
In above snippet we can see that Scaffold component (edge-to-edge) give us `padding` property (our insets) and we are executing JS after page load that sets css properties.
Then HA page instead of relying on `env()` can have custom property that works on both android and ios.
```css
:root {
  --insets-top: var(--android-safe-area-inset-top, env(safe-area-inset-top));
  --insets-bottom: var(--android-safe-area-inset-bottom, env(safe-area-inset-bottom));
}
.main {
[...]

PR #5010, opened 2025-01-25
Display WebView under navigation bar

Gregman-js:
## Summary

Enable EdgeToEdge on WebViewActivity, show WebView under navigation bar.
```
serverManager.getServer()?.version?.isAtLeast(2024, 1)
```
Applying insets depend on server release with [this changes](https://github.com/home-assistant/frontend/pull/23811). We shouldn't display webview under navigation bar before frontend release because some ui elements will overlap with navigation bar.
After releasing this two PR we can even show webview under status bar on Android and iOS in future PR's.
We need to do something in this area because when HA will target api 35 some apis releated to changing status bar and navigation bar color will be not available [docs](https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge)
## Screenshots

Left iOS, Right: Android
image

## Any other notes

Dependency: https://github.com/home-assistant/frontend/pull/23811
Related issue: https://github.com/home-assistant/android/issues/4547

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -46,6 +46,7 @@ import android.widget.FrameLayout
 import android.widget.ImageButton
 import android.widget.Toast
 import androidx.activity.OnBackPressedCallback
+import androidx.activity.enableEdgeToEdge
 import androidx.activity.result.IntentSenderRequest
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.annotation.OptIn
@@ -55,8 +56,11 @@ import androidx.core.app.ActivityCompat
 import androidx.core.content.getSystemService
 import androidx.core.content.res.ResourcesCompat
 import androidx.core.graphics.ColorUtils
+import androidx.core.graphics.Insets
+import androidx.core.view.ViewCompat
 import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.WindowInsetsControllerCompat
+import androidx.core.view.updatePadding
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.lifecycleScope
 import androidx.lifecycle.repeatOnLifecycle
@@ -108,6 +112,7 @@ import io.homeassistant.companion.android.webview.externalbus.ExternalBusMessage
 import java.util.concurrent.Executors
 import javax.inject.Inject
 import javax.inject.Named
+import kotlin.math.roundToInt
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
@@ -239,6 +244,8 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
     private var downloadFileContentDisposition = ""
     private var downloadFileMimetype = ""
     private val javascriptInterface = "externalApp"
+    private var rootInsets: Insets = Insets.NONE
+    private var bottomInsetsApplied = false
 
     @SuppressLint("SetJavaScriptEnabled")
     override fun onCreate(savedInstanceState: Bundle?) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -249,10 +256,27 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
+            val bottomInset = if (bottomInsetsApplied) 0 else insets.bottom
+
+            v.updatePadding(
+                top = insets.top,
+                bottom = bottomInset
+            )
+
+            rootInsets = Insets.of(insets.left, 0, insets.right, insets.bottom - bottomInset)
+
+            WindowInsetsCompat.CONSUMED
```
MindFreeze, line 277: Could make sense to call the default `onApplyWindowInsets` here so the `env` is set properly in case the bug is ever fixed in Android

jpelgrom:
I don't want to do too much review while there is no frontend decision, but why is the status bar/top inset excluded?

Gregman-js:
@jpelgrom From my initial analysis, it turned out that the frontend is not yet adapted to the top insets. Even the iOS app does not draw webview under the status bar(If I haven't missed anything)

MindFreeze:
>I don't want to do too much review while there is no frontend decision

We have decided to accept the overall approach. Just have some nitpicks on the frontend PR

>From my initial analysis, it turned out that the frontend is not yet adapted to the top insets

I think the app should provide all the insets and let frontend decide what to do. As long as we are not breaking it, of course

app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -249,10 +256,27 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
             // Allow showing this on the lock screen when using device controls panel
             setShowWhenLocked(intent.extras?.getBoolean(EXTRA_SHOW_WHEN_LOCKED) ?: false)
         }
-
+        enableEdgeToEdge()
         super.onCreate(savedInstanceState)
 
         binding = ActivityWebviewBinding.inflate(layoutInflater)
+
+        bottomInsetsApplied = serverManager.getServer()?.version?.isAtLeast(2026, 1) == true
```
jpelgrom, line 264: The server/server version might change after the activity is created, most likely in multi-server scenarios. This will only check the current active server which might break UI if at some point it loads another, older frontend version. Not sure what the best fix is but that scenario should be handled.

Gregman-js: Ok, sorry for the delay. I will check what we can do with this.

Gregman-js:
I saw such an article recently https://developer.chrome.com/blog/edge-to-edge

Apparently chrome 135 will support EdgeToEdge and CSS env variables. I don't know yet how it will relate to webview

MindFreeze:
Frontend PR merged. It won't do anything until Android sets the vars but you can test with nightly now

TimoPtr:
@Gregman-js  what needs to be done here ?

Also if you could review this PR https://github.com/home-assistant/android/pull/5345 it would be of a great help :)

Gregman-js:
I will rebase and update this PR.
Do you have any idea how to resolve this comment? https://github.com/home-assistant/android/pull/5010#discussion_r1936117172
I am not sure how to approach this problem.

TimoPtr:
I'm checking your PR currently. I think there are something still missing on the frontend to properly handle the top bar and other screens. I'm going to open a branch with your changes on the latest version of main with my changes. I'll try to handle also the case of the server changing.
image

Gregman-js:
Yes, you are right. We can't render WebView under the status bar because frontend does not support it yet. My frontend PR only adds possibility for it. We have to add top padding (I think my PR does it) or some block above WebView

jpelgrom:
Closing in favor of #5346

PR #5041, opened 2025-02-10
Skip some more notification data items if they are blank

### LINKED ISSUE #5038: Blank subject key causes message key to not be displayed.

**Android version(s):**

Latest stable and beta companion app
2025.1.2-full
2025.2.1-full

**Device model(s):**

Google pixel 9 and 9 pro
Samsung s24 FE
Moto razr 24

**Home Assistant version:**

All recent ones. 2024+

Including a blank subject key causes some other keys, namely message, to not display on the device. 

We use a blueprint for frigate camera notifications which has lots of keys set to nothing by default. I.e `""`

Lots of users have suddenly reported the message key not working. Placing the same text in the title works fine. 

One user has tested a few things and this appears to be caused by the blank "subject" key.  Reports in SgtBatten/HA_blueprints#376

**Companion App Logs:**

```

```

Image

Here you see one notification with message and subject set (only the subject "test" is shown even when expanded).
And the lower notification had title and message both set, with subject blank. Only the title is showing. 

**Additional information:**

I've personally been seeing this for a month or so and couldn't figure it out. I am on the beta program of the app.

jpelgrom:
## Summary

Fixes #5038 

Previously FCM would omit blank keys. This was changed and is now in line with websocket notifications, but in some places the behavior might be unexpected. This PR ignores a couple more keys if blank (like we already do for many) where it seems safe to do so.

## Screenshots

n/a

n/a

## Any other notes

Testing note: if you try to send nothing from dev tools it will send a string `null`, so send a blank string `""` instead

### DIFF
--- a/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/java/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -1215,8 +1215,9 @@ class MessagingManager @Inject constructor(
         builder: NotificationCompat.Builder,
         data: Map<String, String>
     ) {
-        data[SUBJECT]?.let {
-            builder.setContentText(prepareText(it))
+        val subject = data[SUBJECT]
+        if (!subject.isNullOrBlank()) {
+            builder.setContentText(prepareText(subject))
         }
     }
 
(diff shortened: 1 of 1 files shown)


dshokouhi (approved):

SgtBatten:
Thanks so much. 

I'm not overly familiar with the update cycle of the app. Will this appear in the change notes of a future update?

jpelgrom:
No, bugfixes will generally not be called out.

We did mark another beta for release after merging this so it should be available within a day or two from Google Play (version 2025.2.3 of the app).

energywave:
Oh wonderful! I was modifying the Multinotify script to avoid these problems, great!
I think however that even the tag field need that treatment. You can verify by sending two notifications with tag None or "". The second one will overwrite the first, as you specified a valid value for tag.
I'm managing the thing, at the moment, by writing a random string in tag as there is no way to specify a "null" argument field (or I did not find it)

PR #5048, opened 2025-02-18
GitHub: Add dependency uploading

colin-foster-in-advantage:
## Summary

The gradle components don't automatically get detected upon build. Use the dependency-submission API to enable this dependency tracking.

Documentation explaining this:

https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/dependency-graph-supported-package-ecosystems#package-ecosystems-supported-via-dependency-submission-actions

That article links to this action:

https://github.com/marketplace/actions/build-with-gradle#the-dependency-submission-action

## Screenshots

The GitHub front end will change :smile: It currently only displays GitHub and Ruby dependencies, not Gradle.

image

N/A

## Any other notes

This is only for housekeeping / dependencies. I wasn't able to test the action easily due to app secrets, but it should work.

### DIFF
--- a/.github/workflows/onPush.yml
+++ b/.github/workflows/onPush.yml
@@ -118,6 +118,9 @@ jobs:
           AMAZON_APPSTORE_CLIENT_SECRET: ${{ secrets.AMAZON_APPSTORE_CLIENT_SECRET }}
           AMAZON_APPSTORE_APP_ID: ${{ secrets.AMAZON_APPSTORE_APP_ID }}
 
+      - name: Upload Github Dependencies
+        uses: gradle/actions/dependency-submission@v4
+
       - name: Print `git status`
         run: git status
 


jpelgrom:
Thanks for the suggestion, it looks good to me. Could you explain "This is only for housekeeping" a bit more?

colin-foster-in-advantage:
> Thanks for the suggestion, it looks good to me. Could you explain "This is only for housekeeping" a bit more?

Yep. It'll do two things (as far as I understand).

1. It'll populate the dependencies correctly in the screenshot I included. That'll make "Export SBOM" correctly include Gradle dependencies in the SPDX export. It currently only references Ruby Gems and github actions.
2. It'll (likely?) publish security advisories in the "Security" tab. This one might differ based on how the repository is configured.

jpelgrom (approved):

dshokouhi (approved):

colin-foster-in-advantage:
Thanks. Looks like it worked!

image

PR #5078, opened 2025-03-06
Introduce convention gradle plugin for apps

TimoPtr:
## Summary

To avoid duplication of configuration in our build system, I have introduced [convention plugins](https://docs.gradle.org/current/samples/sample_convention_plugins.html). To make it easier to work with, I've created a Gradle project named build-logic and imported it into the existing project using the includeBuild feature. This approach is highly inspired by the [Now in Android repository](https://github.com/android/nowinandroid/tree/e308246131a44cd9a4caf700bee26e8351d06090/build-logic).

## Any other notes

I took the liberty of making some changes, and I will go over them in the comments of this PR.


app/build.gradle.kts
```diff
@@ -1,196 +1,36 @@
...
-    alias(libs.plugins.compose.compiler)
+    alias(libs.plugins.homeassistant.android.dependencies)
 }
 
 android {
-    namespace = "io.homeassistant.companion.android"
-
-    compileSdk = libs.versions.androidSdk.compile.get().toInt()
-
-    ndkVersion = "21.3.6528147"
```
TimoPtr, line 19: I don't see anything in the codebase that needs the NDK and decided to remove this to avoid future issue.

app/build.gradle.kts
```diff
@@ -1,196 +1,36 @@
...
-
         manifestPlaceholders["sentryRelease"] = "$applicationId@$versionName"
         manifestPlaceholders["sentryDsn"] = System.getenv("SENTRY_DSN") ?: ""
 
         bundle {
             language {
+                // We want to keep the translations in the final AAB for all the language
                 enableSplit = false
             }
         }
```
TimoPtr, line 22: I did not move this into the convention plugin based on the fact that it is not set for the `:wear` module. I'm not 100% but is it possible to application bundle for wear apps 🤔?

app/build.gradle.kts
```diff
@@ -1,196 +1,36 @@
...
-
-    testOptions {
-        unitTests.isReturnDefaultValues = true
-    }
-
-    tasks.withType<Test>().configureEach {
-        useJUnitPlatform {
-            includeEngines("spek2")
-        }
-    }
```
TimoPtr, line 111: I took the liberty of removing this since we don't have any tests anymore.

app/build.gradle.kts
```diff
@@ -1,196 +1,36 @@
...
-
-    implementation(libs.bundles.paging)
-
-    implementation(libs.reorderable)
-    implementation(libs.changeLog)
-
-    implementation(libs.zxing)
-    implementation(libs.improv)
-
-    implementation(libs.car.core)
```
TimoPtr, line 193: IMO it was a source of error to have a duplicated list of dependencies. Someone could have missed to add a deps into `:automotive` while adding it to `:app`. 

My 2ct on this is that we might want to consider another approach for this and avoid using the source of `:app` in another module which btw raise a warning about duplicates in Android Studio.

automotive/build.gradle.kts
```diff
@@ -1,33 +1,21 @@
...
 
     defaultConfig {
-        applicationId = "io.homeassistant.companion.android"
+        // Override minSDK since automotive bellow 29 does not exist
         minSdk = libs.versions.androidSdk.automotive.min.get().toInt()
-        targetSdk = libs.versions.androidSdk.target.get().toInt()
-
-        versionName = project.version.toString()
-        // We add 2 because the app, wear (+1) and automotive versions need to have different version codes.
-        versionCode = (System.getenv("VERSION_CODE")?.toIntOrNull() ?: 1) + 3
```
TimoPtr, line 29: It was saying 2 put actually it add 3. I didn't want to break anything so I kept 3 and updated the doc.

build-logic/convention/build.gradle.kts
```diff
@@ -0,0 +1,45 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+
+plugins {
+    `kotlin-dsl`
+    alias(libs.plugins.ktlint)
+}
+
+group = "io.homeassistant.companion.android.buildlogic"
+
+// Configure the build-logic plugins to target JDK 17 and is not related to what is running on device.
```
TimoPtr, line 10: Could be 21 but I kept the same logic as nowinandroid.

build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
```diff
@@ -0,0 +1,122 @@
...
+ * means that it should not in this convention plugin anymore.
+ *
+ * application `version code` can be set in the environment variable VERSION_CODE
+ * keystore information can be set in the environment variables:
+ * - KEYSTORE_PATH for the path to the keystore
+ * - KEYSTORE_PASSWORD for the store password
+ * - KEYSTORE_ALIAS for the alias
+ * - KEYSTORE_ALIAS_PASSWORD for the alias password
+ */
+class AndroidApplicationConventionPlugin : Plugin<Project> {
```
TimoPtr, line 29: I didn't give a lot of though to this class for now it's mostly a copy paste. But in the future we might want to split this into smaller plugins more specifics. Since pieces of this can be shared with `:common`.

build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
```diff
@@ -0,0 +1,122 @@
...
+                flavorDimensions.add("version")
+                productFlavors {
+                    create("minimal") {
+                        applicationIdSuffix = ".minimal"
+                        versionNameSuffix = "-minimal"
+                    }
+                    create("full") {
+                        applicationIdSuffix = ""
+                        versionNameSuffix = "-full"
+                    }
```
TimoPtr, line 101: Is it an issue for the watch to have these flavors?

settings.gradle.kts
```diff
@@ -25,7 +32,16 @@ reckon {
...
 
 dependencyResolutionManagement {
+    repositoriesMode = RepositoriesMode.FAIL_ON_PROJECT_REPOS
     repositories {
+        google {
+            content {
+                includeGroupByRegex("com\\.android.*")
+                includeGroupByRegex("com\\.google.*")
+                includeGroupByRegex("androidx.*")
+                includeGroupByRegex("org\\.chromium.*")
```
TimoPtr, line 42: I took the liberty to make sure we download the libraries from the place we are expecting them to be.

settings.gradle.kts
```diff
@@ -25,7 +32,16 @@ reckon {
 }
 
 dependencyResolutionManagement {
+    repositoriesMode = RepositoriesMode.FAIL_ON_PROJECT_REPOS
```
TimoPtr, line 35: This force us to use the version catalog feature otherwise it doesn't build.

app/build.gradle.kts
```diff
@@ -1,196 +1,36 @@
...
 }
 
 android {
-    namespace = "io.homeassistant.companion.android"
-
-    compileSdk = libs.versions.androidSdk.compile.get().toInt()
-
-    ndkVersion = "21.3.6528147"
-
     useLibrary("android.car")
```
TimoPtr, line 11: Does it make sense to have this also for the `:app`?

settings.gradle.kts
```diff
@@ -25,7 +32,16 @@ reckon {
...
+            content {
+                includeGroupByRegex("com\\.android.*")
+                includeGroupByRegex("com\\.google.*")
+                includeGroupByRegex("androidx.*")
+                includeGroupByRegex("org\\.chromium.*")
+            }
+        }
         mavenCentral()
         google()
         maven("https://jitpack.io")
```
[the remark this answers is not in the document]
jpelgrom: Yes, it's used for the changelog library (and let's not tackle replacing that in this PR)

build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
jpelgrom: You're wasting some resources building it, and there is no such this as a GMS-less Wear OS so no reason for a (minimal) flavor to exist.

app/build.gradle.kts
jpelgrom: Not saying you're wrong (none of us are Gradle experts), and it always felt a little clunky to me, but this was originally contributed by someone from Google's Automotive team.

Missing a dependency is a risk but there is a review / it probably would break when building.

app/build.gradle.kts
jpelgrom: What do you mean with "application bundle for wear apps"?

Everything submitted to Google Play is an aab. I don't believe you can change the language for apps on Wear OS independently of the OS which is why there is no reason to disable that split.

TimoPtr: I was not sure the app for Wear was an AAB or an APK. TBH last time I publish an app for Wear OS AAB was not a thing yet (or I was not using it).

app/build.gradle.kts
TimoPtr: Depending on the deps that is missing, it could manage to build and fail at runtime. I'm thinking about something that would use reflection (even if I think we are not using it here).

If we were to go the path of removing this source sharing, we would need to create another layer with a gradle module and extract some part of `:app` into it. It would allow to have specific code for `automotive` isolated from the rest. WDYT? Could be useful to if we want to have a dedicated experience for AndroidTV.

build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
TimoPtr: If we go further with this PR, I'll extract this into a dedicated convention plugin that is not applied to Wear module.

app/build.gradle.kts
jpelgrom: (If you take a look at the GitHub Actions run for the last beta you can see the files generated and pushed.)

settings.gradle.kts
TimoPtr: Well in build-logic we don't use chromium, but it is also true for `androidx` and I kept it 🤔. Should we match or should we try to extract this into a File that is loaded on both side?

settings.gradle.kts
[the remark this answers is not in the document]
TimoPtr: Agreed I don't want this PR to grow at this point. But if it's used I'm fined with keeping it.

 We could maybe can limit it to this specific lib with an `include` rule WDYT?

settings.gradle.kts
jpelgrom: I'm not sure which option is the best, but don't like inconsistency. If there's a good reason to have them be different that's fine but maybe add a one line comment about it? Honestly not quite sure how this interacts with the one in build-logic/settings.gradle.kts.

settings.gradle.kts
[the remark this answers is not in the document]
jpelgrom: It's already listed last so should be only applied to that dependency, all others are available on other repositories. No need to overcomplicate things.

build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
TimoPtr: @jpelgrom I did extract this into another convention plugin to make it cleaner.

settings.gradle.kts
TimoPtr: I've played with how I can share this and I'm not a big fan of the solution, I had to create another project and add another includeBuild. 
image

It adds a third layer in the equation and a new settings.gradle.kts (that is only referencing `mavenCentral`). Usage of ktlint from `libs` is not possible or I need to duplicate the line `id("dev.panuszewski.typesafe-conventions") version "0.4.1"` somewhere.

Lemme know what do you prefer. From my point of view after some though I don't mind having a different `repositories` settings in the two project. It doesn't leak from one to the other. For instance jitpack is not available in `build-logic`.

jpelgrom: ... another project seems like a lot of complication for aligning one thing that will rarely if ever be touched.

Let's keep it as suggested in the PR for now 😅

app/build.gradle.kts
jpelgrom: Currently I think yes because of it's use in `BaseActivity` and `BaseVehicleScreen`? Otherwise the import breaks (unresolved references).

build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
jpelgrom: + some pieces can be shared with `wear`

app/build.gradle.kts
jpelgrom: Following that logic, I think you can also remove [`AndroidApplicationConventionPlugin` line 95-97](https://github.com/home-assistant/android/pull/5078/files#diff-6f219b940c62174bcc616df1cf6d1e1342046ae168ca268f5d858920d87465eaR95-R97).

wear/build.gradle.kts
```diff
@@ -1,67 +1,17 @@
...
-    alias(libs.plugins.compose.compiler)
 }
 
 android {
-    namespace = "io.homeassistant.companion.android"
-
-    compileSdk = libs.versions.androidSdk.compile.get().toInt()
-
     defaultConfig {
         applicationId = "io.homeassistant.companion.android"
```
jpelgrom, line 8: Is this still needed? I think it is now defined in the application convention plugin.

jpelgrom:
In addition to the specific comments: I generally understand the goal and code here, and think most of it looks good and is an improvement we could adopt. (It builds and avoids duplication. Maybe a bit verbose in some places.)

build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
```diff
@@ -0,0 +1,106 @@
...
+ * - `KEYSTORE_ALIAS_PASSWORD`: The password for the key alias.
+ */
+class AndroidApplicationConventionPlugin : Plugin<Project> {
+    override fun apply(target: Project) {
+        with(target) {
+            apply(plugin = libs.plugins.android.application.getPluginId())
+            apply(plugin = libs.plugins.kotlin.android.getPluginId())
+            apply(plugin = libs.plugins.ksp.getPluginId())
+            apply(plugin = libs.plugins.hilt.getPluginId())
+            apply(plugin = libs.plugins.kotlin.parcelize.getPluginId())
```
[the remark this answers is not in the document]
TimoPtr: Do you think it's an issue? TBH I saw it and was lazy about creating something specific for the wear. I'm guess it impacts the build time but I think it could be négligeable WDYT?

app/build.gradle.kts
TimoPtr: I'm going to keep this in my notes for later :)

app/build.gradle.kts
TimoPtr: I do plan to bring back test pretty soon that's why I kept it.

wear/build.gradle.kts
TimoPtr: You're going to remove it

build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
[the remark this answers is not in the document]
jpelgrom: Could be a slippery slope, one additional things that's not strictly needed now, and then later on another one, and another...

Build time impact probably depends on the system. The 'cost' of doing this only for app/automative is 1 additional line.

build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
jpelgrom: My comment was wrong about wear and I deleted it before your reply but apparently not fast enough :)

build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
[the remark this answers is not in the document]
TimoPtr: Done

jpelgrom (approved):
This generally looks good to me, and if not we can fix it later.

PR #5083, opened 2025-03-11
Assist continue conversation feature while using microphone

TimoPtr:
## Summary

This pull request introduces support for a new flag in the Assist API that indicates when the user should be prompted for input. This flag is currently available when using a Large Language Model (LLM) backend.

To integrate this new functionality with the existing Assist conversation logic, I've ensured that recording only begins after the Text-to-Speech (TTS) output has finished. This prevents the user interface from entering an unrecoverable state, such as the disappearance of buttons.

These changes have been implemented in both the mobile app and the Wear OS app.

To maintain backward compatibility, I've added basic unit tests. Additionally, I've centralized the creation of the Jackson `ObjectMapper` to ensure consistency.

## How to test
To test this feature, you must use the latest nightly build of the Home Assistant, as the flag is not exposed in stable releases. The flag's behavior may vary depending on the LLM used. I found that using [Gemini](https://www.home-assistant.io/integrations/google_generative_ai_conversation/) in English and asking "Set a timer" was an effective way to trigger the flag.

## Any other notes

This pull request is part of a larger effort to enhance the Assist feature. For reference, the corresponding changes in the frontend are available in this [PR](https://github.com/home-assistant/frontend/pull/24511).


common/src/main/java/io/homeassistant/companion/android/common/assist/AssistViewModelBase.kt
```diff
@@ -68,12 +70,12 @@ abstract class AssistViewModelBase(
...
      * @param onMessage callback for messages that should be posted for this pipeline run, with 3
      * arguments: the message, whether the message is input/output/undetermined, whether the message
-     * is an error message
+     * is an error message, whether the conversation should continue
      */
     protected fun runAssistPipelineInternal(
         text: String?,
         pipeline: AssistPipelineResponse?,
-        onMessage: (String, Boolean?, Boolean) -> Unit
+        onMessage: (String, Boolean?, Boolean, Boolean) -> Unit
```
TimoPtr, line 78: I'm not a fan of this callback:
- When using it it's not clear what are the parameters
- The parameters are not always relevant depending on the event we received
- Nullable boolean is kinda weird (3 states instead of 2 it makes the rest of the logic complicated to understand)

We could explore
- Use Flow instead of a callback
- Encapsulate the events into a `sealed class` to have specifics events with proper attributes depending on the context

common/src/main/java/io/homeassistant/companion/android/common/assist/AssistViewModelBase.kt
```diff
@@ -114,21 +116,26 @@ abstract class AssistViewModelBase(
                     AssistPipelineEventType.STT_END -> {
                         stopRecording()
                         (it.data as? AssistPipelineSttEnd)?.sttOutput?.let { response ->
-                            onMessage(response["text"] as String, true, false)
+                            onMessage(response["text"] as String, true, false, false)
```
TimoPtr, line 119: I would like to push for a code style guideline in the codebase.
We should use named parameters when using booleans. Why is that an issue today
- The PR review is harder since you don't know at first sight what is each boolean in the function call. You need to find the function definition to know what it is.
- When using default params you might set the wrong boolean like `error` instead of `continueConversation` (I made the mistakes on this PR 🤦 and lost a lot of time debugging why my json is not properly parsed).

We could enforce that rule with ktlint.

common/src/main/java/io/homeassistant/companion/android/common/assist/AssistViewModelBase.kt
```diff
@@ -51,6 +52,7 @@ abstract class AssistViewModelBase(
 
     private var binaryHandlerId: Int? = null
     private var conversationId: String? = null
+    private var continueConversation = AtomicBoolean(false)
```
TimoPtr, line 55: I decided to use an AtomicBoolean here, because we are not 100% sure about the threading that is coming from the flow (it can change over time while developing) and the value is modified in two different places. To avoid potential issues I preferred to be safe.

common/src/main/java/io/homeassistant/companion/android/common/util/AudioUrlPlayer.kt
```diff
@@ -33,13 +34,20 @@ class AudioUrlPlayer(private val audioManager: AudioManager?) {
...
+    suspend fun playAudio(url: String, isAssistant: Boolean = true, donePlaying: () -> Unit): Boolean = withContext(Dispatchers.IO) {
         if (player != null) {
             stop()
         }
 
+        val refDonePlaying = AtomicReference<(() -> Unit)?>(donePlaying)
+
+        fun donePlayingInvocation() {
+            refDonePlaying.getAndSet(null)?.invoke()
+        }
```
TimoPtr, line 49: I used this complex mechanism to ensure that the callback is only call once.

common/src/main/java/io/homeassistant/companion/android/common/assist/AssistViewModelBase.kt
```diff
@@ -68,12 +70,12 @@ abstract class AssistViewModelBase(
      * @param pipeline information about the pipeline, or `null` to use the server's default
      * @param onMessage callback for messages that should be posted for this pipeline run, with 3
```
jpelgrom, line 71:
```suggestion
     * @param onMessage callback for messages that should be posted for this pipeline run, with 4
```

common/src/main/java/io/homeassistant/companion/android/common/assist/AssistViewModelBase.kt
```diff
@@ -114,21 +116,26 @@ abstract class AssistViewModelBase(
...
+                            onMessage(response, false, false, false)
                         }
                     }
                     AssistPipelineEventType.TTS_END -> {
                         if (!isVoice) return@collect
                         val audioPath = (it.data as? AssistPipelineTtsEnd)?.ttsOutput?.url
                         if (!audioPath.isNullOrBlank()) {
-                            playAudio(audioPath)
+                            playAudio(audioPath) {
+                                // We send the continueConversation flag here after getting it from INTENT_END so that
```
jpelgrom, line 135: Does `INTENT_END` refer to a pipeline stage here? Let's use the app's constants then to avoid confusion.

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/entities/ConversationResponse.kt
```diff
@@ -1,9 +1,10 @@
-package io.homeassistant.companion.android.common.data.websocket.impl.entities
-
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties
-
-@JsonIgnoreProperties(ignoreUnknown = true)
-data class ConversationResponse(
-    val response: ConversationSpeechResponse,
-    val conversationId: String?
-)
+package io.homeassistant.companion.android.common.data.websocket.impl.entities
```
jpelgrom, line 1: I think you're changing the line endings here, causing the entire file to show up as changed, instead of just the 2 changed lines. Can you undo that?

(There is another PR pending to set attributes to make this easier, as you've already discovered.)

common/src/main/java/io/homeassistant/companion/android/common/util/AudioUrlPlayer.kt
```diff
@@ -33,13 +34,20 @@ class AudioUrlPlayer(private val audioManager: AudioManager?) {
...
+     * @param donePlaying callback to be invoked when playback has finished (it covers when it's not playing anything or an error happened)
      * @return `true` if the audio playback started, or `false` if not
      */
-    suspend fun playAudio(url: String, isAssistant: Boolean = true): Boolean = withContext(Dispatchers.IO) {
+    suspend fun playAudio(url: String, isAssistant: Boolean = true, donePlaying: () -> Unit): Boolean = withContext(Dispatchers.IO) {
         if (player != null) {
             stop()
         }
 
+        val refDonePlaying = AtomicReference<(() -> Unit)?>(donePlaying)
```
jpelgrom, line 45: These types don't match - `donePlaying` can be null here but not in the function arguments.

jpelgrom (commented):
I feel like some of your comments here are improvements that are not strictly related to the feature but instead 'found' while working on it. If you see room for improvement, great - but do it in a separate PR to keep this one focused and make it quick/easy to review the improvement.

common/src/main/java/io/homeassistant/companion/android/common/util/AudioUrlPlayer.kt
TimoPtr: Yes that's on purpose. I nullify the reference once the player is done with it. I do that especially because of the error case. It is just for safety if the API in the future changes I don't want this callback to be ever called twice for any reason.

I can add a comment if it's not clear enough.

jpelgrom: Would it make sense to also allow `null` as an argument then? No reason to pass a function that does nothing.

TimoPtr: Indeed I'll make it possible to set this as null. I didn't because it is used in only one place at the moment, but you are right if it used in another place it might not be needed.

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/entities/ConversationResponse.kt
TimoPtr: Indeed https://github.com/home-assistant/android/pull/5115/files#diff-1aa2d56c19899c60887fda68b38768529fd8ec55e0451a817dc3bbe7d49b3530

common/src/main/java/io/homeassistant/companion/android/common/util/AudioUrlPlayer.kt
```diff
@@ -70,10 +78,12 @@ class AudioUrlPlayer(private val audioManager: AudioManager?) {
                     Log.e(TAG, "Media player encountered error: $what ($extra)")
                     releasePlayer()
                     cont.resume(false)
+                    donePlayingInvocation()
```
jpelgrom, line 81: The frontend implementation only uses continue conversation if playback was successful - if the `error` event occurs, `ended` is not triggered. Is this difference intentional?

TimoPtr: Good point! I didn't catch that at first.

Regarding the `MediaPlayer` error handling:
```
mp – the MediaPlayer the error pertains to what – the type of error that has occurred:
MEDIA_ERROR_UNKNOWN
MEDIA_ERROR_SERVER_DIED
 extra – an extra code, specific to the error. Typically implementation dependent.
MEDIA_ERROR_IO
MEDIA_ERROR_MALFORMED
MEDIA_ERROR_UNSUPPORTED
MEDIA_ERROR_TIMED_OUT
MEDIA_ERROR_SYSTEM (-2147483648) - low-level system error.
```

These errors can occur for multiple reasons, including system issues. I've also covered the case where the `MediaPlayer` reports an error while calling `prepare()`.

While writing this code, I considered the case where, if an error occurs, the user could still continue the conversation using voice input. I often use the assistant without TTS on my phone, which is probably why I made that choice. I don't expect the feedback to be required when I speak to it.

@balloob, what are your thoughts on this?

balloob: I would stop. Looks like the OS is messed up when it happens.

balloob: But either way is fine right now. Let’s merge and figure out.

common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/entities/ConversationResponse.kt
```diff
@@ -1,9 +1,10 @@
...
-)
+package io.homeassistant.companion.android.common.data.websocket.impl.entities
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class ConversationResponse(
+    val response: ConversationSpeechResponse,
+    val conversationId: String?,
+    val continueConversation: Boolean = false
```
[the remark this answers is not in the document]
TimoPtr: I agree but I feel always reluctant to treat a Boolean as nullable. It means that it has 3 states. Here we know that if it is not here it should be false in terms of business logic. wdyt?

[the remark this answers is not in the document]
jpelgrom: I guess this is down to personal preference - in my opinion these entities/models should do as little transformation as possible. `continueConversation == true` as a check looks fine to me, and makes it more obvious that it may not be returned by the server. However if you think a default value is better that's also fine, you just asked for an opinion ;)

jpelgrom:
In case you hadn't noticed, this branch needs a rebase/has merge conflicts.

All of my open comments are addressed so good to merge once that is done for me.

PR #5091, opened 2025-03-11
Improve renovate configuration and introduce gradle lockfile

TimoPtr:
## Summary

@frenck suggested to use [Renovate](https://docs.renovatebot.com/) instead of dependabot, to manage the update of our dependencies. It has been added to the repository in a previous PR, but with the basic configuration. This PR add futher configuration to it.
This PR also include the usage of [gradle lockfile](https://docs.gradle.org/current/userguide/dependency_locking.html), which makes https://github.com/home-assistant/android/pull/5079 not needed anymore.
In order to create the lockfile for all the modules I've created a small Gradle task that run `dependencies` on all the subprojects.

## Any other notes

Compare to https://docs.gradle.org/current/userguide/dependency_locking.html this PR doesn't add any check on CI to verify that something got up to date. Simply because gradle use the `lockfile` instead of the versions inside the version catalog.
It comes with the drawback that locally we need to update the lockfile to update a dependency.


jpelgrom:
If we're switching, shouldn't this turn off Dependabot too to avoid duplicates?

TimoPtr:
Yes that is the plan. I'll disable it once the setup of renovate is correct. I'll remove the duplicate manually.

PR #5099, opened 2025-03-12
Extract setup build steps into a dedicated Github Action and export lint reports

TimoPtr:
## Summary

To avoid duplication and have a single source of truth, I've extracted common Github Action steps into a dedicated action. I took the liberty to align the naming of the file to `.yml` since it was the most used. I also ran a yml code style tool on it.
I took the opportunity to upload the lint and ktlint report into Github, it will appear in the PR if something is wrong instead of looking into the logs.


.github/actions/setup-build-env/action.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+        - name: Mock google-services.json
+          shell: bash
+          if: ${{ inputs.mock-google-services == 'true' }}
+          run: |
+              cp .github/mock-google-services.json app/google-services.json
+              cp .github/mock-google-services.json wear/google-services.json
+              cp .github/mock-google-services.json automotive/google-services.json
+
+        - name: Set up JDK 17
```
jpelgrom, line 19: Not sure about this, but maybe https://github.com/home-assistant/android/pull/5108#discussion_r1993341746 is better changed here? The other PR already has a lot going on and needs a rebase anyway if this one is merged.

.github/workflows/pr.yml
```diff
@@ -31,76 +24,77 @@ jobs:
...
+      -   name: Upload lint reports (SARIF)
+          if: ${{ !cancelled() && hashFiles('**/*.sarif') != '' }}
+          uses: github/codeql-action/upload-sarif@v3
+          with:
+              sarif_file: './'
+
   lint:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
```
jpelgrom, line 36: Why is this needed? I do not see it mentioned on https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github.

.github/actions/setup-build-env/action.yml
TimoPtr: I don't mind the rebase but if you prefer I can bring the change to this PR.

TimoPtr: In fact I would prefer to keep it in the other PR since it comes with some explanation in the description of the PR that explain why I did the update.

.github/workflows/pr.yml
TimoPtr: Good point it's a mistake.

jpelgrom (approved):

.github/workflows/pr.yml
TimoPtr: I do remember why I added this permission. If you look in the PR comments Github post a comment. I initially though it needed this permission to push a comment on the PR. I did remove the permission and I'm gonna check in another PR if it works properly without the permission.

PR #5107, opened 2025-03-12
Add new android common gradle convention plugin and update common module

TimoPtr:
## Summary

This is a follow up of https://github.com/home-assistant/android/pull/5078 to avoid duplication in common gradle module.
The Application convention module does call this new module directly and `:common` call it too.

### DIFF
--- a/build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
@@ -4,9 +4,6 @@ import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.kotlin.dsl.apply
 import org.gradle.kotlin.dsl.configure
-import org.gradle.kotlin.dsl.withType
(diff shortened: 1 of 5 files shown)


build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
```diff
@@ -15,6 +12,7 @@ private const val APPLICATION_ID = "io.homeassistant.companion.android"
  * This centralizes configuration, preventing duplication across multiple modules.
  *
  * This plugin applies several Gradle plugins that are commonly used in all application modules.
+ * More specifically it apply the [AndroidCommonConventionPlugin].
```
jpelgrom, line 15: Grammar + indicate that specifically only refers to build-logic and not the other plugins.

```suggestion
 * This plugin applies several Gradle plugins that are commonly used in all application modules,
 * including the [AndroidCommonConventionPlugin].
```

jpelgrom (approved):

Issue #3518, opened 2023-05-06
Jackson 2.14+ requires higher minimum SDK level 


jpelgrom:
**Is your feature request related to a problem? Please describe.**

Jackson, which is used for JSON (de)serialization, requires a minimum SDK level of [26 for version 2.14+](https://github.com/FasterXML/jackson/wiki/Jackson-Releases#:~:text=Last%20branch%20for%20Android%20SDK%2021%2B%20(SDK%2026%20minimum%20for%202.14)). The app's minimum SDK level is currently 21.

Jackson's 2.13 branch will remain supported until at least the end of 2023, but the app should make sure it's using a supported library for critical functionality like this.

**Describe the solution you'd like**

- Increase the app's minimum SDK level (which isn't a bad idea anyway considering how long it's been at 21, though 21->26 is quite a big increase)

and/or

- Switch to another library for JSON (de)serialization like Gson or Moshi (check minimum SDK level and plans for change first)

**Describe alternatives you've considered, if any**

**Additional context**

#3077, #3489

jpelgrom:
Looks like Retrofit 2.10 adds support for kotlinx.serialization - would this be a good candidate for replacement, as it supports whatever Kotlin supports and we're already using Kotlin anyway? I kind of like the idea of less orgs who can make dependencies difficult for us.

JBassett:
Sounds like a good plan

PR #5108, opened 2025-03-12
Introduce instrumentation test

TimoPtr:
## Summary

In this PR I introduce the instrumentation test. To do so I've created a small test that is going to catch that we should not update `jackson` to a version above 2.14 (check https://github.com/home-assistant/android/issues/3518 for more details).
The Github Action job is using https://github.com/marketplace/actions/android-emulator-runner and use the basic configuration that they propose.
In the future we should add more test and platform like automotive and wear.
The current job is using a matrix to run the tests on API 21 and 35. In a follow up PR I'm going to introduce the different platform wear and automotive.
A new job as been created to report the result in the PR itself with annotation/checks/comment. I've made some failure on purpose to showcase how it will be displayed within the PR.

## Any other notes

I've tested https://github.com/mikepenz/action-junit-report?tab=readme-ov-file and https://github.com/EnricoMi/publish-unit-test-result-action. The later have a better documentation and better supports.


TimoPtr:
I'll rebase this PR once we merge https://github.com/home-assistant/android/pull/5099

.github/actions/setup-build-env/action.yml
```diff
@@ -1,26 +1,32 @@
...
-        - name: Set up JDK 17
-          uses: actions/setup-java@v4.7.0
-          with:
-              distribution: "temurin"
-              java-version: "17"
+    - name: Set up JDK 21
+      uses: actions/setup-java@v4.7.0
+      with:
+        distribution: "temurin"
+        java-version: "21"
```
TimoPtr, line 27: ⚠️ ⚠️ ⚠️⚠️ ⚠️ ⚠️⚠️ ⚠️ ⚠️⚠️ ⚠️ ⚠️

I did update the JDK while trying to debug an issue I had. In the end the issue was a file that I forgot to commit ... But today the latest versions of Android Studio comes with the JDK 21 and I've been using it for the last 2 weeks on the project.

My issue led me to investigate why the result was different on CI and I end up here. Everything is working as it should on CI at the moment with 21 so if there are no objections I will keep it like this.

.github/actions/setup-build-env/action.yml
```diff
@@ -1,26 +1,32 @@
...
-        default: "false"
-        required: false
+  mock-google-services:
+    description: "Mock the google-services.json file inside the gradle modules."
+    default: "false"
+    required: false
+  cache-disabled:
+    description: When 'true', all caching is disabled. No entries will be written to or read from the cache.
+    required: false
+    default: "false"
```
TimoPtr, line 11: In order to be able to debug cache issues I introduced this flag, but its usage should not land on `master`.

.github/workflows/pr.yml
```diff
@@ -98,3 +98,93 @@ jobs:
...
+      - uses: ./.github/actions/setup-build-env
+        with:
+          mock-google-services: "true"
+
+      - name: Enable KVM group perms
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+          ls /dev/kvm
```
TimoPtr, line 130: not needed but useful for future debugging.

.github/workflows/pr.yml
```diff
@@ -98,3 +98,93 @@ jobs:
...
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: Instrumentation test results (API${{ matrix.api-level }})
+          path: |
+            **/build/reports/*
+            **/build/outputs/*/connected/*
+
+  publish_test_results:
```
TimoPtr, line 172: We use a dedicated job here since we have a matrix for instrumentationTest and it would lead to strange issues inside the feedback on the PR.

I'll also use this job for the unit tests so they are integrated in the PR too.

common/src/androidTest/kotlin/io/homeassistant/companion/android/JacksonDummyTest.kt
```diff
@@ -0,0 +1,36 @@
+package io.homeassistant.companion.android
+
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.fasterxml.jackson.module.kotlin.readValue
+import org.junit.Assert.assertEquals
+import org.junit.Test
```
TimoPtr, line 6: FYI we can't use junit5 on instrumentationTest since junit5 needs API 26 minimum due to the JAVA_VERSION it uses.

.github/workflows/pr.yml
```diff
@@ -98,3 +98,93 @@ jobs:
...
           path: ./**/*.apk
+
+  instrumentation_test:
+    name: "Instrumentation Tests"
+    runs-on: ubuntu-latest
+    strategy:
+      # We want the result of each device to be reported, so we can't fail-fast
+      fail-fast: false
+      matrix:
+        api-level: [21, 35]
```
TimoPtr, line 109: I think we should at least test the two extreme versions at the moment. In the future if we see that we need more we can add a new one.

jpelgrom:
Not sure about the order you want to merge, but there's a merge conflict and the comments from #5099 need to be addressed here as well.

TimoPtr:
> Not sure about the order you want to merge, but there's a merge conflict and the comments from #5099 need to be addressed here as well.

I'll merge #5099 first and rebase this one on master.

jpelgrom (approved):
Let's do this, we can refine/expand later if needed

PR #5115, opened 2025-03-13
Apply gitattributes

TimoPtr:
## Summary

Based on the PR https://github.com/home-assistant/android/pull/5000. I've asked copilot to generate me a `.gitattributes` from the proposal from the PR and taking into account the code of our repo. I've commit it and renormalize the repo in a separate commit.
The changes are mostly CRLF to LF or the way around depending on what is set.


jpelgrom:
I think you're trusting AI too much

- The extensions list contains a lot of stuff that seems irrelevant for this repository, especially the binary files (icns? flv?).
- "Handle Unity YAML files"? There is no Unity here and I'm not sure how you consider that to become relevant in the future.

TimoPtr:
> I think you're trusting AI too much
> 
>     * The extensions list contains a lot of stuff that seems irrelevant for this repository, especially the binary files (icns? flv?).
> 
>     * "Handle Unity YAML files"? There is no Unity here and I'm not sure how you consider that to become relevant in the future.

Indeed I've been lazy and didn't check all the extension if they were in the repo. I've been over the whole list it should be good now.

I kept the .idea files because I plan to update the gitignore based on Jetbrains recommendations.

jpelgrom (approved):
I've checked .gitattributes which looks good to me now. Quickly scrolled through the other files and will trust that they are only line endings changes.

PR #5119, opened 2025-03-14
Add "Reset frontend cache" option (using new WebStorageCompat)

### LINKED ISSUE #2460: Ability to reload lovelace for features such as dom without data clear

reference:
https://community.home-assistant.io/t/pop-up-cards/351292/17

Currently the only way to make certain things work in lovelace is to fully clear data and re-setup everything.  This is a nuclear option.  My request is for the ability to clear this relevant data within the app rather than having to clear data when making these changes.  To be fair, I am not sure why this is currently required and why it doesn't work like a shift+R refresh in a browser but it would be good to be able to do this.

Ability to refresh the 'browser' cache within the app rather than clearing data.  If there is already a way, that is great.

I tried to clear cache on android, force stop the app.

**Additional context**

jpelgrom:
## Summary

This PR adds an option to clear the frontend cache using the new WebStorageCompat API in androidx.webkit 1.13.0. This new API actually clears cached/stored data by the service worker, unlike the non-compat API.

Clearing cache is a standard [troubleshooting step on the frontend issue form](https://github.com/home-assistant/frontend/blob/d42bd36a3ea40c235a59c2b97fe53b9a1f9983f8/.github/ISSUE_TEMPLATE/bug_report.yml?plain=1#L23-L24), somewhat common for users of custom components (HACS), and this allows us to [match iOS functionality](https://github.com/home-assistant/iOS/blob/871e2dfd0c99bf54fe7deedb9095d41f3d0ba906/Sources/App/Settings/DebugView.swift#L194-L212).

Closes #2460

## Screenshots

|     |Light|Dark|
|----|----|----|
|New item in troubleshooting menu|Troubleshooting menu with option to 'Reset frontend cache', light mode|Troubleshooting menu with option to 'Reset frontend cache', dark mode|
|Tapping the option will show a dialog while working|A dialog 'Resetting...' on top of the troubleshooting menu, light mode|A dialog 'Resetting...' on top of the troubleshooting menu, dark mode|
|Confirmation when done|Toast message 'Reset completed' at the bottom of the screen, light mode|Toast message 'Reset completed' at the bottom of the screen, dark mode|

n/a

## Any other notes

If you keep open Chrome remote debugging while tapping the option
- you can see that after clearing, requests are loaded from the network instead of cache
- the service worker will be unregistered and a new one will be setup and start caching _immediately after_ cache is cleared, and the web app data storage will be empty as well (which wasn't possible with the old API)

image


app/src/main/java/io/homeassistant/companion/android/settings/developer/DeveloperSettingsFragment.kt
```diff
@@ -93,14 +110,24 @@ class DeveloperSettingsFragment : DeveloperSettingsView, PreferenceFragmentCompa
...
             .show()
     }
 
+    override fun onWebViewClearCacheResult(success: Boolean) {
+        lifecycleScope.launch {
+            delay(750L) // Add a delay to prevent the dialog 'flashing' if clearing completed quickly
+            activeTaskDialog?.hide()
+            if (success) {
+                Toast.makeText(requireContext(), commonR.string.clear_webview_cache_success, Toast.LENGTH_SHORT).show()
+            }
```
TimoPtr, line 127: If it fails the user won't know anything about it, and will report in the issue that he did clear the cache, that might be misleading.

jpelgrom: Added a message on failure. This one is a bit theoretical so there's not really anything specific to add to the message, I've yet to see clearing cache fail.

app/src/main/java/io/homeassistant/companion/android/settings/developer/DeveloperSettingsFragment.kt
```diff
@@ -93,14 +110,24 @@ class DeveloperSettingsFragment : DeveloperSettingsView, PreferenceFragmentCompa
...
         AlertDialog.Builder(requireContext())
             .setTitle(commonR.string.thread_debug)
             .setMessage("${if (success == true) "✅" else if (success == null) "⚠️" else "⛔"}\n\n$result")
             .setPositiveButton(commonR.string.ok, null)
             .show()
     }
 
+    override fun onWebViewClearCacheResult(success: Boolean) {
+        lifecycleScope.launch {
+            delay(750L) // Add a delay to prevent the dialog 'flashing' if clearing completed quickly
```
TimoPtr, line 123: The flashing occur only if the cache is small enough, but if the cache is big we are adding 750ms on top of it. It's not a big deal but we could avoid this with a flow and a throttle operator. We can keep it like this since it's a debug screen it's not a big deal.

jpelgrom: That was my thought: this is a lot quicker to setup, keeps it simple and won't bother the user that much if the cache takes for example 30 seconds to clear.

dshokouhi (approved):

PR #5120, opened 2025-03-14
Apply KTLint settings to all projects

TimoPtr:
## Summary

While exporting the SARIF reports I found out that the reports were not generated for subprojects like `:app` it was because the configuration was only apply to the root project.
For `build-logic` I forgot to export the report in SARIF format since it doesn't inherit from the root project.


TimoPtr:
It looks like this 
image

TimoPtr:
I though annotation where generated by Github but from their comment
> Depending on your configuration and choice of analysis tool, future pull requests will be annotated with code scanning analysis results. For more information about GitHub code scanning, check out [the documentation](https://docs.github.com/code-security/code-scanning/introduction-to-code-scanning/about-code-scanning).

It seems that we need another tool to generate the annotation. Even if the SARIF are properly parse they are only visible inside the security section (not even within the PR).

image

The code analysis section in the tab Checks of the PR is empty even if there are issues. I wonder why (is it because we don't run the code analysis on master? I don't know)
image

I guess we could merge this PR because it fixed some configuration issue but it doesn't achieve my goal that was having Github adding specific comments into the PR about the lint issues.

dshokouhi (approved):

.github/workflows/pr.yml
```diff
@@ -8,6 +8,7 @@ jobs:
   ktlint:
     runs-on: ubuntu-latest
     permissions:
+        pull-requests: write
```
TimoPtr, line 11: To remove

PR #5126, opened 2025-03-17
Add dummy instrumentation test for app/tv/wear

TimoPtr:
## Summary

This PR is adding the necessary configuration in order to run instrumentation test on all the platform we support. I've created a very simple test that simply start the `LauncherActivity` and `SplashScreenActivity` it is already a good start and could save us from some crashes since it's going to test min and max versions for each platform.

## Any other notes

I selected different version of API while always including min and max.
Automotive is ready but we can't run test until theses PRs are not merged https://github.com/ReactiveCircus/android-emulator-runner/pull/428 and https://github.com/ReactiveCircus/android-emulator-runner/issues/429

From now on the test comment on the PR is only shown on failure of a test.

In addition now that we use a lot of runners, in order to avoid waste of resources I've made sure that no workflows are running in // for the same PR. It means that if we push a commit while a workflow is running on an opened PR the running workflow is going to be canceled.


jpelgrom:
(Do we need the comment each time? E.g. some projects only post it if a test fails.)

TimoPtr:
> (Do we need the comment each time? E.g. some projects only post it if a test fails.)

I need to check the settings but probably.

dshokouhi (approved):

jpelgrom:
The title notes app/automotive/tv/wear but:

 - automotive is disabled, correct?
 - there is no tv, correct? (In my opinion we should not test for this because we do not support it, mainly due to frontend limitations)

TimoPtr:
> The title notes app/automotive/tv/wear but:
> 
>     * automotive is disabled, correct?
> 
>     * there is no tv, correct? (In my opinion we should not test for this because we do not support it, mainly due to frontend limitations)

Automotive is indeed disabled, with the PR that I've opened on the Action we should get it. About tv I did add one target just to verify it doesn't crash, I think it's good to know the status of the platform even if for now we didn't put a lot of effort into it (from the UX point of view).

Updating the title to remove automotive

PR #5128, opened 2025-03-17
Move from java folders to kotlin folders for sources

TimoPtr:
## Summary

This PR is only about moving files from `java` folders to `kotlin`. It doesn't impact the git history of the files which is important to keep. Now all the source files are located in `kotlin`.


dshokouhi:
Are the code scanning results safe to ignore? typically we expect all actions to have green checkmarks

TimoPtr:
Yes it's because since the files moved it is going to close the known issues and create new ones.

dshokouhi (approved):

jpelgrom (approved):

PR #5129, opened 2025-03-18
Add yamllint

TimoPtr:
## Summary

This PR introduce yamllint as we are doing in https://github.com/home-assistant/core/blob/dev/.yamllint


dshokouhi:
is this in case we ever add YAML in our code or read me? or just to be in alignment with the other repos? the only time i have ever seen yaml used is when a PR demonstrates the example YAML used for the feature like when we add a notification command. Other than that only the docs.

dshokouhi (approved):

.github/workflows/pr.yml
```diff
@@ -1,14 +1,26 @@
 name: Pull Request
 
-on:
+on: # yamllint disable-line rule:truthy
   pull_request:
     branches: [master]
 
 jobs:
+  yamlint:
```
jpelgrom, line 8: Minor typo

```suggestion
  yamllint:
```

TimoPtr: good catch

Issue #3897, opened 2023-09-28
No sound response in Assist


amaciuc:
**Home Assistant Android app version(s):**
`2023.9.2-full`
**Android version(s):**
`12`
**Device model(s):**
`Samsung Note 10+`
**Home Assistant version:**
`2023.9.3`
**Last working Home Assistant release (if known):**

**Description of problem, include YAML if issue is related to notifications:**
When I am using the Voice Assistant, text-to-speech doesn't work somehow. Assistant responses with a text message, but without audio. 
I checked the `tts` notifications by calling a `notify` service, the audio works. 
In Assist pipeline (in configuration popup), at `Text-to-speech` section I tried the voices, they sound. Only when I trigger the Assist popup, the audio doesn't work.

I also tried on another Samsung smartphone (S20 FE), same issue.

Am I missing something? Do I have to change some special setting ?

**Companion App Logs:**

```
09-28 17:54:46.632 14049 14049 D WebSocketRepository: Sending message 9: {type=assist_pipeline/run, start_stage=stt, end_stage=tts, input={sample_rate=44100}, conversation_id=null, pipeline=01h9j6x70v6kxqnct5kpy3e5tj, id=9}
09-28 17:54:46.632 14049 14049 D WebSocketRepository: Message number 9 sent
09-28 17:54:46.676 14049 14270 D WebSocketRepository: Websocket: onMessage (text)
09-28 17:54:46.682 14049 14130 D WebSocketRepository: Message number 8 received
09-28 17:54:46.719 14049 14270 D WebSocketRepository: Websocket: onMessage (text)
09-28 17:54:46.721 14049 14270 D WebSocketRepository: Websocket: onMessage (text)
09-28 17:54:46.721 14049 14130 D WebSocketRepository: Message number 9 received
09-28 17:54:46.723 14049 14130 D WebSocketRepository: Message number 9 received
09-28 17:54:46.724 14049 14130 D WebSocketRepository: Message number 9 received
09-28 17:54:47.015 14049 14270 D WebSocketRepository: Websocket: onMessage (text)
09-28 17:54:47.017 14049 14134 D WebSocketRepository: Message number 9 received
09-28 17:54:48.547 14049 14270 D WebSocketRepository: Websocket: onMessage (text)
09-28 17:54:48.550 14049 14134 D WebSocketRepository: Message number 9 received
[...]

PR #5133, opened 2025-03-19
AudioUrlPlayer invoke donePlaying directly if the volume is 0

TimoPtr:
## Summary

To improve the UX while using assist with the volume down to the minimum, I've updated the AudioUrlPlayer to actually not play anything if the volume is set to 0.

The counterpart is that from now on if you turn up the volume from 0 while getting an answer you won't get any TTS. But if `continue_conversation` flag is set to true you can directly answer once the text is on the screen which make the UX faster for people that don't use TTS instead of waiting without anything happening.

It replicate the behavior of Google Assistant or Siri on iOS.

This feature is not available on the frontend.

## Screenshots

[without_tts.webm](https://github.com/user-attachments/assets/5d932e46-04ff-4552-a8b7-278d536cb02e)

## Any other notes

In the future we might revisit this if we need to still play the sound so that if we turn up the volume we get the TTS. But for now I don't think we need to handle this case.

I removed the donePlaying callback that was not needed in the first place when I introduced it.

I took the opportunity to add unit test to the `AudioUrlPlayer` to test its logic using Robolectric.


bgoncal (approved):
LGTM, can we slowly add some tests to this part of the codebase? Or much needs to be refactored first?

TimoPtr:
I mess up with the coroutine in a previous PR. I'm making it right and I'll re-open the PR after.

jpelgrom:
The volume stream used for assistants (like Assist) is not available to edit when nothing is playing on all devices. If the volume is set to 0, this may remove the option for users to increase it, as the stream is never active - right now we can at least say 'press volume up if you don't hear anything'.

TimoPtr:
Wdym by available to edit?

The idea here is to replicate how other assistants are working. If volume is set to 0 then the assistant is not going to speak and allow for a quicker answer from the user by voice.

Indeed it means that if you turn on the volume back you won't here anything. Maybe we can in the UI indicate that the volume is down.

I did this because that's how I use Google assistant (most of the time without tts) and it is frustrating from Assist to have to wait for something when you already see the answer written.

jpelgrom:
On some devices, there is no slider for the volume level when the assistant stream isn't playing. So if we never play, the slider to raise it above 0 is never available. See #3897 and I think this also applies to most Wear devices but can't find it right now.

TimoPtr:
I didn't even know that there was another stream control for assistant. We need to check how it behaves, because here we are checking the MUSIC stream so if the MUSIC stream is not 0 it will play but the user won't hear anything and we end up into the same situation and the users will be able to change it since it's playing.

On the watch I didn't test because first I need to figure out where to change the volume 🤦🏻.

jpelgrom:
I think technically it's not another stream but sometimes it is treated as such...

Watch volume is in the quick settings panel on most watches.

TimoPtr:
It's seems to also be the issue with Google Assistant on Samsung
https://eu.community.samsung.com/t5/samsung-lounge/low-volume-in-google-assistant/td-p/8837534

TimoPtr:
@jpelgrom I was thinking while using my watch that I never touch the volume and I won't touch the volume while something is playing since I need to go another screen to do that.
I propose to push like this and see the feedback that we have from the user. We could add an UI indicator or whether or not TTS is going to play.

jpelgrom:
> I was thinking while using my watch that I never touch the volume and I won't touch the volume while something is playing since I need to go another screen to do that.

Fair enough.

> We could add an UI indicator or whether or not TTS is going to play.

And what does that allow the user to do?

---

This might not be as big of a deal as I make it out to be but we definitely need to monitor it.

I don't really understand the later test code so cannot approve that part. But if you think it's good you can merge.

TimoPtr:
> And what does that allow the user to do?

Probably nothing just he knows that we are not playing any sounds because the volume is down. It works only when it is tigh to the volume of the music. If it's not then even the icon wouldn't be displayed.

It's probably not worse it, but as you said is users complains we could propose this.

PR #5136, opened 2025-03-20
Add leakCanary for debug builds

TimoPtr:
## Summary

In order to prevents leaks in our apps, I've added [Leak Canary](https://square.github.io/leakcanary/). It is very simple to setup (just 1 line). 
It definitely adds some overheads to the app and make it a bit slower in some cases (especially when it founds a leak). 
When you install the app in debug app it installs a new app named `Leaks`, you'll be able to see the leaks in the notifications and within this new app to investigate where the leaks are.

For now I didn't use the app a lot with it enabled but it didn't spot any leaks.

For the future we should add this to our instrumentation tests, so that the CI benefits from it too.


dshokouhi:
ran locally and dont see any performance impacts on the phone but the watch does worry me as its already slow to start the app, havent had a chance to check the watch but i frequently already hit #3325  in debug builds already

TimoPtr:
While integrating the lib I was considering adding a gradle flag like -PSanitizer that adds the dependency instead of always adding it. But I can actually change the default behavior if it's the app or the watch. Do you have any tough about it?

dshokouhi:
Hmm that could be good too because one thing we do occasionally is ask users to test debug APKs from the actions so being able to add it when we need to would be great.

TimoPtr:
I would like to have LeakCanary running as much as possible. Do you think it would be acceptable to do it the way around instead. Having a flag to disable it?

dshokouhi:
im good with that :)

TimoPtr:
Can be set in the gradle command with the flag `-PnoLeakCanary` for example
```bash
./gradlew app:assembleFullDebug -PnoLeakCanary
```

It can be set at the gradle.properties level too (home level, project level)
```
noLeakCanary=true
```

jpelgrom (approved):

PR #5154, opened 2025-03-26
Detect missing lockfile update on CI

TimoPtr:
## Summary

Now that we are using lockfiles sometimes we forget to update them, which lead to issues that are not easy to associate with a missing update of the lockfile. In order to explicitly spot a missing update I've added a new job that checks if we forget to push the updated lockfiles.

If it is the case it will block the CI and send an error message in the PR code.

## Screenshots

image
This is the message you are getting when you forget to update the lockfile.

## Any other notes

To make more sense into our workflow I propose this new graph
image


dshokouhi (approved):
I like the new workflow, makes sense and should be faster when easy issues are found :)

jpelgrom (approved):

PR #5168, opened 2025-03-30
Add `wear_keep.xml` to prevent ProGuard from removing necessary resou…

yschimke:
## Summary

Add wear_keep in the correct location to ensure proguard (if applied) doesn't strip Wear capabilities.

See https://github.com/google/horologist/issues/2281#issuecomment-2763461479 and https://issuetracker.google.com/issues/348688201 for discussion.

## Any other notes

- [ ] Testing with a release build including shrinkResources

### DIFF
--- a/app/src/full/res/raw/wear_keep.xml
+++ b/app/src/full/res/raw/wear_keep.xml
@@ -1,0 +1,2 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:tools="http://schemas.android.com/tools" tools:keep="@array/android_wear_capabilities"/>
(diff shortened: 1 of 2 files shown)


yschimke:
Note: I am not seeing this lost, perhaps because minify is not enabled. But is mainly preventative.

TimoPtr:
Thanks @yschimke for your contribution. For now minify is indeed not enabled.

app/src/full/res/raw/wear_keep.xml
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:tools="http://schemas.android.com/tools" tools:keep="@array/android_wear_capabilities"/>
```
TimoPtr, line 2: Any reason to put it in `full` flavor and not `main`?

yschimke: I thought main didn't include the capabilities, only full. I'll confirm and move.

In truth we could move this to common and have a single keep file.

TimoPtr: Indeed you are right that wear only works on `full` 👍🏻

PR #5170, opened 2025-03-30
Add WearDns for handling DNS resolution on Wear devices

yschimke:
## Summary

This commit introduces a new `WearDns` class to handle DNS resolution on Wear OS devices. It attempts to use the system DNS and falls back to mobile DNS if there's an issue.

The implementation includes:
- `mobile_network_helper`: Constant for the mobile DNS capability.
- `/dnsLookup`: Constant for the mobile DNS request.

## Any other notes

I'm experimenting to address https://github.com/home-assistant/android/issues/1866

If this isn't working, then next option to explore is requiring Wifi.  See https://google.github.io/horologist/network-awareness/

## TODO

- [x] Add unit tests


yschimke:
Open Questions

- [ ] Should this be a user preference
- [ ] Can we detect specific cases to retry (BT active, lan/local hostname)

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDataListener.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    override fun onRequest(nodeId: String, path: String, request: ByteArray): Task<ByteArray>? {
+        if (path == "/dnsLookup") {
+            val hostname = String(request, Charsets.UTF_8)
+            return mainScope.async(Dispatchers.IO) {
+                Dns.SYSTEM.lookup(hostname).first().address
+            }.asTask()
+        } else {
+            // Not supported by this listener
+            return null
```
TimoPtr, line 27: If I'm not wrong since we have a filter in the manifest based on `/dnsLookup` we should never reach this line correct? If so we should at least add a warning.
```suggestion
            Timber.w("Received a path ($path) that is not supported by this listener. Check the manifest intent-filter")
            return null
```

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDataListener.kt
```diff
@@ -0,0 +1,36 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.async
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.tasks.asTask
+import okhttp3.Dns
+
+@AndroidEntryPoint
+class WearDataListener : WearableListenerService() {
```
TimoPtr, line 15: How do you envision this class to evolve? The name seems quite generic. Could you add a description within the code of the scope of the listener?

yschimke: Yep, happy for guidance here. 

I'll document the current scope (mobile providing helper APIs for networking/data).

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,108 @@
...
+        val cached = dnsHelperCache[hostname]
+        if (cached != null && cached.storedAt.plus(5L, ChronoUnit.MINUTES) > now) {
+            when (cached) {
+                is NegativeCacheHit -> throw cached.exception
+                is PositiveCacheHit -> return cached.value
+            }
+        }
+
+        try {
+            val nodeId = nodeIdWithDns() ?: return null
```
TimoPtr, line 57: You could also throw here saying "No mobile device available for DNS lookup" and always returning a non nullable list.

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private suspend fun attemptLookupViaMobile(
+        hostname: String,
+        e: Exception
+    ): List<InetAddress>? {
+        val now = Instant.now()
+
+        val cached = dnsHelperCache[hostname]
+        if (cached != null && cached.storedAt.plus(5L, ChronoUnit.MINUTES) > now) {
```
TimoPtr, line 49: So if the DNS resolution on mobile failed for any reason you have to wait 5mn to try again on mobile? That's probably too aggressive no?

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,108 @@
...
+        val storedAt: Instant
+    }
+
+    data class NegativeCacheHit(val exception: Exception, override val storedAt: Instant) : CacheResult
+    data class PositiveCacheHit(val value: List<InetAddress>, override val storedAt: Instant) : CacheResult
+
+    companion object {
+        private const val CAPABILITY_DNS_VIA_MOBILE = "mobile_network_helper"
+        private const val REQUEST_DNS_VIA_MOBILE = "/dnsLookup"
+    }
```
TimoPtr, line 107: Could you share this in `:common` Gradle module it will avoid typos if it changed one day.

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            throw e
+        }
+    }
+
+    private suspend fun attemptLookupViaMobile(
+        hostname: String,
+        e: Exception
+    ): List<InetAddress>? {
```
TimoPtr, line 45:
```suggestion
    private suspend fun attemptLookupViaMobile(
        hostname: String
    ): List<InetAddress> {
```

I think this function should not know anything about the previous failure and always return a list or throw.

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDataListener.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    private val mainScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
+
+    override fun onRequest(nodeId: String, path: String, request: ByteArray): Task<ByteArray>? {
+        if (path == "/dnsLookup") {
+            val hostname = String(request, Charsets.UTF_8)
+            return mainScope.async(Dispatchers.IO) {
+                Dns.SYSTEM.lookup(hostname).first().address
```
TimoPtr, line 23: Why only returning the first? If I'm not wrong you could get at two InnetAddress one for ipv4 and one for ipv6. The wear could benefits having the two.

It avoid having some logic here you just forward whatever you received without transformation.

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDataListener.kt
TimoPtr: In that case if you envision it to be for general networking/data I would suggest to follow [Google documentation ](https://developers.google.com/android/reference/com/google/android/gms/wearable/WearableListenerService)

> For best performance, limit the number of listeners and use mutually exclusive filters where there are multiple listeners. A common pattern is to use prefix filters to listen to different namespaces within an application.

And introduce a namespace in the intent-filter something like `android:pathPrefix="/network"` and update the path to look like `/network/dnsLookup`

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,108 @@
...
+    private fun couldBeWearDnsIssue(e: Exception): Boolean {
+        return e is UnknownHostException
+    }
+
+    sealed interface CacheResult {
+        val storedAt: Instant
+    }
+
+    data class NegativeCacheHit(val exception: Exception, override val storedAt: Instant) : CacheResult
+    data class PositiveCacheHit(val value: List<InetAddress>, override val storedAt: Instant) : CacheResult
```
TimoPtr, line 102: Can you make this private?

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,108 @@
...
+    private suspend fun dnsViaMobile(nodeId: String, hostname: String): List<InetAddress> {
+        val response = Wearable.getMessageClient(appContext).sendRequest(
+            nodeId,
+            REQUEST_DNS_VIA_MOBILE,
+            hostname.toByteArray()
+        ).await()
+
+        if (response.isEmpty()) {
+            throw UnknownHostException("Mobile helper unable to resolve $hostname")
+        }
```
TimoPtr, line 87: I'm not sure we should do that, the normal behavior of OkHttp is just passing the empty list down the chain. If you received an empty list for the API forward it OkHttp will fail later.

Since we inject this class in the HomeAssistantAPI I rather replicate exactly what the default behavior does.

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,108 @@
...
+
+        @Suppress("BlockingMethodInNonBlockingContext")
+        return listOf(InetAddress.getByAddress(hostname, response))
```
TimoPtr, line 90:
```suggestion
        return listOf(InetAddress.getByAddress(hostname, response))
```
I'm not sure you need that. The documentation is saying that only the object is created, I don't think it is doing anything blocking here.

TimoPtr (changes requested):
I think that for now we don't need a user preference for it (but it can be added later if needed). But I think it worse mentioning this behavior inside the https://companion.home-assistant.io/ (github https://github.com/home-assistant/companion.home-assistant).

For the retry I don't know we can go with your proposal and see if people are raising new issues about it.

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDataListener.kt
yschimke: Yep. Should I json encode the request and response? Would be more inline with other uses?

This actually reminds me, we should prefer IPv4, as I'm not sure the BT proxied connection supports IPv6.

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
yschimke: I need it because UnknownHostException is declared, as an IOException it flags the linter incorrectly.

TimoPtr:
Would you mind adding some Unit test? I've added recently all the necessary tools to add them.

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDataListener.kt
```diff
@@ -0,0 +1,36 @@
...
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.async
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.tasks.asTask
+import okhttp3.Dns
+
+@AndroidEntryPoint
+class WearDataListener : WearableListenerService() {
+
+    private val mainScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
```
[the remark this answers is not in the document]
yschimke: Ideally each of these would implement LifecycleService, which I believe defaults to Main.

yschimke:
> Would you mind adding some Unit test? I've added recently all the necessary tools to add them.

@TimoPtr testing is hard with the DataLayer. In either instrumentation (need paired devices), or unit test (robolectric with custom shadows?).

I'll refactor to have some API we control for the DataLayer parts and test the rest of the structure.

I'm assuming unit tests based on https://github.com/home-assistant/android/blob/master/common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/entities/ConversationResponseTest.kt

TimoPtr:
> > Would you mind adding some Unit test? I've added recently all the necessary tools to add them.
> 
> @TimoPtr testing is hard with the DataLayer. In either instrumentation (need paired devices), or unit test (robolectric with custom shadows?).
> 
> I'll refactor to have some API we control for the DataLayer parts and test the rest of the structure.
> 
> I'm assuming unit tests based on this being recently added https://github.com/home-assistant/android/blob/master/common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/entities/ConversationResponseTest.kt
> 
> And put them under wear/src/test/kotlin

Yes unit test, unit test with robolectric and instrumentation test have been setup recently. For now it only contains very simple tests, I used them to work on the CI that run everything now. Lemme know if you need anything for writing the tests. I can help since it's still WIP.

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDataListener.kt
TimoPtr: I'm ok with JSON. Even if we should prefer IPv4 let's not take any decision at this level. The wear should deal with that not the mobile.

TimoPtr:
@yschimke do you plan to continue this PR?

yschimke:
Yep - apologies, slipped my mind. I'll try to update in the next 2 weeks.

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,108 @@
...
+
+        try {
+            val nodeId = nodeIdWithDns() ?: return null
+            val addresses = dnsViaMobile(nodeId, hostname)
+
+            dnsHelperCache[hostname] = PositiveCacheHit(addresses, now)
+
+            return addresses
+        } catch (e2: Exception) {
+            e2.addSuppressed(e)
```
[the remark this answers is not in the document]
yschimke: Done, but I think e is too strong, since we expect this. PTAL.

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDataListener.kt
yschimke: Went csv

yschimke:
I'll add unit tests next.  Also still need to test again on device after latest changes.

yschimke:
@TimoPtr I forgot about your suggestion of no setting, and added one. Should I remove?

TimoPtr:
> @TimoPtr I forgot about your suggestion of no setting, and added one. Should I remove?

I'm fine with the settings on the watch. You'll need to add a screenshot in the PR description and also a link to the companion app documentation that adds a mention to this option in the troubleshooting section I would say.

app/build.gradle.kts
```diff
@@ -39,6 +39,8 @@ firebaseAppDistribution {
 dependencies {
     // Most of the dependencies are coming from the convention plugin to avoid duplication with `:automotive` module.
     "fullImplementation"(libs.car.projected)
+
+    implementation(libs.androidx.lifecycle.service)
```
TimoPtr: I think this should go in `AndroidApplicationDependenciesConventionPlugin` otherwise automotive module won't have it.

yschimke:
The setting doesn't work as you can't get to it before selecting a server.  Removing.

yschimke:
I'll follow up with unit tests tomorrow

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDnsRequestListener.kt
```diff
@@ -0,0 +1,46 @@
...
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.tasks.asTask
+import kotlinx.coroutines.withContext
+import okhttp3.Dns
+import timber.log.Timber
+
+@AndroidEntryPoint
+class WearDnsRequestListener : WearableListenerService() {
+
+    private val mainScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
```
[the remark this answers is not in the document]
yschimke: But each individual I/O operation is protected by `withContext(Dispatchers.IO) {`. This is correct code when handling different types of requests, calling a main safe coroutine.

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,109 @@
...
+        val storedAt: Instant
+    }
+
+    private data class NegativeCacheHit(val exception: Exception, override val storedAt: Instant) : CacheResult
+    private data class PositiveCacheHit(val value: List<InetAddress>, override val storedAt: Instant) : CacheResult
+
+    companion object {
+        private const val CAPABILITY_DNS_VIA_MOBILE = "mobile_network_helper"
+        private const val REQUEST_DNS_VIA_MOBILE = "/network/dnsLookup"
+    }
```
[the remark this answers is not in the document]
yschimke: documentation and code are both "/network/dnsLookup"

"mobile_network_helper" is something else

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,109 @@
...
+class WearDns @Inject constructor(@ApplicationContext private val appContext: Context) : Dns {
+    private val dnsHelperCache = ConcurrentHashMap<String, CacheResult>()
+
+    override fun lookup(hostname: String): List<InetAddress> {
+        return try {
+            Dns.SYSTEM.lookup(hostname)
+        } catch (e: UnknownHostException) {
+            return runBlocking {
+                attemptLookupViaMobile(hostname, e)
+            }
```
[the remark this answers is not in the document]
yschimke: The api is from a library okhttp3.Dns that can't be changed

wear/src/test/kotlin/io/homeassistant/companion/android/data/WearDnsTest.kt
```diff
@@ -0,0 +1,43 @@
...
+class WearDnsTest {
+
+    private val context = ApplicationProvider.getApplicationContext<Context>()
+    private val dns = WearDns(FakeMessageClient(context), FakeCapabilityClient(context))
+
+    val homeAssistantLocal = InetAddress.getByAddress("homeassistant.local", byteArrayOf(192.toByte(), 168.toByte(), 0, 23))
+
+    @Test
+    fun `defaults to system dns`() {
+        val shadowDns = shadowOf(Dns.SYSTEM)
```
yschimke, line 46: @TimoPtr this is the approach for unit testing, does it seem ok to you?

app/src/full/AndroidManifest.xml
```diff
@@ -145,6 +145,15 @@
                 android:name="home:0:preferred"
                 android:value=""/>
         </service>
+
+        <service android:name=".data.wear.WearDnsRequestListener" android:exported="true"
+            tools:ignore="ExportedService">
```
TimoPtr, line 150: There is no permission we can attach it to?

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDnsRequestListener.kt
```diff
@@ -0,0 +1,46 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.async
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.tasks.asTask
+import kotlinx.coroutines.withContext
+import okhttp3.Dns
+import timber.log.Timber
+
+@AndroidEntryPoint
```
TimoPtr: I don't see any usage to Hilt I think this should be removed.

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDnsRequestListener.kt
```diff
@@ -0,0 +1,46 @@
...
+class WearDnsRequestListener : WearableListenerService() {
+
+    private val mainScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
+
+    override fun onRequest(nodeId: String, path: String, request: ByteArray): Task<ByteArray>? {
+        if (path == "/network/dnsLookup") {
```
TimoPtr: Could we share this variable between the two modules through common? To make sure that if one day we change something it affect both side.

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDnsRequestListener.kt
```diff
@@ -0,0 +1,46 @@
...
+        } else {
+            Timber.w("Received a path ($path) that is not supported by this listener. Check the manifest intent-filter")
+            return null
+        }
+    }
+
+    private suspend fun dnsRequest(request: ByteArray): ByteArray = withContext(Dispatchers.IO) {
+        val hostname = String(request, Charsets.UTF_8)
+        try {
+            Dns.SYSTEM.lookup(hostname).joinToString(",") { it.address.toHexString() }.encodeToByteArray()
```
TimoPtr: I would be also nice the share a place where we encode/decode the data

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDnsRequestListener.kt
```diff
@@ -0,0 +1,46 @@
...
+        } catch (uhe: UnknownHostException) {
+            byteArrayOf()
+        }
```
TimoPtr, line 41: We could maybe add a warning log here.

app/src/full/kotlin/io/homeassistant/companion/android/data/wear/WearDnsRequestListener.kt
jpelgrom: We have WearDataMessages in common for that purpose (partially)

wear/src/test/kotlin/io/homeassistant/companion/android/data/WearDnsTest.kt
TimoPtr: Yes it's fine 👍🏻 You could also use `mockk` if you want if it makes things a bit easier to setup.

TimoPtr: You have more info here https://developers.home-assistant.io/docs/android/testing/unit_testing if you can follow the Given-When-Then approach while writing the test it would be nice.

wear/src/test/kotlin/io/homeassistant/companion/android/data/ShadowWearable.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.google.android.gms.wearable.Wearable
+import com.google.android.gms.wearable.internal.zzao
+import java.net.InetAddress
+import org.robolectric.annotation.Implementation
+import org.robolectric.annotation.Implements
+import org.robolectric.shadow.api.Shadow
+
+
+@Implements(Wearable::class, isInAndroidSdk = false)
+class ShadowWearable {
```
TimoPtr: I think you are not using this one right?

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,111 @@
+package io.homeassistant.companion.android.data
+
+import com.google.android.gms.wearable.CapabilityClient
+import com.google.android.gms.wearable.MessageClient
+import com.google.android.gms.wearable.Wearable
+import java.net.InetAddress
+import java.net.UnknownHostException
+import java.time.Instant
```
TimoPtr: I would be in favor of using `import kotlin.time.Instant` instead and a `kotlin.time.Clock` that you get in the constructor that default to `Clock.System` so you can override it in your test since you are using `now()` to check the cache.

It will also allow you to make this test a bit nicer
```
        if (cached != null && cached.storedAt + 5.minutes > now) {
```

wear/src/test/kotlin/io/homeassistant/companion/android/data/ShadowWearable.kt
yschimke: I will be, it's in progress. Moved back to a draft.

app/src/full/AndroidManifest.xml
yschimke: Not that I'm aware of, I can ask.

wear/src/test/kotlin/io/homeassistant/companion/android/data/WearDnsTest.kt
yschimke: Does mockk replace Robolectric shadows?

TimoPtr: It probably could it can mock static function you have an example https://github.com/home-assistant/android/blob/fe823549ce08418e15271decc8ffe98b3965ef96/app/src/test/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiverTest.kt#L199 where I used it to mock `GlanceAppWidgetKt`.

wear/src/test/kotlin/io/homeassistant/companion/android/data/ShadowWearable.kt
yschimke: You are right, removed.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/keychain/KeyStoreRepositoryImpl.kt
```diff
@@ -95,12 +96,18 @@ class KeyStoreRepositoryImpl @Inject constructor() : KeyChainRepository {
...
+            keyStore().apply {
                 load(null)
                 setEntry(alias, PrivateKeyEntry(key, chain), null)
             }
         } catch (e: Exception) {
             Timber.e(e, "Exception storing KeyStore.Entry")
         }
     }
+
+    private fun keyStore(): KeyStore = if ("robolectric" == Build.FINGERPRINT) {
```
yschimke, line 108: This doesn't fail the robolectric tests, but without this it's harder to see the real error.

common/src/main/kotlin/io/homeassistant/companion/android/di/DataModule.kt
```diff
@@ -134,6 +137,11 @@ abstract class DataModule {
         @Singleton
         fun providesTextToSpeechClient(@ApplicationContext appContext: Context): TextToSpeechClient =
             TextToSpeechClient(appContext, AndroidTextToSpeechEngine(appContext))
+
+        @OptIn(ExperimentalTime::class)
+        @Provides
+        @Singleton
+        fun clock(): Clock = Clock.System
```
TimoPtr: It feels a bit weird to provide a clock in the DataModule. I would expect the clock to be injected from another module, like `ApplicationModule` since it can be use in a lot of places.

gradle/libs.versions.toml
```diff
@@ -151,7 +153,7 @@ car-core = { module = "androidx.car.app:app", version.ref = "car-versions" }
 car-automotive = { module = "androidx.car.app:app-automotive", version.ref = "car-versions" }
 car-projected = { module = "androidx.car.app:app-projected", version.ref = "car-versions" }
 changeLog = { module = "com.github.AppDevNext:ChangeLog", version.ref = "changeLog" }
-coil-oktthp = { module = "io.coil-kt.coil3:coil-network-okhttp", version.ref = "coil" }
+coil-okhttp = { module = "io.coil-kt.coil3:coil-network-okhttp", version.ref = "coil" }
```
TimoPtr, line 156: Thanks for catching the typo

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,111 @@
...
+                attemptLookupViaMobile(hostname, e)
+            }
+        }
+    }
+
+    private suspend fun attemptLookupViaMobile(hostname: String, e: Exception): List<InetAddress> {
+        val now = clock.now()
+
+        val cached = dnsHelperCache[hostname]
+        if (cached != null && cached.storedAt + 5.minutes > now) {
```
TimoPtr: Nit pick: 5.minutes could be a well name file level val.

wear/src/test/kotlin/io/homeassistant/companion/android/data/WearDnsTest.kt
```diff
@@ -0,0 +1,102 @@
...
+    private val context = ApplicationProvider.getApplicationContext<Context>()
+    val messageClient = FakeMessageClient(context)
+    val capabilityClient = FakeCapabilityClient(context)
+    val clock = FakeClock()
+    private val dns = WearDns(messageClient, capabilityClient, clock)
+
+    val homeAssistantLocal = InetAddress.getByAddress("homeassistant.local", byteArrayOf(192.toByte(), 168.toByte(), 0, 23))
+
+    @Test
+    fun `defaults to system dns`() {
```
TimoPtr:
```suggestion
    fun `Given a hostname when making DNS lookup then returns DNS entry`() {
```
If you could adjust the name of the test to a sentence that self explain the test and give the context of the test 🙏🏻

wear/src/test/kotlin/io/homeassistant/companion/android/data/WearDnsTest.kt
```diff
@@ -0,0 +1,102 @@
...
+
+        // when
+        val exception = assertThrows<UnknownHostException> {
+            dns.lookup("homeassistant.local")
+        }
+
+        // then
+        assertEquals("Mobile helper unable to resolve homeassistant.local", exception.message)
+    }
+}
```
TimoPtr, line 174: We miss tests on the cache mechanism.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/WearDataMessages.kt
```diff
@@ -20,4 +24,20 @@ object WearDataMessages {
...
+            it.address.toHexString()
+        }.encodeToByteArray()
+
+        fun ByteArray.decodeResult(hostname: String): List<InetAddress> = decodeToString().split(",").map {
+            InetAddress.getByAddress(hostname, it.decodeHex().toByteArray())
+        }
+
+        fun String.encodeRequest(): ByteArray = toByteArray()
+
+        fun ByteArray.decodeRequest(): String = decodeToString()
```
TimoPtr:
```suggestion
        fun List<InetAddress>.encodeDNSResult(): ByteArray = joinToString(",") {
            it.address.toHexString()
        }.encodeToByteArray()

        fun ByteArray.decodeDNSResult(hostname: String): List<InetAddress> = decodeToString().split(",").map {
            InetAddress.getByAddress(hostname, it.decodeHex().toByteArray())
        }

        fun String.encodeDNSRequest(): ByteArray = toByteArray()

        fun ByteArray.decodeDNSRequest(): String = decodeToString()
```

To avoid wrong usage of these extension they should be very specific otherwise we might use them for something else.


CC @jpelgrom

wear/src/main/kotlin/io/homeassistant/companion/android/data/WearDns.kt
```diff
@@ -0,0 +1,116 @@
...
+    sealed interface CacheResult {
+        val storedAt: Instant
+    }
+
+    private data class NegativeCacheHit(val exception: Exception, override val storedAt: Instant) : CacheResult
+    private data class PositiveCacheHit(val value: List<InetAddress>, override val storedAt: Instant) : CacheResult
+
+    companion object {
+        val DefaultCacheLifetime = 5.minutes
+    }
```
TimoPtr: Should be private and no need to keep it in companion object according to our best practices since the file only contains one class and the property is private (or VisibleForTest)

wear/src/test/kotlin/io/homeassistant/companion/android/data/WearDnsTest.kt
```diff
@@ -0,0 +1,174 @@
...
+
+        // when
+        val results = dns.lookup("homeassistant.local")
+
+        // then
+        assertEquals(homeAssistantLocal, results.single())
+    }
+
+    @Test
+    fun `falls back to mobile dns when present`() {
```
TimoPtr: Update the 3 remaining titles that are not using GivenWhenThen

app/src/test/kotlin/io/homeassistant/companion/android/data/wear/WearDnsRequestListenerTest.kt
```diff
@@ -0,0 +1,38 @@
...
+        shadowDns.results["homeassistant.local"] = Result.success(listOf(homeAssistantLocal))
+
+        // When
+        val response = runBlocking { service.dnsRequest("homeassistant.local".encodeDNSRequest()) }
+        val addresses = response.decodeDNSResult("homeassistant.local")
+
+        // Then
+        assertEquals("homeassistant.local", addresses.single().hostName)
+        assertEquals("192.168.0.23", addresses.single().hostAddress)
+    }
```
TimoPtr, line 40:
```suggestion
    @Test
    fun `Given domain when request DNS lookup then returns IP address`() = runTest {
        // Given
        val controller = Robolectric.buildService(WearDnsRequestListener::class.java)
        val service = controller.create().get()

        val shadowDns = shadowOf(Dns.SYSTEM)
        shadowDns.results["homeassistant.local"] = Result.success(listOf(homeAssistantLocal))

        // When
        val response = service.dnsRequest("homeassistant.local".encodeDNSRequest())
        val addresses = response.decodeDNSResult("homeassistant.local")

        // Then
        assertEquals("homeassistant.local", addresses.single().hostName)
        assertEquals("192.168.0.23", addresses.single().hostAddress)
    }
```

TimoPtr: You should make `dnsRequest` private and only test the public interface `onRequest`. Tests the two branch when the `path` is PATH_DNS_LOOKUP and something else.

We should also tests what happens when `UnknownHostException` is thrown.

yschimke: It involves coroutines, and didn't just work in the robolectric test, which is why I opted for something simpler.

I would copy the approaches from WearOnboardingListener and PhoneSettingsListener, but it doesn't look they have tests.

yschimke:
Spent too much time on what I thought would be a quick fix. If I find some time later, I'll pick it up and continue.

Thanks for the reviews so far.

TimoPtr:
> Spent too much time on what I thought would be a quick fix. If I find some time later, I'll pick it up and continue.
> 
> Thanks for the reviews so far.

You're very close to the finish line. If you are fine with it I can apply my last comments and merge it 👍🏻

yschimke:
@TimoPtr go for it. Appreciated.

TimoPtr:
@yschimke would you mind have a quick look to my changes?

yschimke:
LGTM, thanks.

I'd suggest living with testing `dnsRequest` instead of `onRequest` because it keeps the tests much simpler.

TimoPtr:
> LGTM, thanks.
> 
> I'd suggest living with testing `dnsRequest` instead of `onRequest` because it keeps the tests much simpler.

Indeed, I didn't see during the review that it was a task from Google. I just tested the else branch

TimoPtr (approved):

dshokouhi:
So after this merge should we reach out to users impacted by #1866 and see if this helps?

yschimke:
Thanks for the help landing this. it was working for me testing locally, but the real test will be on real networks. So any user feedback sooner can confirm or identify any follow up.

PR #5183, opened 2025-04-01
Set Renovate versioning for Guava to use -android and ignore -jre

jpelgrom:
## Summary

This PR tries to 'fix' #5165, with Renovate trying to update our Guava dependency from `version-android` to `version-jre` because j > a. As we are building an Android app we should use the `-android` dependency so that change is unwanted.

Setting the `versionScheme` should prevent this based on [this comment](https://github.com/renovatebot/config-help/issues/232#issuecomment-491826549). Searching for code on GitHub shows the code submitted in this PR as more common than the one in the issue comment, so I picked the search results code.

## Screenshots

n/a

n/a

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -40,7 +40,7 @@
             "packagePatterns": [
                 "^com.google.guava:"
             ],
-            "versionScheme": "docker"
+            "versioning": "docker"
         },
         {
             "groupName": "org.jetbrains.kotlin:*",
--- a/renovate.json5
+++ b/renovate.json5
@@ -37,6 +37,12 @@
             "registryUrls": ["https://repo1.maven.org/maven2/", "https://plugins.gradle.org/m2/", "https://dl.google.com/android/maven2/", "https://jitpack.io"]
         },
         {
+            "packagePatterns": [
+                "^com.google.guava:"
+            ],
+            "versionScheme": "docker"
+        },
+        {
             "groupName": "org.jetbrains.kotlin:*",
             "matchPackageNames": [
                 "org.jetbrains.kotlin:{/,}**",
--- a/renovate.json5
+++ b/renovate.json5
@@ -40,7 +40,7 @@
             "packagePatterns": [
                 "^com.google.guava:"
             ],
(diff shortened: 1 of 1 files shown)


renovate.json5
```diff
@@ -36,6 +36,12 @@
             "matchDatasources": ["maven"],
             "registryUrls": ["https://repo1.maven.org/maven2/", "https://plugins.gradle.org/m2/", "https://dl.google.com/android/maven2/", "https://jitpack.io"]
         },
+        {
+            "packagePatterns": [
+                "^com.google.guava:"
+            ],
+            "versionScheme": "docker"
```
TimoPtr, line 43: If it's working is there any way to find the underlying versionScheme regex and tweak it?

It's weird to see docker here. We could come with a regex that ignore what is after - ?

jpelgrom: I'm not that familiar with Renovate, just searched for the issue and this seems to be a solution that is also used by quite a lot of other repositories. If you can do something with a regex, feel free to do so.

jpelgrom: Based on another search it is renamed to `versioning` (will push a change for that) and a list of values, not a regex: https://docs.renovatebot.com/configuration-options/#versioning

TimoPtr: Actually I think the way to fix what we want is with the parameter above in the documentation `versionCompatibility`

something like this should do the trick

```json
            "versionCompatibility": "^(?<version>.*)-android$"
            "versioning": "semver"
```

I'll push the change and check if it works today

TimoPtr:
It almost worked 
```
DEBUG: GET https://repo1.maven.org/maven2/com/google/guava/guava/33.4.6/guava-33.4.6.pom = (code=ERR_NON_2XX_3XX_RESPONSE, statusCode=404 retryCount=0, duration=4)
DEBUG: Rejected release
{
  "datasource": "maven"
  "packageName": "com.google.guava:guava"
  "registryUrl": "https://repo1.maven.org/maven2/"
  "version": "33.4.6"
}
```
It striped the suffix

PR #5186, opened 2025-04-02
Update readme to introduce developer documentation

TimoPtr:
## Summary

This PR is the follow-up of the creation of the developer documentation https://github.com/home-assistant/developers.home-assistant/pull/2626. Like in core [repository](https://github.com/home-assistant/core) we don't want to repeat ourselves and we want to make sure the documentation is in one place.


jpelgrom:
This is quite different in style from other Home Assistant project readme files. I understand that you're moving documentation to different pages but this feels like too much end user-focused and fluff to me.

(Personally I'm also not a fan of emoji everywhere.)

PR #5190, opened 2025-04-04
Add a setting to split device controls into multiple structures when you have multiple servers

### LINKED ISSUE #5179: Set a structure in device controls to separate multiple servers

Currently, multiple servers are set as being different zones under Home Assistant in device controls. This can make things messy when you have multiple controls from two servers and compromises the separation that otherwise exists within the companion app.

I'd like for Home Assistant to use the built in [`setStructure`](https://developer.android.com/reference/android/service/controls/Control.StatelessBuilder#setStructure(java.lang.CharSequence)) method when building the device controls items to separate controls for different servers into different structures. This would allow for the user to then click a dropdown menu at the top of the device controls panel to switch servers (much like if you have multiple apps in device controls).

The main alternative is to keep the current system. I can understand why some might prefer this, but I think those with many controls and several servers would prefer to be able to separate the servers.

**Additional context**

Here's some screenshots of what it would look like. I've just used a different app for these purposes, but you can get a bit of an idea of how it will operate:

Image

Image

As far as I can tell, it should just be necessary to change the way `setZone` works here: https://github.com/home-assistant/android/blob/587659dc9d95c0733498730d69b3a66f243b9ebf/app/src/main/kotlin/io/homeassistant/companion/android/controls/HaControl.kt#L46 to use `setStructure` in the multi-server scenario. I'm happy to do a PR for this, but thought it was worth checking to see if such a proposal was likely to be accepted first.

SheaSmith:
## Summary
Essentially, this PR provides an option to split multiple home assistant servers into different 'structures' within the Android device controls menu. This means controls for different servers will appear on separate pages within device controls, rather than needing to be bundled into the same page.

## Screenshots

image
image
image
image

Documentation: home-assistant/companion.home-assistant#1190

## Any other notes

This pull request would close #5179


dshokouhi:
Does this have any impact on the add controls screen? Does it change at all from what it was or does this only impact after controls were added?

SheaSmith:
> Does this have any impact on the add controls screen? Does it change at all from what it was or does this only impact after controls were added?

Sorry, should've included some screenshots of that as well. The add controls screen will look like this if you have the structures split enabled:

image
Screenshot_20250405-090333

You can navigate between them by swiping the body of the page.

dshokouhi:
Thanks for sharing thats awesome, but surprisingly subtle and probably going to be easy to miss 😅 out of the apps control obviously.

app/src/main/kotlin/io/homeassistant/companion/android/controls/HaControlInfo.kt
```diff
@@ -9,5 +9,6 @@ data class HaControlInfo(
     val serverName: String? = null,
     val area: AreaRegistryResponse? = null,
     val authRequired: Boolean = false,
-    val baseUrl: String? = null
+    val baseUrl: String? = null,
+    val splitMultiServerIntoStructure: Boolean
```
TimoPtr, line 13: You should probably also add a default value, to follow the design of this class.

app/src/main/kotlin/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
```diff
@@ -249,7 +250,15 @@ class HaControlsProviderService : ControlsProviderService() {
...
         subscriber: Flow.Subscriber<in Control>
     ) {
-        if (serverManager.getServer(serverId) == null) {
+        val serverCount = serverManager.defaultServers.size
+        val server = serverManager.getServer(serverId)
+
+        var serverName: String? = null
+        if (server != null && serverCount > 1) {
+            serverName = server.friendlyName
+        }
```
TimoPtr, line 259: In the scope of this PR I understand why we need this. But imagine opening this function in 6month will you or someone else understand why we do this? The function is already huge it's quite hard to follow. I think that you should add a small comment saying why you are doing this.

SheaSmith: Thanks for the feedback. I've addressed both points in the latest commit.

app/src/main/kotlin/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
```diff
@@ -147,7 +147,8 @@ class HaControlsProviderService : ControlsProviderService() {
                                     entityId = entity.entityId,
                                     serverId = serverId,
                                     serverName = serverNames[serverId],
-                                    area = getAreaForEntity(entity.entityId, serverId)
+                                    area = getAreaForEntity(entity.entityId, serverId),
+                                    splitMultiServerIntoStructure = splitMultiServersIntoStructures()
```
jpelgrom, line 151: You only need to fetch this value once as it is the same for all entities/servers. Very minor performance improvement but can this be placed outside the loop?

app/src/main/kotlin/io/homeassistant/companion/android/settings/controls/views/ManageControlsView.kt
```diff
@@ -119,6 +124,25 @@ fun ManageControlsView(
         }
 
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.UPSIDE_DOWN_CAKE || !panelEnabled) {
+            if (serversList.size > 1) {
+                item {
+                    Row(
+                        verticalAlignment = Alignment.CenterVertically,
+                        modifier = Modifier.padding(start = 16.dp, bottom = 16.dp)
```
jpelgrom, line 131: I feel like this also needs end padding; your screenshot shows the toggle almost touching the right edge of the screen.

SheaSmith: Good point. It would also have been an issue on smaller devices as well, so I've added some padding (and fixed some issues with the text wrapping):
Screenshot_20250412-101247

app/src/main/kotlin/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
```diff
@@ -265,7 +281,8 @@ class HaControlsProviderService : ControlsProviderService() {
                         systemId = it,
                         entityId = entityId,
                         serverId = serverId,
-                        area = getAreaForEntity(entity.entityId, serverId)
+                        area = getAreaForEntity(entity.entityId, serverId),
+                        splitMultiServerIntoStructure = splitMultiServersIntoStructures
```
jpelgrom, line 285: A `serverName` is not supplied here so it will never split, you can remove this line to simplify the code a bit.

dshokouhi (approved):

PR #3061, opened 2022-11-12
Make active websocket message tracking thread-safe

jpelgrom:
## Summary

Make sure that the `activeMessages` map that tracks the active messages/subscriptions on the websocket is thread-safe to prevent exceptions and incorrect responses when it is read/modified from different coroutines. Fixes #3059.

## Screenshots

n/a

## Link to pull request in Documentation repository

n/a

## Any other notes

I'm not using `ConcurrentHashMap` as the crash in the original issue showed that the app was modifying while iterating, and I want to prevent consistency errors. Wrapping map access in a `Mutex` isn't possible as it is also modified [from a `synchronized` block](https://github.com/home-assistant/android/blob/65df31816104fb76a677586e995a6afbc1a6910c/common/src/main/java/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt#L387-L396). As the map should be fairly small (a few messages at a time, maybe 15 at most) blocking the thread and the performance impact should be acceptable.

PR #5197, opened 2025-04-07
Introduce WebSocketCore for isolating the WebSocket low level logic

TimoPtr:
## Summary

In order to make it easier future refactoring and testing of the WebSocket I've decided to extract the low level logic of the `WebSocketRepository` into a dedicated class named `WebSocketCore`.
In this PR I tried to keep things as they are but I did make some small changes described in the notes bellow.

## Any other notes

- I took the opportunity to migrate time constant into proper explicit Duration.
- I added a `CoroutineExceptionHandler` in the ioScope of `WebSocketCore`
- I removed `eventTimeout` from `WebSocketRequest` that was not used anywhere
- I had hard time to keep `AssistInject` because of the dependency to `serverManager` and the need to pass the `serverId` to `WebSocketCore`. In the end I replace it with a manual Factory implementation. If you have any suggestion on how to make it work I'm all hear.
- I added some TODOs that I would like us to close in this PR by adding some comments.


common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -0,0 +1,538 @@
...
+        type: String,
+        data: Map<Any, Any>,
+        timeout: kotlin.time.Duration,
+    ): Flow<T>? {
+        val subscribeMessage = mapOf(
+            "type" to type,
+        ).plus(data)
+
+        eventSubscriptionMutex.withLock {
+            val isNewMessage = synchronized(activeMessages) {
```
TimoPtr, line 249: @jpelgrom I think all the synchronized for the activeMessages are not needed since the map is synchronized already. It would be useful to do this if we wanted to make another operation after getting the activeMessage but I don't think we do in this file at least.

jpelgrom: My understanding of this subject is not perfect, but isn't there another operation after getting the message here?

https://github.com/home-assistant/android/blob/39869d7f4b56a57eb741121198f816c2f4867fa7/common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt#L392

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/WebSocketRequest.kt
```diff
@@ -12,19 +14,16 @@ import kotlinx.coroutines.flow.SharedFlow
...
- * @param eventTimeout timeout in milliseconds for ending the subscription when the flow is no
- * longer collected
  * @param onEvent Channel that can receive events for a subscription, else `null`
  * @param onResponse Continuation for the initial response to this message. Don't set this when
  * creating this class, it will be set when a message is sent on the websocket.
  */
 data class WebSocketRequest(
     val message: Map<*, *>,
-    val timeout: Long = 30000L,
+    val timeout: Duration = 30.seconds,
```
jpelgrom, line 23: The KDoc (line 15) should also be updated to reflect this change

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketConstant.kt
jpelgrom, line 1: There are mutliple so WebSocketConstants.kt?

I think we usually put constants like these in the interface, but not sure if that matches the official style guide.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -0,0 +1,538 @@
...
+) : WebSocketCore, WebSocketListener() {
+
+    private val coroutineExceptionHandler = CoroutineExceptionHandler { ctx, err -> Timber.e(err, "Uncaught exception in WebSocketRepositoryImpl") }
+
+    // TODO We never close the scope? (Should we tight it to some other scope?)
+    // TODO check SupervisorJob or Job?
+    private val ioScope = CoroutineScope(Dispatchers.IO + Job() + coroutineExceptionHandler)
+
+    // TODO any reason to not use a ConcurrentHashMap?
+    private val activeMessages = Collections.synchronizedMap(mutableMapOf<Long, WebSocketRequest>())
```
jpelgrom, line 83: To quote myself from 2.5 years ago (#3061, use git blame!):

> I'm not using `ConcurrentHashMap` as the crash in the original issue showed that the app was modifying while iterating, and I want to prevent consistency errors.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -0,0 +1,538 @@
...
+
+    // Each message that we send needs a unique ID to match it to the answer
+    private val id = AtomicLong(1)
+    private var connection: WebSocket? = null
+    private var connectionState: WebSocketState? = null
+    private var connectionHaVersion: HomeAssistantVersion? = null
+    private val connectedMutex = Mutex()
+
+    // TODO Can't we just keep it in an AtomicBool?
+    private var connected = CompletableDeferred<Boolean>()
```
jpelgrom, line 93: It is awaited so no?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -0,0 +1,538 @@
...
+    private val eventSubscriptionMutex = Mutex()
+
+    // TODO Clarify what is the goal here
+    private val messageQueue = Channel<Job>(capacity = Channel.UNLIMITED).apply {
```
jpelgrom, line 97: This is documented in the place it is used

https://github.com/home-assistant/android/blob/39869d7f4b56a57eb741121198f816c2f4867fa7/common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt#L873

jpelgrom:
(Adding commas at the end of lines where there are otherwise no changes makes this PR unnecessarily noisy in my opinion)

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
```diff
@@ -47,111 +30,53 @@ import io.homeassistant.companion.android.common.data.websocket.impl.entities.Te
...
-import okhttp3.Response
-import okhttp3.WebSocket
 import okhttp3.WebSocketListener
-import okio.ByteString
-import okio.ByteString.Companion.toByteString
-import timber.log.Timber
 
-class WebSocketRepositoryImpl @AssistedInject constructor(
-    private val okHttpClient: OkHttpClient,
+private val matterTimeout = 2.minutes
```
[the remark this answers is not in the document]
TimoPtr: Unfortunately you can't put a `Duration` in a const only primitives and string can be const.

I tend to avoid companion objects unless they're truly necessary.  My reflex is to place private constants outside the class definition.  This practice enhances code organization and readability by reducing boilerplate, resulting in smaller, more focused classes, all without any performance impact.

Companion objects are most beneficial IMO in these specific scenarios (but not only):

-   **Namespacing for External Use:** When you *must* expose utility functions or constants externally (e.g., `internal` or `public`), a companion object provides a clear namespace. This prevents naming collisions and significantly improves code clarity for code outside the file.
-   **Intentional Naming Conflicts:** Within the *same* file, if multiple classes or entities intentionally share the same name for a conceptually similar constant or function (e.g., `EMPTY`, `DEFAULT`), a companion object offers the necessary disambiguation.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketConstant.kt
TimoPtr: Official style guide https://kotlinlang.org/docs/coding-conventions.html#property-names is not mentioning anything regarding where to put the const. 

I usually extract the const into a file when they are used between multiple files or when there are too many const to avoid making the file of the interface too big.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
TimoPtr: You are right I had to read the doc of `synchronizedMap` to find that when using the iterator you need synchronised block. 
```
  * It is imperative that the user manually synchronize on the returned
     * map when traversing any of its collection views via {@link Iterator},
     * {@link Spliterator} or {@link Stream}:
     * <pre>
     *  Map m = Collections.synchronizedMap(new HashMap());
     *      ...
     *  Set s = m.keySet();  // Needn't be in synchronized block
     *      ...
     *  synchronized (m) {  // Synchronizing on m, not s!
     *      Iterator i = s.iterator(); // Must be in synchronized block
     *      while (i.hasNext())
     *          foo(i.next());
     *  }
     * </pre>
     * Failure to follow this advice may result in non-deterministic behavior.
     ```

File: common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/WebSocketRepository.kt [hunk 7]
```diff
@@ -23,9 +23,12 @@ import io.homeassistant.companion.android.common.data.websocket.impl.entities.Te
 import io.homeassistant.companion.android.common.data.websocket.impl.entities.ThreadDatasetResponse
 import io.homeassistant.companion.android.common.data.websocket.impl.entities.ThreadDatasetTlvResponse
 import io.homeassistant.companion.android.common.data.websocket.impl.entities.TriggerEvent
+import javax.inject.Inject
+import javax.inject.Provider
 import kotlinx.coroutines.flow.Flow
 
 interface WebSocketRepository {
+    // TODO Should not return ? but instead throw an exception?
```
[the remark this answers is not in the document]
TimoPtr: Yes for me a enum would be a better fit here. I'll change it in a future PR.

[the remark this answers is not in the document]
TimoPtr: The question still stands for the rest of the methods bellow. Does it make sense to have all of them nullable? If I'm not wrong when it's null it's always because something wrong happened right?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
TimoPtr: The previous impl was using a mutable map (hashmap). Going with a ConcurrentHashMap would prevent the ConcurrentModificationException. From the documentation it says that get operations does not block the map. So for us what is important to check is if we change a value for a given key and I think we don't, the only thing we do is adding value and removing value. So if we get a value being remove from the map it won't be an issue either you will get the value before it gets removed or you will get null.

The second point we need to be careful is that the iterator should only be use in one thread and I think we are good on that front.

I am pretty sure this would not solve our issue of OOM, here it's purely to make the code easier to work with. Using `synchronized` is not straightforward and quite error prone IMO.

I won't touch it for now. But maybe in the future after some further tests. Also I don't think it will improve the performances so not urgent at all.

TimoPtr:
> (Adding commas at the end of lines where there are otherwise no changes makes this PR unnecessarily noisy in my opinion)

Fixed it has been added by Android Studio code style refactor ...

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/WebSocketRepository.kt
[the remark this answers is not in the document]
jpelgrom: Generally yes, `null` is used if the websocket didn't respond to our message or there is an exception during parsing. If we're going to change that, that should be done separately I think as that's quite a big change - but I can see it being useful to have an exception in some cases.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -0,0 +1,543 @@
...
+import okio.ByteString.Companion.toByteString
+import timber.log.Timber
+
+internal class WebSocketCoreImpl(
+    private val okHttpClient: OkHttpClient,
+    private val serverManager: ServerManager,
+    private val serverId: Int,
+) : WebSocketCore, WebSocketListener() {
+
+    private val coroutineExceptionHandler = CoroutineExceptionHandler { ctx, err -> Timber.e(err, "Uncaught exception in WebSocketRepositoryImpl") }
```
jpelgrom, line 76:
```suggestion
    private val coroutineExceptionHandler = CoroutineExceptionHandler { ctx, err -> Timber.e(err, "Uncaught exception in WebSocketCoreImpl") }
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -0,0 +1,543 @@
...
+    private var connection: WebSocket? = null
+    private var connectionState: WebSocketState? = null
+    private var connectionHaVersion: HomeAssistantVersion? = null
+    private val connectedMutex = Mutex()
+
+    /**
+     * A [CompletableDeferred] that completes with `true` when the WebSocket connection is
+     * established and authenticated successfully, or with an [Exception] when the connection.
+     *
+     * A new instance will be created when the connection is close, so it can be reuse.
```
jpelgrom, line 93: is close**d**, so it can be reuse**d**?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -0,0 +1,543 @@
...
+     * Start a subscription for events on the websocket connection and get a Flow for listening to
+     * new messages. When there are no more listeners, the subscription will automatically be cancelled
+     * using `unsubscribe_events`. If the subscription already exists, the existing Flow is returned.
+     *
+     * @param type value for the `type` key in the subscription message, for example `subscribe_events`
+     * @param data a key/value map of additional data to be included in the subscription message, for
+     *             example the `event_type` + value when subscribing with `subscribe_events`
+     * @param timeout timeout until the subscription is ended after the flow is no longer collected
+     * @return a Flow that will emit messages delivered to this subscription, or `null` if an error
+     *         occurred
```
jpelgrom, line 238: Shouldn't this (also) be on the interface, so users can actually use the documentation? You're describing parameters here, not just internals.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
```diff
@@ -47,98 +30,40 @@ import io.homeassistant.companion.android.common.data.websocket.impl.entities.Te
...
-        ioScope.launch {
-            consumeEach { it.join() } // Run a job, and wait for it to complete before starting the next one
-        }
+    override fun shutdown() {
+        return webSocketCore.shutdown()
     }
 
-    override fun getConnectionState(): WebSocketState? = connectionState
-
+    // All of this should be moved and injected through a delegate
```
jpelgrom, line 55: Is this a todo comment?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
jpelgrom: (Technically I'm describing it here as the original author, but you moved it 😉)

TimoPtr: I'm moving it to the interface. I plan to rework the whole documentation of the file in a future PR.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
TimoPtr: TODO, my idea is to have domain specific interface for instance Matter so when injected my Dagger we can get an interface for Matter instead of the whole WS impl. But it is not a must have I would say.

My feeling is that over time this class can grow significantly and it might becomes hard to use with too many methods. That's just a feeling and we can keep it like this for now.

Also the idea is that by using delegation we could keep the existing code as is without changing anything.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -0,0 +1,535 @@
...
+
+    /**
+     * A [CompletableDeferred] that signals the establishment and authentication of a WebSocket connection.
+     *
+     * This deferred value completes with `true` upon successful connection and authentication.
+     * It completes with an [Exception] if [onFailure] is invoked.
+     * It completes with an [AuthorizationException] if the authentication fail.
+     *
+     * A new instance is created when the connection closed. It is not when the authentication fail because it
+     * means it cannot recover without new credentials.
```
TimoPtr, line 97: @jpelgrom for the authentication what will happen a new instance of the Websocket needs to be recreated?

jpelgrom: I don't really follow your comment here or in the code? If authentication fails the server always closes the websocket.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
jpelgrom: What would be the benefit of multiple interfaces for the Home Assistant app, if it is the same implementation? It is not an external API with strict requirements.

As you're already saying it's not a must have, I would suggest removing the comment as there is no final decision yet.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/WebSocketRepository.kt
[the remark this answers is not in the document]
jpelgrom: OK, so maybe update the comment to make it more obvious that it is an option for improvement and applies to all the lines and not just the one directly below it.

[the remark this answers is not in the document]
TimoPtr: TODO needs to be removed before merging this PR

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
TimoPtr: Nevermind. I was wondering what is the behavior for an authentication failure. But now it's clearer for me that we are not retrying to connect on auth failure since we probably need new creds.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketConstant.kt
jpelgrom: This file name still needs to be fixed, it also violates the style from your docs PR right now.

TimoPtr: I'll address this when we merge the guideline about the const https://github.com/home-assistant/developers.home-assistant/pull/2644.

jpelgrom (approved):

PR #5199, opened 2025-04-08
Update pull request template to add more information

TimoPtr:
## Summary

This PR has been created after https://github.com/home-assistant/developers.home-assistant/pull/2626#discussion_r2023416703 on the need to add the `beta` flag in the companion app documentation. I took the opportunity to add more details and look at what we have in core [repository](https://github.com/home-assistant/core/blob/323c459442dd17dc042935f8d7c4dc711c537328/.github/PULL_REQUEST_TEMPLATE.md).

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -24,7 +24,7 @@
 -->
 
 - [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
-- [ ] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices].(https://developers.home-assistant.io/docs/android/best_practices).
+- [ ] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
 - [ ] The changes have been thoroughly tested, and edge cases have been considered.
 - [ ] The changes are backward compatible or any compatibility break is documented with previous versions of Home Assistant Core.
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,15 +1,67 @@
+<!--
+    Please review the contributing guide before submitting: https://developers.home-assistant.io/docs/android/submit
+    Please, complete the following sections to help the processing and review of your changes.
+    Please, DO NOT DELETE ANY TEXT from this template! (unless instructed).
 
-<!-- Thank you for submitting a Pull Request and helping to improve Home Assistant. Please complete the following sections to help the processing and review of your changes. Please do not delete anything from this template. -->
+    Thank you for submitting a Pull Request and helping to improve Home Assistant. You are amazing!
+-->
 
 ## Summary
-<!-- Provide a brief summary of the changes you have made and most importantly what they aim to achieve -->
+<!--
+    Provide a brief summary of the changes you have made and most importantly what they aim to achieve.
+    Don't forget any links that could be useful to the reader. (Github issues, PRs, documentation, articles, ...)
+
+    * What was the motivation behind this change?
+    * What is the impact of the changes on the application?
+-->
+
+## Checklist
+<!--
+    Put an `x` in the boxes that apply. You can also fill these out after
+    creating the PR. If you're unsure about any of them, don't hesitate to ask.
+    We're here to help! This is simply a reminder of what we are going to look
+    for before merging your code.
+-->
+
+- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
+- [ ] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices].(https://developers.home-assistant.io/docs/android/best_practices).
+- [ ] The changes have been thoroughly tested, and edge cases have been considered.
+- [ ] The changes are backward compatible or any compatibility break is documented with previous versions of Home Assistant Core.
 
 ## Screenshots
-<!-- If this is a user-facing change not in the frontend, please include screenshots in light and dark mode. -->
+<!--
+    If this is a user-facing change not in the frontend, please include screenshots in light and dark mode.
 
-## Link to pull request in Documentation repository
-<!-- Pull requests that add, change or remove functionality must have a corresponding pull request in the Companion App Documentation repository (https://github.com/home-assistant/companion.home-assistant). Please add the number of this pull request after the "#" -->
-Documentation: home-assistant/companion.home-assistant#
+    Note: Remove this section if there are no screenshots.
+-->
+
+## Link to pull request in documentation repositories
+<!-- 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,15 +1,67 @@
...
+    Instructions:
+    1. Create a pull request in the documentation repository.
+    2. Add the documentation pull request number after the "#" below.
 
-## Link to pull request in Documentation repository
-<!-- Pull requests that add, change or remove functionality must have a corresponding pull request in the Companion App Documentation repository (https://github.com/home-assistant/companion.home-assistant). Please add the number of this pull request after the "#" -->
-Documentation: home-assistant/companion.home-assistant#
+    Note: Remove this section if there are no PR.
+-->
+Developer Documentation: home-assistant/developers.home-assistant#
```
bgoncal, line 62: This depends on your another PR right? Or if that got merged, disregard my comment

TimoPtr: This PR needs to be merged after the documentation indeed.

.github/pull_request_template.md
```diff
@@ -1,15 +1,67 @@
...
+## Checklist
+<!--
+    Put an `x` in the boxes that apply. You can also fill these out after
+    creating the PR. If you're unsure about any of them, don't hesitate to ask.
+    We're here to help! This is simply a reminder of what we are going to look
+    for before merging your code.
+-->
+
+- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
+- [ ] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices].(https://developers.home-assistant.io/docs/android/best_practices).
```
jpelgrom, line 27: The second link is broken

.github/pull_request_template.md
```diff
@@ -1,15 +1,67 @@
...
+
+## Link to pull request in documentation repositories
+<!-- 
+    This pull request introduces, changes, or removes user-facing functionality.
+    A corresponding update to the Companion App documentation in the documentation repository (https://github.com/home-assistant/companion.home-assistant) is required.
+
+    Instructions:
+    1. Create a pull request in the documentation repository.
+    2. Add the documentation pull request number after the "#" below.
+    3. Add the `<span class='beta'>BETA</span> ` flag in the documentation to mark it as such.
```
jpelgrom, line 46: Wouldn't it make more sense to add this to the docs PR template?

.github/pull_request_template.md
```diff
@@ -1,15 +1,67 @@
...
-<!-- If this is a user-facing change not in the frontend, please include screenshots in light and dark mode. -->
+<!--
+    If this is a user-facing change not in the frontend, please include screenshots in light and dark mode.
+
+    Note: Remove this section if there are no screenshots.
+-->
+
+## Link to pull request in documentation repositories
+<!-- 
+    This pull request introduces, changes, or removes user-facing functionality.
```
jpelgrom, line 40: I wonder if we can merge this comment block and the next one, and end with two places to link documentation, similar to how the Home Assistant Core PR template has one big section to link to other PRs (like docs). They are both about docs and the text is 90% the same. What do you think?

.github/pull_request_template.md
```diff
@@ -1,15 +1,67 @@
...
+    A corresponding update to the Companion App documentation in the documentation repository (https://github.com/home-assistant/companion.home-assistant) is required.
+
+    Instructions:
+    1. Create a pull request in the documentation repository.
+    2. Add the documentation pull request number after the "#" below.
+    3. Add the `<span class='beta'>BETA</span> ` flag in the documentation to mark it as such.
+
+    Note: Remove this section if there are no PR.
```
jpelgrom, line 48: > if there are no PR

That should be "if there are no PRs" or "if there is no PR" (or: "if no documentation is added or updated").

.github/pull_request_template.md
TimoPtr: It could be in both actually that's a good point.

.github/pull_request_template.md
TimoPtr: All the links are broken until the doc PR is merged.

.github/pull_request_template.md
TimoPtr: I though about it while writing it. I took this approach thinking it would be more explicit for ppl that it might need 2 PRs.
I'm not fully convince about this but since the comments is hidden I think we could try like this for a while.

.github/pull_request_template.md
```diff
@@ -1,15 +1,67 @@
...
+    Put an `x` in the boxes that apply. You can also fill these out after
+    creating the PR. If you're unsure about any of them, don't hesitate to ask.
+    We're here to help! This is simply a reminder of what we are going to look
+    for before merging your code.
+-->
+
+- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
+- [ ] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices].(https://developers.home-assistant.io/docs/android/best_practices).
+- [ ] The changes have been thoroughly tested, and edge cases have been considered.
+- [ ] The changes are backward compatible or any compatibility break is documented with previous versions of Home Assistant Core.
```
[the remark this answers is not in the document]
TimoPtr: How do you want to reflect that in the checklist?

.github/pull_request_template.md
TimoPtr: Good catch 👍🏻

.github/pull_request_template.md
jpelgrom: So where is it? You added a template but nothing about marking things as beta in the text.

.github/pull_request_template.md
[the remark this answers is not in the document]
jpelgrom: I'm not completely sure what you expect with "is documented with previous versions of Home Assistant Core", but would suggest applying the measure for any breaking change.

.github/pull_request_template.md
TimoPtr: It is https://github.com/home-assistant/companion.home-assistant/blob/eb5d3a0e7bac20277d7dfe4df419e7a9daf531fb/.github/pull_request_template.md?plain=1#L11

.github/pull_request_template.md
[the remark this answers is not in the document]
jpelgrom: Okay... Still having a hard time seeing how that would work with a change in app behavior that does not depend on the server version. And if you check against a server version that suggests it isn't (always) breaking.

Simply saying backward compatible if possible or documented breaking change (which we can then communicate in a changelog, blogpost, ...) would be enough in my opinion.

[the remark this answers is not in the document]
TimoPtr:
```suggestion
- [ ] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.
```

Would it make more sense?

PR #5215, opened 2025-04-14
Add WebSocketCore unit tests

TimoPtr:
## Summary

It adds 31 unit tests to cover all the features. In order to have the test working I had to modify some pieces of the implementation.

## Any other notes

3 issues have been spotted with the help of this tests and should be address in another PR. This issues are causing OOM in productions. You can search for them with `TODO CRITICAL`

I left some TODOs with questions to address.

This PR is part of a bigger plan. After this one I will create another PR with the fixes for the 3 issues, then adding the documentation with the behavior of the webSocket. This PR then is going to be merged into #5197 and then to main.


common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -67,17 +69,33 @@ import okio.ByteString
 import okio.ByteString.Companion.toByteString
 import timber.log.Timber
 
+private const val WEBSOCKET_PATH = "api/websocket"
+
+private fun URL.toWebSocketURL(): String {
```
TimoPtr, line 74: I had this change for a while but I can remove it from this PR.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -395,10 +425,11 @@ internal class WebSocketCoreImpl(
             if (it.eventFlow == null) {
                 activeMessages.remove(id)
             }
-        }
+        } ?: { Timber.w("Response for message not in activeMessage id($id) skipping") }
     }
 
     private suspend fun handleEvent(response: SocketResponse) {
+        // TODO this probably should be in a separate class to be properly tested
```
TimoPtr, line 432: I decided to not test this part for now since it's mostly routing messages to the proper parser.

common/src/test/kotlin/io/homeassistant/companion/android/common/ConsoleLogTree.kt
```diff
@@ -0,0 +1,19 @@
...
+    override fun log(priority: Int, tag: String?, message: String, t: Throwable?) {
+        if (t != null) {
+            System.err.println("[$tag] - $message - ${t.stackTraceToString()}")
+        } else {
+            if (verbose) {
+                System.err.println("[$tag] - $message")
+            } else {
+                // no-op
+            }
+        }
```
TimoPtr, line 17: Without this there are no logs available in the AS console. First because Timber needs a tree where to send logs.
Then we need to use stderr because Junit is forwarding stdout to build the reports.... I can add a comments if you find it useful.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
TimoPtr: I'm going to remove this and keep previous impl. It's a bad idea to use URL here since it doesn't support ws protocol. This impl is also bad because it forgets the host and port.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -520,14 +551,22 @@ internal class WebSocketCoreImpl(
         }
         // If we still have flows flowing
         val hasFlowMessages = synchronized(activeMessages) { activeMessages.any { it.value.eventFlow != null } }
-        if (hasFlowMessages && ioScope.isActive) {
-            ioScope.launch {
+        if (hasFlowMessages && wsScope.isActive) {
+            wsScope.launch {
+                // TODO Why a delay ? and why 10s I guess you want to wait for the logic above to finish but 10s is quite arbitrary
+                // You could wait for the job above to finish instead
```
TimoPtr, line 557: @jpelgrom I just found out that when you turn your phone into airplane mode or if you loose the connection we retry every 10s is that the expected behavior? And 10s is to avoid trying too often?

TimoPtr: It's a nice feature but it means every 10s we try to reconnect. I do wonder if it impacts the battery usage.

jpelgrom: The delay is there for short interruptions of the connection, for example if you switch access points or when switching from Wi-Fi to mobile data. Very basic but it worked well enough. There is a battery impact but shouldn't be significant beause retrying only happens if there were subscriptions, so the app was already keeping a connection alive.

jpelgrom: Looking at making this 'smarter' in a future PR is fine by me, your question was just why is it 10s :)

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
```diff
@@ -0,0 +1,690 @@
...
+            session = ServerSessionInfo(),
+            user = ServerUserInfo(),
+        )
+
+        mockConnection = mockk(relaxed = true)
+        every { mockServerManager.getServer(any<Int>()) } returns testServer
+        every { mockServerManager.authenticationRepository(testServerId) } returns mockAuthenticationRepository
+        coEvery { mockAuthenticationRepository.retrieveAccessToken() } returns "mock_access_token"
+        // The queue is a channel if we don't use a background scope the test hangs because the channel is never close
+        // In this configuration the channel will be close when the test will end
```
jpelgrom, line 85: Should this be moved to line 104, where you're setting up the `messageQueue`? I don't see anything queue-related here.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
```diff
@@ -0,0 +1,690 @@
...
+
+    @Test
+    fun `Given failure to connect after timeout When connect is invoked Then it returns false and connection state is AUTHENTICATING`() =
+        runTest {
+            setupServer()
+
+            val result = webSocketCore.connect()
+
+            assertFalse(result)
+            // TODO do we still want to be authenticating at this stage?
```
jpelgrom, line 182: Good point, I think the answer is no. That comment should get another place though.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
```diff
@@ -0,0 +1,690 @@
...
+                match<String> { message ->
+                    message.contains(""""type":"auth"""") &&
+                        message.contains(""""access_token":"mock_access_token"""")
+                },
+            )
+        } answers {
+            assertSame(WebSocketState.AUTHENTICATING, webSocketCore.getConnectionState())
+            webSocketListener.onMessage(
+                mockConnection,
+                """{"id":1,"type":"${if (successfulAuth) "auth_ok" else "auth_invalid"}","ha_version":"$haVersion"}""",
```
jpelgrom, line 119: This is wrong? The auth message response does not include an ID

image

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
```diff
@@ -0,0 +1,690 @@
...
+        assertNull(webSocketCore.getConnectionState())
+
+        webSocketCore.sendMessage(mapOf("type" to "test"))
+
+        assertEquals(WebSocketState.ACTIVE, webSocketCore.getConnectionState())
+    }
+
+    @Test
+    fun `Given a not connectable connection When sendMessage is invoked Then it returns null and connection state remains null`() = runTest {
+        setupServer("not_connectable")
```
jpelgrom, line 308: "not connectable connection" is a bit vague in my opinion

```suggestion
    fun `Given an invalid url When sendMessage is invoked Then it returns null and connection state remains null`() = runTest {
        setupServer("invalid_url")
```

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
```diff
@@ -0,0 +1,690 @@
...
+            assertTrue(webSocketCore.connect())
+            assertEquals(WebSocketState.ACTIVE, webSocketCore.getConnectionState())
+
+            val request = slot<String>()
+            every { mockConnection.send(capture(request)) } answers { true }
+
+            webSocketCore.sendMessage(mapOf("type" to "test"))
+            advanceUntilIdle()
+
+            // We sent 1 connection then 5 messages then 1 connection then 1 message
```
jpelgrom, line 356: With `1 connection`, do you mean the `supported_features` message?

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
```diff
@@ -0,0 +1,690 @@
...
+
+            webSocketListener.onMessage(mockConnection, """{"id":2, "type":"event", "event":{"event_type":"state_changed", "time_fired":"2016-11-26T01:37:24.265429+00:00", "data": {"entity_id":"light.bed_light"}}}""")
+            assertEquals("light.bed_light", mainFlow.awaitItem().entityId)
+
+            val secondaryFlow = createSubscription("Secondary flow")
+
+            // event are not replayed when subscribing so we should not get any events here
+            secondaryFlow.expectNoEvents()
+
+            // event are received on both flow
```
jpelgrom, line 592: To make sure you get the new message, wouldn't it make more sense to emit a message with a slightly later timestamp here, or a different entity, and check for that?

jpelgrom (commented):
I like what I'm seeing, even though I wouldn't be able to write tests like these myself at this point :)

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
TimoPtr: I fixed it and remove the TODO

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
TimoPtr: Yes, it was just ignored by the parser I'm removing it 👍🏻

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
TimoPtr: If you mean the timestamp in the message I changed it. But if you mean like adding a delay. Delays in test are skipped to have a better controls. Like everything that is based on time on unit test doesn't rely on real time. `delay(30.s)` will instantly trigger.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
TimoPtr: I will handle that in the follow up PR actually since it's a bit more complicated.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
jpelgrom: I was referring to the message contents 👍

jpelgrom:
I believe you addressed the remaining comments in the code in #5215, if so and the tests pass I'm good with these changes.

PR #5219, opened 2025-04-15
Improve websocket active messages management

TimoPtr:
## Summary

This is the follow up of the changes introduced in #5197 and #5215. It proposes fixes for the TODO CRITICAL and some other improvements needed. All of the changes are controlled with unit tests but should also be checked manually.

## Checklist

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
@@ -43,7 +43,7 @@ import io.homeassistant.companion.android.common.data.websocket.impl.entities.Te
 import io.homeassistant.companion.android.common.data.websocket.impl.entities.TriggerEvent
 import java.io.IOException
 import java.net.URL
-import java.util.Collections
+import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicLong
 import kotlin.coroutines.resumeWithException
 import kotlin.time.Duration.Companion.seconds
@@ -54,11 +54,10 @@ import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.channels.Channel
-import kotlinx.coroutines.channels.ProducerScope
 import kotlinx.coroutines.channels.awaitClose
 import kotlinx.coroutines.channels.consumeEach
-import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.callbackFlow
 import kotlinx.coroutines.flow.consumeAsFlow
@@ -88,7 +87,7 @@ internal class WebSocketCoreImpl(
 ) : WebSocketCore, WebSocketListener() {
 
     @VisibleForTesting
-    val activeMessages: MutableMap<Long, WebSocketRequest> = Collections.synchronizedMap(mutableMapOf<Long, WebSocketRequest>())
+    val activeMessages = ConcurrentHashMap<Long, WebSocketRequest>()
 
     // Each message that we send needs a unique ID to match it to the answer
     private val id = AtomicLong(1)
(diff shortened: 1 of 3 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -275,70 +262,12 @@ internal class WebSocketCoreImpl(
...
-                            ?.let { activeMessages.remove(it.key) }
-                    }
-                    return null
-                }
-            }
+        return eventSubscriptionMutex.withLock<Flow<T>?> {
+            ( // Check for existing subscription before creating a new one
+                activeMessages.entries.firstOrNull { it.value.message == subscribeMessage }?.value?.eventFlow
+                    ?: createSubscriptionFlow<T>(subscribeMessage, timeout)
+                ) as? Flow<T>
```
TimoPtr, line 269: Nothing changed here except the removal of the synchronized block. I've made it this way to avoid iterating multiple times on the activeMessages.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -525,56 +508,50 @@ internal class WebSocketCoreImpl(
...
+                            )
+
+                            val response = sendMessage(newMessage)
+                            if (response == null || response.success != true) {
+                                Timber.e("Issue re-registering subscription with ${oldMessage.message}")
+                            }
                         }
-                    }
                 } else {
                     // TODO in that case what will happen to the active messages?
```
TimoPtr, line 556: I would like to tackle this in this PR but I would like your opinion on what we should do there @jpelgrom

jpelgrom: Do you mean retry vs discard? I'm thinking discard, we already did one 'invisible' retry and after that it is a failure for potentially forever so it should be handed back to the part of the app that requested it.

TimoPtr: If we discard it means that we need to make sure each subscription handle gracefully an exception on the flow. I'm not sure it is the case today but I'll check.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -155,9 +158,14 @@ internal class WebSocketCoreImpl(
...
                             )
                             Timber.d("Sending message ${supportedFeaturesMessage["id"]}: $supportedFeaturesMessage")
-                            it.send(
+                            val result = it.send(
                                 webSocketMapper.writeValueAsString(supportedFeaturesMessage),
                             )
+                            if (!result) {
+                                // Something got wrong when sending the message but we should not change the status of the
+                                // connection here. If an error occur in the WS it will be handled in the onFailure.
+                                Timber.e("Unable to send supported features message")
```
jpelgrom, line 167: As it is not fatal, I would downgrade this to a warning. In case of a major error `onFailure` will log after this one with the error level.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -220,19 +226,9 @@ internal class WebSocketCoreImpl(
 
     override suspend fun sendBytes(data: ByteArray): Boolean? {
         return if (connect()) {
-            // TODO Timeout can never happen here and it should be removed.
-            // since send returns immediately and we can't timeout a synchronized call
-            // but we should check the return of send
```
jpelgrom, line 225: Why did you decide not to check the return of send here?

jpelgrom (commented):
Note I did not check the tests, I'll do that as part of the larger tests PR.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
TimoPtr: Should be tackle in https://github.com/home-assistant/android/issues/5259 I will add it to the TODO in the code.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
TimoPtr: It is sent to the caller I didn't want to change that. I should probably still log it.

TimoPtr: While looking at the usage of this function nothing checks that it went fine. What is your take here? I think it would make sense to create an issue.

> This method returns true if the message was enqueued. Messages that would overflow the outgoing message buffer (16 MiB) will be rejected and trigger a graceful shutdown of this web socket. This method returns false in that case, and in any other case where this web socket is closing, closed, or canceled.

Today I don't think we will ever reach 16MiB since we only use it for voice. But for future usage it might be important to check the size too.

jpelgrom: An issue to track the possible improvement would be fine, but I don't see this becoming an issue anytime soon.

jpelgrom (approved):

PR #5169, opened 2025-03-30
Add todo widget

kamilwawrzyczek:
## Summary
Added todo list widget. Widget can redirect to full list view when clicked on header and it also can mark items as completed when clicked on checkbox. Resolves #2374.

## Screenshots
todo_list_config
todo_list_dark
todo_list_white

## Link to pull request in Documentation repository
Documentation: [home-assistant/companion.home-assistant#](https://github.com/home-assistant/companion.home-assistant/pull/1188)

## Any other notes
- added remote views compact library as it simplified widget with list creation
- why trailing commas in kotlin are not allowed? I needed to fix up all of my files 😅

PR #5252, opened 2025-04-28
Introduce Glance widget and refactor todo widget to use Glance

TimoPtr:
## Summary

This PR introduce the usage of [Glance](https://developer.android.com/develop/ui/compose/glance/build-ui) in the app. To make it more comprehensible with an example of usage I refactored the Todo widget introduced in https://github.com/home-assistant/android/pull/5169. I added unit tests for all the code modified, it highlighted some issues in the first implementation. 

## Checklist

## Screenshots

For reference in the screenshot our app is at the top, on the bottom right you can see Google Keep and on the left Google Calendar.

image
image
image

## Any other notes

The limitations have been documented within the code, but should be also added into the companion app once we agree on the implementation.

Few things still needs to be adjusted like the error management in the configuration screen, or the selection of the theme that by default should be DYNAMICOLOR.

The PR can be review commit per commit to make it easier.


TimoPtr:
@kamilwawrzyczek If you have some time to look at this PR.

app/src/main/AndroidManifest.xml
```diff
@@ -144,15 +144,10 @@
                 android:resource="@xml/entity_widget_info" />
         </receiver>
 
-        <receiver android:name=".widgets.todo.TodoWidget" android:label="@string/widget_todo_label"
+        <receiver android:name=".widgets.todo.TodoWidgetReceiver" android:label="@string/widget_todo_label"
```
dshokouhi, line 147: this may be a breaking change for users, meaning if they had the widget before the update, it will be removed from the home screen after. I think the widgets are linked to the manifest name here, we had an old bug. An easy test is to deploy main and add the widget then update to this branch and see if the widget stays on the home screen. Last time we updated the widget file name we found out the hard way you lose widgets 😬

build-logic/convention/src/main/kotlin/AndroidApplicationDependenciesConventionPlugin.kt
```diff
@@ -86,6 +86,10 @@ class AndroidApplicationDependenciesConventionPlugin : Plugin<Project> {
                     "implementation"(libs.navigation.compose)
                     "implementation"(libs.androidx.lifecycle.runtime.compose)
                     "implementation"(libs.core.remoteviews)
+                    "implementation"(libs.androidx.glance.appwidget)
+                    "implementation"(libs.androidx.glance.material)
+                    "implementation"(libs.androidx.glance.preview)
+                    "implementation"(libs.androidx.glance.appwidget.preview)
```
jpelgrom, line 92: Try bundling these as I don't see using just one happening anytime soon

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/Theme.kt
```diff
@@ -39,15 +49,94 @@ private val haDarkColors = darkColors(
...
         // Copied from MdcTheme:
         CompositionLocalProvider(
             LocalContentColor provides MaterialTheme.colors.onBackground,
-            content = content
+            content = content,
         )
     }
 }
+
+@Composable
```
jpelgrom, line 65: From this point: does it make sense to have this in the same file? You're creating a util/compose/glance package and you're not putting the theme in there.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/glance/GlanceExt.kt
```diff
@@ -0,0 +1,11 @@
+package io.homeassistant.companion.android.util.compose.glance
```
jpelgrom, line 1: Does this really need a package if it is the only thing? The name of the file is already pretty descriptive.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    private fun watchingForEntitiesChanges(context: Context) {
+        setupWidgetScope()
+        if (!serverManager.isRegistered()) {
+            Timber.tag(widgetClassName).d("No server registered won't watch for entities")
+            return
+        }
+
+        Timber.e("Watching for entities $widgetScope")
```
jpelgrom, line 182: Why is this an error? Did you leave in debug logging statements for development and if so do you plan to clean them up?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,289 @@
...
+                    Timber.tag(widgetClassName).d("Entity updates is null for widget $appWidgetId not watching")
+                    // Shouldn't do anything since the job should have been canceled already
+                    removeSubscription(appWidgetId)
+                }
+            }
+        }
+    }
+
+    private fun stopWatchingForEntitiesChanges() {
+        Timber.i("Stop watching for ${glanceAppWidget.javaClass}")
```
jpelgrom, line 215: If you're going to be fancy use the fancy variable ;)

```suggestion
        Timber.i("Stop watching for $widgetClassName")
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidget.kt
```diff
@@ -1,456 +1,344 @@
...
-            val itemsByStatus: Map<Boolean, List<GetTodosResponse.TodoItem>>,
-            val error: Boolean = false,
-        )
+import io.homeassistant.companion.android.util.compose.HomeAssistantGlanceTheme
+import io.homeassistant.companion.android.util.compose.HomeAssistantGlanceTypography
+import io.homeassistant.companion.android.util.compose.glance.glanceStringResource
+import io.homeassistant.companion.android.widgets.todo.TodoState.Companion.getColors
+
+/**
+ * Glance widget for managing and displaying a Todo List.
```
jpelgrom, line 53: I think we have different ideas of documenting code, but some of this feels like fluff / repeats the base Glance widget docs without adding much to it / should be in the developer docs instead of code probably (like: try to follow these guidelines).

common/src/main/res/values/strings.xml
```diff
@@ -1054,9 +1054,8 @@
     <string name="widget_todo_refresh">Refresh to-do</string>
     <string name="widget_todo_label">To-do list</string>
     <string name="widget_todo_empty">You have no to-do items!</string>
+    <string name="widget_todo_no_configuration">Oops. It looks like this widget is not configured.</string>
```
jpelgrom, line 1057: This string is more generic, so maybe it's a good idea to make the name more generic as well.

app/src/main/res/xml/todo_widget_info.xml
```diff
@@ -15,5 +15,4 @@
     android:description="@string/todo_widget_desc"
     android:widgetCategory="home_screen"
     android:previewImage="@drawable/widget_example_todo"
-    android:previewLayout="@layout/widget_todo_wrapper_dynamiccolor"
```
jpelgrom, line 18: Does Glance not support a preview layout? That would be a shame, it's nice to have the widget previews translated and adopt your theme's color.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,289 @@
...
+
+    @Inject
+    lateinit var dao: DAO
+
+    @Inject
+    lateinit var serverManager: ServerManager
+
+    private var widgetScope: CoroutineScope = widgetScopeProvider()
+
+    private val widgetJobs = mutableMapOf<Int, Job>()
```
kamilwawrzyczek, line 104: I believe this will not work as we do not have guarantees that system will reuse broadcast receiver instance.
So you can get screen on on instance A that will have those and screen off on instance B that will have nothing to stop.
On old implementation it as in companion object so it was shared.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,289 @@
...
+
+        // Handle the Glance [androidx.glance.action.Action]
+        super.onReceive(context, intent)
+
+        when (intent.action) {
+            Intent.ACTION_SCREEN_ON -> watchingForEntitiesChanges(context)
+            Intent.ACTION_SCREEN_OFF -> stopWatchingForEntitiesChanges()
+            AppWidgetManager.ACTION_APPWIDGET_UPDATE -> {
+                /*
+                 * WARNING: This action is received from another object thant the other above so you shouldn't entities, otherwise it
```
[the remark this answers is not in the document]
TimoPtr: Basically. This intent is exactly what you described about getting a different instance. We are getting a different instance each time this action is sent. So we can't subscribe here to the entities changes.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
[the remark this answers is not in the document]
TimoPtr: Debug leftover thanks for the catch

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidget.kt
```diff
@@ -1,456 +1,344 @@
...
-            setOnClickFillInIntent(
-                R.id.widget_todo_done_disabled,
-                Intent()
-                    .putExtra(EXTRA_TOGGLE_SUMMARY, summary)
-                    .putExtra(EXTRA_TOGGLE_TARGET_STATUS, GetTodosResponse.TodoItem.COMPLETED_STATUS),
-            )
+@Composable
+private fun ListContent(todoItems: List<TodoItemState>, displayComplete: Boolean) {
+    LazyColumn(modifier = GlanceModifier.padding(top = 4.dp)) {
+        item { HeaderItem(glanceStringResource(commonR.string.widget_todo_active)) }
```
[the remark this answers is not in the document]
TimoPtr: I didn't know if we should do it or not. If it was like this in your impl I will hide it 👍

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
TimoPtr: Since we actually create the receiver in the app when the process is created we are sure that the screen_on and screen_off intent are coming from the same instance. But not the action coming from the system based on the manifest like UPDATE. That's why we can't watch in UPDATE unfortunately.

If a new instance is created it means either we registered twice by mistake or that the process is starting so we should be fine.

app/src/main/res/xml/todo_widget_info.xml
TimoPtr: If I'm not wrong the only way to do it is to have a XML version and TBH I didn't want to do that.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidget.kt
TimoPtr: When writing it I was wondering if it was the place for this. So i'm going to move it. 
My initial idea was to find a way to make this todo widget a bit verbose on purpose to showcase and explain how things work. 
But I might have missed this goal.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
TimoPtr: Mistake, I did clean up but I had so many debugs logs that I might have missed some like this one.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/Theme.kt
TimoPtr: I made this change then went on long weekend and came back and forgot what I was doing 😂.

What is your opinion? Put everything in glance package? Or remove the glance package and move the extension somewhere else? I'm pretty sure we are going to create new extension.

app/src/main/AndroidManifest.xml
TimoPtr: I will try 👍

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,289 @@
...
+                 * won't ever be cancel.
+                 *
+                 * This will trigger an update of all the widget, but won't watch for changes.
+                 *
+                 * <p>This may be sent in response to a new instance for this AppWidget provider having
+                 * been instantiated, the requested {@link AppWidgetProviderInfo#updatePeriodMillis update interval}
+                 * having lapsed, or the system booting.
+                 */
+                widgetScope.launch {
```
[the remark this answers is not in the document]
TimoPtr: Even if it triggers when the screen is off the update all won't do anything since it triggers the composition only if the screen is on.

Basically this event is triggered when you install a new version of the app (useful in debug), probably at some other moments I didn't dig into the docs tbh for other events I needed this one for developing.

bgoncal (changes requested):
Small alignment issues, first can we make sure the leading spacing is the same for the widget title ("Shopping list"), subtitle ("Active completed") and the checkboxes?

Also the spacing between the checkbox and the text should be increased a bit, not sure whats the standard is in android but it needs to be more proportional to the leading spacing.
And last, can we remove the scrolling bar? Or at least make sure it does not overlap with the list content.

438279843-2f9401f0-413c-432f-8991-653df5996d56

app/src/main/AndroidManifest.xml
TimoPtr: Indeed it failed. I renamed the classes. But I would like your opinion to both of you with @jpelgrom. I named the Compose class `TodoGlanceAppWidget` and the receiver `TodoWidget`.

How should we name this classes for future widget. I would like to add it to the documentation.

app/src/main/res/xml/todo_widget_info.xml
TimoPtr: Actually I found this https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:glance/glance-appwidget-preview/src/main/java/androidx/glance/appwidget/preview/GlanceAppWidgetViewAdapter.kt while making the `bundle` I was wondering why there was 2 preview libraries.

I'm gonna try it out.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/Theme.kt
```diff
@@ -39,15 +49,94 @@ private val haDarkColors = darkColors(
...
+fun HomeAssistantGlanceTheme(
+    colors: ColorProviders = HomeAssistantGlanceTheme.colors,
+    content:
```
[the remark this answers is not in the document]
TimoPtr: It's the KTLint format that made this. But indeed it looks weird, changing it.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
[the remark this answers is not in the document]
TimoPtr: I rewrote this part.

app/src/main/AndroidManifest.xml
jpelgrom: > `TodoWidget` for Compose and `TodoWidgetReceiver` for the receiver.

For consistency and to avoid breaking changes for other widgets that would be my preference (and the names are still logical to me)

dshokouhi: agreed, that makes sense to me as well

dshokouhi:
One small comment, when items are marked as completed the text should have a ~~strikethrough~~ to match the frontend and existing design

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,289 @@
...
+        val invalidWidgetIds = entitiesPerServer.keys.minus(systemWidgetIds)
+
+        if (invalidWidgetIds.isNotEmpty()) {
+            Timber.tag(widgetClassName).i(
+                "Found widgets $invalidWidgetIds in database, but not in AppWidgetManager - sending onDeleted",
+            )
+            deleteWidgetsFromDatabase(invalidWidgetIds.toIntArray())
+        }
+
+        val entitiesPerServerWithoutInvalid = entitiesPerServer.toMutableMap().apply {
```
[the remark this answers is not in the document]
TimoPtr: Nice I didn't know this overload 👍🏻

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetConfigureViewModel.kt
```diff
@@ -84,29 +87,57 @@ class TodoWidgetViewModel @Inject constructor(
             selectedEntityId in entities.value.map { it.entityId }
     }
 
-    fun prepareData(componentName: ComponentName): Intent? {
+    fun prepareData() {
```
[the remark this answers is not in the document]
TimoPtr: I'm going to tackle this in the next PR.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetStateUpdater.kt
```diff
@@ -0,0 +1,121 @@
...
+
+        if (entityUpdateFlow == null) {
+            Timber.w("Integration return null for entity update the widget won't update")
+        }
+
+        return entityUpdateFlow?.let { merge(flowOf(currentEntity), it) }
+    }
+
+    private fun getInitialStateFlow(widgetId: Int): Flow<TodoState> {
+        return suspend { todoWidgetDao.get(widgetId) }.asFlow().map {
```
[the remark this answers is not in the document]
TimoPtr: I made it this way so that the function doesn't need to be suspendable which would extend to `stateFlow`.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetToggleActionsTest.kt
```diff
@@ -0,0 +1,124 @@
...
+
+        coVerify(exactly = 1) {
+            entryPoints.dao().get(widgetId)
+            entryPoints.serverManager().webSocketRepository(42)
+            entryPoints.webSocketRepository.updateTodo("HA", "42", null, COMPLETED_STATUS)
+        }
+    }
+
+    @Test
+    fun `toggleStatus mapping is valid`() {
```
dshokouhi, line 119: Looks like this line is missing the Given, When Then format :)

TimoPtr: Fixed

.github/workflows/pr.yml
```diff
@@ -182,7 +182,7 @@ jobs:
           mock-google-services: "true"
 
       - name: Run test
-        run: ./gradlew :common:testDebugUnitTest # Should be adapted once we have more than one tests
+        run: ./gradlew :common:testDebugUnitTest :app:testFullDebugUnitTest # Should be adapted once we have more than one tests
```
jpelgrom, line 185: The test is not full flavor specific. Is there a generic one or if not, can you add a note explaining why you put full here but not minimal?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,284 @@
...
+/**
+ * Base class for Glance widgets that update based on entity state changes.
+ *
+ * This class provides the foundational functionality for managing widget updates, handling lifecycle events,
+ * and observing entity state changes. It is designed to be extended by specific widget implementations.
+ *
+ * ### Key Features:
+ * - **Entity state observation**: Watches for updates to specific entities and triggers widget updates accordingly.
+ * - **Lifecycle management**: Handles widget lifecycle action [Intent.ACTION_SCREEN_ON], [Intent.ACTION_SCREEN_OFF] and [AppWidgetManager.ACTION_APPWIDGET_UPDATE].
+ * - **Real-time updates**: Supports real-time updates for widgets when the user grants the necessary permissions.
```
jpelgrom, line 43: What is "the necessary permission"? Also applies to lines 47-48. Is that documentation needed here or best placed elsewhere?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,284 @@
...
+ * - **Cleanup**: Cancel subscription when widget are removed and make sure the database reflect the current widgets used.
+ *
+ * ### Notes:
+ * - Glance widget, may "freeze" subscriptions after some time unless the user grants
+ *   the `real-time` permission. To unfreeze, call `update` on the widget.
+ * - Ensure that data sources are observed within the composition state (e.g., using [androidx.compose.runtime.collectAsState]) to keep
+ *   widgets updated while the composition is active (less than a minute even with the `real-time` permission).
+ * - Make sur to invoke [registerReceiver] to register actions to manage widget updates effectively.
```
jpelgrom, line 51:
```suggestion
 * - Make sure to invoke [registerReceiver] to register actions to manage widget updates effectively.
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,284 @@
...
+ * - Even if we are watching for entity changes in the receiver the data received cannot be sent to the widget directly,
+ *   if you need to send data to the widget you need to update the [DAO] and call [update].
+ * - The changes are watch only after [Intent.ACTION_SCREEN_ON]. It means that on startup the widget won't watch for changes
+ *   until the screen is turn on and off.
```
jpelgrom, line 55:
```suggestion
 * - The changes are watched only after [Intent.ACTION_SCREEN_ON]. It means that on startup the widget won't watch for changes
 *   until the screen is turned off and on again.
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,284 @@
...
+ *
+ * ### Usage:
+ * Extend this class and implement the abstract method `getWidgetEntitiesByServer` to provide the mapping of widget IDs
+ * to their associated entities.
+ *
+ * ### Example:
```
jpelgrom, line 61: Should this heading be level 4, or maybe removed as the next example doesn't have a heading?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,284 @@
...
+ * // In io.homeassistant.companion.android.HomeAssistantApplication
+ * override fun onCreate() {
+ *     super.onCreate()
+ *     ...
+ *     MyWidgetReceiver().registerReceiver(this)
+ * }
+ * ```
+ *
+ * ### Implementation Details:
+ * - **Coroutine Scope**: A shared [CoroutineScope] is used to manage asynchronous tasks for all widgets of a given class. It uses
```
jpelgrom, line 81: Use sentence case in documentation

```suggestion
 * ### Implementation details:
 * - **Coroutine scope**: A shared [CoroutineScope] is used to manage asynchronous tasks for all widgets of a given class. It uses
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,284 @@
...
+            Intent.ACTION_SCREEN_ON -> startWatchingForEntitiesChanges(context)
+            Intent.ACTION_SCREEN_OFF -> stopWatchingForEntitiesChanges()
+            /**
+             * This action will trigger an update for all widgets but will not monitor for changes.
+             *
+             * WARNING: This action is received from a different instance of this class than the others above, so you should not start something that
+             * needs to be canceled here like [startWatchingForEntitiesChanges] otherwise it won't ever be canceled.
+             *
+             * Extract from the action documentation:
+             * <p>This action may be sent in response to a new instance of this AppWidget provider being instantiated,
```
jpelgrom, line 126: The `<p>` tag here is not ended and seems misplaced.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,284 @@
...
+    private var widgetScope: CoroutineScope = widgetScopeProvider()
+
+    private val widgetJobs = mutableMapOf<Int, Job>()
+
+    // Helper to identify the widget in the logs
+    private val widgetClassName by lazy { glanceAppWidget.javaClass.name }
+
+    override fun onReceive(context: Context, intent: Intent) {
+        val appWidgetId = intent.getIntExtra(AppWidgetManager.EXTRA_APPWIDGET_ID, AppWidgetManager.INVALID_APPWIDGET_ID)
+        Timber.v("Received intent action = ${intent.action} for widget = $appWidgetId ($widgetClassName) $this")
```
jpelgrom, line 111: Do we need debug log statements at the start of most functions?

The documentation you wrote suggests not using excessive logging and it should add value. While it may be true for development once accepted most of these feel less relevant to me.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,284 @@
...
+        val invalidWidgetIds = entitiesPerServer.keys.minus(systemWidgetIds)
+
+        if (invalidWidgetIds.isNotEmpty()) {
+            Timber.tag(widgetClassName).i(
+                "Found widgets $invalidWidgetIds in database, but not in AppWidgetManager - removing it from the database.",
+            )
+            deleteWidgetsFromDatabase(invalidWidgetIds.toIntArray())
+        }
+
+        // strip the server and entities from the map since they are not used anymore
```
jpelgrom, line 243: This comment does not reflect the code below. You're not removing a server?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
```diff
@@ -0,0 +1,284 @@
...
+        val entitiesPerServerWithoutInvalid = entitiesPerServer - invalidWidgetIds
+
+        return entitiesPerServerWithoutInvalid
+    }
+
+    /**
+     * From https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:glance/glance-appwidget/src/main/java/androidx/glance/appwidget/GlanceAppWidget.kt;l=78;drc=852af4153b5fc3f0c92b8356f00ec5d786c5e869
```
jpelgrom, line 250: - Can you clean up the url? `;drc=852af4153b5fc3f0c92b8356f00ec5d786c5e869` isn't necessary.
- Can you permalink the url? `main` will change in the future.

jpelgrom (commented):
Do we

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetState.kt
```diff
@@ -0,0 +1,122 @@
...
+            todoItems.isNotEmpty()
+        } else {
+            todoItems.any { !it.done }
+        }
+    }
+
+    companion object {
+        /**
+         * Create a complete [TodoStateWithData] from the DB and from the server. Set the flag [outOfSync] to false, since the data
+         * should just come from the server.
```
jpelgrom, line 90:
```suggestion
         * includes an updated state from the server.
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetState.kt
```diff
@@ -0,0 +1,122 @@
...
+                listName = entity.friendlyName,
+                todoItems = todos.map(TodoItemState::from),
+                outOfSync = false,
+                showComplete = todoEntity.showCompleted,
+            )
+        }
+
+        /**
+         * Since we only have the DB entity it means that we are out of sync. So when using this function it will
+         * set the flag [outOfSync] to true.
```
jpelgrom, line 107: I think it's nicer to align the descriptions.

```suggestion
         * Create a [TodoStateWithData] with data only from the DB. Set the flag [outOfSync] to true, since the data
         * doesn't have an updated state from the server.
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetStateUpdater.kt
```diff
@@ -0,0 +1,126 @@
...
+import kotlinx.coroutines.flow.flatMapLatest
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.merge
+import kotlinx.coroutines.flow.onCompletion
+import kotlinx.coroutines.flow.onStart
+import timber.log.Timber
+
+/**
+ * This class allow the [io.homeassistant.companion.android.widgets.todo.TodoGlanceAppWidget] to get their
```
jpelgrom, line 25: Same package so I think you don't need to use the fully qualified name, but not 100% sure.
```suggestion
 * This class allow the [TodoGlanceAppWidget] to get their
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetStateUpdater.kt
```diff
@@ -0,0 +1,126 @@
...
+            } else {
+                TodoStateWithData.from(it)
+            }
+        }
+    }
+
+    /**
+     * Observes and provides the state of the widget identified by the given [widgetId].
+     *
+     * ### Flow Details:
```
jpelgrom, line 81: Use sentence case

```suggestion
     * ### Flow details:
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetStateUpdater.kt
```diff
@@ -0,0 +1,126 @@
...
+     * 1. **Initial state flow**: Emits the current state of the widget using the data in the database. If no configuration exists, it emits an empty state.
+     * 2. **Watch for changes flow**: Listens for changes in the widget's configuration or updates from the server. When a change is detected:
+     *    - It fetches the latest entity and todo items from the server.
+     *    - Updates the database with the latest data.
+     *    - Emits the updated state.
+     *
+     * ### Error Handling:
```
jpelgrom, line 88: Use sentence case

```suggestion
     * ### Error handling:
```

app/src/main/res/drawable-nodpi/widget_example_todo.png
jpelgrom, line 1: Is this screenshot accurate? The padding between list items looks wrong to me compared to the padding between a list item and title. (compare milk <> bread and bread <> completed)

app/src/main/res/drawable/ic_refresh_toolbar_glance.xml
jpelgrom, line 1: Is this icon Glance specific somehow? If no drop `_glance` from the file name to allow future re-use.

common/src/main/kotlin/io/homeassistant/companion/android/database/widget/TodoWidgetEntity.kt
```diff
@@ -37,4 +37,11 @@ data class TodoWidgetEntity(
         val summary: String?,
         val status: String?,
     )
+
+    fun isSameConfiguration(other: TodoWidgetEntity): Boolean {
+        // The only field that is not part of the configuration is latestUpdateData
+        // We make copy of the data classes without the latestUpdateData field so that we can use the equals method generated
+        // By doing this we make the check more future proof, if a new configuration field is added
```
jpelgrom, line 44: Should be rewritten to be a KDoc comment on the function instead

app/src/main/res/drawable/ic_refresh_toolbar_glance.xml
TimoPtr: Good catch I didn't remember that I did that ! 
The other icon that we have have a custom attributes for the color and it doesn't work well in glance. So this icon is a duplicate of the one without _glance. Happy to hear ideas on this attributes.

app/src/main/res/drawable-nodpi/widget_example_todo.png
TimoPtr: I took the screenshot out of the preview editor.

TimoPtr: I probably should have take it on a real device.

app/src/main/res/drawable/ic_refresh_toolbar_glance.xml
jpelgrom: Didn't check it but that makes sense. I think to follow the existing pattern this one should be named `ic_refresh` then, see `ic_filter`/`ic_question` for a similar example where a toolbar-specific version exists.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetStateUpdater.kt
TimoPtr: It probably happened when I refactor the names to fix the issue in the manifest.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetConfigureActivity.kt
```diff
@@ -125,29 +127,32 @@ class TodoWidgetConfigureActivity : BaseActivity() {
...
+                    ),
+                )
+        }
     }
 
     private fun onAddWidget() {
-        val intent = viewModel.prepareData(ComponentName(this, TodoWidget::class.java))
+        val intent = viewModel.prepareData()
         if (intent == null) {
+            // TODO show the error when it makes sense
```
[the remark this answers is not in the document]
TimoPtr: I want the refactor this in a second PR dedicated for the configuration screen.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetConfigureViewModel.kt
```diff
@@ -62,7 +63,9 @@ class TodoWidgetViewModel @Inject constructor(
     }
 
     private fun loadPreviousState(widgetId: Int) = viewModelScope.launch {
-        todoWidgetDao.get(widgetId)?.let {
+        val entity = todoWidgetDao.get(widgetId)
+
+        entity?.let {
```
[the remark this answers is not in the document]
TimoPtr: I think it's a leftover of some debugging that I did. Will revert.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetActions.kt
```diff
@@ -0,0 +1,150 @@
...
+@Composable
+internal fun actionRefreshTodo(): Action {
+    return actionRunCallback<RefreshAction>()
+}
+
+/**
+ * Get an Action that will open the [WebViewActivity] for the given [listEntityId]
+ */
+@Composable
+internal fun actionStartWebView(listEntityId: String, serverId: Int): Action {
```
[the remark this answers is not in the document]
TimoPtr: I actually debated with myself about this ! Same for refresh (but more complicated). I wanted to avoid having a path param. I don't like having strings as parameters that can lead to typos not catchable at build time.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoGlanceAppWidget.kt
```diff
@@ -0,0 +1,325 @@
...
+private fun HeaderItem(name: String) {
+    Text(text = name, style = HomeAssistantGlanceTypography.bodyLarge, modifier = GlanceModifier.padding(horizontal = 16.dp))
+}
+
+@OptIn(ExperimentalGlancePreviewApi::class)
+@Preview(250, 320)
+@Composable
+private fun ScreenPreview() {
+    HomeAssistantGlanceTheme {
+        Screen(
```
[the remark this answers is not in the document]
TimoPtr: Yep 👍 I introduced ScreenForState after the previews.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
TimoPtr: I need to check how to do it from cs.android 👍

app/src/main/res/drawable-nodpi/widget_example_todo.png
jpelgrom: Please do to check. The screenshots in the PR descriptions look a better but it's not easy to compare.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
TimoPtr: @dshokouhi I don't know if this has a permission name do you know?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetConfigureActivity.kt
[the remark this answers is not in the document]
jpelgrom: Refactor: OK. But for now this comment still applies? Not suggesting a major rework, just removing a few lines so the code makes sense until you make more changes there.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetActions.kt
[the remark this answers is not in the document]
jpelgrom: > I don't like having strings as parameters that can lead to typos not catchable at build time.

Agreed. What about something like:
```kotlin
fun actionViewTodoList(listEntityId: String, serverId: Int): Action =
    actionStartWebView("$WEBVIEW_PATH_TODO$listEntityId", serverId)
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
jpelgrom: Click the arrow next to the branch name to switch to the latest commit, or View in > View in Gitiles
image

.github/workflows/pr.yml
TimoPtr: Currently we have 
> Candidates are: 'testDebugUnitTest', 'testFullDebugUnitTest', 'testFullReleaseUnitTest', 'testMinimalDebugUnitTest', 'testMinimalReleaseUnitTest', 'testReleaseUnitTest'.

Full and minimal are the same. I decided to not execute it for automotive since they are the same that's why I'm not using the top level calls.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
TimoPtr: I didn't find anything on the doc speaking about this page 
image
it's anoying that this page control widget refresh state... I think it's kinda linked to https://developer.android.com/develop/ui/views/notifications/conversations#conversation-widgets since they need to stay up to date.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
TimoPtr: I kept this under verbose to see if we properly receive events for a given Intent. I think this one should be kept. I will cleanup the rest, but here being on a baseClass I think it is nice to be a bit verbose for people using the class in future development.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
TimoPtr: Done, but I'm curious do you what is DRC ?

app/src/main/res/drawable-nodpi/widget_example_todo.png
TimoPtr: I think it's because of Github rendering it does look ok on my Pixel 6

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
dshokouhi: this is hte intent: https://developer.android.com/reference/android/provider/Settings#ACTION_NOTIFICATION_LISTENER_SETTINGS

and this is hte permission: https://developer.android.com/reference/android/Manifest.permission#BIND_NOTIFICATION_LISTENER_SERVICE

jpelgrom: That is not a widget-related permission though, it makes the system run a service for the app all the time. Presumably you'd see the same behavior for the widget if you enabled the websocket foreground service?

It may be that recent Android changes now mean that you need to have some part of the app active to keep the widget's coroutinescope (and thus subscription) working, whereas previously it worked without, but I would not point to notification access here. That is just misleading IMO.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
jpelgrom: Sorry, no idea what DRC is. Probably some tracking/UI state that isn't relevant to us.

app/src/main/res/drawable-nodpi/widget_example_todo.png
jpelgrom: With your updated screenshots in the PR description now that also looks wrong to me, so I guess the comment is more about the widget design than the preview image.

To illustrate my point:
image

The difference between title and items is much smaller for the Google Keep widget you used as an example
image

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
TimoPtr: I just tested and even with the foreground service up and running. If you don't gives this "real time" permission the widget doesn't refresh after it exits the composition state... 

I know this permission is for notification, but I think Google is using it for the conversation widget to stay in sync all the time. It's only an assumption but it kinda sounds ok with how they've documented it.

This is kind of a hack of the API to keep a widget up to date. We should document that in the companion docs. I wrote it here for future contributors to understand that by default it won't stay up to date.

The reality is that the default behavior is that the widget will enter composition every 30mn for 1mn not more on new version of Android, even if an event occurs it won't do anything.

app/src/main/res/drawable-nodpi/widget_example_todo.png
TimoPtr: image
Should be good now it was missing a vertical padding of 8dp. It does add a bit of space bellow the toolbar but I think it actually looks better.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseGlanceEntityWidgetReceiver.kt
TimoPtr: https://github.com/home-assistant/companion.home-assistant/pull/1203 it's a proposal.

app/src/main/res/drawable-nodpi/widget_example_todo.png
TimoPtr: should be fixed, I've also updated the screenshot in the PR desc.

Issue #3174, opened 2022-12-24
Support UnifiedPush


hockeymikey:
**Is your feature request related to a problem? Please describe.**
[UnifiedPush](https://unifiedpush.org/) is an open source push notification protocol that is implemented by a few apps such as NextPush or ntfysh. After installing one of these implementations (called distributors), you can select them for using them for push notifications in apps such as Element or other [apps](https://unifiedpush.org/users/apps/).
**Describe the solution you'd like**
I'd like to be able to use UnifiedPush for my notifications in the app. This would allow users to have push notifications but not rely on FCM and Google. 

**Describe alternatives you've considered, if any**
There is none.

**Additional context**
I think the links above to their website is a good starting point for understanding the protocol. For implementation on the app side, I think going into one of the apps that support unifiedpush like Element would be a good place to draw inspiration. Here is their [issue ticket for element ](https://github.com/vector-im/element-android/issues/2743)to see their discussion and PRs for the implementation.

dshokouhi:
This needs to first be supported in HA core before the app can consider it. I think we should hold off on this request until then.

dshokouhi:
Let's close this until core has its own request and it gets implemented until then the app can't support this.

hockeymikey:
@dshokouhi But why close it? It's good to track rather than having to duplicate it.

hockeymikey:
https://community.home-assistant.io/t/add-unifiedpush-support-to-home-assistant-android-core/446344

julianfoad:
S1m of the UnifiedPush developers, in [UnifiedPush matrix room](https://matrix.to/#/#unifiedpush:matrix.org), says (paraphrased):

A server supporting webpush is enough to support UnifiedPush. It looks like the HA core supports webpush, so the issue can be reopened :)

There is example code in the [UP-Example](https://github.com/UnifiedPush/android-example) app, for how to get UnifiedPush notifications with webpush.

There are currently open discussions to update the U.P. spec to make it clearly compatible with webpush. ([SERV_1.2.0 #37](https://github.com/UnifiedPush/specifications/pull/37), [Retransmission #34](https://github.com/UnifiedPush/specifications/issues/34)).

Terminology:
  UnifiedPush | Webpush
  ---|---
  End-user app | user agent
  push server | push service
  endpoint | push resources

dshokouhi:
[...]

PR #5261, opened 2025-05-01
Add support for UnifiedPush notifications

lone-faerie:
## Summary

Adds support for using [UnifiedPush](https://unifiedpush.org/) as the source for push notifications. This allows both the full and minimal flavors to receive push notifications without a persistent websocket connection and without reliance on Google's Firebase Cloud Messaging. On launch, UnifiedPush is preferred if a distributor is available, otherwise the full flavor falls back to FCM. A setting has also been added to allow the user to manually select the UnifiedPush distributor to use.

## Checklist

## Screenshots

UnifiedPush Setting
Distributor Selection

## Any other notes

https://github.com/home-assistant/android/issues/3174 mentions Home Assistant Core needing to add support before this would work. I'm not sure if it's changed since then, but in my tests UnifiedPush works with Core as-is. A small change to Core would be needed to support encrypting the notifications, but that isn't necessary for UnifiedPush to work.


jpelgrom:
While this works to get started (and is quite similar in some ways to my [proof of concept from last year](https://github.com/jpelgrom/home-assistant-android/tree/unifiedpush)), I feel like there should be a generic interface here. This approach doesn't scale and the push provider can change without the user noticing.

lone-faerie:
The push provider should only change during onboarding or if the user uninstalls the distributor app. A generic interface does feel like the more proper approach though, perhaps something like the SensorManager. UnifiedPush also has the ability to handle FCM on it's own, though that would require a bit of backend work since it needs a gateway similar to the current cloud functions.

TimoPtr:
Since you are changing the notification, you will also need two others PR in the developer documentation.

Rudd-O:
@lone-faerie I'm willing to fund you finishing this work, and doing whatever the Companion / Core devs want done to get this in.

Please quote me and let's get the details straightened out.

attila-lendvai:
> I'm willing to fund you finishing this work

@Rudd-O i'm also willing to chip in. maybe we could set up some bounty somewhere and advertise it?

is anyone here experienced with an opensource bounty solution that could work here?

PR #5273, opened 2025-05-05
Todo widget configure improvements

TimoPtr:
## Summary

After the migration to Glance in https://github.com/home-assistant/android/pull/5252. The error were not properly handled anymore. The default theme was DAYNIGHT but should have been DYNAMICCOLOR from the start.

## Checklist

## Any other notes

I would like feedback on how we should handle in Compose error from the viewModel back in compose.

Small improvements on how we handle `domains`, by creating a `IntegrationDomains` object that should holds the strings of the `domains` to avoid duplicates.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetConfigureActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetConfigureActivity.kt
@@ -145,11 +145,14 @@ class TodoWidgetConfigureActivity : BaseActivity() {
     }
 
     private fun onAddWidget() {
-        viewModel.prepareData()
-        // TODO show the error when it makes sense
-        setResult(RESULT_OK)
-        viewModel.updateWidget(this)
-        finish()
+        try {
+            viewModel.addWidgetConfiguration()
+            setResult(RESULT_OK)
+            viewModel.updateWidget(this)
+            finish()
+        } catch (_: IllegalStateException) {
+            showAddWidgetError()
+        }
     }
 
     private fun showAddWidgetError() {
--- a/app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetConfigureViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetConfigureViewModel.kt
@@ -89,15 +89,14 @@ class TodoWidgetConfigureViewModel @Inject constructor(
             selectedEntityId in entities.value.map { it.entityId }
     }
 
-    fun prepareData() {
+    fun addWidgetConfiguration() {
         if (!isValidSelection()) {
-            // TODO properly react to this in the activity
             Timber.d("Widget data is invalid")
-            return
+            throw IllegalArgumentException("Widget data is invalid")
         }
         if (widgetId == AppWidgetManager.INVALID_APPWIDGET_ID) {
             Timber.w("Widget ID is invalid")
-            return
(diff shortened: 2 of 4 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetConfigureActivity.kt
```diff
@@ -145,11 +145,14 @@ class TodoWidgetConfigureActivity : BaseActivity() {
...
-        viewModel.updateWidget(this)
-        finish()
+        try {
+            viewModel.addWidgetConfiguration()
+            setResult(RESULT_OK)
+            viewModel.updateWidget(this)
+            finish()
+        } catch (_: IllegalStateException) {
+            showAddWidgetError()
+        }
```
TimoPtr, line 155: I don't really like using exception to manage the error. If you have any other suggestion I'm all hears.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/IntegrationDomains.kt
```diff
@@ -0,0 +1,5 @@
+package io.homeassistant.companion.android.common.data.integration
+
+object IntegrationDomains {
+    const val TODO_DOMAIN = "todo"
+}
```
TimoPtr, line 5: I had something like that in mind at first
```kotlin
private const val TODO_DOMAIN = "todo"


sealed interface IntegrationDomain {
    val name: String
    object Todo: IntegrationDomain {
        override val name: String
            get() = TODO_DOMAIN
    }
    // ...
    data class Unknown(override val name: String): IntegrationDomain
}

val <T> Entity<T>.domain: IntegrationDomain
    get() {
        val domainName = this.entityId.split(".")[0]
        return when (domainName) {
            TODO_DOMAIN -> IntegrationDomain.Todo
            //...
            else -> IntegrationDomain.Unknown(domainName)
        }
    }


suspend fun demoFilterByDomain(serverManager: ServerManager, serverId: Int) {
    serverManager.integrationRepository(serverId)
        .getEntities()
        .orEmpty()
        .filter { entity -> entity.domain is IntegrationDomain.Todo }
}
```

The idea here was to not allow string comparison outside of the creation of the domain in the mapper. But it is a huge refactor and I might miss something.

This simpler proposal is here to remove a duplicate, let's discuss about the more complex one.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetConfigureViewModel.kt
```diff
@@ -46,10 +49,11 @@ class TodoWidgetConfigureViewModel @Inject constructor(
...
                 .getEntities()
                 .orEmpty()
-                .filter { entity -> entity.domain == "todo" }
-        }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(500), emptyList())
+                .filter { entity -> entity.domain == TODO_DOMAIN }
+        }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(500.milliseconds), emptyList())
+
     var selectedEntityId by mutableStateOf<String?>(null)
-    var selectedBackgroundType by mutableStateOf(WidgetBackgroundType.DAYNIGHT)
+    var selectedBackgroundType by mutableStateOf(WidgetBackgroundType.DYNAMICCOLOR)
```
jpelgrom, line 56: Dynamic color is not available on all SDK versions, shouldn't there be a check for that?

jpelgrom (approved):

PR #5281, opened 2025-05-07
Update changelog

TimoPtr:
## Summary

Add missing entry in changlog for beta release.

### DIFF
--- a/app/src/main/res/xml/changelog_master.xml
+++ b/app/src/main/res/xml/changelog_master.xml
@@ -1,15 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <changelog xmlns:tools="http://schemas.android.com/tools"
     tools:ignore="MissingDefaultResource">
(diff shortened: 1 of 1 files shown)


app/src/main/res/xml/changelog_master.xml
```diff
@@ -3,6 +3,7 @@
     tools:ignore="MissingDefaultResource">
     <release version="2025.5.1 - Main" versioncode="3">
         <change>Add more Health Connect sensors</change>
+        <change>New design of To-do list widget</change>
         <change>Bug fixes and dependency updates</change>
```
TimoPtr, line 7: Do we include webSocket stability and crash fixed in this change?

jpelgrom: We generally don't write about specific fixes

jpelgrom:
The version should also be changed

jpelgrom (approved):

PR #5301, opened 2025-05-11
Introduce PushProvider interface for push notification providers

lone-faerie:
## Summary

This PR introduces a generic interface for push notification providers. Currently, the only provider is FCM, but an interface allows supporting additional push providers in the future, such as UnifiedPush.

The implementation is based on the `SensorManager` as only one instance of each provider is needed but they should still be able to be enabled for only a subset of servers. A `PushManager` class has also been added to coordinate the various push providers, which is currently just hard coded to use FCM but would provide the primary API for managing push providers.

## Checklist

## Any other notes

This came as a suggestion in [this PR](https://github.com/home-assistant/android/pull/5261#issuecomment-2846417492) for adding UnifiedPush as an alternative push provider.


app/src/full/kotlin/io/homeassistant/companion/android/notifications/FirebaseCloudMessagingProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+    companion object {
+        const val SOURCE = "FCM"
+    }
+
+    override val setting = PushProviderSetting.FCM
+
+    override fun isAvailable(context: Context): Boolean = true
+
+    private var token: String? = null
+    private val tokenMutex = Mutex()
```
TimoPtr, line 26: Why do you need a mutex? Add documentation explaining why is this needed.

app/src/full/kotlin/io/homeassistant/companion/android/notifications/FirebaseCloudMessagingProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    override suspend fun getToken(): String {
+        return tokenMutex.withLock { token } ?: try {
+            FirebaseMessaging.getInstance().token.await()
+        } catch (e: Exception) {
+            Timber.e(e, "Issue getting token")
+            ""
+        }
+    }
```
TimoPtr, line 39: You are not storing the new token in this class?

app/src/full/kotlin/io/homeassistant/companion/android/notifications/FirebaseCloudMessagingProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.google.firebase.messaging.FirebaseMessaging
+import io.homeassistant.companion.android.common.notifications.PushProvider
+import io.homeassistant.companion.android.database.settings.PushProviderSetting
+import javax.inject.Inject
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.tasks.await
+import timber.log.Timber
+
+class FirebaseCloudMessagingProvider @Inject constructor(
```
TimoPtr, line 13: The class is named MessagingProvider but at first glance it simply store a token in memory and forward a message to the messagingManager with the right SOURCE.

app/src/main/kotlin/io/homeassistant/companion/android/notifications/PushModule.kt
```diff
@@ -0,0 +1,20 @@
...
+import io.homeassistant.companion.android.common.notifications.PushProvider
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+abstract class PushModule {
+    @Binds
+    @Singleton
+    @IntoMap
+    @StringKey(FirebaseCloudMessagingProvider.SOURCE)
```
TimoPtr, line 18:
```suggestion
    @ClassKey(FirebaseCloudMessagingProvider::class)
```
to avoid exposing a string.

app/src/main/kotlin/io/homeassistant/companion/android/notifications/PushManager.kt
```diff
@@ -0,0 +1,21 @@
+package io.homeassistant.companion.android.notifications
+
+import io.homeassistant.companion.android.common.notifications.PushProvider
+import javax.inject.Inject
+
+class PushManager @Inject constructor(
+    val providers: Map<String, @JvmSuppressWildcards PushProvider>
+) {
```
TimoPtr, line 8: Could you create an instrumentation test that checks how the map looks like. It will be useful in the future to make sure the map contains the proper values.

app/src/full/kotlin/io/homeassistant/companion/android/notifications/FirebaseCloudMessagingProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+    private val messagingManager: MessagingManager
+) : PushProvider {
+
+    companion object {
+        const val SOURCE = "FCM"
+    }
+
+    override val setting = PushProviderSetting.FCM
```
TimoPtr, line 21: It's confusing what is setting here?

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
```diff
@@ -10,7 +11,8 @@ import kotlinx.coroutines.launch
 
 abstract class LaunchPresenterBase(
     private val view: LaunchView,
-    internal val serverManager: ServerManager
+    internal val serverManager: ServerManager,
+    internal val pushManager: PushManager
```
TimoPtr, line 15: I think the visibility of the attributes of this class needs to be lower to `protected` not `internal`.

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
```diff
@@ -10,7 +11,8 @@ import kotlinx.coroutines.launch
...
 ) : LaunchPresenter {
 
     internal val mainScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
```
TimoPtr, line 18: If you can check this ones too.

app/src/main/kotlin/io/homeassistant/companion/android/notifications/PushManager.kt
```diff
@@ -0,0 +1,21 @@
...
+        private const val FCM_SOURCE = FirebaseCloudMessagingProvider.SOURCE
+    }
+
+    val defaultProvider: PushProvider get() {
+        return providers[FCM_SOURCE]!!
+    }
+
+    suspend fun getToken(): String {
+        return providers[FCM_SOURCE]!!.getToken()
+    }
```
TimoPtr, line 20: Could you already prepare in this PR the fact that FCM is only available in full by keeping this class agnostic from FCM?

app/src/main/kotlin/io/homeassistant/companion/android/notifications/PushModule.kt
```diff
@@ -0,0 +1,20 @@
...
+    abstract fun bindFirebasePushProvider(firebaseCloudMessagingProvider: FirebaseCloudMessagingProvider): PushProvider
```
TimoPtr, line 19: Why this is not in full?

app/src/minimal/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterImpl.kt
```diff
@@ -3,22 +3,24 @@ package io.homeassistant.companion.android.launch
...
 import kotlinx.coroutines.launch
 import timber.log.Timber
 
 class LaunchPresenterImpl @Inject constructor(
     view: LaunchView,
-    serverManager: ServerManager
-) : LaunchPresenterBase(view, serverManager) {
+    serverManager: ServerManager,
+    pushManager: PushManager
+) : LaunchPresenterBase(view, serverManager, pushManager) {
```
TimoPtr, line 15: Now that you have an abstraction you could potentially already remove the two version of the LaunchPresenterImpl and keep only the one from full. Let's avoid as much as possible duplicates. It makes maintenance of minimal easier.

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/PushProvider.kt
```diff
@@ -0,0 +1,91 @@
...
+import io.homeassistant.companion.android.common.BuildConfig
+import io.homeassistant.companion.android.common.data.integration.DeviceRegistration
+import io.homeassistant.companion.android.common.data.servers.ServerManager
+import io.homeassistant.companion.android.database.AppDatabase
+import io.homeassistant.companion.android.database.settings.PushProviderSetting
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import timber.log.Timber
+
+interface PushProvider {
```
TimoPtr, line 17: Let's make sure this public interface is properly documented. Also it would be nice to create a section in https://developers.home-assistant.io/docs/android/ about how to add a new provider if this is meant to evolve.

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/PushProvider.kt
```diff
@@ -0,0 +1,91 @@
...
+    val setting: PushProviderSetting
+
+    /** @return `true` if this provider is available to the app on this device */
+    fun isAvailable(context: Context): Boolean
+
+    /** @return `true` if this provider is enabled on any server */
+    fun isEnabled(context: Context): Boolean {
+        val settingsDao = AppDatabase.getInstance(context).settingsDao()
+        return serverManager(context).defaultServers.any {
+            settingsDao.get(it.id)?.pushProvider == setting
```
TimoPtr, line 27: Maybe we could consider using a sealed interface instead of an enum, so we don't need to store within the class what the class is already.

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/PushProvider.kt
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /** @return `true` if this provider is enabled for the specified server */
+    fun isEnabled(context: Context, serverId: Int): Boolean {
+        val settingsDao = AppDatabase.getInstance(context).settingsDao()
+        return settingsDao.get(serverId)?.pushProvider == setting
+    }
+
+    /** @return Set of server IDs for which this provider is enabled */
+    fun getEnabledServers(context: Context): Set<Int> {
```
TimoPtr, line 38: Why not returning the Servers directly? It avoid this class to know the internals of the Server class

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/PushProvider.kt
```diff
@@ -0,0 +1,91 @@
...
+                    try {
+                        serverManager.integrationRepository(it).updateRegistration(
+                            deviceRegistration = DeviceRegistration(
+                                pushToken = getToken()
+                            ),
+                            allowReregistration = false
+                        )
+                    } catch (e: Exception) {
+                        Timber.e(e, "Issue updating token")
+                    }
```
TimoPtr, line 73: Could you narrow done the catch to only the potential exception that can occur?

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/PushProvider.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    suspend fun getToken(): String
+
+    fun onMessage(context: Context, notificationData: Map<String, String>)
+
+    /**
+     * Update the device registration using this push provider
+     */
+    suspend fun updateRegistration(context: Context, coroutineScope: CoroutineScope) {
+        coroutineScope.launch {
```
TimoPtr, line 56: Why do you need to get the scope here? Either you enforce the scope within the interface so you control it or you delegate to the caller function to decide which scope to use.

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/PushProvider.kt
TimoPtr, line 17: Could you add unit tests on the default implementation thanks.

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/PushProvider.kt
```diff
@@ -0,0 +1,91 @@
...
+            val serverManager = serverManager(context)
+            if (serverManager.isRegistered()) {
+                Timber.d("Not trying to update registration since we aren't authenticated.")
+                return@launch
```
TimoPtr, line 60:
```suggestion
            if (serverManager.isRegistered()) {
                Timber.d("No server registered skipping update registration.")
                return@launch
```

common/src/main/kotlin/io/homeassistant/companion/android/database/settings/Setting.kt
```diff
@@ -12,5 +12,7 @@ data class Setting(
     @ColumnInfo(name = "websocket_setting")
     var websocketSetting: WebsocketSetting,
     @ColumnInfo(name = "sensor_update_frequency")
-    var sensorUpdateFrequency: SensorUpdateFrequencySetting
+    var sensorUpdateFrequency: SensorUpdateFrequencySetting,
+    @ColumnInfo(name = "push_provider", defaultValue = "FCM")
+    var pushProvider: PushProviderSetting
```
TimoPtr, line 17:
```suggestion
    var pushProvider: PushProviderSetting,
```
It will make future update of this class easier to review

common/src/main/kotlin/io/homeassistant/companion/android/database/settings/Setting.kt
```diff
@@ -12,5 +12,7 @@ data class Setting(
     @ColumnInfo(name = "websocket_setting")
     var websocketSetting: WebsocketSetting,
     @ColumnInfo(name = "sensor_update_frequency")
-    var sensorUpdateFrequency: SensorUpdateFrequencySetting
+    var sensorUpdateFrequency: SensorUpdateFrequencySetting,
+    @ColumnInfo(name = "push_provider", defaultValue = "FCM")
```
TimoPtr, line 16: Could we avoid using a string here for the defaultValue, it's a source of typo and will make refactor harder if one day we want to refactor.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingViewModel.kt
```diff
@@ -42,4 +48,11 @@ class SettingViewModel @Inject constructor(
             settingsDao.update(it)
         }
     }
+
+    fun updatePushProviderSetting(id: Int, setting: PushProviderSetting) {
```
TimoPtr, line 52:
```suggestion
    fun updatePushProviderSetting(serverId: Int, setting: PushProviderSetting) {
```

TimoPtr (commented):
Thanks for pushing forward onto the suggestion made in your previous PR.

app/src/full/kotlin/io/homeassistant/companion/android/notifications/FirebaseCloudMessagingProvider.kt
lone-faerie: I named it that based on FirebaseCloudMessaging and PushProvider, but yeah it is a bit misleading. Something like FirebasePushProviderImpl would probably be clearer.

app/src/full/kotlin/io/homeassistant/companion/android/notifications/FirebaseCloudMessagingProvider.kt
lone-faerie: It is used in `isEnabled()` for checking against the database. That way each implementation just has to define `setting` instead of implementing its own `isEnabled()`. I'll add documentation to explain that, or there could be a better way of handling it.

app/src/full/kotlin/io/homeassistant/companion/android/notifications/FirebaseCloudMessagingProvider.kt
lone-faerie: A mutex might not be necessary, as I'm not 100% positive on the flow of Android services and coroutine scopes. To my understanding, there is potential for a race condition if `onNewToken()` in FirebaseCloudMessagingService gets called at the same time as `getToken()`, such as at launch or during onboarding.

app/src/full/kotlin/io/homeassistant/companion/android/notifications/FirebaseCloudMessagingProvider.kt
lone-faerie: This function was primarily to move `getMessagingToken()` from onboarding into the push provider. The token gets stored with the device registration, and I believe the Firebase library also stores the token.

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/PushProvider.kt
lone-faerie: I'm not sure that'd work, since the implementations are in a different module than the interface. Though that could be solved by moving PushProvider from :common to :app

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/PushProvider.kt
lone-faerie: This function signature was adapted from the one in SensorManager. The only place it's currently used is in `updateRegistration()` which only needs the server ID.

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/PushProvider.kt
lone-faerie: This could be me not fully understanding how Kotlin handles coroutines, but does taking the coroutine scope as an argument not delegate it to the caller function?
Or do you just mean the initial `coroutineScope.launch`? Yeah, that probably isn't needed. I just took `onNewToken()` from FirebaseCloudMessagingService and adapted it for the interface.

common/src/main/kotlin/io/homeassistant/companion/android/database/settings/Setting.kt
lone-faerie: Unfortunately, defaultValue must be a compile-time constant string.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingViewModel.kt
lone-faerie: Should the rest of the methods in SettingViewModel also be changed to match this?

app/src/main/kotlin/io/homeassistant/companion/android/notifications/PushModule.kt
lone-faerie: Any more push providers that are added would also need to be bound in the module, which may not be limited to full. I'm not fully familiar with hilt/dagger, but I imagine it would work to put bindFirebasePushProvider in its own module in full and any providers that work in both full and minimal could be put in one in main.

app/src/test/kotlin/io/homeassistant/companion/android/notifications/PushManagerTest.kt
lone-faerie, line 1: I'm not sure if this is the best way to test the structure of the providers map since it relies on PushManager being in LaunchActivity. I think a better way would be to follow Android's [Hilt testing guide](https://developer.android.com/training/dependency-injection/hilt-testing) but it requires adding the `hilt-android-testing` dependency.

TimoPtr: You can add the new dependency if you need it. The test that I would like to have is a test that gets the map injected by hilt where you can assert the content of the map.

app/src/main/kotlin/io/homeassistant/companion/android/notifications/PushModule.kt
TimoPtr: You can Inject into the map from multiple places. Here we should have a PushModule in full for FCM and another one in minimal without FCM. We don't want FCM reference in minimal.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingViewModel.kt
TimoPtr: You can yes

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/PushProvider.kt
TimoPtr: Let's remove the scope given as parameter.

common/src/main/kotlin/io/homeassistant/companion/android/database/settings/Setting.kt
TimoPtr: Ok the main issue is that the constant is in `:app` so you don't have access it here. I don't like having a reference to FCM in common because it will also end up into `minimal` and it doesn't make sense to have FCM in minimal as default.

I would encourage you to handle the case where the push provider is not set in the `:app` so each flavor can have a proper default value. Also it creates a proper separation from app and common.

TimoPtr:
Multiple comments aren't addressed. Please take a look to them before asking for re-review. Also I would like @jpelgrom opinion on the architecture since he was the one initially thinking about introducing this concept. So take a look to my comments and ask us to review again your changes :)

app/src/minimal/kotlin/io/homeassistant/companion/android/notifications/FirebaseCloudMessagingProvider.kt
```diff
@@ -0,0 +1,33 @@
+package io.homeassistant.companion.android.notifications
+
+import android.content.Context
+import io.homeassistant.companion.android.common.notifications.PushProvider
+import io.homeassistant.companion.android.database.settings.PushProviderSetting
+import javax.inject.Inject
+import kotlinx.coroutines.CoroutineScope
+
+class FirebaseCloudMessagingProvider @Inject constructor() : PushProvider {
```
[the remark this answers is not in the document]
lone-faerie: Actually, I don't think a No-op provider is even necessary. The only thing it would be needed for is returning an empty token, but that can be avoided by just adding null checks elsewhere.

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/PushProvider.kt
lone-faerie: If we wanted to use this same architecture for Wear, then PushProvider would either have to be in :common or duplicated in :wear

lone-faerie: Another option could be implementing an id() function like with SensorManager:
```kotlin
fun PushProvider.id(): String {
    val simpleName = this::class.simpleName ?: this::class.java.name
    return simpleName.lowercase(Locale.US).replace(" ", "_")
}
```
This way the class doesn't need to store what it is, it's just derived from the class itself.

jpelgrom:
How do you imagine UnifiedPush would work with this architecture for choosing between multiple distributors? The current architecture forces each push provider to have a separate class, while my first expectation would be that each distributor could be considered a push provider (for example, you get a list: none / FCM / ntfy / Sunup / ...).

lone-faerie:
The reason for each push provider having its own class is to support multiple protocols, as well as helping to separate things between minimal and full. Like @TimoPtr said, it's a good idea to keep any references to FCM out of minimal. And all of the UnifiedPush distributors (ntfy, Sunup, etc.) would be covered under one class since they all use the same protocol.

The `PushProvider` interface is primarily for translating the various protocols into a common interface. For example, Firebase messages are received as a map but UnifiedPush messages are received as raw bytes, and Firebase just provides a token while UnifiedPush provides a url and public key. The `PushManager` class would then be responsible for coordinating the various push providers. It will provide the list of available providers/distributors as well as handling enabling the selected one.

As for the UI aspect of choosing a distributor, I imagined a setting similar to what I had in [my previous PR](https://github.com/home-assistant/android/pull/5261) though more generic to include other push providers apart from just UnifiedPush. This architecture would also open up the possibility of using a different provider for each server.

lone-faerie:
Ok, I get what you're saying. I was planning on initially just using the default UnifiedPush distributor with `
UnifiedPush.tryUseCurrentOrDefaultDistributor(context)` just to get things working at first. But there's a couple ways I could see for the user choosing a distributor. We could add a `getDistributors()` method to PushProvider or the UnifiedPush library includes [its own dialog](https://unifiedpush.org/kdoc/connector_ui/) for choosing a distributor that we could use.

jpelgrom:
> We could add a `getDistributors()` method to PushProvider

That could be an option. What would you return for that function in the FCM push provider?

> or the UnifiedPush library includes [its own dialog](https://unifiedpush.org/kdoc/connector_ui/) for choosing a distributor

That would require specific UI work outside the provider so kind of defeats the point, I think?

lone-faerie:
> That could be an option. What would you return for that function in the FCM push provider?

For the FCM provider, I'd either return `null` or "FCM" (or "Firebase"). If we allow null, then we could display it to the user as 'Provider (Distributor)' if not null, otherwise just 'Provider'. So for example, 'FCM' and 'UnifiedPush (ntfy)'. If `getDistributor()` always returns a list, so FCM returns "FCM", then we'd just display to the user each distributor.

> That would require specific UI work outside the provider so kind of defeats the point, I think?

For UnifiedPush, the library handles the UI work, though for other providers, yeah it would require work outside of it. The library's UI probably doesn't align with Home Assistant's, so it's probably not the right choice, just wanted to mention it.

jpelgrom:
@lone-faerie Are you still interested in working on this? In case it wasn't obvious with the 👍 reaction to your latest comment, your suggestion sounded OK to me. If you have any questions, feel free to ask. I'll change this to draft for now, whenever you're ready you can mark it as ready for review.

lone-faerie:
Yeah, sorry I've been moving so haven't had any time to work on this. I'm finally settled in so I'll start up on this again soon

app/src/test/kotlin/io/homeassistant/companion/android/notifications/PushManagerTest.kt
lone-faerie: I've updated the instrumentation tests to use `hilt-android-testing` but it requires changing the `testInstrumentationRunner` but this breaks other instrumentation tests that aren't testing injection. There's a few ways to solve this, but I'm not sure which would be best. The first would be annotating test classes for hilt like so:
```kotlin
@HiltAndroidTest
class LaunchActivityTest {
    @get:Rule
    val hiltRule = HiltAndroidRule(this)

    // The rest of the existing LaunchActivityTest
}
```
The other two options would be using Robolectric or adding a separate gradle config for hilt tests.

common/src/main/kotlin/io/homeassistant/companion/android/database/AppDatabase.kt
```diff
@@ -103,7 +104,7 @@ import timber.log.Timber
         Server::class,
         Setting::class
     ],
-    version = 50,
+    version = 51,
```
TimoPtr, line 106: You also need to update the json file of the database.

app/src/test/kotlin/io/homeassistant/companion/android/notifications/PushManagerTest.kt
TimoPtr: Thanks for looking into this. I've pushed to your branch another version that use your suggestion about Robolectric.

We should always prefer Robolectric over Instrumentation Test since it is more reliable and way faster.

PR #5310, opened 2025-05-14
Remove value strings of preference arrays from Lokalise

### LINKED ISSUE #5292: Switching between light/dark theme is not working

**Android version(s):**
Android 15

**Device model(s):**
Pixel 6a

**Home Assistant version:**
2025.05.1

Switching between dark and light in companion app settings does not work if app is set to German.
It does however work, when language is set to English.
This issue old issue, that was fixed in https://github.com/home-assistant/android/issues/913#issue-697515698
And recently reappeared

**Companion App Logs:**

```

```

Image
Image

**Additional information:**

TimoPtr:
## Summary

The switch for the theme was not working because it was using values from `strings.xml` in one side (from the xml arrays) and harcoded values in ThemeManager. Because of that when the translation of the values where different than the hardcoded value it was always defaulting to `AppCompatDelegate.MODE_NIGHT_NO`.

I decided to rename the strings of the values used in the arrays to avoid any issues, when we are going to download the translation for the next release.

⚠️ I need to check that after this PR is merged we are not pushing the new keys into Localize again otherwise the issue will happen again later when someone will translate it again.

I decided to also apply this to the screen orientation that was built the same way. Even if the issue is not as obvious, the issue is only visible if you change the orientation and then you change your language.

## Checklist

## Any other notes

Fixes #5292 
If you test the apk in this PR you cannot reproduce the issue since our repo only contains one language.
⚠️ People that previously had screen orientation set up in another language will get this settings reset if there was a translation.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -900,8 +900,8 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
         WebView.setWebContentsDebuggingEnabled(BuildConfig.DEBUG || presenter.isWebViewDebugEnabled())
 
         requestedOrientation = when (presenter.getScreenOrientation()) {
-            getString(commonR.string.screen_orientation_option_value_portrait) -> ActivityInfo.SCREEN_ORIENTATION_PORTRAIT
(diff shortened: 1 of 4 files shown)


dshokouhi (approved):
Thanks for the quick fix and comment as a reminder to not translate them again :)

TimoPtr:
I did verify the Lokalise configuration and it only pulls the file `common/src/main/res/values/strings.xml`. So anything outside of this won't end up into it 👍🏻 merging this PR.

PR #5322, opened 2025-05-19
Add notification command for turning the hotspot on/off

moemoeq:
## Summary

Added hotspot control functionality through Notification Commands in Home Assistant Android app. This feature allows users to remotely toggle the hotspot function from Home Assistant.

- Uses TetheringManager to control hotspot
   - Supported only on Android 11 (API 30) and above
- Enables hotspot activation/deactivation through notifications

Example to Trigger:
Turn on hotspot:
```
automation:
  - alias: Turn on hotspot
    trigger:
      ...
    action:
      - action: notify.mobile_app_<your_device_id_here>
        data:
          message: "command_hotspot"
          data:
            command: "turn_on"
```

Turn off hotspot:
```
automation:
  - alias: Turn on hotspot
    trigger:
      ...
    action:
      - action: notify.mobile_app_<your_device_id_here>
        data:
          message: "command_hotspot"
          data:
            command: "turn_off"
```

## Checklist

I'll update the documentation PR after getting feedback on this one.

## Any other notes

This feature only works on Android 11 (API 30) and above. The hotspot control functionality is implemented using TetheringManager, and **reflection** is used to access system APIs.

To wider device compatibility, I think also implement a method using ConnectivityManager, although I don't have any device to test this. (> android 11)
However, I've included code with scalability in mind, considering this implementation (even if it seems unnecessary now).

I've verified it on two of my Samsung devices G977N(API 31), G906N(API 34).


app/src/main/kotlin/io/homeassistant/companion/android/util/HotspotHelper.kt
```diff
@@ -0,0 +1,214 @@
...
+        try {
+            return disableHotspot()
+        } catch (e: Exception) {
+            Timber.e(e, "Failed to turn off hotspot")
+            return false
+        }
+    }
+
+    // API 30+ needed for TetheringManager
+    @RequiresApi(Build.VERSION_CODES.R)
```
TimoPtr, line 70: You could probably make the whole class RequiresAPI 30 no?

app/src/main/kotlin/io/homeassistant/companion/android/util/HotspotHelper.kt
```diff
@@ -0,0 +1,214 @@
...
+    private fun enableHotspot(): Boolean {
+        // Try TetheringManager
+        // Support least Android 11(API 30) Code.R
+        if (enableHotspotWithTetheringManager()) {
+            Timber.i("Hotspot enabled successfully using $METHOD_TETHERING_MANAGER")
+            return true
+        }
+
+        // TODO: implement using ConnectivityManager as fallback
```
TimoPtr, line 79: For now I would keep this feature for API 30 with this API and create an issue for previous API (you can link the issue in the TODO). With this you can track how many people would be interested into having this feature in old version, before jumping into using the old API.

app/src/main/kotlin/io/homeassistant/companion/android/util/HotspotHelper.kt
```diff
@@ -0,0 +1,214 @@
...
+                        val builderClass = Class.forName("android.net.TetheringManager\$TetheringRequest\$Builder")
+                        val builder = builderClass.getConstructor(Int::class.java).newInstance(TETHERING_WIFI)
+                        val request = builderClass.getMethod("build").invoke(builder)
+
+                        val startMethod = tetheringManager.javaClass.getMethod(
+                            "startTethering",
+                            tetheringRequestClass,
+                            Executor::class.java,
+                            callbackClass
+                        )
```
TimoPtr, line 126: You should not use reflection here.

```
            val tetheringManager = context.getSystemService<TetheringManager>()
            if (tetheringManager == null) {
                Timber.e("TetheringManager is null")
                return false
            }
            tetheringManager.startTethering(TetheringManager.TetheringRequest.Builder(TetheringManager.TETHERING_WIFI).build(),
                mainExecutor, object : TetheringManager.StartTetheringCallback {})
```

That's how you get the `TetheringManger` you need to adjust the class to not use reflection. 

https://developer.android.com/reference/android/net/TetheringManager the API seems to have been added very recently API 36.

app/src/main/kotlin/io/homeassistant/companion/android/util/HotspotHelper.kt
moemoeq: That sounds like a great idea. If that's the case, I think it's possible make the whole class mark RequiresAPI 30 in this PR

app/src/main/kotlin/io/homeassistant/companion/android/util/HotspotHelper.kt
moemoeq: Thanks for taking a look! I opened this PR mainly because I wanted to get feedback on this specific point. 

During development, I just knows official `TetheringManager` public API (without reflection) is now available starting from API 36 (Android 16), which is still in beta. Because of that, I implemented the current version (using reflection) to support API 30 and above, since that’s where most devices are in use today.

I’ve noticed that similar approaches are used in other apps and some posts, so I followed that pattern. but, my understanding of the long-term implications is still somewhat limited.

One idea I had was to branch the logic by API level — using the public API for API 36+ and falling back to reflection for earlier versions. I’d love to hear your thoughts on whether that would be a more appropriate or acceptable approach.

Also, more generally: is using reflection to access hidden APIs something this project prefers to avoid? Or is it conditionally acceptable depending on use cases like this? If there are any specific guidelines I should follow, I’d really appreciate being pointed in the right direction.

Thanks again for your input!

TimoPtr: Out of curiosity did you try to use the official API on lower version? It might work since it works with reflection. If you look at the impl of `TetheringManager` they doesn't specify any API version so I'm curious.

I get why other app used reflection to access this API if it was not publicly available. But since it has been released and our compile version is 36 it might work. Again not sure at all it needs testing.

moemoeq: Good point! Lucky for me, I’ve got devices on API 31 and 34 in my hand and Android Studio’s already open 😄 I’ll run some tests using with official API on lower versions and let you know how it goes.

And true, I’m with you reflection and private APIs are exactly not ideal. I needed the functionality and, honestly, didn’t find a solid guideline on whether this was “acceptable,” so I went with a solution that seemed to work across more devices for now.

 I’ll report back soon!

moemoeq: I tested it using the official TetheringManager API, and interestingly, it worked fine on both of my devices (API 31 and 34).

That was unexpected, but definitely a nice surprise. maybe we can drop the reflection.
Here’s the snippet I used for testing, just in case:
```
    private fun enableHotspotWithTetheringManager(): Boolean {
        try {
            val tetheringManager = context.getSystemService<TetheringManager>()
            if (tetheringManager == null) {
                Timber.e("TetheringManager is null")
                return false
            }

            tetheringManager.startTethering(TetheringManager.TetheringRequest.Builder(TetheringManager.TETHERING_WIFI).build(),
                mainExecutor,
                object : TetheringManager.StartTetheringCallback {
                    override fun onTetheringStarted() {
                        Timber.d("Callback: Tethering started successfully")
                    }

                    override fun onTetheringFailed(result: Int) {
                        Timber.e("Callback: Tethering failed to start Result: $result")
                    }
                }
            )
```

moemoeq: I assumed it’s been functional since API 30(since reflection access worked back then). but, I’m not very experienced with Android Devs  curious to hear your take!

TimoPtr: The API was there before but not public. They've made it public now without specifying what is the minimal version from an annotation. If you check the source [here](https://cs.android.com/android/platform/superproject/main/+/512046e84bcc51cc241bc6599f83ab345e93ab12:packages/modules/Connectivity/Tethering/common/TetheringLib/src/android/net/TetheringManager.java) you can see the Copyright from 2019.

Most of the functions are marked `@SystemApi`
>  Indicates an API is exposed for use by bundled system applications.
These APIs are not guaranteed to remain consistent release-to-release,
and are not for use by apps linking against the Android SDK.

and the rest with `@SuppressLint("UnflaggedApi")`.

moemoeq: thanks for the clarification, After reading your comment and digging into the AOSP source myself
[android-29](https://android.googlesource.com/platform/prebuilts/fullsdk/sources/android-29/+/refs/heads/androidx-master-release/android/net/
)
[android-30](https://android.googlesource.com/platform/prebuilts/fullsdk/sources/android-30/+/refs/heads/androidx-main-release/android/net/TetheringManager.java)

From these, now i'm clear TetheringManager has existed since API 30 (just marked as @SystemApi back then). With that in mind, I’ll drop the reflection and implement the hotspot functionality directly using the TetheringManager API for API 30+

moemoeq: I have some sad news to share. During our attempt to switch to the official public TetheringManager API, StopTetheringCallback results in a "class not found" error. even StartTetheringCallback works properly from API 30(on my API 34 device)

so, i think we have two options:

1. Implement only the disable hotspot functionality using reflection while keeping the enable functionality with the direct API call
2. Implement both enable and disable functionality consistently with reflection

Which approach would be more maintainable and reliable in the long term? I'm concerned about potential inconsistencies with a mixed approach, but also hesitant to revert working direct API implementations unnecessarily.

andyearnshaw: Given @TimoPtr's aversion to reflection, perhaps the mixed approach would be best? In terms of being "maintainable and reliable":

1. Enable hotspot would continue to work if the reflection implementation was broken by a future Android update
2. Having the direct API approach would provide a reference for someone to update the hotspot disable code in the future when the direct API option is more widely available.

Alternatively, another option could be to have the disabling code not work in older Android versions. Half-support might be good enough for most people. Personally, I only need send the command to enable hotspot, since I want to send it to my car every time it starts up.

TimoPtr: We are on the path to target API 36. I would like someone to test again the StopTetheringCallback there. Indeed I would be in favor to only enable the feature where it is officially supported, meaning it might be for now only available to a subset of devices with recent versions.

TimoPtr: Now that we are targeting API 36 @moemoeq do you plan to continue this PR?

moemoeq: Thanks for checking in @TimoPtr, and yes i think it’s a good time to work on. I’ve actually been using a custom build and it has worked well for my use case. Since there seems to be interest from others as well, I’d love to share this experience to others,  I’ll take another look over the weekend and continue this pr

Soukyuu: Sorry for semi-offtopic, but will this also work with Android 16? I've been using macrodroid but its implementation stopped working after the update to 16 and the dev said that it's because google killed the API for that, so no other solution available. But this being worked on here gives me a little hope.

PR #5324, opened 2025-05-19
Fix always show first view when opening the app

### LINKED ISSUE #5297: 'Always show first view when opening app' setting is not working on HA core 2025.5
**Home Assistant Android app version(s):** 2025.4.3-full

**Android version(s):** 15

**Device model(s):** Samsung S25

**Home Assistant version:** core 2025.5.1 ; supervisor 2025.04.1 ; oprationg system 15.2

**Last working Home Assistant release (if known):** 2025.4.4

**Description of problem, include YAML if issue is related to notifications:** Previously,  with the option checked in the app, the next launch would start on the first page. Now, it starts on the last page;

**Companion App Logs:**

[homeassistant_companion_log_4-11-2025_11-48-2.txt](https://github.com/user-attachments/files/20148260/homeassistant_companion_log_4-11-2025_11-48-2.txt)

```

```

Image

**Additional information:**

TimoPtr:
## Summary

Fixes #5297, by using the new `navigate` external message introduced in https://github.com/home-assistant/frontend/pull/25516

I kept the old implementation but it only works with old version of the frontend. It means that this feature won't work if you have a version of core that is between the two implementation.

## Checklist

## Any other notes

To test it you need a frontend dev environement or wait for the next release of core.
Ideally the message sent to the external bus should be strongly typed so I decided to create a dedicated class for this new message.
The current implementation doesn't erase the backstack, but only replace the head of the backstack. It leads to a weird behavior in the navigation depending on where you were before putting the app in background. To avoid this we could have a command to the frontend to clear the backstack.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -104,6 +104,7 @@ import io.homeassistant.companion.android.util.isStarted
 import io.homeassistant.companion.android.websocket.WebsocketManager
 import io.homeassistant.companion.android.webview.WebView.ErrorType
 import io.homeassistant.companion.android.webview.externalbus.ExternalBusMessage
+import io.homeassistant.companion.android.webview.externalbus.NavigateTo
 import java.util.concurrent.Executors
 import javax.inject.Inject
 import javax.inject.Named
@@ -1662,14 +1663,19 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
         if (presenter.isAlwaysShowFirstViewOnAppStartEnabled() &&
             LifecycleHandler.isAppInBackground()
         ) {
-            // Pattern matches urls which are NOT allowed to show the first view after app is started
-            // This is
-            // /config/* as these are the settings of HA but NOT /config/dashboard. This is just the overview of the HA settings
-            // /hassio/* as these are the addons section of HA settings.
-            if (webView.url?.matches(".*://.*/(config/(?!\\bdashboard\\b)|hassio)/*.*".toRegex()) == false) {
-                Timber.d("Show first view of default dashboard.")
-                webView.evaluateJavascript(
-                    """
+            if (serverManager.getServer(presenter.getActiveServer())?.version?.isAtLeast(2025, 6, 0) == true) {
+                sendExternalBusMessage(
+                    NavigateTo("/config", true)
+                )
+            } else {
+                // Pattern matches urls which are NOT allowed to show the first view after app is started
+                // This is
+                // /config/* as these are the settings of HA but NOT /config/dashboard. This is just the overview of the HA settings
+                // /hassio/* as these are the addons section of HA settings.
+                if (webView.url?.matches(".*://.*/(config/(?!\\bdashboard\\b)|hassio)/*.*".toRegex()) == false) {
+                    Timber.d("Show first view of default dashboard.")
+                    webView.evaluateJavascript(
+                        """
                     var anchor = 'a:nth-child(1)';
                     var defaultPanel = window.localStorage.getItem('defaultPanel')?.replaceAll('"',"");
                     if(defaultPanel) anchor = 'a[href="/' + defaultPanel + '"]';
@@ -1678,10 +1684,11 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
                                                                    .shadowRoot.querySelector('paper-listbox > ' + anchor).click();
                     window.scrollTo(0, 0);
                     """,
-                    null
-                )
-            } else {
-                Timber.d("User is in the Home Assistant config. Will not show first view of the default dashboard.")
+                        null
+                    )
+                } else {
+                    Timber.d("User is in the Home Assistant config. Will not show first view of the default dashboard.")
+                }
             }
         }
     }
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -1663,20 +1663,17 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
         if (presenter.isAlwaysShowFirstViewOnAppStartEnabled() &&
             LifecycleHandler.isAppInBackground()
         ) {
-            /**
-             * Pattern matches urls which are NOT allowed to show the first view after app is started
-             * This is
-             * /config/ as these are the settings of HA but NOT /config/dashboard. This is just the overview of the HA settings
-             * /hassio/ as these are the addons section of HA settings.
-             */
+            // Clearing history and replace the current page with the default page from the frontend.
+            // This way the user have a clear history stack.
+            webView.clearHistory()
+
+            // Pattern matches urls which are NOT allowed to show the first view after app is started
+            // This is
+            // /config/* as these are the settings of HA but NOT /config/dashboard. This is just the overview of the HA settings
+            // /hassio/* as these are the addons section of HA settings.
             if (webView.url?.matches(".*://.*/(config/(?!\\bdashboard\\b)|hassio)/*.*".toRegex()) == false) {
                 Timber.d("Show first view of default dashboard.")
                 if (serverManager.getServer(presenter.getActiveServer())?.version?.isAtLeast(2025, 6, 0) == true) {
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1662,14 +1663,19 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
         ) {
-            // Pattern matches urls which are NOT allowed to show the first view after app is started
-            // This is
-            // /config/* as these are the settings of HA but NOT /config/dashboard. This is just the overview of the HA settings
-            // /hassio/* as these are the addons section of HA settings.
-            if (webView.url?.matches(".*://.*/(config/(?!\\bdashboard\\b)|hassio)/*.*".toRegex()) == false) {
-                Timber.d("Show first view of default dashboard.")
-                webView.evaluateJavascript(
-                    """
+            if (serverManager.getServer(presenter.getActiveServer())?.version?.isAtLeast(2025, 6, 0) == true) {
```
TimoPtr, line 1666: To test this PR this line needs to be adjusted.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1662,14 +1663,19 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
+                sendExternalBusMessage(
+                    NavigateTo("/config", true)
```
TimoPtr, line 1668: Replace is only replacing the head of the history. We could combine this with something like 
```
webView.goBackOrForward(0)
```
Then invoke `NavigateTo` to clear the back history.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: What do you mean? Assuming you're not mixing frontend and core versions, both will have a 2025.6.0devxxxxxxxx version which matches.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: Looking at the [reference](https://developer.android.com/reference/android/webkit/WebView#goBackOrForward(int)), the function you suggests does nothing as the history isn't changed? I'm confused.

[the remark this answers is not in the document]
TimoPtr: 🤦🏻 I commit my test. It should be `/`

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1662,14 +1663,19 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
+                )
+            } else {
+                // Pattern matches urls which are NOT allowed to show the first view after app is started
+                // This is
+                // /config/* as these are the settings of HA but NOT /config/dashboard. This is just the overview of the HA settings
+                // /hassio/* as these are the addons section of HA settings.
```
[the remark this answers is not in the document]
TimoPtr: What we do now is to ask the frontend to open the default view of the current user so, we delegated the responsibility to the user to set whatever he wants as default on the frontend. I don't think we need this check anymore.

BTW like I said in the description this code that we have is not currently working since multiple version of core already.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: I mean that if you want to test this PR you need to change this line to 2025.5.0 and make sure your local version contains my PR on the frontend.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: I didn't look at the impl. I though it was 0 for the first item in the history. Then we could use the size of the history maybe. It was just a suggestion that we could take a look. 

I imagine that someone that never close the app nor restart the device might have an history that can grow pretty big if he navigates within the app a lot. 

We should not tackle this here IMO, we could open an issue to see if people complains about it.

TimoPtr: But I just saw `webView.clearHistory()` we could use this instead it is meant for this.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: So you're mixing frontend and core versions.

Resolving as it is no longer relevant to the review.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: > we could open an issue to see if people complains about it

Because of how long this feature has existed and the lack of comments about it, it looks to me like most people that use this don't consider it a problem/complaint. An issue to track it wouldn't add a lot of value in my opinion.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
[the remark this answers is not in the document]
jpelgrom: The code/comment I highlighted is the pattern matching to avoid showing the first view if the user is on any settings screen. That still seems relevant to me. It was was previously added to avoid breaking flows which switch between the webview and native code and when switching to another app to check things/copy data.

> BTW like I said in the description this code that we have is not currently working since multiple version of core already.

The pattern matching still works as that doesn't depend on the frontend and paths haven't changed. The part that is broken is navigating to a different page.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1662,14 +1663,26 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
         if (presenter.isAlwaysShowFirstViewOnAppStartEnabled() &&
             LifecycleHandler.isAppInBackground()
         ) {
-            // Pattern matches urls which are NOT allowed to show the first view after app is started
-            // This is
-            // /config/* as these are the settings of HA but NOT /config/dashboard. This is just the overview of the HA settings
-            // /hassio/* as these are the addons section of HA settings.
+            /**
```
jpelgrom, line 1666: Why change this to KDoc style? The main advantages (info in tooltips, formatting, generated docs) don't apply anyway inside code blocks, and it is inconsistent with other code.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1662,14 +1663,26 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
             if (webView.url?.matches(".*://.*/(config/(?!\\bdashboard\\b)|hassio)/*.*".toRegex()) == false) {
                 Timber.d("Show first view of default dashboard.")
-                webView.evaluateJavascript(
-                    """
+                if (serverManager.getServer(presenter.getActiveServer())?.version?.isAtLeast(2025, 6, 0) == true) {
+                    /**
+                     * Clearing history and replace the current page with the default page from the frontend.
+                     * This way the user have a clear history stack.
+                     */
+                    webView.clearHistory()
```
jpelgrom, line 1679: If you're doing this for the new implementation, I'd also change the behavior for the old implementation so it behaves the same.

jpelgrom:
Any particular reason this is still in draft? Frontend PR was merged yesterday and your code seems complete.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: Will rollback

TimoPtr:
I didn't want to make it ready for review if it was not testable. But since it's already not working we could still merge it.

PR #5345, opened 2025-05-21
Enable edge to edge for setting screens

TimoPtr:
## Summary

In order to tackle part of #5328, this PR enable support for edge to edge in all the settings screen.

## Checklist

- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

## Screenshots

image
image

## Any other notes

In order to test quicker I've modified the `DevPlaygound` to be launchable from Android Studio directly.

PR #5346, opened 2025-05-22
Feature/edge to edge webviews

TimoPtr:
## Summary

Unfortunately I did not manage to properly handle the change of server since I have to enable edge to edge at the beginning of the activity. Furthermore API 36 enable edge to edge by default so we have to handle it.
See https://github.com/home-assistant/android/pull/5010 for more details. 

## Checklist

## Screenshots

Left iOS, Right: Android
image

## Any other notes

I did not manage to force push to the branch in https://github.com/home-assistant/android/pull/5010 after rebasing the branch so I decided to create another branch.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingActivity.kt
@@ -4,6 +4,7 @@ import android.os.Build
 import android.os.Bundle
 import android.view.KeyEvent
 import androidx.activity.OnBackPressedCallback
+import androidx.activity.enableEdgeToEdge
 import androidx.activity.viewModels
 import androidx.core.app.NotificationManagerCompat
 import androidx.fragment.app.commit
@@ -26,6 +27,7 @@ class OnboardingActivity : BaseActivity() {
     private val viewModel by viewModels<OnboardingViewModel>()
 
     override fun onCreate(savedInstanceState: Bundle?) {
+        enableEdgeToEdge()
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_onboarding)
 
--- a/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingViews.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingViews.kt
@@ -5,12 +5,15 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.ColumnScope
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.safeDrawing
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.windowInsetsPadding
 import androidx.compose.material.MaterialTheme
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
@@ -39,6 +42,7 @@ fun OnboardingScreen(
         val screenWidth = screenWidth()
         Column(
             modifier = modifier
+                .windowInsetsPadding(WindowInsets.safeDrawing)
                 .padding(all = 16.dp)
                 .width(min(screenWidth.value, STEP_SCREEN_MAX_WIDTH_DP).dp)
                 .align(Alignment.Center),
--- a/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/authentication/AuthenticationFragment.kt
@@ -27,6 +27,7 @@ import io.homeassistant.companion.android.common.R as commonR
 import io.homeassistant.companion.android.common.data.HomeAssistantApis
 import io.homeassistant.companion.android.common.data.authentication.impl.AuthenticationService
 import io.homeassistant.companion.android.common.data.keychain.KeyChainRepository
+import io.homeassistant.companion.android.common.util.applyInsets
 import io.homeassistant.companion.android.onboarding.OnboardingViewModel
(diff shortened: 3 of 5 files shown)


TimoPtr:
@Gregman-js could you check here?
We are also currently making sure the frontend supports for edge to edge is extended to the rest of it https://github.com/home-assistant/frontend/pull/25566

Gregman-js (approved):

Gregman-js (changes requested):
Sorry, not this PR https://github.com/home-assistant/android/pull/5345#pullrequestreview-2862526030

jpelgrom:
I think this PR should be changed to draft while the frontend is not yet ready to avoid merging it too early (in case the frontend changes don't make it in by 2025.6).

TimoPtr:
> > (...) since I have to enable edge to edge at the beginning of the activity [and on] API 36 (...) edge to edge [is enabled] by default so we have to handle it.
> 
> And the solution chosen to handle the insets on API 36 generally seems fine to me - adding 'normal'/empty space padding in case of a frontend that does not support it. Why not do the check again after the active server changed? Meaning we're always edge to edge and change the padding in our view as needed.

I initially wanted to do something by giving a lambda to the `applyInsets` function but then it becomes a bit more complicated with managing the nav bar and status bar coloring. We would need to also apply and remove coloring depending on the version of the server. I found it a bit too much for the value.
If you think it is important we can do it.

Gregman-js:
> I initially wanted to do something by giving a lambda to the applyInsets function but then it becomes a bit more complicated with managing the nav bar and status bar coloring. We would need to also apply and remove coloring depending on the version of the server. I found it a bit too much for the value.

`window.statusBarColor` and `window.navigationBarColor` is already deprecated. Maybe it will be easier to replace it with top and bottom Spacer/Box. Then we can hide it to "enable edge to edge" for newer servers. And we can apply background color to this boxes without problem.

jpelgrom:
> `window.statusBarColor` and `window.navigationBarColor` is already deprecated. Maybe it will be easier to replace it with top and bottom Spacer/Box. Then we can hide it to "enable edge to edge" for newer servers. And we can apply background color to this boxes without problem.

That could be nice, but may not be as easy on older API versions because of a scrim the system may try to force. Deprecated doesn't mean we should stopping it on older API versions if it works correctly and the behavior isn't the same on all API versions.

jpelgrom:
You don't have to migrate away if we can't make it work on older API versions. `if (Build.VERSION.SDK_INT => Build.VERSION_CODES.BAKLAVA) { /* or what else the minimum is we can make work with edge to edge, new API */ } else { /* old API }`

TimoPtr:
Could you make your PR to target my branch on the official repo? If your solution works I'm ok with it but we need to make it generic for the 3 places we have a webview
- AuthenticationFragment
- MyActivity
- WebViewActivity

Does that make sense to keep the call to the status bar color if it's deprecated if you provided an alternative solution.

Please remove the modification about the trailing `,` (I might make a big PR that adds all the `,` someday)

Gregman-js:
Okay, I will target your PR.
I kept status bar color because I wasn't able to test HA on older Androids (emulator, login page doesn't show) in context to @jpelgrom comment:
> That could be nice, but may not be as easy on older API versions because of a scrim the system may try to force.

TimoPtr:
I can help if you want an old version of an emulator I did struggle with it too and I've found a workaround and posted it on stackoverflow https://stackoverflow.com/a/79514205/3289338 it helps you find a proper APK of the latest available webview for an old emulator.

Gregman-js:
I created PR here https://github.com/home-assistant/android/pull/5390 for use
I will try to test PR with your webview workaround

timmo001:
Most of the safe area changes in the frontend are now merged to dev, we should be good to test on android now

ref: https://github.com/home-assistant/frontend/issues/26818

TimoPtr:
@timmo001 I've updated the code of this PR so that if it connects to a server on 2025.10 it is going to send the safe areas.

@jpelgrom I think I've managed to found a way to handle the server switch. But to be fair it's a bit messy to find the right way of doing this into this big class.

I sent to @timmo001 remaining issues, but it is already usable if you want to test.

TimoPtr:
@jpelgrom I did test the new version of the frontend 2025.12 and I feel more confident about enabling this. Would you mind double checking?

jpelgrom:
> @jpelgrom I did test the new version of the frontend 2025.12 and I feel more confident about enabling this. Would you mind double checking?

After testing with core 2025.12 and an older version, on a modern Android device, this is looking very promising! (Even if the frontend defaults to a safe area slightly too often IMO.) Just one minor frontend issue:



Do you have a fast route to report those?

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/webview/WebViewContentScreenScreenshotTest.kt
```diff
@@ -27,6 +27,28 @@ class WebViewContentScreenScreenshotTest {
...
             onFullscreenClicked = {},
             onNotificationPermissionResult = {},
+            serverHandleInsets = false,
+        )
+    }
+
+    @PreviewTest
+    @Preview
+    @Composable
+    fun `WebView with app unlocked server handle insets`() {
```
jpelgrom: Visually we cannot see insets, so if something changes the screenshot won't actually change making this test a bit redundant in my opinion.

TimoPtr:
> > @jpelgrom I did test the new version of the frontend 2025.12 and I feel more confident about enabling this. Would you mind double checking?
> 
> After testing with core 2025.12 and an older version, on a modern Android device, this is looking very promising! (Even if the frontend defaults to a safe area slightly too often IMO.) Just one minor frontend issue:
> 
> 
> 
> Do you have a fast route to report those?
> 
> 

I already reported on the task https://github.com/home-assistant/frontend/issues/26818#issuecomment-3665181991

I think most of the major issues has been fixed, and we should release this so that people can open issue on screens that we didn't test.

jpelgrom:
> I already reported on the task https://github.com/home-assistant/frontend/issues/26818#issuecomment-3665181991

Same issue, different dialogs 👍. I was even subscribed to that issue...

> I think most of the major issues has been fixed, and we should release this so that people can open issue on screens that we didn't test.

It'd be good if you rebase and we can get more feedback. Here/using forums at first, but I'm also not opposed to merging this right after we have a stable release and keeping it around for a couple of betas with the risk of minor breaking changes.

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/webview/WebViewContentScreenScreenshotTest.kt
TimoPtr: Removing

Issue #5023, opened 2025-01-30
Allow the app to be the launcher


bs42:
I use a wall mounted tablet for Home Assistant.  It would be very useful if the companion app could be the launcher.  This would permit the app to automatically load on boot as well as help prevent it from getting automatically closed which happens to mine every few days.

dshokouhi:
> This would permit the app to automatically load on boot

already requested #4030 



> help prevent it from getting automatically closed which happens to mine every few days.

this is a symptom of the app crashing or not having background access a bug report instaed of feature request should be filed

What actual launcher features do you want? Pretty sure webview cannot be a launcher element so not sure what the end goal of the request is.

bs42:
I previously used [https://wallpanel.xyz/](url) for my tablet, it has not been updated in ages and so I figured I would try the companion app on this new wall tablet.  With that app it would stay running for weeks at a time without fail.

My only goal is to not have to manually open the app on the tablet every few days.

This tablet does not have a battery and none of the background or battery features are available.  It doesn't say anything about crashing, I just walk past the tablet and it's sitting on the launcher window and the companion app is gone.  I relaunch it and repeat every few days.

If there's something I can do to get some logs from it that explain why to file a bug report I'll do so.

dshokouhi:
> It doesn't say anything about crashing, I just walk past the tablet and it's sitting on the launcher window and the companion app is gone

thats the exact symptom of the app crashing actually, you dont always get a pop-up

Next time that happens when you open the app go to Settings > companion app > troubleshooting > show and share logs and you may see a Recent Crash tab at the top and if that is there you will see the crash and the reason for it. The app becoming a launcher wont fix that crash. You may need to try for a few days if the crash or 2 does not get caught but hopefully the app can save the crash.

bs42:
There wasn't a tab there from the crash earlier, I'll check again next time it happens.

Thank you

bs42:
It has closed down 2 more times, neither time had a Recent Crash tab.

marazmarci:
Maybe the Play Store auto update is closing it?

fpetrovski:
@bs42 there are a few options to achieve something similar 
One is obviously picking HA as the launcher app, the downside of that is the moment you want to access something else on the tablet it becomes a but annoying.

[...]

PR #5348, opened 2025-05-22
Add option to set HA as launcher

fpetrovski:
## Summary

This PR is related to #5023 where it was discussed to allow HA Companion app to become a launcher.
A new option was added to the preference settings, under "Other Settings" Preference Category.

## Checklist


app/src/main/res/xml/preferences.xml
```diff
@@ -37,6 +37,11 @@
     </PreferenceCategory>
     <PreferenceCategory
         android:title="@string/other_settings">
+        <Preference
```
jpelgrom, line 40: Does this need this amount of promotion (in settings! the first item!) for a feature 99% of users will not use?

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -58,6 +58,8 @@ import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import timber.log.Timber
 
+private const val UNKNOWN_LAUNCHER_LABEL = "unknown app"
```
jpelgrom, line 61: Should be translated

jpelgrom (commented):
I'm personally still not convinced this feature needs to be added to the _companion_ app.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
fpetrovski: Absolutely, thanks for pointing it out.

app/src/main/res/xml/preferences.xml
fpetrovski: I decided to position it there since it's somewhat related to fullscreen, screen orientation and keep screen on, in the sense that it changes how the device behaves. 
I can move it below those 3.

dshokouhi: Although this largely compliments most of the apps WebView features this feels like a big behavior change for a device that not every user will want to enable on every device. Maybe a new category is a good idea for this so we can also describe why a user may want to set the app as the launcher? They may not understand that they will need to do things like updating the dashboard to add links to apps etc...

One of the reasons why its not good to list it at the top is that users may enable it on accident and that may not lead to a good experience.

jpelgrom: Another thought: make the launcher an activity-alias that can be enabled/disabled and require the user to opt-in to the setting to choose HA as a launcher (= enable the activity alias). We should avoid the possibility that users accidentally set HA as a launcher.

fpetrovski: > make the launcher an activity-alias that can be enabled/disabled
@jpelgrom I can create a new SwitchPreference for this.

> Maybe a new category is a good idea for this
@dshokouhi, would it be acceptable if I add the switch and based on the state show or hide the current Preference item that navigates to the settings? The new category could be named simply named "Launcher" or "Kiosk", although for it to be truly a kiosk app, a few other things need to be implemented.

Another option is to open a new page like "Manage Sensors" or "Notification History", but I'm not sure if it makes sense to have an entire page just for one option. Additionally if the app is being slowly converted to JetpackCompose, adding another fragment does not help :)

fpetrovski:
Here's an updated video of the feature:
https://github.com/user-attachments/assets/47e36894-3b5c-45f0-8d8c-a53f37721c23

I'm not too confident about the icons and I'm open to suggestions :)

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -359,6 +359,24 @@ class SettingsFragment(
...
+        findPreference<Preference>("set_launcher_app")?.let {
+            it.isVisible = findPreference<SwitchPreference>("enable_ha_launcher")?.isChecked ?: false
+            it.summary = getString(commonR.string.default_launcher_prompt_def, getDefaultLauncherInfo())
+            it.setOnPreferenceClickListener {
+                val intent = Intent(Settings.ACTION_HOME_SETTINGS)
+                    .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                startActivity(intent)
+                true
+            }
+        }
```
TimoPtr: The class is already quite big (same for the function itself) I would be in favor of grouping this two findPref into one private method.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsPresenterImpl.kt
```diff
@@ -295,4 +305,12 @@ class SettingsPresenterImpl @Inject constructor(
             suggestionFlow.emit(null)
         }
     }
+
+    private fun enableLauncherMode(enable: Boolean) {
+        view.getPackageManager()?.setComponentEnabledSetting(
+            launcherAliasComponent,
+            if (enable) PackageManager.COMPONENT_ENABLED_STATE_ENABLED else PackageManager.COMPONENT_ENABLED_STATE_DISABLED,
+            PackageManager.DONT_KILL_APP
```
TimoPtr: Does it make sense to keep this flag when we set the state to disable?

TimoPtr (commented):
Looks good I've played with it a bit. The only thing that is a bit weird is that when your back stack is empty when you press back it reloads the app. We could capture this in the WebViewActivity, but at the same time it's kinda nice when we are in launcher mode to be able to refresh the app.

I'll let @jpelgrom the final word on the PR 👍🏻

common/src/main/res/values/strings.xml
```diff
@@ -1382,4 +1382,10 @@
     <string name="thermostat_tile">Thermostat tile</string>
     <string name="climate_heating">Heating</string>
     <string name="climate_cooling">Cooling</string>
+    <string name="launcher">Launcher</string>
+    <string name="launcher_option_title">Enable Home Assistant launcher mode</string>
+    <string name="launcher_option_summary">Add Home Assistant to the list of available device launchers that can replace your device\'s home screen</string>
+    <string name="default_launcher_prompt">Change launcher app</string>
+    <string name="default_launcher_prompt_def">Open settings to choose a default launcher app. Currently set to %s</string>
```
```suggestion
    <string name="default_launcher_prompt_def">Open settings to choose a default launcher app. Currently set to %s.</string>
```

common/src/main/res/values/strings.xml
```diff
@@ -1382,4 +1382,10 @@
     <string name="thermostat_tile">Thermostat tile</string>
     <string name="climate_heating">Heating</string>
     <string name="climate_cooling">Cooling</string>
+    <string name="launcher">Launcher</string>
+    <string name="launcher_option_title">Enable Home Assistant launcher mode</string>
```
jpelgrom: "launcher mode" suggests there are other changes in the app to me. I was thinking of "Available as launcher app" but that may be too abstract without reading the description.

jpelgrom (commented):
"Launcher" is obvious to me as someone who has used Android for a long time, but maybe not for everyone. The system settings app now seem to call it "home app", should that be mentioned somewhere or used ("launcher/home app" in the description)?

Note there is a merge conflict.

> I'll let @jpelgrom the final word on the PR 👍🏻

No change in my thoughts for if it fits (I think for the x% that may want behavior like this, 90% are served by app pinning). However, it now has enough safeguards that it shouldn't cause major issues, and I won't block it.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsPresenterImpl.kt
```diff
@@ -97,6 +106,9 @@ class SettingsPresenterImpl @Inject constructor(
                         if (value) PackageManager.COMPONENT_ENABLED_STATE_DEFAULT else PackageManager.COMPONENT_ENABLED_STATE_DISABLED,
                         PackageManager.DONT_KILL_APP,
                     )
+                "enable_ha_launcher" -> {
+                    enableLauncherMode(value)
+                }
```
jpelgrom:
```suggestion
                "enable_ha_launcher" -> enableLauncherMode(value)
```

jpelgrom (commented):
Strings look good to me now, we can always tweak later based on feedback.

TimoPtr:
Thanks @fpetrovski for your contribution 💪🏻

PR #5373, opened 2025-05-28
Introduce FailFast utils to provide offensive programing

TimoPtr:
## Summary

While developing sometimes we encounter places where it should never throw an exception but to add more safety this PR introduce a mechanism that will crash crash in DEBUG if this kind of issue are encounter. It makes the error important to investigate ASAP.

It is also useful for unlocking an issue that happen in prod but we don't know how to reproduce it.

The design is flexible so that if we want we can override it so that the errors are reported to Sentry for instance. We could also add an option for us to enable the crash on release too. It would be useful for developer to get the error when we are in production.

## Checklist

## Screenshots

In debug the application process is killed and you have this message in the logs
image

In release the application is kept alive and you have this message in the logs
image

## Any other notes

I've catch two errors that we have on Sentry that we don't know yet how to reproduce.


bgoncal (changes requested):
I would first test with a condition to only crash while in debug, just to have an amostrage of how often this may happen in the current app state, afterwards I totally support the approach you are proposing.

Have in mind that crashes are something we need to fix quicker because depending where they are it can be a total blocker for the user, considering Google's "speed" recently on review approval, I would take that into consideration.

bgoncal (approved):
Approving based on our talk in slack, I missed that this will only happen in debug

jpelgrom:
> In release the application is kept alive and you have this message in the logs image

Regular users don't use logcat. I worry that this is too much formatting (trying to create a box) which is lost when copying from the app's logs section.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManagerImpl.kt
```diff
@@ -89,7 +90,12 @@ class ServerManagerImpl @Inject constructor(
         mutableServers.values.any {
             it.type == ServerType.DEFAULT &&
                 it.connection.isRegistered() &&
-                authenticationRepository(it.id).getSessionState() == SessionState.CONNECTED
+                FailFast.failOnCatch(
+                    message = {
+                        """Fail to get authenticationRepository for ${it.id} current authenticationRepos ids: ${authenticationRepos.keys} """
```
jpelgrom, line 95:
```suggestion
                        """Failed to get authenticationRepository for ${it.id}. Current repository ids: ${authenticationRepos.keys}."""
```

app/src/full/kotlin/io/homeassistant/companion/android/location/HighAccuracyLocationService.kt
```diff
@@ -148,7 +149,12 @@ class HighAccuracyLocationService : Service() {
         notification = notificationBuilder.build()
 
         val type = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) FOREGROUND_SERVICE_TYPE_LOCATION else 0
-        LAUNCHER.onServiceCreated(this, notificationId, notification, type)
+        FailFast.failOnCatch {
+            // Sometimes the service cannot be started as foreground due to the app being in a state where
+            // this is not allowed. We didn't identified yet how to avoid starting the service in this state.
+            // To avoid a crash, we catch it with FailFast, which will only crash in debug builds.
```
jpelgrom, line 155: Isn't this line obvious, so unnecessary?

app/src/full/kotlin/io/homeassistant/companion/android/location/HighAccuracyLocationService.kt
```diff
@@ -148,7 +149,12 @@ class HighAccuracyLocationService : Service() {
         notification = notificationBuilder.build()
 
         val type = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) FOREGROUND_SERVICE_TYPE_LOCATION else 0
-        LAUNCHER.onServiceCreated(this, notificationId, notification, type)
+        FailFast.failOnCatch {
+            // Sometimes the service cannot be started as foreground due to the app being in a state where
+            // this is not allowed. We didn't identified yet how to avoid starting the service in this state.
```
jpelgrom, line 154:
```suggestion
            // this is not allowed. We haven't identified how to avoid starting the service in this state yet.
```

TimoPtr:
image
image
I've rework a bit the display of the header of the error so it looks smaller on a phone. I think we need the header to make it easy to spot.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/FailFast.kt
```diff
@@ -0,0 +1,61 @@
...
+ * [DefaultFailFastHandler] behavior is different based on the build target debug or release.
+ */
+object FailFast {
+    private var handler: FailFastHandler = DefaultFailFastHandler
+
+    fun setHandler(handler: FailFastHandler) {
+        this.handler = handler
+    }
+
+    fun failWhen(condition: Boolean, message: () -> String) {
```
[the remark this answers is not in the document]
jpelgrom: You did full paragraph style again 🙃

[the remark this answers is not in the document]
TimoPtr: It looks better when you get the popup from the IDE when calling the function 😂.

jpelgrom (approved):

Issue #5328, opened 2025-05-20
Edge to edge support is enabled by default


TimoPtr:
Starting with API 36 edge to edge is enable and can't be disabled. We need to support it. We need https://github.com/home-assistant/frontend/pull/24689 to be merged in order to have a proper support within the WebView.

We need to check all the other screens like the settings that are broken at the moment when edge to edge is enabled.

https://developer.android.com/about/versions/16/behavior-changes-16#edge-to-edge

PR #5387, opened 2025-06-03
Enable edge-to-edge support for BaseActivity

TimoPtr:
## Summary

This PR enable edge-edge support at the BaseActivity level. It removes the indivual calls that were made in some other Base classes like for configure widget.

## Checklist

## Any other notes

Closes https://github.com/home-assistant/android/issues/5328. This is the final piece for the native screen of the app to support edge to edge.


jpelgrom:
Should the PR's status be changed to draft while we wait for the WebView changes, as those also extend BaseActivity?

jpelgrom:
I believe it's related to strict mode rather than the code change in this PR, but the barcode scanner immediately crashes for me.

TimoPtr:
> I believe it's related to strict mode rather than the code change in this PR, but the barcode scanner immediately crashes for me.

Indeed it crashes for me too on main. The issue seems to come from wrong context being used.

TimoPtr:
Tracked here https://github.com/journeyapps/zxing-android-embedded/issues/762 but it seems that the lib is not maintained anymore. I'm going to ignore the issue for now.

PR #5390, opened 2025-06-04
Edge to edge support for webviews using <View> bars

Gregman-js:
## Summary
 - Enable `enableEdgeToEdge()` for all
 - Changing active server supported (tested by changing server in settings)
 - Add `<View>` on top and bottom of `<Webview>`
 - Control above views height with `InsetsUtil.kt`
```kotlin
if (!serverHandleInsets) {
    statusBarBackground?.updateLayoutParams {
        height = safeInsets.top
    }
    navigationBarBackground?.updateLayoutParams {
        height = safeInsets.bottom
    }
} else {
// evaluateJavascript...
}
```
 - Support new and old status bar apis
Android >= 15 will use Views background color
Android < 15 will use Views background colors and statusBarColor on top
```kotlin
if (statusBarColor != 0 && !serverHandleInsets) {
    window.statusBarColor = statusBarColor
    binding.statusBarBackground.setBackgroundColor(statusBarColor)
} else {
    Timber.e("Skipping coloring status bar...")
}
if (navigationBarColor != 0 && !serverHandleInsets) {
    window.navigationBarColor = navigationBarColor
    binding.navigationBarBackground.setBackgroundColor(navigationBarColor)
} else {
    Timber.e("Skipping coloring navigation bar...")
}
```

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -1273,23 +1273,28 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
 
         // Set background colors
         if (statusBarColor != 0 && !serverHandleInsets) {
-            window.statusBarColor = statusBarColor
             binding.statusBarBackground.setBackgroundColor(statusBarColor)
         } else {
             Timber.e("Skipping coloring status bar...")
         }
-        if (navigationBarColor != 0 && !serverHandleInsets) {
-            window.navigationBarColor = navigationBarColor
-            binding.navigationBarBackground.setBackgroundColor(navigationBarColor)
+
+        if (navigationBarColor != 0) {
+            // Below Android Q windowInsetsController.isAppearanceLightNavigationBars does not set navigation background color
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                window.navigationBarColor = ColorUtils.setAlphaComponent(navigationBarColor, 128)
+            }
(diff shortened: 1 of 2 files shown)


TimoPtr:
@Gregman-js I'm working on migrating the webViews to compose so that we can handle this into one place.

TimoPtr:
@Gregman-js here is the PR https://github.com/home-assistant/android/pull/5419 for the migration to Compose. It will be easier to apply your fix in compose and it will work for the other screens too.

You are more than welcome to review the PR :)

TimoPtr:
@Gregman-js The PR about compose has been merged. Could you impl your proposal in compose only and apply it to the two places that uses a WebView. FYI we can't get the version of a server without being fully authenticated so in the onboarding we cannot assume that the frontend support edge to edge yet.

TimoPtr:
It seems that in some case we have the capability to get the version of the server through mDNS https://github.com/home-assistant/android/blob/d3408300ddf984fb8363bd148142a7767d7eb146/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/discovery/HomeAssistantSearcher.kt#L100

Then it won't work for ppl that set the address manually. But it would cover most of the case which is nice. 

It's important because sometimes the users are making the setup of their home assistant instances directly on their phone so we want to make sure that edge to edge is properly handled to not hide anything.

TimoPtr:
I did the inital version in compose here https://github.com/home-assistant/android/pull/5507 it adds the bar with the right color. I'll make a second PR with the needed changes to react to a frontend that is compatible. @Gregman-js if you want to take a look.

TimoPtr:
Closing this PR since now we are using compose 👍🏻

PR #5408, opened 2025-06-11
Introduce HAMediaPlayer to migrate existing player to compose

TimoPtr:
## Summary

In order to migrate to compose the `WebViewActivity` we need to have a compose view of the media player. By default only the surface is provided not the control, this PR adds a very basic implementation of the controls. It is not feature complete. Since the WebView at the moment is only using this player to play a camera, the implementation is currently limited to this use case.

I decided to first make a PR that replace the player in `DevPlayground` so it can be use later in the PR for the WebView to compose.

## Checklist

## Any other notes

I'm not a compose expert so any comment on how to make it better is very much appreciated.


jpelgrom:
I think this would qualify as visual changes not in the frontend, so please include a screenshot in the future for similar changes ;)

TimoPtr:
I will the only thing for now is that until it is not integrated in the app it's not very clear how it looks like. I'm going to add how it looks like on the dev activity.

app/src/main/kotlin/io/homeassistant/companion/android/player/BufferingState.kt
```diff
@@ -0,0 +1,48 @@
+package io.homeassistant.companion.android.player
```
jpelgrom: Is this the right package? It feels like we generally only put app features on this level, this may be better placed in util (like the Composables in util.icondialog).

app/src/main/kotlin/io/homeassistant/companion/android/player/FakePlayer.kt
```diff
@@ -0,0 +1,526 @@
+package io.homeassistant.companion.android.player
```
jpelgrom: This is in the main source set because of the preview, correct? It feels wrong, but I don't know how to fix it other than removing the preview.

app/src/main/kotlin/io/homeassistant/companion/android/player/HAMediaPlayer.kt
```diff
@@ -0,0 +1,377 @@
...
+    if (autoHideControl) {
+        LaunchedEffect(showControls) {
+            if (showControls) {
+                delay(AUTO_HIDE_DELAY)
+                showControls = false
+            }
+        }
+    }
+
+    // TODO we could move the Box outside and make BoxScope.Content instead and emit a fullScreenOnClick callback to let the caller handle the fullscreen button
```
jpelgrom: At what point do you think letting the caller handling the fullscreen button would be useful? (Trying to avoid features and their added complexity here if we don't need it.)

app/src/main/kotlin/io/homeassistant/companion/android/player/HAMediaPlayer.kt
```diff
@@ -0,0 +1,377 @@
...
+private fun BottomControls(
+    player: Player,
+    showControls: Boolean,
+    isFullScreen: Boolean,
+    onClickFullscreen: () -> Unit,
+    modifier: Modifier,
+) {
+    Row(
+        modifier = modifier
+            .height(60.dp)
```
jpelgrom: The buttons placed inside are 64dp, is this on purpose?

app/src/main/kotlin/io/homeassistant/companion/android/player/HAMediaPlayer.kt
```diff
@@ -0,0 +1,377 @@
...
+                releasePlayer()
+            }
+        }
+    }
+
+    player?.let { player ->
+        // Allow the PiP to have control of the player
+        MediaSession.Builder(context, player)
+            .setId(UUID.randomUUID().toString())
+            .build()
```
jpelgrom: While this could be nice, I'm cautious as it might create more issues like home-assistant/frontend#23230. We currently don't have it so I'm leaning towards not introducing it in this PR, so if any issues arise we know it isn't because of this.

app/src/main/kotlin/io/homeassistant/companion/android/player/MuteUnmuteButtonState.kt
```diff
@@ -0,0 +1,88 @@
...
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.media3.common.Player
+import androidx.media3.common.listen
+import androidx.media3.common.util.UnstableApi
+import androidx.media3.common.util.Util.shouldShowPlayButton
+
+// TODO update docs
```
jpelgrom: Dev docs or ...?

app/src/main/kotlin/io/homeassistant/companion/android/player/BufferingState.kt
```diff
@@ -0,0 +1,48 @@
...
+import androidx.media3.common.util.UnstableApi
+
+/**
+ * Remembers the value of [MuteUnmuteButtonState] created based on the passed [Player] and launch a
+ * coroutine to listen to [Player's][Player] changes. If the [Player] instance changes between
+ * compositions, produce and remember a new value.
+ */
+
+@Composable
+@OptIn(UnstableApi::class)
```
jpelgrom: Do we need this opt in, as the `BufferingState` already opts in? AFAIK you only need it at the location you directly use the `@UnstableApi`.

app/src/main/kotlin/io/homeassistant/companion/android/player/HAMediaPlayer.kt
```diff
@@ -0,0 +1,377 @@
...
+    if (muteState.isEnabled) {
+        IconButton(
+            modifier = Modifier
+                .align(Alignment.CenterVertically)
+                .size(64.dp),
+            onClick = muteState::onClick,
+            enabled = showControls,
+        ) {
+            val icon =
+                if (muteState.showMute) Icons.AutoMirrored.Filled.VolumeMute else Icons.AutoMirrored.Filled.VolumeUp
```
jpelgrom: Might be personal preference, but `VolumeOff` seems like a better fit than `VolumeMute`, for example the system volume level also uses a slashed icon when at 0.

jpelgrom (commented):
Seems like a pretty good start for a basic player!

If I turn off the screen, I lose video progress and it starts back at the start of the video. Not entirely sure if that is the demo video or the Compose implementation.

Small size player highlights layout issues:

 - the bottom bar partially obscures the play/pause icon
 - start and end padding isn't the same because of the icon button, I think the text needs more padding at the start

image

LeakCanary popped up after exiting the activity, saying there was an issue:

<details>
<summary>LeakCanary trace</summary>

```
┬───
│ GC Root: System class
│
├─ android.provider.FontsContract class
│    Leaking: NO (HomeAssistantApplication↓ is not leaking and a class is never
│    leaking)
│    ↓ static FontsContract.sContext
├─ io.homeassistant.companion.android.HomeAssistantApplication instance
│    Leaking: NO (Application is a singleton)
│    mBase instance of android.app.ContextImpl
│    ↓ Application.mLoadedApk
│                  ~~~~~~~~~~
├─ android.app.LoadedApk instance
│    Leaking: UNKNOWN
│    Retaining 10.1 kB in 162 objects
│    mApplication instance of io.homeassistant.companion.android.
│    HomeAssistantApplication
│    Receivers
│    ..DemoExoPlayerActivity@38767496
│    ....MediaSessionLegacyStub$MediaButtonReceiver@39579224
│    ..HomeAssistantApplication@33649392
│    ....MediaPlayerControlsWidget@35483216
│    ....SensorReceiver@35328608
│    ....VisibilityTracker@34591576
│    ....cs@39014000
│    ....TemplateWidget@35487080
│    ....io@39255048
│    ....ButtonWidget@35467640
│    ....WebsocketBroadcastReceiver@35102864
│    ....TodoWidget@35494184
│    ....it@39172856
│    ....NetworkTypeObserver$Receiver@39433520
│    ....EntityWidget@35474104
│    ....ik@39255632
│    ↓ LoadedApk.mReceivers
│                ~~~~~~~~~~
├─ android.util.ArrayMap instance
│    Leaking: UNKNOWN
│    Retaining 9.6 kB in 148 objects
│    ↓ ArrayMap.mArray
│               ~~~~~~
├─ java.lang.Object[] array
│    Leaking: UNKNOWN
│    Retaining 9.5 kB in 146 objects
│    ↓ Object[0]
│            ~~~
╰→ io.homeassistant.companion.android.developer.DemoExoPlayerActivity instance
​     Leaking: YES (ObjectWatcher was watching this because io.homeassistant.
​     companion.android.developer.DemoExoPlayerActivity received
​     Activity#onDestroy() callback and Activity#mDestroyed is true)
​     Retaining 280.9 kB in 4236 objects
​     key = 6f808c2c-a2cc-4915-8dc3-e1ac7f0be014
​     watchDurationMillis = 50022
​     retainedDurationMillis = 45018
​     mApplication instance of io.homeassistant.companion.android.
​     HomeAssistantApplication
​     mBase instance of androidx.appcompat.view.ContextThemeWrapper

METADATA

Build.VERSION.SDK_INT: 36
Build.MANUFACTURER: Google
LeakCanary version: 2.14
App process name: io.homeassistant.companion.android.debug
Class count: 41689
Instance count: 274500
Primitive array count: 197937
Object array count: 40711
Thread count: 50
Heap total bytes: 36924602
Bitmap count: 3
Bitmap total bytes: 36163
Large bitmap count: 0
Large bitmap total bytes: 0
Db 1: open /data/user/0/io.homeassistant.companion.android.
debug/no_backup/androidx.work.workdb
Db 2: open /data/user/0/io.homeassistant.companion.android.
debug/databases/HomeAssistantDB
Stats: LruCache[maxSize=3000,hits=133643,misses=254955,hitRate=34%]
RandomAccess[bytes=12572701,reads=254955,travel=136972692539,range=45653516,size
=56454593]
Analysis duration: 79901 ms
```
</details>

I learned some new KDoc syntax, nice.

app/src/main/kotlin/io/homeassistant/companion/android/player/MuteUnmuteButtonState.kt
TimoPtr: The whole docs of the class need to be revisited I've forgot to update after copying the impl of play pause from media3-ui-compose

TimoPtr:
About the Leak did it happen with PiP? Are you able to reproduce it all the time?

app/src/main/kotlin/io/homeassistant/companion/android/player/HAMediaPlayer.kt
TimoPtr: Removing it from the PR since it was also badly implemented on my end ... I forgot to release the session that leads to another leak.

app/src/main/kotlin/io/homeassistant/companion/android/player/HAMediaPlayer.kt
TimoPtr: You're right it looks better

app/src/main/kotlin/io/homeassistant/companion/android/player/HAMediaPlayer.kt
TimoPtr: I was not a fan of exposing two modifier, that's why I've added this. But I'm not a fan to also delegate the click ... I will remove the TODO but if you have a suggestion let me know.

app/src/main/kotlin/io/homeassistant/companion/android/player/FakePlayer.kt
TimoPtr: Yeah I think we do this in other places like in `io.homeassistant.companion.android.util.PreviewData`. It's a shame that this has to be in the main sourceSet ...

app/src/main/kotlin/io/homeassistant/companion/android/player/HAMediaPlayer.kt
TimoPtr: FYI https://github.com/home-assistant/frontend/pull/25767

TimoPtr:
On small layout the current layout (from media3-ui) hide the time and display only the progress
image

I decided to hide the bottom controls when there is not enough space for it.

To handle
> If I turn off the screen, I lose video progress and it starts back at the start of the video. Not entirely sure if that is the demo video or the Compose implementation.

I would be in favor of introducing a viewModel. I didn't here but it's getting more and more complicated and a VM would make it cleaner. I don't think we need that for the first impl for the WebView.

TimoPtr:
While working on the migration of the WebViewActivity I came around the fact that I should have extract the Player initialization out of the Compose definition. 🤦🏻

jpelgrom:
> About the Leak did it happen with PiP? Are you able to reproduce it all the time?

I was able to reproduce it all the time, but not anymore since you removed the MediaSession. Just opening the activity, and going back using the system back button was enough to trigger it.

> On small layout the current layout (from media3-ui) hide the time and display only the progress
> I decided to hide the bottom controls when there is not enough space for it.

OK, seems reasonable. New issue that also would apply to the classic view ExoPlayer: the preview in the playground activity can no longer be full screened, maybe it's better to make that one bigger so we can test all functionality (making it smaller is easy enough with splitscreen).

> I don't think we need that for the first impl for the WebView.

OK, let's keep that out of scope for this PR.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/media/player/HAMediaPlayer.kt
```diff
@@ -0,0 +1,346 @@
...
+    modifier: Modifier = Modifier,
+    fullscreenModifier: Modifier = Modifier,
+) {
+    var showControls by remember { mutableStateOf(true) }
+
+    LaunchedEffect(showControls) {
+        if (showControls) {
+            delay(AUTO_HIDE_DELAY)
+            showControls = false
+            Timber.e("Hiding controls")
```
jpelgrom: Should be removed before merging or at least not an error

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/media/player/HAMediaPlayer.kt
```diff
@@ -0,0 +1,346 @@
...
+        onClick = onClickFullscreen,
+        modifier = Modifier
+            .align(Alignment.CenterVertically)
+            .size(BottomControlsHeight),
+    ) {
+        Icon(
+            if (isFullScreen) Icons.Default.FullscreenExit else Icons.Default.Fullscreen,
+            contentDescription = stringResource(R.string.fullscreen),
+            tint = Color.White,
+            modifier = Modifier.size(BottomControlButtonSize),
```
jpelgrom: Nitpick: for the `MuteButton` the order is reversed

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/media/player/HAMediaPlayer.kt
```diff
@@ -0,0 +1,346 @@
...
+    }
+}
+
+@Composable
+private fun RowScope.FullscreenButton(isFullScreen: Boolean, onClickFullscreen: () -> Unit) {
+    IconButton(
+        onClick = onClickFullscreen,
+        modifier = Modifier
+            .align(Alignment.CenterVertically)
+            .size(BottomControlsHeight),
```
jpelgrom, line 256: Nitpick: for the `MuteButton` the order is reversed

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/media/player/HAMediaPlayer.kt
```diff
@@ -0,0 +1,346 @@
...
+        while (true) {
+            currentPosition = player.currentPositionDuration()
+            delay(TIME_TRACKING_UPDATE_DELAY)
+        }
+    }
+
+    Text(
+        currentPosition.toComponents { hours, minutes, seconds, _ ->
+            if (hours > 0L) {
+                String.format(Locale.US, "%02d:%02d:%02d", hours, minutes, seconds)
```
jpelgrom: I'm not aware of any locale specific notation for a timer, but why not use `Locale.getDefault()`?

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/media/player/MuteUnmuteButtonState.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    var showMute by mutableStateOf(shouldShowMuteUnmuteButton(player))
+        private set
+
+    /**
+     * Handles the interaction with the MuteUnmute button according to the current state of the
+     * [Player].
+     *
+     * If the volume is set to 0 then it will set it to 1 (unmute) otherwise it will set to 0 (mute).
+     *
```
jpelgrom:
```suggestion
```

app/src/main/kotlin/io/homeassistant/companion/android/player/MuteUnmuteButtonState.kt
jpelgrom: Looks to me like it is done, please mark this as resolved if done.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/media/player/HAMediaPlayer.kt
TimoPtr: I'm not to I just fixed a local but I will use default.

PR #5409, opened 2025-06-11
Use new frontend entity more info deeplink

jpelgrom:
## Summary

The latest frontend release randomly added an official option to open the more info dialog programmatically, [using a query paramater](https://github.com/home-assistant/frontend/pull/25733), after we have been unable to get such a frontend change for the app's deeplinks for years. Better late than never, I guess.

This PR switches from the workaround to the new query parameter to open more info, if the server version is recent enough.

## Checklist

## Screenshots

Using two examples, but this obviously applies to all. Also confirmed it works correctly if you changed the default dashboard/url. Really nice how responsive this now feels!

 - From a push notification

https://github.com/user-attachments/assets/0604d1be-2bf6-409c-8f27-2edc04f3c0fc

 - From device controls

https://github.com/user-attachments/assets/5fa702c5-f17b-4f7a-8721-3a83ded1f042

n/a

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -1157,14 +1157,22 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
         super.onWindowFocusChanged(hasFocus)
         if (hasFocus && !isFinishing) {
             unlockAppIfNeeded()
-            val path = intent.getStringExtra(EXTRA_PATH)
-            presenter.onViewReady(path)
+            var path = intent.getStringExtra(EXTRA_PATH)
             if (path?.startsWith("entityId:") == true) {
                 // Get the entity ID from a string formatted "entityId:domain.entity"
                 // https://github.com/home-assistant/core/blob/dev/homeassistant/core.py#L159
                 val pattern = "(?<=^entityId:)((?!.+__)(?!_)[\\da-z_]+(?<!_)\\.(?!_)[\\da-z_]+(?<!_)$)".toRegex()
-                moreInfoEntity = pattern.find(path)?.value ?: ""
+                val entity = pattern.find(path)?.value ?: ""
+                if (
+                    entity.isNotBlank() &&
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1157,14 +1157,22 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
+                if (
+                    entity.isNotBlank() &&
+                    serverManager.getServer(presenter.getActiveServer())?.version?.isAtLeast(2025, 6, 0) == true
+                ) {
+                    path = "/?more-info-entity-id=$entity"
+                } else {
+                    moreInfoEntity = entity
+                }
             }
+            presenter.onViewReady(path)
```
TimoPtr, line 1175: FYI @bgoncal tried to make it work with the new `navigate` external command but for now it doesn't work.

TimoPtr (approved):

TimoPtr:
Do you know if it's properly documented that we can do this? Also maybe you could add a section in the changelog about this new capability?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: We've had this for a long time so just switching that implementation in this PR. Using navigate for changing paths would be a different PR.

jpelgrom:
Usually the frontend and core teams don't document this separately. App functionality is documented on the companion app docs.

I'll update the changelog when I'm able to tonight (CEST).

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: Once this is merged https://github.com/home-assistant/frontend/pull/28375 we should be able to use redirect and avoid the webview to reload to give a better experience.

TimoPtr: @jpelgrom I think that while we are using the WebViewActivity it is not helping since each time we open a link we end up into the LinkActivity that is going to force the recreation of the WebViewActivity.

PR #5425, opened 2025-06-17
Rework MyActivity to support Invite and improve UX

TimoPtr:
## Summary

This PR enhances deep link handling in `MyActivity` by:
1.  Introducing support for `https://my.home-assistant.io/invite#url=...` universal links (and deep link `homeassistant://invite#url=...` ), directing users to the server onboarding flow with the specified URL.
2.  Improving the user experience for `https://my.home-assistant.io/redirect/...` universal links by navigating directly to the relevant content within the app, instead of opening a WebView with a button.

**Motivation & Background:**

*   **Invite Links:** To streamline server setup by allowing users to directly onboard via a shared link containing the server URL. The server URL is passed in the URL fragment for security (to prevent it from being sent to `my.home-assistant.io`) so we have to manipulate the fragment and transform it into a query.
*   **Redirect Links UX:** Previously, `redirect` links opened a WebView as a security measure. After discussion with the frontend team, it was determined that since the frontend already validates these links and user interaction is expected, a more seamless in-app navigation is preferable. This change also ensures these links work even if `my.home-assistant.io` is temporarily unavailable or the user doesn't have internet (on their local network).
*   **Task Affinity Bug:** Resolved an issue where `MyActivity`, due to its `standard` launchMode, would incorrectly appear as part of the calling app's task (e.g., Home Assistant appearing within Google Keep's task in the recents screen). `MyActivity` now correctly launches in its own task.
*   **URL Parameter Encoding:** Dedicated effort was made to ensure URLs with percent-encoded parameters are handled correctly throughout the linking process, preventing the generation of invalid URLs from valid incoming links.

**Impact:**

*   Users can now be invited to a Home Assistant instance via a simple link, automatically pre-filling the server URL for onboarding.
*   Navigating via `redirect` links from `my.home-assistant.io` will feel more integrated, taking users directly to the content within the app.
*   Fixes a bug where the app could be incorrectly grouped with the launching app in the Android task switcher.

## Checklist

## Screenshots

**Before:** Home Assistant shown as part of the calling app's task (e.g., Google Keep) and opening a WebView.
https://github.com/user-attachments/assets/0982b09d-032b-4e89-bed7-015c4086ee51

**Now:** Home Assistant correctly opens in its own task and supports invite without opening an intermediate WebView.
https://github.com/user-attachments/assets/6e42f506-0e58-4a1c-ac41-dba62ed5bd65

## Any other notes

`MyActivity` has been renamed to `LinkActivity` to be more generic.
This functionality mirrors a similar feature added on iOS (see https://github.com/home-assistant/iOS/pull/3591).
The `FailFast` utility has been extended with a simpler method `fail` to cover more use cases.


app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchActivity.kt
```diff
@@ -42,9 +43,19 @@ import kotlinx.coroutines.launch
 import retrofit2.HttpException
 import timber.log.Timber
 
+private const val TO_SERVER = "go_to_server"
+
 @AndroidEntryPoint
 class LaunchActivity : AppCompatActivity(), LaunchView {
 
+    companion object {
+        fun newInstanceToSpecificServer(context: Context, serverUrl: String): Intent {
```
TimoPtr: I'm not sure at all about the naming of the variables in the `Lauch*` classes. So if you have suggestion be my guest.

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchActivity.kt
```diff
@@ -125,8 +136,8 @@ class LaunchActivity : AppCompatActivity(), LaunchView {
         overridePendingTransition(0, 0) // Disable activity start/stop animation
     }
 
-    override fun displayOnBoarding(sessionConnected: Boolean) {
-        registerActivityResult.launch(OnboardApp.Input())
+    override fun displayOnBoarding(sessionConnected: Boolean, gotoServer: String?) {
+        registerActivityResult.launch(OnboardApp.Input(url = gotoServer))
```
Marchat9:
```suggestion
    override fun displayOnBoarding(sessionConnected: Boolean, goToServer: String?) {
        registerActivityResult.launch(OnboardApp.Input(url = goToServer))
```

to keep camelcase practice

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchActivity.kt
jpelgrom: I wouldn't expect "goto server" from launching the app to open onboarding, but rather a server in the app (in which case you'd use the server ID). Why not something like `onboardServer` or `addServer`? That feels obvious to me...

TimoPtr:
I made a mistake in the implementation it should support deep link rather than universal link, that is different from the iOS impl. It means that it doesn't open the my website. 

Also another mistake is that I should treat the fragment as an URI not using simple string manipulation.

I decided to make it better and introduce another class that will be agnostic from the activity so that I can write Unit Test about it. I don't think it needs to be a ViewModel but it could.

app/src/main/kotlin/io/homeassistant/companion/android/launch/my/MyLinkHandler.kt
```diff
@@ -0,0 +1,129 @@
...
+        if (uri.scheme != "https" || uri.host != MY_BASE_DOMAIN) {
+            FailFast.fail {
+                "Invalid deep link double check the host or the scheme: $uri"
+            }
+            return LinkDestination.NoDestination
+        }
+
+        return when {
+            path.startsWith(REDIRECT_URL_PATH) -> handleNavigateDeepLink(uri)
+            path.startsWith(INVITE_URL_PATH) -> handleInviteUniversalLink(uri)
```
balloob: this will also match `https://my.home-assistant/invite-blablablabla`

balloob: Netlify hosts My Home Assistant and it automatically adds trailing slashes to the URL when visiting.

TimoPtr: Thanks for the clarification 👍. I need to update the reason for this in the code, I initially thought that the PR on the frontend was on my repository.

So the reason why I remove the trailing slash is for people that use an old version of the frontend that doesn't support trailing slash.

bgoncal (approved):

app/src/main/res/xml/changelog_master.xml
```diff
@@ -1,15 +1,17 @@
...
+        <change>Redirect links from my.home-assistant.io directly open the app without confirmation</change>
         <change>Bug fixes and dependency updates</change>
     </release>
-    <release version="2025.6.3 - Wear" versioncode="2">
+    <release version="2025.6.4 - Wear" versioncode="2">
         <change>Bug fixes and dependency updates</change>
     </release>
-    <release version="2025.6.3 - Automotive" versioncode="1">
+    <release version="2025.6.4 - Automotive" versioncode="1">
         <change>Bug fixes and dependency updates</change>
```
dshokouhi, line 17: This feature should also work for Automotive right? Might be nice to mention it so they know they dont have to type in the full URL and can use an invite link too?

TimoPtr: Good point 👍

dshokouhi:
is there an easy way to test this? For some reason the debug build does not register teh deep links nor allows me to select them, so the browser keeps opening telling me to install the app.

image

image

Testing this by saving links in gmail and clicking on them to trigger the process

TimoPtr:
It's because your production app already captures the links. We should have universal links for debug but that's a bigger task.

jpelgrom:
This seems incomplete to me for the flow where a user is completely new and doesn't have the app. My page opens in browser (because there is no app) > user taps Google Play badge to install the app > user returns to browser > tapping 'Accept invite' doesn't do anything because you haven't added support for `homeassistant://invite#url=http%3A%2F%2Fhomeassistant.local%3A8123`. So this either needs a change on the my page, or in the app. I'm guessing changing the app is easier and backwards compatible ;)

app/src/main/kotlin/io/homeassistant/companion/android/launch/my/MyLinkHandler.kt
```diff
@@ -0,0 +1,131 @@
...
+        if (uri.getQueryParameter(MOBILE_PARAM) == MOBILE_VALUE) {
+            FailFast.fail { "$MOBILE_PARAM flag already present and set to 1, the deep link is already handled" }
+            return LinkDestination.NoDestination
+        }
+
+        val path = uri.buildUpon()
+            // We strip last / to handle old links created before https://github.com/home-assistant/frontend/pull/25841
+            // A trailing slash is always added by Netlify that is used to host https://my.home-assistant.io/, but
+            // the frontend was not supporting having a trailing slash before https://github.com/home-assistant/frontend/pull/25841
+            // in order to backward compatible we remove the trailing slash here.
```
jpelgrom: This is more than one sentence 


```suggestion
            // We strip the last / to handle old links created before https://github.com/home-assistant/frontend/pull/25841.
            // A trailing slash is always added by Netlify that is used to host https://my.home-assistant.io/, but
            // the frontend did not support having a trailing slash before https://github.com/home-assistant/frontend/pull/25841.
            // For backward compatibility, we remove the trailing slash here.
```

app/src/main/res/xml/changelog_master.xml
```diff
@@ -1,15 +1,19 @@
 <?xml version="1.0" encoding="utf-8"?>
 <changelog xmlns:tools="http://schemas.android.com/tools"
     tools:ignore="MissingDefaultResource">
-    <release version="2025.6.3 - Main" versioncode="3">
+    <release version="2025.6.4 - Main" versioncode="3">
```
```suggestion
    <release version="2025.6.5 - Main" versioncode="3">
```

app/src/main/res/xml/changelog_master.xml
```diff
@@ -1,15 +1,19 @@
...
         <change>The app now uses all available space on the screen (edge-to-edge support)</change>
         <change>Viewing more info for entities from device controls, notifications, shortcuts and tiles is now more responsive</change>
+        <change>Support invitation links from my.home-assistant.io</change>
+        <change>Redirect links from my.home-assistant.io directly open the app without confirmation</change>
         <change>Bug fixes and dependency updates</change>
     </release>
-    <release version="2025.6.3 - Wear" versioncode="2">
+    <release version="2025.6.4 - Wear" versioncode="2">
```
```suggestion
    <release version="2025.6.5 - Wear" versioncode="2">
```

app/src/main/res/xml/changelog_master.xml
```diff
@@ -1,15 +1,19 @@
...
         <change>Bug fixes and dependency updates</change>
     </release>
-    <release version="2025.6.3 - Automotive" versioncode="1">
+    <release version="2025.6.4 - Automotive" versioncode="1">
```
```suggestion
    <release version="2025.6.5 - Automotive" versioncode="1">
```

jpelgrom (commented):
Minor comments about text, otherwise code looks good.

jpelgrom (approved):

PR #5458, opened 2025-06-26
Disable ha_launcher by default and hide it in automotive

TimoPtr:
## Summary

The new feature about being able to use HA as a launcher was not properly setup in the SettingsFragment. It was enabled by default where it should have been disabled by default.

We also don't want to expose this feature in the automotive variant of the app.

## Checklist

CC @fpetrovski

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
@@ -557,11 +557,10 @@ class SettingsFragment(
     }
 
     private fun setupLauncherPrefs() {
-        val launcherCategory = findPreference<PreferenceCategory>("launcher_category")
         val launcherSwitchPref = findPreference<SwitchPreference>("enable_ha_launcher")
         val launcherPref = findPreference<Preference>("set_launcher_app")
 
-        launcherCategory?.isVisible = true
+        findPreference<PreferenceCategory>("launcher_category")?.isVisible = true
 
         launcherSwitchPref?.setOnPreferenceClickListener {
             launcherPref?.isVisible = launcherSwitchPref.isChecked
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
@@ -360,7 +360,9 @@ class SettingsFragment(
             }
         }
 
-        setupLauncherPrefs()
+        if (!isAutomotive) {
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -555,9 +557,12 @@ class SettingsFragment(
     }
 
     private fun setupLauncherPrefs() {
+        val launcherCategory = findPreference<PreferenceCategory>("launcher_category")
```
jpelgrom: We don't need a variable here as it is only used in one place?

```suggestion
        findPreference<PreferenceCategory>("launcher_category")?.isVisible = true
```

\+ remove the other two lines for the launcher category

jpelgrom:
Using this PR to fix some weirdness from the launcher setting: I propose reverting the change from [this suggestion](https://github.com/home-assistant/android/pull/5348#discussion_r2158338771), so always passing `PackageManager.DONT_KILL_APP` as the flag.
It prevents the current behavior where disabling the setting in the app results in the app restarting, which looks like a crash, and doesn't seem to cause any issues (the launcher apps list in Settings still updates immediately, and swiping home returns you to a normal launcher).

jpelgrom:
Other changes get a ✅ from me

fpetrovski:
> Using this PR to fix some weirdness from the launcher setting: I propose reverting the change from [this suggestion](https://github.com/home-assistant/android/pull/5348#discussion_r2158338771), so always passing `PackageManager.DONT_KILL_APP` as the flag. It prevents the current behavior where disabling the setting in the app results in the app restarting, which looks like a crash, and doesn't seem to cause any issues (the launcher apps list in Settings still updates immediately, and swiping home returns you to a normal launcher).

I have another launcher app besides Samsung's OneUI and when the app closed I got the prompt to chose a default launcher, but if you don't have another launcher app, I see how it can be perceived as a crash. 

Also the Automotive variant didn't even cross my mind. I'll remember that for some future contribution

jpelgrom:
> I have another launcher app besides Samsung's OneUI and when the app closed I got the prompt to chose a default launcher, but if you don't have another launcher app, I see how it can be perceived as a crash.

The behavior I'm describing happens even without changing the default launcher. Just toggling it on and off is enough.

TimoPtr:
> > I have another launcher app besides Samsung's OneUI and when the app closed I got the prompt to chose a default launcher, but if you don't have another launcher app, I see how it can be perceived as a crash.
> 
> The behavior I'm describing happens even without changing the default launcher. Just toggling it on and off is enough.

Reverted

Issue #5472, opened 2025-07-01
Device controls sometimes get stuck on loading until app is restarted


dshokouhi:
**Home Assistant Android app version(s):**

2025.6.7

**Android version(s):**

16

**Device model(s):**

Pixel 9 pro xl

**Home Assistant version:**

2025.6.x

**Last working Home Assistant release (if known):**

at least a few months ago as I have not seen this issue before and it occurs pretty frequntly lately

**Description of problem, include YAML if issue is related to notifications:**

**Companion App Logs:**

```
06-30 21:05:35.476  9596  9747 D HaControlsProviderService$createPublisherFor: request 100000
06-30 21:05:35.480  9596 27142 D ServerConnectionInfo: Using external URL
06-30 21:05:35.480  9596  2525 D WebSocketCoreImpl$sendMessage: Sending message 252: {type=config/area_registry/list, id=252}
06-30 21:05:35.480  9596  2525 D WebSocketCoreImpl$sendMessage: Message number 252 sent awaiting answer from WebSocket
06-30 21:05:35.481  9596 32756 D WebSocketCoreImpl$sendMessage: Sending message 253: {type=config/entity_registry/list, id=253}
06-30 21:05:35.481  9596 32756 D WebSocketCoreImpl$sendMessage: Message number 253 sent awaiting answer from WebSocket
06-30 21:05:35.481  9596 27141 D WebSocketCoreImpl$sendMessage: Sending message 254: {type=config/device_registry/list, id=254}
06-30 21:05:35.482  9596 27141 D WebSocketCoreImpl$sendMessage: Message number 254 sent awaiting answer from WebSocket
06-30 21:05:35.490  9596  7105 D WebSocketCoreImpl: Websocket: onMessage (text)
06-30 21:05:35.490  9596  7105 D WebSocketCoreImpl: Message id 252 received
06-30 21:05:35.491  9596  7105 E LogOnlyFailFastHandler: ██████████████████████
06-30 21:05:35.491  9596  7105 E LogOnlyFailFastHandler: !!! CRITICAL FAILURE: FAIL-FAST !!!
06-30 21:05:35.491  9596  7105 E LogOnlyFailFastHandler: ██████████████████████
06-30 21:05:35.491  9596  7105 E LogOnlyFailFastHandler: 
06-30 21:05:35.491  9596  7105 E LogOnlyFailFastHandler: The error has been ignored to avoid a crash, but it should be handled.
[...]

PR #5459, opened 2025-06-26
FailFast when WS message were not proceed properly

TimoPtr:
## Summary

We want to investigate a bit more why sometimes the `messageQueue` is not able to proceed with a message. To do so I've used the `FailFast` API instead of a simple log and I've added more information to the message.

## Checklist

- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [ ] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

PR #5477, opened 2025-07-02
Use SupervisedJob in WebSocketCoreImpl to avoid cancelation of the scope

TimoPtr:
## Summary

With the new logs introduced in #5459, we found out that the issue #5472 seems to be because a coroutine cancel the whole `wsScope`. It is due to the fact that we use a `Job`, which means that if one `Job` is canceled within the scope it cancel the whole scope. To avoid this behavior this PR change the `Job` to `SupervisedJob` in the `wsScope`.

## Checklist


dshokouhi (approved):
Thanks for the fix :)

Issue #1511, opened 2021-04-14
F-Droid: Geolocation


newhinton:
**Is your feature request related to a problem? Please describe.**
Geolocation is not available for F-Droid. See #1335 

**Describe the solution you'd like**
I found a newer library: Its last commit was from late 2020, so it seems active enough. Also, the implementation seems rather straitforward, implementing a sample application took less than 10 minutes.

**Describe alternatives you've considered, if any**
See #1335 

**Additional context**

dshokouhi:
Nice find! Will need to research this a bit.  Also this won't just benefit the minimal flavor if it works we will use it for all flavors!

gdt:
My impression is that UnifiedNLP is available on all the ROMs that don't have gmscore, and that it's straightforward to use.  The only issue is not insisting that the provider of location services be signed by Google; on many ROMs there isn't "signature spoofing".   As an example, OsmAnd (navigation) uses GPS and "network location" just fine on Lineage.   In my case "network" is a plugin called DejaVu, that keeps on-device records of wifi/cell locations, basically like querying a network database except it's local.

newhinton:
Id like to keep this open as this creates feature parity between stores.

bd4:
I just ran into this with latest stable HA and app from f-droid, on CalyxOS Pixel 3a. Would be great to have location support on de-googled phones like Calyx and /e/.

d-two:
Same here... My wife has a Huawei Honor 10 X without google apps and all stuff work except the Location :-(

mike-lloyd03:
Is it possible to implement something like [UnifiedNlp](https://github.com/microg/UnifiedNlp) or Mozilla's [MLS](https://location.services.mozilla.com/) on the minimal version?

gdt:
Basically it needs to use the AOSP location API, which has both GPS and network, instead of using the Play Services API.   Google's docs are buggy in that they tell people to use the Play APIs :-)

dshokouhi:
Part of the reason why we went with Play Services is that its very battery friendly and also very easy to implement, it is not just a simple switch of the API. 

Google does indeed recommend using play services API in this case as we can see from this link: 

https://developer.android.com/reference/android/location/package-summary

We are open source so anyone who wants to develop this feature is more than welcome to.

answer35:
[...]

PR #5478, opened 2025-07-02
Handle GPS week rollover

Pawik55:
## Checklist

## Any other notes

This pull request adds handling of the GPS week number rollover in the minimal flavour. The problem that was present is described in this issue comment:  https://github.com/home-assistant/android/issues/1511#issuecomment-2946265967

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
@@ -31,7 +31,7 @@ class GeocodeSensorManager : SensorManager {
         private const val SETTING_ACCURACY = "geocode_minimum_accuracy"
         const val SETTINGS_INCLUDE_LOCATION = "geocode_include_location_updates"
         private const val DEFAULT_MINIMUM_ACCURACY = 200
-        private val LOCATION_OUTDATED_THRESHOLD = 5.minutes
+        val LOCATION_OUTDATED_THRESHOLD = 5.minutes
 
         /**
          * Represents the duration of 1024 weeks, used to detect and correct the GPS week rollover bug.
--- a/app/src/main/kotlin/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
@@ -19,6 +19,7 @@ import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 import kotlin.coroutines.suspendCoroutine
 import kotlin.time.Clock
+import kotlin.time.Duration.Companion.days
 import kotlin.time.Duration.Companion.minutes
 import kotlin.time.ExperimentalTime
 import timber.log.Timber
@@ -30,6 +31,7 @@ class GeocodeSensorManager : SensorManager {
         const val SETTINGS_INCLUDE_LOCATION = "geocode_include_location_updates"
         private const val DEFAULT_MINIMUM_ACCURACY = 200
         val LOCATION_OUTDATED_THRESHOLD = 5.minutes
+        private val GPS_ROLLOVER_WEEKS1024 = (7 * 1024).days
         val geocodedLocation = SensorManager.BasicSensor(
             "geocoded_location",
             "sensor",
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
```diff
@@ -30,6 +30,7 @@ class GeocodeSensorManager : SensorManager {
         const val SETTINGS_INCLUDE_LOCATION = "geocode_include_location_updates"
         private const val DEFAULT_MINIMUM_ACCURACY = 200
         val LOCATION_OUTDATED_THRESHOLD = 5.minutes
+        val GPS_ROLLOVER_WEEKS1024 = (60 * 24 * 7 * 1024).minutes
```
TimoPtr:
```suggestion
        private val GPS_ROLLOVER_WEEKS1024 = (7 * 1024).days
```
Could you make the one above private too 🙏🏻

app/src/main/kotlin/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
```diff
@@ -99,8 +100,12 @@ class GeocodeSensorManager : SensorManager {
         }
 
         if (Clock.System.now() - location.instant() > LOCATION_OUTDATED_THRESHOLD) {
-            Timber.w("Skipping geocoded update due to old timestamp ${location.instant()}")
-            return
+            if ((Clock.System.now() - location.instant() >= GPS_ROLLOVER_WEEKS1024) && (Clock.System.now() - location.instant() < GPS_ROLLOVER_WEEKS1024 + LOCATION_OUTDATED_THRESHOLD)) {
```
TimoPtr: Extract the diff between `now` and `instant` in a variable

app/src/main/kotlin/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
```diff
@@ -99,8 +100,12 @@ class GeocodeSensorManager : SensorManager {
...
+                // Timestamp OK. GPS week 1024 rollover bug
```
TimoPtr: A an info log instead of a comment.

TimoPtr:
Thank you for your contribution 🙏🏻

TimoPtr:
@Pawik55 I've commit to your branch a suggestion. Lemme know what you think about it. The reason behind it is that I would like to keep the information about we are doing this check within the code.

I also wanted to avoid the nested `if`

jpelgrom (approved):

app/src/main/kotlin/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
```diff
@@ -120,6 +132,21 @@ class GeocodeSensorManager : SensorManager {
...
     }
 
+    /**
+     * Some GPS receivers produce wrong time data due to the 1024 weeks rollover bug.
+     * This function attempts to detect and correct the time difference if it's affected by this bug.
+     *
+     * @return The sanitized time difference, potentially corrected for the rollover bug.
+     */
+    private fun Duration.sanitizeGPSTimeDifference(): Duration {
+        return if (this >= GPS_ROLLOVER_WEEKS1024) {
```
dshokouhi: Maybe its the logic that slightly confuses me. But how does this code change protect against Google sending a real report that is a week or 2 or even a month older as those are valid? I see we have a greater than or equals to check but that means anything over than a week is considered right? Isnt this bug very precise in that it should be the EXACT number and not over it?

TimoPtr: Here we are checking if the time difference between now and the value we got is higher than 1024 weeks so almost 20years. This function is strictly checking this and if the difference is bigger we remove these 1024 weeks from the time difference to only keep the real difference.

If Google send us data that is 2 weeks old we won't end up into this check since it will be bellow 1024 weeks.

dshokouhi: thanks for clarifying :) just wanted to make sure we dont reintroduce the old teleporting issue which was when we introduced the time checks.

TimoPtr: I'm going to add some unit test 👍🏻 and I'm also going to move the geocode invocation after the time check. No need to make a call to the geocoder if the data is too old.

TimoPtr: @dshokouhi if you can double check the tests that I wrote it should cover what you've pointed out.

Pawik55: @TimoPtr Your changes are ok with me. 👍

PR #5482, opened 2025-07-03
Handle missing field exception in CompressedEntityState

TimoPtr:
## Summary

After migrating to Kotlinx serialization the initialization of the classes are more strict than before. So a field that doesn't have a default value needs to be present in the JSON while decoding. In this case it is possible that the field `attributes` is empty or null (for reference see iOS impl https://github.com/home-assistant/HAKit/pull/49/files#diff-becad62c1e040b15f3ebfefa6bb6d7bc25de151dcc3eddc1599d27d76bb11f69).

## Checklist

## Any other notes

I saw a `!!` while converting to an `Entity` and I changed it to `checkNotNull` for more readability if it ever happens.
I'm wondering if we actually should support `null` for the `state` I'm going to follow up with core on this.

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/entities/CompressedEntity.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/entities/CompressedEntity.kt
@@ -62,9 +62,9 @@ data class CompressedEntityState(
     fun toEntity(entityId: String): Entity {
         return Entity(
             entityId = entityId,
-            state = state!!,
+            state = checkNotNull(state) { "State must not be null" },
             attributes = attributes,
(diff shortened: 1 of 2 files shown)


jpelgrom:
For the test: wouldn't it make more sense to have a test for a state without attributes, like what caused the issue? `{}` is unrealistic.

TimoPtr:
> For the test: wouldn't it make more sense to have a test for a state without attributes, like what caused the issue? `{}` is unrealistic.

I made this test on purpose to show that today impl is very permissive even if the code behind is not. 

I will add a second one that replicates a bit more the issue. But actually we don't know if the message received was complete since it crashed before going further and as you can see the model is quite permissive.

dshokouhi (approved):
Dont have a way to directly reproduce this but given I see it almost every other day best to merge and see if the issue disappears. Thanks for the fix :)

PR #5483, opened 2025-07-03
Support strict intent matching

### LINKED ISSUE #5330: Enforce Intent matching
Starting with API 36 Google introduced a new security feature for Intent matching in the manifest https://developer.android.com/about/versions/16/behavior-changes-16#implementation

We should adjust our implementation to support this new feature.

TimoPtr:
## Summary

In order to prepare ourselves from the changes that Google is going to push in the new version of Android, but also detect potential issues. I've enable some flags in DEBUG from `StrictMode`.

I enabled Intent matching in the manifest from this issue https://github.com/home-assistant/android/issues/5330.

This flags only impacts the API 36, I did some testing and it seems to work properly.

## Checklist

Developer Documentation: home-assistant/developers.home-assistant#2719

## Any other notes

Enabling all the checks is making the app crashing from the beginning about the socket that does network call without being properly tags. I only picked some flags that I found useful.

Closes #5330

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -71,7 +71,8 @@
         android:networkSecurityConfig="@xml/network_security_config"
         android:enableOnBackInvokedCallback="true"
         android:localeConfig="@xml/locales_config"
-        tools:ignore="GoogleAppIndexingWarning"
+        android:intentMatchingFlags="enforceIntentFilter"
+        tools:ignore="GoogleAppIndexingWarning,UnusedAttribute"
         tools:targetApi="tiramisu">
 
         <!-- Start things like SensorWorker on device boot -->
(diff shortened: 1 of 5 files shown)


dshokouhi:
anything in particular we need to test with this? like do we need to look at last update sensor intents to see if thsoe still get picked up?

TimoPtr:
So until we change the targetSDK to 36 nothing will happen except if you use a DEBUG build where you might get crashes now if we are making mistakes.

Once we target API 36 or if you make a build with API 36 the system will block Intent that are not valid from reaching our app.

dshokouhi (approved):

jpelgrom:
- Instead of forcing strict mode in debug, would it maybe be a good idea to have this behind a flag like LeakCanary (so it can still be on by default in debug)? You may encounter an issue when developing and these flags force a crash, which you may not want to/are able to resolve immediately. You can always edit the code to disable it of course but a flag seems a bit cleaner.
- Why add `detectLeakedRegistrationObjects()` if LeakCanary is supposed to detect that as well?

TimoPtr:
I'm going to add the flag to disable it like leak canary and update the doc.

For the `registrationObjects` vs leak canary I'm not 100% sure leak canary catch this issues. I rather have a duplicate than missing one.

jpelgrom (approved):

Issue #5486, opened 2025-07-03
Quick Settings - Device Controls stopped working


wary-hermit:
**Home Assistant Android app version(s):**  2025.6.6-full

**Android version(s):** 14

**Device model(s):** Pixel 7

**Home Assistant version:** 2025.7.0

**Last working Home Assistant release (if known):** Don't remember

**Description of problem, include YAML if issue is related to notifications:**
Quick Settings -> Device Control tiles not working 
I can add/edit/remove controls, see current status (on/off for automations, percentage for lightbulbs), but trying to turn on/off/change brightness causes `Error, try again` error message. 

**Probably relevant logs with LogFox:**

```
1751585356.817 10248  2463  2463 D ControlsProviderLifecycleManager: subscribe ComponentInfo{io.homeassistant.companion.android/io.homeassistant.companion.android.controls.HaControlsProviderService} - [1.light.ikea_color, 1.light.ikea_warm, 1.automation.motion_alarm, 1.automation.turn_on_purifier]
1751585359.624 10248  2463  2463 D ControlsProviderLifecycleManager: onAction ComponentInfo{io.homeassistant.companion.android/io.homeassistant.companion.android.controls.HaControlsProviderService} - 1.light.ikea_color
1751585360.665 10248  2463  2463 D ControlsProviderLifecycleManager: onAction ComponentInfo{io.homeassistant.companion.android/io.homeassistant.companion.android.controls.HaControlsProviderService} - 1.light.ikea_warm
1751585360.832 10636  7681 11396 E HaControlsProviderService$performControlAction: io.homeassistant.companion.android.common.data.integration.IntegrationException: Error calling integration request call_service
1751585360.832 10636  7681 11396 E HaControlsProviderService$performControlAction: 	at io.homeassistant.companion.android.common.data.integration.impl.IntegrationRepositoryImpl.callAction(IntegrationRepositoryImpl.kt:310)
1751585360.832 10636  7681 11396 E HaControlsProviderService$performControlAction: 	at io.homeassistant.companion.android.common.data.integration.impl.IntegrationRepositoryImpl$callAction$1.invokeSuspend(Unknown Source:15)
1751585360.832 10636  7681 11396 E HaControlsProviderService$performControlAction: 	at io.homeassistant.companion.android.common.data.integration.impl.IntegrationRepositoryImpl$callAction$1.invokeSuspend(Unknown Source:15)
[...]

PR #5487, opened 2025-07-04
Improve error handling when using Response from Retrofit

TimoPtr:
## Summary

While looking at #5486 I found out that we were simply not logging the potential error that we get from the API when calling the webhook endpoint.

To illustrate the issue I made sure that a server return a 404 when calling the webhook. As you can see bellow it only says that an `IntegrationException` was thrown and nothing else. We don't know what caused this failure.
Screenshot 2025-07-04 at 11 07 46

It makes the debugging of the issue almost impossible because we don't know why it failed. This is meanly due to the fact that we were using `Response<ResponseBody>`. When doing so Retrofit doesn't throw an error when it is an HTTP error (only when it is a network error like DNS), it sends back a Result with the code and the error in it that we were just ignoring.

This PR address this issue for every places where we uses `Response<>` from Retrofit to make sure we properly display the error.

After this PR is merged the error would becomes
Screenshot 2025-07-04 at 11 13 00

## Checklist

## Any other notes

I took the liberty to also use named argument for `ServerSessionInfo`, because most of the parameters are strings and we might one day inverse the order of the parameters and it won't be catch at build time. I'm thinking about introducing a lint rule that would force the usage of named parameters when a type is used more than once in an invocation. It will make refactors safer.

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/authentication/impl/AuthenticationRepositoryImpl.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/authentication/impl/AuthenticationRepositoryImpl.kt
@@ -44,11 +44,11 @@ class AuthenticationRepositoryImpl @AssistedInject constructor(
             serverManager.updateServer(
                 server.copy(
                     session = ServerSessionInfo(
-                        it.accessToken,
-                        it.refreshToken!!,
-                        System.currentTimeMillis() / 1000 + it.expiresIn,
-                        it.tokenType,
-                        installId,
+                        accessToken = it.accessToken,
+                        refreshToken = it.refreshToken!!,
+                        tokenExpiration = System.currentTimeMillis() / 1000 + it.expiresIn,
+                        tokenType = it.tokenType,
+                        installId = installId,
                     ),
                 ),
             )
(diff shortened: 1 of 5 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/data/authentication/impl/AuthenticationRepositoryImpl.kt
```diff
@@ -173,7 +173,7 @@ class AuthenticationRepositoryImpl @AssistedInject constructor(
             ) {
                 revokeSession()
             }
-            throw AuthorizationException()
+            throw AuthorizationException("Fail to refresh token", it.code(), it.errorBody())
```
jpelgrom:
```suggestion
            throw AuthorizationException("Failed to refresh token", it.code(), it.errorBody())
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
```diff
@@ -797,6 +668,42 @@ class IntegrationRepositoryImpl @AssistedInject constructor(
         }
     }
 
+    private suspend fun ServerConnectionInfo.callWebhookOnUrls(
```
jpelgrom: Why did you make this an extension function on `ServerConnectionInfo`? I don't see a reason to tightly couple it instead of a normal function. You only save a tiny bit of code in line 677.

TimoPtr: I will change it to the Server. It is just to avoid repeating multiple times the same code that doesn't bring value. It makes it more concise. It will look like this 
```
server.callWebhookOnUrls(...)
```

Since the extension is fully private and it's a call on a server I don't mind the coupling. But happy to discuss. Indeed it was not very useful to tight it to `ServerConnectionInfo`

jpelgrom: I guess my point was: `private suspend fun callWebhookOnUrls` also works and doesn't  add a lot of additional code (where you added `connection` in your latest commit, it'd be `server.connection`)?

If it's private like this using extensions adds little value while it does create room for confusion IMO (where did the `connection` come from, if you don't look carefully at the function), but not a blocker.

TimoPtr: For me it was like reading in English, server.callWehbook, it will handle find the right url to call.

Issue #5502, opened 2025-07-07
Error in template in Widget template


Mincka:
**Home Assistant Android app version(s):**
2025.6.6-full

**Android version(s):**
15

**Device model(s):**
Pixel 7 Pro

**Home Assistant version:**
2025.7.1

**Last working Home Assistant release (if known):**
Unknown

**Description of problem, include YAML if issue is related to notifications:**
A template I previously used in "Template widget" is now invalid, I get "Error in template".

```
{%  set distance = distance('device_tracker.telephone_julien', 'device_tracker.telephone_morgane') | round(default=0) %} 
{% if distance == 0 %}
💕
{% else %}
{{ distance }}
{% endif %}
```

Valid alternatives:
```
{%  set distance = 0 | round(default=0) %} 
{% if distance == 0 %}
💕
{% else %}
{{ distance }}
{% endif %}
```
or
```
{%  set distance = distance('device_tracker.telephone_julien', 'device_tracker.telephone_morgane') | round(default=0) %} 
{% if distance == 0 %}
💕
{% else %}
{{ distance }}
{% endif %}.    <-- any character here
```

**Companion App Logs:**
```
07-07 12:27:41.805 31159 23363 E TemplateWidgetConfigureActivity$renderTemplateText: Exception while rendering template
07-07 12:27:41.805 31159 23363 E TemplateWidgetConfigureActivity$renderTemplateText: io.homeassistant.companion.android.common.data.integration.IntegrationException: kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 12: Expected quotation mark '"', but had '6' instead at path: $['template']
07-07 12:27:41.805 31159 23363 E TemplateWidgetConfigureActivity$renderTemplateText: JSON input: {"template":6}
07-07 12:27:41.805 31159 23363 E TemplateWidgetConfigureActivity$renderTemplateText: 	at io.homeassistant.companion.android.common.data.integration.impl.IntegrationRepositoryImpl.renderTemplate(IntegrationRepositoryImpl.kt:231)
07-07 12:27:41.805 31159 23363 E TemplateWidgetConfigureActivity$renderTemplateText: 	at io.homeassistant.companion.android.common.data.integration.impl.IntegrationRepositoryImpl$renderTemplate$1.invokeSuspend(Unknown Source:15)
[...]

PR #5506, opened 2025-07-08
Handle multiple returns types from renderTemplate

TimoPtr:
## Summary

Fixes https://github.com/home-assistant/android/issues/5502. The issue was introduced because Kotlin serialization is strict on the JSON is receiving and we were expecting the JSON strcuture to only contains strings by using a `Map<String,String>`. This PR is handling the case where the rendered template returns other value than a string and convert it to a valid string to be displayed.
Tests have been added to verified that the behavior is the one intended.

## Checklist

## Screenshots

image
The template is now properly render without adding an extra character to make the return type a string.


TimoPtr:
@Mincka if you can install the APK and test your use case to validate the behavior it would be nice.

Mincka:
> @Mincka if you can install the APK and test your use case to validate the behavior it would be nice.

Tested on this [build](https://github.com/home-assistant/android/actions/runs/16135918918) and I confirm it's fixed.
Thanks a lot!

jpelgrom (approved):

PR #5507, opened 2025-07-08
Support Edge to edge default mode in WebViews

### LINKED ISSUE #5332: Replace usage of deprecated attributes for status and navigation bars
Starting with the API 35 Google is deprecating and disabling some attributes to get/set status and navigation bar colors.

https://developer.android.com/about/versions/15/behavior-changes-15#deprecated-apis

We should migrate away from this API or make sure it is not used in the new android versions.

TimoPtr:
## Summary

Since the frontend is not ready I decided to split https://github.com/home-assistant/android/pull/5346 in two. This first PR is making sure that if edge to edge is enable (which is going to be the case in API 36) the UI of the WebViews is not broken but it won't be edge to edge until the frontend fully supports it https://github.com/home-assistant/frontend/pull/25566.

It is inspired from https://github.com/home-assistant/android/pull/5390 but adjusted for compose. 

## Checklist

## Screenshots

image
image
image
image

## Any other notes

Part of https://github.com/home-assistant/android/issues/5328
Closes #5332

Add a new method to `FailFast` API and skip well known issue in chromium client that trigger strict mode.


TimoPtr:
After this PR is merged we can move to https://github.com/home-assistant/android/pull/5387 and edge to edge support should be complete.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1223,16 +1231,17 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
             Timber.e("Cannot set status bar color. Skipping coloring...")
         }
         if (navigationBarColor != 0) {
-            window.navigationBarColor = navigationBarColor
+            this.navigationBarColor.value = Color(navigationBarColor)
         } else {
             Timber.e("Cannot set navigation bar color. Skipping coloring...")
         }
 
+        // Adjust the color of the font based on the color of the theme used in HA
```
jpelgrom: If you're adding this the comment on the line below can probably be removed?

jpelgrom:
This is a familiar issue: when enabling edge to edge, you also need to handle left/right insets or cutouts are going to cause issues like in the menu here (nothing was obscured, but could be worse on other devices):
image

(Sorry for the delay testing, I also wanted to change target SDK between 34 and 36 before commenting which required using my computer.)

TimoPtr:
> This is a familiar issue: when enabling edge to edge, you also need to handle left/right insets or cutouts are going to cause issues like in the menu here (nothing was obscured, but could be worse on other devices):
> image
> 
> (Sorry for the delay testing, I also wanted to change target SDK between 34 and 36 before commenting which required using my computer.)

Good catch. This one is going to be a bit tricky to handle, we need to a bar on the left and right and use the background color of the theme.

TimoPtr:
image
Left and Right insets are now handled too.

I've also fixed the bluring that was not applied to the statusbar.

jpelgrom:
Minor nitpicks, otherwise good: AuthenticationFragment/SafeHAWebView should probably use the same comment about frontend support or at least both include `TODO` for easy reference.

(OnboardingActivity is not yet using edge to edge but I assume that is intentional. I did force it for the AuthenticationFragment and that one looked good.)

PR #5510, opened 2025-07-08
Request PERMISSION_READ_HEALTH_DATA_IN_BACKGROUND for HealtConnectSensor

### LINKED ISSUE #5327: Use granular permissions for health sensors
Starting with API 36 Google spitted https://developer.android.com/reference/android/Manifest.permission#BODY_SENSORS into multiple more granular permissions. We need to handle this case when asking for permissions.

https://developer.android.com/about/versions/16/behavior-changes-16#health-fitness-permissions

TimoPtr:
## Summary

Following the recent changes from Android for API 36 we most probably (I did not manage to reproduce the issue I once had) need to request [READ_HEALTH_DATA_IN_BACKGROUND](https://developer.android.com/reference/android/health/connect/HealthPermissions#READ_HEALTH_DATA_IN_BACKGROUND) according to the [documentation](https://developer.android.com/about/versions/16/behavior-changes-16#health-fitness-permissions). This PR adds this permission to the app.

It does not check if the feature is available and I don't know how I could test the scenario where it is not available. If anyone have idea on how to test it. In any case if one of our users is impacted by this the fix is for him to update his Health Connect app from the Play Store.

I tested with the https://developer.android.com/health-and-fitness/guides/health-connect/test/health-connect-toolbox and I was able to retrieve my data with the app targeting API 36 in background.

⚠️ I think this will break existing setup because it will misses one permission. So users will need to go in the settings again and turn on the 

## Checklist

## Screenshots

Classic flow to enable a Health sensor with the new permission

https://github.com/user-attachments/assets/20fae462-4d2f-473b-8855-211c2fddabe6

Enabling back a Health sensor that was enabled on a previous version. It simply ask for the background permission and only once since this permission is shared for all the sensors.

https://github.com/user-attachments/assets/d2cfdae0-f671-446a-9579-312512165526

## Any other notes

I took the liberty to replace the `when` with a lookup map to split the function and make it faster and easier to read. Alongside with adding some `Failfast` usage to detect missing permission for a given sensorID.

Closes #5327


dshokouhi:
> ⚠️ I think this will break existing setup because it will misses one permission. So users will need to go in the settings again and turn on the

Users should be prompted to grant the permission upon upgrading to this version, they may get more than 1 notification too depending on how many health connect sensors they have enabled. This is unavoidable though given its a permission.

app/src/main/res/xml/changelog_master.xml
```diff
@@ -1,13 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <changelog xmlns:tools="http://schemas.android.com/tools"
     tools:ignore="MissingDefaultResource">
-    <release version="2025.7.1 - Main" versioncode="3">
+    <release version="2025.7.2 - Main" versioncode="3">
+        <change>Health sensors now requires read health data in background permission</change>
```
```suggestion
        <change>Health Connect sensors now require permission to access health data in the background</change>
```

Health Connect, match wording from background permission prompt

jpelgrom:
> It does not check if the feature is available and I don't know how I could test the scenario where it is not available.

Previously it was not available and we could read Health Connect data using the SensorWorker, so I doubt it is something we can realistically test. If there is a scenario where background reads stopped working, that would probably be a Health Connect update that includes this permission?

Issue #5344, opened 2025-05-21
Migrate SetttingsActivity to Jetpack compose


TimoPtr:
This is a huge piece and needs to be tackled into multiple small PRs.

The Settings screen is made of an activity that host multiple fragments we should migrate the UI to Jetpack compose instead of xml to align with the rest of the application.

**Requirements**
The screens should have edge-to-edge supports

KvRae:
Hey @TimoPtr, 
I want to contribute to this

TimoPtr:
Hey @KvRae anyone can take it, so feel free to do it :)

yashc18:
Hello, @TimoPtr  If No one is working on this, can it be assigned to me?

TimoPtr:
To be fair this one is huge, and should be done in multiple PRs. So yes you can take this but make small PRs.

yashc18:
Sure! Let me see it through first!

PR #5512, opened 2025-07-08
Make gestures configurable (+ breaking: swap left/right)

jpelgrom:
## Summary

This PR makes it possible to configure the gestures in the WebViewActivity as desired. Note that the aim of this PR is to make the _existing_ gesture functionality configurable, but you get 2 finger swipes in this PR as well as that was barely any more code. I've stayed pretty close to existing Compose settings instead of overhauling everything for a fragment-less settings (#5344).

**Breaking change**: the default for three finger swipe left/right have been swapped to match the [defaults](https://github.com/home-assistant/iOS/blob/69f91142d48e57e09c42d75c00afd68f7b4506ae/Sources/Shared/AppGesture.swift#L195) on iOS. If you prefer how it was you can now change it back 😉

## Checklist

## Screenshots

|     |Light|Dark|
|----|----|----|
|New setting in main list|Settings row 'Gestures' with a squiggle icon below the title 'Other settings', light mode|Settings row 'Gestures' with a squiggle icon below the title 'Other settings', dark mode|
|Gesture settings screen*|Settings screen with options to configure swipes to the left, right, up and down with two and three fingers, light mode|Settings screen with options to configure swipes to the left, right, up and down with two and three fingers, dark mode|

\* The description here includes a generic note about system-wide gestures overriding HA gestures, similar to what is currently in the documentation (main example: OnePlus screenshots).

I intend to update docs with the next PR (see below), describing Android and iOS as one and dropping the current Android-specific gestures section.

## Any other notes

I intend to follow up after this is accepted with another PR to add more _actions_, which will add what is available in the iOS app (different quick bars / Assist / app settings). I'm not planning to add more gestures or add more defaults from iOS: shake/single finger swipes are uncommon or conflict with system gestures, and don't make sense as defaults on Android.

### DIFF
--- a/app/src/screenshotTest/kotlin/io/homeassistant/companion/android/settings/views/SettingsRowScreenshotTest.kt
+++ b/app/src/screenshotTest/kotlin/io/homeassistant/companion/android/settings/views/SettingsRowScreenshotTest.kt
@@ -30,9 +30,11 @@ class SettingsRowScreenshotTest {
             primaryText = "Title",
             secondaryText = "Subtitle",
             icon = {
-                Spacer(Modifier
-                    .border(width = 1.dp, color = Color.Magenta)
-                    .size(width = 56.dp, height = 16.dp))
+                Spacer(
+                    Modifier
+                        .border(width = 1.dp, color = Color.Magenta)
+                        .size(width = 56.dp, height = 16.dp),
+                )
             },
             onClicked = {},
         )


bgoncal:
Looking good! What are the actions that Android supports at this moment? Can you share a screenshot from the actions list as well?

jpelgrom:
> Looking good! What are the actions that Android supports at this moment? Can you share a screenshot from the actions list as well?

Search entities/server list/next server/previous server. As mentioned in the description, that is what was already supported and I've intentionally not expanded that list in this PR.

bgoncal:
Sounds good, if this list expands in the future consider not using drop down UI element and instead a full page to pick the action.

But for now it's great, thanks for adding.

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/GesturesViewModel.kt
```diff
@@ -0,0 +1,25 @@
...
+import io.homeassistant.companion.android.common.util.GestureAction
+import io.homeassistant.companion.android.common.util.HAGesture
+import javax.inject.Inject
+import kotlinx.coroutines.runBlocking
+
+@HiltViewModel
+class GesturesViewModel @Inject constructor(
+    private val prefsRepository: PrefsRepository,
+    application: Application,
+) : AndroidViewModel(application) {
```
TimoPtr:
```suggestion
) : ViewModel() {
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/GesturesViewModel.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    fun getGestureAction(gesture: HAGesture) = runBlocking {
+        prefsRepository.getGestureAction(gesture)
+    }
+
+    fun setGestureAction(gesture: HAGesture, action: GestureAction) = runBlocking {
+        prefsRepository.setGestureAction(gesture, action)
+    }
```
TimoPtr: Any reason to not be async here?

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesView.kt
```diff
@@ -0,0 +1,114 @@
...
+                    getAction = { onGetAction(gesture) },
+                    onSetAction = { action -> onSetAction(gesture, action) },
+                )
+            }
+        }
+    }
+}
+
+@Composable
+fun GestureSettingRow(
```
TimoPtr:
```suggestion
private fun GestureSettingRow(
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesView.kt
```diff
@@ -0,0 +1,114 @@
...
+            Text(
+                text = stringResource(R.string.gestures_description),
+                modifier = Modifier.padding(horizontal = 16.dp),
+            )
+            Divider(
+                modifier = Modifier.padding(all = 16.dp),
+            )
+        }
+
+        val gesturesGrouped = HAGesture.entries.groupBy { it.direction }
```
TimoPtr, line 48:
```suggestion
        val gesturesGrouped = HAGesture.entries.groupBy(HAGesture::direction)
```
Not mandatory I often find it more readable than using `it`.

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesView.kt
```diff
@@ -0,0 +1,114 @@
...
+    LazyColumn(
+        contentPadding = PaddingValues(vertical = 16.dp) + safeBottomPaddingValues(applyHorizontal = false),
+    ) {
+        item {
+            Text(
+                text = stringResource(R.string.gestures_description),
+                modifier = Modifier.padding(horizontal = 16.dp),
+            )
+            Divider(
+                modifier = Modifier.padding(all = 16.dp),
```
TimoPtr, line 44: You already have a bottom padding of 16.dp with the `contentPadding` so the padding between the divider and next item is 32.dp was it on purpose?

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesView.kt
```diff
@@ -0,0 +1,114 @@
...
+    var currentAction by remember(gesture.name) { mutableStateOf(getAction()) }
+    var expanded by remember { mutableStateOf(false) }
+
+    Box {
+        SettingsRow(
+            primaryText = when (gesture.pointers) {
+                2 -> stringResource(R.string.gestures_pointers_two)
+                3 -> stringResource(R.string.gestures_pointers_three)
+                else -> {
+                    FailFast.fail { "Missing pointer count string for $gesture" }
```
TimoPtr: I really like that you are using FailFast and it is indeed a relevant usage. But here I would be in favor to replace the type of pointers by an enum to leverage the `when` builtin check that will break if one day we add 4 fingers for instance. Otherwise it will break only when we would open the Gesture screen.

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesView.kt
```diff
@@ -0,0 +1,114 @@
...
+            },
+            secondaryText = stringResource(currentAction.description),
+            icon = null,
+            onClicked = { expanded = true },
+        )
+        DropdownMenu(
+            expanded = expanded,
+            onDismissRequest = { expanded = false },
+            // Match text padding
+            offset = DpOffset(x = 8.dp, y = 0.dp),
```
TimoPtr: Image
You want to align the text?

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesView.kt
```diff
@@ -0,0 +1,114 @@
...
+                    Text(stringResource(action.description))
+                }
+            }
+        }
+    }
+}
+
+@Preview
+@Composable
+fun PreviewGesturesView() {
```
TimoPtr:
```suggestion
private fun PreviewGesturesView() {
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/GesturesViewModel.kt
TimoPtr: We should as much as possible avoid relying on using Application/Context in VM IMO

app/src/main/kotlin/io/homeassistant/companion/android/settings/sensor/views/SensorListView.kt
```diff
@@ -42,24 +36,15 @@ fun SensorListView(
...
-                            Text(
-                                text = stringResource(manager.name),
-                                style = MaterialTheme.typography.body2,
-                                fontWeight = FontWeight.Bold,
-                                color = MaterialTheme.colors.primary,
-                            )
-                        }
+                        SettingsSubheader(
+                            text = stringResource(manager.name),
+                            paddingForIcon = true,
```
TimoPtr: Maybe not relevant but 

Image

Don't we want the icon to be display and not being cut by `App sensors`?

TimoPtr:
```suggestion
                            modifier = Modifier.padding(start = 72.dp),
```
I guess this would respect more the way of creating reusable composable components. It would make the `SettingsSubheader` agnostic from the Icon.

TimoPtr: And you could probably remove the box if you gives the right modifier to the `SettingsSubheader`

app/src/main/kotlin/io/homeassistant/companion/android/settings/views/SettingsRow.kt
```diff
@@ -25,13 +25,55 @@ import androidx.compose.ui.unit.dp
...
+                    colorFilter = ColorFilter.tint(
+                        if (enabled) {
+                            MaterialTheme.colors.primary
+                        } else {
+                            contentColorFor(backgroundColor = MaterialTheme.colors.background)
+                        },
+                    ),
+                )
+            } else {
+                Spacer(modifier = Modifier.width(24.dp))
```
TimoPtr: extract 24.dp into a variable if one day we change the size of the icon it would need to be adjusted.

app/src/main/kotlin/io/homeassistant/companion/android/settings/views/SettingsSubheader.kt
```diff
@@ -0,0 +1,52 @@
...
+
+@Preview
+@Composable
+fun PreviewSettingsSubheadingDefault() {
+    SettingsSubheader("Attributes")
+}
+
+@Preview
+@Composable
+fun PreviewSettingsSubheadingWithPadding() {
```
TimoPtr:
```suggestion
private fun PreviewSettingsSubheadingDefault() {
    SettingsSubheader("Attributes")
}

@Preview
@Composable
private fun PreviewSettingsSubheadingWithPadding() {
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/views/SettingsRow.kt
```diff
@@ -25,13 +25,55 @@ import androidx.compose.ui.unit.dp
 import com.mikepenz.iconics.compose.Image
 import com.mikepenz.iconics.typeface.IIcon
 
+/**
```
TimoPtr, line 28: You could add screenshot tests here too.

app/src/main/kotlin/io/homeassistant/companion/android/settings/views/SettingsSubheader.kt
```diff
@@ -0,0 +1,52 @@
...
+    paddingForIcon: Boolean = false,
+) {
+    Row(
+        modifier = Modifier
+            .padding(top = 8.dp)
+            .heightIn(min = 40.dp)
+            .padding(start = if (paddingForIcon) 72.dp else 16.dp, end = 16.dp),
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        Text(
```
TimoPtr: Do we need a Row?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -302,28 +304,11 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
-                                        supportFragmentManager.clearFragmentResultListener(ServerChooserFragment.RESULT_KEY)
-                                    }
-                                    serverChooser.show(supportFragmentManager, ServerChooserFragment.TAG)
-                                }
-
-                                SwipeDirection.DOWN -> {
-                                    dispatchKeyEvent(KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_E))
-                                }
-                            }
+                        if (pointerCount > 1 && velocity >= 75 && !appLocked.value) {
```
TimoPtr, line 311: The behavior changed no? Before you were not checking the appLocked status.  It kinda makes more sense now. 

But the general logic is missing the case where the even is not consumed. We should return `false` in the `else` branch if I'm not mistaking.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -862,6 +847,33 @@ class WebViewActivity : BaseActivity(), io.homeassistant.companion.android.webvi
...
+                    serverChooser.show(supportFragmentManager, ServerChooserFragment.TAG)
+                }
+
+                GestureAction.QUICKBAR_DEFAULT -> {
+                    webView.dispatchKeyEvent(KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_E))
+                }
+
+                GestureAction.NONE -> {
+                    // Do nothing
+                }
```
TimoPtr, line 884: In that case should we consume the event or not?

app/src/main/res/xml/preferences_server.xml
```diff
@@ -8,7 +8,7 @@
         app:isPreferenceVisible="false"/>
     <Preference
         android:key="activate_server_hint"
-        android:summary="@string/server_activate_swipe_hint"
+        android:summary="@string/server_activate_gesture_hint"
```
TimoPtr, line 11: It's going to break all the existing translations for this key no?

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/settings/views/SettingsRowScreenshotTest.kt
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Preview
+    @Composable
+    fun `SettingsRow with custom icon composable`() {
+        SettingsRow(
+            primaryText = "Title",
+            secondaryText = "Subtitle",
+            icon = {
+                Spacer(Modifier.size(width = 56.dp, height = 24.dp))
```
TimoPtr: Not really an icon ^^

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
```diff
@@ -260,6 +263,26 @@ class PrefsRepositoryImpl @Inject constructor(
...
+            action != null -> action
+            // Defaults
+            gesture == HAGesture.SWIPE_UP_THREE -> GestureAction.SERVER_LIST
+            gesture == HAGesture.SWIPE_DOWN_THREE -> GestureAction.QUICKBAR_DEFAULT
+            gesture == HAGesture.SWIPE_LEFT_THREE -> GestureAction.SERVER_PREVIOUS
+            gesture == HAGesture.SWIPE_RIGHT_THREE -> GestureAction.SERVER_NEXT
+            // No user preference and not default
+            else -> GestureAction.NONE
+        }
+    }
```
TimoPtr, line 282: We could write a simple unit test for this one. The goal would be to avoid breaking this logic by mistake while refactoring, so I suggest to not use the const variables in the test but hardcoded strings.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/Gestures.kt
```diff
@@ -0,0 +1,69 @@
+package io.homeassistant.companion.android.common.util
+
+import io.homeassistant.companion.android.common.R
+import kotlin.math.abs
+
+enum class GestureAction(val description: Int) {
```
TimoPtr:
```suggestion
enum class GestureAction(@StringRes val description: Int) {
```

common/src/main/kotlin/io/homeassistant/companion/android/common/util/Gestures.kt
```diff
@@ -0,0 +1,69 @@
...
+    NONE(R.string.none),
+    QUICKBAR_DEFAULT(R.string.gestures_action_quickbar_default),
+    SERVER_LIST(R.string.gestures_action_server_list),
+    SERVER_NEXT(R.string.gestures_action_server_next),
+    SERVER_PREVIOUS(R.string.gestures_action_server_previous),
+}
+
+enum class GestureDirection(val description: Int) {
```
TimoPtr:
```suggestion
enum class GestureDirection(@StringRes val description: Int) {
```

common/src/main/kotlin/io/homeassistant/companion/android/common/util/Gestures.kt
```diff
@@ -0,0 +1,69 @@
...
+    SWIPE_UP_TWO(GestureDirection.UP, 2),
+    SWIPE_UP_THREE(GestureDirection.UP, 3),
+    SWIPE_DOWN_TWO(GestureDirection.DOWN, 2),
+    SWIPE_DOWN_THREE(GestureDirection.DOWN, 3),
+    ;
+
+    companion object {
+        fun fromSwipeListener(direction: GestureDirection, pointers: Int): HAGesture? =
+            entries.firstOrNull { it.direction == direction && it.pointers == pointers }
+    }
```
TimoPtr: We could FailFast here

TimoPtr (changes requested):
Thank you very much for taking this feature. We finally catch up with iOS 💪🏻

jpelgrom:
> Sounds good, if this list expands in the future consider not using drop down UI element and instead a full page to pick the action.

I think a large dialog is a more common pattern on Android. But definitely not using a dropdown if a lot more/different actions are added.

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesView.kt
jpelgrom: `it` is more obvious to me, like reading a sentence...

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesView.kt
jpelgrom: `contentPadding` is before the first item and after the last item, not between the divider and first subheader which are in the middle.

There is slightly more padding between the intro text/divider and first subheader than other subheaders, which is on purpose. (This pattern in general is somewhat inconsistently styled, and I tried to match one of them, but that shouldn't be addressed in this PR.)

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesView.kt
jpelgrom: Outdated comment... In the end I tried to avoid the dropdown menu hitting the left screen edge which looked bad, not align text.

app/src/main/kotlin/io/homeassistant/companion/android/settings/sensor/views/SensorListView.kt
jpelgrom: Not related indeed

The header is sticky, wouldn't it be weird to have the icon scroll behind but the row title/subtitle obscured?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: We never consume the event / the touch is always also passed to the WebView if unlocked (line 313). This behavior is unchanged.

app/src/main/kotlin/io/homeassistant/companion/android/settings/views/SettingsRow.kt
jpelgrom: In this file or ...? I already added screenshot tests for the settings row. Don't they need to be in debug to avoid shipping them?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: The behavior technically changed because I'm quitting early if the app is locked. However, if that is the case the blur should be on top blocking the touch anyway so I don't expect this case to happen in real life.

Regarding consuming or not, let's keep that discussion limited to https://github.com/home-assistant/android/pull/5512#discussion_r2194735168

app/src/main/res/xml/preferences_server.xml
jpelgrom: Yes, that is intentional to avoid other languages still having the string with three finger swipe (because of outdated translations), which may be incorrect.

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/settings/views/SettingsRowScreenshotTest.kt
jpelgrom: It doesn't need to be an icon, it is a Composable for a reason and this makes the test faster than rendering an icon.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/Gestures.kt
jpelgrom: That would be annoying. If you accidentally swipe with 4 fingers on the WebView the app would crash. A swipe (which can be anything detected by Android) without a gesture (supported by the app) is valid.

app/src/main/kotlin/io/homeassistant/companion/android/settings/sensor/views/SensorListView.kt
jpelgrom: > I guess this would respect more the way of creating reusable composable components. It would make the SettingsSubheader agnostic from the Icon.

While reusing is nice, this is supposed to be quite a specific component and the current parameter is much more descriptive than you having to provide 16/72 dp. We're not building a toolkit where everything has to be customizable.

I've adjusted it to provide a modifier slot, but also added a parameter for the text padding with named options (similar to something like `ButtonDefaults` in Compose).

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/GesturesViewModel.kt
jpelgrom: (Copying patterns from other settings screens with Compose.) I've adjusted it, theoretically this means the settings may show up as blank for a split second when loading but in reality that won't really happen.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
jpelgrom: I've added tests which I think cover what you're asking for, but I do not yet have a good feeling for writing these tests so please check them and let me know if you had something else in mind.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: If I'm not wrong we are sending the value of `return appLocked.value` which btw if you know why I would like a small comment in the code about it 🙏🏻.

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/settings/views/SettingsRowScreenshotTest.kt
TimoPtr: I think that for the sake of the screenshot visualization it would be nice to add a background to it, like a red one.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/Gestures.kt
TimoPtr: Make sense maybe a debug log could help to debug still.

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/GesturesViewModel.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    val gestureActions = mutableStateMapOf<HAGesture, GestureAction>()
+
+    init {
+        viewModelScope.launch {
+            HAGesture.entries.forEach {
+                gestureActions[it] = prefsRepository.getGestureAction(it)
+            }
+        }
+    }
```
TimoPtr, line 26: I'm not sure if we want to tackle this here but we should be sure that getGestureAction internal IO call to the storage is made on IO thread. Right now even if it async `viewModelScope` by default use the main thread group.

TimoPtr: It is probably a bit more complex than I though since the underlying is a SharedPref https://www.reddit.com/r/androiddev/comments/12lve85/why_is_disk_io_on_the_main_thread_using/

jpelgrom: That is out of scope for this PR in my opinion and should be tackled in the LocalStorage interface then.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
TimoPtr: Looks perfect thanks 👍🏻, it will be very useful if one day we decide to change a name it will automatically detect that it will break existing configuration.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/Gestures.kt
jpelgrom: I would rather not; limit noise in the logs and log only when it is something that we can use.

PR #5516, opened 2025-07-09
Apply bottom inset to snackbar in SettingsFragment

TimoPtr:
## Summary

I only noticed the issue because I was testing on API 21 another PR. The snackbar was display behind the navigation bar when adding a new server. 
image
It happens because even if we apply the insets properly on the `listView` the snackbar is using the underlying `CoordinatorLayout` where we don't want to apply the insets. So to fix it I simply add a new listener to get the exact bottom inset value and apply it to the snackbar before displaying it.

## Checklist

## Screenshots

image

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
@@ -13,10 +13,13 @@ import android.os.Bundle
 import android.os.PowerManager
 import android.provider.Settings
 import android.view.View
+import android.view.ViewGroup
 import androidx.activity.result.contract.ActivityResultContracts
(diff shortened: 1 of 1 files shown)


jpelgrom:
Your second screenshot looks like it has a double inset - it should float just above the navigation bar, now it is 2x navigation bar height? Also all list items look wrong.

TimoPtr:
> Your second screenshot looks like it has a double inset - it should float just above the navigation bar, now it is 2x navigation bar height? Also all list items look wrong.

Thanks for this comment. It reminded me that we were consuming the Insets and with the new listener that I added we were not consuming the inset anymore hence why it was applied twice. So the real fix is just about about setting the proper flag to false and it works as expected. 
image
Now it works as expected 🎉

PR #5517, opened 2025-07-09
Make SensorDAO methods suspensible

TimoPtr:
## Summary

After enabling `StrictMode` in #5483 I found out that we are doing many db calls on the main thread where we shouldn't. It most probably causes some ANR. This PR is the first of many PRs regarding the `SensorDAO` all the DAO methods should be marked as `suspend`.

Note: This PR only add the suspend keyword and move some call to IO but it does not covers all the changes that are needed to make sure no one is invoking any method from the MainThread (that should be tackle in a second PR).

This issue won't happen again with https://github.com/home-assistant/android/pull/5526

Part of https://github.com/home-assistant/android/issues/5520

## Checklist

## Any other notes

This will probably causes some new bugs since many things were running on the main thread and now some have been moved to IO.


dshokouhi:
> This will probably causes some new bugs since many things were running on the main thread and now some have been moved to IO.

That may be difficult to find as we probably will need to have all sensors enabled and keep it running for a couple days. This sounds like something we will want to soak in beta for a week or 2.

With that said the PR does not have any issues when I run it and sensors still update.

TimoPtr:
> > This will probably causes some new bugs since many things were running on the main thread and now some have been moved to IO.
> 
> That may be difficult to find as we probably will need to have all sensors enabled and keep it running for a couple days. This sounds like something we will want to soak in beta for a week or 2.
> 
> With that said the PR does not have any issues when I run it and sensors still update.

Agreed, btw I was thinking about making in the DevPlaygound or even in the sensor list a way to trigger enable all sensors in one shot (that will ask all permissions one by one)

dshokouhi:
> even in the sensor list a way to trigger enable all sensors in one shot (that will ask all permissions one by one)

we used to have that, but removed it because many users were enabling it and draining their battery if not setup properly so we removed it https://github.com/home-assistant/android/pull/2037

TimoPtr:
> > even in the sensor list a way to trigger enable all sensors in one shot (that will ask all permissions one by one)
> 
> we used to have that, but removed it because many users were enabling it and draining their battery if not setup properly so we removed it #2037

Nice I'll enable it only in debug build in that case what do you think?

dshokouhi:
> > > even in the sensor list a way to trigger enable all sensors in one shot (that will ask all permissions one by one)
> > 
> > 
> > we used to have that, but removed it because many users were enabling it and draining their battery if not setup properly so we removed it #2037
> 
> Nice I'll enable it only in debug build in that case what do you think?

Sure sounds good to me :) anything to make testing easier when we need to test all sensors in one go lol

I would love to have it back in the prod app but not sure how we can handle it in a way that users wont turn it on by default. Earlier obviously the implementation was not the best since it was the first thing you saw on the sensor screen 😅 However with sane settings the drain is not horrible and often at times I'll have 40% or more battery end of day 🤷

TimoPtr:
> > > > even in the sensor list a way to trigger enable all sensors in one shot (that will ask all permissions one by one)
> > > 
> > > 
> > > we used to have that, but removed it because many users were enabling it and draining their battery if not setup properly so we removed it #2037
> > 
> > 
> > Nice I'll enable it only in debug build in that case what do you think?
> 
> Sure sounds good to me :) anything to make testing easier when we need to test all sensors in one go lol
> 
> I would love to have it back in the prod app but not sure how we can handle it in a way that users wont turn it on by default. Earlier obviously the implementation was not the best since it was the first thing you saw on the sensor screen 😅 However with sane settings the drain is not horrible and often at times I'll have 40% or more battery end of day 🤷

 Maybe we could but it a bit hidden like a long press on something.

jpelgrom:
The easiest option to enable all right now is
- grant all permissions to the app in advance using system settings
- bulk enable on the server
- the app will enable everything the next time sensors sync

This isn't that much work but still allows a lot faster enabling than going one by one in the app interface.

TimoPtr:
> The easiest option to enable all right now is
> - grant all permissions to the app in advance using system settings
> - bulk enable on the server
> - the app will enable everything the next time sensors sync
> 
> This isn't that much work but still allows a lot faster enabling than going one by one in the app interface.

Let's do something that we can use in one click or in one shell command that we put in the doc.

dshokouhi (approved):

app/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -269,7 +271,7 @@ class MessagingManager @Inject constructor(
         const val THIS_SERVER_ID = "server_id"
     }
 
-    private val mainScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
+    private val ioScope: CoroutineScope = CoroutineScope(Dispatchers.IO + Job())
```
TimoPtr: I'm not sure about this change. Also the `handleMessage` is so big ... I actually wonder if it would make sense to just make the whole thing running directly within the mainScope

TimoPtr: @jpelgrom do you have any idea here?

dshokouhi: > I'm not sure about this change. Also the `handleMessage` is so big ... I actually wonder if it would make sense to just make the whole thing running directly within the mainScope

I think that makes sense, this method has so much going on that its hard to escape from. its really the only way we can control the phone from the server

jpelgrom: What do I need to have an idea about? Making it a coroutine function sounds like a nice idea. Some of these actions may need to be running on the UI/main thread, like activity launches, so whatever we do we should test it.

TimoPtr: I've rollback to Main dispatcher and move the whole execution in the coroutine.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/SensorManager.kt
```diff
@@ -38,6 +38,7 @@ interface SensorManager {
 
     val name: Int
 
+    // TODO any reason to use mainScope here and not iO?
```
jpelgrom: I do not have a good answer. The history seems to be: a pattern found in a few sensors was put here when needed in more places, to make it common/less code. So adding new scopes separately in each sensor like you're doing kind of contradicts that, maybe add a `sensorWorkerIoScope` as well? Considering the amount of changes in this PR I wouldn't also change the existing scope.

TimoPtr: I'm not really comfortable with this global scopes not being tight to any lifecycle and them using `Job` instead of `SupervisedJob`, if for some reason an exception is raised or `cancel` is invoked on the scope then the scope is not usable anymore like what happened in the WebSocket recently.

I'm going to create an issue to follow up on this and merge this PR.

TimoPtr: https://github.com/home-assistant/android/issues/5585

PR #5527, opened 2025-07-10
Make SensorDAO.add/update suspendable

TimoPtr:
## Summary

This PR is the last step to make the whole SensorDAO suspendable before checking that we are doing DB operation on the IO thread and not the main thread.


dshokouhi:
Not sure if related but while testing the debug APK found this leak: https://github.com/home-assistant/android/issues/5546 the rest of the chagnes look good to me. I am unsure if the leak is associated with this or not. Assuming not but safer to ask :)

PR #5528, opened 2025-07-10
Make settingsDao fully suspensible

TimoPtr:
## Summary

SettingsDao was not properly made, all the methods should have been suspensible to make sure the call are not blocking the main thread if they are called from the main thread.

## Checklist


jpelgrom:
> // TODO why 0 ????

This database table has columns for both app-level settings and server-level settings, so as a compromise it is using row with 'server ID' `0` for the app-level settings and rows with real server IDs for the server-level settings.

TimoPtr:
> > // TODO why 0 ????
> 
> This database table has columns for both app-level settings and server-level settings, so as a compromise it is using row with 'server ID' `0` for the app-level settings and rows with real server IDs for the server-level settings.

It probably make sense to split this in two if we use 0 as app-level settings.

PR #5538, opened 2025-07-13
More actions to configure for gestures

jpelgrom:
## Summary

This PR brings in the remaining gesture actions [from iOS](https://github.com/home-assistant/iOS/blob/69f91142d48e57e09c42d75c00afd68f7b4506ae/Sources/Shared/AppGesture.swift#L28-L45) as options for gestures in the app, but:

 - Excluding page back because we have a system button/gesture for that
 - Added action for go to default dashboard (suggestion in comment)
 - Added action for reload page (has been mentioned a couple of times in the past, like a pull to refresh, and this feels like a good place to add it without making it required and makes sense next to page forward)

## Checklist

## Screenshots

|Light|Dark|
|----|----|
|Screen with gesture actions, a very long list, longer than a normal phone screen, light mode|Screen with gesture actions, a very long list, longer than a normal phone screen, dark mode|

User Documentation: home-assistant/companion.home-assistant#1224


victorigualada:
Awesome! Would be great to have a an option to return to the default dashboard. As I find myself hitting the back button or the back gesture multiple times to reach it.

jpelgrom:
> Would be great to have a an option to return to the default dashboard.

Thanks for the suggestion! Added as we already have similar functionality with a different trigger.

https://github.com/user-attachments/assets/31de8340-4d2e-4b63-9bdf-1d3b2cce10f7

jpelgrom:
FYI, UI changes mentioned in the description have been split out into #5578. Once that is approved I'll update this PR, which will only contain the new actions and related info.

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/settings/gestures/GesturesFragmentScreenshotTest.kt
```diff
@@ -0,0 +1,29 @@
...
+import io.homeassistant.companion.android.common.util.HAGesture
+import io.homeassistant.companion.android.settings.gestures.views.GestureActionsView
+import io.homeassistant.companion.android.settings.gestures.views.GesturesListView
+
+class GesturesFragmentScreenshotTest {
+
+    @Preview
+    @Composable
+    fun `Gestures list with no action for each gesture`() {
+        GesturesListView(
```
TimoPtr: Could you use the HA theme ? Otherwise we are not using the right colors.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsActivity.kt
```diff
@@ -42,8 +43,16 @@ class SettingsActivity : BaseActivity() {
     private var externalAuthCallback: ((Int) -> Boolean)? = null
 
     companion object {
-        fun newInstance(context: Context): Intent {
-            return Intent(context, SettingsActivity::class.java)
+        const val SCREEN_DEVELOPER = "developer"
+        const val SCREEN_NOTIFICATION_HISTORY = "notification_history"
+        const val SCREEN_WEBSOCKET = "websocket"
+
+        fun newInstance(context: Context, screen: String? = null): Intent {
```
TimoPtr: Make an enum out of the potential screens.

TimoPtr (approved):
Minor things to address but overall all good.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsActivity.kt
jpelgrom: Adjusted, even though this significantly increases the number of changes not strictly related to the feature.

jpelgrom: Adjusted, even though this significantly increases the number of changes not strictly related to the feature.

@TimoPtr Could you check if this looks good to you? (if so and no new comments it can be merged, otherwise I'll address them on Saturday)

TimoPtr: All good thanks 👍🏻. In the end we could have use a sealed class to handle the parameters of the deep link but it's good enough to expose the API of the fragment.

PR #5526, opened 2025-07-10
Introduce Coroutine Dao functions lint rule

TimoPtr:
## Summary

In order to avoid such situation in the future https://github.com/home-assistant/android/pull/5517 this PR makes sure that all the function within a Room DAO are using `suspend` or `Flow`.

## Checklist

- [x] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

## Screenshots

image

## Any other notes

For now I've ignored all the issue in `:common` but need to tackle then one by one.

Issue #5520, opened 2025-07-09
DB calls read/write on mainThread


TimoPtr:
We have a lot of calls that are made from the mainThread we should not do this. When you open the app in debug you can see many warnings from StrictMode about it.

You can take one of them and try to fix it.

jpelgrom:
I'm a bit confused - Room famously does not allow using the database on the main thread, and Android will crash by default if you do network access on the main thread. Do you mean that we should try to move more work around it to a different thread? Is that really necessary in all cases where we already use suspending functions, so that should prevent blocking the UI?

TimoPtr:
If you start the app in debug from main. You can see all the strict mode warning about room making a violation call to the database from our application object. Once migrated to suspend function the issue is gone. I cannot check right now but I do think it's because when a function is not suspendable it is behind the scene a run locking operation. I don't think it would crash the system like the network calls. 

The title is probably wrong about network call it's storage access read and write. I'm going to fix it.

PR #5545, opened 2025-07-15
Make ServerDao suspensible

TimoPtr:
## Summary

After enabling `StrictMode` in #5483 I found out that we are doing many db calls on the main thread where we shouldn't. It most probably causes some ANR. This PR is the first of many PRs regarding the `ServerDAO` all the DAO methods should be marked as `suspend`, to not make a giant PR this one focus mostly on the `get` function and the needed adjustments. 

Note: This PR only add the suspend keyword and move some call to IO but it does not covers all the changes that are needed to make sure no one is invoking `get` from the MainThread (that should be tackle in a second PR).

This issue won't happen again with https://github.com/home-assistant/android/pull/5526

Part of https://github.com/home-assistant/android/issues/5520

## Checklist

## Any other notes

This will probably causes some new bugs since many things were running on the main thread and now some have been moved to IO.

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManager.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManager.kt
@@ -89,18 +89,3 @@ interface ServerManager {
      */
     fun webSocketRepository(serverId: Int = SERVER_ID_ACTIVE): WebSocketRepository
 }
-
-/**
- * Helper class to retrieve a [Server] using [ServerManager.getServer]. This class will cache the
- * server after the first retrieval. Note that this class does **not** observe any server changes.
- *
- * @property serverManager the [ServerManager] instance to use
- * @property serverId the ID of the server to retrieve
- */
-class ServerRetriever(val serverManager: ServerManager, val serverId: Int) {
-    private var sever: Server? = null
-
-    suspend operator fun invoke(): Server {
-        return sever ?: checkNotNull(serverManager.getServer(serverId)) { "No server found for id $serverId" }
-    }
-}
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/settings/server/ServerSettingsPresenterImpl.kt
```diff
@@ -128,15 +128,21 @@ class ServerSettingsPresenterImpl @Inject constructor(
...
 
     override fun onFinish() {
-        if (serverManager.getServer()?.id != serverId) {
-            setAppActive(false)
+        runBlocking {
+            if (serverManager.getServer()?.id != serverId) {
+                setAppActive(false)
+            }
         }
         mainScope.cancel()
```
TimoPtr, line 136: If we were using a ViewModel we should use the VMScope but here we can't for now I decided to make it like it was before.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManager.kt
```diff
@@ -89,3 +89,18 @@ interface ServerManager {
      */
     fun webSocketRepository(serverId: Int = SERVER_ID_ACTIVE): WebSocketRepository
 }
+
+/**
+ * Helper class to retrieve a [Server] using [ServerManager.getServer]. This class will cache the
+ * server after the first retrieval. Note that this class does **not** observe any server changes.
```
jpelgrom: I did not immediately encounter issues during testing, but am concerned about caching the server without watching for changes. If something like the authentication token changes, it is not immediately reflected. Why did you change this, when most functions in repositories are suspending?

common/src/main/kotlin/io/homeassistant/companion/android/database/sensor/SensorDao.kt
```diff
@@ -72,6 +72,7 @@ interface SensorDao {
 
     @Transaction
     suspend fun removeServer(serverId: Int) {
+        // TODO improve the transaction to minimize the number of DB queries
```
jpelgrom: This does not belong in this PR, maybe in the sensor DAO PR.

TimoPtr: It was not meant to be committed 👍🏻

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManager.kt
TimoPtr: Good point I was trying to make something smart but in the end it is not 👍🏻 removing.

PR #5556, opened 2025-07-16
Bump target API to 36 🚀🚀🚀

TimoPtr:
## Summary

We are finally ready to target API 36, https://github.com/orgs/home-assistant/projects/23 is almost fully done. This PR should be merged only after https://github.com/home-assistant/android/pull/5387.

This change is needed for us to continue to publish application on the Play Store 
> Starting August 31 2025:
>
>    New apps and app updates must target Android 15 (API level 35) or higher to be submitted to Google Play; except for Wear OS, Android Automotive OS, and Android TV apps, which must target Android 14 (API level 34) or higher.
    Existing apps must target Android 14 (API level 34) or higher to remain available to new users on devices running Android OS higher than your app's target API level. Apps that target Android 13 (API level 33) or lower, including Android 12 (API level 31) or lower for Wear OS and Android TV, will only be available on devices running Android OS that are the same or lower than your app's target API level.


TimoPtr:
@jpelgrom I don't think it is necessary to add a dedicated entry in the changelog for the bump.

dshokouhi (approved):
Ran the debug APK for about a full day and did not encounter any issues in my usage

PR #5558, opened 2025-07-17
Open a PR to update changelog on each weekly workflow

TimoPtr:
## Summary

To avoid to forget to bump the version within the changelog file, this PR is updating the weekly workflow so that it automatically create a PR after each weekly workflow run with the changelog version field updated.

## Checklist

## Screenshots

image

image

Developer Documentation: home-assistant/developers.home-assistant#2729

## Any other notes

it won't work if we ever manually trigger a beta from main without using the weekly workflow. In that case the changelog won't be updated manually.


.github/workflows/weekly.yml
```diff
@@ -35,3 +35,44 @@ jobs:
...
+            github.rest.pulls.create({
+              title: `Bump changelog_master.xml for weekly release ${version}`,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              head: 'weekly_changelog_bump',
+              base: 'main',
+              body: [
+                'Automated PR to update changelog_master.xml for weekly release.',
+                'Double check if we didn't forget to update the changelog changes.',
+              ].join('\n'),
```
TimoPtr: Open for any suggestion on the content of the PR.

jpelgrom: > after each weekly workflow run 

With that timing, the text to check for if we forgot to mention anything doesn't make a lot of sense. We're already too late.

TimoPtr: Indeed removing this part.

.github/workflows/weekly.yml
```diff
@@ -35,3 +35,44 @@ jobs:
...
+      - name: Create changelog update branch
+        env:
+          GITHUB_TOKEN: ${{ secrets.TAG_PUSH_TOKEN }}
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git fetch origin
+          git checkout -B weekly_changelog_bump
+          git add app/src/main/res/xml/changelog_master.xml
+          git commit -m "Bump changelog_master.xml for weekly release"
```
jpelgrom: Would be nice to also include the version number here like the PR title

TimoPtr:
> This is a nice idea while we keep using the current changelog library.
> 
> However, could you maybe split it up in another workflow and trigger it when a new tag is created on the main branch? Smaller file + right now it will give you the wrong version number in weeks when the month changes (it'll just add .0.1 while it should change to match the month).

Do you have an idea for the issue at the end of the month? It's the same for the end of the year

jpelgrom:
The month/year change is done by pushing a tag (see the monthly workflow). That's why I suggested using a new tag on main as a trigger ;)

.github/workflows/weekly.yml
TimoPtr: I added it even if in the end this will never reach `main`since everything is squashed

PR #5563, opened 2025-07-17
Fix timestamp being in the future for GPS rollover bug

Pawik55:
## Summary

Fixes the case, when the GPS timestamp is a few seconds in the future and the GPS week rollover bug is present, being treated as too old because it is a few seconds earlier than 1024 weeks in the past.
See https://github.com/home-assistant/android/issues/1511#issuecomment-3075284598

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
@@ -25,6 +25,7 @@ import kotlin.time.Clock
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.days
 import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
 import kotlin.time.ExperimentalTime
 import timber.log.Timber
 
@@ -37,6 +38,13 @@ private const val DEFAULT_MINIMUM_ACCURACY = 200
  * For more information, see: https://github.com/home-assistant/android/issues/1511#issuecomment-2946265967
  */
 private val GPS_ROLLOVER_WEEKS1024 = (7 * 1024).days
+
+/**
+ * Sometimes the GPS timestamp is a few seconds in the future from the devices clock.
+ * To still catch these cases that are affected by this and the GPS rollover
+ * this value is subtracted from GPS_ROLLOVER_WEEKS1024.
+ */
+private val TIME_INACCURACY_FACTOR = 10.seconds
 private const val SETTING_ACCURACY = "geocode_minimum_accuracy"
 class GeocodeSensorManager : SensorManager {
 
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt
```diff
@@ -37,6 +38,13 @@ private const val DEFAULT_MINIMUM_ACCURACY = 200
  * For more information, see: https://github.com/home-assistant/android/issues/1511#issuecomment-2946265967
  */
 private val GPS_ROLLOVER_WEEKS1024 = (7 * 1024).days
+
+/**
+ * Sometimes the GPS timestamp is a few seconds in the future from the devices clock.
+ * To still catch these cases that are affected by this and the GPS rollover
+ * this value is subtracted from GPS_ROLLOVER_WEEKS1024.
+ */
+private val TIME_INACCURACY_FACTOR = 10.seconds
```
TimoPtr, line 47: Any particular reason to take 10s here? I would be curious to see what others are doing in this case.

Pawik55: No particular reason. I just thought 10 seconds be enough to not widen the timeframe to much.

TimoPtr: If this occur without the 1024 issue then it won't either. Like if we get a time in the future of 1s the difference is going to be negative and it will return false.

Could you write a test with this value.
```
val time = now + 5.seconds
```

I read a bit more about the drift and ask a famous AI here the answer it gave me

> Okay, let's discuss the issue of future GPS timestamps in an Android app.
> In general, GPS-enabled devices can sometimes report future timestamps due to a few potential reasons:
> - Clock Drift: The internal clock of the GPS receiver or the device itself may not be perfectly synchronized with the atomic clocks used by the GPS satellite system. This can lead to small discrepancies in the reported timestamps.
> - Leap Seconds: Occasionally, leap seconds are added to the UTC time system to account for variations in the Earth's rotation. This can cause the reported GPS time to temporarily be ahead of the local device time.
> - Ephemeris Data Issues: The GPS receiver relies on ephemeris data from the satellites to calculate its position and time. If this data is inaccurate or delayed, it can result in timestamps that appear to be in the future.
> In practice, the amount of time the GPS timestamp can be in the future is typically quite small, usually on the order of a few seconds at most. Anything significantly greater than that may indicate a more serious issue with the GPS data.
> If you encounter GPS timestamps that are more than a few seconds in the future, you may want to implement some validation or filtering logic in your app to detect and handle these cases appropriately. This could involve:
> - Checking the timestamp against the device's current time and discarding any that are too far in the future.
> - Implementing a maximum acceptable future timestamp threshold and clamping the values accordingly.
> - Logging and monitoring any instances of future timestamps to help identify the root cause.
> By being aware of this potential issue and implementing robust handling mechanisms, you can ensure your Android app can properly manage and utilize the GPS data it receives.
> 

So I'm fine about keeping 10.seconds, but you should extend it to not only the 1024 issue.

Pawik55: Hope I understood you right. The problem doesn't exist when GPS time isn't affected by the rollover. Substracting a time from the future with the current time will yield a negative number which is not greater then 5 minutes threshold. This already happens when the rollover time is from future and the 1024 are substracted. The result is a negative number which will never be greater than the 5 minutes threshold.
I added a unit test for the future time without the rollover bug but I think this test is redunant because the rollover test is already covering this case.

A reason for a bigger timefrime could be if someone sets the time manually. The GPS time could then be minutes in the future but for this a setting to disable the timechecks completely would be a better option in my opinion.

PR #5578, opened 2025-07-21
Update gesture action selection UI

jpelgrom:
## Summary

This PR updates the UI for selecting a gesture action from a dropdown to a full screen, grouping actions by a category.

Based on [design input](https://discord.com/channels/330944238910963714/1346948551892009101/1395373488512176158), while using existing components as much as possible. These changes are done for #5538 which will add a lot of actions, but I'm submitting this separately for easier review as it contains a new pattern with Compose Navigation code in a fragment. 

## Checklist

  _Adding Compose Navigation tests here feels out of scope as it requires a general Compose testing setup/this is not a navigation framework for the entire app._

## Screenshots

|Light|Dark|
|---|---|
|A screen titled 'Swipe up using three fingers', showing radio buttons and headings, light mode|A screen titled 'Swipe up using three fingers', showing radio buttons and headings, light mode|

n/a

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GestureActionsView.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GestureActionsView.kt
@@ -1,5 +1,6 @@
 package io.homeassistant.companion.android.settings.gestures.views
 
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
@@ -20,20 +21,21 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import io.homeassistant.companion.android.common.util.GestureAction
 import io.homeassistant.companion.android.settings.views.SettingsSubheader
+import io.homeassistant.companion.android.util.plus
 import io.homeassistant.companion.android.util.safeBottomPaddingValues
 
 /**
- * View showing all actions for a gesture, grouped by category, and the
- * currently configured action. Select to update the configured action.
+ * View showing all actions for a gesture, grouped by category, and the currently
+ * configured action. Clicking on an action calls [onActionClicked] with the action.
  *
  * @param selectedAction The current action
- * @param onActionSelected Called when an action is selected
+ * @param onActionClicked Called when an action is selected
  */
 @Composable
-fun GestureActionsView(selectedAction: GestureAction, onActionSelected: (GestureAction) -> Unit) {
+fun GestureActionsView(selectedAction: GestureAction, onActionClicked: (GestureAction) -> Unit) {
     val actionsGrouped = GestureAction.entries.minus(GestureAction.NONE).groupBy { it.category }
     LazyColumn(
-        contentPadding = safeBottomPaddingValues(applyHorizontal = false),
+        contentPadding = PaddingValues(vertical = 16.dp) + safeBottomPaddingValues(applyHorizontal = false),
         modifier = Modifier.selectableGroup(),
     ) {
         item {
@@ -41,7 +43,7 @@ fun GestureActionsView(selectedAction: GestureAction, onActionSelected: (Gesture
             GestureActionItem(
                 action = GestureAction.NONE,
                 checked = selectedAction == GestureAction.NONE,
-                onClick = { onActionSelected(GestureAction.NONE) },
+                onClick = { onActionClicked(GestureAction.NONE) },
             )
         }
         actionsGrouped.forEach { (category, actions) ->
@@ -52,7 +54,7 @@ fun GestureActionsView(selectedAction: GestureAction, onActionSelected: (Gesture
                 GestureActionItem(
                     action = action,
                     checked = selectedAction == action,
-                    onClick = { onActionSelected(action) },
+                    onClick = { onActionClicked(action) },
                 )
             }
         }
@@ -60,7 +62,7 @@ fun GestureActionsView(selectedAction: GestureAction, onActionSelected: (Gesture
 }
 
 @Composable
-fun GestureActionItem(action: GestureAction, checked: Boolean, onClick: () -> Unit) {
+private fun GestureActionItem(action: GestureAction, checked: Boolean, onClick: () -> Unit) {
     // Based on androidx.compose.material.samples.RadioGroupSample
     Row(
         Modifier
@@ -80,7 +82,7 @@ fun GestureActionItem(action: GestureAction, checked: Boolean, onClick: () -> Un
         )
         Text(
             text = stringResource(action.description),
-            style = MaterialTheme.typography.body1.merge(),
+            style = MaterialTheme.typography.body1,
             modifier = Modifier.padding(start = 32.dp),
         )
     }
@@ -88,9 +90,9 @@ fun GestureActionItem(action: GestureAction, checked: Boolean, onClick: () -> Un
 
 @Preview
 @Composable
-fun PreviewGestureActionsView() {
+private fun PreviewGestureActionsView() {
     GestureActionsView(
         selectedAction = GestureAction.QUICKBAR_DEFAULT,
-        onActionSelected = {},
+        onActionClicked = {},
     )
 }
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesListView.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesListView.kt
@@ -19,14 +19,14 @@ import io.homeassistant.companion.android.util.plus
 import io.homeassistant.companion.android.util.safeBottomPaddingValues
 
 /**
- * Shows all gestures with the configured action for it. Select a gesture
- * to change the configured action.
+ * Shows all gestures with the configured action for it. Clicking on a
+ * gesture calls [onGestureClicked] with the gesture.
  *
  * @param gestureActions User settings (gestures and the current action)
- * @param onSelected Called when a gesture is selected
+ * @param onGestureClicked Called when a gesture is selected
  */
 @Composable
-fun GesturesListView(gestureActions: Map<HAGesture, GestureAction>, onSelected: (HAGesture) -> Unit) {
+fun GesturesListView(gestureActions: Map<HAGesture, GestureAction>, onGestureClicked: (HAGesture) -> Unit) {
     LazyColumn(
         contentPadding = PaddingValues(vertical = 16.dp) + safeBottomPaddingValues(applyHorizontal = false),
     ) {
@@ -48,7 +48,7 @@ fun GesturesListView(gestureActions: Map<HAGesture, GestureAction>, onSelected: 
                     primaryText = stringResource(gesture.pointers.description),
                     secondaryText = action?.let { stringResource(it.description) } ?: "",
                     icon = null,
-                    onClicked = { onSelected(gesture) },
+                    onClicked = { onGestureClicked(gesture) },
                 )
             }
         }
@@ -60,6 +60,6 @@ fun GesturesListView(gestureActions: Map<HAGesture, GestureAction>, onSelected: 
 private fun PreviewGesturesListView() {
     GesturesListView(
         gestureActions = HAGesture.entries.associateWith { GestureAction.NONE },
-        onSelected = { _ -> },
+        onGestureClicked = { _ -> },
     )
 }
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesScreen.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesScreen.kt
@@ -5,18 +5,20 @@ import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.ui.platform.LocalContext
 import androidx.navigation.NavDestination.Companion.hasRoute
 import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
 import androidx.navigation.compose.rememberNavController
 import androidx.navigation.toRoute
+import io.homeassistant.companion.android.common.R
 import io.homeassistant.companion.android.common.util.GestureAction
 import io.homeassistant.companion.android.common.util.HAGesture
 import kotlinx.serialization.Serializable
 
-@Serializable object GesturesRoute
+@Serializable private data object GesturesRoute
 
-@Serializable data class ActionsRoute(val gesture: HAGesture)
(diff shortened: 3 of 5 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesListView.kt
jpelgrom, line 1: This is renamed from `GesturesView` with line 47-52 replacing the previous GestureSettingRow composable.

common/src/main/res/values/strings.xml
```diff
@@ -298,11 +298,20 @@
...
     <string name="gestures_direction_right">Swipe right</string>
     <string name="gestures_direction_up">Swipe up</string>
+    <string name="gesture_name_swipe_down_two">Swipe down using two fingers</string>
+    <string name="gesture_name_swipe_down_three">Swipe down using three fingers</string>
+    <string name="gesture_name_swipe_left_two">Swipe left using two fingers</string>
+    <string name="gesture_name_swipe_left_three">Swipe left using three fingers</string>
+    <string name="gesture_name_swipe_right_two">Swipe right using two fingers</string>
+    <string name="gesture_name_swipe_right_three">Swipe right using three fingers</string>
+    <string name="gesture_name_swipe_up_two">Swipe up using two fingers</string>
+    <string name="gesture_name_swipe_up_three">Swipe up using three fingers</string>
```
jpelgrom, line 314: These strings are new because simply using `direction + " " + number of pointers` may not work for all languages.

TimoPtr:
image
I think it misses one vertical padding around the list 'None' is too close from the top bar and with more item I think the last item would be too close to the navigation bar.
The vertical padding on the GestureView seems more accurate to the design.

Also on the design the category seems to have an extra padding between them which is not really visible in the impl (to be verified).

Navigation is kinda weird, when I click on a specific gesture and I press back in the topbar instead to be brought back to the GestureScreen I'm ending up into the Settings screen (but not with system backpress).

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GestureActionsView.kt
```diff
@@ -0,0 +1,96 @@
...
+                    checked = selectedAction == action,
+                    onClick = { onActionSelected(action) },
+                )
+            }
+        }
+    }
+}
+
+@Composable
+fun GestureActionItem(action: GestureAction, checked: Boolean, onClick: () -> Unit) {
```
TimoPtr:
```suggestion
private fun GestureActionItem(action: GestureAction, checked: Boolean, onClick: () -> Unit) {
```

jpelgrom:
> Also on the design the category seems to have an extra padding between them which is not really visible in the impl

Correct: the existing component for the header had slightly different padding with less difference between top and bottom.

> Navigation is kinda weird, when I click on a specific gesture and I press back in the topbar

Hadn't found that one in testing, good catch!

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GestureActionsView.kt
```diff
@@ -0,0 +1,96 @@
...
+            .padding(horizontal = 16.dp),
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        RadioButton(
+            selected = checked,
+            onClick = null, // Handled by parent Row
+        )
+        Text(
+            text = stringResource(action.description),
+            style = MaterialTheme.typography.body1.merge(),
```
TimoPtr: Why merge?

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GestureActionsView.kt
```diff
@@ -0,0 +1,96 @@
...
+            modifier = Modifier.padding(start = 32.dp),
```
TimoPtr, line 86: Why 32dp here? Shouldn't we have the same padding on the left and right of the RadioButton?

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GestureActionsView.kt
```diff
@@ -0,0 +1,96 @@
...
+        )
+    }
+}
+
+@Preview
+@Composable
+fun PreviewGestureActionsView() {
```
TimoPtr:
```suggestion
private fun PreviewGestureActionsView() {
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesListView.kt
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Shows all gestures with the configured action for it. Select a gesture
+ * to change the configured action.
+ *
+ * @param gestureActions User settings (gestures and the current action)
+ * @param onSelected Called when a gesture is selected
+ */
+@Composable
+fun GesturesListView(gestureActions: Map<HAGesture, GestureAction>, onSelected: (HAGesture) -> Unit) {
```
TimoPtr: While doing the onboarding I also came across this naming while creating my composables and it was confusing because `View` are an in-heritage from XML UI. I think we should use `GesturesList` or `GestureColumn`?

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GestureActionsView.kt
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * View showing all actions for a gesture, grouped by category, and the
+ * currently configured action. Select to update the configured action.
+ *
+ * @param selectedAction The current action
+ * @param onActionSelected Called when an action is selected
+ */
+@Composable
+fun GestureActionsView(selectedAction: GestureAction, onActionSelected: (GestureAction) -> Unit) {
```
TimoPtr: Same here we should drop `View`

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GestureActionsView.kt
```diff
@@ -0,0 +1,96 @@
...
+import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import io.homeassistant.companion.android.common.util.GestureAction
+import io.homeassistant.companion.android.settings.views.SettingsSubheader
+import io.homeassistant.companion.android.util.safeBottomPaddingValues
+
+/**
+ * View showing all actions for a gesture, grouped by category, and the
+ * currently configured action. Select to update the configured action.
```
TimoPtr:
```suggestion
 * currently configured action. An item can be selected and will trigger [onActionSelected].
```
The composable doesn't contain the logic it is hoist to its parent, the documentation should focus on the composable.

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesListView.kt
```diff
@@ -0,0 +1,65 @@
...
+import io.homeassistant.companion.android.common.util.GestureAction
+import io.homeassistant.companion.android.common.util.HAGesture
+import io.homeassistant.companion.android.settings.views.SettingsRow
+import io.homeassistant.companion.android.settings.views.SettingsSubheader
+import io.homeassistant.companion.android.util.plus
+import io.homeassistant.companion.android.util.safeBottomPaddingValues
+
+/**
+ * Shows all gestures with the configured action for it. Select a gesture
+ * to change the configured action.
```
TimoPtr:
```suggestion
 * Shows all gestures with the configured action for it. Clicking on a gesture triggers [onSelected].
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesListView.kt
TimoPtr: I would replace `onSelected` with `onGestureClick`

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesScreen.kt
```diff
@@ -0,0 +1,77 @@
...
+import androidx.navigation.compose.composable
+import androidx.navigation.compose.rememberNavController
+import androidx.navigation.toRoute
+import io.homeassistant.companion.android.common.util.GestureAction
+import io.homeassistant.companion.android.common.util.HAGesture
+import kotlinx.serialization.Serializable
+
+@Serializable object GesturesRoute
+
+@Serializable data class ActionsRoute(val gesture: HAGesture)
```
TimoPtr:
```suggestion
@Serializable private data object GesturesRoute

@Serializable private data class ActionsRoute(val gesture: HAGesture)
```
Having the route a `data` class or object makes it comparable but also nice to see in the logs it will strop the hascode https://kotlinlang.org/docs/object-declarations.html#data-objects

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesScreen.kt
```diff
@@ -0,0 +1,77 @@
...
+ *  @param gestureActions User settings (gestures and the current action)
+ *  @param onSetAction Called when the action for a gesture should be changed
+ *  @param setToolbarTitleFor Called when the screen changes, to update the
+ *         Activity's Toolbar title to the main title (`null`) or a gesture's title
+ */
+@Composable
+fun GesturesScreen(
+    gestureActions: Map<HAGesture, GestureAction>,
+    onSetAction: (HAGesture, GestureAction) -> Unit,
+    setToolbarTitleFor: (HAGesture?) -> Unit,
```
TimoPtr:
```suggestion
    onToolbarTitleChanged: (String) -> Unit,
```
So we can keep the UI logic in this file. It avoid dealing with a null value IMO.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/Gestures.kt
```diff
@@ -55,15 +62,52 @@ enum class GesturePointers(val asInt: Int, @StringRes val description: Int) {
...
-    SWIPE_LEFT_TWO(GestureDirection.LEFT, GesturePointers.TWO),
-    SWIPE_LEFT_THREE(GestureDirection.LEFT, GesturePointers.THREE),
-    SWIPE_RIGHT_TWO(GestureDirection.RIGHT, GesturePointers.TWO),
-    SWIPE_RIGHT_THREE(GestureDirection.RIGHT, GesturePointers.THREE),
-    SWIPE_UP_TWO(GestureDirection.UP, GesturePointers.TWO),
-    SWIPE_UP_THREE(GestureDirection.UP, GesturePointers.THREE),
-    SWIPE_DOWN_TWO(GestureDirection.DOWN, GesturePointers.TWO),
-    SWIPE_DOWN_THREE(GestureDirection.DOWN, GesturePointers.THREE),
+@Keep
+enum class HAGesture(
```
TimoPtr, line 66: ℹ️ Since `HAGesture` ends up into a compose object (the route) it triggers the lint rule `MissingKeepAnnotation from androidx.navigation.compose` which is valid when R8 is enabled.

I don't know if we will ever enable R8 tbh.

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GestureActionsView.kt
jpelgrom: To align the text with the title in the toolbar, otherwise you have text at three different positions from the start which looks messy.

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GesturesListView.kt
jpelgrom: I don't fully agree; having something like `View` as a suffix makes it more obvious that it is a larger Composable for standalone use/usage in a specific place, instead of a small reusable component.

The Compose equivalent of what you mention is an `AndroidView` so not too confusing IMO. Now in Android seems to use `Screen` as a suffix, that could also work.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/Gestures.kt
jpelgrom: I've added the annotation because lint required it. Suppressing will also require an annotation so it is easier to do it 'right', even if we are unlikely to need it.

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GestureActionsView.kt
jpelgrom: This was taken from the sample mentioned in the comment. Doesn't have a negative impact but I can remove it.

TimoPtr: For me it does nothing if we don't give it a parameter right?

common/src/main/kotlin/io/homeassistant/companion/android/common/util/Gestures.kt
TimoPtr: Yep this comment was mostly informative for the future to explain why this has been added.

The "issue" here is we would need a lot of @Keep if we ever want to have R8 working. I did this in a huge codebase in the past it's quite hard to do right.

app/src/main/kotlin/io/homeassistant/companion/android/settings/gestures/views/GestureActionsView.kt
jpelgrom: I misunderstood that docs comment as merging with the current local text style, but after looking at the source code it does nothing... Removed.

jpelgrom:
> I think it misses one vertical padding around the list

Fixed:

|Scrolled to top|Scrolled to bottom|
|---|---|
|image|image|

> > Navigation is kinda weird, when I click on a specific gesture and I press back in the topbar
> 
> Hadn't found that one in testing, good catch!

Fixed: the settings activity actually had a weird setup where it would check the SupportFragmentManager backstack first, only then the back handler. Changed to always use back handler as those will always match.

TimoPtr:
I think we should add some screenshots tests on this PR or in the next one with the new gestures.

TimoPtr (approved):

jpelgrom:
A screenshot test matching the preview? I'd prefer to add it to the next PR then because the new actions will cause the screen to change again.

TimoPtr:
> A screenshot test matching the preview? I'd prefer to add it to the next PR then because the new actions will cause the screen to change again.

Yes fine for me :)

jpelgrom:
@TimoPtr Good to merge this now then?

TimoPtr:
> @TimoPtr Good to merge this now then?

Yes I approved with giving you the choice to make the screenshots here or in the next PR 😊

PR #5581, opened 2025-07-22
Add network monitor during app start

### LINKED ISSUE #5524: Home App (Launcher) fails to load on first start, network not ready

**Android version(s):** 15 and 11

**Device model(s):** Lenovo Yoga Tab

**Home Assistant version:** App 2025.6.6-full Core 2025.7.1 OS 16

**Companion App Logs:**

When setting Home Assistant as the Home App Launcher on an Android device, the app launches correctly at startup. However, it starts a few seconds before the network is fully initialized. As a result, it fails to connect to Home Assistant because the network isn’t ready yet (shows "Unable to connect to Home Assistant"), it has to be retried manually.

A few seconds later, once the Wi-Fi connects, everything works fine if I manually refresh the external URL.

It would be helpful if there was a way to:

Delay the loading of the URL until the network is connected,
Automatically retry the connection once the network is available,
Or make the app wait a few seconds before trying to load.

```

```

Image

**Additional information:**

fpetrovski:
## Summary

This change adds a network monitor on app start, before launching the WebView. With this addition the launch activity presenter can observe the network status and react on 4 different states:
    READY_LOCAL,
    READY_REMOTE,
    CONNECTING,
    UNAVAILABLE,
 
If there is no network available an alert dialog is displayed, notifying the user to check their connection.

This change fixes #5524 

## Checklist

## Screenshots

The message used in the alert dialog is: Network unavailable, please enable your connection
Alert Message


app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchActivity.kt
```diff
@@ -142,6 +145,21 @@ class LaunchActivity :
...
+                .show()
+        }
+    }
+
+    override fun dismissDialog() {
+        networkDialog?.dismiss()
+        networkDialog = null
+    }
+
     override fun onDestroy() {
```
TimoPtr, line 166: Shouldn't we also dismiss the dialog in onDestroy and remove the reference?

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
```diff
@@ -1,15 +1,23 @@
...
+import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.launch
 
-abstract class LaunchPresenterBase(private val view: LaunchView, internal val serverManager: ServerManager) :
-    LaunchPresenter {
+abstract class LaunchPresenterBase(
+    private val view: LaunchView,
+    internal val serverManager: ServerManager,
+    internal val networkStatusMonitor: NetworkStatusMonitor,
```
TimoPtr:
```suggestion
    private val networkStatusMonitor: NetworkStatusMonitor,
```
If I'm not wrong it is only used in this file.

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
```diff
@@ -24,12 +32,20 @@ abstract class LaunchPresenterBase(private val view: LaunchView, internal val se
...
+                if (
                     serverManager.isRegistered() &&
-                    serverManager.authenticationRepository().getSessionState() == SessionState.CONNECTED
+                    serverManager.authenticationRepository().getSessionState() == SessionState.CONNECTED &&
+                    activeServer != null
                 ) {
-                    resyncRegistration()
-                    view.displayWebview()
+                    networkStatusMonitor.observeNetworkStatus(activeServer.connection)
+                        .distinctUntilChanged()
```
TimoPtr: `observeNetworkStatus` already have a distinctUntilChanged.

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
```diff
@@ -24,12 +32,20 @@ abstract class LaunchPresenterBase(private val view: LaunchView, internal val se
...
+                val activeServer = serverManager.getServer(ServerManager.SERVER_ID_ACTIVE)
+                if (
                     serverManager.isRegistered() &&
-                    serverManager.authenticationRepository().getSessionState() == SessionState.CONNECTED
+                    serverManager.authenticationRepository().getSessionState() == SessionState.CONNECTED &&
+                    activeServer != null
                 ) {
-                    resyncRegistration()
-                    view.displayWebview()
+                    networkStatusMonitor.observeNetworkStatus(activeServer.connection)
```
TimoPtr: This is started inside the `mainScope` that is not tight to a lifecycle. It means that it will continue to observeNetworkStatus while the app is in background and probably at some point invoke `view.displayWebView` which I think is wrong. I'm not even sure what would happen tbh.

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
```diff
@@ -24,12 +32,20 @@ abstract class LaunchPresenterBase(private val view: LaunchView, internal val se
...
+                        .distinctUntilChanged()
+                        .collectLatest { state ->
+                            if (handleNetworkState(state)) cancel()
```
TimoPtr: Cancel will cancel the entire `mainScope` is it what you expect?

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
```diff
@@ -39,6 +55,25 @@ abstract class LaunchPresenterBase(private val view: LaunchView, internal val se
...
 
+    private fun handleNetworkState(state: NetworkState): Boolean = when (state) {
+        NetworkState.READY_LOCAL, NetworkState.READY_REMOTE -> {
+            view.dismissDialog()
+            resyncRegistration()
+            view.displayWebView()
+            true
+        }
+
+        NetworkState.CONNECTING -> {
```
TimoPtr, line 57: This state is kinda new compare to before. How the UI will look like in this state?

Note: In the work I'm doing to redo the whole onboarding/start of the app, this could be useful to show the splashscreen until the webview is fully loaded.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkCapabilitiesChecker.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import android.net.ConnectivityManager
+import android.net.NetworkCapabilities
+import android.net.NetworkInfo
+import android.os.Build
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class NetworkCapabilitiesChecker @Inject constructor(private val connectivityManager: ConnectivityManager) {
```
TimoPtr:
```suggestion
internal class NetworkCapabilitiesChecker @Inject constructor(private val connectivityManager: ConnectivityManager) {
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkCapabilitiesChecker.kt
```diff
@@ -0,0 +1,40 @@
+package io.homeassistant.companion.android.common.data.network
+
+import android.net.ConnectivityManager
+import android.net.NetworkCapabilities
+import android.net.NetworkInfo
+import android.os.Build
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
```
TimoPtr, line 18: Any particular reason to have this class a Singleton? 

Also I would like some documentation on the purpose of the class, explaining what you get out of it.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkCapabilitiesChecker.kt
```diff
@@ -0,0 +1,40 @@
...
+        if (!info.isConnected) return false
+
+        return when (transport) {
+            NetworkCapabilities.TRANSPORT_WIFI -> info.type == ConnectivityManager.TYPE_WIFI
+            NetworkCapabilities.TRANSPORT_CELLULAR -> info.type == ConnectivityManager.TYPE_MOBILE
+            NetworkCapabilities.TRANSPORT_ETHERNET -> info.type == ConnectivityManager.TYPE_ETHERNET
+            NetworkCapabilities.TRANSPORT_VPN -> info.type == ConnectivityManager.TYPE_VPN
+            else -> false
+        }
+    }
```
TimoPtr, line 47: All of this methods are deprecated, is there no alternative?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkHelper.kt
```diff
@@ -0,0 +1,42 @@
+package io.homeassistant.companion.android.common.data.network
+
+import android.net.NetworkCapabilities
+import android.os.Build
+import javax.inject.Inject
+import javax.inject.Singleton
+
+interface NetworkHelper {
```
TimoPtr, line 8:
```suggestion
internal interface NetworkHelper {
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkHelper.kt
```diff
@@ -0,0 +1,42 @@
...
+    /** Returns if the active data connection is using ethernet */
+    fun isUsingEthernet(): Boolean
+
+    /** Returns if the active data connection is a VPN connection */
+    fun isUsingVpn(): Boolean
+}
+
+@Singleton
+class NetworkHelperImpl @Inject constructor(private val capabilitiesChecker: NetworkCapabilitiesChecker) :
+    NetworkHelper {
```
TimoPtr:
```suggestion
@Singleton
private class NetworkHelperImpl @Inject constructor(private val capabilitiesChecker: NetworkCapabilitiesChecker) :
    NetworkHelper {
```

Any reason to have a Singleton?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitor.kt
```diff
@@ -0,0 +1,71 @@
...
+     */
+    fun observeNetworkStatus(serverConfig: ServerConnectionInfo): Flow<NetworkState>
+}
+
+/**
+ * Tracks network changes using ConnectivityManager callbacks.
+ * Automatically handles callback registration/cleanup.
+ */
+@Singleton
+class NetworkStatusMonitorImpl @Inject constructor(
```
TimoPtr:
```suggestion
@Singleton
internal class NetworkStatusMonitorImpl @Inject constructor(
```

No need to expose the impl outside of `:common`. Still the same question do we need it to be a Singleton.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitor.kt
```diff
@@ -0,0 +1,71 @@
...
+    private val connectivityManager: ConnectivityManager,
+    private val networkHelper: NetworkHelper,
+) : NetworkStatusMonitor {
+
+    override fun observeNetworkStatus(serverConfig: ServerConnectionInfo): Flow<NetworkState> = callbackFlow {
```
TimoPtr, line 62: Would it make sense (and possible) to do this in a IO dispatcher?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitor.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    private fun getCurrentNetworkState(serverConfig: ServerConnectionInfo): NetworkState = when {
+        !networkHelper.hasActiveNetwork() -> NetworkState.UNAVAILABLE
+        serverConfig.isInternal(requiresUrl = false) -> NetworkState.READY_LOCAL
+        networkHelper.isNetworkValidated() -> NetworkState.READY_REMOTE
+        else -> NetworkState.CONNECTING
+    }
+}
+
+enum class NetworkState {
```
TimoPtr: Move this close to the interface definition and add some precise documentation of each state meaning. It is not really clear what READY_LOCAL means for instance.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitor.kt
```diff
@@ -0,0 +1,71 @@
...
+
+        awaitClose {
+            connectivityManager.unregisterNetworkCallback(callback)
+        }
+    }.distinctUntilChanged()
+
+    private fun getCurrentNetworkState(serverConfig: ServerConnectionInfo): NetworkState = when {
+        !networkHelper.hasActiveNetwork() -> NetworkState.UNAVAILABLE
+        serverConfig.isInternal(requiresUrl = false) -> NetworkState.READY_LOCAL
+        networkHelper.isNetworkValidated() -> NetworkState.READY_REMOTE
```
TimoPtr, line 102: Both could be true at the same time, I think it would be beneficial to document the logic in this function

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/WifiHelper.kt
```diff
@@ -17,10 +17,4 @@ interface WifiHelper {
     fun isUsingSpecificWifi(networks: List<String>): Boolean
     fun getWifiSsid(): String?
     fun getWifiBssid(): String?
-
```
TimoPtr, line 20: 🙏🏻 thanks for removing irrelevant function from this class.

common/src/main/res/values/strings.xml
```diff
@@ -257,6 +257,7 @@
     <string name="error_auth_revoked">It appears that your authorization was revoked, please reconnect to Home Assistant.</string>
     <string name="error_connection_failed">Unable to connect to Home Assistant.</string>
     <string name="error_connection_failed_to">Unable to connect to %1$s</string>
+    <string name="error_connection_failed_no_network">Network unavailable, please enable your connection</string>
```
TimoPtr:
```suggestion
    <string name="error_connection_failed_no_network">Network unavailable, please check your connection</string>
```

common/src/test/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitorImplTest.kt
```diff
@@ -0,0 +1,153 @@
...
+            )
+        } just Runs
+
+        every {
+            connectivityManager.unregisterNetworkCallback(any<ConnectivityManager.NetworkCallback>())
+        } just Runs
+    }
+
+    @Test
+    fun `given monitoring flow when monitoring is canceled then flow unregisters callback`() = runTest {
```
TimoPtr: Nitpick but we usually start with `Given` like a sentence.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitorImplTest.kt
```diff
@@ -0,0 +1,153 @@
...
+        }
+        advanceUntilIdle()
+        job.cancel()
+        advanceUntilIdle()
+
+        verify { connectivityManager.unregisterNetworkCallback(callbackSlot.captured) }
+    }
+
+    @Test
+    fun `given no active network when monitoring starts then emit UNAVAILABLE`() = runTest {
```
TimoPtr:
```suggestion
    fun `given no active network when monitoring starts then emits UNAVAILABLE`() = runTest {
```
To confirm with @jpelgrom my english grammar is far from being accurate ...

common/src/test/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitorImplTest.kt
```diff
@@ -0,0 +1,153 @@
...
+        every { networkHelper.isNetworkValidated() } returns false
+
+        val states = mutableListOf<NetworkState>()
+        val job = launch {
+            networkMonitor.observeNetworkStatus(serverConfig).take(2).toList(states)
+        }
+        advanceUntilIdle()
+        callbackSlot.captured.onAvailable(network)
+        advanceUntilIdle()
+        job.cancel()
```
TimoPtr, line 132: Do you need the advanceUntilIdle and cancel here I'm not sure.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkCapabilitiesChecker.kt
dshokouhi: Do we need to also consider the bluetooth transport type for Wear OS? I know this PR is for the phone but technically watches use bluetooth for data when actively connected to a phone

common/src/test/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitorImplTest.kt
jpelgrom: Different suggestion to avoid the discussion and for consistency with the final two tests:

```suggestion
    fun `given no active network when monitoring starts then state is UNAVAILABLE`() = runTest {
```

FWIW I think emits is correct, but you're missing the actor in the title (_who/what_ emits?). emit is what the function will do (via `trySend`) but that's not what you're testing, you're testing the result.

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
fpetrovski: Yes the mainScope is used in a few places and I don't know the reason behind that decision so I didn't want to refactor that part. The good thing is that there is a onFinish() method that cancels the scope. 
It's currently called from onDestroy in the activity. Should I move it to onStop() and keep mainScope, or refactor and change the coroutine scope?

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
fpetrovski: I believe, since we are launching a new coroutine from the mainScope (mainScope.launch{...}), cancel will stop only the new created child coroutine. 
Also we do need to cancel this since the job is done: the network is acquired, we've connected to the server and the webView is launched.

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
fpetrovski: I think I got confused that there is a loader indicator, but that might actually be from the webView later on. I will double check.
I can also mark this PR dependant of yours if you'd like?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkCapabilitiesChecker.kt
fpetrovski: Regarding the Singleton annotation for this and the other related classes, if you are asking why they are annotated both here and in the NetworkModule, I guess out of habit 😅. Technically only one is needed, but I think it's harmless.
If you are asking why they are singleton at all, it's because they are related to the network status and that's a system wide service. It makes sense to have one instance of those classes and have one monitor at a time.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkCapabilitiesChecker.kt
fpetrovski: This is only a fallback used for devices running below API 23 where getNetworkCapabilities isn't available. Since the minsdk will get a bump, we can remove this soon.
Regarding WearOS, honestly I'm not that upto date with it and I'm open to suggestions @dshokouhi .

common/src/test/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitorImplTest.kt
fpetrovski: In this case we do need advanceUntilIdle to process the emissions and also to allow the onAvailable callback to execute its code. Otherwise we would get a value not yet captured error. 
Regarding the cancelation, you are right, I think we don't really need that.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitor.kt
fpetrovski: Yes, good point calling this out. The feature actually started as a fix for a bug where the app tried to connect to the server too early: #5524 
So I guess having this in the back of my mind it evolved to not be just a generic check, but also whether the current network allows reaching the server locally or remotely. 
I'll definitely add some more docs across the feature.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkCapabilitiesChecker.kt
dshokouhi: Ah right my comment was probably more for the method `hasActiveNetwork` we should probably extend that to support any of the transports that can have a data connection instead of just the 4 listed? My watch is just an example but I think the other transports may be valid for other use cases. We actually have that data as a sensor "Network type".

app/src/full/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterImpl.kt
```diff
@@ -5,34 +5,36 @@ import dagger.hilt.android.qualifiers.ActivityContext
...
-                    } catch (e: Exception) {
-                        Timber.e(e, "Issue updating Registration")
-                    }
-                }
+class LaunchPresenterImpl @Inject constructor(
+    @ActivityContext context: Context,
+    serverManager: ServerManager,
+    networkStatusMonitor: NetworkStatusMonitor,
+) : LaunchPresenterBase(context as LaunchView, serverManager, networkStatusMonitor) {
+    override suspend fun resyncRegistration() = withContext(Dispatchers.IO) {
```
TimoPtr: Actually I don't think we need to change the dispatcher here, it was using IO before because we didn't have any scope, but now with this change we can delegate to whatever is calling. The network and DB call are made on a dedicated dispatcher already.

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenter.kt
```diff
@@ -1,12 +1,12 @@
 package io.homeassistant.companion.android.launch
 
+import kotlinx.coroutines.CoroutineScope
+
 interface LaunchPresenter {
 
-    fun onViewReady(serverUrlToOnboard: String? = null)
+    fun onViewReady(serverUrlToOnboard: String? = null, coroutineScope: CoroutineScope)
```
TimoPtr: You could directly make this suspendable and start it with the right scope in the activity directly. Btw it's going to take few weeks but I'm redoing this part and getting ride of the presenter in favor of a ViewModel and compose navigation.

app/src/minimal/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterImpl.kt
```diff
@@ -2,28 +2,32 @@ package io.homeassistant.companion.android.launch
...
-                    serverManager.webSocketRepository(it.id).getCurrentUser() // Update cached data
-                } catch (e: Exception) {
-                    Timber.e(e, "Issue updating Registration")
-                }
+class LaunchPresenterImpl @Inject constructor(
+    view: LaunchView,
+    serverManager: ServerManager,
+    networkStatusMonitor: NetworkStatusMonitor,
+) : LaunchPresenterBase(view, serverManager, networkStatusMonitor) {
+    override suspend fun resyncRegistration() = withContext(Dispatchers.IO) {
```
TimoPtr: same here

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkCapabilitiesChecker.kt
```diff
@@ -0,0 +1,47 @@
...
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * A utility class that provides access to current network transport capabilities.
+ * It is primarily a wrapper around [ConnectivityManager] used to determine if the device is connected
+ * via specific transport type.
+ * It providing both modern checks using [NetworkCapabilities] and fallback logic
+ * using [NetworkInfo] for devices running pre-Marshmallow (API < 23).
+ */
```
TimoPtr:
```suggestion
/**
 * A utility class providing access to current network transport capabilities.
 * It wraps [ConnectivityManager] to determine if the device is connected
 * via a specific transport type.
 *
 * It offers checks using [NetworkCapabilities] (for Android Marshmallow/API 23 and above)
 * and provides fallback logic using [NetworkInfo] for older devices (pre-Marshmallow/API < 23).
 */
```

Don't hesitate to use AI like gemini from Android Studio to help you check the grammar and wording of the documentation.

common/src/main/kotlin/io/homeassistant/companion/android/di/NetworkModule.kt
```diff
@@ -0,0 +1,48 @@
...
+
+@Module
+@InstallIn(SingletonComponent::class)
+internal abstract class NetworkModule {
+
+    companion object {
+        @Provides
+        @Singleton
+        fun provideConnectivityManager(@ApplicationContext appContext: Context) =
+            appContext.getSystemService<ConnectivityManager>()!!
```
TimoPtr: Replace with checkNotNull and explicit error message even if it should never happen.

TimoPtr (approved):
Looks good 👍🏻 Let's wait for @jpelgrom to take a look too he has more history than me on these topics.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkCapabilitiesChecker.kt
fpetrovski: Funny thing is I did use AI for this, but then I decided to be smart and modify it... badly 😅

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenter.kt
fpetrovski: That’s awesome! I’d actually been weighing the idea of moving the presenter to MVVM/MVI myself, but as a newer contributor I figured it wasn’t my place to do such a big change 😄

TimoPtr: Well if you are willing to review my work 🙏🏻 it would be very helpful. If you do want to tackle a big piece we can still discuss on how to tackle this in smaller pieces.

app/src/full/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterImpl.kt
```diff
@@ -5,34 +5,36 @@ import dagger.hilt.android.qualifiers.ActivityContext
...
+        if (!serverManager.isRegistered()) return@withContext
+
+        serverManager.defaultServers.forEach {
```
jpelgrom, line 25: By removing the `launch` inside, if you have multiple servers to update they now run in sequence instead of parallel, potentially slowing down launch. (This function isn't supposed to be blocking, it's a best effort update.)

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenter.kt
jpelgrom: > You could directly make this suspendable and start it with the right scope in the activity directly. 

This makes more sense to me... Passing the CoroutineScope is unusual and I don't see a particular reason why you couldn't do it from the activity.

app/src/full/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterImpl.kt
TimoPtr: Good point, we should just use launch it will use the scope of the caller to get the result, but the network/db calls are still going to be on a background thread.

fpetrovski: > By removing the `launch` inside, if you have multiple servers to update they now run in sequence instead of parallel, potentially slowing down launch. (This function isn't supposed to be blocking, it's a best effort update.)

This makes sense, I'll remove the context switching and just use coroutineScope + launch to achieve parallelism.

TimoPtr (approved):

PR #5591, opened 2025-07-24
Fix disk violations and enable strict mode threading death penalty

### LINKED ISSUE #5520: DB calls read/write on mainThread

TimoPtr:
## Summary

This PR is the last one to closes #5520, by fixing and ignoring the last bits of code that were doing disk access on the main thread. Since all the issues seems to have been fixed this PR enable death penalty for threading issues.

## Checklist

## Any other notes

I introduce a `SuspendProvider` to bypass the limitation of dagger that doesn't allow use to work with coroutine at the DI level. Instead we simply wrap the coroutine in a `SuspendProvider` so that we can get the value in a coroutine scope later.


common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
```diff
@@ -1,73 +1,112 @@
...
+ * [SharedPreferences] only once in the lifecycle of this class.
+ */
+class LocalStorageImpl(sharedPreferences: suspend () -> SharedPreferences) : LocalStorage {
+    private val sharedPreferences = SharedPreferenceRetriever(sharedPreferences)
 
     override suspend fun putString(key: String, value: String?) {
-        sharedPreferences.edit().putString(key, value).apply()
+        withContext(Dispatchers.IO) {
+            sharedPreferences().edit { putString(key, value) }
+        }
```
jpelgrom: Why is this 3 lines when `putBoolean` and `putStringSet`, which are longer, are 1 line?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
```diff
@@ -5,267 +5,289 @@ import io.homeassistant.companion.android.common.data.LocalStorage
 import io.homeassistant.companion.android.common.data.integration.ControlsAuthRequiredSetting
 import io.homeassistant.companion.android.common.util.GestureAction
 import io.homeassistant.companion.android.common.util.HAGesture
+import java.util.concurrent.atomic.AtomicBoolean
 import javax.inject.Inject
 import javax.inject.Named
-import kotlinx.coroutines.runBlocking
```
jpelgrom, line 10: Moving all the constants here makes the diff a mess, is that really necessary / can't you do it separately?

TimoPtr:
> Tested and did not see any major crashes when running, so I'm generally ok with it as any potential impact will be limited to developers.
> 
> Like server, having to add `()` at the end of localStorage feels a bit like a downgrade and less Kotlin-y.

I agree about the less kotlin-y do you have any idea on how to avoid this. It's mostly because getting a server now is a suspend function.

common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
TimoPtr: no reason I will make it one line

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
TimoPtr: TBH I didn't took this in consideration while doing it. But now that I'm looking at it since I've already modify every method because of `()` I don't think it is a big deal don't you think?

jpelgrom:
> I agree about the less kotlin-y do you have any idea on how to avoid this. It's mostly because getting a server now is a suspend function.

Unfortunately not, I tried some things and searched but couldn't find a good solution. Just mentioning it here for future reference ;)

For local storage: I considered (re)moving the injection in the interface and doing it on demand when calling a suspending function, so you can still inject in the repositories, but that's more boilerplate for each function. For server: I don't see a way without decoupling it from the data class but that seems like more work than it's worth.

common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
TimoPtr: I'm going to keep 3 lines when we have remove I find it more readable.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
jpelgrom: Maybe not, but it did make me miss that there are actually changes to the migration.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
```diff
@@ -5,267 +5,289 @@ import io.homeassistant.companion.android.common.data.LocalStorage
...
+                        PREF_CONTROLS_AUTH_REQUIRED,
+                        PREF_CONTROLS_AUTH_ENTITIES,
+                        PREF_FULLSCREEN_ENABLED,
+                        PREF_KEEP_SCREEN_ON_ENABLED,
+                        PREF_PINCH_TO_ZOOM_ENABLED,
+                        PREF_AUTOPLAY_VIDEO,
+                        PREF_ALWAYS_SHOW_FIRST_VIEW_ON_APP_START,
+                        PREF_WEBVIEW_DEBUG_ENABLED,
+                    ).forEach { key ->
+                        integrationStorage.getString(key)?.let {
```
jpelgrom: You're now always using `getString` for each pref in the migration vs `getString`, `getStringSet` or `getBoolean` as needed.

common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
jpelgrom: The `remove` function was already 1 line? I care more about consistency and ktlint being ok with it than whether it is 1 or 3 lines.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
TimoPtr: 😱😱 good catch I went too fast on this. I felt too confident, I'm going to revert this change.

PR #5600, opened 2025-07-28
Use on prereleased instead of tag to update changelog

TimoPtr:
## Summary

It seems that our `onTag` workflow was not triggered by the creation of a release. So this PR instead switch the trigger to `prereleased` which will do exactly the same has what we wanted with the tag. I didn't use `publish` or `release` since we always first create a `pre-release`.

## Checklist

### DIFF
--- a/.github/workflows/onPrereleased.yml
+++ b/.github/workflows/onPrereleased.yml
@@ -4,9 +4,8 @@ name: Prepare next release
 
 on: # yamllint disable-line rule:truthy
   workflow_dispatch:
-  push:
-    tags:
-      - '20[2-9][0-9].[0-9]+.[0-9]+' # Simple matching for CalVer, even if it's not perfect it is enough for detecting that it is a version tag.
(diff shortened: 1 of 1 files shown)


jpelgrom:
> this PR instead switch the trigger to `prereleased` which will do exactly the same has what we wanted with the tag.

No, it's missing the month/year change which is a tag but not a pre release.

TimoPtr:
> > this PR instead switch the trigger to `prereleased` which will do exactly the same has what we wanted with the tag.
> 
> No, it's missing the month/year change which is a tag but not a pre release.

I was wondering actually on this. It will get a pre release on the next weekly. I don't think we need to handle it don't you think?

jpelgrom:
#5601 updated the version to `2025.7.5`, but the actual push (and the next prerelease on GitHub) this week will be `2025.8.1`. So it will be wrong unless we manually update it this week.

PR #5606, opened 2025-07-29
Support update pull request and monthly tag for update changelog version

TimoPtr:
## Summary

Based on the comment https://github.com/home-assistant/android/pull/5600#issuecomment-3127722725 we should also handle the case of tags to support monthly and yearly changes push through a tag not a release like weekly.
Since we might have update overlapping over each other, I've modify the workflow so it handle updating an open PR if it is not merged yet. 

## Any other notes

Once merged I'll create a dumb tag to make sure it works.


TimoPtr:
Tag got trigger here https://github.com/home-assistant/android/actions/runs/16589205370

TimoPtr:
PR got created 
image

TimoPtr:
Update of a PR also works https://github.com/home-assistant/android/actions/runs/16589252052/job/46920738950

PR #5609, opened 2025-07-29
Extract common configuration + theme from webview

TimoPtr:
## Summary

This PR is part of a larger effort to extract the logic of the webView into dedicated functions/classes. Here it's focus on the common settings and `theme` that has been renamed to `nightModeTheme` and replace by an enum instead of a string.

## Checklist


app/src/main/kotlin/io/homeassistant/companion/android/themes/NightModeManager.kt
```diff
@@ -0,0 +1,37 @@
+package io.homeassistant.companion.android.themes
+
+import android.os.Build
+import io.homeassistant.companion.android.common.data.prefs.NightModeTheme
+import io.homeassistant.companion.android.common.data.prefs.PrefsRepository
+import javax.inject.Inject
+import timber.log.Timber
+
+class NightModeManager @Inject constructor(private val prefsRepository: PrefsRepository) {
```
TimoPtr, line 21: I might be wrong but for me it is mostly focus on whether or not it's in night mode?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
```diff
@@ -1,17 +1,47 @@
 package io.homeassistant.companion.android.common.data.prefs
 
+import androidx.appcompat.app.AppCompatDelegate
 import io.homeassistant.companion.android.common.data.integration.ControlsAuthRequiredSetting
 import io.homeassistant.companion.android.common.util.GestureAction
 import io.homeassistant.companion.android.common.util.HAGesture
 
+enum class NightModeTheme(val storageValue: String) {
```
TimoPtr, line 7: I was wondering if this is at the right place if you have suggestions.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
```diff
@@ -1,17 +1,47 @@
...
+    LIGHT("light"),
+    DARK("dark"),
+
+    // TODO double check this
+    @Deprecated("Kept for backwards compatibility")
+    ANDROID("android"),
```
TimoPtr: I'm not sure about this one.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/HAWebView.kt
```diff
@@ -29,9 +55,72 @@ fun HAWebView(modifier: Modifier = Modifier, configure: WebView.() -> Unit = {},
...
 }
+
+fun WebView.settings(configureDsl: WebSettings.() -> Unit) {
+    settings.apply(configureDsl)
+}
+
+@SuppressLint("SetJavaScriptEnabled")
+private fun WebSettings.defaultSettings() {
+    // https://github.com/home-assistant/android/pull/3353
+    minimumFontSize = 5
```
TimoPtr, line 96: Changing the settings of the WebView makes the screenshot testing generate a blank screenshot...

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/HAWebView.kt
```diff
@@ -29,9 +56,84 @@ fun HAWebView(modifier: Modifier = Modifier, configure: WebView.() -> Unit = {},
...
     )
 }
+
+fun WebView.settings(configureDsl: WebSettings.() -> Unit) {
+    try {
+        settings.configureDsl()
+    } catch (e: NoSuchMethodError) {
+        // While displaying the WebView within a Preview or while making screenshot test getSettings is throwing
+        // `java.lang.NoSuchMethodError` we catch the error to be able to continue to use Preview and screenshot tests.
+        Timber.e(
```
jpelgrom: As this is expected, maybe downgrade it to a warning?

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/HAWebView.kt
jpelgrom: Isn't this fixed by the DSL above?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
jpelgrom: The enum is OK I think, but having `AppCompatDelegate` in here is a bit weird. Why not put that in NightModeManager (in the app module)?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
jpelgrom: That's correct. There used to be different settings for `android` and `system` but they did the same, see my PR where I merged them #2923.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
TimoPtr: while looking at this more carefully it's kinda weird that we can change the default night mode outside of the Manager, it means that at some point we could store a value that is different from the one we apply if we make a mistake somewhere.

Shouldn't we read and apply the default night mode only in the settings when we change the value and from the HomeAssistantApplication ?

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/HAWebView.kt
```diff
@@ -29,9 +56,84 @@ fun HAWebView(modifier: Modifier = Modifier, configure: WebView.() -> Unit = {},
...
+    javaScriptEnabled = true
+    domStorageEnabled = true
+    // https://github.com/home-assistant/android/pull/2252
+    displayZoomControls = false
+    userAgentString += " ${HomeAssistantApis.USER_AGENT_STRING}"
+}
+
+@Suppress("DEPRECATION")
+private fun WebSettings.setNightModeTheme(nightModeTheme: NightModeTheme, uiMode: Int) {
+    nightModeTheme.setAsDefaultNightMode()
```
TimoPtr: @jpelgrom I guess we apply the defaut here in case it has been reset by the system? If not could you let me know why we do it here? If yes then I think it might be better in the Application class wdyt?

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/HAWebView.kt
TimoPtr: it is indeed

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/HAWebView.kt
TimoPtr: My last commit remove this from the webView and move it to the Application lemme know what you think.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
jpelgrom: I don't quite follow your comment. At what point is the light/dark theme changed outside the manager class?

TimoPtr: In my previous impl it was in the manager and on the WebView and in the initial impl it was indeed in the manager but applied when we were loading the WebView.

I was wondering why applying the night mode theme when we load the WebView and not when we start the application (that's my latest impl).

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/HAWebView.kt
jpelgrom: If that works consistently (also on older Android versions / if not launched via the LaunchActivity), that seems like a nice cleanup. I don't think there is a specific reason for why it is applied in these places other than how it was added piece by piece.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
jpelgrom: OK let's just continue that conversation in https://github.com/home-assistant/android/pull/5609#discussion_r2244737531, two review comment threads about the same subject.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
```diff
@@ -4,14 +4,29 @@ import io.homeassistant.companion.android.common.data.integration.ControlsAuthRe
 import io.homeassistant.companion.android.common.util.GestureAction
 import io.homeassistant.companion.android.common.util.HAGesture
 
+enum class NightModeTheme(val storageValue: String) {
+    LIGHT("light"),
+    DARK("dark"),
+
+    // TODO double check this
```
jpelgrom:
```suggestion
```

jpelgrom (approved):

Issue #5605, opened 2025-07-29
Add Health Connect sensor to minimal app


MA3o:
**Is your feature request related to a problem? Please describe.**
The Health Connect sensor in the companion app is currently only available in the full version of the app, not in the minimal version.  I wonder if it could to be added to the minimal flavor as well.

Health connect is part of the Android framework since Android version 14 ( as per the [API docs](https://developer.android.com/health-and-fitness/guides/health-connect/develop/get-started) ) and seems to no longer rely on Google Play Services or require a third party app to be installed. It is even available on some highly privacy-focused Android distributions like GrapheneOS ( as they have discussed [here](https://discuss.grapheneos.org/d/8285-will-health-connect-be-removed-hopefully) ). 

**Describe the solution you'd like**
The Health Connect implementation already exist in the full version. I suggest that they are made available in the minimal version of the app.

**Describe alternatives you've considered, if any**
Installing the full version of the Companion App is a fully working alternative. 

**Additional context**
I am specifically bringing this up because the developers of [Gadgetbridge](https://codeberg.org/Freeyourgadget/Gadgetbridge), a FOSS and privacy-friendly app that allows to interact with many Bluetooth gadgets and aims to replace their vendor specific apps, is currently working on a [PR](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/4481) to integrate Health Connect functionality. Since Gadgetbridge follows a similar philosophy as Home Assistant (privacy-respecting, local-first, against vendor lock-in a.s.o.), combining these two could be a great way to integrate data from a lot of devices into home assistant in a privacy-friendly way. Adding this sensor to the minimal version would allow for this functionality to be available on devices that don't have Google services installed.

dshokouhi:
Based on my understanding play services is a hard requirement. The only difference Android 14 made was no need for the health connect app as it's baked in.

https://developer.android.com/health-and-fitness/guides/health-connect/frequently-asked-questions#play-services

MA3o:
I assumed that this was a specific requirement of the health connect app, which is now no longer needed since the functionality has been merged into AOSP.

Since I'm not sure myself, I asked people over at the Gadgetbridge PR ( [link](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/4481#issuecomment-6042487) ) for clarification on what they have found out while implementing this. One user reports to have used health connect on their de-googled phone, another confirmed that Google play services were only required for the separate app, which is no longer needed.

[...]

PR #5612, opened 2025-07-31
Move Health Connect to main module

dshokouhi:
## Summary

Based on #5605 health connect should work on the minimal build if the SDK is available. We have a user who can test on a de-googled phone to confirm it works for them. Our hasSensor check should work as is here.

## Checklist

User Documentation: home-assistant/companion.home-assistant#1230

## Any other notes

We should probably get some users to test and confirm, on my end sensors continue to work as expected.

I used android studio to move the files and committed as is, the `Companion` addition in multiple locations surprised me but works.

### DIFF
--- a/app/src/full/AndroidManifest.xml
+++ b/app/src/full/AndroidManifest.xml
@@ -33,17 +33,6 @@
             </intent-filter>
         </activity>
 
-        <activity-alias
-            android:name="ViewHealthConnectPermissionUsageActivity"
-            android:exported="true"
-            android:targetActivity=".settings.sensor.HealthConnectPermissionActivity"
-            android:permission="android.permission.START_VIEW_PERMISSION_USAGE">
-            <intent-filter>
-                <action android:name="android.intent.action.VIEW_PERMISSION_USAGE" />
-                <category android:name="android.intent.category.HEALTH_PERMISSIONS" />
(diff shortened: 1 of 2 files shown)


jpelgrom:
Why remove the `import io.homeassistant.companion.android.common.R` alias? If we keep that there are less changes and I think it's a nice alias to avoid confusion about which module's R is used.

dshokouhi:
> Why remove the `import io.homeassistant.companion.android.common.R` alias? If we keep that there are less changes and I think it's a nice alias to avoid confusion about which module's R is used.

i can switch it back, that was all android studio lol

jpelgrom:
Simple move and changing the dependency looks good, also after reviewing the docs, but let's wait for user feedback before merging.

TimoPtr (approved):

dshokouhi:
We have received end user confirmation on a custom ROM :)

https://github.com/home-assistant/android/issues/5605#issuecomment-3553854450

with that said not sure how we feel about merging something health connect related due to the recent rash of rejections😂

TimoPtr:
I don't see any issue about merging this since it is already in the full flavor 👍🏻. The only issue could be if F-droid are forbidding the use of this specific library but I don't think so since it's `androidx`.

PR #5622, opened 2025-08-05
Add to dialog (widget, android auto favorite) 

TimoPtr:
## Summary

We want to make the native features of the companion apps more accessible to the user. When the user is going to click on `Add to` from the frontend it will display a bottom modal showing the list of potential actions. This PR covers the needed change to react to this click, inform the frontend that we support this feature and expose a way to add an Android Auto favorite or a widget depending on the domain of the entity. Follow-up PRs are going to add shortcut, tile and watch favorite support.

## Checklist

## Screenshots

image
image

## Any other notes

This features depends on the frontend, it will only work when this https://github.com/home-assistant/frontend/pull/26346 is going to be merged.


TimoPtr:
Screenshot are failing most probably because I've generated them on macOS and the verification runs on linux. You can see the diff
image

Once this PR have been reviewed I will either generate the image on linux or I will update the threshold to make the tests pass.

app/src/main/kotlin/io/homeassistant/companion/android/util/vehicle/DomainChecks.kt
```diff
@@ -39,6 +40,10 @@ fun isVehicleDomain(entity: Entity): Boolean {
         canNavigate(entity)
 }
 
+fun isValidTileDomain(entity: Entity): Boolean {
```
jpelgrom: This function's location/package isn't logical, should be placed elsewhere

app/src/main/res/xml/changelog_master.xml
```diff
@@ -2,6 +2,7 @@
 <changelog xmlns:tools="http://schemas.android.com/tools"
     tools:ignore="MissingDefaultResource">
     <release version="2025.8.2 - Main" versioncode="3">
+        <change>Easily add entities to widgets or favorites via their more-info dialog settings</change>
```
jpelgrom: This should mention a version requirement

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/AddToAction.kt
```diff
@@ -0,0 +1,207 @@
...
+            Icon(
+                imageVector = ImageVector.vectorResource(R.drawable.ic_car),
+                contentDescription = null,
+                modifier = modifier,
+            )
+        }
+
+        @Composable
+        override fun text(): String {
+            return stringResource(commonR.string.add_to_android_auto_favorite)
```
jpelgrom: This should have a different string for Automotive as the feature is called driving favorites for that platform.

TimoPtr:
This PR will be reworked since we are most probably move the UI on the frontend side. All the logic of this PR will be kept except that the frontend will send us an event with the `entity_id` to get the list of potential actions. Then the frontend will send us the action to execute (like we are doing in the dialog in the current state of this PR).

app/src/main/res/xml/changelog_master.xml
TimoPtr: I don't know when this will be available in the frontend so we should postpone this then ? To not forget about it we could just comment this line.

app/src/main/kotlin/io/homeassistant/companion/android/util/vehicle/DomainChecks.kt
TimoPtr: 🤦🏻 I didn't look at the package I though it was a general file to validate a domain, wdyt about moving the whole file closer to the Entity definition?

jpelgrom: Everything else is quite specific about Auto(motive) support, no? I can understand moving the function you added, not the others. And it shouldn't be moved too close (common package).

app/src/main/res/xml/changelog_master.xml
jpelgrom: Considering your latest comment that it more may move to the frontend I wonder if it should be mentioned at all in the app changelog.

jpelgrom:
> This PR will be reworked since we are most probably move the UI on the frontend side. All the logic of this PR will be kept except that the frontend will send us an event with the `entity_id` to get the list of potential actions. Then the frontend will send us the action to execute (like we are doing in the dialog in the current state of this PR).

OK, until that is decided (not sure where) and implemented this PR can be drafted then?

app/src/main/kotlin/io/homeassistant/companion/android/util/vehicle/DomainChecks.kt
TimoPtr: Either we move into a common place all the 'isValid*Domain' or we have them into dedicated places like this one for each domain we are interested in.

TimoPtr: Either we move into a common place all the 'isValid*Domain' or we have them into dedicated places like this one for each domain we are interested in.

TimoPtr:
> > This PR will be reworked since we are most probably move the UI on the frontend side. All the logic of this PR will be kept except that the frontend will send us an event with the `entity_id` to get the list of potential actions. Then the frontend will send us the action to execute (like we are doing in the dialog in the current state of this PR).
> 
> OK, until that is decided (not sure where) and implemented this PR can be drafted then?

Yes indeed I can't do it on my phone but you can or I'll do it tomorrow.

dshokouhi:
> All the logic of this PR will be kept except that the frontend will send us an event with the `entity_id` to get the list of potential actions. Then the frontend will send us the action to execute (like we are doing in the dialog in the current state of this PR).

This is out of scope for this PR but I always wondered if it was the correct approach for the app and frontend to communicate like so for every action. Open app and it says what can and cannot be supported, then now it looks like even further communication is requested for whats supported at different actions. Would it make more sense for the app to first register with the server what it can and cannot support like in some JSON format so we rely a bit less on external bus for whats supported and keep it strictly for actions (i.e. opening settings, triggering dialogs etc....). The features that are supported are on a per device and app version level so the initial communication when we do (re)registration should contain everything needed up until needing to send an action to the app or vice versa. For example if a device doesnt have NFC then it will never have it, so why ask every time. Unless its a user facing setting some of these things will change very little or only on app update.

app/src/main/kotlin/io/homeassistant/companion/android/util/vehicle/DomainChecks.kt
jpelgrom: Are you talking about Home Assistant domains here?

In case it wasn't obvious from the last comments, I'd favor a different file for the feature. Keep things contained and specific.

jpelgrom:
> Yes indeed I can't do it on my phone but you can or I'll do it tomorrow.

Use a browser on your phone 😉 Changed

TimoPtr:
@jpelgrom I've addressed your comments and the frontend has been updated.

app/src/main/res/xml/changelog_master.xml
TimoPtr: I rework a bit the wording I think we should still keep it there and also in the frontend.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/camera/CameraWidgetConfigureActivity.kt
```diff
@@ -28,6 +32,18 @@ import timber.log.Timber
 
 @AndroidEntryPoint
 class CameraWidgetConfigureActivity : BaseWidgetConfigureActivity<CameraWidgetEntity, CameraWidgetDao>() {
+    companion object {
+        private const val FOR_ENTITY = "for_entity"
```
jpelgrom: I'm wondering whether this should be part of the BaseWidgetConfigureActivity instead if it's going to exist for multiple of these activities.

app/src/main/res/xml/changelog_master.xml
```diff
@@ -2,6 +2,7 @@
 <changelog xmlns:tools="http://schemas.android.com/tools"
     tools:ignore="MissingDefaultResource">
     <release version="2025.11.2 - Main" versioncode="3">
+        <change>Starting from Home Assistant 2025.12 easily add entities to widgets or favorites via their more-info dialog settings</change>
```
```suggestion
        <change>Easily add entities to widgets or favorites from the more info dialog (requires Home Assistant 2025.12 or newer)</change>
```

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
```diff
@@ -0,0 +1,191 @@
...
+    fun details(context: Context): String? = null
+
+    /**
+     * Action to add an entity to Android Auto or Android Automotive favorites.
+     *
+     * The display text varies depending on whether the app is running on an automotive head unit
+     * or on a phone.
+     */
+    @Serializable
+    data object AndroidAutoFavorite : EntityAddToAction {
```
jpelgrom, line 72: It's a bit weird to me to have these _inside_ the interface

app/src/test/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToActionTest.kt
```diff
@@ -0,0 +1,167 @@
...
+        val action = EntityAddToAction.Shortcut(enabled = false)
+
+        assertEquals("mdi:open-in-new", action.mdiIcon)
+        assertEquals("Add to home screen", action.text(context))
+        assertEquals(false, action.enabled)
+        assertEquals("Maximum shortcuts reached", action.details(context))
+    }
+
+    @Test
+    fun `Given Tile when testing attributes then return correct values`() {
```
jpelgrom, line 82: I'm honestly not sure what the value of these tests for always available actions is when the possible options are defined in code. Avoid people changing them? It is a string and defaults.

common/src/main/res/values/strings.xml
```diff
@@ -15,6 +15,17 @@
     <string name="add_favorite">Add favorite</string>
     <string name="add_action_data_field">Add field</string>
     <string name="add_shortcut">Add shortcut</string>
+    <string name="add_to_android_auto_favorite">Add to Android Auto favorite</string>
```
```suggestion
    <string name="add_to_android_auto_favorite">Add to Android Auto favorites</string>
```

common/src/main/res/values/strings.xml
```diff
@@ -15,6 +15,17 @@
...
+    <string name="add_to_android_auto_driving_favorite">Add to Driving favorites</string>
+    <string name="add_to_camera_widget">Add camera/picture to home screen</string>
+    <string name="add_to_entity_widget">Add entity state to home screen</string>
+    <string name="add_to_media_player_widget">Add media player to home screen</string>
+    <string name="add_to_shortcut">Add shortcut</string>
```
```suggestion
    <string name="add_to_shortcut">Add pinned shortcut</string>
```

It's different from a dynamic shortcut

common/src/main/res/values/strings.xml
```diff
@@ -15,6 +15,17 @@
...
+    <string name="add_to_shortcut_limit">Already 5 shortcuts set</string>
+    <string name="add_to_tile">Add tile</string>
```
```suggestion
    <string name="add_to_tile">Add quick settings tile</string>
```

"tile" may not be obvious where it ends up, also because different Android versions/manufacturers display system UI differently

common/src/main/res/values/strings.xml
```diff
@@ -15,6 +15,17 @@
     <string name="add_favorite">Add favorite</string>
     <string name="add_action_data_field">Add field</string>
     <string name="add_shortcut">Add shortcut</string>
+    <string name="add_to_android_auto_favorite">Add to Android Auto favorite</string>
+    <string name="add_to_android_auto_driving_favorite">Add to Driving favorites</string>
```
```suggestion
    <string name="add_to_android_auto_driving_favorite">Add to driving favorites</string>
```

app/src/test/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToActionTest.kt
```diff
@@ -0,0 +1,167 @@
...
+
+    @AfterEach
+    fun tearDown() {
+        unmockkStatic("io.homeassistant.companion.android.common.util.ContextExtKt")
+    }
+
+    @Test
+    fun `Given automotive context when testing AndroidAutoFavorite then return correct attributes`() {
+        every { context.isAutomotive() } returns true
+        every { context.getString(commonR.string.add_to_android_auto_driving_favorite) } returns "Add to driving favorites"
```
jpelgrom, line 33: These mock strings are different from the actual string, what is the point in testing them?

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
```diff
@@ -0,0 +1,191 @@
...
+            return if (enabled) null else context.getString(commonR.string.add_to_shortcut_limit)
+        }
+    }
+
+    /**
+     * Action to add an entity to Quick Settings as a tile.
+     */
+    @Serializable
+    data object Tile : EntityAddToAction {
+        override val mdiIcon: String = "mdi:cog"
```
jpelgrom: I'd use the same icon as in the app's settings, or if you think that one is wrong I don't think a cog makes sense as an icon for a quick settings tile.

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
```diff
@@ -0,0 +1,191 @@
...
+            return context.getString(commonR.string.add_to_tile)
+        }
+    }
+
+    /**
+     * Action to add a general entity widget to the home screen.
+     */
+    @Serializable
+    data object EntityWidget : EntityAddToAction {
+        override val mdiIcon: String = "mdi:widgets"
```
jpelgrom: I'd use the same icon as in the app's settings

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
```diff
@@ -0,0 +1,191 @@
...
+            return context.getString(commonR.string.add_to_camera_widget)
+        }
+    }
+
+    /**
+     * Action to add a to-do list widget to the home screen.
+     */
+    @Serializable
+    data object TodoWidget : EntityAddToAction {
+        override val mdiIcon: String = "mdi:checkbox-marked-circle-plus-outline"
```
jpelgrom: I'd use the same icon as in the app's settings

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
```diff
@@ -0,0 +1,191 @@
...
+ * Represents an action that allows users to add a Home Assistant entity to various Android
+ * platform features (widgets, shortcuts, tiles, etc.) or connected devices (watches, automotive).
+ *
+ * This sealed interface is used for bidirectional communication between the Android app and the
+ * Home Assistant frontend. The frontend can query available actions, and the user can select
+ * which action to perform for a specific entity.
+ *
+ * Each implementation represents a different platform integration point where entities can be
+ * added, such as Android Auto favorites, home screen shortcuts, Quick Settings tiles, or various
+ * widget types.
```
jpelgrom:
```suggestion
```

You've already said this in the first paragraph

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
```diff
@@ -0,0 +1,191 @@
...
+            return if (context.isAutomotive()) {
+                context.getString(commonR.string.add_to_android_auto_driving_favorite)
+            } else {
+                context.getString(commonR.string.add_to_android_auto_favorite)
+            }
+        }
+    }
+
+    /**
+     * Action to add an entity shortcut to the device home screen.
```
jpelgrom:
```suggestion
     * Action to add a pinned shortcut to the entity to the device home screen.
```

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
```diff
@@ -0,0 +1,191 @@
...
+    data object Tile : EntityAddToAction {
+        override val mdiIcon: String = "mdi:cog"
+
+        override fun text(context: Context): String {
+            return context.getString(commonR.string.add_to_tile)
+        }
+    }
+
+    /**
+     * Action to add a general entity widget to the home screen.
```
jpelgrom:
```suggestion
     * Action to add an entity state widget to the home screen.
```

app/src/main/kotlin/io/homeassistant/companion/android/webview/externalbus/ExternalBusMessage.kt
```diff
@@ -62,9 +69,60 @@ class ExternalConfigResponse(
...
+    val details: String?,
+    val mdiIcon: String,
+) {
+    companion object {
+        /**
+         * We encode the app payload into Base64 so that we are sure that the data remains the same while going
+         * to the frontend and coming back.
+         */
+        fun fromAction(context: Context, action: EntityAddToAction): ExternalEntityAddToAction {
+            return ExternalEntityAddToAction(
```
jpelgrom:
```suggestion
        fun fromAction(context: Context, action: EntityAddToAction): ExternalEntityAddToAction {
            // Encode the app payload into Base64 to ensure that the data remains the same while going
            // to the frontend and coming back.
            return ExternalEntityAddToAction(
```

This doesn't describe the function so shouldn't be on the function but inside

app/src/test/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToHandlerTest.kt
```diff
@@ -0,0 +1,254 @@
...
+        val entityId = "light.nonexistent"
+        coEvery { integrationRepository.getEntity(entityId) } returns null
+
+        val actions = handler.actionsForEntity(entityId)
+
+        assertEquals(emptyList<EntityAddToAction>(), actions)
+    }
+
+    @Test
+    fun `Given standard entityId when getting potentialActions then it emits first empty then EntityWidget`() = runTest {
```
jpelgrom: Except it doesn't emit an empty list at first, nor do you test for it (this applies to most of the following tests)

TimoPtr:
> > the frontend has been updated
> 
> Frontend PR is still pending review, no?
> 
> I'm also still seeing commented out code/unimplemented actions. Let's remove that, there's a lot going on. You already have issues to track them.

It should be merged after the release of Wednesday. I meant that the PR on the frontend has been updated not the frontend itself.

I'll remove the todos.

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
TimoPtr: It's a sealed class, I'm used to writing it like this but we could put it outside 👍. It just needs to be in the same package.

app/src/test/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToActionTest.kt
TimoPtr: Just testing that we are using the right string id. I don't want the test to use the translation.

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
TimoPtr: Do you happen to know their MDI name?

app/src/test/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToHandlerTest.kt
TimoPtr: It's a leftover from the previous implementation good catch 👍. 

I used Claude to try to help me spot tests that I've missed but it seems that it was not 'smart' enough to look at previous tests.

app/src/test/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToActionTest.kt
TimoPtr: Mostly if we do refactoring things, these tests could be helpful to make sure we don't mess things up.

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
jpelgrom: Not a dealbreaker and may just be personal preference, my (limited) experience with sealed classes put it outside like in the official docs. Inside makes me think of enums ;)

app/src/test/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToActionTest.kt
jpelgrom: In that case I would return something different so it is also clear to future contributors that you do not test for that and do not need to keep strings in sync.

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
jpelgrom: [`mdi:tune`](https://pictogrammers.com/library/mdi/icon/tune/)

app/src/test/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToActionTest.kt
jpelgrom: Don't mess up what? It's a couple of values defined in one place. You are using a sealed interface so no way to break an ID string and all possible variants are already handled.

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
TimoPtr: I've used sealed class when they first introduced it and it was mostly to "replace" enums. That should be why I'm used to do that 👍🏻 but fine for me, one less indentation.

TimoPtr: Wait no, it has a reason ^^ if we don't put it in the sealed class the usage is going to be 

```kotlin
val action = Tile
```

where if it is inside

```kotlin
val action = EntityAddToAction.Tile
```
I think it's better for context while working with an action, don't you think?
(even if we could make an import that removes the first part)

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
TimoPtr: mdi:clipboard-list

app/src/test/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToActionTest.kt
TimoPtr: Like adding some logic within the text() or icon() methods, like we do in automotive. The tests are there already let see if it becomes painful to maintain we can remove them.

app/src/main/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToAction.kt
jpelgrom: Now it's a naming question. Changing them so you have `TileAddToAction`, `CameraWidgetAddToAction`, etc. would also work in my opinion. 

For now fine either way, let's not have the PR get stuck on this discussion.

PR #5641, opened 2025-08-14
Trigger prepareNextRelease using Github Action API to bypass limitation

TimoPtr:
## Summary

The prepareNextRelease workflow is not triggered since Github added a security that prevent recursive workflows when we use the token `GITHUB_TOKEN` available in the action. 

> When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. 

https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#triggering-a-workflow-from-a-workflow

In order to by-pass this limitation I've manually trigger the workflow from the other workflow (weekly and monthly).

I've also set a concurrency limit to avoid running the workflow in //.

### DIFF
--- a/.github/workflows/monthly.yml
+++ b/.github/workflows/monthly.yml
@@ -19,3 +19,14 @@ jobs:
         uses: rickstaa/action-create-tag@v1.7.2
         with:
           tag: ${{ steps.current-time.outputs.year }}.${{ steps.current-time.outputs.month }}.0
+
+      - name: Trigger Prepare Next Release
+        uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.actions.createWorkflowDispatch({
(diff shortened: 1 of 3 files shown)


jpelgrom:
Fine with merging for testing it but shouldn't you use another token?

TimoPtr:
> Fine with merging for testing it but shouldn't you use another token?

We need another token if we when to use the Github Action events here we use the API and you can see an example where we are actually already doing this. It makes things a bit more manual and error prone if we rename a workflow or if we add another one. It's a best effort solution I would say.

Issue #4018, opened 2023-11-15
Hide Changelog Popup


histefanhere:
**Is your feature request related to a problem? Please describe.**

Upon opening the app after an update (which in most cases is automatic), the changelog of the update appears. This _was_ an acceptable(ish) way of informing users of changes when the issue was brought up 3 years ago (#1162) and when the feature was introduced a year ago (#2195). Immediately after, a suggestion was made to make this behavior togglable (#2280), but it was denied. Since then, the target audience of Home Assistant has broadened and now covers the more general user - examples of this include the releases of [HA Yellow](https://www.home-assistant.io/yellow/) and [HA Green](https://www.home-assistant.io/green/) which are advertised as "plug and play" solutions with "seamless" mobile apps. Now, burdening _all_ mobile users with the technical details of updates through their changelogs isn't acceptable, and there should be an option to disable this behavior.

The concern of making the application more accessible for friends and family members, especially those who are older with technical illiteracy and weak English (because it is by far not the only language in the world), still persists and is more important now than ever in the history of HA. The only chance my 80-year-old Grandma, whom I have to read out and explain emails for, has of using the app comes from its friendly, house-looking logo. Hell, I'm sure the great majority of people who aren't familiar with software development have never even heard of a "changelog" before because it's so irregular to be _forced_ to see it. Google Home, for example, doesn't notify users of updates, especially complex technical ones. Even Home Assistant itself doesn't notify _all users of the platform_ that there has been an update to it, and its breaking changes are arguably much more important than the apps.

 **The changelog should definitely not be hidden from sight for all users** - I understand the arguments for why the changelog should appear as, I quote, "changes to the app impact every user and not just the admin" - but they do not necessarily _concern_ all users. If anything happens to the companion app to my Grandma or any family member of mine, I am informed anyway in the same way that an issue with HA itself like cameras being down or sensors being unavailable get relayed back to me by confused relatives, regardless of if they were prompet with a changelog or not. I am a strong advocate for the transparency of applications, and I enjoy reading the changelogs for new features that I have been waiting on. However, not every user is in this boat and there is no reason why this cannot be accommodated for on a case-by-case basis.

**Describe the solution you'd like**

An opt-out toggle switch in the companion app settings that hides the changelog popup when the app is first turned on after an update. Consider warning the user that breaking changes may be introduced that produce unintended behavior in the app, and ask if the user is sure they want to hide the dialog from appearing following updates.

**Describe alternatives you've considered, if any**

 
[...]

PR #5651, opened 2025-08-16
Add feature to allow not displaying change log on update

loganrosen:
## Summary
As requested in #4018, adding a setting to allow not showing the change log popup on every update of the Home Assistant companion app. The default is to show it, but users can now opt out with this new toggle.

## Checklist

## Screenshots

image
image

## Any other notes

As part of this change, in `PrefsRepositoryImpl`, I moved the line where `migrationChecked` is set to `true` outside of the conditional of `currentVersion == null || currentVersion < 1`, as I experienced a deadlock when trying to toggle the new setting without this change. This seems more logically correct anyway because the migration was checked at that point, regardless of whether the version condition matched – please let me know if this should be reverted/addressed differently.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/util/ChangeLog.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/util/ChangeLog.kt
@@ -6,12 +6,24 @@ import android.view.ContextThemeWrapper
 import info.hannes.changelog.ChangeLog
 import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.common.data.prefs.NightModeTheme
+import io.homeassistant.companion.android.common.data.prefs.PrefsRepository
 import io.homeassistant.companion.android.themes.NightModeManager
 import javax.inject.Inject
 import kotlinx.coroutines.runBlocking
 
-class ChangeLog @Inject constructor(val nightModeManager: NightModeManager) {
+class ChangeLog @Inject constructor(
+    val nightModeManager: NightModeManager,
+    private val prefsRepository: PrefsRepository
+) {
+    private fun createChangeLog(context: Context): ChangeLog {
+        return ChangeLog(context)
+    }
     fun showChangeLog(context: Context, forceShow: Boolean) {
+        // Check if user has enabled change log popup or this is a forced show
+        if (!forceShow && !runBlocking { prefsRepository.isChangeLogPopupEnabled() }) {
+            return
+        }
+
         val isDarkTheme = when (runBlocking { nightModeManager.getCurrentNightMode() }) {
             NightModeTheme.ANDROID, NightModeTheme.SYSTEM -> {
                 val nightModeFlags =
@@ -27,7 +39,7 @@ class ChangeLog @Inject constructor(val nightModeManager: NightModeManager) {
                 darkThemeChangeLog.fullLogDialog.show()
             }
         } else {
-            val changeLog = ChangeLog(context)
+            val changeLog = createChangeLog(context)
             if ((!changeLog.isFirstRunEver && changeLog.isFirstRun) || forceShow) {
                 changeLog.fullLogDialog.show()
             }
--- a/app/src/main/res/xml/preferences.xml
+++ b/app/src/main/res/xml/preferences.xml
@@ -224,6 +224,11 @@
     </PreferenceCategory>
     <PreferenceCategory
         android:title="@string/app_version_info">
+        <SwitchPreference
+            android:key="change_log_popup_enabled"
+            android:icon="@drawable/ic_changelog"
+            android:title="@string/enable_change_log_popup"
+            android:summary="@string/enable_change_log_popup_summary" />
         <Preference
             android:key="changelog_prompt"
             android:icon="@drawable/ic_changelog"
--- a/app/src/test/kotlin/io/homeassistant/companion/android/util/ChangeLogTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/util/ChangeLogTest.kt
@@ -1,0 +1,56 @@
+package io.homeassistant.companion.android.util
+
+import android.app.AlertDialog
+import android.content.Context
+import io.homeassistant.companion.android.common.data.prefs.PrefsRepository
+import io.homeassistant.companion.android.themes.NightModeManager
+import io.mockk.coEvery
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.spyk
+import io.mockk.verify
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+class ChangeLogTest {
+    private val context = mockk<Context>(relaxed = true)
+    private val prefsRepository = mockk<PrefsRepository>()
+    private val nightModeManager = mockk<NightModeManager>()
+    private val hannesChangeLog = mockk<info.hannes.changelog.ChangeLog>(relaxed = true)
+    private val dialog = mockk<AlertDialog>(relaxed = true)
+
+    @BeforeEach
+    fun setup() {
+        coEvery { nightModeManager.getCurrentNightMode() } returns io.homeassistant.companion.android.common.data.prefs.NightModeTheme.LIGHT
+        every { hannesChangeLog.fullLogDialog } returns dialog
+        every { hannesChangeLog.isFirstRun } returns true
+        every { hannesChangeLog.isFirstRunEver } returns false
+    }
+
+    @Test
+    fun `Given change log popup is not enabled when showing change log then popup is not shown`() {
+        coEvery { prefsRepository.isChangeLogPopupEnabled() } returns false
+        val changeLog = spyk(ChangeLog(nightModeManager, prefsRepository), recordPrivateCalls = true)
+        every { changeLog["createChangeLog"](any<Context>()) } returns hannesChangeLog
+        changeLog.showChangeLog(context, forceShow = false)
+        verify(exactly = 0) { dialog.show() }
+    }
+
+    @Test
+    fun `Given change log popup is enabled when showing change log then popup is shown`() {
+        coEvery { prefsRepository.isChangeLogPopupEnabled() } returns true
+        val changeLog = spyk(ChangeLog(nightModeManager, prefsRepository), recordPrivateCalls = true)
+        every { changeLog["createChangeLog"](any<Context>()) } returns hannesChangeLog
+        changeLog.showChangeLog(context, forceShow = false)
+        verify(exactly = 1) { dialog.show() }
+    }
+
+    @Test
+    fun `Given forceShow is true when showing change log then popup is shown regardless of enabled setting`() {
+        coEvery { prefsRepository.isChangeLogPopupEnabled() } returns false
+        val changeLog = spyk(ChangeLog(nightModeManager, prefsRepository), recordPrivateCalls = true)
+        every { changeLog["createChangeLog"](any<Context>()) } returns hannesChangeLog
+        changeLog.showChangeLog(context, forceShow = true)
+        verify(exactly = 1) { dialog.show() }
+    }
+}
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
@@ -41,6 +41,7 @@ private const val PREF_AUTO_FAVORITES = "auto_favorites"
 private const val PREF_LOCATION_HISTORY_DISABLED = "location_history"
 private const val PREF_IMPROV_PERMISSION_DISPLAYED = "improv_permission_displayed"
 private const val PREF_GESTURE_ACTION_PREFIX = "gesture_action"
+private const val PREF_CHANGE_LOG_POPUP_ENABLED = "change_log_popup_enabled"
 
 /**
  * This class ensure that when we use the local storage in [PrefsRepositoryImpl] the migrations has been made
@@ -84,8 +85,8 @@ private class LocalStorageWithMigration(
                     }
 
                     localStorage.putInt(MIGRATION_PREF, MIGRATION_VERSION)
-                    migrationChecked.set(true)
                 }
+                migrationChecked.set(true)
             }
         }
     }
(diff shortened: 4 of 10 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/util/ChangeLog.kt
```diff
@@ -6,12 +6,24 @@ import android.view.ContextThemeWrapper
...
+class ChangeLog @Inject constructor(
+    val nightModeManager: NightModeManager,
+    private val prefsRepository: PrefsRepository,
+) {
+    private fun createChangeLog(context: Context): ChangeLog {
+        return ChangeLog(context)
+    }
     fun showChangeLog(context: Context, forceShow: Boolean) {
+        // Check if user has enabled change log popup or this is a forced show
+        if (!forceShow && !runBlocking { prefsRepository.isChangeLogPopupEnabled() }) {
```
TimoPtr: We should not use runBlocking it will block the UI on some device especially on low end device (or when the device is already doing a lot). It won't happen all the time but it might if the XML file behind the SharedPref is not yet loaded in memory.

This is resulting into an ANR being reported in our crash console on the play store. I'm trying to get rid of all the runBlocking since we have a lot of this ANR unfortunately and it makes hard to spot more important issue.

From the user perspective he might not see the issue at all except the UI freezing a bit but we should not accept this.

app/src/test/kotlin/io/homeassistant/companion/android/util/ChangeLogTest.kt
```diff
@@ -0,0 +1,56 @@
...
+        coEvery { nightModeManager.getCurrentNightMode() } returns io.homeassistant.companion.android.common.data.prefs.NightModeTheme.LIGHT
+        every { hannesChangeLog.fullLogDialog } returns dialog
+        every { hannesChangeLog.isFirstRun } returns true
+        every { hannesChangeLog.isFirstRunEver } returns false
+    }
+
+    @Test
+    fun `Given change log popup is not enabled when showing change log then popup is not shown`() {
+        coEvery { prefsRepository.isChangeLogPopupEnabled() } returns false
+        val changeLog = spyk(ChangeLog(nightModeManager, prefsRepository), recordPrivateCalls = true)
```
TimoPtr: I didn't yet look at the impl but from the test perspective I prefer to expose a function with VisibleForTesting rather than using a spy. We should focus on testing the public interface and the behavior of the target.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
```diff
@@ -84,8 +85,8 @@ private class LocalStorageWithMigration(
                     }
 
                     localStorage.putInt(MIGRATION_PREF, MIGRATION_VERSION)
-                    migrationChecked.set(true)
                 }
+                migrationChecked.set(true)
```
TimoPtr: Good catch 👍🏻

TimoPtr: Could you cover this with a unit test?

app/src/main/res/xml/preferences.xml
```diff
@@ -224,6 +224,11 @@
     </PreferenceCategory>
     <PreferenceCategory
         android:title="@string/app_version_info">
+        <SwitchPreference
```
TimoPtr: I’d avoid putting this so high in the list. Most users benefit from the changelog, which we try to keep understandable for Home Assistant users; while some may not follow it, my feeling tells me that it is a minority (that we should still consider of course). We really want to highlight the app’s added value over a plain webpage, we shouldn’t encourage disabling the changelog. I would be in favor to put it bellow View full change log.

What's your opinion @dshokouhi @jpelgrom

dshokouhi: I agree @TimoPtr below the change log makes the most sense

jpelgrom: Agreed as well, maybe also without an icon and simply "Show change log after updating"?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
```diff
@@ -316,6 +317,15 @@ class PrefsRepositoryImpl @Inject constructor(
         localStorage().putString("${PREF_GESTURE_ACTION_PREFIX}_${gesture.name}", action.name)
     }
 
+    override suspend fun isChangeLogPopupEnabled(): Boolean {
+        // Default to true if not set
+        return localStorage().getBoolean(PREF_CHANGE_LOG_POPUP_ENABLED) ?: true
```
jpelgrom: Did you mean `getBooleanOrNull`? Because `getBoolean` has a non-null return type. And if so, the comment seems unnecessary.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImplTest.kt
```diff
@@ -63,4 +63,35 @@ class PrefsRepositoryImplTest {
 
         assertEquals(GestureAction.NONE, result)
     }
+
+    @Test
+    fun `Given no preference set when checking change log popup enabled then default is true`() = runTest {
+        coEvery { localStorage.getBoolean("change_log_popup_enabled") } returns true
```
jpelgrom: You're testing no preference set according to the function name, but here you're mocking a Boolean return value for the preference (so it _is_ set)?

app/src/main/kotlin/io/homeassistant/companion/android/util/ChangeLog.kt
loganrosen: Makes sense. Updated the function/call sites to avoid blocking accordingly.

app/src/test/kotlin/io/homeassistant/companion/android/util/ChangeLogTest.kt
loganrosen: Updated to use `@VisibleForTesting` helper methods instead of using a spy.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
loganrosen: Done!

app/src/main/res/xml/preferences.xml
loganrosen: Updated based on your feedback.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
loganrosen: Yes, good catch. Updated, thanks.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImplTest.kt
loganrosen: Yep, similar issue to the one you caught above. Updated, thank you.

app/src/main/kotlin/io/homeassistant/companion/android/util/ChangeLog.kt
```diff
@@ -22,12 +41,12 @@ class ChangeLog @Inject constructor(val nightModeManager: NightModeManager) {
...
-            val darkThemeChangeLog = DarkThemeChangeLog(context)
+            val darkThemeChangeLog = createDarkThemeChangeLog(context)
             if ((!darkThemeChangeLog.isFirstRunEver && darkThemeChangeLog.isFirstRun) || forceShow) {
                 darkThemeChangeLog.fullLogDialog.show()
             }
         } else {
-            val changeLog = ChangeLog(context)
+            val changeLog = createChangeLog(context)
             if ((!changeLog.isFirstRunEver && changeLog.isFirstRun) || forceShow) {
                 changeLog.fullLogDialog.show()
```
TimoPtr: You've made the whole function suspend which is nice, but you need to make sure that `.show()` is invoked on the Main thread within this function. We cannot assume that it's always going to be invoked from Main we should enforce it.

Please double check what happens if you force IO dispatcher for instance for the View creation and also for show() invocation.

app/src/main/kotlin/io/homeassistant/companion/android/util/ChangeLog.kt
```diff
@@ -3,16 +3,35 @@ package io.homeassistant.companion.android.util
...
 import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.common.data.prefs.NightModeTheme
+import io.homeassistant.companion.android.common.data.prefs.PrefsRepository
 import io.homeassistant.companion.android.themes.NightModeManager
 import javax.inject.Inject
-import kotlinx.coroutines.runBlocking
 
-class ChangeLog @Inject constructor(val nightModeManager: NightModeManager) {
-    fun showChangeLog(context: Context, forceShow: Boolean) {
-        val isDarkTheme = when (runBlocking { nightModeManager.getCurrentNightMode() }) {
```
TimoPtr, line 15: One less runBLocking 💪🏻

app/src/main/res/xml/preferences.xml
```diff
@@ -229,6 +229,10 @@
             android:icon="@drawable/ic_changelog"
             android:title="@string/show_changelog"
             android:summary="@string/show_changelog_summary" />
+        <SwitchPreference
```
TimoPtr, line 232: Please update the screenshot in the PR description 🙏🏻

app/src/test/kotlin/io/homeassistant/companion/android/util/ChangeLogTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    private val prefsRepository = mockk<PrefsRepository>()
+    private val nightModeManager = mockk<NightModeManager>()
+    private val hannesChangeLog = mockk<info.hannes.changelog.ChangeLog>(relaxed = true)
+    private val darkThemeChangeLog = mockk<DarkThemeChangeLog>(relaxed = true)
+    private val dialog = mockk<AlertDialog>(relaxed = true)
+    private lateinit var changeLog: TestableChangeLog
+
+    @BeforeEach
+    fun setup() {
+        coEvery { nightModeManager.getCurrentNightMode() } returns io.homeassistant.companion.android.common.data.prefs.NightModeTheme.LIGHT
```
TimoPtr: Could you make this a static import?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
TimoPtr: thanks 🙏🏻

app/src/main/res/xml/preferences.xml
```diff
@@ -229,6 +229,10 @@
...
+            android:key="change_log_popup_enabled"
+            android:title="@string/enable_change_log_popup"
```
TimoPtr, line 235: You forgot to bring back the icon?

jpelgrom: I suggested no more icon in https://github.com/home-assistant/android/pull/5651#discussion_r2282638435, because I thought the double icons looked weird and it is a "sub-setting" of the main changelog one.

TimoPtr: image
It feels weird because none of the other items in this list has this "sub-settings" usually it's another page.

jpelgrom: Looking at your screenshot: yes, that doesn't look as good as I imagined, also with the repeated "Show" for both settings.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
TimoPtr: @loganrosen I did bring your changes into another PR https://github.com/home-assistant/android/pull/5690 to get it merge ASAP since it might be the cause of some ANRs on some devices.

app/src/main/res/xml/preferences.xml
loganrosen: Added the icon back. Please see the updated screenshots in the PR description.

app/src/test/kotlin/io/homeassistant/companion/android/util/ChangeLogTest.kt
loganrosen: Done!

app/src/main/res/xml/preferences.xml
loganrosen: Updated the screenshots in the description.

app/src/main/kotlin/io/homeassistant/companion/android/util/ChangeLog.kt
loganrosen: I've updated the call to ensure it happens on the Main thread.

app/src/main/res/xml/preferences.xml
TimoPtr: Let's wait for @jpelgrom if he comes up with a nice idea :)

jpelgrom: My alternative idea would be to show it in the dialog itself, a switch at the bottom "Show after updating", but at that point we should get rid of the entire library and redo it ourselves in Compose IMO. Out of scope for now, let's stick with this option :)

app/src/main/kotlin/io/homeassistant/companion/android/util/ChangeLog.kt
```diff
@@ -21,15 +42,19 @@ class ChangeLog @Inject constructor(val nightModeManager: NightModeManager) {
...
+                val darkThemeChangeLog = createDarkThemeChangeLog(context)
+                if ((!darkThemeChangeLog.isFirstRunEver && darkThemeChangeLog.isFirstRun) || forceShow) {
+                    darkThemeChangeLog.fullLogDialog.show()
+                }
+            } else {
+                val changeLog = createChangeLog(context)
+                if ((!changeLog.isFirstRunEver && changeLog.isFirstRun) || forceShow) {
+                    changeLog.fullLogDialog.show()
+                }
             }
```
jpelgrom:
```suggestion
            val changeLog = if (isDarkTheme) {
                createDarkThemeChangeLog(context)
            } else {
                createChangeLog(context)
            }
            if ((!changeLog.isFirstRunEver && changeLog.isFirstRun) || forceShow) {
                changeLog.fullLogDialog.show()
            }
```

Keep logic for first run / force show in one place

Issue #5221, opened 2025-04-16
Help Improve Our Codebase by Fixing Linter Warnings!


TimoPtr:
We've recently enforced new linting rules to enhance code quality and maintain consistency across our project. To prevent an overwhelming number of changes in a single pull request (PR), we've added the current warnings to baseline files.

We invite you to help us improve the codebase by addressing these linter issues! This is a fantastic opportunity to get familiar with the codebase and contribute to the project.

## Guidelines:

- Check out our [Linter Documentation](https://developers.home-assistant.io/docs/android/linter) to understand the lint rules and how to run them.

- Scope Your Work:
  - No need to fix all issues—we prefer manageable PRs over giant ones.
  - Focus on fixing linter warnings within a specific Kotlin package.

- Submitting Pull Requests:
  - Don't create a separate PR for each issue.
  - Do group related fixes together in a single PR.

- Some issues in the `:automotive` Gradle module may not be relevant due to its dependency on `:app` we recommend you to Ignore linter issues from `:automotive` for now.

- You can find the `lint-baseline.xml` file in each Gradle module directory.

**We appreciate your help and look forward to your contributions!**

If you have questions or need assistance, check our [documentation](https://developers.home-assistant.io/docs/android/).

PR #5659, opened 2025-08-17
Fix various `ObsoleteSdkInt` and `SuspiciousImport` lint warnings

loganrosen:
## Summary

Fixing various `ObsoleteSdkInt` and `SuspiciousImport` lint warnings and updating the baseline accordingly, as requested in #5221. I've also configured the Ktlint plugin in Android Studio to automatically fix formatting issues that it finds.

## Checklist


app/src/main/kotlin/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -271,18 +271,14 @@ class CarSensorManager :
...
-            emptyList()
-        }
+        return sensorsList
     }
 
     @Suppress("KotlinConstantConditions")
     override fun hasSensor(context: Context): Boolean {
         this.latestContext = context.applicationContext
 
-        return isAutomotive || (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && BuildConfig.FLAVOR == "full")
```
dshokouhi, line 285: This check is required to prevent showing the sensor on incompatible android versions.

.idea/ktlint-plugin.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="com.nbadal.ktlint.KtlintProjectSettings">
+    <ktlintMode>DISTRACT_FREE</ktlintMode>
```
TimoPtr, line 4: I'm not sure about this one. Could you make a small video of how it behaves?

app/src/main/kotlin/io/homeassistant/companion/android/sensors/CarSensorManager.kt
loganrosen: Okay, I think I understand what's happening here. The lint warning is showing up because `CarSensorManager` is annotated as `@RequiresApi(Build.VERSION_CODES.O)`, so it theoretically shouldn't be instantiated on any Android version lower than O, which would make these conditionals unnecessary. However, [`SensorReceiver` instantiates it regardless of the API version](https://github.com/home-assistant/android/blob/9100fc3ae6c617a924047bbe6aa8c1c69bb63d64/app/src/main/kotlin/io/homeassistant/companion/android/sensors/SensorReceiver.kt#L59), which is causing [a _different_ `NewApi` lint warning that's part of the baseline](https://github.com/home-assistant/android/blob/9100fc3ae6c617a924047bbe6aa8c1c69bb63d64/app/lint-baseline.xml#L158-L167).

We have two options accordingly:

1. Remove the `RequiresApi` annotation on `CarSensorManager` so that it can/should be instantiated on Android versions below O. Reinstate the removed conditionals in this PR.
2. Keep the annotation but change `SensorReceiver` to only instantiate `CarSensorManager` if the Android version is >= O (would this cause problems?).

Which option do you recommend?

.idea/ktlint-plugin.xml
loganrosen: It applies the changes on save. Here's a video of me removing the trailing comma and then hitting save, upon which it invokes Ktlint and restores the trailing comma:

https://github.com/user-attachments/assets/72375079-9ddb-4e37-b3f6-cc64b50715f5

app/src/main/kotlin/io/homeassistant/companion/android/sensors/CarSensorManager.kt
jpelgrom: Option 1 would be preferred, as that better reflects how our sensors architecture works. I would argue that the annotation was an error, not intentional. The class itself already has checks to prevent usage of the API on older Android versions.

loganrosen: Updated accordingly, thanks.

app/src/main/kotlin/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -344,6 +349,9 @@ class CarSensorManager :
 
     @androidx.annotation.OptIn(androidx.car.app.annotations.ExperimentalCarApi::class)
     private fun setListener(l: Listener, enable: Boolean) {
+        if (!areCarSensorApisAvailable) {
+            return
+        }
```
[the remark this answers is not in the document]
loganrosen: This is necessary to avoid a lint warning for calling `HaCarAppService.carInfo` on the next line, as the `HaCarAppService` class is annotated as `@RequiresApi(Build.VERSION_CODES.O)`.

.idea/ktlint-plugin.xml
TimoPtr: I'm fine trying it. If it becomes an issue for any reason we will disable it again 👍🏻

app/src/main/kotlin/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -219,6 +217,9 @@ class CarSensorManager :
 
     private lateinit var latestContext: Context
 
+    @Suppress("ObsoleteSdkInt") // Needed for app module (minSdk 21) but obsolete for automotive (minSdk 29)
+    private val areCarSensorApisAvailable get() = Build.VERSION.SDK_INT >= Build.VERSION_CODES.O
```
TimoPtr:
```suggestion
    private val areCarSensorApisAvailable = Build.VERSION.SDK_INT >= Build.VERSION_CODES.O
```

I don't know why we have this ugly pattern in this class if it is to get different values over time it shouldn't be an attributes but a function (CC @jpelgrom ). At least for the one you've added should be updated.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -49,34 +49,29 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository)
         // Let the WebViewActivity know the endpoint requires TLS Client Auth
         isTLSClientAuthNeeded = true
 
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
```
TimoPtr, line 52: I'm not sure we should remove this check. I think we should remove the `@RequiresApi(Build.VERSION_CODES.M)` instead, since this API is called by the system not by us.

loganrosen: Sounds good, updated. I had to suppress the lint rule still because these source files are part of the `app` and `automotive` modules, each of which have their own `minSdk` versions. Because the `automotive` minimum version is higher, this warns in that module only.

Have we considered moving these files to a shared module that both `app` and `automotive` depend on instead of sharing the sources between them?

app/src/main/kotlin/io/homeassistant/companion/android/sensors/CarSensorManager.kt
loganrosen: Updated accordingly.

app/src/main/kotlin/io/homeassistant/companion/android/sensors/CarSensorManager.kt
```diff
@@ -219,6 +218,9 @@ class CarSensorManager :
 
     private lateinit var latestContext: Context
 
+    @SuppressLint("ObsoleteSdkInt") // Needed for app module (minSdk 21) but obsolete for automotive (minSdk 29)
+    private val areCarSensorApisAvailable = Build.VERSION.SDK_INT >= Build.VERSION_CODES.O
```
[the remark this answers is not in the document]
loganrosen: This wouldn't achieve the intended effect.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
TimoPtr: > Have we considered moving these files to a shared module that both app and automotive depend on instead of sharing the sources between them?

This automotive setup that shares all the sources is not something that I like, but it's too big to change in short term.

PR #5667, opened 2025-08-20
Replace Named annotation with custom qualifier for the key chain/store

TimoPtr:
## Summary

In order to avoid using string that could lead to typo I replace two Named annotation with custom qualifiers for the `KeyChainRepository` implementations.

## Checklist

## Any other notes

IMO we should forbid the usage of the `Named` annotation and privilege custom qualifier. If we agree we should migrate the remaining `Named` annotation and make this a best practice in the dev doc.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -21,6 +21,7 @@ import coil3.SingletonImageLoader
 import coil3.network.okhttp.OkHttpNetworkFetcherFactory
 import dagger.hilt.android.HiltAndroidApp
 import io.homeassistant.companion.android.common.data.keychain.KeyChainRepository
+import io.homeassistant.companion.android.common.data.keychain.NamedKeyChainRepository
 import io.homeassistant.companion.android.common.data.prefs.PrefsRepository
 import io.homeassistant.companion.android.common.sensors.AudioSensorManager
 import io.homeassistant.companion.android.common.sensors.LastUpdateManager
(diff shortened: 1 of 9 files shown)


jpelgrom:
This change's motivation feels a bit hypothetical, as realistically you're not touching these regularly, but I agree it may be a good best practice for the others (which are preferences).

---

In my opinion it makes more sense to put all qualifiers for it in one file rather than two.

Your description for `@NamedKeyChainRepository` in the file already drops "repository" - for consistency I would also drop it from the qualifier itself.

TimoPtr:
I'm going to add this as a best practice and create an issue on the repo to migrate the remaining `@Named` annotations.

TimoPtr:
The dev docs update https://github.com/home-assistant/developers.home-assistant/pull/2764

TimoPtr:
And the new issue https://github.com/home-assistant/android/issues/5684

jpelgrom:
> In my opinion it makes more sense to put all qualifiers for it in one file rather than two.

Did you disagree with this one or missed it? Thinking about LocalStorage/SharedPreferences where it is cleaner to have one file with all annotations for it.

TimoPtr:
> > In my opinion it makes more sense to put all qualifiers for it in one file rather than two.
> 
> Did you disagree with this one or missed it? Thinking about LocalStorage/SharedPreferences where it is cleaner to have one file with all annotations for it.

Missed it sorry

TimoPtr:
> > In my opinion it makes more sense to put all qualifiers for it in one file rather than two.
> 
> Did you disagree with this one or missed it? Thinking about LocalStorage/SharedPreferences where it is cleaner to have one file with all annotations for it.

fixed

jpelgrom (approved):

PR #5669, opened 2025-08-20
Setup onboarding module

TimoPtr:
## Summary

This PR is the foundation of the new onboarding, it doesn't do anything except adding all the needed files. The plan is to make small PRs that we would merge on `main`until we reach a state where we fill confident about replacing the onboarding with the new one.

## Checklist

## Any other notes

I don't know yet if we will keep a separate module or not for the onboarding, but for the time of the implementation it will stay in this dedicated module to not leak into the production app. It means that we might need to duplicate to some stuff from `:app`.

It already has some dependencies that are going to be use in the implementation of the onboarding, to avoid having to push the lockfile on every PR that are going to follow this one.

The code is going to contains `TODO` but they should be removed or tracked by an issue when we want to release the onboarding in production.


TimoPtr:
I struggled to get the instrumentation test working, the setup is kinda weird I was not expecting to have to do this to have it working properly...

onboarding/build.gradle.kts
```diff
@@ -0,0 +1,48 @@
...
+android {
+    namespace = "io.homeassistant.companion.android.onboarding"
+
+    defaultConfig {
+        testInstrumentationRunner = "io.homeassistant.companion.android.CustomTestRunner"
+    }
+    testOptions {
+        // This is needed otherwise we are not able to run instrumentation test in :onboarding
+        targetSdk = libs.versions.androidSdk.compile.get().toInt()
+    }
```
TimoPtr, line 17: Without this the test hangs indefinitely with a screen asking for permissions.

onboarding/README.md
jpelgrom, line 1: Do we need a file here?

We do not for other modules (there is the developers website) and in the IDE it defaults to a view that doesn't use the filesystem.

onboarding/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,23 @@
...
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
+
+<!-- TODO support leanback launcher -->
+<!--            <intent-filter>-->
+<!--                <action android:name="android.intent.action.MAIN" />-->
+<!--                <category android:name="android.intent.category.LEANBACK_LAUNCHER" />-->
+<!--            </intent-filter>-->
```
jpelgrom, line 19: Avoid commented out code. Leanback/tv support is way out of scope for onboarding.

onboarding/src/main/kotlin/io/homeassistant/companion/android/Launcher.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
+import dagger.hilt.android.AndroidEntryPoint
+import io.homeassistant.companion.android.common.compose.theme.HATheme
+
+/**
+ * Main entry point of the application, it manages the splash screen.
+ */
+@AndroidEntryPoint
+class Launcher : AppCompatActivity() {
```
jpelgrom: Is this a simple placeholder or intended to stay? If it stays:

 - Will this move to a subpackage (`io.homeassistant.companion.android.launcher`?) and if so can you use that package for now? Dumping files in the root will quickly become a mess.
 - An activity without `Activity` suffix feels weird.
 - Why set a theme if there is no content?

onboarding/README.md
TimoPtr: Let's keep it for now, since it's not yet documented anywhere and I don't know if we are going to keep the module. If someone end up here I want him to know that it's WIP.

onboarding/src/main/AndroidManifest.xml
TimoPtr: It's Launcher here not onboarding. In the follow up PR's the Launcher becomes the holder of the whole start of the app.

onboarding/src/main/kotlin/io/homeassistant/companion/android/Launcher.kt
TimoPtr: It is intended. I'll move it in `launcher` and rename it. For the theme it's changing in the next PR that is on top of this one 👍🏻

TimoPtr: For now the module onboarding contains class that should not be there. I'll adjust at the very end when I'll move the LauncherActivity to the app module.

onboarding/src/main/AndroidManifest.xml
TimoPtr: I'm going to keep this as a TODO like I said before releasing I need to go over all the TODO of the module.

jpelgrom:
Don't ask for a review if you're going to merge it 1 minute later, even skipping CI checks

TimoPtr:
> Don't ask for a review if you're going to merge it 1 minute later, even skipping CI checks

Sorry 🙏🏻 I merged it because I'm rebasing the next PR. And ask for another view in case you wanted me to change something which I would do in the next PR. By trying to make things reviewable and not waiting to move on it makes the management of the PR and branch quite complex. 

I ignored the UI test that is a pain today for some reason.

PR #5670, opened 2025-08-20
Run instrumentation test on API 36 instead of 35

TimoPtr:
## Summary

Now that we are targeting API 36 we should have instrumentation test running using this API. In order to save some resources I replaced the API 35 by 36.


dshokouhi:
hmm the job is stuck on API 35 lol?

jpelgrom:
> hmm the job is stuck on API 35 lol?

It's marked as required which is why it's stuck. Needs someone with more permissions on the repo to change that :)

TimoPtr:
Yep it was expected I wanted to see if it goes through first. I'll change the checks requirements and push the PR.

PR #5687, opened 2025-08-25
Introduced HATextField

TimoPtr:
## Summary

The `TextField` is not yet fully done since we are waiting for some confirmation from the design team (CC @marcinbauer85), but I wanted to have a first working solution to continue to move on the onboarding implementation.

## Checklist

## Screenshots

See within the PR

### DIFF
--- a/app/src/debug/kotlin/io/homeassistant/companion/android/developer/catalog/HAComposeCatalogScreen.kt
+++ b/app/src/debug/kotlin/io/homeassistant/companion/android/developer/catalog/HAComposeCatalogScreen.kt
@@ -16,7 +16,7 @@ import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Add
 import androidx.compose.material.icons.filled.Close
 import androidx.compose.material.icons.filled.MoreVert
-import androidx.compose.material.icons.outlined.Rectangle
+import androidx.compose.material.icons.outlined.Info
 import androidx.compose.material3.DropdownMenu
 import androidx.compose.material3.DropdownMenuItem
 import androidx.compose.material3.ExperimentalMaterial3Api
@@ -326,7 +326,7 @@ private fun LazyListScope.input() {
                 onValueChange = { value5 = it },
                 leadingIcon = {
                     Icon(
-                        imageVector = Icons.Outlined.Rectangle,
+                        imageVector = Icons.Outlined.Info,
                         contentDescription = null,
                     )
                 },


jpelgrom:
On the catalog screen: the first input is missing padding on the bottom / the second input with placeholder is missing padding on the top. All other fields have some padding between them.

What do you want us to review? Right now it is not much more than a wrapper around OutlinedTextField and we cannot validate the design specs.

TimoPtr:
> On the catalog screen: the first input is missing padding on the bottom / the second input with placeholder is missing padding on the top. All other fields have some padding between them.
> 
> What do you want us to review? Right now it is not much more than a wrapper around OutlinedTextField and we cannot validate the design specs.

The padding added for the other inputs are for the label to move up and for the error text to go bellow.

Well the docs and if it makes sense for you to have it this way. Also give your opinion on the design 👍🏻

app/src/debug/kotlin/io/homeassistant/companion/android/developer/catalog/HAComposeCatalogScreen.kt
```diff
@@ -255,6 +260,102 @@ private fun LazyListScope.textStyles() {
...
+                        style = HATextStyle.UserInput.copy(color = Color.Unspecified),
+                    )
+                },
+            )
+            HATextField(
+                value = value5,
+                onValueChange = { value5 = it },
+                leadingIcon = {
+                    Icon(
+                        imageVector = Icons.Outlined.Rectangle,
```
jpelgrom: Maybe use info like the screenshot text? I tapped it more than once while using the catalog thinking it was a checkbox 😅

app/src/debug/kotlin/io/homeassistant/companion/android/developer/catalog/HAComposeCatalogScreen.kt
```diff
@@ -255,6 +260,102 @@ private fun LazyListScope.textStyles() {
...
+                    }
+                },
+            )
+            HATextField(
+                value = value2,
+                onValueChange = { value2 = it },
+                placeholder = {
+                    Text(
+                        text = "Placeholder",
+                        style = HATextStyle.UserInput.copy(color = Color.Unspecified),
```
jpelgrom, line 294: Assuming this is for the preview only, but otherwise: making this look like normal text made me want to select it to clear it.

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HATextField.kt
```diff
@@ -0,0 +1,75 @@
...
+        // The color is controlled from the [colors] attribute
+        textStyle = HATextStyle.UserInput.copy(color = Color.Unspecified),
+        keyboardOptions = keyboardOptions,
+        keyboardActions = keyboardActions,
+        placeholder = placeholder,
+        isError = isError,
+        label = label,
+        colors = LocalHAColorScheme.current.textField(),
+        modifier = modifier
+            .widthIn(max = MaxButtonWidth)
```
jpelgrom, line 72: No way to override max width?

jpelgrom (commented):
Cannot verify the design.

The composable setup generally looks OK, which is expected as you're building a wrapper around Google's work with defaults.

> The padding added for the other inputs are for the label to move up and for the error text to go bellow.

Makes sense, but it does look 'wrong', so should probably be avoided in production use.

app/src/debug/kotlin/io/homeassistant/companion/android/developer/catalog/HAComposeCatalogScreen.kt
TimoPtr: This part is where I need UI helps from the design team.

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HATextField.kt
TimoPtr: Well for now I would like to keep it like the button and follow material guidance.

TimoPtr:
> Makes sense, but it does look 'wrong', so should probably be avoided in production use.

I think it's fine like this for now since it is on the catalog and not meant to be seen by anyone else than developer or UI/UX team.

Issue #5665, opened 2025-08-20
100% freeze rate on entering settings


RubenKelevra:
**Home Assistant Android app version(s):**

**Android version(s):**

12

**Device model(s):**

Redmi Note 10 Pro

**Home Assistant version:**

2025.8.1

**Last working Home Assistant release (if known):**

Not related

**Description of problem, include YAML if issue is related to notifications:**

App freezes when entering settings for app. Tapping anywhere and waiting around 20 seconds will show androids "app does not respond" window to close it. 

**Companion App Logs:**

Not possible. 

**Screenshot or video of problem:**

https://github.com/user-attachments/assets/98127a23-d1d8-40c4-82a9-a1c72aabbfc2

**Additional information:**

jpelgrom:
Did this previously work for you, or is this issue recent?

RubenKelevra:
This did previously work fine for me.

I had issues using the app on my smartwatch: No actions worked.

So I thought: Let's remove and reinstall the apps.

Now the settings don't work anymore in the main app and I also can't pair my smartwatch anymore: Ai select the instance on my watch, and continue then on my phone, but then the finishing step doesn't work.

So yeah, pretty bummed, as this blocks access from my smartwatch and remote access to my HA instance, as I can't add the definition for "external" where it should try to use the VPN IP, as well as I'm unable to provide the wifi which is considered "home".

TimoPtr:
Which version of the app are you on? Beta or 2025.7.1?

RubenKelevra:
@TimoPtr I'm on the beta channel in the Playstore, which showed me the version I wrote in the ticket.

RubenKelevra:
Weirdly I *was* able to enter the settings yesterday, and add the wifi name/internal and external (VPN) IPs, but now it hangs again.

AWIDIMAIG:
I’ve had the same problem since the 2025.8.7 update. I tried clearing the app cache through the Android settings, but that didn’t solve the issue. After that, I completely cleared all data, and now I’m stuck at the server configuration. The app detects the server, but when I click on it, a blank white screen appears and stays there indefinitely.

**Edit:** The server configuration issue does not appear to be related to the 2025.8.7 update, as it also occurs with older versions.
Regarding the freeze issue, I only encounter the problem on my Lenovo Smart Tab M10 Plus (TB-X606F). On my mobile phone (OnePlus 8 Pro), the app works as expected.

[...]

PR #5690, opened 2025-08-25
Fix/anr settings

TimoPtr:
## Summary

Potential fix of https://github.com/home-assistant/android/issues/5665. I was able to reproduce an ANR by making an artificial delay within the bounds of the `runBlocking` of the `isActive` method. By making sure that it doesn't block the UI thread the ANR is gone with this artificial delay. 

It might be only one part of the issue so we need people impacted by this issue to test the fix to see if it is enough. On a longer term we need to get rid of almost of `runBlocking` since we do have other ANRs.

I suspect the issue to only appear on old devices with a small amount of available threads in the IO thread pool (since it's based on the number of thread available on the CPU).

## Checklist

## Any other notes

Credits to @loganrosen from https://github.com/home-assistant/android/pull/5651 for the fix on the migration.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsActivity.kt
@@ -10,6 +10,7 @@ import android.view.ViewGroup
 import androidx.biometric.BiometricManager
 import androidx.core.content.IntentCompat
 import androidx.fragment.app.commit
+import androidx.lifecycle.lifecycleScope
 import dagger.hilt.EntryPoint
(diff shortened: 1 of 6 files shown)


jpelgrom:
Did you test this with app lock enabled? That is a feature which should not be delayed in case you want the lock to work all the time.

TimoPtr:
> Did you test this with app lock enabled? That is a feature which should not be delayed in case you want the lock to work all the time.

It seems to behave as before on my pixel when I compare to the app without this change. I tested on a very old device that I just found (One plus 3T with Android 9) and I cannot see the content behind even very fast.

Issue #5689, opened 2025-08-25
App takes a long time to open after update to 2025.8.7


dougle03:
UPDATE:
THIS BUG IS RELATED TO HAVING MORE THAN ONE SERVER IN THE APP. IF YOU ONLY HAVE ONE SERVER YOU MAY NOT HAVE A PROBLEM.

**Home Assistant Android app version(s):**
2025.8.7

**Android version(s):**
16
**Device model(s):**
Google Pixel 6 Pro
**Home Assistant version:**
2025.8.2
**Last working Home Assistant release (if known):**

**Description of problem, include YAML if issue is related to notifications:**
On opening the app the loading spinning wheel is presented for anywhere from 10s to over 2m
This is a new behaviour since the app was updated from the play store two days ago. Prior to this the app opened almost instantly to my dashboard without any noticeable delay.
This is the same on WiFi at home where HA is installed, and away from Wifi on mobile.
It usually eventually loads, but the long delay is making it almost unusable.

**Companion App Logs:**

```

```

**Screenshot or video of problem:**

**Additional information:**

SirSimon97:
I also see this behaviour on my Galaxy S24 after updating to 2025.8.7 for any restart of the app. I also observed, that when I switch Wifi or Mobile Internet off and on again during this loading screen, the app is loaded immediately.

Nytromix:
Same issues here on my s24. It started right after the last update. Using the method that @SirSimon97 mentioned seems to help. Thanks SS97. Hopefully there will be a quick fix because I depend on the app more than hopping on my computer.

TimoPtr:
I've made a first attempt to fix this issue if someone wants to try it https://github.com/home-assistant/android/pull/5693

dougle03:
> I've made a first attempt to fix this issue if someone wants to try it [#5693](https://github.com/home-assistant/android/pull/5693)

Awsome. Can't try it, don't have the skills...
Any idea when this will be released through the play store?
Thanks.

TimoPtr:
> > I've made a first attempt to fix this issue if someone wants to try it [#5693](https://github.com/home-assistant/android/pull/5693)
> 
> Awsome. Can't try it, don't have the skills...
> Any idea when this will be released through the play store?
> Thanks.

No timeline yet for a release on the play store but an APK can be found in the PR. It will be on the next beta if you want. But like I said I'm not sure it fixes completely the issue.

Nytromix:
[...]

PR #5693, opened 2025-08-25
Stop observing network state after being READY

TimoPtr:
## Summary

We potentially had a race condition with `collectLatest` and how we observe the network status. By simply making a synthetic implementation of my assumption

```kotlin
override fun observeNetworkStatus(serverConfig: ServerConnectionInfo): Flow<NetworkState> {
        return flow {
            emit(NetworkState.READY_LOCAL)
            emit(NetworkState.CONNECTING)
        }
}
```

I was stuck on the loading animation, I suspect that on some phone this situation might happen, but I have no idea how and why.

This PR is a proposal of how to avoid this situation by making sure that if that the flow is emitting `READY` at any point we proceed. I came to this conclusion based on the workaround exposed in https://github.com/home-assistant/android/issues/5689#issuecomment-3220109955 that says that to unlock this situation the user simply toggle WIFI on/off that actually will end up into observeNetworkStatus and unblock the situation.

## Checklist

## Any other notes

@fpetrovski would you mind testing it with your initial use case since I've replace the `collectLatest` with a simple `collect` and a `takeWhile`.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
@@ -37,7 +37,7 @@ abstract class LaunchPresenterBase(
             ) {
                 networkStatusMonitor.observeNetworkStatus(activeServer.connection)
                     .takeWhile { state ->
-                        // Until the network is ready we continue to observer network status changes
+                        // Until the network is ready we continue to observe network status changes
                         !handleNetworkState(state)
                     }.collect()
             } else {


fpetrovski (approved):
This seems the right approach and it should fix the issue.

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
```diff
@@ -35,13 +36,15 @@ abstract class LaunchPresenterBase(
                 activeServer != null
             ) {
                 networkStatusMonitor.observeNetworkStatus(activeServer.connection)
-                    .collectLatest { state ->
-                        if (handleNetworkState(state)) return@collectLatest
-                    }
+                    .takeWhile { state ->
+                        // Until the network is ready we continue to observer network status changes
```
jpelgrom:
```suggestion
                        // Until the network is ready we continue to observe network status changes
```

jpelgrom (approved):

PR #5633, opened 2025-08-09
Open tile settings from newly created tile

Martreides:
## Summary

This PR is intended to add buttons to newly created tiles that send the user directly to the settings page in the app where the user can configure the tile.

This enhancement is intended to resolve https://github.com/home-assistant/android/issues/4918.

## Checklist

- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [ ] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [ ] The changes have been thoroughly tested, and edge cases have been considered.
- [ ] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

## Link to pull request in documentation repositories

User Documentation: home-assistant/companion.home-assistant#

Developer Documentation: home-assistant/developers.home-assistant#

## Any other notes

This PR is under development.

PR #5626, opened 2025-08-05
Use sealed class instead of an optional param for SettingsActivity links

TimoPtr:
## Summary

In order to ease the navigation within the settings I replace the optional parameter of `SettingsActivity.newInstance` with a sealed class. It is going to make easier the work in `Add to` shortcut and tile.

## Checklist

- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

PR #5708, opened 2025-08-28
Add open settings button to camera tile

Martreides:
## Summary
This PR adds an "Open settings" button to the camera tile if it is not configured. This functionality is similar as was included in https://github.com/home-assistant/android/pull/5633 for the thermostat tile.

## Checklist

### DIFF
--- a/wear/src/main/kotlin/io/homeassistant/companion/android/home/HomeActivity.kt
+++ b/wear/src/main/kotlin/io/homeassistant/companion/android/home/HomeActivity.kt
@@ -54,8 +54,7 @@ class HomeActivity :
                 putExtra(EXTRA_FROM_ONBOARDING, fromOnboarding)
             }
         }
-
-        fun getLaunchAction(packageName: String, tileId: Int): ActionBuilders.LaunchAction {
+        fun getLaunchAction(packageName: String, tileId: Int, launchMode: String): ActionBuilders.LaunchAction {
             val androidActivity = ActionBuilders.AndroidActivity.Builder()
                 .setPackageName(packageName)
                 .setClassName(
@@ -64,7 +63,7 @@ class HomeActivity :
                 .addKeyToExtraMapping(
                     LAUNCH_MODE,
                     ActionBuilders.AndroidStringExtra.Builder().setValue(
-                        OpenTileSettingsActivity.CONFIG_THERMOSTAT_TILE,
+                        launchMode,
                     )
                         .build(),
                 )
@@ -114,12 +113,14 @@ class HomeActivity :
         super.onCreate(savedInstanceState)
         // Get rid of me!
         presenter.init(this)
-
-        if (intent.getStringExtra(LAUNCH_MODE) == OpenTileSettingsActivity.CONFIG_THERMOSTAT_TILE) {
+        val launchMode = intent.getStringExtra(LAUNCH_MODE)
+        if (launchMode == OpenTileSettingsActivity.CONFIG_THERMOSTAT_TILE ||
+            launchMode == OpenTileSettingsActivity.CONFIG_CAMERA_TILE
+        ) {
             startActivity(
                 OpenTileSettingsActivity.newInstance(
                     this@HomeActivity,
-                    OpenTileSettingsActivity.CONFIG_THERMOSTAT_TILE,
+                    launchMode,
                     intent.getIntExtra(OpenTileSettingsActivity.TILE_ID_KEY, 0),
                 ),
             )
--- a/wear/src/main/kotlin/io/homeassistant/companion/android/tiles/CameraTile.kt
+++ b/wear/src/main/kotlin/io/homeassistant/companion/android/tiles/CameraTile.kt
@@ -83,7 +83,12 @@ class CameraTile : TileService() {
                 .setTileTimeline(
                     if (serverManager.isRegistered()) {
                         if (tileConfig?.entityId.isNullOrBlank()) {
-                            getNotConfiguredTimeline(this@CameraTile, requestParams)
+                            getNotConfiguredTimeline(
+                                this@CameraTile,
+                                requestParams,
+                                commonR.string.camera_tile_no_entity_yet,
+                                OpenTileSettingsActivity.CONFIG_CAMERA_TILE,
+                                )
                         } else {
                             timeline(
                                 requestParams.deviceConfiguration.screenWidthDp,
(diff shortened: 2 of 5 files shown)


Martreides:
Both for the thermostat and the camera tile I have now created function getNotConfiguredTimeline. I expect it is preferable to only define it once somewhere centrally. Any ideas on where I can put it? Should we create something to put generic tile functions?

wear/src/main/kotlin/io/homeassistant/companion/android/home/HomeActivity.kt
```diff
@@ -54,8 +54,7 @@ class HomeActivity :
                 putExtra(EXTRA_FROM_ONBOARDING, fromOnboarding)
             }
         }
-
-        fun getLaunchAction(packageName: String, tileId: Int): ActionBuilders.LaunchAction {
+        fun getLaunchAction(packageName: String, tileId: Int, launchMode: String): ActionBuilders.LaunchAction {
```
TimoPtr: Let's avoid using string for launchMode and create a sealed class it will allow you to use the when operator properly. 

I just realized that we didn't put this into our documentation, so I've wrote this https://github.com/home-assistant/developers.home-assistant/pull/2768 let me know if you find it useful.

Also you have a concrete example on this PR https://github.com/home-assistant/android/pull/5626

wear/src/main/kotlin/io/homeassistant/companion/android/tiles/CameraTile.kt
```diff
@@ -126,7 +135,7 @@ class CameraTile : TileService() {
                         withContext(Dispatchers.IO) {
                             val response = okHttpClient.newCall(Request.Builder().url(url).build()).execute()
                             byteArray = response.body.byteStream().readBytes()
-                            byteArray?.let {
+                            byteArray.let {
```
TimoPtr: It's nice to remove the warning but here the let is useless and should be removed.

TimoPtr:
> Both for the thermostat and the camera tile I have now created function getNotConfiguredTimeline. I expect it is preferable to only define it once somewhere centrally. Any ideas on where I can put it? Should we create something to put generic tile functions?

Indeed you can probably put it under `TileViews`

Martreides:
@timoptr, I don't have time to work on this PR this and next week. Will pick it up thereafter, though.

wear/src/main/kotlin/io/homeassistant/companion/android/home/HomeActivity.kt
Martreides: Could you check whether this makes sense: https://github.com/home-assistant/android/pull/5708/commits/9a8fa62b84867c7813aff09ff340005aaed25a3b?

wear/src/main/kotlin/io/homeassistant/companion/android/tiles/CameraTile.kt
Martreides: Could you check this as well https://github.com/home-assistant/android/pull/5708/commits/e836744b14e66d25fae24217211a335456e72d3a?

wear/src/main/kotlin/io/homeassistant/companion/android/home/HomeActivity.kt
```diff
@@ -55,7 +55,12 @@ class HomeActivity :
             }
         }
 
-        fun getLaunchAction(packageName: String, tileId: Int): ActionBuilders.LaunchAction {
+        sealed class LaunchMode(val mode: String) {
+            object ThermostatTile : LaunchMode(OpenTileSettingsActivity.CONFIG_THERMOSTAT_TILE)
+            object CameraTile : LaunchMode(OpenTileSettingsActivity.CONFIG_CAMERA_TILE)
+        }
```
TimoPtr: You don't need a `mode`

```suggestion
        sealed interface LaunchMode {
            object ThermostatTile : LaunchMode
            object CameraTile : LaunchMode
        }
```

When you do this then you can do something like 
```kotlin
when(launchMode) {
ThermostatTile -> // DO watever you want for thermostat
CameraTile -> // Same for Camera
```

With this pattern if you even add a new object the compiler is going to warn you at the `when` level that you need to handle the new case.

In your case maybe you want to create the sealed interface at the OpenTileSettingsActivity since the two options are for this activity not really for HomeActivity.

Martreides: Changed it: [a7aabbe](https://github.com/home-assistant/android/pull/5708/commits/a7aabbeae7121d020e3283c109b5797ff156ee39)

wear/src/main/kotlin/io/homeassistant/companion/android/tiles/CameraTile.kt
```diff
@@ -127,33 +136,31 @@
...
+                                imageWidth = maxWidth.toInt()
+                                imageHeight = maxHeight.toInt()
+                                if (currentRatio > screenRatio) {
+                                    imageWidth = (maxHeight * currentRatio).toInt()
                                 } else {
-                                    imageWidth = bitmap.width
-                                    imageHeight = bitmap.height
+                                    imageHeight = (maxWidth / currentRatio).toInt()
+                                }
+                                bitmap = Bitmap.createScaledBitmap(bitmap, imageWidth, imageHeight, true)
```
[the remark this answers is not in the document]
TimoPtr: I think it's because you've rebase your branch. This has been fixed since you've started your branch.

wear/src/main/kotlin/io/homeassistant/companion/android/home/HomeActivity.kt
TimoPtr: Thanks. I looked a bit more and I've assumed wrong that it was possible to give an object to the `ActionBuilders` but it is not. So your proposal is good for me 👍🏻

wear/src/main/kotlin/io/homeassistant/companion/android/tiles/CameraTile.kt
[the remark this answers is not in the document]
Martreides: Should be resolved now. There was indeed a small merge conflict yesterday and I guess I made a mistake in resolving that.

Martreides:
I think the groundwork has been done in this PR and the one for the thermostat tile, so I expect the remaining tiles should be a bit quicker :)

TimoPtr:
> I think the groundwork has been done in this PR and the one for the thermostat tile, so I expect the remaining tiles should be a bit quicker :)

Indeed :)

PR #5711, opened 2025-08-31
Fixup misleading nomenclature: s/frontend cache/client cache/g

nethershaw:
## Summary

* What was the motivation behind this change?

This commit replaces every instance of a misleading reference to a "frontend cache" where there is no such thing with a correct, however generalized, reference to the "client cache."

In objective fact, a "frontend cache" is a term reserved for server-side caching reverse proxies centralized on and attributable to a frontend itself. No such resource exists in any Home Assistant deployment. This incorrect use of terms is confusing and misleading, and therefore is herein replaced. I will confess to having been thusly misled despite (and perhaps _because of_) being employed in the relevant technology industries for going on two decades.

The only relevant cache is attributable to the client, whether it is a browser or whether it is a dedicated platform client. This change categorically prevents the mistaken apprehension that any server-side cache attributable to the frontend requires resetting, and at the very least, is *less incorrect.* If the word "client" is objectionable, valid alternatives e.g. "browser" and "user," or simply eschewing the misleading qualifier entirely (not recommended), remain available.

* What is the impact of the changes on the application?

Functionally, none. For this repository, the change represents barely a single string.

## Checklist

- [ x ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
  (strictly irrelevant to this PR)
- [ x ] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
  (no code has been changed; only a localization string)
- [ x ] The changes have been thoroughly tested, and edge cases have been considered.
  (there is no possibility of relevance here)
- [ x ] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.
  (no string matching the case-insensitive expression "frontend cache" appears in the documentation (indeed, this is the reason for the other corrections); therefore, no correction is required here)

## Any other notes

cf. https://community.home-assistant.io/t/how-to-clear-the-frontend-cache/670491/91

This pull request is part of a set against the following repositories (crosslinks may be updated after the PR is opened):
  - home-assistant/android
  - home-assistant/iOS
  - hacs/integration

### DIFF
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -146,7 +146,7 @@
     <string name="choose_server">Choose server</string>
     <string name="clear_favorites">Clear favorites</string>
     <string name="clear_search">Clear search</string>
-    <string name="clear_webview_cache">Reset frontend cache</string>
+    <string name="clear_webview_cache">Reset client cache</string>
     <string name="clear_webview_cache_active">Resetting…</string>
     <string name="clear_webview_cache_failed">Reset failed</string>
     <string name="clear_webview_cache_success">Reset completed</string>


TimoPtr:
Well we could argue that it is the frontend cache on the client. @jpelgrom @bgoncal @dshokouhi let's discuss here about this change.

jpelgrom:
> Well we could argue that it is the frontend cache on the client.

Yes - specifically for apps it is IMO _more_ wrong than before because there are other kinds of cache which are not included here. And even on a computer "client cache" doesn't mean anything to me when using a website, "browser cache" seems more common.

Specifically for HACS, why wouldn't a quick "How do I do this?" question in that dialog help? The terminology is consistent at least. I can imagine a similar forum thread existing if "client cache" was used 🤷

bgoncal:
In an App context saying "browser cache" could be confusing because for the user it's an App and not a browser.

On the other hand "frontend" is not something that is clear for all users either, I don't have a better  proposal, all caching on the webview level is generated by the frontend (web UI) right? So... it's not wrong in my opinion.

MindFreeze:
The frontend repo is called "frontend" so that is already correct. If you want to be more specific, you can just call it webview cache.

TimoPtr:
I like the french translation. It says re-initialize cache user interface and it's under the troubleshooting section so I expect that you end up here if you actually have an issue.

Frontend doesn't mean anything for a non developer in my opinion, but cache and client too.

TimoPtr:
No activity on the PR I'm closing until we agree on something.

Issue #5688, opened 2025-08-25
Remove any usage of `runBlocking`


TimoPtr:
We do have a lot of `runBlocking` usage in the app. Most of them could be replaced by a proper coroutine. It leads to some ANR because it actually blocks the Main thread which we must not do under any circumstances. 

It is not an easy task since some of them are quite hard to replace properly.

PR #5722, opened 2025-09-01
Remove runBlocking from ServerManager.activeServerId

TimoPtr:
## Summary

Part of the effort describe in https://github.com/home-assistant/android/issues/5688 we should remove all the `runBlocking`. There was one in `ServerManager` to get the active server.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/util/vehicle/GridItems.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/util/vehicle/GridItems.kt
@@ -77,7 +77,7 @@ fun getChangeServerGridItem(
 fun getNavigationGridItem(
     carContext: CarContext,
     screenManager: ScreenManager,
-    integrationRepository: IntegrationRepository,
+    integrationRepositoryProvider: suspend () -> IntegrationRepository,
     allEntities: Flow<Map<String, Entity>>,
     entityRegistry: List<EntityRegistryResponse>?,
 ): GridItem.Builder {
@@ -100,7 +100,7 @@ fun getNavigationGridItem(
             screenManager.push(
                 MapVehicleScreen(
                     carContext,
-                    integrationRepository,
+                    integrationRepositoryProvider,
                     allEntities.map {
                         it.values.filter { entity ->
                             entity.domain in MAP_DOMAINS &&
@@ -158,6 +158,7 @@ fun getDomainList(
                 domainIsEmpty = it.isEmpty()
             }
         }
+        // TODO most probably this is always false since the launch will update the boolean after we actually test
         if (!domainIsEmpty) {
             listBuilder.addItem(
                 GridItem.Builder().apply {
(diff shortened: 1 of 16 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/util/vehicle/GridItems.kt
```diff
@@ -158,6 +158,7 @@ fun getDomainList(
                 domainIsEmpty = it.isEmpty()
             }
         }
+        // TODO most probably this is always false since the launch will update the boolean after we actually test
```
TimoPtr, line 161: @dshokouhi I didn't test but just by reading the code I can say that most probably `domainIsEmpty` is always false is it expected?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManagerImpl.kt
```diff
@@ -203,30 +203,33 @@ class ServerManagerImpl @Inject constructor(
...
-    override fun authenticationRepository(serverId: Int): AuthenticationRepository {
+    override suspend fun authenticationRepository(serverId: Int): AuthenticationRepository {
         val id = if (serverId == SERVER_ID_ACTIVE) activeServerId() else serverId
         return authenticationRepos[id] ?: run {
             if (id == null || mutableServers[id] == null) throw IllegalArgumentException("No server for ID")
-            authenticationRepos[id] = authenticationRepositoryFactory.create(id)
-            authenticationRepos[id]!!
+            val repository = authenticationRepositoryFactory.create(id)
+            authenticationRepos[id] = repository
+            checkNotNull(authenticationRepos[id]) { "Should not be null since we've just called create ($repository)" }
```
TimoPtr, line 212: Added more logs here because I still see some crash on this repo being null and I cannot say how it could be null...

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
```diff
@@ -133,33 +134,36 @@ class IntegrationRepositoryImpl @AssistedInject constructor(
...
-            webSocketRepository.getCurrentUser() // To get user info stored
+            webSocketRepository().getCurrentUser() // To get user info stored
         } catch (e: Exception) {
             Timber.e(e, "Unable to save device registration")
         }
     }
 
     override suspend fun updateRegistration(deviceRegistration: DeviceRegistration, allowReregistration: Boolean) {
         val request = RegisterDeviceIntegrationRequest(createUpdateRegistrationRequest(deviceRegistration))
-        server().callWebhookOnUrls(request, onSuccess = { response ->
```
jpelgrom, line 144: Does ktlint require this? Otherwise I'd prefer to keep it as it is to avoid more changed lines in this PR, even if at a glance it looks the same.

TimoPtr: It is updated automatically with the ktlint plugin that is now enabled by default.

PR #5723, opened 2025-09-01
Fix fire and forget resync to avoid blocking UI at startup

### LINKED ISSUE #5689: App takes a long time to open after update to 2025.8.7

TimoPtr:
## Summary

In https://github.com/home-assistant/android/pull/5581 we've introduce a regression by making the `resyncRegistration` part of the flow. The regression is only visible when one server is not responding or any call during the registration. It was causing the start to be postpone until the server timeout (could be up to 30s).

 This PR does move the registration back into a dedicated coroutine scope. This fix the slow delay issue but it does mean that the activity could out live it's lifecycle, which is bad but it is limited to maximum the timeout of the server so 30s.

## Checklist

## Any other notes

Fixes #5689 

We could consider adding a new state to a server to mark it as disabled and avoid calling a server if we know that it might not respond. It might be useful for ppl that are managing servers for other and where they are not able to reach out to this server but they still want the server in the app for when they are able to reach out.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
@@ -30,8 +30,7 @@ abstract class LaunchPresenterBase(
      *
      * **Note:** Using a dedicated scope like this means that the underlying Activity might not be
      * destroyed immediately if these background tasks are still running. This is a trade-off
-     * for improved UI responsiveness in scenarios with potential network delays. If we
-     * introduce a new server state like disable we could maybe bring this back in the flow.
+     * for improved UI responsiveness in scenarios with potential network delays.
      *
      * We use a [SupervisorJob] on purpose because it ensures that if one task fails, it won't
      * impact the others.


TimoPtr:
We could consider moving this to a Worker in a second step.

jpelgrom:
Couldn't the variable live in `LaunchPresenterBase` instead? Now you're duplicating a lot of lines.

Adjusted title because the intention was always that this action was a best effort sync but not blocking/"fire and forget".

> We could consider adding a new state to a server to mark it as disabled and avoid calling a server if we know that it might not respond. It might be useful for ppl that are managing servers for other and where they are not able to reach out to this server but they still want the server in the app for when they are able to reach out.
>
> We could consider moving this to a Worker in a second step.

I'm not in favour of adding more configuration like that to the app unless absolutely necessary. Just a silent failure which isn't blocking should be enough. A worker could be nice (there is even a comment that has existed for a long time asking if this is the right place, but it was never a priority or issue), but out of scope for this PR.

TimoPtr:
> Couldn't the variable live in `LaunchPresenterBase` instead? Now you're duplicating a lot of lines.
> 
> Adjusted title because the intention was always that this action was a best effort sync but not blocking/"fire and forget".
> 
> > We could consider adding a new state to a server to mark it as disabled and avoid calling a server if we know that it might not respond. It might be useful for ppl that are managing servers for other and where they are not able to reach out to this server but they still want the server in the app for when they are able to reach out.
> > We could consider moving this to a Worker in a second step.
> 
> I'm not in favour of adding more configuration like that to the app unless absolutely necessary. Just a silent failure which isn't blocking should be enough. A worker could be nice (there is even a comment that has existed for a long time asking if this is the right place, but it was never a priority or issue), but out of scope for this PR.

Indeed it can I'm going to adjsut. Then I'm going to open an issue with the details 👍🏻

TimoPtr:
https://github.com/home-assistant/android/issues/5724

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchPresenterBase.kt
```diff
@@ -17,6 +20,26 @@ abstract class LaunchPresenterBase(
...
+     * This is crucial for operations like server resynchronization, especially when dealing with
+     * multiple servers where one might be unresponsive.
+     *
+     * By launching tasks in this scope, we ensure that the UI remains responsive and is displayed
+     * promptly, even if a server connection times out.
+     *
+     * **Note:** Using a dedicated scope like this means that the underlying Activity might not be
+     * destroyed immediately if these background tasks are still running. This is a trade-off
+     * for improved UI responsiveness in scenarios with potential network delays. If we
+     * introduce a new server state like disable we could maybe bring this back in the flow.
```
jpelgrom:
```suggestion
     * for improved UI responsiveness in scenarios with potential network delays.
```

Considering the other adjusted comment and how we need to keep the app launch as fast as possible, I don't think we ever want to have this blocking the flow, so remove the line?

TimoPtr: Done

PR #5725, opened 2025-09-01
Ignore SSL errors for non‑critical in Ingress WebView

tzagim:
Ignore SSL errors for non‑critical external resources (e.g. Google Tag Manager) in Ingress WebView

## Summary
When using add‑ons via Ingress in the Android app, if a third‑party resource such as https://www.googletagmanager.com/gtm.js is blocked at the network/DNS level (e.g. via Pi‑hole, AdGuard), the WebView receives an onReceivedSslError for that resource. Currently, the app treats any SSL error as a fatal connection failure and shows the "The certificate authority for Home Assistant is not trusted" dialog, even though the main Home Assistant server certificate is valid.

This behavior is misleading and interrupts the user experience for a non‑critical resource.

This PR should resolve issues like [this](https://community.home-assistant.io/t/android-app-unable-to-connect-to-home-assistant-error/866366)

## Checklist

## Any other notes
This change is intentionally scoped to a specific known domain to minimize risk. If accepted, it could be extended in the future to ignore SSL errors for any non‑critical external host that is not the active Home Assistant server.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -415,8 +415,14 @@ class WebViewActivity :
                     authenticationDialog(handler, host, resourceURL, realm, authError)
                 }
 
-                override fun onReceivedSslError(view: WebView?, handler: SslErrorHandler?, error: SslError?) {
-                    Timber.e("onReceivedSslError: $error")
+                override fun onReceivedSslError(view: WebView? ,handler: SslErrorHandler? ,error: SslError?) {
+                    val url = error?.url ?: ""
+                    Timber.e("onReceivedSslError: $url -> $error")
+                    if (url.contains("googletagmanager.com")) {
+                        Timber.w("Ignoring SSL error for non-critical resource: $url")
+                        handler?.proceed()
+                        return
+                    }
                     showError(
                         ErrorType.SSL,
                         error,


dshokouhi:
This is not a good idea. We should not merge. The blocking is user controlled based on the description.

tzagim:
Hi @dshokouhi 
Thanks for the feedback. I understand the concern about not overriding intentional user‑controlled blocking. The intent of this PR is not to bypass a deliberate block, but to address a UX issue where blocking a non‑critical third‑party resource (like GTM) via DNS results in the app showing a fatal SSL error for the entire Ingress session - even though the main Home Assistant server certificate is valid.

In this scenario, the user did not block Home Assistant itself, and the blocked resource is not required for the add‑on to function. The current behavior misleads the user into thinking there is a problem with their HA server’s SSL configuration, when in fact the issue is with an unrelated external script.

This PR scopes the ignore logic to a specific known domain to minimize risk and avoid masking real SSL issues for the HA server. If the concern is about generalizing this behavior, I’m open to:

1. Making it opt‑in via a setting, or
2. Limiting it to a small, vetted list of known non‑critical domains

The goal is to improve the user experience and reduce confusion, not to override intentional blocking decisions.

TimoPtr:
This is a quite advanced topic for a user of the application. From my point of view the actual issue is the error message. We should show a more relevant message maybe with the domain to help the user pin point the issue in a first step.

Then this is specific to third party addons that might have tracking enable and the user block this tracking on his own by his own mean. I don't think we can have a hard coded list like this in the app, with that being said I'm going to close this PR. But you can make another one at least to improve the error message.

tzagim:
Hi @TimoPtr 
Thanks for the feedback and for clarifying the preferred direction. As suggested, I’ve opened a new PR that focuses on improving the SSL error message for external resources, showing the domain name to help users identify the cause, while keeping the original behavior for the main HA host.

You can find the new PR here: #5729

PR #5729, opened 2025-09-02
Improve SSL error handling in WebViewActivity for external resources

tzagim:
## Summary
Improve SSL error handling in WebViewActivity for external resources.

When an SSL error occurs for a resource whose host is different from the active
Home Assistant server (e.g. third-party scripts like Google Tag Manager),
the app will now:

- Log a warning with the failing URL.
- Show a non-blocking Toast message with the external domain name.
- Proceed with loading the rest of the Ingress content instead of stopping.

This prevents misleading "certificate authority not trusted" dialogs when the
main HA server certificate is valid, but a non-critical external resource fails
due to DNS blocking or certificate issues.

If the SSL error is for the active Home Assistant host, the existing error
handling remains unchanged.

Closes: #5725 

This PR should resolve issues like [this](https://community.home-assistant.io/t/android-app-unable-to-connect-to-home-assistant-error/866366)

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -416,12 +416,23 @@ class WebViewActivity :
                 }
 
                 override fun onReceivedSslError(view: WebView?, handler: SslErrorHandler?, error: SslError?) {
+                    val failingUrl = error?.url ?: ""
+                    val activeHost = Uri.parse(
+                        serverManager.getServer(presenter.getActiveServer())?.connection?.getUrl(false)
+                    ).host ?: ""
+                
+                    if (Uri.parse(failingUrl).host != activeHost) {
+                        Timber.w("SSL error for external resource: $failingUrl")
+                        Toast.makeText(
+                            this,
+                            getString(R.string.webview_error_SSL_EXTERNAL_RESOURCE, Uri.parse(failingUrl).host),
+                            Toast.LENGTH_LONG
+                        ).show()
+                        handler?.proceed()
+                        return
+                    }
                     Timber.e("onReceivedSslError: $error")
-                    showError(
-                        ErrorType.SSL,
-                        error,
-                        null,
-                    )
+                    showError(ErrorType.SSL, error, null)
                 }
 
                 override fun onRenderProcessGone(view: WebView?, handler: RenderProcessGoneDetail?): Boolean {
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -1037,6 +1037,7 @@
     <string name="webview_error_SSL_INVALID">A generic SSL error occurred when loading Home Assistant, please review the Home Assistant certificate or the connection settings and try again.</string>
     <string name="webview_error_SSL_NOTYETVALID">The Home Assistant certificate is not yet valid, please review the Home Assistant certificate or the connection settings and try again.</string>
     <string name="webview_error_SSL_UNTRUSTED">The Home Assistant certificate authority is not trusted, please review the Home Assistant certificate or the connection settings and try again.</string>
+    <string name="webview_error_SSL_EXTERNAL_RESOURCE">SSL error while loading resource from %1$s.</string>
     <string name="webview_error">There was an error loading Home Assistant, please review the connection settings and try again. We will attempt to try another provided URL when you select Refresh.</string>
     <string name="welcome_hass_desc">This app connects to your Home Assistant server and allows integrating data about you and your phone.\n\nHome Assistant is free and open source home automation software with a focus on local control and privacy.</string>
     <string name="welcome_hass">Welcome to Home Assistant Companion!</string>


app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -416,12 +416,23 @@ class WebViewActivity :
...
+                    ).host ?: ""
+                
+                    if (Uri.parse(failingUrl).host != activeHost) {
+                        Timber.w("SSL error for external resource: $failingUrl")
+                        Toast.makeText(
+                            this,
+                            getString(R.string.webview_error_SSL_EXTERNAL_RESOURCE, Uri.parse(failingUrl).host),
+                            Toast.LENGTH_LONG
+                        ).show()
+                        handler?.proceed()
```
TimoPtr: To be honest I don't like this. It's very risky to let this happen. Imagine if you are not the admin of a Home Assistant instance and the admin is malicious he could to some very ugly things without any trouble. Like phishing to only quote one security issue. 

I'm ok merging the toast but not the proceed.

tzagim: I understand your security concern about handler.proceed() and the potential for abuse if the HA admin is malicious. My goal here is to prevent Ingress from breaking when a clearly non‑critical external resource (e.g. analytics script) fails SSL, while keeping the main HA host fully protected.

Would you be open to a more restricted approach where:

1. We only call proceed() if the failing host is different from the active HA host,
2. The resource is loaded over HTTPS, and
3. (Optionally) this behavior is behind an opt‑in setting in the app?

This way, we avoid blocking the user experience for harmless blocked resources, but still protect against any SSL issues on the main HA server

TimoPtr: The 3 would be the minimum and I think it would still not be secure so we should not do it. We cannot identify `non‑critical external resource`. If for instance I were to host a malicious website on a compromised network (poisoned DNS server) to steal information from you, you wouldn't see it at all and you could like I don't know your Google credentials or bank or anything.

For technical user they might be able to see the issue but imagine being a non tech user and opt-in to this dangerous setting without understanding all the risk or the admin tells you to opt-in.

bgoncal: Related to the toast, we should not expose technical terms to end users right away, let's follow that error screent that iOS has which first shows a user friendly text such as "We couldn't connect to Home Assistant" then the technical details inside an "Advanced/More Info" part of the error screen.

tzagim: I understand the concern about an admin being able to push malicious domains if this was a server‑controlled setting. My proposal is to make this a local, per‑device blocklist stored only in the Android app.

1. The list is empty by default.

2. Only the end‑user can add or remove domains.

It never syncs with the HA server or any admin‑controlled config. This way, the feature empowers users to block specific external resources (e.g., analytics scripts like Google Tag Manager) without introducing the risk of an admin injecting harmful domains. Would you be open to considering such a local‑only implementation.

Thanks @bgoncal , I’ll try to implement your idea as well.

dshokouhi: We need to remove the proceed call. Even Google documentation says not to proceed. We need to follow the guidelines put forth by Google.

https://developer.android.com/reference/android/webkit/SslErrorHandler#proceed()

tzagim: Hi @TimoPtr, @bgoncal, @dshokouhi 
I’ve implemented an update that introduces two new capabilities:

*External Domain Blocking*

1. Added a new preference (block_external_domains_enabled) in Settings.

2. Created a DomainBlocklist utility to store and normalize blocked domains locally (lowercase + IDN normalization).

3. Updated WebViewActivity to check this list and block requests to listed domains before an SSL error occurs.

*Friendly SSL Error Screen for External Resources*

1. Added FriendlyErrorActivity with a dedicated layout that shows a user‑friendly message and an “Advanced” button to reveal technical details.

2. In onReceivedSslError, if the error is for a domain other than the active server, this screen is shown instead of the standard SSL error UI.

*Goals of the change:*

1. Improve user experience for SSL errors on external resources by showing a clear, friendly message instead of a technical error page.

2. Give users control over blocking external resources for improved security and privacy.

Please let me know if this approach works for you, and if there are any changes you’d like to see before merging.

dshokouhi: I will be honest I am not a fan of the app maintaining any list as proposed, I am in favor of making the error message more user friendly so the user can tell what they need to do to fix the issue.

The proposed code changes also feel like they were written by AI.

TimoPtr:
The code is not working (nothing is adding a domain to the whitelist), and to follow @dshokouhi I don't feel confident about doing this, it doesn't seem like a recurring issue for most of our users. It means that it is something that we have to maintain in the future.

It's nice of you to contribute and it's ok to use AI, but when you do so you need to do few things
- read the contributor documentation https://developers.home-assistant.io/docs/android/ to guide your agent properly. Today in the PR I see patterns that are not allowed in our best practices.
- heavily test the application to ensure it does what you are asking for. Here the feature simply doesn't work
- self review the PR before asking for review

I would encourage you to change this PR to only display the error message and nothing else. @bgoncal is correct but for now we don't have such screen. Either you can try to make one, or for now you stick to the exiting error dialog that we use and open an issue to improve the whole dialog in a second step (for all the errors not only this one).

PR #5730, opened 2025-09-02
Imrpove UI/UX of todo widget

TimoPtr:
## Summary

Based on feedbacks from @balloob and @joostlek on the android discord project [channel](https://discord.com/channels/330944238910963714/1346948551892009101/1411051649572798505). This PR remove the `active` title and auto select the first entity from the entity picker if none are currently selected.

## Checklist

## Screenshots

image

## Any other notes

Added a mutex to ensure race condition when updating a widget.
I kept the `complete` section since this section is already optional.


jpelgrom:
PR description screenshot shows the old design 🙈

>  I kept the complete section since this section is already optional.

Wasn't the feedback to [also](https://discord.com/channels/330944238910963714/1346948551892009101/1411333448995438632) remove that header? Removing Active but keeping Completed looks weird to me.

TimoPtr:
> PR description screenshot shows the old design 🙈
> 
> >  I kept the complete section since this section is already optional.
> 
> Wasn't the feedback to [also](https://discord.com/channels/330944238910963714/1346948551892009101/1411333448995438632) remove that header? Removing Active but keeping Completed looks weird to me.

Well I'm not sure about removing it. It is like this also in Google keep they have a title for the completed items and since it's optional I thought we could keep it. But happy to remove it if it doesn't make sense.

jpelgrom:
I clicked approve because both options are valid, we did try to match Keep closely last time but it was specific feedback.

("looks weird" also could be read as "need to get used to it")

balloob:
I think that we can remove it. It's double to have the label and the checkbox

TimoPtr:
> I think that we can remove it. It's double to have the label and the checkbox

It looks like this. I do prefer with the header, but if we agree to remove it let's remove it.

image

image

jpelgrom:
Looks good enough to me, but fine with both options.

TimoPtr:
> Looks good enough to me, but fine with both options.

All right then I'll remove it 👍🏻

PR #5731, opened 2025-09-02
Add new onboarding server discovery

TimoPtr:
## Summary

This PR adds the server discovery screen after the welcome screen. 
It contains also few new things
- `HATopBar` a generic composable built as a wrapper on top of `TopAppBar`
- `Flow.delayFirst` a custom flow operator to add a fake delay to the screen while not blocking the discovery process
- `HomeAssistantSearcher` a new version of the one in `:app` but using a simpler interface and with unit tests

## Checklist

## Screenshots

https://github.com/user-attachments/assets/e545fbb3-3a94-49d4-ae6b-ded8069f8edc

https://github.com/user-attachments/assets/f59174b1-f943-467b-9c4a-b2b5afba68c7

## Any other notes

The PR contains some TODO that needs to be address before the release of the whole onboarding. For now It's mostly notes for us to not forget.


jpelgrom:
Design feedback based on your video:
- The transition when a bottom sheet pops up looks rough, it dims the background too quickly
- Maybe round the bottom sheet corners a bit more?
- The different spinners before an instance are found and after one has been found are jarring. Shouldn't the second spinner also use HA blue?

onboarding/src/screenshotTestDebug/reference/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryScreenshotTest/ServerDiscoveryScreen with multiple servers found_Light_b29dc7a7_0.png
jpelgrom, line 1: "with multiple servers found" - doesn't look like it

TimoPtr: I don't know why but the preview is broken at the moment you have a lazy list :/ even in the IDE.

jpelgrom: Compose previews in Android Studio misbehaving is something I'm used to, but that does make it a bad screenshot test ;)

TimoPtr: I decided to keep it so if one day an update fix it we actually see it.

TimoPtr:
> Maybe round the bottom sheet corners a bit more?

Thanks for catching that it was a mistake from my side while looking at the reference design. Fixed

> The transition when a bottom sheet pops up looks rough, it dims the background too quickly

I had two implementation and I was not able to say what was wrong, but it seems that my simplistic implementation of the Modal makes the animation go wrong. I've made some adjustment using an intermediate state and use the hide() function to let the modal plays its animation correctly 👍🏻 

>  The different spinners before an instance are found and after one has been found are jarring. Shouldn't the second spinner also use HA blue?

Good catch 👍🏻, I'm gonna open another PR with a dedicated composable for loading. I need to adjust it also in the Launcher.

TimoPtr:
> The different spinners before an instance are found and after one has been found are jarring. Shouldn't the second spinner also use HA blue?

I have updated the loading animation with `HALoading`. It is still not the same blue, since the first blue is using the brand color where the other is using the primary color.

image
image

TimoPtr:
After a discussion with the design team, the dots are going to be using the same colors as the loader (but not the HA icons that will continue to use the brand colors)
image

TimoPtr:
Since screenshot are not working for the list of server I wanted to share how it looks on big screen
image

image

TimoPtr:
> Design feedback based on your video:
> 
>     * The transition when a bottom sheet pops up looks rough, it dims the background too quickly
> 
>     * Maybe round the bottom sheet corners a bit more?
> 
>     * The different spinners before an instance are found and after one has been found are jarring. Shouldn't the second spinner also use HA blue?

All your remarks for the design should have been addressed. I'm going to update the video in the description

TimoPtr:
> Since screenshot are not working for the list of server I wanted to share how it looks on big screen image
> image

I have some doubt on the max width size that I'm discussing, I might update the value.

bgoncal:
In iOS there is a "readable margins layout guide", made to make any text into a comfortable width size so user dont need to move the head/eyes that much, check if you dont have the same in Android

common/src/main/kotlin/io/homeassistant/companion/android/util/FlowUtil.kt
```diff
@@ -18,3 +26,36 @@ fun <T> Flow<T>.throttleLatest(delayMillis: Long): Flow<T> = this
...
+    }.onEach {
+        val currentTime = clock.now()
+        if (currentTime < endDelay) {
+            delay(endDelay - currentTime)
+        }
+    }
+}
+
+/**
+ * Delays the emission of the first item from the upstream flow by the specified [delayDuration].
```
jpelgrom:
```suggestion
 * Delays the first emission of items from the upstream flow by the specified [delayDuration].
```

 - It delays emission of all items received for the specified duration, not just the first item
 - It delays the first emission from the upstream flow start; not the first item received (for example: with a delay of 500ms and the first item arriving after 1s, with your original text it would arrive at 1.5s, whereas it actually arrives at 1s)

(These are behaviors you might want to test)

common/src/main/kotlin/io/homeassistant/companion/android/util/FlowUtil.kt
```diff
@@ -18,3 +26,36 @@ fun <T> Flow<T>.throttleLatest(delayMillis: Long): Flow<T> = this
...
+ * Subsequent items are emitted without delay. This is useful when you want to delay the
+ * initial processing of a flow without delaying the start of the upstream flow itself.
+ * It also adds a buffer to the flow to ensure that the upstream flow can continue emitting items
+ * while the first item is being delayed.
+ *
+ * Unlike using `onStart { delay(duration) }`, which postpones the start of the entire flow,
+ * this operator allows the upstream flow to start immediately, and only delays the downstream
+ *
```
jpelgrom:
```suggestion
 * this operator allows the upstream flow to start immediately, and only delays the downstream flow.
```

common/src/screenshotTest/kotlin/io/homeassistant/companion/android/compose/composable/HAModalScreenshotTest.kt
```diff
@@ -0,0 +1,51 @@
...
+    @PreviewLightDark
+    @PreviewTest
+    @Composable
+    fun `HAModalBottomSheet hidden`() {
+        HAModalBottomSheetTest(sheetValue = Hidden)
+    }
+
+    @Composable
+    private fun HAModalBottomSheetTest(sheetValue: SheetValue) {
+        // TODO update to HAThemeForPreview once https://github.com/home-assistant/android/pull/5733 is merged
```
jpelgrom: It has been merged and you're using it, outdated comment?

common/src/screenshotTest/kotlin/io/homeassistant/companion/android/compose/composable/HAModalScreenshotTest.kt
```diff
@@ -0,0 +1,51 @@
...
+        HAThemeForPreview {
+            HAModalBottomSheet(
+                bottomSheetState = SheetState(
+                    skipPartiallyExpanded = false,
+                    Density(LocalDensity.current.density),
+                    initialValue = sheetValue,
+                ),
+            ) {
+                Text("Hello :wave:", style = HATextStyle.Headline)
```
jpelgrom: Let's put the actual emoji in then

```suggestion
                Text("Hello 👋", style = HATextStyle.Headline)
```

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/HomeAssistantSearcher.kt
```diff
@@ -0,0 +1,254 @@
...
+        }
+
+        return serviceFlow
+            .onCompletion {
+                hasCollector.set(false)
+            }
+            .onStart {
+                FailFast.failWhen(hasCollector.get()) {
+                    "Something already started to collect, this flow is designed to only be collected by one collector at the time."
+                }
```
jpelgrom, line 79: Duplicate of line 69, I think, or do you see a race condition happening here?

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/HomeAssistantSearcher.kt
```diff
@@ -0,0 +1,254 @@
...
+         * The static [AtomicBoolean] acts as a safeguard against potential Hilt misconfiguration
+         * (hence its static nature) or unintended multiple concurrent collections of the flow.
+         * It ensures that only one collector can be active at any given time.
+         */
+        @VisibleForTesting val hasCollector = AtomicBoolean(false)
+    }
+
+    override fun discoveredInstanceFlow(): Flow<HomeAssistantInstance> {
+        FailFast.failWhen(hasCollector.get()) {
+            "Something has already call instancesFlow() and didn't close the flow yet."
```
jpelgrom:
```suggestion
            "Something has already called discoveredInstanceFlow() and didn't close the flow yet."
```

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/HomeAssistantSearcher.kt
```diff
@@ -0,0 +1,254 @@
...
+
+        override fun onDiscoveryStopped(serviceType: String?) {
+            Timber.d("Service discovery stopped")
+        }
+
+        override fun onServiceFound(serviceInfo: NsdServiceInfo?) {
+            serviceInfo?.let {
+                Timber.i("Service discovery found HA: $serviceInfo")
+                trySend(serviceInfo)
+            } ?: Timber.w("Service found but is null skipping")
```
jpelgrom:
```suggestion
            } ?: Timber.w("Service found but info is null, skipping")
```

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/HomeAssistantSearcher.kt
```diff
@@ -0,0 +1,254 @@
...
+        override fun onServiceLost(serviceInfo: NsdServiceInfo?) {
+            Timber.d("Service lost: $serviceInfo")
+        }
+
+        override fun onStartDiscoveryFailed(serviceType: String?, errorCode: Int) {
+            close(DiscoveryFailedException("Start discovery failed with error code $errorCode"))
+        }
+
+        override fun onStopDiscoveryFailed(serviceType: String?, errorCode: Int) {
+            Timber.e("Stop discovery failed: Error code:$errorCode")
```
jpelgrom:
```suggestion
            Timber.e("Stop discovery failed with error code $errorCode")
```

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryScreen.kt
```diff
@@ -0,0 +1,464 @@
...
+import io.homeassistant.companion.android.common.compose.theme.LocalHAColorScheme
+import io.homeassistant.companion.android.common.compose.theme.MaxButtonWidth
+import io.homeassistant.companion.android.common.data.HomeAssistantVersion
+import io.homeassistant.companion.android.compose.HAPreviews
+import io.homeassistant.companion.android.compose.alpha
+import io.homeassistant.companion.android.compose.composable.HATopBar
+import io.homeassistant.companion.android.onboarding.R
+import java.net.URL
+
+private val ICON_SIZE = 64.dp
```
jpelgrom, line 87: Also defined exactly the same in the previous screen - should this be a shared constant somewhere else?

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryScreen.kt
```diff
@@ -0,0 +1,464 @@
...
+            onConnectClick = onConnectClick,
+            onDismiss = onDismissOneServerFound,
+            discoveryState = discoveryState,
+        )
+    }
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+private fun OneServerFound(
```
jpelgrom, line 145: A bottom sheet when one is found is a weird flow to me, why not show it in the list?

onboarding/src/screenshotTestDebug/reference/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryScreenshotTest/ServerDiscoveryScreen with one server found_phone_landscape_9e00b29d_0.png
jpelgrom, line 1: This is technically correct but it may not be obvious to the user to drag the sheet up to find and tap the connect button. Consider a smaller icon or dropping the icon if height is limited.

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/HomeAssistantSearcherImplTest.kt
```diff
@@ -0,0 +1,299 @@
...
+    @CsvSource(
+        nullValues = ["null"],
+        value = [
+            "null,null",
+            "http://localhost:8123,null",
+            "http://localhost:8123,wrong_version",
+            "malformed://localhost:8123,null",
+        ],
+    )
+    fun `Given one discoverable instance with wrong attributes when discoveredInstanceFlow is called then nothing is emit`(
```
jpelgrom:
```suggestion
    fun `Given one discoverable instance with wrong attributes when discoveredInstanceFlow is called then nothing is emitted`(
```

Also should be placed below the next test (place impossible to resolve before failures when resolving)

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/HomeAssistantSearcherImplTest.kt
```diff
@@ -0,0 +1,299 @@
...
+            resolveListener.captured.onServiceResolved(serviceInfo)
+
+            expectNoEvents()
+        }
+
+        verifyDiscoveryStartedAndStopped(discoveryListener.captured, resolveListener.captured)
+    }
+
+    @Test
+    fun `Given one discoverable instance impossible to resolve when discoveredInstanceFlow is called then nothing is emit`() = runTest {
```
jpelgrom:
```suggestion
    fun `Given one discoverable instance impossible to resolve when discoveredInstanceFlow is called then nothing is emitted`() = runTest {
```

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/HomeAssistantSearcherImplTest.kt
```diff
@@ -0,0 +1,299 @@
...
+            resolveListener.captured.onResolveFailed(serviceInfo, 0)
+
+            expectNoEvents()
+        }
+
+        verifyDiscoveryStartedAndStopped(discoveryListener.captured, resolveListener.captured)
+    }
+
+    @Test
+    fun `Given one discoverable instance with null service when discoveredInstanceFlow is called then nothing is emit`() = runTest {
```
jpelgrom:
```suggestion
    fun `Given one discoverable instance with null service when discoveredInstanceFlow is called then nothing is emitted`() = runTest {
```

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/HomeAssistantSearcherImplTest.kt
```diff
@@ -0,0 +1,299 @@
...
+            val error = awaitError()
+            assertTrue(error is DiscoveryFailedException)
+            assertEquals("Start discovery failed with error code 0", error.message)
+        }
+
+        verifyDiscoveryStartedAndStopped(discoveryListener.captured, null)
+    }
+
+    @Test
+    fun `Given two instance of discoveredInstanceFlow when collecting at the same time then it throws FailFastException`() = runTest {
```
jpelgrom, line 223: You're testing debug behavior here? Shouldn't you also test production behavior then where it doesn't throw?

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryViewModelTest.kt
```diff
@@ -0,0 +1,197 @@
...
+        searcher = mockk()
+        every { searcher.discoveredInstanceFlow() } returns discoveredInstanceFlow
+    }
+
+    private fun createViewModel() {
+        viewModel = ServerDiscoveryViewModel(searcher)
+    }
+
+    @Test
+    fun `Given initial state when view model created then state is Started`() = runTest {
```
jpelgrom:
```suggestion
    fun `Given initial state when view model created then discovery state is Started`() = runTest {
```

Also for the next test tiles, ViewModel state is not what you're checking and could be different (lifecycle).

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryViewModelTest.kt
```diff
@@ -0,0 +1,197 @@
...
+            assertEquals(Started, expectMostRecentItem())
+            advanceTimeBy(DELAY_BEFORE_DISPLAY_DISCOVERY)
+            runCurrent()
+            val discoveredState = expectMostRecentItem()
+            assertTrue(discoveredState is ServerDiscovered)
+        }
+    }
+
+    @Test
+    fun `Given one server found before TIMEOUT when view model created then state is ServerDiscovered`() = runTest {
```
jpelgrom: There is no constant named `TIMEOUT`

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/HomeAssistantSearcher.kt
TimoPtr: Well you can get the flow but not collecting it and getting the flow a second time. So I prefer to be safe here.

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/HomeAssistantSearcherImplTest.kt
TimoPtr: Well in prod it won't do anything except logging. The behavior is going to be undefined I would say and would depends on the device.

common/src/main/kotlin/io/homeassistant/companion/android/util/FlowUtil.kt
TimoPtr: So if we have item0 at t0 = 0s, item1 at t1 = 100ms, item2 at t2 = 550ms then we would received 
item0 and item1 at 500ms then item2 at 550ms.

I think the existing tests are covering that. Maybe the title are wrong or I miss one?
```kotlin
    @Test
    fun `Given event sent before delay when consuming then should return all after delay without delay`() = runTest {
        flowOf(1, 2, 3)
            .delayFirst(1.seconds)
            .test {
                expectNoEvents()
                advanceTimeBy(500.milliseconds)
                expectNoEvents()
                advanceTimeBy(500.milliseconds)
                assertEquals(1, awaitItem())
                assertEquals(2, awaitItem())
                assertEquals(3, awaitItem())
                awaitComplete()
            }
    }
```

all the items are sent before the delay.

TimoPtr: But your suggestion is indeed correct I'm updating the doc accordingly.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryScreen.kt
TimoPtr: I've open the discussion with the design team to add more size in `HASize` to cover the icon size.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryScreen.kt
TimoPtr: Because most of the time you will only ever see one server. To make the flow more smooth having a modal is nicer. Like when your phone is coming near by a BLE device like a headset/earbuds

image

I only found this ugly screenshot ... But this appears out of nowhere when a device is near by in a specific mode. 

CC @bgoncal

onboarding/src/screenshotTestDebug/reference/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryScreenshotTest/ServerDiscoveryScreen with one server found_phone_landscape_9e00b29d_0.png
TimoPtr: Icon needs to revisited for sure. For the size I think this specific case is fine since the user is in landscape with a phone. He can expect to scroll don't you think?

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryViewModelTest.kt
TimoPtr: I've updated the titles using AI I had hard time to properly name them...

common/src/main/kotlin/io/homeassistant/companion/android/util/FlowUtil.kt
jpelgrom: > So if we have item0 at t0 = 0s, item1 at t1 = 100ms, item2 at t2 = 550ms then we would received
item0 and item1 at 500ms then item2 at 550ms.

Correct

> I think the existing tests are covering that. Maybe the title are wrong or I miss one?

The one quoted covers >1 items before the delay are received after the delay ends.

I don't believe a test covers "(for example: with a delay of 500ms and the first item arriving after 1s, with your original text it would arrive at 1.5s, whereas it actually arrives at 1s)": you never test for the first item arrives after the initial delay has passed is delivered instantly.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryScreen.kt
jpelgrom: OK, but it doesn't have to be an "official" size. We can also define sizes we re-use because they're handy.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryScreen.kt
jpelgrom: The fast pair dialog, yes, also used for Matter accessoires.

But IMO the critical difference is: that dialog is from the system/Play Services, on top of whatever other app you were using, with a different action. (Like a very prominent heads up notification.) The proposed dialog is from the same app, with the same action: search result + connect.

We do use a bottom sheet like that for Improv, but that is a different flow on top of the existing screen, and you're doing the entire flow in the bottom sheet.

I won't block it, but am still not convinced.

onboarding/src/screenshotTestDebug/reference/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryScreenshotTest/ServerDiscoveryScreen with one server found_phone_landscape_9e00b29d_0.png
jpelgrom: Maybe, guess we'll see if it is obvious enough.

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/HomeAssistantSearcherImplTest.kt
jpelgrom: Do we have a guarantee tests will always run like a debug build?

onboarding/src/main/res/values/strings.xml
```diff
@@ -1,6 +1,9 @@
 <resources>
     <string name="home_assistant_branding_icon_content_description">Home\u00A0Assistant branding icon</string>
     <string name="loading_content_description">Loading</string>
+    <string name="searching_home_network">Searching on\nhome network</string>
+    <string name="server_discovery_no_server_info">Check that your Home\u00A0Assistant is powered on and you\'re connected to the same network. You can enter the address manually if you know it.</string>
```
jpelgrom:
```suggestion
    <string name="server_discovery_no_server_info">Check that your Home\u00A0Assistant server is powered on and you\'re connected to the same network. You can enter the address manually if you know it.</string>
```

Updated initial screen text now includes "server" so let's use that here as well

jpelgrom:
Potential issue with (scrolling on) small screens: the scrollable area does not include the title or 'Enter address manually' button. In my opinion, scroll should be on the outermost content container, so including both of those elements.

On my average size Pixel 7a with default size, that means I have approximately one discovered server row of height of space to scroll (I have to scroll down for the loader). On a smaller screen/with a higher zoom levels, not even one row discovered may fit.

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/HomeAssistantSearcherImplTest.kt
TimoPtr: Well in this test I do set a custom handler so the behavior is controller and I can assert the condition that triggers the FailFast invocation. So we know that in debug it will crash and in release it would just log. But if one day we decide to change this behavior the test will continue to work since we are actually asserting that it trigger FailFast that's all.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryScreen.kt
TimoPtr: Indeed but at the same time I'm not sure we should normalize the size of icons from different screen. If we do that and the size in the end was not standard we might change the look of a screen that we don't want. A bit like the `MaxContentSize` that I kept at the screen level even if for now it uses the `MaxButtonSize`.

Let see how it evolves

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryScreen.kt
TimoPtr: Let see how the users are going to react to this. Another solution would be to have the full screen to be the modal and have a button `Not this serve` that would redirect to the list. The main idea is to not show the list in most of the case to streamline the onboarding, very few users have multiple instances on their network.

jpelgrom (approved):
Looks like everything was addressed, thanks for your patience 🙏

No more comments for now, also because it is a dev only module, if needed we can fix things later.

PR #5736, opened 2025-09-04
Add onboarding manual server setup

TimoPtr:
## Summary

This PR adds the manual server setup screen to the new onboarding flow.

## Checklist

## Screenshots

| Light | Dark |
|---|---|
| image | image |
| image | image |

## Any other notes

The PR contains some TODO that needs to be address before the release of the whole onboarding. For now It's mostly notes for us to not forget.


jpelgrom:
UX question: has it been considered to pre-fill http/https and/or auto-suggest it if that is the only thing missing and/or make it a dropdown? Makes it a bit easier when typing a domain name or IP address, just like your browser will add that to the address if not present.

TimoPtr:
> UX question: has it been considered to pre-fill http/https and/or auto-suggest it if that is the only thing missing and/or make it a dropdown? Makes it a bit easier when typing a domain name or IP address, just like your browser will add that to the address if not present.

We didn't go into the details of this but here how it looks on iOS 
image
we could probably do something similar.

@bgoncal do you have any suggestions?

bgoncal:
Actually I have updated this screen in iOS to match the new styling we are designing, what is in figma is wrong, I'll send a screenshot here in a minute

bgoncal:
Nevermind, I remember it wrong, it's on my list to update it, put an image, title and text field like the other onboarding screens but I haven't done it yet, for now it looks like this 

image

TimoPtr:
I propose that we merge this and iterate after on adding suggestion when we agree on how it should behave and looks. 
We could go quite far like detecting if it is IPv4 if so hint also the 8123 port.

@bgoncal @jpelgrom

jpelgrom:
When typing text it immediately shows an error state. Technically correct, but maybe add a small delay in validation or displaying the error so you don't get it while actively typing?

This looks wrong, padding cuts off the button. Padding should be inside the scrollable container.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/manualserver/ManualServerScreen.kt
```diff
@@ -0,0 +1,216 @@
...
+            )
+        },
+        keyboardOptions = KeyboardOptions(imeAction = ImeAction.Done),
+        keyboardActions = KeyboardActions(
+            onDone = {
+                if (isServerUrlValid) {
+                    onConnectClick()
+                }
+                // This is going to hide the keyboard in any case
+                defaultKeyboardAction(ImeAction.Done)
```
jpelgrom, line 187: \+ also clear focus from the text input field

TimoPtr: Wdym ? you want me to update the description or call the freefocus?

TimoPtr:
> When typing text it immediately shows an error state. Technically correct, but maybe add a small delay in validation or displaying the error so you don't get it while actively typing?

I'm going to only display the error after the 8th char since the first part should be http:// wdyt?

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/manualserver/ManualServerScreen.kt
jpelgrom: Also call clearFocus - so when tapping done on the keyboard, the keyboard hides (= default action) and the text field is no longer selected.

jpelgrom (approved):

PR #5759, opened 2025-09-08
Add onboarding connection screen

TimoPtr:
## Summary

This screen is the equivalent of `AuthenticationFragment` fully rewritten to follow the new architecture and adding all the tests missing. I duplicated the `HAWebView` since I was cold feet about moving it to `common` but happy to hear your thought about it.

This also could be reuse as a base to rewrite the whole `WebViewActivity`, take this in consideration when reviewing.

## Checklist

## Screenshots

See inside the PR

## Any other notes

For now it doesn't support Deeplink and the returned auth code is not persisted anywhere.


dshokouhi:
> I duplicated the `HAWebView` since I was cold feet about moving it to `common` but happy to hear your thought about it.

I dont think that would work because WebView does not exist on Wear OS which has common included?

TimoPtr:
> > I duplicated the `HAWebView` since I was cold feet about moving it to `common` but happy to hear your thought about it.
> 
> I dont think that would work because WebView does not exist on Wear OS which has common included?

Well it would since at runtime it won't be executed, but it's bad to add code that won't be use... Another option would be to break down common or create another module... For now I think we can live with the duplicate until I reach the end of the new onboarding.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/navigation/ConnectionNavigation.kt
```diff
@@ -0,0 +1,77 @@
...
+    onBackClick: () -> Unit,
+) {
+    val context = LocalContext.current
+    LaunchedEffect(viewModel) {
+        viewModel.navigationEventsFlow.collect {
+            when (it) {
+                is ConnectionNavigationEvent.Authenticated -> onAuthenticated(it.url, it.authCode)
+                is ConnectionNavigationEvent.Error -> {
+                    // TODO use full screen error when available
+                    onShowSnackbar(
```
jpelgrom, line 65: I like using Snackbars more, but these error messages can be quite long (too long for the Snackbar) + the function only shows them for a short duration so they are easily missed.

jpelgrom:
Screenshot tests are even less useful with webview...

I did expect to see screenshot tests of the error handling. Maybe not specific to this screen, but it is the only place so far you're using Snackbars.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/navigation/ConnectionNavigation.kt
TimoPtr: I left a TODO here because I do share the same feeling. Unfortunately today we are lacking a full screen to display an error. I did start working on one but it not done and I would like designer PoV. So in the scope of this PR I've decided to keep it in a snackbar and redirect the user back.

TimoPtr:
> Screenshot tests are even less useful with webview...
> 
> I did expect to see screenshot tests of the error handling. Maybe not specific to this screen, but it is the only place so far you're using Snackbars.

I should have tested this into the `HAAppTest` but unfortunately we have to use a coroutine to show the snackbar and because of that the screenshot would not work like for the LazyColumn.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionScreen.kt
```diff
@@ -0,0 +1,90 @@
...
+                modifier = Modifier
+                    .fillMaxSize()
+                    .windowInsetsPadding(WindowInsets.safeDrawing),
+                configure = {
+                    this.webViewClient = webViewClient
+                    loadUrl(url)
+                },
+                onBackPressed = onBackClick,
+            )
+        } ?: FailFast.fail { "ConnectionScreen: url is null" }
```
jpelgrom: Doesn't this introduce a race condition? url is collected from the urlFlow, which has a null default value. In the viewmodel's init, you do call a function to build it, but it's inside a `launch` so runs async.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
```diff
@@ -0,0 +1,237 @@
...
+                    .build()
+                    .toString(),
+            )
+        } catch (e: Exception) {
+            Timber.e(e, "Unable to build authentication URL")
+            _navigationEventsFlow.emit(ConnectionNavigationEvent.Error(R.string.connection_screen_malformed_url))
+        }
+    }
+
+    private fun shouldRedirect(view: WebView?, url: Uri): Boolean {
```
jpelgrom: This also does the actual navigation event, so it is not only "should". 

Initially I was a bit confused about the purpose because redirects can also happen inside a webview (webpage redirect). I'd suggest to change the name to be more descriptive, like `interceptRedirectIfRequired`.

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
```diff
@@ -0,0 +1,392 @@
...
+        turbineScope {
+            val urlFlow = viewModel.urlFlow.testIn(backgroundScope)
+            val isLoadingFlow = viewModel.isLoadingFlow.testIn(backgroundScope)
+            val navigationEventsFlow = viewModel.navigationEventsFlow.testIn(backgroundScope)
+
+            // Initial state
+            assertTrue(isLoadingFlow.awaitItem())
+            assertEquals(null, urlFlow.awaitItem())
+
+            val expectedAuthUrl = "$baseUrl/auth/authorize?response_type=code&client_id=https://home-assistant.io/android&redirect_uri=homeassistant://auth-callback"
```
jpelgrom: Client ID is hardcoded, maybe use the constant? That is not part of the ConnectionViewModel so should not be tested here.

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
```diff
@@ -0,0 +1,392 @@
...
+            advanceUntilIdle()
+
+            assertEquals(expectedAuthUrl, urlFlow.awaitItem())
+
+            viewModel.webViewClient.onPageFinished(mockk(), null)
+
+            assertFalse(isLoadingFlow.awaitItem())
+
+            navigationEventsFlow.expectNoEvents() // No errors expected
```
jpelgrom:
```suggestion
            navigationEventsFlow.expectNoEvents() // No authenticated or error events expected
```

jpelgrom (commented):
Have you tested mTLS/client certificate auth in this flow (not writing tests, but running the app, selecting a certificate and seeing it works)?

Not reviewing HAWebView / TLSWebViewClient as it'll be replaced by the one in :app, and if that one needs changes we can discuss when you update that one

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionScreen.kt
TimoPtr: Indeed good catch. I think that it is fast enough on my Pixel 6 that it never happens during my tests. Replacing with a simple log.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
TimoPtr: Speaking of this I'm going to emit a new event instead of calling startActivity in the function.

TimoPtr:
> Have you tested mTLS/client certificate auth in this flow (not writing tests, but running the app, selecting a certificate and seeing it works)?

I honestly didn't I need to bring back my setup. I'll try to do that this afternoon or next week.

TimoPtr:
I tested mTLS and it works as expected. There is a StrictMode violation on Disk read in chromium, but this existed already.

jpelgrom:
Technically it looks OK, just UI notes after testing:

 - When going to the login screen from the server discovered bottom sheet, the bottom sheet hides 'too late' in my opinion (when the transition has finished, instead of when the transition starts).
 - The idea of an intermediate loading screen can be nice, but that one also hides too quickly and you still see the frontend loading:

https://github.com/user-attachments/assets/86985662-3c7b-40e6-a1aa-2927f5d8798d


 - Snackbar errors show on top of the webview, is that supposed to happen? I'd expect the webview to immediately hide and see the Snackbar error on the previous screen:

https://github.com/user-attachments/assets/e298c982-dd7e-492b-a428-1d108e4213e0

PR #5761, opened 2025-09-09
Add onboarding name your device

TimoPtr:
## Summary

This PR add NameYourDevice screen. Compare to the actual flow, once we arrive on this screen if we go back it would go to the server discovery and not back to the webView. Once we've save on this screen the device will be registerer on core.

## Checklist

## Screenshots

image
image

### DIFF
--- a/onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceScreen.kt
+++ b/onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceScreen.kt
@@ -49,8 +49,8 @@ internal fun NameYourDeviceScreen(
     modifier: Modifier = Modifier,
 ) {
     val deviceName by viewModel.deviceNameFlow.collectAsStateWithLifecycle()
-    val saveClickable by viewModel.isSaveClickable.collectAsStateWithLifecycle(false)
-    val isSaving by viewModel.isSaving.collectAsStateWithLifecycle()
+    val saveClickable by viewModel.isSaveClickableFlow.collectAsStateWithLifecycle(false)
+    val isSaving by viewModel.isSavingFlow.collectAsStateWithLifecycle()
 
     NameYourDeviceScreen(
         onBackClick = onBackClick,
--- a/onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
+++ b/onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
@@ -73,9 +73,9 @@ internal class NameYourDeviceViewModel @VisibleForTesting constructor(
 
     private val _isValidNameFlow = MutableStateFlow(isValidName(deviceNameFlow.value))
     val isValidNameFlow = _isValidNameFlow.asStateFlow()
-    private val _isSaving = MutableStateFlow(false)
-    val isSaving = _isSaving.asStateFlow()
-    val isSaveClickable = combine(isSaving, isValidNameFlow) { isSaving, isValidName ->
+    private val _isSavingFlow = MutableStateFlow(false)
+    val isSavingFlow = _isSavingFlow.asStateFlow()
+    val isSaveClickableFlow = combine(isSavingFlow, isValidNameFlow) { isSaving, isValidName ->
         !isSaving && isValidName
     }
 
@@ -87,7 +87,7 @@ internal class NameYourDeviceViewModel @VisibleForTesting constructor(
     fun onSaveClick() {
         viewModelScope.launch {
             try {
-                _isSaving.emit(true)
+                _isSavingFlow.emit(true)
                 val serverId = addServer()
                 _navigationEventsFlow.emit(NameYourDeviceNavigationEvent.DeviceNameSaved(serverId))
             } catch (e: Exception) {
@@ -100,7 +100,7 @@ internal class NameYourDeviceViewModel @VisibleForTesting constructor(
                 }
                 _navigationEventsFlow.emit(NameYourDeviceNavigationEvent.Error(messageRes))
             } finally {
-                _isSaving.emit(false)
+                _isSavingFlow.emit(false)
             }
         }
     }
--- a/onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModelTest.kt
+++ b/onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModelTest.kt
@@ -83,8 +83,8 @@ class NameYourDeviceViewModelTest {
     }
 
     @Test
-    fun `Given viewModelInitialized when isSaveClickable then emits true`() = runTest {
-        viewModel.isSaveClickable.test {
+    fun `Given viewModelInitialized when isSaveClickableFlow then emits true`() = runTest {
+        viewModel.isSaveClickableFlow.test {
             assertTrue(awaitItem())
         }
     }
@@ -350,8 +350,8 @@ class NameYourDeviceViewModelTest {
     }
 
     @Test
-    fun `Given adding server when onSaveClick then isSaving flow is properly updated`() = runTest {
-        viewModel.isSaving.test {
+    fun `Given adding server when onSaveClick then isSavingFlow is properly updated`() = runTest {
+        viewModel.isSavingFlow.test {
             assertFalse(awaitItem())
             coEvery { serverManager.addServer(any()) } coAnswers {
                 assertTrue(awaitItem())
@@ -364,8 +364,8 @@ class NameYourDeviceViewModelTest {
     }
 
     @Test
-    fun `Given currently saving when isSaveClickable collecting then isSaveClickable is updating accordingly`() = runTest {
-        viewModel.isSaveClickable.test {
+    fun `Given currently saving when isSaveClickableFlow collecting then isSaveClickable is updating accordingly`() = runTest {
+        viewModel.isSaveClickableFlow.test {
             assertTrue(awaitItem())
             coEvery { serverManager.addServer(any()) } coAnswers {
                 assertFalse(awaitItem())
@@ -379,8 +379,8 @@ class NameYourDeviceViewModelTest {
     }
 
     @Test
-    fun `Given updating device name when isSaveClickable collecting then isSaveClickable is updating according to isValidDeviceName`() = runTest {
-        viewModel.isSaveClickable.test {
+    fun `Given updating device name when isSaveClickableFlow collecting then isSaveClickable is updating according to isValidDeviceName`() = runTest {
+        viewModel.isSaveClickableFlow.test {
             assertTrue(awaitItem())
 
             viewModel.onDeviceNameChange("")


onboarding/src/main/res/values/strings.xml
```diff
@@ -8,6 +8,10 @@
     <string name="welcome_details">Access your Home\u00A0Assistant server on the go using this app.\n\nHome\u00A0Assistant is open source, advocates for privacy and runs locally in your home.</string>
     <string name="welcome_connect_to_ha">Connect to my Home\u00A0Assistant server</string>
     <string name="welcome_learn_more">Learn more</string>
+    <string name="name_your_device_title">How would you like to name this device?</string>
+    <string name="name_your_device_content">This is used to identify your device in your Home\u00A0Assistant.</string>
```
jpelgrom: Not sure about the emphasis on _your_ HA here.

onboarding/src/main/res/values/strings.xml
```diff
@@ -8,6 +8,10 @@
...
+    <string name="name_your_device_save">Save</string>
+    <string name="name_your_device_clear_name">Clear device name</string>
```
jpelgrom: Very descriptive, but to make it easier for translators maybe it is time to start using "Clear text".

TimoPtr: You mean also for the manual server URL?

TimoPtr:
@jpelgrom I suspect this screen to also have the issue when the keyboard is open on your phone.

onboarding/src/main/res/values/strings.xml
jpelgrom: Yes, so we can have one string instead of n strings different for each text input.

jpelgrom:
> @jpelgrom I suspect this screen to also have the issue when the keyboard is open on your phone.

I didn't test this one on a real device, but looking at your other fix yes I suspect it would have had the same issue.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
```diff
@@ -64,10 +68,30 @@ internal fun NavGraphBuilder.onboarding(
...
         )
+        nameYourDeviceScreen(
+            onBackClick = navController::popBackStack,
+            onDeviceNamed = { serverId ->
+                // TODO if external URL or cloud URL available go to Location otherwise go to local first
+            },
+            onShowSnackbar = onShowSnackbar,
+            onHelpClick = {
+                // TODO validate the URL to use
+                navController.navigateToUri("https://www.home-assistant.io/installation/")
```
jpelgrom, line 96: Definitely past installation now, but I don't think there's a good page on the main site for this step.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
```diff
@@ -0,0 +1,148 @@
...
+
+    private val _isValidNameFlow = MutableStateFlow(isValidName(deviceNameFlow.value))
+    val isValidNameFlow = _isValidNameFlow.asStateFlow()
+
+    fun onDeviceNameChange(name: String) {
+        _deviceNameFlow.update { name }
+        validateName(name)
+    }
+
+    fun onSaveClick() {
```
jpelgrom, line 87: Based on Discord chat: while this is running the ~~continue~~ save button should be disabled to avoid running the request twice

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceScreen.kt
```diff
@@ -0,0 +1,195 @@
...
+        },
+        keyboardOptions = KeyboardOptions(imeAction = ImeAction.Done),
+        keyboardActions = KeyboardActions(
+            onDone = {
+                if (saveClickable) {
+                    onSaveClick()
+                }
+                // This is going to hide the keyboard in any case
+                defaultKeyboardAction(ImeAction.Done)
+            },
```
TimoPtr, line 179: Update text and add clearFocus

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceScreen.kt
```diff
@@ -0,0 +1,195 @@
...
+            saveClickable = saveClickable,
+            onSaveClick = onSaveClick,
+        )
+
+        Spacer(modifier = Modifier.weight(1f))
+
+        HAAccentButton(
+            text = stringResource(R.string.name_your_device_save),
+            onClick = onSaveClick,
+            enabled = saveClickable,
```
TimoPtr, line 139: Disable the button when pressing save or keyboard done to avoid clicking twice

onboarding/src/main/res/values/strings.xml
TimoPtr: I'll fix that once we reach the point of merging this PR otherwise it's going to conflict with the PR on manual server.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
TimoPtr: With @bgoncal we think that we should redo the companion app documentation about onboarding and update the links here.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
TimoPtr: In fact we should disable the text field also when we pressed save and bring it back to life if the register failed.

TimoPtr: done

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
```diff
@@ -0,0 +1,157 @@
...
+            }
+        }
+    }
+
+    private fun validateName(name: String) {
+        _isValidNameFlow.update { isValidName(name) }
+    }
+
+    private fun isValidName(name: String): Boolean {
+        // TODO ?
```
jpelgrom: Is this a question for us the other maintainers or someone else?

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceScreen.kt
```diff
@@ -0,0 +1,211 @@
...
+        ),
+        enabled = deviceNameEditable,
+        modifier = modifier
+            .focusRequester(focusRequester)
+            .testTag(DEVICE_NAME_TEXT_FIELD_TAG),
+    )
+    // Request focus on the text field when the screen is shown
+    LaunchedEffect(Unit) {
+        focusRequester.requestFocus()
+    }
```
jpelgrom: Is this necessary? I'd argue the default name can be good enough. If we want to keep this: maybe put the cursor at the end of the text instead of the start.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceScreen.kt
```diff
@@ -0,0 +1,211 @@
...
+            onDeviceNameChange = {},
+            saveClickable = true,
+            deviceNameEditable = true,
+            onSaveClick = {},
+        )
+    }
+}
+
+// TODO check that we call device registration with the name set
+// TODO write all the tests
```
jpelgrom: - You are doing that
- I'm not sure if it's all you wanted, but I'd remove it or be more specific

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceScreen.kt
TimoPtr: We could indeed not focus the text field.


Btw on iOS most of the devices are named with the name of the user like iPhone of John.
 
We could use the name of the user and the name of the device to build the name of the device in HA.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceScreen.kt
TimoPtr: Leftovers to be removed.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceScreen.kt
```diff
@@ -0,0 +1,211 @@
...
+            onDone = {
+                focusRequester.freeFocus()
+                if (saveClickable) {
+                    onSaveClick()
+                }
+                // This is going to hide the keyboard and clear focus on the text field
+                defaultKeyboardAction(ImeAction.Done)
+            },
+        ),
+        enabled = deviceNameEditable,
```
jpelgrom, line 181: This does not disable the cross to clear the field input. I'm not sure whether that should be handled by this option, or in the trailingIcon compose function.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
TimoPtr: Well I was wondering if we should check something else than this. Can we name a device with only one char?

jpelgrom: Warn the user if the device name is already taken? That creates a confusing setup.

A short device name should work, it can even be blank technically as core will [fallback](https://github.com/home-assistant/core/blob/1bb3c96fc150ac6b8ba2d3da4b4ed570667639b3/homeassistant/components/mobile_app/http_api.py#L85) to the device ID.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceScreen.kt
jpelgrom: > We could indeed not focus the text field.

That or putting it at the end of the text is both fine with me with your explanation, but focusing and putting the cursor at the start when there's already a value makes no sense here.

> We could use the name of the user and the name of the device to build the name of the device in HA.

I expect the user's name to be hidden for apps (couldn't find it in a quick 2 min search) as it would be very easy to profile otherwise. Unless it's one line of code I'd keep changing the default device name out of scope to limit the PR size.

TimoPtr: That should be rather easy to get the current user like we do in the frontend, but we should do it in a separate PR indeed.
https://github.com/home-assistant/android/issues/5816

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceScreen.kt
TimoPtr: I did use the same variable to disable the cross within the trailingIcon since yes setting this doesn't impact the trailingIcon. I guess if we were to wrap even more the composable we could manage it in one variable but for now it is not possible.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceScreen.kt
jpelgrom: I thought you meant name of the Android user (on device)

PR #5770, opened 2025-09-11
Introduce maybeAskForIgnoringBatteryOptimizations extension

TimoPtr:
## Summary

This is a small PR to remove some duplicate code, it is going to be use also in the new onboarding.

## Checklist

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/util/ContextExt.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/util/ContextExt.kt
@@ -43,6 +43,8 @@ fun Context.isAutomotive(): Boolean {
 /**
  * If the app is not already ignoring battery optimizations, this function will open the system
  * settings page to allow the user to grant this permission.
+ *
+ * TODO this should not be exposed to the wear module https://github.com/home-assistant/android/discussions/5771
  */
 fun Context.maybeAskForIgnoringBatteryOptimizations() {
     if (!isIgnoringBatteryOptimizations()) {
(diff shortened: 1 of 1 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/util/ContextExt.kt
```diff
@@ -33,3 +39,30 @@ suspend fun Context.getSharedPreferencesSuspend(name: String, mode: Int = Contex
 fun Context.isAutomotive(): Boolean {
     return packageManager.isAutomotive()
 }
+
+/**
+ * If the app is not already ignoring battery optimizations, this function will open the system
+ * settings page to allow the user to grant this permission.
+ */
+fun Context.maybeAskForIgnoringBatteryOptimizations() {
```
jpelgrom, line 49: This function is not used in the Wear OS app (and you cannot change this setting on Wear OS), so should live in `app` instead.

TimoPtr: I would like to use it in :onboarding. At some point I'm going to split :common but not today.

jpelgrom: Then let's add a TODO note to not forget it and avoid accidental use on Wear OS.

(Also not sure about splitting common for onboarding use, as that module was intended to be temporary. Let's have that discussion later somewhere else, out of scope for this PR.)

TimoPtr: Added

PR #5774, opened 2025-09-12
Introduce HARadioGroup

TimoPtr:
## Summary

Introduce a wrapper around material `RadioButton` specific for HomeAssistant. For now it is quite simple but should cover our needs.

## Checklist

## Screenshots

image
image


common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HARadioGroup.kt
```diff
@@ -0,0 +1,190 @@
...
+            .padding(HASpacing.S),
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        RadioButton(
+            selected = isSelected,
+            enabled = enabled,
+            onClick = null,
+            colors = radioButtonColors(),
+        )
+        Column(modifier = Modifier.padding(start = HASpacing.XS)) {
```
jpelgrom, line 189: Is different padding here compared to the row on purpose? Now left and right of the radio button have different padding values (S vs XS).

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HARadioGroup.kt
```diff
@@ -0,0 +1,190 @@
...
+                        shape = shape,
+                    )
+                } else {
+                    Modifier
+                },
+            )
+            .selectable(
+                enabled = enabled,
+                selected = isSelected,
+                interactionSource = remember { MutableInteractionSource() },
```
jpelgrom: Is this needed? If kept as `null` Compose should create and manage one itself, and you don't appear to use it anywhere else.

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HARadioGroup.kt
```diff
@@ -0,0 +1,190 @@
...
+ * @param modifier An optional [Modifier] to be applied to the radio group.
+ * @param selectedOption The currently selected [RadioOption]. If `null`, no option is selected.
+ * @param spaceBy The vertical spacing between radio buttons. Defaults to [HASpacing.XL].
+ */
+@Composable
+fun <T> HARadioGroup(
+    @SuppressLint("ComposeUnstableCollections") options: List<RadioOption<T>>,
+    onSelect: (RadioOption<T>) -> Unit,
+    modifier: Modifier = Modifier,
+    selectedOption: RadioOption<T>? = null,
```
jpelgrom: Does it make sense to duplicate the option here, instead of using the `selectionKey`?

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HARadioGroup.kt
TimoPtr: image
That's the reference design. The padding to the left of the radio button is the padding of the box and is the same as the one at the top. This XS padding is for the content between the radio button and the text.

@marcinbauer85 maybe you want to double check how it looks?

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HARadioGroup.kt
TimoPtr: Indeed it can be null

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HARadioGroup.kt
TimoPtr: It's not duplicating, but you're right that we could simplify the API with the key directly. I'll make a second version with the key 👍🏻. It's even better because I was comparing the selectedOption instead of the key. Good catch.

TimoPtr: I kept the two version since in some cases I think we might want to keep the RadioOption reference in the state not only the key. It's doesn't add a complexity to keep two version.

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HARadioGroup.kt
jpelgrom: If null you can remove the entire line...

TimoPtr:
> I may be biased towards Material Design, but the border around the option suggests to me that it is _not_ part of a group.

Here a real world example of the group used. We can discussed with the design team if you want on discord about it 👍🏻.

image
image

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HARadioGroup.kt
TimoPtr: Unfortunately no, that's why in the first place I've added it. This version of `selectable` doesn't have default value on `interactionSource`

```kotlin
fun Modifier.selectable(
    selected: Boolean,
    interactionSource: MutableInteractionSource?,
    indication: Indication?,
    enabled: Boolean = true,
    role: Role? = null,
    onClick: () -> Unit,
)
```

jpelgrom: Sorry, I was only glancing at docs instead of ctrl - click in the IDE to check it. Good to keep it like this then.

TimoPtr: I've set it to null, since it is not adding a lot of value to set the value yet. In the future we could potentially take this as a parameter it might be useful.

jpelgrom:
> > I may be biased towards Material Design, but the border around the option suggests to me that it is _not_ part of a group.
> 
> Here a real world example of the group used. 

That doesn't look too bad, but as there is only one group on the screen and you need to choose something to continue you don't need to know that it is a group. My concern is more with 2+ groups.

TimoPtr:
> > > I may be biased towards Material Design, but the border around the option suggests to me that it is _not_ part of a group.
> > 
> > 
> > Here a real world example of the group used.
> 
> That doesn't look too bad, but as there is only one group on the screen and you need to choose something to continue you don't need to know that it is a group. My concern is more with 2+ groups.

For 2 groups I think we would need the design team to help us. I suppose it's a mater of what is around the group and if you add a title in between.

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HARadioGroup.kt
TimoPtr: As you can see I didn't implement the component fully. For now I don't see the need of the icons, but it would be rather easy to implement. I'm thinking that the `RadioOption` could contain a `@Composable () -> Unit` to give some freedom on the content.

jpelgrom: OK. There are some other differences compared to the design you shared (subtitle top margin, subtitle text color, radio button circle alignment to title instead of parent container), but I'm fine with both the design and the current implementation there.

jpelgrom (approved):

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HARadioGroup.kt
TimoPtr: I'm going to merge it like this. But since I've made the first version of the design system the text style were not final. I need to catch up and migrate to the new text styles, In that case it should align the colors.

TimoPtr: https://github.com/home-assistant/android/issues/5793 (no details yet I need to catch up first)

PR #5794, opened 2025-09-15
Add onboarding location sharing screens

TimoPtr:
## Summary

This PR introduces two new onboarding screens related to location permissions and security in the app. The goal is to make users aware of potential risks when sharing (or not sharing) their location, and to give them clearer control over how the app behaves.

### Location Sensor Opt-in

- The first new screen explicitly asks whether the user wants to enable the location sensor, which allows their phone’s position to be sent to Home Assistant.
- This is an optional step users can choose whether or not to share this data.

### Security Awareness for Location Sharing

- The second new screen is shown only if the user declines to enable full location sharing and their setup is deemed insecure (e.g., using http instead of https).
- They are then given two options:
  - Most Secure: Requires the user to define their Home location. If they are not currently at home, the app will display a blocking screen when opened, preventing insecure queries and explaining why they are blocked. This screen will also provide guidance on adjusting location settings.
  - Less Secure: Allows the user to proceed with reduced protection, while clearly stating the risks involved. In this case the user won't ever be blocked

The blocking screen is going to be designed specifically to be easier for non-technical users to understand compared to an overloaded onboarding explanation. For example, it will explain:

> You are not considered at home. We cannot perform queries securely in this setup without risking credential leakage. Would you like to review your Home location settings?

> OR Would you like to give location permission?

Users will also have a way to switch from the most secure option back to less secure if they decide.

The main objective of these changes is to improve user understanding of security trade-offs in relation to location data and connection security. By presenting the information contextually at the time it matters we help users make informed choices about protecting their personal data.

⚠️ Note: This PR is part of the broader work in progress to improve security.

## Checklist

## Screenshots

| Light | Dark |
|---|---|
| image | image |
| image | image |

## Any other notes

A final screen is going to be introduce to get information about the `Home` to help us deciding on whether or not the connection is "secure". 

The permission about battery optimization is going to be asked on the WebViewActivity alongside with the notification permission (with a bottom sheet before explaining why).


jpelgrom:
> The permission about battery optimization is going to be asked on the WebViewActivity alongside with the notification permission (with a bottom sheet before explaining why).

Doesn't this add too much clutter and different designs? You think you're in, but first even more popups...

A bottom sheet is also easier to accidentally dismiss. Keeping it in the onboarding flow at least lets the user know what to expect.

onboarding/src/main/res/values/strings.xml
```diff
@@ -17,7 +17,18 @@
...
     <string name="connection_screen_malformed_url">Verify that your Home\u00A0Assistant address is correct</string>
-    <string name="local_first_title">Local by default.\nRemote when you’re ready.</string>
+    <string name="local_first_title">Local by default.\nRemote when you\'re ready.</string>
     <string name="local_first_content">Your Home\u00A0Assistant is only accessible on your home network. To control your smart home from anywhere, you can set up remote access later in your settings.\n\nFor now, you\'re securely connected to your local network.</string>
     <string name="local_first_next">Got it</string>
+    <string name="location_sharing_title">Use this device\'s location for automations</string>
+    <string name="location_sharing_content">Location sharing enables powerful automations, such as turning off the heating when you leave home. This option shares the device\'s location only with your Home Assistant system.\n\nThis data stays in your home and is never sent to third parties. It also helps strengthen the security of your connection to Home Assistant.</string>
+    <string name="location_sharing_share">Share with my location</string>
```
jpelgrom:
```suggestion
    <string name="location_sharing_share">Share my location</string>
```

onboarding/src/main/res/values/strings.xml
```diff
@@ -17,7 +17,18 @@
...
     <string name="manual_server_wrong_url">Invalid Home\u00A0Assistant address</string>
     <string name="connection_screen_malformed_url">Verify that your Home\u00A0Assistant address is correct</string>
-    <string name="local_first_title">Local by default.\nRemote when you’re ready.</string>
+    <string name="local_first_title">Local by default.\nRemote when you\'re ready.</string>
     <string name="local_first_content">Your Home\u00A0Assistant is only accessible on your home network. To control your smart home from anywhere, you can set up remote access later in your settings.\n\nFor now, you\'re securely connected to your local network.</string>
     <string name="local_first_next">Got it</string>
+    <string name="location_sharing_title">Use this device\'s location for automations</string>
+    <string name="location_sharing_content">Location sharing enables powerful automations, such as turning off the heating when you leave home. This option shares the device\'s location only with your Home Assistant system.\n\nThis data stays in your home and is never sent to third parties. It also helps strengthen the security of your connection to Home Assistant.</string>
```
jpelgrom, line 23: > It also helps strengthen the security of your connection to Home Assistant.

This is added at the end without details, raising questions. How does it help strengthen the security? I understand trying to keep it simple, but maybe you need to add some text from the second screen here as well.

onboarding/src/main/res/values/strings.xml
```diff
@@ -17,7 +17,18 @@
     <string name="manual_server_clear_url">Clear server url</string>
     <string name="manual_server_wrong_url">Invalid Home\u00A0Assistant address</string>
     <string name="connection_screen_malformed_url">Verify that your Home\u00A0Assistant address is correct</string>
-    <string name="local_first_title">Local by default.\nRemote when you’re ready.</string>
+    <string name="local_first_title">Local by default.\nRemote when you\'re ready.</string>
     <string name="local_first_content">Your Home\u00A0Assistant is only accessible on your home network. To control your smart home from anywhere, you can set up remote access later in your settings.\n\nFor now, you\'re securely connected to your local network.</string>
     <string name="local_first_next">Got it</string>
+    <string name="location_sharing_title">Use this device\'s location for automations</string>
```
jpelgrom, line 22: This is one option, but you can also:
 - simply track the device
 - use different URLs
 - disable app lock when at home
It also doesn't immediately mean you get an automation based on device location.

We shouldn't cover all that with one title, but I think the current title is too limited.

TimoPtr: The idea is that if you say no to this screen we disable the location sensor which doesn't disable the features that you've described. In that case if you say no we enter the second screen `LocationForSecureConnection` which speaks about your Home.

With @bgoncal we also want to add a final last screen that asks `Are you Home?` if the user grant the permission on this Location screen. 
This screen would show you your wifi and if you are over VPN and ethernet (this options wouldn't be shown if they are not currently in use).

onboarding/src/main/res/values/strings.xml
TimoPtr: I think we can keep the copy like this for now. I know @laupalombi spent a lot of time already on this screen.

onboarding/src/main/res/values/strings.xml
jpelgrom: > With @bgoncal we also want to add a final last screen that asks Are you Home? if the user grant the permission on this Location screen.

Good idea to make it more explicit, but couldn't it be shown when agreeing to "Use this device's location for automations" or "Let us help secure your remote connection", instead of only the first screen? We still need to confirm the user is currently on their home network is even if you enable location tracking.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/locationsharing/LocationSharingViewModel.kt
```diff
@@ -0,0 +1,56 @@
...
+                        // TODO add sensor ID from `LocationSensorManager` instead of string
+                        "location_background",
+                        "zone_background",
+                        "accurate_location",
+                    ),
+                    serverId = serverId,
+                    enabled = enabled,
+                )
+                // We don't allow the insecure connection if the user give his permission to use location.
+                // This can be overridden in th LocationForSecureConnectionScreen.
```
jpelgrom: Don't assume gender + typo

```suggestion
                // We don't allow the insecure connection if the user gives their permission to use location.
                // This can be overridden in the LocationForSecureConnectionScreen.
```

onboarding/src/main/kotlin/io/homeassistant/companion/android/compose/LocationPermissionState.kt
```diff
@@ -0,0 +1,66 @@
...
+ * @see Manifest.permission.ACCESS_COARSE_LOCATION
+ * @see Manifest.permission.ACCESS_BACKGROUND_LOCATION
+ * @see MultiplePermissionsState
+ * @see rememberPermissionState
+ * @see rememberMultiplePermissionsState
+ */
+@OptIn(ExperimentalPermissionsApi::class)
+@Composable
+fun rememberLocationPermission(onPermissionResult: (Boolean) -> Unit): MultiplePermissionsState {
+    // TODO verify that we should ask on Q because we see R in some other functions
```
jpelgrom: The permission was added in Android 10 (Q), the flow where we need to request it separately is added in Android 11 (R)

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
```diff
@@ -108,10 +121,54 @@ internal fun NavGraphBuilder.onboarding(
...
+                navController.navigateToUri("https://www.home-assistant.io/installation/")
+            },
+            onGotoNextScreen = { serverId ->
+                onOnboardingDone()
+            },
+            onShowSnackbar = onShowSnackbar,
+            // We don't have back button since after name your device the device is registered
+        )
+
+        // TODO ask for background permission (using the next extension) on the dashboard with the notification permission
```
jpelgrom: I think this comment is outdated?

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/locationsharing/LocationSharingScreen.kt
```diff
@@ -0,0 +1,152 @@
...
+
+@Composable
+@OptIn(ExperimentalPermissionsApi::class)
+private fun ColumnScope.BottomButtons(
+    onGoToNextScreen: () -> Unit,
+    onLocationSharingResponse: (enabled: Boolean) -> Unit,
+) {
+    val permissions = rememberLocationPermission(
+        onPermissionResult = {
+            // We ignore the result and proceed even if the user reject the permission and we should proceed
```
jpelgrom:
```suggestion
            // We ignore the result and proceed even if the user rejected the permission
```

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
```diff
@@ -38,8 +38,9 @@ internal sealed interface NameYourDeviceNavigationEvent {
      * Represents the navigation event that is triggered when the device name is saved.
      *
      * @param serverId The ID of the server for which the device name was saved.
+     * @param hasPlainTextAccess Boolean that define if the current server has a plain text url
```
jpelgrom:
```suggestion
     * @param hasPlainTextAccess Boolean that defines if the server has a plain text URL.
```

onboarding/src/main/res/values/strings.xml
```diff
@@ -16,7 +16,18 @@
...
+    <string name="location_sharing_content">Location sharing enables powerful automations, such as turning off the heating when you leave home. This option shares the device\'s location only with your Home Assistant system.\n\nThis data stays in your home and is never sent to third parties. It also helps strengthen the security of your connection to Home Assistant.</string>
+    <string name="location_sharing_share">Share my location</string>
+    <string name="location_sharing_no_share">Do not share my location</string>
+    <string name="location_secure_connection_title">Let us help secure your remote connection</string>
+    <string name="location_secure_connection_content">It appears that your connection is not encrypted (using HTTP). If this app knows when you\'re away from home, it can choose a more secure way to connect to your Home Assistant system if available. This requires location services to be enabled.</string>
```
jpelgrom:
```suggestion
    <string name="location_secure_connection_content">It appears that your connection is not encrypted (using HTTP). If the app knows when you\'re away from home, it can choose a more secure way to connect to your Home Assistant system if available. This requires location services to be enabled.</string>
```

onboarding/src/main/res/values/strings.xml
```diff
@@ -16,7 +16,18 @@
...
+    <string name="location_secure_connection_most_secure">Most secure: Allow this app to know when you\'re home</string>
+    <string name="location_secure_connection_less_secure">Less secure: Do not allow this app to know when you\'re home</string>
+    <string name="location_secure_connection_hint">This data will never be shared with the Home Assistant project or third parties.</string>
+    <string name="location_secure_connection_next">Next</string>
+    <string name="location_secure_connection_discard_permission">You rejected the location permission, you should then choose the Less secure option.</string>
```
jpelgrom: Let the user know they can still choose to allow access + rewrite to make the text with that additional info fit in a Snackbar (although German is going to cause issues...).

```suggestion
    <string name="location_secure_connection_discard_permission">Location permission denied. Allow access or choose the \'Less secure\' option.</string>
```

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/locationsharing/LocationSharingScreen.kt
```diff
@@ -0,0 +1,152 @@
...
+
+    HAAccentButton(
+        text = stringResource(R.string.location_sharing_share),
+        onClick = {
+            onLocationSharingResponse(true)
+            permissions.launchMultiplePermissionRequest()
+        },
+    )
+
+    HAPlainButton(
```
jpelgrom, line 131: Nudging people towards location sharing is OK, but was it considered to make this one filled so it has at least some emphasis? Now it feels like accepting cookies dark patterns, and we know there are quite some people who don't want to use location all the time.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
TimoPtr: We still need to do it, but I didn't want to commit this TODO.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/locationsharing/LocationSharingScreen.kt
TimoPtr: I agree, to avoid too much discussion on this PR.

The plan is to actually give a debug APK, once I merge all the PR and ask you and OHF employees to test the flow and give feedbacks so that we can iterate on the details, copy, design, ...

I propose that we discuss this at this moment if that's fine for you.

TimoPtr:
> Add a new TODO: before replacing the onboarding: expose a toggle for the allow insecure preference. (This feels like something that could be done in the main app before we have the updated onboarding.)

Indeed it can be made before, it's in my list of things to do before replacing the whole flow.

onboarding/src/main/kotlin/io/homeassistant/companion/android/compose/LocationPermissionState.kt
TimoPtr: [Screen_recording_20250929_121314.webm](https://github.com/user-attachments/assets/9d611cde-c21f-40bd-86b7-14aad91ccb29)
Tested on Android 10 and I think it's acceptable to keep it like this. I removed the TODO 👍🏻

PR #5815, opened 2025-09-23
Improve ux server discovery and connection screen

TimoPtr:
## Summary

Based on the feedback https://github.com/home-assistant/android/pull/5759#issuecomment-3320991477 this PR adjust few UI elements and behavior to improve the UX of the server discovery and connection screen.

- Increase the delay of 0.5s before displaying the first server found in the modal
- Connection screen is displayed only when the modal is fully hidden
- An error placeholder has been added to avoid seeing the ugly webview error when there is one

## Checklist

## Screenshots

https://github.com/user-attachments/assets/6ce9c3f1-c43e-4ccd-adbc-f0e9d58c1299


jpelgrom:
There is a conflict in one file.

> An error placeholder has been added to avoid seeing the ugly webview error when there is one

You didn't answer the original question (it wasn't really about the webview error page):

Snackbar errors show on top of the webview, is that supposed to happen? I'd expect the webview to immediately hide and see the Snackbar error on the previous screen.

Also, in [another comment](https://github.com/home-assistant/android/pull/5759#discussion_r2353516152) you suggested that another error screen would be implemented. I think we still need that before onboarding is replaced with this module because of the Snackbar limitations I mentioned regarding message length/timing.

TimoPtr:
I went this way to hide to very ugly error from the webview and I found the UX of displaying the error on the previous screen a bit jumpy. Indeed I want the full screen error with a call to action button, but I think it can be in combination with this one to always make sure we hide the ugly webview error.

PR #5818, opened 2025-09-23
Add onboarding deeplinks (invite and redirect)

TimoPtr:
## Summary

`LauncherActivity` now can be started with a `DeepLink` sealed interface instance. For now it only support Invite and NavigateTo. I decided to keep the `LinkActivity` to handle the server selection and potential modification/validation of the data received on the app. 

The current state of the navigation framework doesn't allow us to use the `deeplink` capability since it has too many restrictions. So for now everything is using the startDestination. Unfortunately the sub-navigation-graph cannot read the value from the startDestination, that's why I passed a `serverToOnboard` parameter to the `OnbardingNavigation` instead of simply using the route. IMO it is also a valid pattern that I'm going to use to morph the onboarding graph for the wear.

I fixed an issue where when we open the Frontend from the Launcher it was keeping the Launcher in the backstack with a white screen. For now it is a workaround with a dummy route that doesn't do anything except redirecting to the activity and finish the current activity. Later we are going to remove this once the WebViewActivity is using the navigation framework.

## Checklist

## Any other notes

For now the `homeassistant://navigate` links are still handled by the WebViewActivity directly.

### DIFF
--- a/onboarding/src/main/kotlin/io/homeassistant/companion/android/launcher/LauncherActivity.kt
+++ b/onboarding/src/main/kotlin/io/homeassistant/companion/android/launcher/LauncherActivity.kt
@@ -37,7 +37,6 @@ class LauncherActivity : AppCompatActivity() {
     }
 
     companion object {
-
         fun newInstance(context: Context, deepLink: DeepLink? = null): Intent {
             return Intent(context, LauncherActivity::class.java).apply {
                 if (deepLink != null) {


onboarding/src/main/kotlin/io/homeassistant/companion/android/launcher/LauncherActivity.kt
```diff
@@ -1,28 +1,59 @@
...
 class LauncherActivity : AppCompatActivity() {
-    private val viewModel: LauncherViewModel by viewModels()
+    @Parcelize
+    sealed interface DeepLink : Parcelable {
+        data class Invite(val url: String) : DeepLink
+        data class NavigateTo(val path: String?, val serverId: Int) : DeepLink
+    }
+
+    companion object {
+
```
jpelgrom:
```suggestion
```

jpelgrom:
Testing LinkActivity: while comparing the current production app and this version, this version is noticeably slower than production and has a weird flash at the top of the screen. Both are cold starts (app removed from recents).

Current prod:

https://github.com/user-attachments/assets/ff513b08-209c-4afb-90d6-51dc7c2bda55

This version with new deeplinks:

https://github.com/user-attachments/assets/e9079151-4cc8-4d08-8dea-9c1025b80a95

jpelgrom:
After [discussion on Discord](https://discord.com/channels/330944238910963714/1346948551892009101/1422953685876740226), I tested a debug build of the latest commit of main and performance/behavior for deeplinks seems similar to this PR branch, so the observed difference is probably because of debug and not the changes in this PR.

https://github.com/user-attachments/assets/822fd33e-96d2-43f1-bca0-f3583d2c51b1

PR #5828, opened 2025-09-26
Update HAButton and HATextField acording to design system

TimoPtr:
## Summary

Home Assistant design system has 3 different basic sizes for buttons that we need to support. The text field was too limited and was not exposing enough of the underlying component. 

## Checklist

## Screenshots

image
image

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HAButtons.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HAButtons.kt
@@ -72,6 +72,10 @@ enum class ButtonSize(val value: Dp) {
  * @param modifier Optional [androidx.compose.ui.Modifier] to be applied to the button.
  * @param variant The [ButtonVariant] that determines the button's color scheme. Defaults to [ButtonVariant.PRIMARY].
  * @param enabled Controls the enabled state of the button. When `false`, the button will not be clickable.
+ * @param size The size of the button. Defaults to [ButtonSize.LARGE].
+ * @param textOverflow How visual overflow should be handled. Defaults to [TextOverflow.Clip].
+ * @param maxLines An optional maximum number of lines for the text to span, wrapping if necessary.
+ *  If the text exceeds the given number of lines, it will be truncated according to [textOverflow].
  * @param prefix Optional composable content to be displayed at the start of the button, before the text.
  * @param suffix Optional composable content to be displayed at the end of the button, after the text.
  */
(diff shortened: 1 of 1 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HAButtons.kt
```diff
@@ -253,17 +285,23 @@ private fun RowScope.ButtonContent(
...
+        maxLines = maxLines,
+        overflow = textOverflow,
         modifier = Modifier
             .padding(
                 // Adjust padding based on the presence of prefix/suffix
                 start = if (prefix != null) HASpacing.X2S else HASpacing.M,
                 end = if (suffix != null) HASpacing.X2S else HASpacing.M,
             )
+            // Minimum padding when the text is multi-lines
+            .padding(vertical = HASpacing.X2S)
```
jpelgrom: But this is now always applied, also with single line content?

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HAButtons.kt
```diff
@@ -253,17 +285,23 @@ private fun RowScope.ButtonContent(
...
     suffix: @Composable (BoxScope.() -> Unit)?,
+    textOverflow: TextOverflow,
+    maxLines: Int,
 ) {
     ButtonDecorator(ButtonDecoratorType.PREFIX, prefix)
     Text(
         text = text,
         style = HATextStyle.Button,
+        maxLines = maxLines,
+        overflow = textOverflow,
```
jpelgrom: It annoys me that this order doesn't match the function arguments' order 😅

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HAButtons.kt
TimoPtr: Yes but you don't see it even on a small button since it's smaller than the container.

jpelgrom: Visually yes it's no issue, but then your comment is false ;)

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HAButtons.kt
TimoPtr: I'll address it 👍. To be fair I'm still struggling with myself adhering to a convention for the parameters order outside of the practical cases like a lambda at the end or default at the end.

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HAButtons.kt
TimoPtr:
```
// Apply a small padding that is only be visible when the text is going to be on multiple lines, to avoid touching the border of the background
````

Something like that might fit better?

jpelgrom: - that is only be visible > that is only visible
 - comment probably needs multiple lines because of length

Otherwise nice improvement 👍

PR #5829, opened 2025-09-26
Stop loading when the webview is released

TimoPtr:
## Summary

In the specific case of the onboarding we are using a WebView with Javascript enabled. It seems that after that we catpure the authcode and proceed with the rest of the onboarding. The WebView is still active and keeps a reference to the `LauncherActivity`. This PR makes sure that when the activity is destroyed we invoke `stopLoading` from the WebView.

image

I've apply the changed to the two `HAWebView` even if we never kill the WebViewActivity.

## Checklist

### DIFF
--- a/onboarding/src/main/kotlin/io/homeassistant/companion/android/compose/composable/HAWebView.kt
+++ b/onboarding/src/main/kotlin/io/homeassistant/companion/android/compose/composable/HAWebView.kt
@@ -47,7 +47,7 @@ fun HAWebView(
         },
         modifier = modifier.testTag(HA_WEBVIEW_TAG),
         onRelease = {
-            Timber.d("OnRelease webview stop loading")
+            Timber.d("onRelease WebView, stopping loading")
             it.stopLoading()
             webview = null
         },


jpelgrom:
The debug statements are a bit much to keep, right? We don't annotate every part of the lifecycle, this is dev only stuff in my opinion.

TimoPtr:
> The debug statements are a bit much to keep, right? We don't annotate every part of the lifecycle, this is dev only stuff in my opinion.

The WebView is central in the app I think it's nice to know when we release it. We should almost never see it.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/HAWebView.kt
```diff
@@ -73,6 +73,10 @@ fun HAWebView(
                 webView.settings.setNightModeTheme(it, uiMode)
             }
         },
+        onRelease = {
+            Timber.d("OnRelease webview stop loading")
```
jpelgrom:
```suggestion
            Timber.d("onRelease WebView, stopping loading")
```

onboarding/src/main/kotlin/io/homeassistant/companion/android/compose/composable/HAWebView.kt
```diff
@@ -46,6 +46,11 @@ fun HAWebView(
             }
         },
         modifier = modifier.testTag(HA_WEBVIEW_TAG),
+        onRelease = {
+            Timber.d("OnRelease webview stop loading")
```
jpelgrom:
```suggestion
            Timber.d("onRelease WebView, stopping loading")
```

PR #5845, opened 2025-09-29
Add onboarding for wear

TimoPtr:
## Summary

This PR adds a dedicated navigation graph for wear. When started it can go to Conenction directly if the URL is specified otherwise it goes to ServerDiscovery with the known instances added. Once authenticated we simply ask the user the name of the device and stop the onboarding.

## Checklist

## Any other notes

This PR also contains a modification of the ConenctionScreen to return the current status of mTLS to know if we need or not to send the cert to the wear. Another PR is going to add the necessary screen to select a cert and the password before quitting the wear onboarding.

The graph might change later once we have everything in the navigation graph since we might not want to use a start destination.


onboarding/src/main/kotlin/io/homeassistant/companion/android/launcher/LauncherActivity.kt
```diff
@@ -81,7 +84,16 @@ class LauncherActivity : AppCompatActivity() {
                                     FrontendRoute(event.path, event.serverId)
                                 }
                             }
+
                             is LauncherNavigationEvent.Onboarding -> OnboardingRoute(event.url)
+                            is LauncherNavigationEvent.WearOnboarding -> {
+                                // TODO fail when not in FULL variant (maybe make a dedicated screen explaining that
+                                //  the wear only work with FULL variant)
```
jpelgrom, line 90: Any Wear option is completely unavailable in other flavors so there is no need to create a screen, or at least not here

jpelgrom:
I'm not exactly sure where to put the comment, but: when Wear onboarding is opened using a link (so you tapped a server URL on the watch), the first back action doesn't work. Only the second time you press/swipe back, it'll actually go back closing the onboarding.

jpelgrom:
I tested the flow as part of #5848, and it seems to complete correctly.

TimoPtr:
> I'm not exactly sure where to put the comment, but: when Wear onboarding is opened using a link (so you tapped a server URL on the watch), the first back action doesn't work. Only the second time you press/swipe back, it'll actually go back closing the onboarding.

I'm going to look at this into a separate PR

TimoPtr:
> > I'm not exactly sure where to put the comment, but: when Wear onboarding is opened using a link (so you tapped a server URL on the watch), the first back action doesn't work. Only the second time you press/swipe back, it'll actually go back closing the onboarding.
> 
> I'm going to look at this into a separate PR

It comes from the fact that the `HAWebView` has a `BackHandler` when we are at the very starts of graph. It kinda reproduce the issue we have in the navigation tests. We stay on the screen because it simply pop the item from the graph but then the backstack is empty with nowhere to go. 

If I remove the `BackHandler` it works as expected but I've added it to avoid checking doUpdateVisitedHistory, I need to see how to work with it.

PR #5849, opened 2025-09-30
Change PressureSensor device_class to atmospheric_pressure

### LINKED ISSUE #5847: Pressure in Android Companion app: device_class should be atmospheric_pressure
2025.8.7-full
**Android version(s):**
16 (GrapheneOS)
**Device model(s):**
Pixel 6a
**Home Assistant version:**
2025.9.4

There are currently two device classes available in Home Assistant for pressure: `pressure` and `atmospheric_pressure`.
Both have the exact same allowed units: cbar, bar, hPa, mmHg, inHg, kPa, mbar, Pa or psi
The `device_class` `atmospheric_pressure` has been added in 2022 via [PR #83455](https://github.com/home-assistant/core/pull/83455) as was requested in [discussion #839](https://github.com/home-assistant/architecture/discussions/839)
This new `device_class` was specially created to distinct between atmospheric pressure and other pressure sensors like for tyre pressure.

The problem however is that the atmospheric pressure sensor as delivered through the Home Assistant Companion app for Android is still using the old `device_class` `pressure`.
This means that if the pressure sensor from the Companion app is combined in a History graph with other pressure sensors that are using the correct `device_class` `atmospheric_pressure` these are, although these have the same unit, shown in separate graphs, like is shown below.
All these sensors have the same unit hPa.

My understanding is that this is caused by not all sensors having the same `device_class`.

So the request is to change the device_class for the Android companion app pressure sensor into `atmospheric_pressure`.

**Companion App Logs:**

```

```

Image

**Additional information:**

TimoPtr:
## Summary

Change the device class of the pressure sensor to use the atmospheric_pressure.

## Checklist

## Any other notes

Fixes #5847

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/PressureSensorManager.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/PressureSensorManager.kt
@@ -27,7 +27,7 @@ class PressureSensorManager :
             commonR.string.sensor_name_pressure,
             commonR.string.sensor_description_pressure_sensor,
             "mdi:gauge",
-            deviceClass = "pressure",
+            deviceClass = "atmospheric_pressure",
             unitOfMeasurement = "hPa",
             stateClass = SensorManager.STATE_CLASS_MEASUREMENT,
         )


TimoPtr:
@jpelgrom It was introduced in 2022 I didn't add backward support. I'm not sure of all the implication of such change today if you can enlighten me.

jpelgrom:
> @jpelgrom It was introduced in 2022 I didn't add backward support. I'm not sure of all the implication of such change today if you can enlighten me.

I'm not sure I fully understand your question. 

For existing users the statistics will have a warning that the device class changed, but otherwise it shouldn't cause issues.

In the code where we register sensors you can add an exception for the device class to avoid breaking support on older installs (looks like atmospheric_pressure first appeared in core 2023.1).

https://github.com/home-assistant/android/blob/45838ad3067970fd1f52ac320abcc7f5ee6caffb/common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt#L616

PR #5855, opened 2025-10-01
Add missing GitHub action explicit permissions

TimoPtr:
## Summary

Fix missing permission properties reported by CodeQL on its first run. The permission needs to be explicit and not to widely open at the workflow level so that we can control during review process what we are exposing. It is also helpful to limit what can a third party action can do in our CI env. 

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistActivity.kt
@@ -42,12 +42,7 @@ class AssistActivity : BaseActivity() {
             serverId: Int = -1,
             pipelineId: String? = null,
             startListening: Boolean = true,
-            fromFrontend: Boolean = true
-
-
-
-  
-
+            fromFrontend: Boolean = true,
         ): Intent {
             return Intent(context, AssistActivity::class.java).apply {
                 putExtra(EXTRA_SERVER, serverId)


.github/workflows/prepareNextRelease.yml
```diff
@@ -13,6 +13,9 @@ jobs:
   tag:
     name: Prepare next release
     runs-on: ubuntu-latest
+    permissions:
+      actions: write
+      contents: write
```
jpelgrom, line 18:
```suggestion
    permissions:
      actions: write
      contents: write
      pull_requests: write
```

It'll also need permissions to create a pull request

.github/workflows/pr.yml
```diff
@@ -8,6 +8,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
```
jpelgrom, line 11: Doesn't it need permission to create inline comments/issues for lint issues that were found?

TimoPtr: Well I tried to test and I didn't manage to get any annotations to be displayed... I tried for yaml I should probably try for ktlint. 
With the current permission the annotations are in the checks.

There is something I don't fully understand on the checks yet. I also looked in the documentation of the permissions and I didn't find things related to annotations.

.github/workflows/prepareNextRelease.yml
TimoPtr: I'm not sure we need this one. The pull_requedt permission from the documentation is for modifying an existing pull request like adding a label https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-pull-requests

I think I'll have to test ...

jpelgrom: On that page it says that `POST /repos/{owner}/{repo}/pulls` is part of `pull_requests`, suggesting to me that we do need it. But I'm also willing to test without, not a major issue if this workflow fails once because of missing permissions.

.github/workflows/pr.yml
TimoPtr: I tested with ktlint and it works 👍🏻

.github/workflows/prepareNextRelease.yml
TimoPtr: It failed https://github.com/home-assistant/android/actions/runs/18279276341/job/52038354748 I'm gonna add the pull_requests permission while fixing missing perms for other workflows. (I created a fake TAG that I'm going to remove for the test)

TimoPtr: I'm going to actually remove the 2 permissions and only keep pull-requests

PR #5856, opened 2025-10-01
Add onboarding set home network

TimoPtr:
## Summary

This screen is the last one of the onboarding flow asking to set the home network in a simplified form (compare to the settings)

This screen is displayed when
- Using HTTP connection to your instance
- Selecting Most secure or allow location sharing

The WIFI SSID is going to be prefilled. The VPN and ETHERNET switch will be shown only is currently in used.

The goal is to having the user validating that the home network is with this WIFI, using VPN/ETHERNET.
Most of the time the user should simply skip this screen.

## Checklist

## Screenshots

image
image

The day we drop the support of HTTP this screen might become obsolete in the flow.

### DIFF
--- a/onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkViewModelTest.kt
+++ b/onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkViewModelTest.kt
@@ -1,7 +1,7 @@
 package io.homeassistant.companion.android.onboarding.sethomenetwork
 
 import app.cash.turbine.test
-import io.homeassistant.companion.android.common.data.network.NetworkCapabilitiesChecker
+import io.homeassistant.companion.android.common.data.network.NetworkHelper
 import io.homeassistant.companion.android.common.data.network.WifiHelper
 import io.homeassistant.companion.android.common.data.servers.ServerManager
 import io.homeassistant.companion.android.database.server.Server
@@ -33,7 +33,7 @@ private const val MOCK_INITIAL_WIFI_SSID_QUOTED = "\"MyHomeWifi\""
 class SetHomeNetworkViewModelTest {
 
     private val serverManager: ServerManager = mockk()
-    private val networkCapabilitiesChecker: NetworkCapabilitiesChecker = mockk(relaxed = true)
+    private val networkHelper: NetworkHelper = mockk(relaxed = true)
     private val wifiHelper: WifiHelper = mockk()
     private lateinit var viewModel: SetHomeNetworkViewModel
 
@@ -47,7 +47,7 @@ class SetHomeNetworkViewModelTest {
         viewModel = SetHomeNetworkViewModel(
             serverId = MOCK_SERVER_ID,
             serverManager = serverManager,
-            networkCapabilitiesChecker = networkCapabilitiesChecker,
+            networkHelper = networkHelper,
             wifiHelper = wifiHelper,
         )
     }
@@ -67,7 +67,7 @@ class SetHomeNetworkViewModelTest {
 
     @Test
     fun `Given view model initialized and ethernet connected, when observing is using ethernet, then emits true`() {
-        every { networkCapabilitiesChecker.hasEthernetConnection() } returns true
+        every { networkHelper.isUsingEthernet() } returns true
         initializeViewModel()
         assertTrue(viewModel.isUsingEthernet.value)
     }
@@ -80,7 +80,7 @@ class SetHomeNetworkViewModelTest {
 
     @Test
     fun `Given view model initialized and VPN connected, when observing is using VPN, then emits true`() = runTest {
-        every { networkCapabilitiesChecker.hasVPNConnection() } returns true
+        every { networkHelper.isUsingVpn() } returns true
         initializeViewModel()
         assertTrue(viewModel.isUsingVpn.value)
     }


onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkViewModel.kt
```diff
@@ -0,0 +1,111 @@
...
+@Suppress("DEPRECATION")
+private fun ConnectivityManager.hasVPNConnection(): Boolean {
+    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+        allNetworks.any {
+            getNetworkCapabilities(it)?.hasTransport(NetworkCapabilities.TRANSPORT_VPN) == true
+        }
+    } else {
+        getNetworkInfo(ConnectivityManager.TYPE_VPN)?.isConnected == true
+    }
+}
```
TimoPtr: We could move this into a shared place but I'm not sure if it useful in another place than here.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkScreen.kt
```diff
@@ -0,0 +1,281 @@
...
+        HASwitch(
+            checked = checked,
+            onCheckedChange = onCheckedChange,
+            modifier = Modifier.testTag(testTag),
+        )
+    }
+}
+
+@Composable
+private fun Hint(modifier: Modifier = Modifier) {
```
jpelgrom, line 254: You've also used a very similar composable in the previous screen about using location to secure the connection - is it time to extract this? That would also allow us to easily use it in a couple of other places.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkScreen.kt
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        if (showEthernet) {
+            SelectableOption(
+                isUsingEthernet,
+                onUsingEthernetChange,
+                text = stringResource(commonR.string.manage_ssids_ethernet),
+                testTag = ETHERNET_TAG,
+            )
+        }
```
jpelgrom, line 166: I agree with putting VPN before ethernet, but let's also change the order of the main settings screen for home network to match then.

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkScreenTest.kt
```diff
@@ -0,0 +1,160 @@
...
+    val composeTestRule = createAndroidComposeRule<HiltComponentActivity>()
+
+    @Before
+    fun setup() {
+        Timber.plant(ConsoleLogTree)
+        ConsoleLogTree.verbose = true
+    }
+
+    @Test
+    fun `Given no vpn and no ethernet when screen is displayed then show no options and handle interactions`() {
```
jpelgrom:
```suggestion
    fun `Given no vpn and no ethernet when screen is displayed then only show Wi-Fi and handle interactions`() {
```

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkScreenTest.kt
```diff
@@ -0,0 +1,160 @@
...
+                onNodeWithContentDescription(stringResource(R.string.clear_text)).assertIsNotDisplayed()
+
+                onNodeWithText(stringResource(R.string.set_home_network_next)).performScrollTo().assertIsDisplayed().performClick()
+                assertTrue(nextClicked)
+            }
+        }
+    }
+
+    @Test
+    fun `Given vpn and no ethernet when screen is displayed then show no options and handle interactions`() {
```
jpelgrom:
```suggestion
    fun `Given vpn and no ethernet when screen is displayed then show Wi-Fi and vpn only and handle interactions`() {
```

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkScreenTest.kt
```diff
@@ -0,0 +1,160 @@
...
+                onNodeWithContentDescription(stringResource(R.string.clear_text)).assertIsNotDisplayed()
+
+                onNodeWithText(stringResource(R.string.set_home_network_next)).performScrollTo().assertIsDisplayed().performClick()
+                assertTrue(nextClicked)
+            }
+        }
+    }
+
+    @Test
+    fun `Given no vpn and ethernet when screen is displayed then show no options and handle interactions`() {
```
jpelgrom:
```suggestion
    fun `Give ethernet and no vpn when screen is displayed then show Wi-Fi and ethernet only and handle interactions`() {
```

It's unclear whether "no" refers only to "vpn" or to "vpn and ethernet"

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkViewModel.kt
```diff
@@ -0,0 +1,111 @@
...
+        viewModelScope.launch {
+            val currentWifiNetwork = currentWifiNetwork.value
+            serverManager.getServer(serverId)?.let { server ->
+                serverManager.updateServer(
+                    server.copy(
+                        connection = server.connection.copy(
+                            internalVpn = isUsingVpn.value,
+                            internalEthernet = isUsingEthernet.value,
+                            // We don't want to add an empty network in the list
+                            internalSsids = currentWifiNetwork.takeIf { it.isNotBlank() }
```
jpelgrom: This may be too niche for onboarding to support, but: #3354

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkViewModel.kt
jpelgrom: Yes - for when we check whether the connection has VPN/ethernet for home network. Why not use the existing NetworkHelper here, which has [similar functions](https://github.com/home-assistant/android/blob/02380ef786a74acfedf4197399aded08f3ea313a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkHelper.kt#L15-L19)?

TimoPtr: I though about moving it there but I was not fully convinced, but since you also pointed it out I'm going to move it and properly test the checker.

TimoPtr: In the end I wont add the test yet otherwise I have to modify how we check the current SDK version and I don't like how it looks. Once we've dropped API 21 and 22 we should be good and writing a test for this class should be easy.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkScreen.kt
TimoPtr: Indeed, I'm going to extract it in a follow up PR 👍🏻

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkScreen.kt
TimoPtr: I didn't even noticed I've changed the order 🤦🏻, I'll update the settings screen

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkViewModel.kt
TimoPtr: Since it is already supported we should too in this screen.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkScreen.kt
TimoPtr: https://github.com/home-assistant/android/issues/5887

TimoPtr:
@jpelgrom I've added the small icon for the VPN and Ethernet like in the settings page. For the OR/AND let see in the review phase what ppl are proposing.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkViewModel.kt
jpelgrom: You've added it to another class than the one I linked... now we have two competing implementations?!

I'm only now noticing that your version checks all networks, while the PR description says "The VPN and ETHERNET switch will be shown only **_is currently in used._**" which is what the existing check I linked does.

TimoPtr: Ok I completely missed this. About all networks I was wondering actually if it was relevant to consider that we could have a connection to a VPN not considered as the active one? I didn't look into the doc.

TimoPtr: > Ok I completely missed this. About all networks I was wondering actually if it was relevant to consider that we could have a connection to a VPN not considered as the active one? I didn't look into the doc.

Like we show the switch of we have a connected VPN or ethernet even if it not active.

jpelgrom: > Like we show the switch [i]f we have a connected VPN or ethernet even if it not active.

But we're not doing that for Wi-Fi, that one is only prefilled if it is the active data connection. (Just tested it to confirm by plugging in ethernet to my phone, Wi-Fi was still connected according to system settings but the app did not get an SSID.)

Showing VPN/ethernet if it is on any connection feels like unnecessary complexity right now, and a chance for confusion in the future for devs using the wrong function. If the user cares that much they will probably seek out the settings later.

TimoPtr: Fair enough 👍🏻 I'm going to remove my changes

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkCapabilitiesChecker.kt
```diff
@@ -18,9 +18,27 @@ import javax.inject.Singleton
 @Singleton
 internal class NetworkCapabilitiesChecker @Inject constructor(private val connectivityManager: ConnectivityManager) {
 
+    /**
```
TimoPtr, line 21: I kept this documentation even if it out of scope of the PR it doesn't hurt.

PR #5861, opened 2025-10-02
Build release APK on PR

TimoPtr:
## Checklist

## Any other notes

For now I decided to not store the APK generated since ppl might install them and we don't want people to install the app from a PR they should do it from the release.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -187,7 +187,22 @@ jobs:
         with:
           cache-encryption-key: ${{ secrets.GRADLE_ENCRYPTION_KEY }}
           mock-google-services: "true"
+      - name: Generate dummy keystore
+        run: |
+          keytool -genkeypair -v \
+          -alias androiddebugkey \
+          -keyalg RSA \
+          -keysize 2048 \
+          -validity 10000 \
+          -keystore /home/runner/work/android/android/app/release_keystore.keystore \
+          -dname "CN=Android CI Debug, O=Android, C=FR" \
+          -storepass android \
+          -keypass android
       - name: Build Release
+        env:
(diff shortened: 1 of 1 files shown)


jpelgrom:
Did you consider waiting on ktlint to complete successfully, to avoid unnecessary builds/resource usage? Other than the flaky tests ktlint is the most common other failure, and a ktlint failure always needs to be resolved before continuing.

TimoPtr:
> Did you consider waiting on ktlint to complete successfully, to avoid unnecessary builds/resource usage? Other than the flaky tests ktlint is the most common other failure, and a ktlint failure always needs to be resolved before continuing.

Indeed we could since today the pipeline duration doesn't seem to be a blocker. I'll update also for pr_build

jpelgrom:
> Indeed we could since today the pipeline duration doesn't seem to be a blocker.

Not a technical blocker, but we should try to keep it fast / parallel if reasonable. Waiting an hour for CI to complete isn't fun (see: recent flaky tests needing lots of retries).

TimoPtr:
> > Indeed we could since today the pipeline duration doesn't seem to be a blocker.
> 
> Not a technical blocker, but we should try to keep it fast / parallel if reasonable. Waiting an hour for CI to complete isn't fun (see: recent flaky tests needing lots of retries).

I used to try to target 30mn max. But TBH I didn't look how we are on the repo since it's mostly the two of us that are heavily using the CI.

TimoPtr:
I'm going to merge this PR and add a new one to modify the flow. I'm going to start ktlint right away instead of waiting for lockfile and yamllint. Doing this won't add any time to the build and will avoid unnecessary jobs if ktlint fails. I'm keeping lint out since it is quite long.

PR #5876, opened 2025-10-06
Make new onboarding available in debug

TimoPtr:
## Summary

This PR makes the changes to replace the `LaunchActivity` with `LauncherActivity` to finally be able to distribute a test APK to some ppl to validate the new flow.

Known issues/limitations
- Deep link `homeassistant://navigate` is not supported
- background and notification permissions are not requested properly in the flow
- Missing error screen (today snackbars, but some errors are going to be moved as full screen actionable errors)
- URL protection `most secure` is not yet enforced
- Changelog dialog is missing
- After deleting the only server connected to going back lead to a loader (also the case in prod)
- Server discovery hide existing server while adding a new server from the settings is not working (we see the server we are connected to if visible)
- Minimal build (without google services) is still asking for location tracking even if there is no location tracking available
- Buttons are not using the right size LARGE instead of MEDIUM
- Documentation for the new flow is not ready (links in the screen are probably wrong, help us finding the right section if it exists already 🙏🏻)
- We have some known leaks while reaching the WebView
- TV onboarding is not yet tested and adjusted

## Checklist

## Any other notes

I renamed the deeplinks to have an action in their name and be more generic.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardApp.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardApp.kt
@@ -5,6 +5,7 @@ import android.content.Intent
 import android.os.Build
 import androidx.activity.result.contract.ActivityResultContract
 import io.homeassistant.companion.android.BuildConfig
+import io.homeassistant.companion.android.launch.intentLauncherOnboarding
 import io.homeassistant.companion.android.launch.intentLauncherWearOnboarding
 
 class OnboardApp : ActivityResultContract<OnboardApp.Input, OnboardApp.Output?>() {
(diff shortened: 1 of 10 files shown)


jpelgrom:
> Minimal build (without google services) is still asking for location tracking even if there is no location tracking available

This probably needs a bit more context than just hiding; using location for home network is still desirable.

> Buttons are not using the right size LARGE instead of MEDIUM

Fixed, other PR merged first.

TimoPtr:
> > Minimal build (without google services) is still asking for location tracking even if there is no location tracking available
> 
> This probably needs a bit more context than just hiding; using location for home network is still desirable.

Indeed, but not necessary for the first testing phase. (I'll do it before the release)


> 
> > Buttons are not using the right size LARGE instead of MEDIUM
> 
> Fixed, other PR merged first.

I updated the description 👍

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardApp.kt
```diff
@@ -84,8 +85,14 @@ class OnboardApp : ActivityResultContract<OnboardApp.Input, OnboardApp.Output?>(
     }
 
     override fun createIntent(context: Context, input: Input): Intent {
-        return if (input.isWatch && BuildConfig.DEBUG) {
-            context.intentLauncherWearOnboarding(input.defaultDeviceName, input.url)
+        return if (BuildConfig.DEBUG) {
+            if (input.isWatch) {
+                context.intentLauncherWearOnboarding(input.defaultDeviceName, input.url)
+            } else {
+                context.intentLauncherOnboarding(input.url)
```
jpelgrom, line 92: Behavior difference with the current OnboardingActivity: the first screen welcoming you to the companion app is no longer skipped if onboarding is started from settings (when adding another server). I understand why (the URL extra check), but do think we should skip that screen when started from settings.

TimoPtr: I'll address this in another PR since it needs to modify the navigation and this PR is the base of 2 others.

PR #5880, opened 2025-10-06
Disable test in WearDnsRequestListenerTest that breaks following tests

TimoPtr:
## Summary

While investigating why we have flaky tests on some very simple tests like the Changelog. I found out a pattern that it always happens when the tests are executed after the one in `WearDnsRequestListenerTest` while digging deeper I remember that `.asTask` was not testable easily since it's a layer on an obfuscated library from Google. After disabling the test that was invoking `.asTask` the tests started to behave normally again. Since the test is pretty dumb I decided to disable it but keep it to remember that it was causing an issue.

### DIFF
--- a/app/src/test/kotlin/io/homeassistant/companion/android/data/wear/WearDnsRequestListenerTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/data/wear/WearDnsRequestListenerTest.kt
@@ -3,17 +3,15 @@ package io.homeassistant.companion.android.data.wear
 import io.homeassistant.companion.android.common.util.WearDataMessages.DnsLookup.PATH_DNS_LOOKUP
 import io.homeassistant.companion.android.common.util.WearDataMessages.DnsLookup.decodeDNSResult
 import io.homeassistant.companion.android.common.util.WearDataMessages.DnsLookup.encodeDNSRequest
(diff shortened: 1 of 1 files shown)


TimoPtr:
credits to @dlevache for helping me

jpelgrom:
Maybe this is my limited experience, but why not mock that function call to do nothing instead?

TimoPtr:
> Maybe this is my limited experience, but why not mock that function call to do nothing instead?

It's an extension, and it's quite annoying to mock. It's always better to try without mocking first.
Since the test is not testing much here just the creation of the task I don't think we should invest time on this.

jpelgrom:
So is there an issue to keep track with Google/Jetbrains?

I can understand not wanting to invest time, but you do keep the other test for returning null. If this part breaks you don't know because the entire test isn't running, it might be returning null 🤷‍♂️

TimoPtr:
> So is there an issue to keep track with Google/Jetbrains?
> 
> I can understand not wanting to invest time, but you do keep the other test for returning null. If this part breaks you don't know because the entire test isn't running, it might be returning null 🤷‍♂️

Probably but it's very hard to identify where it's coming from without knowing the details.

Thanks for pushing me to think about this. I'm going to simply inject a scope instead of overriding the Main dispatcher through the extension. By doing this it should avoid the issue since the impact of whatever they are doing is going to be limited to the test scope created especially for this test.

(I'll push tomorrow)

PR #5895, opened 2025-10-08
Disable lint tasks while building release in PR workflow

TimoPtr:
## Summary

Since we are already running `lint` in another job we should avoid in PR running the lint tasks while building the release. It will reduce the resources used and the time to build.

The hidden goal is to avoid having the job to fail because we are running out of memory.


TimoPtr:
The benefits are not visible compare to a PR where we have a cache of the lint execution. From the logs we can see that lint is not executed anymore.

PR #5897, opened 2025-10-08
Add skip welcome screen option to onboarding

TimoPtr:
## Summary

Skip welcome screen when started from Settings.

This PR also fixes some UI issues:
- back button in Server discovery shouldn't be displayed when the back stack is empty otherwise clicking on it does nothing (like Welcome Screen). This will automatically adjust the moment we only have one big navigation graph.
- discovered server URL was displayed using wrong text align
- missing padding in small screen for discovered server

## Checklist


onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigationTest.kt
```diff
@@ -193,6 +195,24 @@ internal class OnboardingNavigationTest {
...
 
+    @Test
+    fun `Given skipWelcome without urlToOnboard when starting then show ServerDiscovery`() {
+        testNavigation(skipWelcome = true) {
+            assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<ServerDiscoveryRoute>() == true)
+        }
+    }
+
+    @Test
+    fun `Given OnboardingRoute with skipWelcome with urlToOnboard as start when starts then show ServerDiscovery and no back arrow`() {
```
jpelgrom:
```suggestion
    fun `Given skipWelcome and urlToOnboard when starting then show ServerDiscovery and no back arrow`() {
``` 

The original title here feels very convoluted and inconsistent

TimoPtr:
> Merge conflicts everywhere. I've ignored the first three commits when reviewing, I think that's other PRs I've just merged and the cause of conflicts.

Sorry I forgot to warn you that the PR were on top of each other again since it was modifying same code...


> 
> > back button in Server discovery shouldn't be displayed when the back stack is empty otherwise clicking on it does nothing (like Welcome Screen).
> 
> Except this also hides it when you open it from Settings > Companion app > Add server. In that case I expect to see a back/up button, the fact that it is a different activity doesn't matter to me as an average user. (We don't have a back/up button in the current onboarding either but let's ignore that.)



The issue is going to be solved once we have only one graph (so it's going to take time). It is mostly because the nav controller cannot finish the activity so when the stack is empty we cannot do anything except doing back gesture (with predictive back) like on welcome screen.

jpelgrom:
> The issue is going to be solved once we have only one graph (so it's going to take time). It is mostly because the nav controller cannot finish the activity so when the stack is empty we cannot do anything except doing back gesture (with predictive back) like on welcome screen.

That's going to take a long time and we may not ever have one graph for everything. Why is it not possible to just show the icon, but treat it as back instead?

TimoPtr:
> > The issue is going to be solved once we have only one graph (so it's going to take time). It is mostly because the nav controller cannot finish the activity so when the stack is empty we cannot do anything except doing back gesture (with predictive back) like on welcome screen.
> 
> That's going to take a long time and we may not ever have one graph for everything. Why is it not possible to just show the icon, but treat it as back instead?

The current nav library doesn't expose an API other than popbackstack or navigateUP. Both return false when you are on the start destination. We would need to get the activity and call finish manually.

jpelgrom:
> We would need to get the activity and call finish manually.

If we can, why not do that? That creates a more cohesive experience for the user who doesn't know about navigation graphs and activities :)

TimoPtr:
> > We would need to get the activity and call finish manually.
> 
> If we can, why not do that? That creates a more cohesive experience for the user who doesn't know about navigation graphs and activities :)

I brought back the back button and it closes the activity

PR #5898, opened 2025-10-08
Adjust onboarding UI

TimoPtr:
## Summary

Tweak some UI elements
- All the buttons now are taking the max size (fixed an issue in the button wrapper that was allowing the size to go above the limit)
- Reduce space between buttons when two are close to each other
- Remove a spacer to give some room in the SetHomeNetwork (should avoid scrolling)

## Checklist


jpelgrom:
Wide buttons look weird (not really Android-y) but after the size change it all feels weird, so maybe it takes some getting used to.

> Remove a spacer to give some room in the SetHomeNetwork (should avoid scrolling)

There's always someone with a smaller screen so you cannot avoid scrolling.

After looking at all the updated screenshots, wouldn't it make sense to do the same thing for the Wear client certificate section? That one is weirdly centered while other screens like home network with controls move it all to the top.

TimoPtr:
> Wide buttons look weird (not really Android-y) but after the size change it all feels weird, so maybe it takes some getting used to.
> 
> > Remove a spacer to give some room in the SetHomeNetwork (should avoid scrolling)
> 
> There's always someone with a smaller screen so you cannot avoid scrolling.
> 
> After looking at all the updated screenshots, wouldn't it make sense to do the same thing for the Wear client certificate section? That one is weirdly centered while other screens like home network with controls move it all to the top.

image

Something like that? (it's extreme since on tablet)

jpelgrom:
> image
> 
> Something like that? (it's extreme since on tablet)
> 

Yes that looks more consistent and nicer to me. It doesn't shift around after selecting a file because the choose file button will show the file that was selected, right?

TimoPtr:
> > image
> > Something like that? (it's extreme since on tablet)
> 
> Yes that looks more consistent and nicer to me. It doesn't shift around after selecting a file because the choose file button will show the file that was selected, right?


https://github.com/user-attachments/assets/80b4b3cc-f529-4898-9442-111f38077285

PR #5902, opened 2025-10-09
Adjust server discovery flow

TimoPtr:
## Summary

We want to align the logic between iOS and Android. With @bgoncal we agreed on the fact that the modal should not be automatically cancelled. If tow server are discovered during the initial delay of 1.5s then we show the list directly without showing the modal.
When the modal is display we show the loading animation in background until we receive another server in that case the modal stay and we hide the loading animation and we display the list.

We display a message if we don't find a server after 5s. This message is discarded when we find a server.

## Checklist

## Screenshots

Android:

https://github.com/user-attachments/assets/ef9df964-495c-4ec6-973f-2f7201562c8e

For comparison here the iOS version

https://github.com/user-attachments/assets/92fbee6b-22e2-41e2-8d0f-74d7c0827f53

https://github.com/user-attachments/assets/6c36d8ed-3f3c-4e81-b7be-5ee34b0bb340

https://github.com/user-attachments/assets/52d37582-3a98-48fc-8ce0-34d3d42faac2

https://github.com/user-attachments/assets/8324d2e5-d740-4b03-b104-65432a79eff8

## Any other notes

I had to come up with a new flow operator a bit more complex. I'm not sure about the naming. I used AI to generate the tests.

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/util/FlowUtil.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/util/FlowUtil.kt
@@ -4,14 +4,10 @@ import androidx.annotation.VisibleForTesting
 import kotlin.time.Clock
 import kotlin.time.Duration
 import kotlin.time.ExperimentalTime
-import kotlin.time.Instant
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.buffer
 import kotlinx.coroutines.flow.channelFlow
 import kotlinx.coroutines.flow.conflate
-import kotlinx.coroutines.flow.onEach
-import kotlinx.coroutines.flow.onStart
 import kotlinx.coroutines.flow.transform
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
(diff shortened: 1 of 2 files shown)


TimoPtr:
@bgoncal if you can download the APK (or build this branch) to test the behavior it would be nice :)

jpelgrom:
> I had to come up with a new flow operator a bit more complex.

`delayFirst` is/was only used for this onboarding flow, so I think it makes sense to remove that one for now?

For all tests: "then should emits" should be "then emits" or "then should emit".

TimoPtr:
> > I had to come up with a new flow operator a bit more complex.
> 
> `delayFirst` is/was only used for this onboarding flow, so I think it makes sense to remove that one for now?

It could, it was breaking my heart to do it, since it was a bit painful to get right to be honest 🤣. I had some hope to reuse it somewhere else, one day.

> 
> For all tests: "then should emits" should be "then emits" or "then should emit".

jpelgrom:
> It could, it was breaking my heart to do it, since it was a bit painful to get right to be honest 🤣. I had some hope to reuse it somewhere else, one day.

It's in git history! I'd remove dead code if there is no planned use right now.

TimoPtr:
> > It could, it was breaking my heart to do it, since it was a bit painful to get right to be honest 🤣. I had some hope to reuse it somewhere else, one day.
> 
> It's in git history! I'd remove dead code if there is no planned use right now.

Yes I'm going to 👍🏻. I'm also going to keep it in a Gist otherwise I'm going to forget about it.

https://gist.github.com/TimoPtr/876163eca43d4c89ae10772b19b8cee9

TimoPtr:
@bgoncal tested the behavior and it works like we agreed on.

PR #5912, opened 2025-10-10
Make sure to clear the back stack after NameYourDevice is done

TimoPtr:
## Summary

While testing the new onboading @klaasnicolaas found out that we can come back to name your device when adding a server from the settings. It put the apps in a weird state because the `auth_token` is already consumed so we cannot do anything on this screen and it leads to this error 

image

This is due to the recent changes that skip the welcome screen, we were poping up up to WelcomeScreen but since we've skipped it the framework is not able to clear the back stack properly. This PR ensure that we clear up to the startDestination instead.

## Checklist

## Screenshots

https://github.com/user-attachments/assets/b95a1d9f-4f24-4066-acd1-56cb5e6c9d94

### DIFF
--- a/onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigationTest.kt
+++ b/onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigationTest.kt
@@ -370,10 +370,23 @@ internal class OnboardingNavigationTest {
     }
 
     @Test
+    fun `Given device named and skip welcome with url when pressing next then show LocalFirst then goes back stop the app`() {
+        localFirstTest(true, "http://ha.local")
+    }
+
+    @Test
+    fun `Given device named and skip welcome without url when pressing next then show LocalFirst then goes back stop the app`() {
+        localFirstTest(true, null)
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigationTest.kt
```diff
@@ -369,11 +369,24 @@ internal class OnboardingNavigationTest {
...
+        localFirstTest(true, null)
+    }
+
     @Test
     fun `Given device named when pressing next then show LocalFirst then goes back stop the app`() {
-        val instanceUrl = "http://ha.local"
-        testNavigation {
-            navController.navigateToNameYourDevice(instanceUrl, "code")
+        localFirstTest(false, null)
+    }
```
TimoPtr, line 385: @jpelgrom since this class is using JUnit 4 I can't use ParametrizedTest 👎🏻

jpelgrom: Thanks for the explanation. More specific test titles are good to have here anyway.

PR #5919, opened 2025-10-12
Fix `ObsoleteSdkInt` lint warnings in `wear` module

loganrosen:
## Summary
Fixing all of the `ObsoleteSdkInt` lint warnings in the `wear` module and updating the lint baseline accordingly, as part of #5221.

## Checklist

### DIFF
--- a/wear/src/main/kotlin/io/homeassistant/companion/android/sensors/SensorReceiver.kt
+++ b/wear/src/main/kotlin/io/homeassistant/companion/android/sensors/SensorReceiver.kt
@@ -9,7 +9,6 @@ import android.content.Intent
 import android.media.AudioManager
 import android.net.wifi.WifiManager
 import android.nfc.NfcAdapter
-import android.os.Build
 import android.os.PowerManager
 import androidx.core.app.TaskStackBuilder
 import androidx.core.net.toUri


wear/src/main/kotlin/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
```diff
@@ -154,10 +154,6 @@ class HealthServicesSensorManager : SensorManager {
         return arrayOf(Manifest.permission.ACTIVITY_RECOGNITION)
     }
 
-    override fun hasSensor(context: Context): Boolean {
-        return Build.VERSION.SDK_INT >= Build.VERSION_CODES.R
-    }
```
TimoPtr, line 159: I think it would make actual more sense to keep the version check here and always add the SensorManager. You will need to remove the @ RequiresApi at the top of the file.

@jpelgrom wdyt? I think it keeps the contextual information closer to the implementation.
In the future I would probably like to inject all the managers instead of manually adding them to a list.

jpelgrom: The annotation on the class is not enough, this function is always called so should be kept.

TimoPtr: I want to keep this and remove the class annotation like we did with the car recently.

loganrosen: @jpelgrom Is that true? It's only conditionally added [here](https://github.com/home-assistant/android/blob/87c461a705a321cd5c05d03b9328f25896de3a12/wear/src/main/kotlin/io/homeassistant/companion/android/sensors/SensorReceiver.kt#L86).

@TimoPtr Sure, I can make that change.

TimoPtr: Thanks for the changes 👍🏻. @jpelgrom from my point of view this is how it should be done, but lemme know if you think otherwise.

jpelgrom: No, agreed this is what is preferred.

PR #5920, opened 2025-10-12
Remove usage of `runBlocking` from `SettingsActivity`

loganrosen:
## Summary
As part of #5688, removing usage of `runBlocking` from `SettingsActivity`.

The `setAppActive` and `isAppLocked` logic has been moved into a new `SettingsActivityViewModel`, using `viewModelScope` instead of `lifecycleScope`. This ensures the app lock state is persisted even if the Activity lifecycle scope is cancelled (e.g. during `onPause` before `onDestroy`). `SettingsFragment` now accesses the shared ViewModel via `activityViewModels()` instead of casting to the Activity.

## Checklist


app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -545,7 +545,11 @@ class SettingsFragment(private val presenter: SettingsPresenter, private val lan
 
     private fun onServerLockResult(result: Int): Boolean {
         if (result == Authenticator.SUCCESS && serverAuth != null) {
-            (activity as? SettingsActivity)?.setAppActive(serverAuth, true)
+            (activity as? SettingsActivity)?.let { settingsActivity ->
+                lifecycleScope.launch {
+                    settingsActivity.setAppActive(serverAuth, true)
+                }
+            }
             parentFragmentManager.commit {
```
TimoPtr: @jpelgrom @dshokouhi could you test if you see any issue on this change? 

I did test on Android 9 old device, that is quite slow with 2 servers and 1 lock and it seems to be ok.

jpelgrom: The fragment's comment suggests a possible issue when toggling it on/off:

https://github.com/home-assistant/android/blob/e40092508e8f9b203cc791e535ceada68ad0e85e/app/src/main/kotlin/io/homeassistant/companion/android/settings/server/ServerSettingsFragment.kt#L340-L341

Now the code runs async so it might end up locking the app? Is that what you tested or just app lock on the main settings screen?

TimoPtr: I simply tested app lock in both dashboard and settings

jpelgrom:
After thinking about it a bit more: by setting the active state in the `lifecycleScope`, we risk that the app is not set as inactive before the activity/fragment/app is quit, as the scope may be cancelled. That should never happen as it defeats the purpose of an app lock.

(This may also apply in other places, not only here)

loganrosen:
> After thinking about it a bit more: by setting the active state in the `lifecycleScope`, we risk that the app is not set as inactive before the activity/fragment/app is quit, as the scope may be cancelled. That should never happen as it defeats the purpose of an app lock.
> 
> (This may also apply in other places, not only here)

@jpelgrom Do we need to use an application-level scope here?

jpelgrom:
> @jpelgrom Do we need to use an application-level scope here?

I missed that this PR was waiting for a response from me, sorry.

I'm tempted to say yes, but would like another opinion. Maybe I'm making this too big, and it would increase the scope of the PR. @TimoPtr what do you think?

TimoPtr:
> > @jpelgrom Do we need to use an application-level scope here?
> 
> I missed that this PR was waiting for a response from me, sorry.
> 
> I'm tempted to say yes, but would like another opinion. Maybe I'm making this too big, and it would increase the scope of the PR. @TimoPtr what do you think?

In general I'm not a big fan of application scope. It could be in a viewModel.

Now that I think a bit more about it, we could still have race condition if for some reason the localStorage is slow. We might need to make the `IntegrationRepositoryImpl.appActive` thread safe. 

It could be made in a first PR before this one. Also I would like to have tests for such changes.

loganrosen:
Addressed the review feedback:

- Moved `setAppActive` and `isAppLocked` into a new `SettingsActivityViewModel`, using `viewModelScope` instead of `lifecycleScope` so the operation outlives the Activity lifecycle. The fragment now uses `activityViewModels()` to access it directly instead of casting to the Activity.
- Added unit tests for the ViewModel (both methods, error handling, null server IDs).

Thread-safety of `IntegrationRepositoryImpl.appActive` is not addressed here per @TimoPtr's suggestion to handle that in a separate PR.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModel.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    suspend fun isAppLocked(serverId: Int?): Boolean {
+        return serverManager.getServer(serverId ?: ServerManager.SERVER_ID_ACTIVE)?.let {
+            try {
+                serverManager.integrationRepository(it.id).isAppLocked()
+            } catch (e: IllegalArgumentException) {
+                Timber.w(e, "Cannot determine app locked state")
+                false
+            }
+        } ?: false
```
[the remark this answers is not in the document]
loganrosen: Keeping the `getServer()` null check — a missing server is not an error (can happen when a server is removed), so a silent no-op is the right behavior. Removing it would use exceptions for control flow.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModel.kt
```diff
@@ -0,0 +1,44 @@
...
+import androidx.lifecycle.viewModelScope
+import dagger.hilt.android.lifecycle.HiltViewModel
+import io.homeassistant.companion.android.common.data.servers.ServerManager
+import javax.inject.Inject
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import timber.log.Timber
+
+@HiltViewModel
+class SettingsActivityViewModel @Inject constructor(private val serverManager: ServerManager) : ViewModel() {
```
TimoPtr: I'm not a fan of the name of the ViewModel, especially because we already have a SettingViewModel.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModel.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Set the app active state for a specific server. Uses [viewModelScope] to ensure the
+     * operation completes even if the calling Activity/Fragment lifecycle scope is cancelled,
+     * which is critical for the app lock feature.
+     */
+    fun setAppActive(serverId: Int?, active: Boolean) {
+        viewModelScope.launch(Dispatchers.IO) {
```
TimoPtr:
```suggestion
        viewModelScope.launch(Dispatchers.Default) {
```
We don't need to set it to IO, the setAppActive is main safe already.

app/src/test/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModelTest.kt
```diff
@@ -0,0 +1,138 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.RegisterExtension
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class SettingsActivityViewModelTest {
+
+    companion object {
+        @JvmField
+        @RegisterExtension
+        val mainDispatcherExtension = MainDispatcherJUnit5Extension()
```
TimoPtr: This should be using in @ExtendWith annotation not in the companion object

app/src/test/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModelTest.kt
```diff
@@ -0,0 +1,138 @@
...
+    fun setup() {
+        coEvery { serverManager.getServer(any<Int>()) } returns mockk<Server>(relaxed = true) {
+            every { id } returns 1
+        }
+        coEvery { serverManager.integrationRepository(any()) } returns integrationRepository
+        viewModel = SettingsActivityViewModel(serverManager)
+    }
+
+    @Test
+    fun `setAppActive true calls integrationRepository setAppActive with true`() = runTest {
```
TimoPtr: Rewrite the test name using Given When Then

app/src/test/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModelTest.kt
```diff
@@ -0,0 +1,138 @@
...
+        coJustRun { integrationRepository.setAppActive(any()) }
+
+        viewModel.setAppActive(1, true)
+        advanceUntilIdle()
+
+        coVerify { integrationRepository.setAppActive(true) }
+    }
+
+    @Test
+    fun `setAppActive false calls integrationRepository setAppActive with false`() = runTest {
```
TimoPtr: You can use @ParametrizedTest to reduce the numbers of tests.

app/src/test/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModelTest.kt
```diff
@@ -0,0 +1,138 @@
...
+    @Test
+    fun `setAppActive handles IllegalArgumentException gracefully`() = runTest {
+        coEvery { serverManager.integrationRepository(any()) } throws IllegalArgumentException("test")
+
+        viewModel.setAppActive(1, true)
+        advanceUntilIdle()
+
+        // Should not crash
+    }
+
```
TimoPtr: You can use @Nested to group tests that are for the same method.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModel.kt
loganrosen: Renamed to `AppLockViewModel`.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModel.kt
loganrosen: Done.

app/src/test/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModelTest.kt
loganrosen: Done.

app/src/test/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModelTest.kt
loganrosen: Done.

app/src/test/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModelTest.kt
loganrosen: Done — combined the true/false variants for both `setAppActive` and `isAppLocked` with `@ValueSource(booleans = [true, false])`.

app/src/test/kotlin/io/homeassistant/companion/android/settings/SettingsActivityViewModelTest.kt
loganrosen: Done — grouped into `SetAppActive` and `IsAppLocked` inner classes.

jpelgrom:
This became more elegant than I expected. Is the intention to use this in other activities with app lock checks as well, as the ViewModel is very generic?

TimoPtr:
> This became more elegant than I expected. Is the intention to use this in other activities with app lock checks as well, as the ViewModel is very generic?

We could but in another PR, merging this one.

PR #5926, opened 2025-10-13
Add privacy hint on sensors details view

TimoPtr:
## Summary

Add a small hint as suggested by @jpelgrom to inform how the sensor data of the user are going to be used.

## Checklist

## Screenshots

image
image

## Any other notes

I didn't know how to store the information about the user discarding the hint. I used the Session storage but I'm not a fan at all. I'm under the impression that we are lacking a storage for application specific user pref.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
@@ -18,7 +18,7 @@ import androidx.lifecycle.viewModelScope
 import dagger.hilt.android.lifecycle.HiltViewModel
 import io.homeassistant.companion.android.common.R as commonR
 import io.homeassistant.companion.android.common.bluetooth.BluetoothUtils
-import io.homeassistant.companion.android.common.data.LocalStorage
+import io.homeassistant.companion.android.common.data.prefs.PrefsRepository
 import io.homeassistant.companion.android.common.data.servers.ServerManager
 import io.homeassistant.companion.android.common.sensors.BluetoothSensorManager
 import io.homeassistant.companion.android.common.sensors.NetworkSensorManager
@@ -31,7 +31,6 @@ import io.homeassistant.companion.android.database.sensor.SensorWithAttributes
 import io.homeassistant.companion.android.database.sensor.toSensorsWithAttributes
 import io.homeassistant.companion.android.database.settings.SensorUpdateFrequencySetting
 import io.homeassistant.companion.android.database.settings.SettingsDao
-import io.homeassistant.companion.android.di.qualifiers.NamedSessionStorage
 import io.homeassistant.companion.android.sensors.LastAppSensorManager
 import io.homeassistant.companion.android.sensors.SensorReceiver
 import javax.inject.Inject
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
```diff
@@ -283,6 +296,14 @@ class SensorDetailViewModel @Inject constructor(
         sensorSettingsDialog = null
     }
 
+    fun discardShowPrivacyHint() {
+        _showPrivacyHint.update { false }
+        viewModelScope.launch {
+            // 0 is used for storing app level settings
+            localStorage.putBoolean(PREF_SHOW_PRIVACY_HINT, false)
```
jpelgrom: > I didn't know how to store the information about the user discarding the hint. I used the Session storage but I'm not a fan at all. I'm under the impression that we are lacking a storage for application specific user pref.

Generally: we don't edit prefs directly in UI screens or associated ViewModels/Fragments/Activities, but use repositories instead. The class using the pref doesn't know whether it is shared preferences or a database or ...

`PrefsRepository` exists exactly for this purpose: global/application level preferences.

common/src/main/res/values/strings.xml
```diff
@@ -1166,6 +1166,7 @@
     <string name="sensor_update_type_info_worker_setting">You can change how often this sensor is updated by using the \'%1$s\' setting.</string>
     <string name="sensor_update_type_info_location">For more information about location updates, please check the documentation.</string>
     <string name="sensor_update_type_info_custom">This sensor updates on a non-standard schedule, please check the documentation.</string>
+    <string name="sensor_privacy">You have full control over your data. When enabled, sensor data is only sent to your own Home Assistant server. It will never be shared with the Home\u00A0Assistant project or third parties.</string>
```
jpelgrom: No non-breaking space for the first HA?

TimoPtr: Good catch.

app/src/main/kotlin/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
TimoPtr: I'm going to change thanks for helping, I struggled a lot on where to put it that I went the easy path. 
I think the `NamedThemesStorage` is badly named.

I'm fully align that we shouldn't work on the data layer directly from the VM, it goes also for the DAO IMO too.

jpelgrom: > I think the `NamedThemesStorage` is badly named.

Originally it was for themes but the file name doesn't change... The repository does have a logical name?
 
> I'm fully align that we shouldn't work on the data layer directly from the VM, it goes also for the DAO IMO too. 

Let's not tackle that here.

TimoPtr: > > I think the `NamedThemesStorage` is badly named.
> 
> Originally it was for themes but the file name doesn't change... The repository does have a logical name?

Yes indeed. I guess we could adjust while migrating to DataStore.

> 
> > I'm fully align that we shouldn't work on the data layer directly from the VM, it goes also for the DAO IMO too.
> 
> Let's not tackle that here.

Indeed.

Issue #5931, opened 2025-10-14
Bluetooth permission should only be requested when needed for location sensor


TimoPtr:
### Description

Today we are asking for Bluetooth permission when we want to enable the LocationSensors. Without this we are disabling the sensor.

Bluetooth in location tracking is only used for high accuracy tracking with some condition like being connected to a given device.

We should be able to enable the location without asking for Bluetooth permission. This would make one less popup during the onboarding of a user. Also the location should continue to work even if the user denied Bluetooth permission.

TimoPtr:
## Idea

### Having a permission at a sensor settings level

We could attach a permission (or a set of permission)  to a settings, so that we check the permission when we check the settings. We can then ask for permission when the user actually wants to turn on this setting.

jpelgrom:
> ## Idea

Instead of complicating sensors with permissions per setting (which are dynamically created), my suggestion would be to break location tracking out from the sensor interface. The three main 'sensors' are not actually sensors and already behave differently.

dshokouhi:
I do recall that we did add location to sensor manager at the time because location tracking was a bit unreliable then due to bugs (both in app and play services). It was nice to have work manager there because it was able to periodically do checks like if we are still getting updates as expected.  Which was for another bug fix as sometimes play services stops sending us updates. 

I do like the permission at sensor setting level but I cant think of any other sensor that would need a permission like this one for the setting.

jpelgrom:
We could still use WorkManager for location tracking setup/periodic checks, I'm just suggesting that we should not treat it like a sensor when this is making things needlessly complicated for actual sensors. (And I think it also is confusing to normal users to see it as a sensor while it does not behave like one.)

PR #5909, opened 2025-10-10
[DONOTMERGE] New Onboarding feedbacks

TimoPtr:
> [!CAUTION]
> This PR is not meant to be merged.
> It serves as a distribution point for a trusted debug APK, available under the Checks tab or via the following link: [debug APK][APK_LINK].

## Summary

:tada: Introducing the first version of the new onboarding experience!

After a few months of work, we’ve reached an exciting milestone the first version of our new onboarding flow is ready for you to try and share feedback.

:warning: **Heads up:** this is an early, unreleased build. We’ll list known issues and unsupported items below :point_down: please focus your testing beyond those areas for now.
On Android, this release is a big step forward: onboarding now runs on a solid new architecture, with tests in place :muscle: (though that also means possible regressions compared to the current app).
We’d love for you to install the app, complete the onboarding, and tell us what you think! Our goal is to make onboarding more modern, polished, and welcoming this is just the beginning.
:pray: Thanks for testing and helping us improve!

**Known issues/limitations**
- Splashscreen is not the right one
- background and notification permissions are not requested properly in the flow (they are going to be asked while on the dashboard at the very end)
- URL protection `most secure` is not yet enforced
- After deleting the only server connected to going back lead to a loader (also the case in prod)
- Minimal build (without google services) is still asking for location tracking even if there is no location tracking available
- Documentation for the new flow is not ready (links in the screen are probably wrong, help us finding the right section if it exists already :pray_tone2:)
- We have some known leaks while reaching the WebView
- TV onboarding is not yet tested and adjusted
- Notification are not working with Firebase in Debug

## How to install the app

Download and install the provided [debug APK][APK_LINK] on your Android device or emulator.

image

> [!IMPORTANT]  
> This is the debug version of the app (red icon).
> It will not replace your existing Home Assistant app or erase your settings.

## How to give feedbacks

We appreciate your time and input! Please help us:
    👍 Leave a reaction on this PR if you enjoyed testing it.
    🐞 Found an issue? Add a comment describing steps to reproduce.
    💡 Have an idea for improvement? Share your thoughts in a comment.

Let’s keep discussions constructive and kind we truly value your help.

> [!NOTE]
> We'll review all feedback, though we may not be able to apply every suggestion right away.

## Screenshots

Screenshots of all screen variants are available in the PR files for review.

## Any other notes

The CI is expected to fail (intentional).
I'm going to update the PR (rebase) with potential fixes when they are merged to main.

[...]

PR #5932, opened 2025-10-14
Add Bluetooth permission to LocationPermissionState

TimoPtr:
## Summary

Thanks to @dshokouhi we found out that the new onboarding was not enabling the location sensor https://github.com/home-assistant/android/pull/5909#issuecomment-3398416248. The issue was that I removed the Bluetooth permission from being requested in the onboarding. I didn't understand at first that it would have for impact to disable the Location sensors entirely because the manager list Bluetooth permission as a hard requirement for the sensor to be enabled.

This should not be a hard requirements but it requires a significant change in the `LocationSensorManager` which is not a priority for now. I decided to create a follow up task https://github.com/home-assistant/android/issues/5931 and make a quick fix in this PR by adding back the permission to the list of permission requested in the onboarding.

## Checklist


TimoPtr:
> This works, but I'm not the biggest fan of hardcoding these permissions here 🤔 At the same time, this should be picked up when doing full/minimal handling as we need a different permission level then anyway.

I don't liked it either, the main reason behind is that the sensor are not known in the onboarding module so I cannot use the permissions defined there. Also this weird logic of asking for background after make me think that we might want something specific for the UI.

PR #5933, opened 2025-10-14
Display LocalFirstScreen only when necessary

TimoPtr:
## Summary

We want to avoid displaying the LocalFirstScreen when it doesn't make sense, like when we are sure that the URL used is publicly accessible. This PR make sure of that by displaying the screen only when it makes sense.

It address the feedback from @dshokouhi https://github.com/home-assistant/android/pull/5909#issuecomment-3398335900

## Checklist

## Any other notes

It needs to be merged after https://github.com/home-assistant/android/pull/5912 to avoid conflicts.

### DIFF
--- a/onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
+++ b/onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
@@ -128,7 +128,7 @@ internal fun NavGraphBuilder.onboarding(
                         inclusive = true
                     }
                 }
-                if (hasPlainTextAccess || !isPubliclyAccessible) {
+                if (!isPubliclyAccessible) {
                     navController.navigateToLocalFirst(
                         serverId = serverId,
                         hasPlainTextAccess = hasPlainTextAccess,


common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt
```diff
@@ -96,3 +102,50 @@ object UrlUtil {
...
+    if (localTlds.any { fqdn.endsWith(it, ignoreCase = true) }) {
+        return false
+    }
+
+    // Resolve and check IP
+    return try {
+        val addresses = withContext(Dispatchers.IO) {
+            InetAddress.getAllByName(fqdn)
+        }
+        addresses.none { it.isPrivateOrLocal() }
```
jpelgrom, line 137: What about people using a public FQDN, but running an internal DNS server so it resolves to the internal IP when at home (with a short TTL so it is refreshed frequently)? I've seen that more than once. Maybe a simple check based on TLD is enough.

common/src/test/kotlin/io/homeassistant/companion/android/util/UrlUtilTest.kt
```diff
@@ -0,0 +1,59 @@
...
+        // Using a well-known public domain that should resolve to public IPs
+        val url = URL("https://www.home-assistant.io")
+        assertTrue(url.usesPublicFqdn())
+    }
+
+    @Test
+    fun `Given URL with Google DNS when checking usesPublicFqdn then returns true`() = runTest {
+        // Google's public DNS IP
+        val url = URL("http://8.8.8.8:80")
+        assertTrue(url.usesPublicFqdn())
```
jpelgrom: An IP address is not a FQDN...

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
```diff
@@ -39,8 +41,11 @@ internal sealed interface NameYourDeviceNavigationEvent {
      *
      * @param serverId The ID of the server for which the device name was saved.
      * @param hasPlainTextAccess Boolean that defines if the server has a plain text URL.
+     * @param isRemotelyAccessible Boolean that defines if the server is publicly accessible.
      */
-    data class DeviceNameSaved(val serverId: Int, val hasPlainTextAccess: Boolean) : NameYourDeviceNavigationEvent
+    data class DeviceNameSaved(val serverId: Int, val hasPlainTextAccess: Boolean, val isRemotelyAccessible: Boolean) :
```
jpelgrom: Your description already tells us that you actually mean public ;)

```suggestion
    data class DeviceNameSaved(val serverId: Int, val hasPlainTextAccess: Boolean, val isPubliclyAccessible: Boolean) :
```

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
```diff
@@ -90,10 +95,12 @@ internal class NameYourDeviceViewModel @VisibleForTesting constructor(
             try {
                 _isSavingFlow.emit(true)
                 val serverId = addServer()
+                val url = route.url
                 _navigationEventsFlow.emit(
                     NameYourDeviceNavigationEvent.DeviceNameSaved(
                         serverId,
-                        route.url.startsWith("http://"),
+                        hasPlainTextAccess = url.startsWith("http://"),
+                        isRemotelyAccessible = runCatching { URL(url).usesPublicFqdn() }.getOrDefault(false),
```
jpelgrom: We don't really care about remote/public access if it is plaintext, so should skip the check in that case to prevent a network request.

jpelgrom (commented):
Not sure if you wanted to do this in another PR, but to fully address the feedback, the messaging should also be changed for plaintext + publicly accessible (like `http://home.example.com`, or simple port forwarding without SSL).

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt
TimoPtr: Well I think it makes sense when the call is actually made, for such setup ppl "know" what they are doing and most probably they have HTTPS.

But some ppl have static entry in their local DNS for a domain that is not registered externally.

common/src/test/kotlin/io/homeassistant/companion/android/util/UrlUtilTest.kt
TimoPtr: Indeed the test is actually not only about the FQDN but is it publicly addressable. `isPubliclyAddressable` (or Accessible) might be a better name?

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
TimoPtr: I get your point but to be fair, I'm pretty sure this won't do a network call since the domain is already resolved before for the connection. I prefer to keep two distinct boolean in case the logic behind changes I don't want to be mislead thinking that `isRemotelyAccessible` is accurate when it's not.

common/src/test/kotlin/io/homeassistant/companion/android/util/UrlUtilTest.kt
TimoPtr: I changed the name to `isPubliclyAccessible`

TimoPtr:
> Not sure if you wanted to do this in another PR, but to fully address the feedback, the messaging should also be changed for plaintext + publicly accessible (like `http://home.example.com`, or simple port forwarding without SSL).

@bgoncal suggested to drop the last sentence in this screen
image

If the users agrees to share his location in any case we are going to choose `Most secure` for him and if he doesn't we are going to show the LocationForSecureScreen where we explicitly say that the connection is unencrypted. 

I'm not sure we need to add a mention on this screen wdyt? (both of you @bgoncal and @jpelgrom)

bgoncal:
I don't think it's needed. Keep it simple

bgoncal:
I created a PR for iOS similar to this but I will leave DNS resolution out of the equation until I get feedback that it's needed https://github.com/home-assistant/iOS/pull/3900

jpelgrom:
> > Not sure if you wanted to do this in another PR, but to fully address the feedback, the messaging should also be changed for plaintext + publicly accessible (like `http://home.example.com`, or simple port forwarding without SSL).
> 
> @bgoncal suggested to drop the last sentence in this screen image

But it still mentions "only accessible on your home network", "set up remoe access later", while it is remotely accessible. This is not going to be fixed by removing/adding one sentence.

@dshokouhi since it is based on your feedback, this is what I find confusing but maybe you have more/other comments

bgoncal:
That will be "handled" by the check that @TimoPtr made to detect if it's internal or external URL.

The endgame is to always receive from core what is the internal and what is the remote for that server, but we don't have that yet.

jpelgrom:
> That will be "handled" by the check that @TimoPtr made to detect if it's internal or external URL.

Which check are you referring to? I don't recognize anything like that in the code changes or are able to get it to behave differently, and this PR does not depend on another branch.

bgoncal:
Perhaps I misunderstood, Timo can confirm tomorrow but in my mind the check for if the URL is publicly available was to determine during onboarding if we are onboarding a local or remote URL.

In iOS specifically this means that we put the URL in the correct "place" in the configuration, as you know in iOS we call internal/external URL.

Given the above we could determine if we need to display this disclaimer screen or not.

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt
dshokouhi: were probably still going to have cases where users who feel shouldnt see this screen still do

jpelgrom:
> Perhaps I misunderstood, Timo can confirm tomorrow but in my mind the check for if the URL is publicly available was to determine during onboarding if we are onboarding a local or remote URL.
> 
> In iOS specifically this means that we put the URL in the correct "place" in the configuration, as you know in iOS we call internal/external URL.

That may be the goal, but is not what the PR is doing. There is also no such thing as only having an "internal" URL in the Android app architecture so that would require bigger changes which are definitely out of scope of this PR.

> Given the above we could determine if we need to display this disclaimer screen or not.

... but we're not there yet, the reason mentioned is a specific comment from @dshokouhi and I agree that it is somewhat confusing to see this screen when using something like `http(s)://home.example.com` :)

dshokouhi:
Had a chance to test the PR, at least for my setup I no longer see the screen which is what I would expect. Generally speaking I dont think we will be handle to every check to ensure this doesnt show up for every user and every configuration. Handling the majority probably makes the most sense.

bgoncal:
@jpelgrom  In that case I'll leave for @TimoPtr to take over the discussion, if it can't be determined if remote access exist (in a best effort manner), then this screen shouldn't be shown at all in Android.

But the way I see it, a quick fix is:

- Onboarding coming from mdns discovery? No cloud URL fetch from core? Show the screen.

- Onboarding coming from manual entry URL and detected that is a remote URL? (doesn't need to be 100% accurate, covering most cases is good enough) Do not show the screen

Not saying this PR does that, that's just my opinion on what should happen.

TimoPtr:
I think all this confusion can be lift off by simplifying the condition. Today we show this screen if it plaintext OR if it is publicly accessible. We do have another screen later to notify the user if the connection is not encrypted (LocationForSecureConnection and later the "blocking" screen). I propose to simply the condition to show this screen to only when we know it is a non public address. 

Yes it means that ppl with a complex setup like changing the DNS entry using a short TTL are going to see this screen. But technically this screen is true the user is onboarding on a private address, I don't think it is going to bother the user to see this screen. Mostly because he knows his setup. It is going to be an issue for a secondary user, but I would assume that in that case the secondary user is going to ask the admin to help him setup remote access.

WDYT @bgoncal @jpelgrom @dshokouhi ?

bgoncal:
Looks good to me, my main concern is the average user.

jpelgrom:
Agreed we should focus on average. Simplifying it to only show if it's a non-public address works for me to as a solution to avoid confusion about "only accessible on your home network" 👍

common/src/test/kotlin/io/homeassistant/companion/android/util/UrlUtilTest.kt
```diff
@@ -0,0 +1,59 @@
...
+        ],
+    )
+    fun `Given local URL when checking isPubliclyAccessible then returns false`(url: String) = runTest {
+        val url = URL(url)
+        assertFalse(url.isPubliclyAccessible())
+    }
+
+    @Test
+    fun `Given unresolved TLD when checking isPubliclyAccessible then returns false`() = runTest {
+        val url = URL("http://this-domain-definitely-does-not-exist-12345.com:8123")
```
jpelgrom, line 42: If someone wants to troll us they'll buy the domain

PR #5943, opened 2025-10-16
Add HADetails composable

TimoPtr:
## Summary

Add a simple version of HA-details from Home Assistant design system. The implementation is not yet complete, since the component is still under construction, but this current implementation is functional.

## Checklist

## Screenshots

image
image

## Any other notes

The design system got updated and we need to change the background color to a darker color that's why today we don't see the background of the hadetails.


jpelgrom:
> The design system got updated and we need to change the background color to a darker color that's why today we don't see the background of the hadetails.

(main/surface background changed, I'm assuming)

If the background color is that important, then it should be changed first and the screenshot test should make sure that main vs. details background is visible (add some padding?).

TimoPtr:
> > The design system got updated and we need to change the background color to a darker color that's why today we don't see the background of the hadetails.
> 
> (main/surface background changed, I'm assuming)
> 
> If the background color is that important, then it should be changed first and the screenshot test should make sure that main vs. details background is visible (add some padding?).


For now it's going to stay like this, we'll go back to this in 2 weeks. The current design should work.

Issue #5935, opened 2025-10-14
Support multiple languages in F-droid


TimoPtr:
Store the translations from lokalise in our release in a zip. Modify build pipeline from F-droid to pull the zip and extract it before building the app.

Check https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/io.homeassistant.companion.android.minimal.yml

TimoPtr:
We could store the translation in a zip in our release and then do something like
```bash
   - export VERSION_NO_FLAVOR="${$$VERSION$$%-minimal}"
   - export RELEASE_URL="https://github.com/home-assistant/android/releases/download/${VERSION_NO_FLAVOR}/translati
  on.zip"
    - echo "Downloading translations from: $RELEASE_URL"
    - curl -L -f "$RELEASE_URL" -o translation.zip
    - unzip -o translation.zip -d app/src/main/res/
```
in the `prebuild` section assuming we have `curl`

TimoPtr:
https://gitlab.com/fdroid/fdroiddata/-/merge_requests/28278 Once this Merge Request is merged the and the PR #5946 the translations should be available on F-droid.

licaon-kter:
why aren't these in the repo here?  https://github.com/home-assistant/android/tree/main/app/src/main/res

jpelgrom:
Because translating is handled using Lokalise, a web UI lowering barrier to entry and with some language-specific features which are much nicer than XML files (see [contributor docs](https://developers.home-assistant.io/docs/translations/)). Every release dynamically pulls the latest strings from Lokalise.

licaon-kter:
And it can't be integrated here in git like Weblate, Poedit, Crowdin or Transifex portals do?

dshokouhi:
> And it can't be integrated here in git like Weblate, Poedit, Crowdin or Transifex portals do?

we dont store them in github to prevent unwanted edits, that is the reason they are only pulled when we build a release. We only want users to edit them in Lokalise and if they see it in git they will submit PRs to update which is what we want to avoid.

licaon-kter:
You can reject MRs here that edit them, just like other MRs you don't like :)

dshokouhi:
> You can reject MRs here that edit them, just like other MRs you don't like :)

That does not align with the HA project as a whole.

TimoPtr:
Everything has been merged and users of F-droid should have translations now.

PR #5946, opened 2025-10-16
Store translations in github release for F-droid

TimoPtr:
## Summary

In order to offer all the translations we have on the minimal version distributed on F-droid we need first to expose them in a way that it can be retrieved when F-droid build our app. https://github.com/home-assistant/android/issues/5935

F-droid is ready to use this files https://gitlab.com/helloworldFr/fdroiddata/-/jobs/11744052536 you can see the logs where it tries to pull them. 

They've asked me to wait for the next release to merge the changes I've made.

### DIFF
--- a/.github/workflows/onPush.yml
+++ b/.github/workflows/onPush.yml
@@ -99,7 +99,6 @@ jobs:
             ./automotive/build/outputs/apk/minimal/release/automotive-minimal-release.apk
             ./strings.zip
             ./app/src/main/res/xml/locales_config.xml
-            ./wear/src/main/res/xml/locales_config.xml
 
       - name: Deploy to Firebase
         env:


jpelgrom:
Solution looks OK to me and the comment on the MR suggests this would also be accepted (?).

Can you update the docs page https://developers.home-assistant.io/docs/android/ci/#release-on-f-droid to mention these two files as well? We can approve/merge shortly before doing a release so it's not forgotten.

TimoPtr:
> Solution looks OK to me and the comment on the MR suggests this would also be accepted (?).
> 
> Can you update the docs page https://developers.home-assistant.io/docs/android/ci/#release-on-f-droid to mention these two files as well? We can approve/merge shortly before doing a release so it's not forgotten.

Done https://github.com/home-assistant/developers.home-assistant/pull/2840

PR #5950, opened 2025-10-18
Replace retrofit api dependency with implementation

### LINKED ISSUE #5421: Retrofit should be not exposed outside of common module
Today `retrofit` is an `api` dependency of `:common`. It should be an internal dependency there is no reason to expose it. The only usage is about an exception catching in another Gradle module. We could simply catch the exception in common and rethrow our own error.

**Expectation**
- `retrofit` is now a `implementation` dependency of `:common`

boguszpawlowski:
## Summary

Replaced `api` Retrofit dependency from `common` module with `implementation`.
Closes #5421 

## Checklist


common/src/main/kotlin/io/homeassistant/companion/android/common/data/HttpExceptionCallAdapterFactory.kt
```diff
@@ -0,0 +1,63 @@
...
+import okio.Timeout
+import retrofit2.Call
+import retrofit2.CallAdapter
+import retrofit2.Callback
+import retrofit2.Response
+import retrofit2.Retrofit
+import java.lang.reflect.ParameterizedType
+import java.lang.reflect.Type
+
+class HttpExceptionCallAdapterFactory : CallAdapter.Factory() {
```
boguszpawlowski: Added this adapter to minimize the amount of code changes. IMO we should replace it with an `Result` style adapter to remove exception-based error handling. Im happy to do it in this /separate PR.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/HttpExceptionCallAdapterFactory.kt
```diff
@@ -0,0 +1,53 @@
+package io.homeassistant.companion.android.common.data
+
+import io.homeassistant.companion.android.common.exception.HttpException
```
TimoPtr, line 3:
```suggestion
import io.homeassistant.companion.android.common.data.exception.HttpException
```
And you should also move `MalformedHttpUrlException`

common/src/main/kotlin/io/homeassistant/companion/android/common/data/HttpExceptionCallAdapterFactory.kt
```diff
@@ -0,0 +1,53 @@
...
+import java.lang.reflect.ParameterizedType
+import java.lang.reflect.Type
+import retrofit2.Call
+import retrofit2.CallAdapter
+import retrofit2.Callback
+import retrofit2.Response
+import retrofit2.Retrofit
+
+class HttpExceptionCallAdapterFactory : CallAdapter.Factory() {
```
TimoPtr: From https://developers.home-assistant.io/docs/android/best_practices#documentation `Future-proof: Ask yourself, "Will I understand what I did in 6 months?"`

I think we would benefits from having a documentation on the class definition explaining what is the goal of this factory.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/HttpExceptionCallAdapterFactory.kt
TimoPtr: Changing every API to return a Result<> is a big task. I invite you to first create an issue with proper explanations. It needs to be in another PR (maybe multiples one), then adding a lint rule to force the usage of `Result`.

Currently exception handling in the application is not the best and would need huge changes in general (stopping using `catch(e:Exception)` that is way too broad for instance)

common/src/main/kotlin/io/homeassistant/companion/android/common/data/HttpExceptionCallAdapterFactory.kt
TimoPtr:
```suggestion
internal class HttpExceptionCallAdapterFactory : CallAdapter.Factory() {
```
This should not be exposed outside of the common module.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/HttpExceptionCallAdapterFactory.kt
```diff
@@ -0,0 +1,53 @@
...
+        delegate.enqueue(object : Callback<R> {
+            override fun onResponse(call: Call<R>, response: Response<R>) {
+                callback.onResponse(call, response)
+            }
+
+            override fun onFailure(call: Call<R>, t: Throwable) {
+                val convertedException = if (t is retrofit2.HttpException) {
+                    HttpException(
+                        code = t.code(),
+                        message = t.response()?.errorBody()?.string() ?: t.message(),
```
TimoPtr: I would simply use the message from the exception they already handle everything properly.
```suggestion
                        message = t.message,
```

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlBuilder.kt
```diff
@@ -0,0 +1,29 @@
+package io.homeassistant.companion.android.util
+
+import javax.inject.Inject
+import javax.inject.Singleton
+import okhttp3.HttpUrl
+import okhttp3.HttpUrl.Companion.toHttpUrl
+
+@Singleton
+class UrlBuilder @Inject constructor() {
```
TimoPtr, line 9: I don't think this is relevant to the current PR.

TimoPtr (changes requested):
Today our application is not using `retrofit2.Call` API from retrofit but the coroutine capability of Retrofit, so the factory that you've created won't do anything unfortunately.

Furthermore you are using the `onFailure` from the Call API. The documentation is clear

```
Invoked when a network exception occurred talking to the server or when an unexpected exception occurred creating the request or processing the response.
```

Error like 404 are still handled by `onResponse` according to the documentation
```
Invoked for a received HTTP response.
Note: An HTTP response may still indicate an application-level failure such as a 404 or 500. Call Response.isSuccessful() to determine if the response indicates success.
```

It means that if we merge your PR today the app won't display the right message in `LaunchActivity` since it is going to be the retrofit `HttpException` thrown not yours.

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlBuilder.kt
boguszpawlowski: It is, as without it we were using Retrofit classes inside of the `ConnectionViewModel`

common/src/main/kotlin/io/homeassistant/companion/android/common/data/HttpExceptionCallAdapterFactory.kt
boguszpawlowski: Since this is moved to onResponse - no longer relevant.

boguszpawlowski:
@TimoPtr Thanks for the review and apologies for long inactivity on this. 
> Today our application is not using retrofit2.Call API from retrofit but the coroutine capability of Retrofit, so the factory that you've created won't do anything unfortunately.

All suspending calls use `Call` internally. See: https://github.com/square/retrofit/blob/trunk/CHANGELOG.md#260---2019-06-05, this factory will work as intended. 

> Error like 404 are still handled by onResponse according to the documentation

Great call, forgot about it 👌. Changed the implementation.

Do you maybe know how to reach one of the states in the application that uses the http exception in order to test it? Namely `onComplicationRequest` in `EntityStateDataSourceService` or error case in `registerAndOpenWebview` in `LaunchActivity`.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/HttpExceptionCallAdapterFactory.kt
TimoPtr: Still relevant

common/src/main/kotlin/io/homeassistant/companion/android/common/exception/HttpException.kt
```diff
@@ -0,0 +1,3 @@
+package io.homeassistant.companion.android.common.exception
+
+data class HttpException(val code: Int, override val message: String?) : Exception(message)
```
TimoPtr, line 3: Let's add some documentation since this Exception is public.

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlBuilder.kt
TimoPtr: I'm not sure we need a class for this it could be a utility function within the `UrlUtil` file.

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlBuilder.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    fun buildUrl(base: String, pathSegments: String, parameters: Map<String, String> = emptyMap()): String {
+        val baseUrl = base.toHttpUrl()
```
TimoPtr, line 12: Shouldn't this be a bit safer? with toHttpUrlOrNull. If we use this in another context it might be unsafe.

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlBuilder.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import javax.inject.Inject
+import javax.inject.Singleton
+import okhttp3.HttpUrl
+import okhttp3.HttpUrl.Companion.toHttpUrl
+
+@Singleton
+class UrlBuilder @Inject constructor() {
+
+    fun buildUrl(base: String, pathSegments: String, parameters: Map<String, String> = emptyMap()): String {
```
TimoPtr, line 11: It lacks documenation

common/src/test/kotlin/io/homeassistant/companion/android/util/UrlBuilderTest.kt
```diff
@@ -0,0 +1,243 @@
...
+
+    @Test
+    fun `Given base URL with trailing slash and path segments when building URL then it combines them correctly`() {
+        val result = urlBuilder.buildUrl(
+            base = "http://example.com/",
+            pathSegments = "api/v1/users",
+        )
+
+        assertEquals("http://example.com/api/v1/users", result)
+    }
```
TimoPtr, line 30: I think more tests could be moved into ParametrizedTest.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/HttpExceptionCallAdapterFactory.kt
```diff
@@ -0,0 +1,61 @@
...
+import retrofit2.Retrofit
+
+/**
+ * A factory converting Retrofit-specific HttpException into domain-defined [HttpException].
+ * Prevents leaking Retrofit types into the domain layer.
+ */
+internal class HttpExceptionCallAdapterFactory : CallAdapter.Factory() {
+    override fun get(returnType: Type, annotations: Array<out Annotation>, retrofit: Retrofit): CallAdapter<*, *>? {
+        if (getRawType(returnType) != Call::class.java) {
+            return null
```
TimoPtr, line 19: Do we want to capture when this happen? Like if it shouldn't ever happen let's use `FailFast` here.

TimoPtr (commented):
Sorry for the delay to get back to your comment. I don't have an easy path to reproduce the error sorry. Maybe @jpelgrom have an idea.

I've reviewed your changed and left some other comments on top of the Copilot ones. Lemme know if you have any issue with resolving the conflicts. You can also reach out on discord if needed.

PR #5963, opened 2025-10-24
Disable lint tasks while building release

TimoPtr:
## Summary

It seems that `-x lint` was not enough to avoid the lint task in release build
image

This PR tries to use a Gradle property instead.

On this build https://github.com/home-assistant/android/actions/runs/18784473104/job/53599246854?pr=5963 we can see that no lint tasks are triggered we went from 10mn (when lint task is cached) of build to 8mn.

The changes does also apply to the build happening on `main`. This should help us save some ram and avoiding having some builds to fail. The counterpart is that we are not checking the lint issues on main anymore. In theory it should be ok since we do validate this on PRs, but since we are not enforcing the PR to be up to date on main we might actually merge lint issue on main when merging branches that are not up to date.

I think that for now it is acceptable, if it not we could bring it back or enforce being up to date on main to merge (or use the merge queue)

### DIFF
--- a/app/lint-baseline.xml
+++ b/app/lint-baseline.xml
@@ -1,0 +1,930 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<issues format="6" by="lint 8.13.0" type="baseline" client="gradle" dependencies="false" name="AGP (8.13.0)" variant="all" version="8.13.0">
+
+    <issue
+        id="MissingPermission"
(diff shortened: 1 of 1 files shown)


jpelgrom:
> In theory it should be ok since we do validate this on PRs, but since we are not enforcing the PR to be up to date on main we might actually merge lint issue on main when merging branches that are not up to date.

Which historically has been unlikely, and at which point it'll be noticed shortly after a new PR is submitted. Lint issues rarely break the build or create major issues, so I think that's a reasonable compromise.

PR #5976, opened 2025-10-27
Replaced org.json usages with kotlinx serialization

### LINKED ISSUE #5306: Remove usage of org.json in favor of Kotlinx serialization
To keep things uniform in the app we should replace the usage of `org.json` in the app with Kotlinx serialization.

asiridol:
## Summary

Replaced org.json usages with kotlinx serialization (resolves #5306)

## Checklist

## Any other notes

No new unit tests were added as this is a direct replacement of the serialization library

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
@@ -1,5 +1,6 @@
 package io.homeassistant.companion.android.common.util
 
+import io.homeassistant.companion.android.common.BuildConfig
 import io.homeassistant.companion.android.common.data.websocket.impl.entities.SocketResponse
 import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter
@@ -40,6 +41,15 @@ import kotlinx.serialization.serializer
 import kotlinx.serialization.serializerOrNull
 import timber.log.Timber
 
+/**
+ * Converts a JsonArray to a list of strings by extracting the primitive content from each element.
+ *
+ * This assumes all elements in the array are JsonPrimitive values. If any element is not a
+ * JsonPrimitive (for example, a JsonObject or JsonArray), this will throw an exception.
+ *
+ * @return A list containing the string content of each array element
+ * @throws IllegalArgumentException if any element is not a JsonPrimitive
+ */
 fun JsonArray.toStringList(): List<String> = List(size) { i ->
     val element = this[i]
     element.jsonPrimitive.content
@@ -269,130 +279,237 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
 }
 
 /**
- * Returns current element as JsonObject. If current element is not a JsonObject (ex/ JsonPrimitive) this will return null
+ * Safely casts this JsonElement to a JsonObject.
+ *
+ * @return The element as a JsonObject if it is one, null otherwise (for example, if it's a
+ *         JsonPrimitive, JsonArray, or JsonNull)
  */
 fun JsonElement.jsonObjectOrNull(): JsonObject? = this as? JsonObject
 
 /**
- * Returns current element as JsonArray. If current element is not a JsonArray (ex/ JsonPrimitive) this will return null
+ * Safely casts this JsonElement to a JsonArray.
+ *
+ * @return The element as a JsonArray if it is one, null otherwise (for example, if it's a
+ *         JsonPrimitive, JsonObject, or JsonNull)
  */
 fun JsonElement.jsonArrayOrNull(): JsonArray? = this as? JsonArray
 
 /**
- * Converts a map to a JsonObject
+ * Converts a map with string keys and arbitrary values to a JsonObject.
+ *
+ * This uses [MapAnySerializer] to handle the conversion, which supports primitive types,
+ * collections, nested maps, and serializable objects. See [MapAnySerializer] documentation
+ * for details on supported types and limitations.
+ *
+ * Example:
+ * ```kotlin
+ * val data = mapOf(
+ *     "name" to "Alice",
+ *     "age" to 30,
+ *     "active" to true
+ * )
+ * val json = data.toJsonObject()
+ * ```
+ *
+ * @return A JsonObject representation of this map
+ * @throws IllegalArgumentException if the map contains unsupported value types or non-string keys in nested maps
  */
-
 fun Map<String, Any?>.toJsonObject(): JsonObject {
     return kotlinJsonMapper.encodeToJsonElement(MapAnySerializer, this) as JsonObject
 }
 
 /**
- * Parses a string to a JsonObject. This will return null if;\
- * the string is empty or\
- * if the given string is not a valid json or\
- * if the given string cannot be mapped to a json object
+ * Parses a JSON string to a JsonObject, returning null if parsing fails.
+ *
+ * This function safely handles various failure scenarios including empty strings, malformed JSON,
+ * and JSON that represents non-object types (for example, arrays or primitives).
+ *
+ * Example:
+ * ```kotlin
+ * val json = """{"name": "Alice", "age": 30}""".toJsonObjectOrNull()
+ * // json is a JsonObject with properties "name" and "age"
+ *
+ * val invalid = "not valid json".toJsonObjectOrNull()
+ * // invalid is null
+ *
+ * val array = """["item1", "item2"]""".toJsonObjectOrNull()
+ * // array is null (because it's a JSON array, not an object)
+ * ```
+ *
+ * @return A JsonObject if the string is valid JSON representing an object, null otherwise
  */
 fun String.toJsonObjectOrNull(): JsonObject? {
     if (this.isEmpty()) return null
     return runCatching {
         Json.parseToJsonElement(this) as? JsonObject
     }.onFailure {
-        Timber.w("Failed to convert to a json object $this")
+        Timber.w("Failed to convert to a json object: ${if (BuildConfig.DEBUG) this else "HIDDEN"}")
     }.getOrNull()
 }
 
 /**
- * Returns the value to which the specified key is mapped, or null if;\
- * this JsonObject contains no mapping for the key or\
- * this mapping for the key is JsonNull or\
- * this mapping for the key is not a JsonPrimitive or\
- * this mapping for the key is not a string
- * @param key a string value to lookup json structure
+ * Retrieves a string value from this JsonObject, returning null if the value cannot be accessed.
+ *
+ * This function safely handles various scenarios including missing keys, null values, and
+ * type mismatches. It will return null if:
+ * - The key does not exist in the JsonObject
+ * - The value for the key is JsonNull
+ * - The value is not a JsonPrimitive (for example, it's a JsonObject or JsonArray)
+ * - The primitive value is not a string (for example, it's a number or boolean)
+ *
+ * Example:
+ * ```kotlin
+ * val json = buildJsonObject {
+ *     put("name", JsonPrimitive("Alice"))
+ *     put("age", JsonPrimitive(30))
+ * }
+ * json.getStringOrNull("name")  // Returns "Alice"
+ * json.getStringOrNull("age")   // Returns null (it's a number, not a string)
+ * json.getStringOrNull("email") // Returns null (key doesn't exist)
+ * ```
+ *
+ * @param key The key to look up in the JsonObject
+ * @return The string value if present and valid, null otherwise
  */
 fun JsonObject.getStringOrNull(key: String): String? {
     return runCatching {
-        if (this.containsKey(key)) {
-            val value = this[key]
-            if (value is JsonNull) null else value?.jsonPrimitive?.content
-        } else {
-            null
-        }
+        val value = this[key]
+        if (value is JsonNull) null else value?.jsonPrimitive?.content
     }.onFailure {
-        Timber.w("Failed to get string value for $key in jsonObject $this")
+        Timber.w("Failed to get string value for $key in jsonObject: ${if (BuildConfig.DEBUG) this else "HIDDEN"}")
     }.getOrNull()
 }
 
 /**
- * Returns the value to which the specified key is mapped, or fallback if;\
- * this JsonObject contains no mapping for the key or\
- * this mapping for the key is JsonNull or\
- * this mapping for the key is not a JsonPrimitive or\
- * this mapping for the key is not a string
- * @param key a string value to lookup json structure
- * @param fallback a string used as the fallback if the key is missing or value is null
+ * Retrieves a string value from this JsonObject, returning a fallback value if the value cannot be accessed.
+ *
+ * This is a convenience function that combines [getStringOrNull] with a default value. It will
+ * return the fallback if the key doesn't exist, the value is null, or the value cannot be
+ * converted to a string.
+ *
+ * Example:
+ * ```kotlin
+ * val json = buildJsonObject {
+ *     put("name", JsonPrimitive("Alice"))
+ * }
+ * json.getStringOrElse("name", "Unknown")    // Returns "Alice"
+ * json.getStringOrElse("email", "No Email")  // Returns "No Email"
+ * ```
+ *
+ * @param key The key to look up in the JsonObject
+ * @param fallback The default value to return if the key is missing or invalid
+ * @return The string value if present and valid, the fallback value otherwise
  */
 fun JsonObject.getStringOrElse(key: String, fallback: String): String = this.getStringOrNull(key) ?: fallback
 
 /**
- * Returns the value to which the specified key is mapped, or null if;\
- * this JsonObject contains no mapping for the key or\
- * this mapping for the key is JsonNull or\
- * this mapping for the key is not a JsonPrimitive or\
- * this mapping for the key is not a boolean
- * @param key a string value to lookup json structure
+ * Retrieves a boolean value from this JsonObject, returning null if the value cannot be accessed.
+ *
+ * This function safely handles various scenarios including missing keys, null values, and
+ * type mismatches. It will return null if:
+ * - The key does not exist in the JsonObject
+ * - The value for the key is JsonNull
+ * - The value is not a JsonPrimitive (for example, it's a JsonObject or JsonArray)
+ * - The primitive value is not a boolean (for example, it's a string or number)
+ *
+ * Example:
+ * ```kotlin
+ * val json = buildJsonObject {
+ *     put("active", JsonPrimitive(true))
+ *     put("name", JsonPrimitive("Alice"))
+ * }
+ * json.getBooleanOrNull("active")  // Returns true
+ * json.getBooleanOrNull("name")    // Returns null (it's a string, not a boolean)
+ * json.getBooleanOrNull("deleted") // Returns null (key doesn't exist)
+ * ```
+ *
+ * @param key The key to look up in the JsonObject
+ * @return The boolean value if present and valid, null otherwise
  */
 fun JsonObject.getBooleanOrNull(key: String): Boolean? {
     return runCatching {
-        if (this.containsKey(key)) {
-            val value = this[key]
-            if (value is JsonNull) null else value?.jsonPrimitive?.booleanOrNull
-        } else {
-            null
-        }
+        val value = this[key]
+        if (value is JsonNull) null else value?.jsonPrimitive?.booleanOrNull
     }.onFailure {
-        Timber.w("Failed to get boolean value for $key in jsonObject $this")
+        Timber.w("Failed to get boolean value for $key in jsonObject: ${if (BuildConfig.DEBUG) this else "HIDDEN"}")
     }.getOrNull()
 }
 
 /**
- * Returns the value to which the specified key is mapped, or fallback if;\
- * this JsonObject contains no mapping for the key or\
- * this mapping for the key is JsonNull or\
- * this mapping for the key is not a JsonPrimitive or\
- * this mapping for the key is not a boolean
- * @param key a string value to lookup json structure
- * @param fallback a boolean value used as the fallback if the key is missing or value is null
+ * Retrieves a boolean value from this JsonObject, returning a fallback value if the value cannot be accessed.
+ *
+ * This is a convenience function that combines [getBooleanOrNull] with a default value. It will
+ * return the fallback if the key doesn't exist, the value is null, or the value cannot be
+ * converted to a boolean.
+ *
+ * Example:
+ * ```kotlin
+ * val json = buildJsonObject {
+ *     put("active", JsonPrimitive(true))
+ * }
+ * json.getBooleanOrElse("active", false)  // Returns true
+ * json.getBooleanOrElse("deleted", false) // Returns false
+ * ```
+ *
+ * @param key The key to look up in the JsonObject
+ * @param fallback The default value to return if the key is missing or invalid
+ * @return The boolean value if present and valid, the fallback value otherwise
  */
 fun JsonObject.getBooleanOrElse(key: String, fallback: Boolean): Boolean = this.getBooleanOrNull(key) ?: fallback
 
 /**
- * Returns the value to which the specified key is mapped, or null if;\
- * this JsonObject contains no mapping for the key or\
- * this mapping for the key is JsonNull or\
- * this mapping for the key is not a JsonPrimitive or\
- * this mapping for the key is not an integer
- * @param key a string value to lookup json structure
+ * Retrieves an integer value from this JsonObject, returning null if the value cannot be accessed.
+ *
+ * This function safely handles various scenarios including missing keys, null values, and
+ * type mismatches. It will return null if:
+ * - The key does not exist in the JsonObject
+ * - The value for the key is JsonNull
+ * - The value is not a JsonPrimitive (for example, it's a JsonObject or JsonArray)
+ * - The primitive value is not an integer (for example, it's a string or boolean, or a number too large for Int)
+ *
+ * Note: This will return null for floating-point numbers. Use JsonPrimitive.int only for integer values.
+ *
+ * Example:
+ * ```kotlin
+ * val json = buildJsonObject {
+ *     put("age", JsonPrimitive(30))
+ *     put("name", JsonPrimitive("Alice"))
+ * }
+ * json.getIntOrNull("age")   // Returns 30
+ * json.getIntOrNull("name")  // Returns null (it's a string, not an int)
+ * json.getIntOrNull("score") // Returns null (key doesn't exist)
+ * ```
+ *
+ * @param key The key to look up in the JsonObject
+ * @return The integer value if present and valid, null otherwise
  */
 fun JsonObject.getIntOrNull(key: String): Int? {
     return runCatching {
-        if (this.containsKey(key)) {
-            val value = this[key]
-            if (value is JsonNull) null else value?.jsonPrimitive?.intOrNull
-        } else {
-            null
-        }
+        val value = this[key]
+        if (value is JsonNull) null else value?.jsonPrimitive?.intOrNull
     }.onFailure {
-        Timber.w("Failed to get integer value for $key in jsonObject $this")
+        Timber.w("Failed to get integer value for $key in jsonObject: ${if (BuildConfig.DEBUG) this else "HIDDEN"}")
     }.getOrNull()
 }
 
 /**
- * Returns the value to which the specified key is mapped, or fallback if;\
- * this JsonObject contains no mapping for the key or\
- * this mapping for the key is JsonNull or\
- * this mapping for the key is not a JsonPrimitive or\
- * this mapping for the key is not an integer
- * @param key a string value to lookup json structure
- * @param fallback an integer used as the fallback if the key is missing or value for key is null
+ * Retrieves an integer value from this JsonObject, returning a fallback value if the value cannot be accessed.
+ *
+ * This is a convenience function that combines [getIntOrNull] with a default value. It will
+ * return the fallback if the key doesn't exist, the value is null, or the value cannot be
+ * converted to an integer.
+ *
+ * Example:
+ * ```kotlin
+ * val json = buildJsonObject {
+ *     put("age", JsonPrimitive(30))
+ * }
+ * json.getIntOrElse("age", 0)   // Returns 30
+ * json.getIntOrElse("score", 0) // Returns 0
+ * ```
+ *
+ * @param key The key to look up in the JsonObject
+ * @param fallback The default value to return if the key is missing or invalid
+ * @return The integer value if present and valid, the fallback value otherwise
  */
 fun JsonObject.getIntOrElse(key: String, fallback: Int): Int = this.getIntOrNull(key) ?: fallback
--- a/common/src/test/kotlin/io/homeassistant/companion/android/common/util/JsonUtilTest.kt
+++ b/common/src/test/kotlin/io/homeassistant/companion/android/common/util/JsonUtilTest.kt
@@ -131,18 +131,6 @@ class JsonUtilTest {
     }
 
     @Test
-    fun `Given a map with an object when toJsonObjectOrNull is called then it doesn't throw`() {
-        val map = mapOf<String, Any?>("key1" to Stub("value"))
-
-        try {
-            kotlinJsonMapper.encodeToString(MapAnySerializer, map)
-            fail { "encodeToString should fail with unsupported type" }
-        } catch (e: IllegalArgumentException) {
-            assertEquals("Unsupported type: ${Stub::class}", e.message)
-        }
-    }
-
-    @Test
     fun `Given a json array of strings when toStringList is called then returns a list of strings`() {
         val jsonArray = buildJsonArray {
             repeat(5) {
@@ -152,7 +140,7 @@ class JsonUtilTest {
 
         val stringList = jsonArray.toStringList()
         assertNotNull(stringList)
-        assertTrue { stringList.isNotEmpty() }
+        assertTrue(stringList.isNotEmpty())
         assertEquals(5, stringList.count())
         assertEquals(listOf("Item 0", "Item 1", "Item 2", "Item 3", "Item 4"), stringList)
     }
@@ -448,7 +436,6 @@ class JsonUtilTest {
         assertTrue(value4.isNullOrEmpty())
     }
 
-    @OptIn(ExperimentalSerializationApi::class)
     @Test
     fun `Given an existing key with value empty string when getStringOrNull on JsonObject then returns empty string`() {
         val jsonObject = buildJsonObject {
@@ -462,9 +449,8 @@ class JsonUtilTest {
         assertEquals("", value4)
     }
 
-    @OptIn(ExperimentalSerializationApi::class)
     @Test
-    fun `Given an existing key with value empty string when getStringOrNull on JsonObject then returns value`() {
+    fun `Given an existing key with value when getStringOrNull on JsonObject then returns value`() {
         val jsonObject = buildJsonObject {
             put("key1", JsonPrimitive("value1"))
             put("key2", JsonPrimitive("value2"))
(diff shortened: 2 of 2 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepositoryImpl.kt
```diff
@@ -139,10 +140,10 @@ class WearPrefsRepositoryImpl @Inject constructor(
     }
 
     private suspend fun setTileShortcuts(map: Map<Int?, List<String>>) {
-        val jsonArrayMap = map.map { (tileId, entities) ->
-            tileId.toString() to JSONArray(entities)
-        }.toMap()
-        val jsonStr = JSONObject(jsonArrayMap).toString()
+//        val jsonArrayMap = map.map { (tileId, entities) ->
+//            tileId.toString() to entities.toJsonArray()
+//        }.toMap()
```
jpelgrom: Please remove commented out code if it's no longer necessary.

(I thought we had an automatic check for this)

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1122,7 +1155,7 @@ class WebViewActivity :
                     },
                 )
                 prepare()
-                volume = if (isMuted) 0f else 1f
+                volume = if (isMuted == true) 0f else 1f
```
jpelgrom: This seems more specific than it needs to be as your extension returns a non-null Boolean? 


```suggestion
                volume = if (isMuted) 0f else 1f
```

app/src/main/kotlin/io/homeassistant/companion/android/barcode/BarcodeScannerViewModel.kt
```diff
@@ -42,7 +44,7 @@ class BarcodeScannerViewModel @Inject constructor(
             ).collect { message ->
                 when (val type = BarcodeActionType.fromExternalBus(message.getString("type"))) {
                     BarcodeActionType.NOTIFY -> frontendActionsFlow.emit(
-                        BarcodeScannerAction(type, message.getJSONObject("payload").getString("message")),
+                        BarcodeScannerAction(type, message.get("payload")?.jsonObject?.getString("message")),
```
TimoPtr: if for any reason the payload is not an object this is going to crash
```kotlin
public val JsonElement.jsonObject: JsonObject
    get() = this as? JsonObject ?: error("JsonObject")
```

We would create `jsonObjectOrNull` or do like I did in other places `as? JsonObject`.

TimoPtr: IDE is suggesting and I prefer it too.
```suggestion
                        BarcodeScannerAction(type, message["payload"]?.jsonObject?.getString("message")),
```

TimoPtr: I get that the previous implementation was using empty string. But here you are creating a new API on top of the Json API of Kotlinx serialization and I would like to follow Kotlin design in terms of APIs.

As example the `get` operator on an Iterable
- `.get(index)` you are expecting the index to be within the bounds of the iterable if it is not it throw
- `.getOrNull(index)` means that you are not sure if index is going to be within the bounds
- `.getOrElse()`means that it might not be there but you want to handle this case differently (small helper arround getOrNull in the end)

So I would apply the same for your getString and most probably for now only implement `getOrNull` variant.

Since you do also create this new API I want tests for it.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -721,30 +727,30 @@ class WebViewActivity :
 
                     @JavascriptInterface
                     fun getExternalAuth(payload: String) {
-                        JSONObject(payload).let {
+                        payload.toJsonObject().let {
```
TimoPtr: I think we would benefit here from something like
`payload.toJsonObjectOrNull` here.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -721,30 +727,30 @@ class WebViewActivity :
...
                             presenter.onGetExternalAuth(
                                 this@WebViewActivity,
                                 it.getString("callback"),
-                                it.has("force") && it.getBoolean("force"),
+                                it.containsKey("force") && it.getBoolean("force"),
```
TimoPtr:
```suggestion
                                it.getBooleanOrElse("force", false),
```

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -822,8 +835,8 @@ class WebViewActivity :
                                 "tag/write" ->
                                     writeNfcTag.launch(
                                         WriteNfcTag.Input(
-                                            tagId = json.getJSONObject("payload").getString("tag"),
-                                            messageId = JSONObject(message).getInt("id"),
+                                            tagId = json.get("payload")?.jsonObject?.getString("tag"),
```
TimoPtr:
```suggestion
                                            tagId = json["payload"]?.jsonObject?.getString("tag"),
```

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -838,15 +851,26 @@ class WebViewActivity :
                                 }
 
                                 "bar_code/scan" -> {
-                                    val payload = if (json.has("payload")) json.getJSONObject("payload") else null
-                                    if (payload?.has("title") != true || !payload.has("description")) return@post
+                                    val payload = if (json.containsKey(
+                                            "payload",
+                                        )
+                                    ) {
+                                        json.get("payload")?.jsonObject
```
TimoPtr:
```suggestion
                                        json["payload"]?.jsonObject
```

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -862,15 +886,24 @@ class WebViewActivity :
...
-                                    val payload = if (json.has("payload")) json.getJSONObject("payload") else null
-                                    if (payload?.has("name") != true) return@post
+                                    val payload = if (json.containsKey(
+                                            "payload",
+                                        )
+                                    ) {
+                                        json.get("payload")?.jsonObject
+                                    } else {
+                                        null
+                                    }
```
TimoPtr: Could be simplify with getOrNull

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -862,15 +886,24 @@ class WebViewActivity :
...
+                                    if (payload?.containsKey("name") != true) return@post
                                     configureImprovDevice(payload.getString("name"))
                                 }
 
                                 "exoplayer/play_hls" -> exoPlayHls(json)
                                 "exoplayer/stop" -> exoStopHls()
                                 "exoplayer/resize" -> exoResizeHls(json)
-                                "haptic" -> processHaptic(json.getJSONObject("payload").getString("hapticType"))
+                                "haptic" -> processHaptic(
+                                    json.get("payload")?.jsonObject?.getString("hapticType") ?: "",
```
TimoPtr:
```suggestion
                                    json["payload"]?.jsonObject?.getString("hapticType") ?: "",
```

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1152,8 +1185,8 @@ class WebViewActivity :
     }
 
     @OptIn(UnstableApi::class)
-    fun exoResizeHls(json: JSONObject) {
-        val payload = json.getJSONObject("payload")
+    fun exoResizeHls(json: JsonObject) {
+        val payload = json.get("payload")?.jsonObject ?: return
```
TimoPtr: Simplify

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/NetworkSensorManager.kt
```diff
@@ -556,9 +556,9 @@ class NetworkSensorManager : SensorManager {
                 override fun onResponse(call: Call, response: Response) {
                     if (!response.isSuccessful) throw IOException("Unexpected response code $response")
                     try {
-                        val jsonObject = JSONObject(response.body.string())
+                        val jsonObject = response.body.string().toJsonObject()
                         ip = jsonObject.getString("ip")
-                    } catch (e: JSONException) {
+                    } catch (e: Exception) {
```
TimoPtr: Please avoid global catching here you should be able to determine the potential exception and handle them properly.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +266,76 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
         }
     }
 }
+
+/**
+ * Converts a map to a JsonObject
+ */
+@Suppress("UNCHECKED_CAST")
+fun Map<String, Any?>.toJsonObject(): JsonObject {
```
TimoPtr, line 317: This looks a lot to what we already have in `io.homeassistant.companion.android.common.util.MapAnySerializer` I would suggest to reuse what we have.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +266,76 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
...
+    }
+    return JsonObject(content)
+}
+
+/**
+ * Parses a string to a JsonObject
+ */
+fun String.toJsonObject(): JsonObject {
+    return Json.parseToJsonElement(this) as JsonObject
+}
```
TimoPtr: The cast might not be safe. Check the Kotlinx serialization impl the cast should always be `as?` and then `?: error(...)` if you want to return a non null object. On our side it could even use the `FailFast` api

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +266,76 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
...
+
+/**
+ * Returns a string value for a key, if the key is missing or null an empty string is returned
+ * @param key a string value to lookup json structure
+ */
+fun JsonObject.getString(key: String): String {
+    return if (this.containsKey(key)) {
+        this[key]?.jsonPrimitive?.content ?: ""
```
TimoPtr: containsKey doesn't mean that the jsonElement is a primitive it can throw an error.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +266,76 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
...
+    }
+}
+
+/**
+ * Returns a boolean value for a key, if the key is missing or null false is returned
+ * @param key a string value to lookup json structure
+ */
+fun JsonObject.getBoolean(key: String): Boolean {
+    return if (this.containsKey(key)) {
+        this[key]?.jsonPrimitive?.booleanOrNull ?: false
```
TimoPtr: same here

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +266,76 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
...
+    }
+}
+
+/**
+ * Returns an integer value for a key, if the key is missing or null 0 is returned
+ * @param key a string value to lookup json structure
+ */
+fun JsonObject.getInt(key: String): Int {
+    return if (this.containsKey(key)) {
+        this[key]?.jsonPrimitive?.intOrNull ?: 0
```
TimoPtr: and here

onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/wearmtls/WearMTLSViewModelTest.kt
```diff
@@ -47,11 +53,13 @@ class WearMTLSViewModelTest {
             every { contentResolver } returns this@WearMTLSViewModelTest.mockContentResolver
         }
         viewModel = WearMTLSViewModel(context)
+        Dispatchers.setMain(testDispatcher)
     }
 
     @AfterEach
     fun tearDown() {
         unmockkStatic(KeyStore::class)
+        Dispatchers.resetMain()
```
TimoPtr: This is unreleated to this PR and also not needed since this class is annotated with `@ExtendWith(MainDispatcherJUnit5Extension::class)`. 

The test `Given viewModel, when onPasswordChanged is called multiple times rapidly, then only the last value should trigger validation` is indeed flaky and should be fixed but this change is not the way to do it.

wear/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -46,7 +46,7 @@ class MessagingManager @Inject constructor(
             val now = System.currentTimeMillis()
 
             val jsonData = notificationData as Map<String, String>?
-            val jsonObject = jsonData?.let { JSONObject(it) }
+            val jsonObject = jsonData?.let { it.toJsonObject() }
```
TimoPtr:
```suggestion
            val jsonObject = jsonData?.toJsonObject()
```

app/src/main/kotlin/io/homeassistant/companion/android/barcode/BarcodeScannerViewModel.kt
asiridol: Here jsonObject is a convenience method provided by the SDK, but I get your point about making it more robust. Will do some more testing around this.

asiridol:
> In your previous PR you mentioned that you didn't run/compile the app because of timeouts. Did you fix that and run the app on a device or emulator this time?

Yes it was a local issue and works fine now.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -721,30 +730,30 @@ class WebViewActivity :
...
                         Timber.d("External bus $message")
                         webView.post {
-                            val json = JSONObject(message)
-                            when (json.get("type")) {
+                            val json = message.toJsonObjectOrNull() ?: return@post
+                            when (json.getStringOrNull("type")) {
                                 "connection-status" -> {
-                                    isConnected = json.getJSONObject("payload")
-                                        .getString("event") == "connected"
+                                    isConnected = json["payload"]?.jsonObject
```
TimoPtr: `.jsonObject` could throw here

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -791,25 +800,20 @@ class WebViewActivity :
                                 }
 
                                 "assist/show" -> {
-                                    val payload = if (json.has("payload")) json.getJSONObject("payload") else null
+                                    val payload = if (json.containsKey(
+                                            "payload",
+                                        )
+                                    ) {
+                                        json["payload"]?.jsonObject
```
TimoPtr: could throw

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -822,8 +826,8 @@ class WebViewActivity :
                                 "tag/write" ->
                                     writeNfcTag.launch(
                                         WriteNfcTag.Input(
-                                            tagId = json.getJSONObject("payload").getString("tag"),
-                                            messageId = JSONObject(message).getInt("id"),
+                                            tagId = json["payload"]?.jsonObject?.getStringOrNull("tag"),
```
TimoPtr: Could throw

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -838,22 +842,26 @@ class WebViewActivity :
                                 }
 
                                 "bar_code/scan" -> {
-                                    val payload = if (json.has("payload")) json.getJSONObject("payload") else null
-                                    if (payload?.has("title") != true || !payload.has("description")) return@post
+                                    val payload = if (json.containsKey(
+                                            "payload",
+                                        )
+                                    ) {
+                                        json["payload"]?.jsonObject
```
TimoPtr: same

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -862,15 +870,24 @@ class WebViewActivity :
...
+                                    val deviceName = payload?.getStringOrNull("name") ?: return@post
+                                    configureImprovDevice(deviceName)
                                 }
 
                                 "exoplayer/play_hls" -> exoPlayHls(json)
                                 "exoplayer/stop" -> exoStopHls()
                                 "exoplayer/resize" -> exoResizeHls(json)
-                                "haptic" -> processHaptic(json.getJSONObject("payload").getString("hapticType"))
+                                "haptic" -> processHaptic(
+                                    json["payload"]?.jsonObject?.getStringOrNull("hapticType") ?: "",
```
TimoPtr: could throw

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1099,10 +1116,10 @@ class WebViewActivity :
         }
     }
 
-    fun exoPlayHls(json: JSONObject) {
-        val payload = json.getJSONObject("payload")
-        val uri = payload.getString("url").toUri()
-        val isMuted = payload.optBoolean("muted")
+    fun exoPlayHls(json: JsonObject) {
+        val payload = json["payload"]?.jsonObject
```
TimoPtr: could throw

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: Still valid

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1152,27 +1169,28 @@ class WebViewActivity :
     }
 
     @OptIn(UnstableApi::class)
-    fun exoResizeHls(json: JSONObject) {
-        val payload = json.getJSONObject("payload")
+    fun exoResizeHls(json: JsonObject) {
+        val payload = json["payload"]?.jsonObject ?: return
```
TimoPtr: could throw

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepositoryImpl.kt
```diff
@@ -108,26 +109,28 @@ class WearPrefsRepositoryImpl @Inject constructor(
...
                 },
                 onFailure = {
                     // backward compatibility with the previous format when there was only one Shortcut Tile:
-                    val jsonArray = JSONArray(jsonStr)
-                    val entities = jsonArray.toStringList()
-                    mapOf(
-                        // the key is null since we don't (yet) have the tileId
-                        null to entities,
-                    )
+                    val jsonArray = jsonStr.toJsonObjectOrNull()?.jsonArray
```
TimoPtr: could throw

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -32,15 +32,17 @@ import kotlinx.serialization.json.floatOrNull
...
 import kotlinx.serialization.serializer
 import kotlinx.serialization.serializerOrNull
-import org.json.JSONArray
+import timber.log.Timber
 
-fun JSONArray.toStringList(): List<String> = List(length()) { i ->
-    getString(i)
+fun JsonArray.toStringList(): List<String> = List(size) { i ->
+    val element = this[i]
+    element.jsonPrimitive.content
```
TimoPtr, line 55: Why changing this?

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +267,125 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
         }
     }
 }
+
+/**
+ * Converts a map to a JsonObject
+ */
+
+fun Map<String, Any?>.toJsonObjectOrNull(): JsonObject {
```
TimoPtr:
```suggestion
 */
fun Map<String, Any?>.toJsonObjectOrNull(): JsonObject {
```

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +267,125 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
...
+    val content = this.mapValues { (_, value) ->
+        value.toJsonElementOrNull()
+    }
+    return JsonObject(content)
+}
+
+/**
+ * Parses a string to a JsonObject
+ */
```
TimoPtr, line 340: Add the case of an empty string and when the return is going to be null. The documentation on these public extension needs to explain the behavior of each.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +267,125 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
...
+    if (this.isEmpty()) return null
+    return runCatching {
+        Json.parseToJsonElement(this) as? JsonObject
+    }.onFailure {
+        Timber.w("Failed to convert to a json object $this")
+    }.getOrNull()
+}
+
+/**
+ * Returns a string value for a key, if the key is missing will return null
```
TimoPtr: not only if the key is missing the return is null.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +267,125 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
...
+    }.onFailure {
+        Timber.w("Failed to get string value for $key in jsonObject $this")
+    }.getOrNull()
+}
+
+/**
+ * Returns a string value for a key, if the key is missing or null will return fabllback value
+ * @param key a string value to lookup json structure
+ * @param fallback a string used as the fallback if the key is missing or value is null
+ */
```
TimoPtr, line 402: You can almost copy paste the documentation from the Kotlin sources for these extensions.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
TimoPtr: This extension never return null, so the name is wrong.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +267,125 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
...
+
+/**
+ * Returns an integer value for a key, if the key is missing or null will return fallback value
+ * @param key a string value to lookup json structure
+ * @param fallback an integer used as the fallback if the key is missing or value for key is null
+ */
+fun JsonObject.getIntOrElse(key: String, fallback: Int): Int = this.getIntOrNull(key) ?: fallback
+
+@Suppress("UNCHECKED_CAST")
+private fun Any?.toJsonElementOrNull(): JsonElement = when (this) {
```
TimoPtr: Why are you creating a new extension?

```kotlin
kotlinJsonMapper.encodeToJsonElement(value)
```

This is going to invoke `toJsonElement` private function with all the edge cases already handled. Since you've made already the tests, changing the implementation should be rather easy and your test should continue to pass.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
asiridol: Was using JSONArray from org.json

TimoPtr: I missed it 👍

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
asiridol: oops missed it

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
asiridol: Will throw SerializationException as a serializer for type Any is not registered. Instead using this method directly I'll just add this to else block like this;

        val serializer = runCatching {
            kotlinJsonMapper.serializersModule.serializer(this::class.java)
        }.getOrNull()
        // try to use a known serializer
        if (serializer != null) {
            kotlinJsonMapper.encodeToJsonElement(serializer, this)
        } else {
            // fallback
            JsonPrimitive(this.toString())
        }

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -791,25 +801,20 @@ class WebViewActivity :
...
                                 "assist/show" -> {
-                                    val payload = if (json.has("payload")) json.getJSONObject("payload") else null
+                                    val payload = if (json.containsKey(
+                                            "payload",
+                                        )
+                                    ) {
+                                        json["payload"]?.jsonObjectOrNull()
+                                    } else {
+                                        null
+                                    }
```
TimoPtr:
```suggestion
                                    val payload = json["payload"]?.jsonObjectOrNull()
```

[] returns null already so no need to check for contains anymore

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -838,22 +843,26 @@ class WebViewActivity :
...
-                                    val payload = if (json.has("payload")) json.getJSONObject("payload") else null
-                                    if (payload?.has("title") != true || !payload.has("description")) return@post
+                                    val payload = if (json.containsKey(
+                                            "payload",
+                                        )
+                                    ) {
+                                        json["payload"]?.jsonObjectOrNull()
+                                    } else {
+                                        null
+                                    }
```
TimoPtr:
```suggestion
                                    val payload = json["payload"]?.jsonObjectOrNull()
```

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -862,15 +871,24 @@ class WebViewActivity :
...
-                                    if (payload?.has("name") != true) return@post
-                                    configureImprovDevice(payload.getString("name"))
+                                    val payload = if (json.containsKey(
+                                            "payload",
+                                        )
+                                    ) {
+                                        json["payload"]?.jsonObjectOrNull()
+                                    } else {
+                                        null
+                                    }
```
TimoPtr:
```suggestion
                                    val payload = json["payload"]?.jsonObjectOrNull()
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/WearPrefsRepositoryImpl.kt
```diff
@@ -108,26 +111,40 @@ class WearPrefsRepositoryImpl @Inject constructor(
...
                             val intKey = stringKey.takeUnless { it == "null" }?.toInt()
-                            val jsonArray = jsonObject.getJSONArray(stringKey)
-                            val entities = jsonArray.toStringList()
+                            val jsonArray = runCatching {
+                                jsonObject[stringKey]?.jsonArray
+                            }.onFailure { ex ->
+                                Timber.w(
+                                    "Failed to get current element as JsonArray ${jsonStr.toJsonObjectOrNull()}, exception: $ex",
+                                )
+                            }.getOrNull()
```
TimoPtr: You should make `jsonArrayOrNull`

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +267,160 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
...
 }
+
+/**
+ * Returns current element as JsonObject. If current element is not a JsonObject (ex/ JsonPrimitive) will swallow the exception and return null
+ */
+fun JsonElement.jsonObjectOrNull(): JsonObject? = runCatching {
+    this.jsonObject
+}.onFailure { ex ->
+    Timber.w("Failed to get current element as JsonObject $this, exception: $ex")
+}.getOrNull()
```
TimoPtr:
```suggestion
fun JsonElement.jsonObjectOrNull(): JsonObject? = this as? JsonObject
```

We don't need to do more than this, if you check the implementation it only throws when `this` is not an object. Don't forget to update the doc

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
TimoPtr: We should not have a fallback here
```
fun Map<String, Any?>.toJsonObject(): JsonObject {
    return kotlinJsonMapper.encodeToJsonElement(MapAnySerializer, this) as JsonObject
}
```

This should be enough, it has been made exactly for serializing Map<String,Any>

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +267,160 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
...
+/**
+ * Converts a map to a JsonObject
+ */
+
+fun Map<String, Any?>.toJsonObject(): JsonObject {
+    val content = this.mapValues { (_, value) ->
+        value.toJsonElementOrNull()
+    }
+    return JsonObject(content)
+}
```
TimoPtr, line 319:
```suggestion
fun Map<String, Any?>.toJsonObject(): JsonObject {
    return kotlinJsonMapper.encodeToJsonElement(MapAnySerializer, this) as JsonObject
}
```

TimoPtr: I tried it you will have to update your tests because we should not be able to serialize `Stub` since it is not Serializable. We've even made a lint rule for that, but I don't think it can works since it's within the map so not directly detectable by our linter.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/JsonUtil.kt
```diff
@@ -265,3 +267,160 @@ private fun toJsonElement(encoder: JsonEncoder, value: Any?): JsonElement {
...
+
+/**
+ * Parses a string to a JsonObject. If the string is empty this method will return null
```
TimoPtr: It will also return null if it is not a JsonObject. Take the time to go over the doc again 🙏🏻 public API like this needs to be well documented so we don't check the implementation while using it.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -767,7 +777,7 @@ class WebViewActivity :
                                         }
                                     sendExternalBusMessage(
                                         ExternalConfigResponse(
-                                            id = JSONObject(message).get("id"),
+                                            id = message.toJsonObjectOrNull()?.getIntOrNull("id") ?: 0,
```
TimoPtr:
```suggestion
                                            id = json.getIntOrElse("id", 0),
```

Double check but I think this should do the same.

common/src/test/kotlin/io/homeassistant/companion/android/common/util/JsonUtilTest.kt
```diff
@@ -114,12 +129,624 @@ class JsonUtilTest {
...
+    fun `Given a json array of strings when toStringList is called then returns a list of strings`() {
+        val jsonArray = buildJsonArray {
+            repeat(5) {
+                add(JsonPrimitive("Item $it"))
+            }
+        }
+
+        val stringList = jsonArray.toStringList()
+        assertNotNull(stringList)
+        assertTrue { stringList.isNotEmpty() }
```
TimoPtr:
```suggestion
        assertTrue(stringList.isNotEmpty())
```

TimoPtr:
Hey @asiridol I took the liberty to address some nitpicking
- We should avoid logging the JSON content to avoid leaking any secrets going through this (I've dump the content only in debug)
- Remove some redundant tests and annotations
- Update test names
- Remove some `containt(key)` since `[key]` is already checking the key
- Rework the documentation with the help of AI to make it complete

TimoPtr (approved):

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1127,7 +1124,7 @@ class WebViewActivity :
         }
         sendExternalBusMessage(
             ExternalBusMessage(
-                id = json.get("id"),
+                id = json["id"],
```
TimoPtr, line 1127: I missed this ... But currently this crashes because it is a JsonElement which is not properly handle by our custom serializer. I'm going to fix it in a new PR.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -767,7 +777,7 @@ class WebViewActivity :
                                         }
                                     sendExternalBusMessage(
                                         ExternalConfigResponse(
-                                            id = JSONObject(message).get("id"),
+                                            id = json.getIntOrElse("id", 0),
```
TimoPtr, line 780: We should not assume it's a Int, on the frontend it is a Number.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: Nvm the issue is not with JsonElement not handle. It is handled. The issue is that in some case JsonObject contains JsonLiteral that are not serializables. (it doesn't trigger on this PR but it does in my PR https://github.com/home-assistant/android/pull/5622 where I use `get("id")` and not assuming "id" as a Int.

PR #5979, opened 2025-10-28
Remove horizontal safe insets for ModalBottomSheet

TimoPtr:
## Summary

As pointed out in https://github.com/home-assistant/frontend/issues/26818#issuecomment-3451160171 the horizontal insets were wrongly applied. This PR simply removes the horizontal insets.


jpelgrom:
Your frontend screenshot also shows something weird with shadow clipping above the dialog, but I cannot replicate that in either the current code or this branch's code.

jpelgrom:
Nested scrolling also appears to be broken, not sure when that happened. Notice that the button to skip/continue granting permissions is unavailable in the dialog in landscape on a regular size phone and you cannot 'pull up' the content to get to it.

To fix it, we need to add two modifiers. See this commit: https://github.com/jpelgrom/home-assistant-android/commit/b1eb0067a668824b30b14c631967b590c260b790. Tested with server chooser (short height content that doesn't scroll) and Improv permissions (medium height content that does scroll). Feel free to add it as it's based on top of this branch, or let me know if you want a separate PR.

TimoPtr:
Thanks for the fix I've added it to the current PR.

PR #5981, opened 2025-10-29
Connection secure level settings

TimoPtr:
## Summary

Add an entry in the server settings to be able to update the value of the `AllowInsecureConnection`. 

## Checklist

## Screenshots

image

https://github.com/user-attachments/assets/0f052019-55f3-4f98-95a9-c8e0fb94a773

## Any other notes

I've also introduce a high level flag for the onboarding within `WIPFeature`.

### DIFF
--- a/app/src/release/kotlin/io/homeassistant/companion/android/settings/ConnectionSecurityLevelFragment.kt
+++ b/app/src/release/kotlin/io/homeassistant/companion/android/settings/ConnectionSecurityLevelFragment.kt
@@ -9,6 +9,6 @@ class ConnectionSecurityLevelFragment : Fragment() {
 
     companion object {
         const val RESULT_KEY = "connection_security_level_result"
-        const val EXTRA_SEVER_ID_KEY = "server_id"
+        const val EXTRA_SERVER = "server_id"
     }
 }
--- a/onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigationTest.kt
+++ b/onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigationTest.kt
@@ -549,7 +549,7 @@ internal class OnboardingNavigationTest {
             navController.navigateToLocationForSecureConnection(42)
             assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<LocationForSecureConnectionRoute>() == true)
 
-            onNodeWithText(stringResource(R.string.location_secure_connection_less_secure)).performScrollTo().performClick()
+            onNodeWithText(stringResource(commonR.string.connection_security_less_secure)).performScrollTo().performClick()
             onNodeWithText(stringResource(R.string.location_secure_connection_next)).performScrollTo().performClick()
 
             assertTrue(onboardingDone)


jpelgrom:
Is the goal here to have a quick solution to change it, or for the setting to stay like this?

In case it's permanent:

 - I'd expect the screen to have a back/up button when accessed from settings. The giant image and centered title also a bit weird, it is very different from the other settings screens and doesn't make a lot of sense when changing one settings vs. in a flow.

 - Somewhat related to our earlier conversations about this screen, I wonder whether in settings we should just call it "Allow insecure connections". I have no idea how "Remote connection security level": "Allow this app to know when you're home" can be understood as "you need a secure, HTTPS connection to use the app when remote". 

 - Should there be a condition for home network to be set / reminder to do this somewhere? It's not that obvious in my opinion that there after selecting most secure you **must** define your home network for the app to work.

 - The string suggest you intend to allow for a third (default?) option where neither of the settings is selected. How would that be treated differently from allow/disallow?

TimoPtr:
> The string suggest you intend to allow for a third (default?) option where neither of the settings is selected. How would that be treated differently from allow/disallow?

The third state is a temporary state, only possible for user making an update to this new version. We plan to have a screen before the WebView if this is not set and using HTTP (if using HTTPS we simply pick Most secure for the user without a screen), that force the user to pick one or the other.

 > Should there be a condition for home network to be set / reminder to do this somewhere? It's not that obvious in my opinion that there after selecting most secure you must define your home network for the app to work.

The block screen would contains this information about why it is blocking currently.

> Is the goal here to have a quick solution to change it, or for the setting to stay like this?
>
>In case it's permanent:
>
>    I'd expect the screen to have a back/up button when accessed from settings. The giant image and centered title also a bit weird, it is very different from the other settings screens and doesn't make a lot of sense when changing one settings vs. in a flow.

I want to say it's a quick solution, but I don't see the settings adjust to this new design soon, seeing the pace of how we are doing things on the onboarding.
I'll add a back button 👍🏻

> Somewhat related to our earlier conversations about this screen, I wonder whether in settings we should just call it "Allow insecure connections". I have no idea how "Remote connection security level": "Allow this app to know when you're home" can be understood as "you need a secure, HTTPS connection to use the app when remote".

I'm going to discuss this point with @bgoncal

bgoncal:
> I'd expect the screen to have a back/up button when accessed from settings. The giant image and centered title also a bit weird, it is very different from the other settings screens and doesn't make a lot of sense when changing one settings vs. in a flow.

I agree with the back button, but regarding this screen looking different from the rest, this will happen more and more from now on since we are redesigning many parts of Home Assistant itself, with time it will look more seemless. For this screen specifically the reason to keep it the same UI as in onboarding is to keep consistency on the message we want to pass, so the user can remember "Oh, so this was what I chose during onboaridng and here is where to change it".

> Somewhat related to our earlier conversations about this screen, I wonder whether in settings we should just call it "Allow insecure connections". I have no idea how "Remote connection security level": "Allow this app to know when you're home" can be understood as "you need a secure, HTTPS connection to use the app when remote".

So this is again related to what I said above, if we use "Allow insecure connections" we are creating a new "term" in the app, while we already have explained that during onboarding but using a different term "Connection security level", when the user opens the screen it will have more information about that. "Allow insecure connections" is something too direct does not provide clear explanation to what it is really happening, perhaps if you feel that the current screen also doesn't, then we need to improve the copy there as well.

> Should there be a condition for home network to be set / reminder to do this somewhere? It's not that obvious in my opinion that there after selecting most secure you must define your home network for the app to work.

If the user didn't configure it's home network, when they try to access their dashboard a "block connection" screen will be presented explaing what is remaining, check this iOS PR: https://github.com/home-assistant/iOS/pull/3920

> The string suggest you intend to allow for a third (default?) option where neither of the settings is selected. How would that be treated differently from allow/disallow?

The "hidden" third option is the state where the user just migrated to this version of the app with servers already onboarded, so they have to opt to one of those options, there are on an "undefined" state.

jpelgrom:
> The third state is a temporary state, only possible for user making an update to this new version. We plan to have a screen before the WebView if this is not set and using HTTP (if using HTTPS we simply pick Most secure for the user without a screen), that force the user to pick one or the other.

> The "hidden" third option is the state where the user just migrated to this version of the app with servers already onboarded, so they have to opt to one of those options, there are on an "undefined" state.

If it is a hidden state, why create a string for it that will be translated but in practice never be shown?

(FWIW it is a good string, a better explanation what the feature does than everything else.)

app/src/debug/kotlin/io/homeassistant/companion/android/settings/ConnectionSecurityLevelFragment.kt
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * ## Migration Note
+ * This fragment is temporary and should be removed once the app fully migrates to Compose Navigation.
+ */
+@AndroidEntryPoint
+class ConnectionSecurityLevelFragment : Fragment() {
+
+    companion object {
+        const val RESULT_KEY = "connection_security_level_result"
+        const val EXTRA_SEVER_ID_KEY = "server_id"
```
jpelgrom:
```suggestion
        const val EXTRA_SERVER= "server_id"
```

Typo and use the same extra key as other settings fragments

app/src/debug/kotlin/io/homeassistant/companion/android/settings/ConnectionSecurityLevelFragment.kt
```diff
@@ -0,0 +1,110 @@
...
+                    ) {
+                        LocationForSecureConnectionScreen(
+                            viewModel = viewModel,
+                            onGoToNextScreen = {
+                                setFragmentResult(RESULT_KEY, Bundle())
+                                parentFragmentManager.popBackStack()
+                            },
+                            onHelpClick = {
+                                // TODO validate the URL to use
+                                uriHandler.openUri("https://www.home-assistant.io/installation/")
```
jpelgrom: This should link to the pending docs page, I think

app/src/main/kotlin/io/homeassistant/companion/android/settings/server/ServerSettingsFragment.kt
```diff
@@ -363,6 +387,25 @@ class ServerSettingsFragment :
...
         presenter.updateServerName()
         presenter.updateUrlStatus()
+        updateSecurityLevelSummary()
+        potentiallyShowSecurityLevel()
+    }
+
+    private fun potentiallyShowSecurityLevel() {
+        lifecycleScope.launch {
+            findPreference<Preference>("connection_security_level")?.let {
+                it.isVisible = USE_NEW_LAUNCHER && presenter.hasHTTPURL()
```
jpelgrom: This hides the setting even if I previously set it to allow insecure connections, and changed to an HTTPS URL. The setting is still set to allow insecure connections though, which I think is unwanted.

app/src/main/kotlin/io/homeassistant/companion/android/settings/server/ServerSettingsPresenterImpl.kt
```diff
@@ -192,4 +195,17 @@ class ServerSettingsPresenterImpl @Inject constructor(
     override suspend fun serverURL(): String? {
         return serverManager.getServer(serverId)?.connection?.getUrl()?.toString()
     }
+
+    override suspend fun securityLevelSummary(): Int? {
+        val allowInsecureConnection = serverManager.integrationRepository(serverId).getAllowInsecureConnection()
+        return when (allowInsecureConnection) {
```
jpelgrom: Returning the string from this class is a different pattern from most places, where the class returns the settings value and the fragment decides how to display it and set a summary (see: ServerSettingsView functions).

app/src/main/res/xml/preferences_server.xml
```diff
@@ -36,6 +36,12 @@
             android:key="connection_internal"
             android:title="@string/pref_connection_internal"
             app:useSimpleSummaryProvider="true"/>
+        <Preference
+            app:isPreferenceVisible="false"
+            android:key="connection_security_level"
+            android:icon="@drawable/ic_lock"
+            android:title="@string/connection_security_level_title"
+            android:summary="@string/connection_security_level_default_summary"/>
```
jpelgrom: Don't set the default summary to a long string to avoid the screen moving around immediately when opened due to the summary changing.

jpelgrom (commented):
While testing, I did find another scenario where I end up with a null preference: the server added during onboarding had HTTPS, and I manually changed it to HTTP later. The preference should probably be set to true (secure by default) in that case instead of null.

In the screen I'd expect "Save" instead of "Next" when accessed from settings.

app/src/debug/kotlin/io/homeassistant/companion/android/settings/ConnectionSecurityLevelFragment.kt
TimoPtr: Good catch, now that it is merge I'll update. I'm going also to update it in the onboarding flow

app/src/main/res/xml/preferences_server.xml
TimoPtr: I'm going to keep it blank then, it is going to be set programmatically only

app/src/main/kotlin/io/homeassistant/companion/android/settings/server/ServerSettingsFragment.kt
TimoPtr: @bgoncal didn't make this hide action. I think based on your comments it might make sense to also keep it. At least for now even on HTTPS

TimoPtr:
> While testing, I did find another scenario where I end up with a null preference: the server added during onboarding had HTTPS, and I manually changed it to HTTP later. The preference should probably be set to true (secure by default) in that case instead of null.
> 
> In the screen I'd expect "Save" instead of "Next" when accessed from settings.

I'm going to use Save instead of Next also for the onboarding. I think save is valid in both context wdyt?

jpelgrom:
> I'm going to use Save instead of Next also for the onboarding. I think save is valid in both context wdyt

Save works for both, but I think Next is nicer during onboarding as it is a flow with clear previous/next screens.

Thinking more about it, why do we need a save button when accessed from settings anyway? Simply save when the setting is changed, which is a lot more like how other settings on Android function (not just in HA, but any app).

TimoPtr:
> > I'm going to use Save instead of Next also for the onboarding. I think save is valid in both context wdyt
> 
> Save works for both, but I think Next is nicer during onboarding as it is a flow with clear previous/next screens.
> 
> Thinking more about it, why do we need a save button when accessed from settings anyway? Simply save when the setting is changed, which is a lot more like how other settings on Android function (not just in HA, but any app).

We could do like the notification permission? It doesn't close itself you have list of multiple choices and you have to go back yourself. Would that work?

jpelgrom:
@TimoPtr Not sure if you've seen the reaction, but the approach in your last comment sounds good to me. Also [this comment](https://github.com/home-assistant/android/pull/5981#discussion_r2491948516) is still unaddressed.

TimoPtr:
> @TimoPtr Not sure if you've seen the reaction, but the approach in your last comment sounds good to me. Also [this comment](https://github.com/home-assistant/android/pull/5981#discussion_r2491948516) is still unaddressed.

I wanted to do something else today to change my mind :). I should be back on this on Monday and address this two comments.

Did you see that I added some logic that won't work until the race condition is fixed in the server manager?

jpelgrom:
> Did you see that I added some logic that won't work until the race condition is fixed in the server manager?

Yes, but I haven't played with your suggested fixes for that in the issue yet.

TimoPtr:
> > I'm going to use Save instead of Next also for the onboarding. I think save is valid in both context wdyt
> 
> Save works for both, but I think Next is nicer during onboarding as it is a flow with clear previous/next screens.
> 
> Thinking more about it, why do we need a save button when accessed from settings anyway? Simply save when the setting is changed, which is a lot more like how other settings on Android function (not just in HA, but any app).

After trying to implement something like that I found out that the UX would be weird if we go this path especially because we might ask location permission when selecting Most secure (or when going back) and it feels bad. I'm going to go the path of Next/Save by changing the value with a flag.

jpelgrom:
When going to the screen and back, the position of the up arrow seems to shift around a little. Is that something we can fix or a M2/M3 incompatibility? 

Despite all previous arguments, I still don't like how alien this feels in the settings and how vague the description is (needing to read the docs to understand a setting is a red flag). Yet at the same time I don't have a good alternative so I won't reject it.

TimoPtr:
> When going to the screen and back, the position of the up arrow seems to shift around a little. Is that something we can fix or a M2/M3 incompatibility?
> 
> Despite all previous arguments, I still don't like how alien this feels in the settings and how vague the description is (needing to read the docs to understand a setting is a red flag). Yet at the same time I don't have a good alternative so I won't reject it.

It seems that the issue is that the navbar height is not the same between the two version and creates this shift... This should fix itself once we migrate the settings to Material3

TimoPtr:
> Despite all previous arguments, I still don't like how alien this feels in the settings and how vague the description is (needing to read the docs to understand a setting is a red flag). Yet at the same time I don't have a good alternative so I won't reject it.

If you come up with something better 👍🏻 lemme know so we can discuss. Ideally this screen should disappear if HA goes full HTTPS one day.

PR #5982, opened 2025-10-29
Support minimal onboarding

TimoPtr:
## Summary

Modify the onboarding flow to detect when the location sensor are not present using the package name. It simply skips the `LocationSharingScreen`.

## Checklist

## Any other notes

I took the time to document the onboarding graph, even if the function are not public I think the documentation is important on this file.

### DIFF
--- a/onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
+++ b/onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
@@ -92,11 +92,7 @@ internal fun NavController.navigateToOnboarding(
  * 2. Server discovery (only shown if [urlToOnboard] is empty)
  * 3. Connection
  * 3. Device naming and registration
- * 4. Location/security configuration (conditional):
- *    - Public servers + full flavor: Location sharing
- *    - Local servers: Local-first explanation then location sharing (full) or security config (minimal)
- *    - Minimal flavor + HTTP: Location for secure connection detection
- *    - Minimal flavor + HTTPS: Complete
+ * 4. Location/security configuration (conditional)
  * 5. Home network configuration (if applicable)
  *
  * @param navController Navigation controller for managing navigation actions


jpelgrom:
The diff is somewhat annoying to read here because all the added documentation makes it less obvious what changed...

> using the package name

Why not `BuildConfig.FLAVOR`?

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
```diff
@@ -90,6 +114,7 @@ internal fun NavGraphBuilder.onboarding(
     urlToOnboard: String?,
     hideExistingServers: Boolean,
     skipWelcome: Boolean,
+    hasLocationSensors: Boolean = navController.context.packageName?.contains(".minimal")?.not() == true,
```
jpelgrom:
```suggestion
    hasLocationTracking: Boolean = navController.context.packageName?.contains(".minimal")?.not() == true,
```

Personal annoyance, but I wouldn't call it sensors as the feature may change and they are not sensors for core. (https://github.com/home-assistant/android/issues/5931#issuecomment-3403376832)

TimoPtr: It makes sense 👍🏻

TimoPtr:
> The diff is somewhat annoying to read here because all the added documentation makes it less obvious what changed...
> 
> > using the package name
> 
> Why not `BuildConfig.FLAVOR`?

FLAVOR doesn't exist in the onboarding and I didn't want to add a flavor to the Gradle module.

jpelgrom:
> FLAVOR doesn't exist in the onboarding and I didn't want to add a flavor to the Gradle module.

Do you plan to merge onboarding into the app module once finished? Otherwise we'll end up with different checks depending on the module for the same thing, which isn't nice. Maybe it is possible to 'forward' that info somehow.

onboarding/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
```diff
@@ -78,10 +78,34 @@ internal fun NavController.navigateToOnboarding(
...
+ * ## Flow Overview
+ * 1. Welcome screen (only shown if [skipWelcome] is false)
+ * 2. Server discovery (only shown if [urlToOnboard] is empty)
+ * 3. Connection
+ * 3. Device naming and registration
+ * 4. Location/security configuration (conditional):
+ *    - Public servers + full flavor: Location sharing
+ *    - Local servers: Local-first explanation then location sharing (full) or security config (minimal)
+ *    - Minimal flavor + HTTP: Location for secure connection detection
+ *    - Minimal flavor + HTTPS: Complete
```
jpelgrom:
```suggestion
 * 4. Location/security configuration (conditional)
```

This repeats the more detailed description on the functions below and risks diverging from that, so I wouldn't add a lot of detail here.

jpelgrom (commented):
Tested a couple of variations and I think this is working as intended, nice job!

TimoPtr:
> > FLAVOR doesn't exist in the onboarding and I didn't want to add a flavor to the Gradle module.
> 
> Do you plan to merge onboarding into the app module once finished? Otherwise we'll end up with different checks depending on the module for the same thing, which isn't nice. Maybe it is possible to 'forward' that info somehow.

For now my plan is to keep onboarding to see if I can split things by big features. My guess is that we are going inject things in the dagger graph instead of using the package name, like the sensors so we can check if a sensor is present or not.

jpelgrom:
> For now my plan is to keep onboarding to see if I can split things by big features. My guess is that we are going inject things in the dagger graph instead of using the package name, like the sensors so we can check if a sensor is present or not.

I'm not completely sold, but we'll see as time goes on.

For now: my current problem with the check is that it checks for "not minimal", while the check should be "is full". This will break if we introduce another flavor or someone uses a different package name.

TimoPtr:
> > For now my plan is to keep onboarding to see if I can split things by big features. My guess is that we are going inject things in the dagger graph instead of using the package name, like the sensors so we can check if a sensor is present or not.
> 
> I'm not completely sold, but we'll see as time goes on.
> 
> For now: my current problem with the check is that it checks for "not minimal", while the check should be "is full". This will break if we introduce another flavor or someone uses a different package name.


Lemme reformulate this need to change before the release and I would like to do it once I start moving parts around.

jpelgrom (approved):
Approving based on the understanding that flavor/minimal detection will be revisited before new onboarding is released

PR #5998, opened 2025-11-04
WebsocketManager start method use coroutine instead of a blocking call

TimoPtr:
## Summary

The start method used to start the WebSocketManager worker was using a blocking call that could cause some hangs when loading the WebView for instance. This PR makes sure we use a coroutine to start the Worker so it doesn't block any thread while waiting for the result.

## Any other notes

There was a typo in the TAG of the worker too.


app/src/main/kotlin/io/homeassistant/companion/android/websocket/WebsocketBroadcastReceiver.kt
```diff
@@ -3,9 +3,17 @@ package io.homeassistant.companion.android.websocket
...
+import kotlinx.coroutines.launch
 
 class WebsocketBroadcastReceiver : BroadcastReceiver() {
+    private val ioScope: CoroutineScope = CoroutineScope(Dispatchers.IO + Job())
+
     override fun onReceive(context: Context, intent: Intent) {
-        WebsocketManager.start(context)
+        ioScope.launch {
+            WebsocketManager.start(context)
+        }
```
TimoPtr, line 17: @jpelgrom I think we should do something about all these scopes that we are created in the app without a lifecycle attached. We should maybe use the GlobalScope instead of creating all these scopes (since it is basically the same), or at least a wrapper around the GlobalScope and avoid using it whenever we can...

I also think we could make something like this
```kotlin
abstract class CoroutineBroadcastReceiver : BroadcastReceiver() {
    final override fun onReceive(context: Context, intent: Intent) {
        GlobalScope.launch {
            withTimeout(10.seconds) {
                onReceiveIntent(context, intent)
            }
        }
    }

    abstract suspend fun onReceiveIntent(context: Context, intent: Intent)
}
```
In the same spirit as CoroutineWorker

jpelgrom:
> There was a typo in the TAG of the worker too.

Not a typo, intentional to keep the tag short in logs but you ditched that when switching to Timber 😊

app/src/main/kotlin/io/homeassistant/companion/android/websocket/WebsocketBroadcastReceiver.kt
jpelgrom: You think we should do something about it but are also doing it here?!

The example looks nice but I wouldn't introduce it at the same time for every potential use combined with this change, as it may have side effects we do not anticipate. So just for this PR for now, and switch others later after testing.

TimoPtr: I stick to the old way because I didn't want to introduce something like this in this PR.

I wanted to start opening the discussion with an example. I've created a new task for this new API https://github.com/home-assistant/android/issues/6000

TimoPtr: This one is more an audit I would say https://github.com/home-assistant/android/issues/6001

PR #6014, opened 2025-11-08
Disable weekly push

TimoPtr:
Google is currently testing a version, we don't want to cancel their test by pushing a new version tomorrow morning.


jpelgrom:
Disabled manually in the UI

PR #6030, opened 2025-11-12
Expose LocationTrackingSupport at the app level

TimoPtr:
## Summary

Address the comment from @jpelgrom about using the package name https://github.com/home-assistant/android/pull/5982#issuecomment-3499019175

## Checklist

## Any other notes

In a longer run maybe we could think of keeping an enum of all potential sensors in `common` and retrieve the list of current sensor available to check if a given sensor is available.


jpelgrom:
That's a lot of places to inject the boolean :)

> In a longer run maybe we could think of keeping an enum of all potential sensors in common and retrieve the list of current sensor available to check if a given sensor is available.

What does this have to do with sensors?

TimoPtr:
> That's a lot of places to inject the boolean :)
> 
> > In a longer run maybe we could think of keeping an enum of all potential sensors in common and retrieve the list of current sensor available to check if a given sensor is available.
> 
> What does this have to do with sensors?

Ideally I would like to not inject a boolean. I would like to inject the list of available sensors and we can check if a sensor is enabled or not but that would be quite huge in terms of changes I think.

jpelgrom:
Oh right, you're thinking of location tracking it as a sensor. I wouldn't do that but that's a well known argument at this point.

Issue #5846, opened 2025-09-30
Auto enable/disable logs in unit test using a local variable


TimoPtr:
### Description

The CI fails to read the result of the tests when the xml files are too big https://github.com/home-assistant/android/pull/5845/checks?check_run_id=51524177822 It happens because we are enabling the logs and that it can generates a lot of noise, in this specific case it generated a 10Mb file. 

The logs are important but for 90% of the case we don't need them on CI, mostly while debugging. We should disable them on CI by default and enable them only when we rerun the action with the flag more logs https://docs.github.com/en/actions/how-tos/monitor-workflows/enable-debug-logging

In local environment the logs for the logs should be enabled by default.

Today we have to manually add the `Timber.plant` on every test file. We should come up with a more automated approach that enables the logs for any test files without having to add it manually.

PR #6031, opened 2025-11-12
Rework UrlUtil handle to avoid crashes while converting URI to URL

TimoPtr:
## Summary

While looking at Sentry crashes I found out that there was a case not handled by the `handle` function in `UrlUtil`, where a string like `http://h:8123None` can be parsed as a URI but then `.toURL` would crash with `NumberFormatException`.

To fix this I've asked Claude to first generate a set of tests, to be able to reproduce the issue (it was not trivial to found a case that triggers the issue from the crash only). 
Then once identified I took the liberty to change the implementation with more details.

image
This only crashes when something is wrong in the URL. But it might happens since it's retrieved from a notification so a user could make a mistake.

## Checklist


common/src/test/kotlin/io/homeassistant/companion/android/util/UrlUtilTest.kt
```diff
@@ -2,14 +2,135 @@ package io.homeassistant.companion.android.util
...
+        // instead of the normalized input, so "homeassistant://navigate/https://..." is treated as relative
+        val input = "homeassistant://navigate/https://example.com/path"
+
+        val result = UrlUtil.handle(baseUrl, input)
+
+        // Current buggy behavior: addPathSegments treats "https://example.com/path" as path segments
+        // splitting on "/" and adding each segment, resulting in a malformed URL
+        assertNotNull(result)
+        // assertEquals("https://example.com:8123/https:/example.com/path", result.toString())
+        assertEquals("https://example.com:8123/path", result.toString())
```
TimoPtr: @jpelgrom could you check this test in particular? The comments are from Claude and I didn't want to apply his suggestion until we discuss it. Which are to test the `isAbsolute` onto the `normalizedInput` instead of the `input`.

TimoPtr:
Test reporting are failing because the XML contains all the logs when the tests are running and it's too big for the tool we use to proceed.

https://github.com/home-assistant/android/issues/5846

common/src/test/kotlin/io/homeassistant/companion/android/util/UrlUtilTest.kt
jpelgrom: Nice find, I suppose.

It is supposed to be a custom URI scheme specifically to be used for creating a URL _relative to your Home Assistant server_, which is needed because your base URL may change (internal/external). Always treating it as relative therefore makes the most sense to me.

The current check does that, indirectly (`homeassistant` protocol != `http(s)`, so not absolute). After thinking about it, I'd probably keep it as it is right now because:
 - changing it to check the normalized input does mean that weird input like here can end up being treated like a absolute url, which is explicitly not intended
 - we have a custom scheme so can simply say it's not supported (and HA doesn't have paths like this)
 - checking it as absolute may also makes it easier to have a redirect attack (external app opens LinkActivity with `homeassistant://navigate/a`; currently you end up on your server, and we have checks to get out of the WebViewActivity if the host is different, but why risk it?)

TimoPtr: I've updated the test to remove the comments and make the title more explicit.

PR #6034, opened 2025-11-12
Fix: properly setup console logs for testing

### LINKED ISSUE #5846: Auto enable/disable logs in unit test using a local variable

TimoPtr:
## Summary

I never took the time to debug why the logs were displayed multiple times in the console since I've introduce the ConsoleLog object. But it turns out that we can call multiple times `Timber.plant` and it is going to add multiple times the same `Tree`. Since most of the test were setting up the ConsoleLog into `Before` or `BeforeEach` it was adding the `Tree` each time a test was starting so for a class with a lot of test the logs were growing exponentially in the test report. 

This PR make sure to use a proper rule and extension for the logger. It is still manual but it makes the logs cleaner by setting it up according to the lifecycle of the tests.

## Checklist

## Any other notes

I've splited the PR in 2, the first part is the implementation and the second part replace the invocation everywhere. 

This PR closes #5846 since now it shouldn't be an issue to keep the logs all the time.

### DIFF
--- a/onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
+++ b/onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
@@ -32,6 +32,7 @@ import kotlinx.coroutines.test.runTest
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertFalse
 import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertNotNull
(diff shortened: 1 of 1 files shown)


jpelgrom:
One minor merge conflict to resolve

PR #6037, opened 2025-11-13
Make sure onSensorUpdated happens on Default dispatcher to avoid ANRs

TimoPtr:
## Summary

It happens that the `onSensorUpdated` is taking too long and should not be invoked from the `Main` dispatcher. It is too long because it does Json de-serialization. There are no reason to keep the execution of this function in `Main` if it is called from `Main` which it is the case for instance in `SensorDetailViewModel` that invoke the `requestSensorUpdate` that could invoke `onSensorUpdated` (we have a sample with the `BedTimeSensor`).

```kotlin
fun setSetting(setting: SensorSetting) {
        viewModelScope.launch {
            sensorDao.add(setting)
            try {
                sensorManager?.requestSensorUpdate(getApplication())
            } catch (e: Exception) {
                Timber.e(e, "Exception while requesting update for sensor $sensorId")
            }
            refreshSensorData()
        }
    }
```

## Checklist

## Any other notes

It is probably not the only problem we have like this, but this one is new and impacting some users in the last release.
image

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/SensorManager.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/SensorManager.kt
@@ -24,6 +24,7 @@ import java.util.Locale
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.withContext
 
 interface SensorManager {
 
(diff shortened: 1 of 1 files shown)


jpelgrom (approved):

jpelgrom:
This change seems good to me, but maybe it could even be moved to IO?

TimoPtr:
> This change seems good to me, but maybe it could even be moved to IO?

Based on the documentation, I think it could stay in default they kinda share the same pool in the end. We are not making any network or disk call.

jpelgrom:
Fair enough, Room will already move database calls to IO. Ready to merge?

TimoPtr:
> Fair enough, Room will already move database calls to IO. Ready to merge?

I think so yes 🙂‍↕️

PR #6039, opened 2025-11-13
Remove WindowsInsets override in LocationForSecureConnectionScreen

TimoPtr:
## Summary

While testing on Android Automotive I found out that the Next button was not visible anymore. I've overridden the WindowsInsets while fixing another issue while exposing the `LocationForSecureConnectionScreen`. But it was not necessary and should have been removed before merging.

## Checklist

## Screenshots

image
image


jpelgrom:
Various other onboarding screens use `contentWindowInsets = WindowInsets.safeDrawing`. When do you use that and when do you use nothing? (This is a detail that wasn't really discussed while you were working on the initial implementation of all screens, and everything seemed to work, but now might be a good time to ask.)

TimoPtr:
> WindowInsets.safeDrawing

Good catch. It works because the default value in `Scafold` is almost the same as `safeDrawing` it is `WindowInsets.systemBarsForVisualComponents`.

But I'm going to align all the screen by using `safeDrawing`.

jpelgrom:
If the other screens were working why not stick to the default?

TimoPtr:
> If the other screens were working why not stick to the default?

None lemme remove. I think at first I've added it to be more explicit.

TimoPtr:
@jpelgrom done I had to trick a bit in the Fragment for the connection security level.

jpelgrom:
With these changes, the fragment no longer appears to be edge to edge in settings. Is that expected? Most of the other onboarding screens also isn't so it is more consistent in onboarding now.

(see gesture navigation handle)

image

TimoPtr:
> With these changes, the fragment no longer appears to be edge to edge in settings. Is that expected? Most of the other onboarding screens also isn't so it is more consistent in onboarding now.
> 
> (see gesture navigation handle)
> image

I think it's fine especially because making go behind it add extra logic that we might break like I did already... So keeping it simple, also the screen was initially meant to be not scrollable (obviously it is not on all the screens).

PR #6041, opened 2025-11-13
Hide Add to widget on Automotive

TimoPtr:
## Summary

Add to widget actions should not be displayed on Automotive since we can't add widget there.

## Checklist

## Screenshots

image

### DIFF
--- a/app/src/test/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToHandlerTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToHandlerTest.kt
@@ -104,7 +104,7 @@ class EntityAddToHandlerTest {
         val entityId = "light.test"
         coEvery { serverManager.getServer() } returns null
 
-        val actions = handler.actionsForEntity(false, entityId)
+        val actions = handler.actionsForEntity(isAutomotive = false, isQuest = false, entityId = entityId)
 
         assertEquals(emptyList<EntityAddToAction>(), actions)
     }
@@ -114,7 +114,7 @@ class EntityAddToHandlerTest {
         val entityId = "light.nonexistent"
         coEvery { integrationRepository.getEntity(entityId) } returns null
 
-        val actions = handler.actionsForEntity(false, entityId)
+        val actions = handler.actionsForEntity(isAutomotive = false, isQuest = false, entityId)
 
         assertEquals(emptyList<EntityAddToAction>(), actions)
     }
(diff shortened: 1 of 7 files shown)


jpelgrom:
Looking at our settings:

https://github.com/home-assistant/android/blob/a8b5adf22f8e4b15e4da14772b6a9fe72080009a/app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt#L192

These actions should also be hidden for Meta Quest devices which do not support widgets.

TimoPtr:
> Looking at our settings:
> 
> https://github.com/home-assistant/android/blob/a8b5adf22f8e4b15e4da14772b6a9fe72080009a/app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt#L192
> 
> These actions should also be hidden for Meta Quest devices which do not support widgets.

Nice catch. I've handle it and also took the opportunity to centralize the check of Quest into a class to ease future refactor.

app/src/test/kotlin/io/homeassistant/companion/android/webview/addto/EntityAddToHandlerTest.kt
```diff
@@ -237,6 +239,70 @@ class EntityAddToHandlerTest {
...
+            "light.test",
+            "alarm_control_panel.test",
+        ],
+    )
+    fun `Given standard entity on Quest when getting actionsForEntity then returns auto favorite`(entityId: String) = runTest {
+        mockGetEntity(entityId)
+
+        val actions = handler.actionsForEntity(isAutomotive = false, isQuest = true, entityId)
+
+        assertEquals(listOf(EntityAddToAction.AndroidAutoFavorite), actions)
```
jpelgrom, line 286: This test may technically work but you realistically won't have Android Auto on a Meta Quest (supposed to use minimal flavor, no Google services)

TimoPtr: We should in that case make sure to not return this Action

jpelgrom: Yes - but maybe not in this PR to avoid even more scope creep. Agreed?

TimoPtr: Indeed, in that case are you ok merging like it is today?

For the AndroidAuto we could check if we are on a minimal build when not automotive?

jpelgrom: > Indeed, in that case are you ok merging like it is today?

Yes

> For the AndroidAuto we could check if we are on a minimal build when not automotive?

The check should be inverted, [like](https://github.com/home-assistant/android/blob/a6ed21cdb08b7a20f4f6cbe8f242031388d079f8/app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt#L385-L387) in settings: flavor is full (or system is Automotive) (and also check SDK level)

PR #6058, opened 2025-11-17
Add back missing ruby deps

TimoPtr:
## Summary

After merging https://github.com/home-assistant/android/pull/6057 `fastlane` started to fail
https://github.com/home-assistant/android/actions/runs/19426545153/job/55578827605
image

This PR brings back the missing deps in the lock file


TimoPtr:
I tested locally before opening the PR and I was able to reproduce the issue.

PR #6063, opened 2025-11-18
Update translation for Set Home network

TimoPtr:
## Summary

Based on feedbacks and modification made on iOS this PR update the translation for the Set Home Network.

### DIFF
--- a/onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkScreenTest.kt
+++ b/onboarding/src/test/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkScreenTest.kt
@@ -146,7 +146,7 @@ class SetHomeNetworkScreenTest {
 
             onNodeWithText(stringResource(R.string.set_home_network_title)).performScrollTo().assertIsDisplayed()
             onNodeWithText(stringResource(R.string.set_home_network_content)).assertIsDisplayed()
-            onNodeWithText(stringResource(commonR.string.manage_ssids_wifi)).assertIsDisplayed()
+            onNodeWithText(stringResource(R.string.set_home_network_wifi_name)).assertIsDisplayed()
 
             dsl()
         }


jpelgrom:
The image files in this PR appear to be corrupted for me, both when loading them in the browser and after downloading the file. Do they show correctly for you?

I'm unable to find these strings in the iOS repo, can you link to the PR that changed them?

TimoPtr:
> The image files in this PR appear to be corrupted for me, both when loading them in the browser and after downloading the file. Do they show correctly for you?
> 
> I'm unable to find these strings in the iOS repo, can you link to the PR that changed them?

Maybe it was due to the Cloudflare outrage yesterday. I can see them properly.

For iOS the translations are on Lokalize even the English ones. Right @bgoncal?

bgoncal:
https://github.com/home-assistant/iOS/pull/3992

The strings start on iOS codebase but once they are merged, it is uploaded to Lokalise, from this moment on they need to be updated there instead the codebase.

Finally, it gets imported back to the project through a PR like this.

jpelgrom:
GitHub had a disruption of their own yesterday when I was reviewing this. Images load correctly now and look good 👍


Weird flow for iOS strings in my opinion, and I didn't check that PR assuming it was only other languages, so thanks for explaining.

common/src/main/res/values/strings.xml
```diff
@@ -479,7 +479,7 @@
     <string name="manage_ssids_input">New Wi-Fi SSID</string>
     <string name="manage_ssids_introduction">Indicate when you\'re connected to your home network, based on Wi-Fi, ethernet or VPN. This can be used to, for example, use the internal connection URL or disable the app lock when at home.</string>
     <string name="manage_ssids_permission">Home Assistant needs additional access to read Wi-Fi information.</string>
-    <string name="manage_ssids_wifi">Wi-Fi network connected</string>
+    <string name="manage_ssids_wifi">Name of the Wi-Fi network at home</string>
```
jpelgrom: I can understand and support this change in the onboarding flow, but in the app's settings you end up with a weird setting name because
 - it is followed by a text field to add more Wi-Fi networks
 - the text field already has a hint
 - the text field also allows BSSID entry (hidden power user feature)

To fix that, I'd keep the current string and introduce a new string with the suggested change, specifically for the simpler field in onboarding.

image

PR #6095, opened 2025-11-25
Add battery optimization permission to onboarding

TimoPtr:
## Summary

Add missing battery optimization to the onboarding flow. It was there before the PR is bringing this to the onboarding. The permission is going to be displayed on the Set home network screen.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/locationsharing/LocationSharingScreen.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/locationsharing/LocationSharingScreen.kt
@@ -1,5 +1,7 @@
 package io.homeassistant.companion.android.onboarding.locationsharing
 
+import androidx.activity.compose.rememberLauncherForActivityResult
+import androidx.activity.result.contract.ActivityResultContracts
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/onboarding/locationforsecureconnection/LocationForSecureConnectionScreen.kt
```diff
@@ -147,6 +150,7 @@ private fun LocationForSecureConnectionContent(
         val permissions = rememberLocationPermission(
             onPermissionResult = { permissionGranted ->
                 if (permissionGranted) {
+                    context.maybeAskForIgnoringBatteryOptimizations()
```
jpelgrom: The placement on this screen feels a bit random to me. We prefer no background restrictions for location tracking to receive timely updates. It's not really related to more/less secure or needed to read the SSID, so the user may not understand why this is shown after selecting the option.

TimoPtr: Fair enough I'm going to remove it from here, I didn't think that through TBH.

jpelgrom (approved):

jpelgrom:
Onboarding already continues in the background while the permission pop-up is shown, but I don't think that is an issue as we can continue without approval.

jpelgrom:
Actually, I can't get the dialog to show when onboarding with a https url, probably because location sharing is the last step and continue means the activity transitions. Did you test that flow?

TimoPtr:
> Actually, I can't get the dialog to show when onboarding with a https url, probably because location sharing is the last step and continue means the activity transitions. Did you test that flow? 

Nope I didn't test with HTTPS you're right... But then we have an issue until the WebView is part of the graph ....and not a standalone activity.

Or we need to show this dialog after the notification permission but that kinda sucks. Any other ideas?

jpelgrom:
`startActivityForResult` and wait for it to return? (not tested)

TimoPtr:
> `startActivityForResult` and wait for it to return? (not tested)

Good call 👍🏻 It does work as intended, I tested on Android 9 and 6.

jpelgrom:
Tested again and now it waits before continuing, and also works as expected with a https url.

PR #6097, opened 2025-11-25
Add set security level when not set while opening a server

TimoPtr:
## Summary

This PR is going to show the ConnectionSecurityLevelFragment if the current server has a plain text URL in its URLs and the allowInsecure flag is not set yet.

## Checklist

### DIFF
--- a/app/src/test/kotlin/io/homeassistant/companion/android/util/compose/HAAppTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/util/compose/HAAppTest.kt
@@ -43,7 +43,8 @@ import io.mockk.verify
 import kotlinx.coroutines.test.runTest
 import org.junit.Rule
 import org.junit.Test
-import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertTrue
 import org.junit.jupiter.api.assertNull
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
@@ -112,7 +113,7 @@ class HAAppTest {
     @Test
     fun `Given OnboardingRoute with skipWelcome without urlToOnboard as start when starts then show ServerDiscovery`() {
         testApp(OnboardingRoute(hasLocationTracking = true, skipWelcome = true)) {
-            Assertions.assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<ServerDiscoveryRoute>() == true)
+            assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<ServerDiscoveryRoute>() == true)
         }
     }
 
@@ -131,7 +132,7 @@ class HAAppTest {
     @Test
     fun `Given FrontendRoute as start when starts then navigate to Frontend and finish current activity`() {
         testApp(FrontendRoute()) {
-            Assertions.assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<FrontendRoute>() == true)
+            assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<FrontendRoute>() == true)
             verify(exactly = 1) {
                 activityNavigator.navigate(
                     match {
@@ -143,14 +144,14 @@ class HAAppTest {
                 )
             }
             // TODO remove this once we are using WebViewActivity anymore
-            Assertions.assertTrue(activity.isFinishing)
+            assertTrue(activity.isFinishing)
         }
     }
 
     @Test
     fun `Given WearOnboardingRoute with url to onboard as start when starts then navigate to ConnectionScreen`() {
         testApp(WearOnboardingRoute("wear", "http://ha")) {
-            Assertions.assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<ConnectionRoute>() == true)
+            assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<ConnectionRoute>() == true)
             onNodeWithTag(HA_WEBVIEW_TAG).assertIsDisplayed()
         }
     }
@@ -171,7 +172,7 @@ class HAAppTest {
             onNodeWithText(stringResource(R.string.connection_security_less_secure)).performScrollTo().performClick()
             onNodeWithText(stringResource(R.string.location_secure_connection_next)).performScrollTo().assertIsEnabled().assertIsDisplayed().performClick()
 
-            Assertions.assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<FrontendRoute>() == true)
+            assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<FrontendRoute>() == true)
         }
     }
 


app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenter.kt
```diff
@@ -52,6 +52,15 @@ interface WebViewPresenter {
...
     suspend fun isSsidUsed(): Boolean
 
+    /**
+     * Checks if the security level has been configured for the current server.
+     *
+     * @return `true` if the server has no plain text URLs, or if the allowInsecureConnection
+     *         flag has been explicitly set (to either true or false). Returns `false` if
+     *         the server has plain text URLs and the user hasn't yet chosen a security level.
+     */
+    suspend fun isSecurityLevelSet(): Boolean
```
jpelgrom: I find the function name here a bit misleading because even if it isn't set, you return `true` if there's only HTTPS. That still doesn't automatically put the security level on most secure anywhere?

TimoPtr: Updated 👍🏻

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1401,6 +1410,32 @@ class WebViewActivity :
...
+                waitForConnection()
+            }
+        }
+
+        val fragment = ConnectionSecurityLevelFragment().apply {
+            arguments = Bundle().apply {
+                putInt(ConnectionSecurityLevelFragment.EXTRA_SERVER, serverId)
+                putBoolean(ConnectionSecurityLevelFragment.EXTRA_HANDLE_ALL_INSETS, true)
+            }
+        }
```
jpelgrom: Now that there are two arguments and we create the fragment in two places, I kind of prefer the `newInstance` function we have in a lot of places to get an intent to a fragment without needing to build the bundle yourself. You'd get something like:

```kotlin
ConnectionSecurityLevelFragment.newInstance(serverId, handleAllInsets = true)
```

TimoPtr: I'm having hard time to work on these screen because I would like to get rid of this and use the navigation graph... Which is not possible until we move the WebViewActivity.

TimoPtr:
> Do you have a good way to test this without manually editing the app code to skip this step? 😅

Not really unfortunately ... I did tweak the code, you could also use a different version of the app without the flag (before the new onboarding like 2025.8.7)

> Why is there an up arrow if we want the user to choose before continuing?

Well we had a long disussion with @bgoncal about this screen being optional, he made it optional on iOS with a cross. If the user close it with the cross it's going to be shown again when opening the app again next time. At first I wanted to make this mandatory but I get the point that a user might just want to do something quick in the app and doesn't have the time to read.

jpelgrom:
Easiest way I found to test: onboard using https url which doesn't set the preference, then change it to http afterwards.

---

 - When temporarily skipping the screen, I can use the up button on screen and this will load the dashboard. However, using the system back button doesn't do that and instead I end up with a blank screen that does nothing.

 - If the app is paused and resumed with the connection security level screen visible, you still get an overlay for the frontend timeout ("Could not connect to Home Assistant"). This can happen with the user hiding/resuming the app, but also if you get a permission dialog for 'most secure' > deny and wait a bit.

> Well we had a long disussion with @bgoncal about this screen being optional, he made it optional on iOS with a cross. If the user close it with the cross it's going to be shown again when opening the app again next time. 

OK, I see how not forcing but showing it every time until a choice is made is nice. However, maybe we should also show a close/cross button in this situation to make that more obvious? It's not uncommon for Android apps to have that as the first icon in the top app bar. (I'm also thinking of a redesigned frontend timeout screen, where you want a close instead of up button.)

TimoPtr:
> I'm also thinking of a redesigned frontend timeout screen, where you want a close instead of up button.

What do you mean?

I'm going to make it a cross instead, to make it simpler is it fine for you to also use a cross when the fragment is opened from the settings too?

jpelgrom:
> > I'm also thinking of a redesigned frontend timeout screen, where you want a close instead of up button.
> 
> What do you mean?

My assumption was that you were working on replacing the frontend timeout pop-up with a full screen dialog that guides the user more on how to fix it like iOS / the onboarding connection error screen? In that one I'd expect a top app bar with a refresh or close icon in the top left.

> I'm going to make it a cross instead, to make it simpler is it fine for you to also use a cross when the fragment is opened from the settings too?

Well in the settings screen an up arrow makes more sense 😅 not a dialog but a normal full screen in that case

TimoPtr:
> > > I'm also thinking of a redesigned frontend timeout screen, where you want a close instead of up button.
> > 
> > What do you mean?
> 
> My assumption was that you were working on replacing the frontend timeout pop-up with a full screen dialog that guides the user more on how to fix it like iOS / the onboarding connection error screen? In that one I'd expect a top app bar with a refresh or close icon in the top left.
> 

I'm going to do this once we drop the WebViewActivity not before.


> > I'm going to make it a cross instead, to make it simpler is it fine for you to also use a cross when the fragment is opened from the settings too?
> 
> Well in the settings screen an up arrow makes more sense 😅 not a dialog but a normal full screen in that case

🥲 I'm going to handle both cases on Wednesday.

TimoPtr:
Done 👍🏻

jpelgrom:
> If the app is paused and resumed with the connection security level screen visible, you still get an overlay for the frontend timeout ("Could not connect to Home Assistant"). This can happen with the user hiding/resuming the app, but also if you get a permission dialog for 'most secure' > deny and wait a bit.

This still hasn't been addressed.

TimoPtr:
> > If the app is paused and resumed with the connection security level screen visible, you still get an overlay for the frontend timeout ("Could not connect to Home Assistant"). This can happen with the user hiding/resuming the app, but also if you get a permission dialog for 'most secure' > deny and wait a bit.
> 
> This still hasn't been addressed.

I thought I did by checking if a fragment is displayed when calling wait for connection. But maybe I'm wrong and the fragment is not displayed anymore.

TimoPtr:
> > > If the app is paused and resumed with the connection security level screen visible, you still get an overlay for the frontend timeout ("Could not connect to Home Assistant"). This can happen with the user hiding/resuming the app, but also if you get a permission dialog for 'most secure' > deny and wait a bit.
> > 
> > 
> > This still hasn't been addressed.
> 
> I thought I did by checking if a fragment is displayed when calling wait for connection. But maybe I'm wrong and the fragment is not displayed anymore.

@jpelgrom I'm curious about how to reproducing it, because I've added some logs and I can see that
- When opening the app for the first time we don't waitForConnection because we don't have the loaded URL yet, then we set the URL but we don't call waitForConnection since we show the fragment.
- When putting on background and then foreground waitForConnection is called but it doesn't do anything since there is a fragment being displayed so it simply continue.

jpelgrom:
> @jpelgrom I'm curious about how to reproducing it, because I've added some logs and I can see that
> 
>     * When opening the app for the first time we don't waitForConnection because we don't have the loaded URL yet, then we set the URL but we don't call waitForConnection since we show the fragment.
> 
>     * When putting on background and then foreground waitForConnection is called but it doesn't do anything since there is a fragment being displayed so it simply continue.

I've just tried reproducing it again and I'm also unable to with the latest commit on this branch. Maybe I broke something with a manual edit or it was fixed in a later commit. Sorry to keep you waiting, in case this turned out to be a real issue we'll surely encounter it again after merging.

PR #6103, opened 2025-11-27
Block insecure connection on the WebView

TimoPtr:
## Summary

Add a new screen on top of the WebViewActivity that prevents loading the URL if we are in an insecure state and the user picked the `Most secured` option. This screen change based on the current state and can display banner with button to help the user 'fix' the situation.

## Checklist

## Screenshots

image
image
image

## Any other notes

The screen doesn't refresh on its own.


common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
```diff
@@ -142,6 +152,39 @@ data class ServerConnectionInfo(
...
+     * When insecure, [SecurityStatus.Insecure] provides guidance on what's missing:
+     * - [SecurityStatus.Insecure.missingHomeSetup]: No internal URL, VPN, or Ethernet configured
+     * - [SecurityStatus.Insecure.missingLocation]: Location permission denied or location services disabled,
+     *   which prevents SSID-based internal network detection
+     *
+     * @param context Android context for checking location permission and services
+     * @param url The URL being used for the connection
+     * @return [SecurityStatus.Secure] if the connection is secure, [SecurityStatus.Insecure] otherwise
+     */
+    fun currentSecurityStatusForUrl(context: Context, url: String): SecurityStatus {
```
jpelgrom, line 172: "for a given url" where you provide the url shouldn't exist when it is an extension function on the class that holds the URLs. It should not be an extension function here, determine the url to check automatically, or allow you to pass an enum for internal/external/cloud.

app/src/main/kotlin/io/homeassistant/companion/android/webview/insecure/BlockInsecureFragment.kt
```diff
@@ -0,0 +1,112 @@
...
+                    putInt(EXTRA_SERVER, serverId)
+                    putBoolean(EXTRA_MISSING_HOME_SETUP, missingHomeSetup)
+                    putBoolean(EXTRA_MISSING_LOCATION, missingLocation)
+                }
+            }
+        }
+    }
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+        val serverId = arguments?.getInt(EXTRA_SERVER, -1) ?: -1
```
jpelgrom: Use `ServerManager.SERVER_ID_ACTIVE` instead of `-1`

app/src/main/kotlin/io/homeassistant/companion/android/webview/insecure/BlockInsecureScreen.kt
```diff
@@ -0,0 +1,219 @@
...
+            horizontalAlignment = Alignment.CenterHorizontally,
+        ) {
+            Column(
+                modifier = Modifier
+                    .fillMaxSize()
+                    .verticalScroll(rememberScrollState())
+                    .padding(horizontal = HADimens.SPACE4),
+                horizontalAlignment = Alignment.CenterHorizontally,
+                verticalArrangement = Arrangement.spacedBy(HADimens.SPACE6),
+            ) {
```
jpelgrom: Why are there two nested columns here instead of just one?

app/src/main/kotlin/io/homeassistant/companion/android/webview/insecure/BlockInsecureScreen.kt
```diff
@@ -0,0 +1,219 @@
...
+    )
+}
+
+@Composable
+private fun ColumnScope.Header() {
+    Icon(
+        modifier = Modifier
+            .padding(all = 20.dp)
+            .size(120.dp),
+        imageVector = Icons.Default.Lock,
```
jpelgrom: We had some [discussions about this on Discord.](https://discord.com/channels/330944238910963714/1346948551892009101/1443629721756766230) My vote is for `mdi:lock-open-alert` as you're talking about http(s) in the screen and users have been trained for open lock = http, closed lock = https in web browsers, so showing a closed lock because they are using http feels wrong.

build-logic/convention/src/main/kotlin/AndroidComposeConventionPlugin.kt
```diff
@@ -38,7 +38,7 @@ class AndroidComposeConventionPlugin : Plugin<Project> {
                 // Hack until we get the update of the screenshot libray
                 // https://issuetracker.google.com/issues/444048026
                 // 3g is the minimal value for our tests to pass currently
-                maxHeapSize = "3g"
+                maxHeapSize = "4g"
```
jpelgrom, line 41: Don't forget to update the comment above

common/src/main/res/values/strings.xml
```diff
@@ -1472,6 +1472,15 @@
     <string name="location_secure_connection_next">Next</string>
     <string name="location_secure_connection_save">Save</string>
     <string name="location_secure_connection_discard_permission">Location permission denied. Allow access or choose the \'Less secure\' option.</string>
+    <string name="block_insecure_title">Insecure connection blocked</string>
+    <string name="block_insecure_content">This connection is not secure because it uses HTTP. You have chosen to block insecure connections when not at home.</string>
+    <string name="block_insecure_missing_home_setup">The app cannot determine if you are at home because no home network has been configured.</string>
```
jpelgrom: Not 100% sure about this: _if_ is for conditions, but "a configured home network" is the condition here

```suggestion
    <string name="block_insecure_missing_home_setup">The app cannot determine whether you are at home because no home network has been configured.</string>
```

common/src/main/res/values/strings.xml
```diff
@@ -1472,6 +1472,15 @@
...
+    <string name="block_insecure_missing_location">Location is required to detect your home network. Android requires this because network information can be used to infer your location.</string>
```
```suggestion
    <string name="block_insecure_missing_location">Location access is required to detect your home network. Android requires this because network information can be used to infer your location.</string>
```

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1410,6 +1416,46 @@ class WebViewActivity :
...
 
+    /**
+     * Make sure we only load the url if the securityLevel and current states allow it.
+     */
+    private suspend fun secureLoadUrl(url: String) {
+        fun loadAndWait() {
+            webView.loadUrl(url)
+            isInsecureState = false
+            waitForConnection()
+        }
```
jpelgrom, line 1443: Not a fan of putting functions inside functions, easy to lose track of them

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1410,6 +1416,46 @@ class WebViewActivity :
...
+
+        if (url.startsWith("https")) {
+            loadAndWait()
+            return
+        }
+
+        val allowInsecureConnection = serverManager.integrationRepository(
+            presenter.getActiveServer(),
+        ).getAllowInsecureConnection()
```
jpelgrom: I think this should be exposed via the presenter instead, the activity should try to avoid using repositories directly

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: I don't want this one to be outside of the scope of this function. This should be used only when the function is small to avoid repetition.

app/src/main/kotlin/io/homeassistant/companion/android/webview/insecure/BlockInsecureFragment.kt
TimoPtr: This should never happens since the Fragment can only be created through the `newInstance`. I really hope that we can get rid of this fragments at some point to avoid checking this states that are only there because of the framework we use.

app/src/main/kotlin/io/homeassistant/companion/android/webview/insecure/BlockInsecureScreen.kt
TimoPtr: That's me doing something over multiple days sorry ... removed.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: I had the same feelings but we are already calling `getServer` `removeServer` that we shouldn't. 
I'm going to move this to the presenter in any case.

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
jpelgrom: You sent me a DM on Discord about creating a new `UrlHelper` class for this one. Feel free to experiment but please comment here once you decided whether to introduce that or keep this.

PR #6115, opened 2025-12-01
Rework OnboardApp to adjust to new architecture

TimoPtr:
## Summary

This PR cleans up the `OnboardApp` class and keep it only for the wear onboarding since it is not needed for the rest while removing all the useless parameters.

## Checklist

## Any other notes

I tested the onboarding on TV, mobile and wear.


app/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
```diff
@@ -172,8 +172,13 @@ internal class NameYourDeviceViewModel @VisibleForTesting constructor(
                     messagingTokenProvider(),
                 ),
             )
-            return serverManager.convertTemporaryServer(tempServerId)
+            val serverId = serverManager.convertTemporaryServer(tempServerId)
                 ?: throw IllegalStateException("Server still temporary")
+
+            // Active the newly added server
+            serverManager.activateServer(serverId)
```
jpelgrom, line 164: I don't think this is for Wear, and it's also a behavior change I don't fully understand the reasoning behind. Can you elaborate / pull it out into another PR?

TimoPtr: It was called in the results of the onboarding after adding a server. If you look at the old code activity result in settings you would see that. I moved it here so it replicates the same behavior as before.

jpelgrom: For app usage it may replicate the same usage, but now it also tries to active the Wear server on the phone that started onboarding it, no?

TimoPtr: It's probably not clear enough and would deserve a comment in the nameyourweardevice package that actually it uses the screen but not the view model actually.

So no it doesn't activate the server on the wear onboarding.

jpelgrom: Yes please add a comment as that seems like something you may easily miss like I did

TimoPtr: Done I didn't add it in the function but on the screen usage and the viewmodel itself.

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourweardevice/navigation/NameYourWearDeviceNavigation.kt
```diff
@@ -46,6 +46,10 @@ internal fun NavGraphBuilder.nameYourWearDeviceScreen(
         // If the logic of the screen evolve we should consider introducing a viewModel instead.
         var deviceName by rememberSaveable { mutableStateOf(route.defaultDeviceName) }
 
+        // Unlike phone/tablet onboarding, we don't use NameYourDeviceViewModel here.
+        // In Wear onboarding, the phone acts as a proxy: it collects the device name and auth code,
```
jpelgrom:
```suggestion
        // In Wear onboarding, the phone/tablet acts as a proxy: it collects the device name and auth code,
```

PR #6131, opened 2025-12-05
Refactor connection state management

TimoPtr:
## Summary

This PR introduces a major refactoring of how the app manages server connection state, along with the block insecure HTTP connections when not on the home network.

   **Key changes:**

   1. **New `ServerConnectionStateProvider`** - A reactive provider that determines the appropriate URL based on current network conditions (internal vs external), manages security checks, and emits updates when network conditions change. This replaces the scattered URL resolution logic previously in
   `ServerConnectionInfo`.

   2. **Refactored `ServerConnectionInfo`** - Moved URL parsing logic to lazy cached `HttpUrl` properties and extracted connection state logic to `ServerConnectionStateProvider`. And it now contains `allowInsecureConnection` instead of the local storage, it is more relevant and make the logic easier to pick a URL.

   3. Replace all the usage of `getUrl` with the new flow when it makes sense. Try to uniform how we decide to pick a URL instead of doing differently in different places (even in the webview itself)

## Checklist

## Any other notes

Ui is not yet fully ready and tested.
Some TODO are still in the code and might need to be address otherwise we need to removed them.
This is a major change and might introduce regressions.

I've changed how we open the integration after Improv suggest since I wanted to remove the getUrl and it was easier to simply use NavigateTo and improve the experience of the user.

I've used Claude to help me generate an extensive set of tests since it's a critical piece of the app I want it to be fully tested. It introduce the usage of `nested` test and it is quite nice when running inside the IDE to create categories. 
I've on purpose didn't make all the tests Parametrized to try to keep the name meaningful as much as possible. Claude was going to hard on this and making tests unreadable.

**Important:** This PR is the first one of the refactor. A second one is needed to change the usage of `getApiUrls` since today it doesn't take in account the allowInsecure flag. Another one can be made later once everything is merged to handle more state while getting the URL to actually handle better when network in unavailable for the WebView and WebSocket for instance.

I don't like the name `UrlState` if you have a better name please share it.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
@@ -24,6 +24,10 @@ import io.homeassistant.companion.android.common.util.GestureDirection
 import io.homeassistant.companion.android.common.util.HAGesture
 import io.homeassistant.companion.android.common.util.cancelOnLifecycle
 import io.homeassistant.companion.android.database.server.ServerConnectionInfo
+import io.homeassistant.companion.android.database.settings.SensorUpdateFrequencySetting
+import io.homeassistant.companion.android.database.settings.Setting
+import io.homeassistant.companion.android.database.settings.SettingsDao
+import io.homeassistant.companion.android.database.settings.WebsocketSetting
 import io.homeassistant.companion.android.improv.ImprovRepository
 import io.homeassistant.companion.android.matter.MatterManager
 import io.homeassistant.companion.android.thread.ThreadManager
--- a/app/src/test/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImplTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImplTest.kt
@@ -15,6 +15,7 @@ import io.homeassistant.companion.android.common.data.servers.ServerManager
 import io.homeassistant.companion.android.common.data.servers.UrlState
 import io.homeassistant.companion.android.database.server.Server
 import io.homeassistant.companion.android.database.server.ServerConnectionInfo
+import io.homeassistant.companion.android.database.settings.SettingsDao
 import io.homeassistant.companion.android.improv.ImprovRepository
 import io.homeassistant.companion.android.matter.MatterManager
 import io.homeassistant.companion.android.testing.unit.ConsoleLogExtension
@@ -108,6 +109,7 @@ class WebViewPresenterImplTest {
     private val prefsRepository: PrefsRepository = mockk(relaxed = true)
     private val matterManager: MatterManager = mockk(relaxed = true)
     private val threadManager: ThreadManager = mockk(relaxed = true)
+    private val settingsDao: SettingsDao = mockk(relaxed = true)
     private val authenticationRepository: AuthenticationRepository = mockk(relaxed = true)
     private val connectionStateProvider: ServerConnectionStateProvider = mockk(relaxed = true)
 
@@ -143,6 +145,7 @@ class WebViewPresenterImplTest {
             prefsRepository = prefsRepository,
             matterUseCase = matterManager,
             threadUseCase = threadManager,
+            settingsDao = settingsDao,
         )
         return presenter
     }
--- a/app/lint-baseline.xml
+++ b/app/lint-baseline.xml
@@ -85,7 +85,7 @@
         errorLine2="                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt"
-            line="1053"
+            line="1055"
             column="17"/>
     </issue>
 
@@ -129,7 +129,7 @@
         errorLine2="                             ~~~~~~~~~~~~~">
         <location
             file="src/main/kotlin/io/homeassistant/companion/android/controls/HaControlsProviderService.kt"
-            line="46"
+            line="47"
             column="30"/>
     </issue>
 
--- a/automotive/lint-baseline.xml
+++ b/automotive/lint-baseline.xml
@@ -85,7 +85,7 @@
         errorLine2="                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/notifications/MessagingManager.kt"
-            line="1053"
+            line="1055"
             column="17"/>
     </issue>
 
@@ -151,7 +151,7 @@
         errorLine2="                             ~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/controls/HaControlsProviderService.kt"
-            line="46"
+            line="47"
             column="30"/>
     </issue>
 
@@ -1218,7 +1218,7 @@
         errorLine2="            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/notifications/MessagingManager.kt"
-            line="619"
+            line="621"
             column="13"/>
     </issue>
 
@@ -1229,7 +1229,7 @@
         errorLine2="                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/notifications/MessagingManager.kt"
-            line="985"
+            line="987"
             column="17"/>
     </issue>
 
@@ -1240,7 +1240,7 @@
         errorLine2="            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/notifications/MessagingManager.kt"
-            line="1040"
+            line="1042"
             column="13"/>
     </issue>
 
@@ -1251,7 +1251,7 @@
         errorLine2="                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/notifications/MessagingManager.kt"
-            line="1086"
+            line="1088"
             column="21"/>
     </issue>
 
@@ -1262,7 +1262,7 @@
         errorLine2="                                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/notifications/MessagingManager.kt"
-            line="1117"
+            line="1119"
             column="50"/>
     </issue>
 
@@ -1273,7 +1273,7 @@
         errorLine2="            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/notifications/MessagingManager.kt"
-            line="1667"
+            line="1684"
             column="13"/>
     </issue>
 
@@ -2109,7 +2109,7 @@
         errorLine2="            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/webview/WebViewActivity.kt"
-            line="289"
+            line="310"
             column="13"/>
     </issue>
 
@@ -2120,7 +2120,7 @@
         errorLine2="                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/webview/WebViewActivity.kt"
-            line="565"
+            line="586"
             column="21"/>
     </issue>
 
@@ -2131,7 +2131,7 @@
         errorLine2="            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/webview/WebViewActivity.kt"
-            line="675"
+            line="696"
             column="13"/>
     </issue>
 
@@ -2142,7 +2142,7 @@
         errorLine2="                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/webview/WebViewActivity.kt"
-            line="1274"
+            line="1299"
             column="21"/>
     </issue>
 
@@ -2153,7 +2153,7 @@
         errorLine2="                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/webview/WebViewActivity.kt"
-            line="1393"
+            line="1418"
             column="17"/>
     </issue>
 
@@ -2164,7 +2164,7 @@
         errorLine2="            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/websocket/WebsocketManager.kt"
-            line="231"
+            line="230"
             column="13"/>
     </issue>
 
@@ -2175,7 +2175,7 @@
         errorLine2="                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="${:automotive*fullDebug*MAIN*sourceProvider*0*javaDir*5}/io/homeassistant/companion/android/websocket/WebsocketManager.kt"
-            line="286"
+            line="285"
             column="17"/>
     </issue>
 
--- a/common/lint-baseline.xml
+++ b/common/lint-baseline.xml
@@ -8,7 +8,7 @@
         errorLine2="                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="src/main/kotlin/io/homeassistant/companion/android/database/AppDatabase.kt"
-            line="1034"
+            line="1035"
             column="17"/>
     </issue>
 
@@ -19,7 +19,7 @@
         errorLine2="                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="src/main/kotlin/io/homeassistant/companion/android/database/AppDatabase.kt"
-            line="1034"
+            line="1035"
             column="17"/>
     </issue>
 
@@ -228,7 +228,7 @@
         errorLine2="            ~~~~~~~~~~~~~~~~">
         <location
             file="src/main/kotlin/io/homeassistant/companion/android/common/data/authentication/impl/AuthenticationRepositoryImpl.kt"
-            line="77"
+            line="81"
             column="13"/>
     </issue>
 
@@ -250,7 +250,7 @@
         errorLine2="                                                                ~~~~~~~~~~~~~~~~">
         <location
             file="src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt"
-            line="233"
+            line="236"
             column="65"/>
     </issue>
 
@@ -261,7 +261,7 @@
         errorLine2="                                                                    ~~~~~~~~~~~~~~~~">
         <location
             file="src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt"
-            line="273"
+            line="276"
             column="69"/>
     </issue>
 
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
@@ -722,7 +722,7 @@ class IntegrationRepositoryImpl @AssistedInject constructor(
         localStorage.putBoolean("${serverId}_$PREF_ASK_NOTIFICATION_PERMISSION", shouldAsk)
     }
 
-    private suspend fun Server.callWebhookOnUrls(
+    private suspend fun callWebhookOnUrls(
         request: IntegrationRequest,
         onSuccess: suspend (response: Response<ResponseBody>) -> Unit = {},
         isValidResponse: (response: Response<ResponseBody>) -> Boolean = { response -> response.isSuccessful },


app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -85,64 +88,93 @@ class WebViewPresenterImpl @Inject constructor(
...
-                    url = url.toString().toUri()
-                        .buildUpon()
-                        .appendQueryParameter("external_auth", "1")
-                        .build()
-                        .toString(),
-                    keepHistory = oldUrlForServer == urlForServer,
-                    openInApp = url?.baseIsEqual(baseUrl) ?: false,
-                )
+        var shouldConsumePath = path != null
+        urlFlowJob = CoroutineScope(currentCoroutineContext()).launch {
```
TimoPtr: This needs to be break down into multiple pieces and it needs unit tests.

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
```diff
@@ -44,149 +58,101 @@ data class ServerConnectionInfo(
...
-            context,
-            Manifest.permission.ACCESS_FINE_LOCATION,
-        ) == PackageManager.PERMISSION_GRANTED
-        val isLocationEnabled = DisabledLocationHandler.isLocationEnabled(context)
-
-        return SecurityStatus.Insecure(
-            missingHomeSetup = missingHomeSetup,
-            missingLocation = !hasLocationPermission || !isLocationEnabled,
-        )
+    fun isKnownUrl(url: String): Boolean {
```
TimoPtr, line 150: We might want to move this out of here.

jpelgrom:
> Ui is not yet fully ready and tested

What UI are you referring to? The PR description makes it sound like there are no UI changes.

TimoPtr:
> > Ui is not yet fully ready and tested
> 
> What UI are you referring to? The PR description makes it sound like there are no UI changes.

The base of this PR is the PR of the blocking UI. And it needs some adjustments on top of the fact that I didn't test properly if it works as it should outside of the happy path.

TimoPtr:
https://github.com/user-attachments/assets/11a6586a-79da-4f2e-a18a-548da2e8cdcb

The improv flow using navigate instead of opening the URL again.

app/src/main/kotlin/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
```diff
@@ -301,7 +302,10 @@ class HaControlsProviderService : ControlsProviderService() {
             }
         }
         val entities = mutableMapOf<String, Entity>()
-        val baseUrl = serverManager.getServer(serverId)?.connection?.getUrl()?.toString()?.removeSuffix("/") ?: ""
+        // TODO see what we do about this one we should get the url only when we need it not prefetching it
```
jpelgrom: This is for the camera control. It could be fetched only if the entities list includes a domain that uses that control and otherwise remain null.

app/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -1257,10 +1259,15 @@ class MessagingManager @Inject constructor(
...
-                builder.setLargeIcon(bitmap)
+            val urlState = serverManager.connectionStateProvider(serverId).urlFlow().first()
+            if (urlState is UrlState.HasUrl) {
+                val url = UrlUtil.handle(urlState.url, dataIcon)
+                val bitmap = getImageBitmap(serverId, url, !UrlUtil.isAbsoluteUrl(dataIcon))
+                if (bitmap != null) {
+                    builder.setLargeIcon(bitmap)
+                }
+            } else {
+                Timber.w("Not fetching image since URL is insecure")
```
jpelgrom:
```suggestion
                Timber.w("Not fetching icon since URL is insecure")
```

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
```diff
@@ -144,7 +144,15 @@ internal class NameYourDeviceViewModel @VisibleForTesting constructor(
             // Until https://github.com/home-assistant/android/issues/6005 is not addressed this is going to fail.
```
jpelgrom: By switching it to being info on the server class, we can actually avoid the issue in this case and set the value in `addServer()` when we insert it into the database.

app/src/main/kotlin/io/homeassistant/companion/android/settings/url/ExternalUrlViewModel.kt
```diff
@@ -37,9 +37,9 @@ class ExternalUrlViewModel @Inject constructor(
         state.get<Int>(ExternalUrlFragment.EXTRA_SERVER)?.let { serverId = it }
         viewModelScope.launch {
             serverManager.getServer(serverId)?.let {
-                canUseCloud = it.connection.canUseCloud()
+                canUseCloud = it.connection.canUseCloud
                 useCloud = it.connection.useCloud
-                externalUrl = it.connection.getUrl(isInternal = false, force = true).toString()
+                externalUrl = serverManager.getServer(it.id)?.connection?.externalUrl ?: ""
```
jpelgrom:
```suggestion
                externalUrl = it.connection.externalUrl
```

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1388,77 +1402,63 @@ class WebViewActivity :
...
         ) { _, _ ->
             Timber.d("Security level screen exited by user, proceeding with URL loading")
             supportFragmentManager.clearFragmentResultListener(ConnectionSecurityLevelFragment.RESULT_KEY)
 
-            if (::loadedUrl.isInitialized) {
-                lifecycleScope.launch {
-                    secureLoadUrl(loadedUrl)
-                }
+            // We don't care of the result we always assume that the fragment bash been closed and we should
+            // show it again.
```
jpelgrom: I don't follow your comment here / what are you trying to tell other than the fragment has been closed?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1388,77 +1402,63 @@ class WebViewActivity :
...
+            connectionSecurityLevelFragmentClosed[serverId] = true
+
+            lifecycleScope.launch {
+                // Trigger a reload of the URL to trigger to trigger loadUrl in the view,
+                // it is important to use the presenter to apply the potential changes
+                // made in the fragment.
```
jpelgrom:
```suggestion
                // Trigger a reload of the URL via the presenter to trigger loadUrl in the view.
                // The presenter will apply the potential changes made in the fragment.
```

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1475,20 +1475,30 @@ class WebViewActivity :
             .commit()
     }
 
+    override fun showBlockInsecure(serverId: Int, missingHomeSetup: Boolean, missingLocation: Boolean) {
+        showBlockInsecureFragment(serverId, missingHomeSetup, missingLocation)
+    }
```
jpelgrom: Why have a different function here if it's forwarding to `showBlockInsecureFragment` without any additions/defaults?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1661,20 +1672,16 @@ class WebViewActivity :
...
-                        failedConnection = if (buttonRefreshesInternal) "internal" else "external"
-                        if (url != null) {
-                            loadUrl(url = url, keepHistory = true, openInApp = true)
-                        } else {
-                            waitForConnection()
-                        }
+                        // TODO Once we do this it's going to keep the override until load is called again
+                        //  in onWindowFocusChanged that should happen right after the dialog is canceled.
+                        //  Then it might fail again if the user internal URL is always failing but the user is
+                        //  at home. (I guess)
```
jpelgrom: I think this needs to be handled before merging, but haven't tested it yet

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -2040,18 +2050,15 @@ class WebViewActivity :
...
-                            loadUrl(url = url, keepHistory = true, openInApp = true)
+                        // Using My links since config/integrations/dashboard/add doesn't open the dialog
+                        // when already on the integrations page.
+                        val path = "/_my_redirect/config_flow_start?domain=$improvDomain"
+                        val version = serverManager.getServer(presenter.getActiveServer())?.version
+
+                        if (NavigateTo.isAvailable(version)) {
+                            sendExternalBusMessage(NavigateTo(path))
+                        } else {
+                            presenter.load(lifecycle, path)
```
jpelgrom: Does this need to be in this PR?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -85,68 +87,180 @@ class WebViewPresenterImpl @Inject constructor(
...
+        if (shouldConsumePath) {
+            loadInitialUrl(
+                baseUrl = urlState.url,
+                path = path,
+                isNewServer = isNewServer,
+            )
+        } else {
+            // TODO not sure if we need to handle the path here, since if we end up into this
+            //  we would have already consumed the path and we are changing to another base URL
+            //  like going from internal to external.
```
jpelgrom: I would argue the path should be ignored in that case, as path is only for initial load of the activity, but any navigation after that will change it which we don't track.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -85,68 +87,180 @@ class WebViewPresenterImpl @Inject constructor(
...
+        } ?: Timber.w("Url is null")
+    }
+
+    /**
+     * Loads a server URL directly without path modification.
+     *
+     * @param serverUrl the URL to load
+     * @param isNewServer whether this is a new server (affects history behavior)
+     */
+    private suspend fun loadServerUrl(serverUrl: URL?, isNewServer: Boolean) {
```
jpelgrom: Why is this a different function and not simply `loadInitialUrl` with a null path?

app/src/main/kotlin/io/homeassistant/companion/android/webview/insecure/BlockInsecureScreen.kt
```diff
@@ -107,12 +114,26 @@ internal fun BlockInsecureScreen(
...
+                onClick = {
+                    scope.launch {
+                        rotation.animateTo(
+                            targetValue = rotation.value - 360f,
+                            animationSpec = tween(durationMillis = ROTATION_ANIMATION_DURATION_MS),
+                        )
+                    }
+                    onRetry()
+                },
+            ) {
```
jpelgrom, line 132: This is nice but is flavor that can be in the initial PR that adds the screen

app/src/main/kotlin/io/homeassistant/companion/android/widgets/camera/CameraWidget.kt
```diff
@@ -127,10 +129,18 @@ class CameraWidget : AppWidgetProvider() {
         if (widget != null) {
             try {
                 val entityPictureUrl = retrieveCameraImageUrl(widget.serverId, widget.entityId)
-                val baseUrl = serverManager.getServer(
+
+                // TODO we should change the look of the widget in that case
```
jpelgrom: Just use the error state for the widget?

app/src/test/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImplTest.kt
```diff
@@ -0,0 +1,525 @@
...
+        backgroundScope.launch {
+            presenter.load(lifecycle, path = "entityId:light.living_room", isInternalOverride = null)
+        }
+
+        val urlSlot = slot<Uri>()
+        verify { webView.loadUrl(capture(urlSlot), any(), any()) }
+
+        // entityId paths should be ignored and base URL should be loaded
+        assertTrue(urlSlot.captured.toString().startsWith("https://example.com"))
+        assertTrue(!urlSlot.captured.toString().contains("entityId"))
```
jpelgrom: Why not use `assertFalse`?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/SecurityInfo.kt
```diff
@@ -0,0 +1,4 @@
+package io.homeassistant.companion.android.common.data.servers
+
+// This should never be persisted and we should always check the info before making a call
+class SecurityInfo(val isOnInternalNetwork: Boolean, val hasHomeSetup: Boolean, val locationEnabled: Boolean)
```
jpelgrom: This is a state, why not call it SecurityState so it is more obvious it shouldn't be persisted?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProvider.kt
```diff
@@ -0,0 +1,145 @@
...
+sealed interface UrlState {
+    /**
+     * A URL is available and safe to use.
+     *
+     * This state is emitted when:
+     * - The URL uses HTTPS, or
+     * - The device is on an internal network (HTTP is acceptable internally), or
+     * - The user has explicitly allowed insecure connections
+     *
+     * @property url the server URL to use, may be `null` if no valid URL is configured
```
jpelgrom, line 38: > A URL is available

> may be `null` if no valid URL is configured

This seems like a contradiction :)

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProvider.kt
```diff
@@ -0,0 +1,145 @@
...
+ * This provider determines the appropriate URL to use based on the current network state
+ * (internal vs external), manages security checks for connections, and emits reactive
+ * updates when network conditions change.
+ *
+ * @see ServerConnectionStateProviderFactory for creating instances
+ */
+interface ServerConnectionStateProvider {
+
+    /**
+     * Indicates if the device's current connection should be treated as internal for this server.
```
jpelgrom: A device connection should be treated as on the home network. When on the home network, the internal URL should be used. There is never an "internal network".

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProvider.kt
```diff
@@ -0,0 +1,145 @@
...
+    suspend fun getExternalUrl(): URL?
+
+    /**
+     * Gets all API URLs for webhook communication, ordered by priority.
+     *
+     * The order depends on whether the device is on an internal network:
+     * - Internal: internal URL, then cloudhook, then external
+     * - External: cloudhook first (if available), then external
+     *
+     * Each URL has `/api/webhook/{webhookId}` appended for webhook calls.
```
jpelgrom: Not for cloudhooks. This also seems unnecessary as your first sentence already says they are URLs for webhook communication.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImpl.kt
```diff
@@ -0,0 +1,262 @@
...
+        if (!connection.isKnownUrl(url)) return false
+
+        // URL is HTTP and belongs to this server - check if it's safe
+        val allowInsecure = connection.allowInsecureConnection ?: true
+        return allowInsecure || isInternal(requiresUrl = false)
+    }
+
+    // If permission is denied then the process restart so initial state is going to be updated
+
+    // TODO this needs to be shared within this class to avoid multiple registration
```
jpelgrom: TODO

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImpl.kt
```diff
@@ -0,0 +1,262 @@
...
+    private fun observeConnectionInfoChanges(): Flow<ServerConnectionInfo> {
+        return serverDao.observeServer(serverId)
+            .mapNotNull { it?.connection }
+            .distinctUntilChanged()
+    }
+
+    // TODO this could be cache too, the idea would be to have better logging experience regarding
+    //  which URL is currently being used
```
jpelgrom, line 148: I don't see caching yet so this would need an issue for tracking (even though I'm not sure currently what logging experience we need)

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManagerImpl.kt
```diff
@@ -51,6 +52,7 @@ class ServerManagerImpl @Inject constructor(
     private val authenticationRepos = mutableMapOf<Int, AuthenticationRepository>()
     private val integrationRepos = mutableMapOf<Int, IntegrationRepository>()
     private val webSocketRepos = mutableMapOf<Int, WebSocketRepository>()
+    private val connectionRepos = mutableMapOf<Int, ServerConnectionStateProvider>()
```
jpelgrom: So... why did you not call it a ConnectionRepository?

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
```diff
@@ -1,26 +1,40 @@
...
 
+/**
+ * Holds connection configuration for a Home Assistant server.
+ *
+ * This data class is embedded in [Server] and stored in Room database. It contains all URL
+ * endpoints and network detection settings used to connect to the server.
+ *
+ * @property externalUrl The primary URL used to connect to the server from outside the home network.
+ * @property internalUrl Optional URL used when the device is on the home network.
+ * @property cloudUrl Optional Home Assistant Cloud remote UI URL. When available and [useCloud] is enabled.
```
jpelgrom: The URL is always stored, also if `useCloud` is set to `false`. I think it's also good to clarify this is the value that automatically updates (from get_config), not something the user entered.

```suggestion
 * @property cloudUrl Optional Home Assistant Cloud remote UI URL, automatically updated from the server.
```

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
```diff
@@ -1,26 +1,40 @@
...
+ *   home network.
+ * @property internalSsids List of Wi-Fi network names (SSIDs) that indicate the device is on
+ *   the home network.
+ * @property internalEthernet When `true`, Ethernet connections are treated as being on the home
+ *   network. `null` means the user hasn't configured this setting.
+ * @property internalVpn When `true`, VPN connections are treated as being on the home network.
+ *   `null` means the user hasn't configured this setting.
+ * @property prioritizeInternal When `true`, always try the internal URL first regardless of
+ *   detected network state.
+ * @property allowInsecureConnection Controls whether HTTP (non-HTTPS) connections are permitted
```
jpelgrom: Insecure is when not HTTPS. We should make sure to consistently check for it that way.

```suggestion
 * @property allowInsecureConnection Controls whether non-HTTPS connections are permitted
```

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
```diff
@@ -1,26 +1,40 @@
...
+ *   when not on the home network. `null` means the user hasn't set their preference yet (legacy
+ *   installs), `true` allows insecure connections, `false` blocks them and shows a warning.
```
jpelgrom: Blocking should be handled by classes using this value. Showing a warning is the UI behavior in one specific location.

```suggestion
 *   installs), `true` allows insecure connections, when `false` insecure connections should be blocked.
```

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
```diff
@@ -44,149 +58,110 @@ data class ServerConnectionInfo(
...
-    fun isRegistered(): Boolean = getApiUrls().isNotEmpty()
-
-    fun getApiUrls(): List<URL> {
-        // If we don't have a webhook id we don't have any urls.
-        if (webhookId.isNullOrBlank()) {
-            return emptyList()
-        }
+    /**
+     * Keep the parsing of the URLs from String to HTTPUrl to avoid doing it multiple times
+     */
```
jpelgrom: This doesn't describe the value so a normal comment

```suggestion
    // Keep the parsing of the URLs from String to HttpUrl to avoid doing it multiple times
```

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
```diff
@@ -44,149 +58,110 @@ data class ServerConnectionInfo(
...
+            externalHttpUrl,
+            internalHttpUrl,
+            cloudHttpUrl,
+            cloudhookHttpUrl,
+        )
+    }
 
-        return retVal.filterNotNull()
+    /**
+     * Indicates whether at least one URL (external, internal, cloud, or cloudhook) is configured
```
jpelgrom: Why would you want cloud OR cloudhook?

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerDao.kt
```diff
@@ -22,6 +22,9 @@ interface ServerDao {
     @Query("SELECT * FROM servers ORDER BY `list_order` ASC")
     fun getAllFlow(): Flow<List<Server>>
 
+    @Query("SELECT * FROM servers WHERE id = :serverId")
+    fun observeServer(serverId: Int): Flow<Server?>
```
jpelgrom: For consistency I would use this (and move it up before `getAll`)

```suggestion
    fun getFlow(serverId: Int): Flow<Server?>
```

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt
```diff
@@ -168,3 +168,38 @@ private fun InetAddress.isPrivateOrLocal(): Boolean {
...
+        port == other.port
+}
+
+/**
+ * Checks if this Uri has the same base (scheme, host, and port) as the other Uri.
+ *
+ * @param other the Uri to compare against
+ * @return `true` if both URIs have the same scheme, host, and port
+ */
+fun Uri.hasSameBase(other: Uri?): Boolean {
```
jpelgrom: Why is it called base here, but origin for `HttpUrl.hasSameOrigin`?

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt
```diff
@@ -168,3 +168,38 @@ private fun InetAddress.isPrivateOrLocal(): Boolean {
...
+        host.equals(other.host, ignoreCase = true) &&
+        port == other.port
+}
+
+/**
+ * Checks if this Uri has a meaningful path (not empty, not just "/").
+ *
+ * @return `true` if the Uri has a path that is not blank and not just "/"
+ */
+fun Uri.hasMeaningfulPath(): Boolean {
```
jpelgrom: You need to check the KDoc here to understand what is "meaningful". I don't really like it and would prefer something more intuitive like Kotlins `isNullOrBlank`, but `Uri.pathIsNullOrBlankOrSlash()` doesn't quite have the same ring to it :)

common/src/test/kotlin/io/homeassistant/companion/android/util/UrlUtilTest.kt
```diff
@@ -172,4 +179,29 @@ class UrlUtilTest {
...
+        "https://example.com/path, https://example.com, true",
+        "https://example.com?query=1, https://example.com, true",
+        "https://example.com:443, https://example.com, true",
+        "http://example.com:80, http://example.com, true",
+        "https://example.com:8123, https://example.com:8123, true",
+        "https://example.com, https://other.com, false",
+        "https://example.com, http://example.com, false",
+        "https://example.com:8123, https://example.com:8124, false",
+        "https://example.com, https://example.com.evil.com, false",
+        "https://sub.example.com, https://example.com, false",
```
jpelgrom, line 195: This is a different list from UriExtensionsTest, why?

common/src/test/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImplTest.kt
```diff
@@ -0,0 +1,756 @@
...
+
+        @Test
+        fun `Given SSID configured but not using WiFi when calling isInternal then returns false`() = runTest {
+            val provider = createServerConnectionStateProvider(
+                externalUrl = "https://external.example.com",
+                internalUrl = "http://192.168.1.1:8123",
+                internalSsids = listOf("HomeWiFi"),
+            )
+            every { wifiHelper.isUsingSpecificWifi(listOf("HomeWiFi")) } returns true
+            every { wifiHelper.isUsingWifi() } returns false
```
jpelgrom, line 187: This combination isn't something that could happen, I think.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImplTest.kt
```diff
@@ -0,0 +1,756 @@
...
+
+            assertFalse(provider.isInternal())
+        }
+    }
+
+    @Nested
+    inner class GetSecurityInfo {
+
+        @Test
+        fun `Given device is on internal network via ethernet then canMakePlainTextCallSecurely returns true`() = runTest {
```
jpelgrom: Stale title

common/src/test/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImplTest.kt
```diff
@@ -0,0 +1,756 @@
...
+            every { DisabledLocationHandler.isLocationEnabled(context) } returns false
+
+            val result = repository.getSecurityInfo()
+
+            assertFalse(result.isOnInternalNetwork)
+            assertFalse(result.locationEnabled)
+        }
+
+        @Test
+        fun `Given null internalVpn and internalEthernet then hasHomeSetup is false`() = runTest {
```
jpelgrom:
```suggestion
        fun `Given no internalSsids and null internalVpn and internalEthernet then hasHomeSetup is false`() = runTest {
```

common/src/test/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfoTest.kt
```diff
@@ -1,335 +1,349 @@
...
-            internalEthernet = true,
-        )
-        every { networkHelper.isUsingEthernet() } returns true
-        val httpUrl = "http://192.168.1.1:8123"
+            assertTrue(connection.hasHomeNetworkSetup)
+        }
 
-        val result = serverInfo.currentSecurityStatusForUrl(context, httpUrl)
+        @Test
+        fun `Given internalVpn disabled then hasHomeNetworkSetup is false`() {
```
jpelgrom: Similar to the previous comment I think this one needs to be more complete as it is misleading otherwise

```suggestion
        fun `Given no internalSsids and internalVpn disabled and internalEthernet null then hasHomeNetworkSetup is false`() {
```

common/src/test/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfoTest.kt
```diff
@@ -1,335 +1,349 @@
...
+            )
 
-        val result = serverInfo.currentSecurityStatusForUrl(context, httpUrl)
+            assertTrue(connection.hasHomeNetworkSetup)
+        }
 
-        assertInstanceOf(SecurityStatus.Secure::class.java, result)
-    }
+        @Test
+        fun `Given internalEthernet disabled then hasHomeNetworkSetup is false`() {
```
jpelgrom:
```suggestion
        fun `Given no internalSsids and internalVpn null and internalEthernet disabled then hasHomeNetworkSetup is false`() {
```

jpelgrom (commented):
There's a lot (too much, probably) happening here, I might come back with more comments later

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: Well I though about it indeed and I decided to keep it here otherwise it would have cause a conflict when merging because it replaces the getUrl(). Since it's going to already complicated to merge I took a shortcut here since the change is not that big.

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/NameYourDeviceViewModel.kt
TimoPtr: Good catch 💪🏻

app/src/main/kotlin/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
TimoPtr: But when fectching it should only fetch when it can. I'm not sure to fully follow how it is used in the HAControls. But I feel like the retrieval of the URL should happen when we actually need it not before.

app/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
TimoPtr: I've replaced `insecure` with `unavailable` because at the moment we are going to extend to no network the log is going to be wrong.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: Sorry It should be `shouldn't` the idea here is that we don't care about the result the only thing we know is that we don't want to see the fragment again in the scope of this session for this server. I'm going to adjust the comment.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: Leftover 👍🏻 from previous iteration

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: I felt a bit overwhelmed by this logic, what does it means for the rest of the app. Like the widget won't benefits from the switch to internal or external here. If for instance I'm on internal and it fails to load the webview and press go external it won't apply to the widget nor anything else than the webview.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
TimoPtr: Yeah I share the same feeling but I wanted to discuss it here since I was not sure if it was really only for the first load.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
TimoPtr: Leftover of a super complicated design ... simplified.

app/src/main/kotlin/io/homeassistant/companion/android/webview/insecure/BlockInsecureScreen.kt
TimoPtr: Indeed 👍🏻 Gonna move it there.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/camera/CameraWidget.kt
TimoPtr: It should already handle it properly then 👍🏻

app/src/test/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImplTest.kt
TimoPtr: Good catch I've addressed all the usage like this ...

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProvider.kt
TimoPtr: How would you describe this state? It's kinda weird that it can be null but that's our design, even the name HasUrl is wrong.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImpl.kt
TimoPtr: I'm going to tackle that later. I don't think it has a huge impact if we do register multiple time. wdyt?

TimoPtr: https://github.com/home-assistant/android/issues/6147

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImpl.kt
TimoPtr: What I don't like today in our logs is that we are spam with the logs from `getUrl`

`Using internal URL`
`Using cloud / remote UI URL`
...

We should have less of that to be more meaningful.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManagerImpl.kt
TimoPtr: I didn't call it a repository because from my perspective it is not storing any data but you might have a different opinion?

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
TimoPtr: I took a generic approach here. I should probably remove it from the doc?

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt
TimoPtr: Should be origin https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Same-origin_policy I think going to update.

TimoPtr: And also update the condition to test in the same ways

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt
TimoPtr: Went with `hasNonRootPath`

common/src/test/kotlin/io/homeassistant/companion/android/util/UrlUtilTest.kt
TimoPtr: Lemme make it the same

common/src/test/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImplTest.kt
TimoPtr: Even if it in theory cannot happen having a test is not a bad thing I would say.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImplTest.kt
TimoPtr: You mean it should be home network right?

app/src/main/kotlin/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
jpelgrom: This is the only usage:

https://github.com/home-assistant/android/blob/f1f3415d71b65ea6f0c68895fa6be61e283e9626/app/src/main/kotlin/io/homeassistant/companion/android/controls/CameraControl.kt#L37

So if we provide the control (= camera domain) when getting controls for added entities, we need the URL. That seems like a simple check.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: It is a temporary override for the webview as far as I know, but I must admint I've always found it a bit confusing as well. @dshokouhi is the advocate for this feature :)

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProvider.kt
jpelgrom: What determines if the external URL is 'valid'? I'm struggling to see how you would end up in that state and why it should be nullable.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImpl.kt
jpelgrom: Agreed it's a small optimization that's not required at this point, but that wasn't obvious from your code comment

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImpl.kt
jpelgrom: I agree about more meaningful logging, but
- a cache here seems not easily achieved
- it does give us a lot of reference points for when it is fetching the URL and when it changes
- you're adding a lot of debug log lines in this PR as well which also result in clutter...

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManagerImpl.kt
jpelgrom: Fair, it is only a manager/helper (we don't have a good name for that), but the (internal) variable is still calling it a repository 😉

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
jpelgrom: Not sure.

The 'normal' external/internal URLs can be used for both normal operations (opening pages and websocket) and webhooks. Cloud and cloudhook are distinct, so it feels odd to me to put it all in one variable, which I presume is going to be used specifically for one of those two use cases, not both at the same time.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImplTest.kt
jpelgrom: No it was about `canMakePlainTextCallSecurely` - but internal network should also be home network :)

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: Not the biggest fan since this would've been very quick to review, approve and merge to main as a separate PR, but OK. I don't think you can avoid large merge conflicts/rebase with this mega branch you're building anyway.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/authentication/impl/AuthenticationRepositoryImpl.kt
```diff
@@ -34,9 +36,13 @@ class AuthenticationRepositoryImpl @AssistedInject constructor(
         return checkNotNull(serverManager.getServer(serverId)) { "No server found for id $serverId" }
     }
 
+    private suspend fun connectionStateProvider() = serverManager.connectionStateProvider(serverId)
+
     override suspend fun registerAuthorizationCode(authorizationCode: String) {
         val server = server()
-        val url = server.connection.getUrl()?.toHttpUrlOrNull()
+        val url = connectionStateProvider().urlFlow().firstUrlOrNull {
+            "Unavailable URL to register auth code"
```
jpelgrom:
```suggestion
            "No URL available to register auth code"
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/authentication/impl/AuthenticationRepositoryImpl.kt
```diff
@@ -62,12 +68,14 @@ class AuthenticationRepositoryImpl @AssistedInject constructor(
     }
 
     override suspend fun registerRefreshToken(refreshToken: String) {
-        val url = server().connection.getUrl()?.toHttpUrlOrNull()
+        val url = connectionStateProvider().urlFlow().firstUrlOrNull {
+            "Unavailable URL to register refresh token"
```
jpelgrom:
```suggestion
            "No URL available to register refresh token"
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/authentication/impl/AuthenticationRepositoryImpl.kt
```diff
@@ -89,7 +97,9 @@ class AuthenticationRepositoryImpl @AssistedInject constructor(
 
     override suspend fun revokeSession() {
         val server = server()
-        val url = server.connection.getUrl()?.toHttpUrlOrNull()
+        val url = connectionStateProvider().urlFlow().firstUrlOrNull {
+            "Unavailable URL to revoke session"
```
jpelgrom:
```suggestion
            "No URL available to revoke session"
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -174,13 +175,17 @@ internal class WebSocketCoreImpl(
...
+
     override suspend fun connect(): Boolean {
         connectedMutex.withLock {
             if (connection != null && connected.isCompleted) {
                 return !connected.isCancelled
             }
 
-            val url = server()?.connection?.getUrl()
+            val url = connectionStateProvider().urlFlow().firstUrlOrNull {
+                "Unavailable URL not opening WebSocket connection"
```
```suggestion
                "No URL available to open WebSocket connection"
```

app/src/main/kotlin/io/homeassistant/companion/android/controls/HaControlsProviderService.kt
TimoPtr: OK as far as it is only used to build the thumbnail when we create the control and not in an async manner it's fine 👍🏻

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProvider.kt
TimoPtr: It might happen only if the externalUrl is not parsing properly using .toHttpUrlOrNull()

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImpl.kt
TimoPtr: > it does give us a lot of reference points for when it is fetching the URL and when it changes

I would like to make it more obvious by only log it once. But it's not for now :)

Regarding the clutter about my logs I'm cleaning it up a bit to remove duplicate logs.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManagerImpl.kt
TimoPtr: I think I did address the naming to 
```
private val connectionStateProviders = mutableMapOf<Int, ServerConnectionStateProvider>()
```

common/src/test/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImplTest.kt
TimoPtr: Oh you'r right I removed this method 👍🏻 Updating the test name

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
TimoPtr: I'm not sure what to do here TBH. 
- `hasAtLeastOneUrl` could only check external/internal/cloud
- `isRegister` could check all
- `isKnownUrl` could check all

WDYT? I'm not sure if it's a big issue if hasAtLeastOneUrl also test the cloudhook

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
jpelgrom: Why? I don't see an issue with keeping this closely tied to this class.

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
jpelgrom: What is the method used for when called - use the correct URLs depending on that, I'd say. There is/was a difference because of where registering the app's config entry happened, so you could have a URL without a webhook ID, and you needed to continue.

I don't want to make things too complicated but instinctively my answer would be:
- `hasAtLeastOneUrl` is for continuing to launch/WebViewActivity so external/internal/cloud
- `isRegistered` is somewhat inconsistent but originally for webhook-related calls so external/internal/cloudhook
- `isKnownUrl` is new and for loading pages in WebView only so external/internal/cloud

jpelgrom (commented):
I've checked all the code and done tests with various scenarios, and almost everything looks good (although it is so big there's a reasonable chance I've missed something, especially when considering the base branch is also changing). Not returning a URL when insecure will probably cause some unforeseen issues but so far most features seem to fail gracefully.

The only issue I'm consistently seeing is with migration (choose a security level for existing server with http) + granting permissions, it seems to need another refresh for the state after the last permission is granted to pick it up.

https://github.com/user-attachments/assets/7ea4c263-3e5f-4d41-a797-6e3f189bf80c

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -263,7 +265,26 @@ class WebViewActivity :
...
+     * When set, this function is passed to [WebViewPresenter.load] to override the automatic
+     * internal/external URL detection. This is used when the user explicitly requests to refresh
+     * using the internal or external URL from the error dialog (e.g., after a connection failure).
+     *
+     * The override persists for the activity's lifecycle so that subsequent loads
+     * continue to use the user's preference.
+     *
+     * Keep in mind that it only applies to the Webview not for anything else like
+     * background sync, widgets, ... It is used as a temporary fix for the user to
+     * access his instance, but something might need to be fixed on user side.
```
jpelgrom: Don't assume gender + the app uses server instead of instance

```suggestion
     * access their server, but something might need to be fixed on user side.
```

PR #6141, opened 2025-12-08
Collect more information on crash (thread dump and view hierarchy)

TimoPtr:
## Summary

We have few ANRs with stack that is not very useful in Sentry. This PR adds two new flags about collecting [threadDump](https://docs.sentry.io/platforms/android/configuration/app-not-respond/#attaching-thread-dump) and [view hierarchy](https://docs.sentry.io/platforms/android/configuration/options/#attachViewHierarchy) to help debug.

### DIFF
--- a/app/src/full/kotlin/io/homeassistant/companion/android/CrashHandling.kt
+++ b/app/src/full/kotlin/io/homeassistant/companion/android/CrashHandling.kt
@@ -27,6 +27,8 @@ fun initCrashReporting(context: Context, enabled: Boolean) {
 
     SentryAndroid.init(context) { options ->
         options.isEnableAutoSessionTracking = true
+        options.isAttachAnrThreadDump = true
+        options.isAttachViewHierarchy = true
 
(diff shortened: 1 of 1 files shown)


jpelgrom:
Why do you need view hierarchy info?

This led me to check the privacy policy which doesn't mention crash statistics or processors at all. I think that should be added + users should be explicitly informed we're changing how much info is collected.

TimoPtr:
> Why do you need view hierarchy info?
> 
> This led me to check the privacy policy which doesn't mention crash statistics or processors at all. I think that should be added + users should be explicitly informed we're changing how much info is collected.

I do have crashes that are only native (C/C++) side, which leads to stacktrace not being very useful. I would like to know on which screen the user is when it crashes.

But fair point about disclosure. I'm going to put on hold this PR until someone with more legal knowledge look at the privacy policy again.

TimoPtr:
Let's reopen if we ever need more information.

PR #6148, opened 2025-12-10
GetApiUrl check allow insecure connection and refactor IntegrationRepository

TimoPtr:
## Summary

This PR makes sure to not provide URL that are not secure to use in the current context based on the allowInsecure flag. It also mutualise the logic in the integration repository to handle errors in the same way everywhere.

## Checklist

## Any other notes

Based onto #6131

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
@@ -323,12 +323,10 @@ class IntegrationRepositoryImpl @AssistedInject constructor(
         val requestBody = RateLimitRequest(pushToken)
 
         return try {
-            checkNotNull(
-                integrationService.getRateLimit(
-                    RATE_LIMIT_URL,
-                    requestBody,
-                ).rateLimits,
-            ) { "Rate limit response is null" }
+            integrationService.getRateLimit(
+                RATE_LIMIT_URL,
+                requestBody,
+            ).rateLimits
         } catch (e: Exception) {
             Timber.e(e, "Unable to get notification rate limits")
             throw IntegrationException(e)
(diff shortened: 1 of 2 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
```diff
@@ -396,39 +356,21 @@ class IntegrationRepositoryImpl @AssistedInject constructor(
     }
 
     override suspend fun isHomeAssistantVersionAtLeast(year: Int, month: Int, release: Int): Boolean {
-        if (!isRegistered()) return false
+        if (!server().connection.isRegistered) return false
```
TimoPtr, line 360: I think it was wrong to only test the URL before, but lemme know if you think otherwise. If I had kept it like it was in insecure place this function would have return false even if the server (not available at the moment) was at least the given version. 

Indeed it is going to fail later in the flow, but with a proper error not with an error coming from a flow bellow a given version.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt
```diff
@@ -350,25 +318,15 @@ class IntegrationRepositoryImpl @AssistedInject constructor(
...
         val pushToken = localStorage.getString(PREF_PUSH_TOKEN) ?: ""
         val requestBody = RateLimitRequest(pushToken)
-        var checkRateLimits: RateLimitResponse? = null
-
-        var causeException: Exception? = null
 
-        try {
-            checkRateLimits =
-                integrationService.getRateLimit(RATE_LIMIT_URL, requestBody).rateLimits
+        return try {
```
TimoPtr, line 325: This change was simple to simplify the code. Now all the code in this file are handle almost in the same way.

PR #6150, opened 2025-12-11
Websocket react on URL change 

TimoPtr:
## Summary

This is an attempt to make the Websocket impl to react to a URL change.

## Checklist

## Any other notes

This is an attempt because I'm not sure it is actually doing anything, I don't know if there are a use case where the URL change that is not already disconnecting the web socket. Maybe if the internal is still accessible, but changing wifi SSID does disconnect briefly the WS already.

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
@@ -287,7 +287,8 @@ internal class WebSocketCoreImpl(
                         } else {
                             Timber.w("URL changed, disconnecting immediately.")
                         }
-                        // we abruptly cancel the connection since we might be in an insecure state
+                        // Set state before cancel() since cancel triggers onFailure -> handleClosingSocket
+                        connectionState = WebSocketState.CLOSED_URL_CHANGE
                         connection?.cancel()
                     }
                 }
@@ -654,13 +655,20 @@ internal class WebSocketCoreImpl(
     private fun handleClosingSocket() {
         urlObserverJob?.cancel()
         urlObserverJob = null
+        // Capture current state before it gets modified in the mutex block
+        val closingState = connectionState
         val cancelPendingMessagesJob = wsScope.launch {
             connectedMutex.withLock {
                 connected = CompletableDeferred()
                 connection = null
                 connectionHaVersion = null
                 connectedUrl = null
-                if (connectionState != WebSocketState.CLOSED_AUTH) {
+                // Preserve specific closure states, otherwise set to CLOSED_OTHER
+                if (connectionState !in listOf(
+                        WebSocketState.CLOSED_AUTH,
+                        WebSocketState.CLOSED_URL_CHANGE,
+                    )
+                ) {
                     connectionState = WebSocketState.CLOSED_OTHER
                 }
                 activeMessages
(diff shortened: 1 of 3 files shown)


jpelgrom:
> I don't know if there are a use case where the URL change that is not already disconnecting the web socket.

Changing from 5G (with public URL that also works when at home) to Wi-Fi (with internal IP) doesn't immediately trigger a disconnect for me, it takes the system a while to actively stop using 5G for network access for active connections and disconnect the websocket. With this change, I suppose that may happen sooner from the app's end now instead of waiting for the system to disconnect + the app to notice?

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
```diff
@@ -266,7 +288,71 @@ connect()
         assertTrue(result)
         // auth
         coVerify(exactly = 1) { mockConnection.send(any<String>()) }
-        assertSame(WebSocketState.ACTIVE, webSocketCore.getConnectionState())
+        assertEquals(WebSocketState.ACTIVE, webSocketCore.getConnectionState())
+    }
+
+    @Test
+    fun `Given an active connection When URL changes Then it cancels the connection`() = runTest {
```
jpelgrom:
```suggestion
    fun `Given an active connection When URL changes to a different URL Then it cancels the connection`() = runTest {
```

Make the difference with the second test more obvious

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -242,15 +248,54 @@ internal class WebSocketCoreImpl(
...
+                        isFirstEmission = false
+                        firstUrlDeferred.complete(url)
+                    } else if (url != connectedUrl) {
+                        if (urlState is UrlState.InsecureState) {
+                            Timber.w("Insecure state, disconnecting immediately.")
+                        } else {
+                            Timber.w("URL changed, disconnecting immediately.")
+                        }
+                        // we abruptly cancel the connection since we might be in an insecure state
+                        connection?.cancel()
```
jpelgrom: - Maybe introduce `WebSocketState.CLOSED_URL_CHANGE` for this?
- In case the URL changes, do we want the same delay for reconnecting (10s) or none / a shorter delay?

TimoPtr:
> > I don't know if there are a use case where the URL change that is not already disconnecting the web socket.
> 
> Changing from 5G (with public URL that also works when at home) to Wi-Fi (with internal IP) doesn't immediately trigger a disconnect for me, it takes the system a while to actively stop using 5G for network access for active connections and disconnect the websocket. With this change, I suppose that may happen sooner from the app's end now instead of waiting for the system to disconnect + the app to notice?

Interesting, in that case I'm curious if you could test if you see a difference. I only tested changing Wi-Fi to another one.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
TimoPtr: Done 👍🏻

TimoPtr: If it's easy for you to test with 5G lemme know how it behaves on your side.

jpelgrom: Shouldn't be too difficult, I'll test it tomorrow

jpelgrom:
Sorry for the late response: testing when switching between 5G/Wi-Fi now shows immediate disconnects, with the system also immediately switching to the new connection when the websocket is restarted, and also immediate disconnects for trusted Wi-Fi/other Wi-Fi.

Testing note: if the websocket is returning a null URL it should probably be handled in the WebSocketManager's job to not indicate a connection in the notification while there is actually none, but that is out of scope of this PR.

jpelgrom (approved):

PR #6159, opened 2025-12-15
Fix app startup hang on LAN-only networks without internet

### LINKED ISSUE #6099: LAN-Only Android Companion App Stuck on Launch
2025.11.3

**Android version(s):**
13

**Device model(s):**
Pixel 4XL

**Home Assistant version:**
current Docker image

2025.7.1

initial app setup works (select server, log in, "Firebase error" displays then dashboard loads).  after an app or device restart it will only display spinning "busy" signal.

**Additional information:**
android device is on an IOT VLAN with no internet access, guessing it hangs waiting for Firebase or some other cloud connection even though the connection persistence is set to "always".  tried "full" and "minimal" releases, same result.  reverting to 2025.7.1 fixes issue.

vaind:
## Summary

This PR fixes a critical startup hang issue where the Home Assistant Companion app would hang indefinitely on the splash screen when connected to LAN-only networks without internet access (e.g., isolated IoT VLANs).

**Root Cause:**
- The network monitoring logic relied on Android's `NET_CAPABILITY_VALIDATED` flag
- On LAN-only networks without internet, this flag is never set
- The app would wait indefinitely in the `CONNECTING` state

**Solution:**
- When a network is active but not validated, check if the external URL is a private/local address
- Private addresses include: `192.168.x.x`, `10.x.x.x`, `.local`, `.home`, `.lan`, `.internal` domains
- If the URL is local, proceed with `READY_NET_LOCAL` state instead of waiting
- Let the WebView/API layer handle actual connectivity failures

**Additional Changes (per reviewer feedback):**
- Renamed `NetworkState` enum values for clarity:
  - `READY_LOCAL` → `READY_INTERNAL` (on configured home network via SSID/VPN/Ethernet)
  - `READY_REMOTE` → `READY_NET_VALIDATED` (internet connectivity validated by Android)
  - New: `READY_NET_LOCAL` (network not validated, but URL is private/local)
- Fixed `CancellationException` handling to properly propagate coroutine cancellation
- Adapted to new `ServerConnectionStateProvider` API after merging with main

**Related Issue:** Fixes #6099

## Checklist

- [x] New or updated tests have been added to cover the changes
  - Added test case for LAN-only network scenario (`READY_NET_LOCAL`)
  - Updated all existing tests with new enum names
- [x] The code follows the project's code style and best practices
  - Code formatted with ktlint
  - Follows existing architecture patterns
- [x] The changes have been thoroughly tested
  - **Verified on real Android device with LAN-only network (no internet)**
  - Confirmed app loads successfully with `.home` domain on unvalidated network
  - Verified normal internet-connected networks still work
- [x] Changes are backward compatible
  - No breaking API changes
  - Only affects edge case that was previously broken

## Test Results

Verified on real device connected to network without internet:
- Network capabilities show `INTERNET` but NOT `VALIDATED`
- External URL: `http://ha.home:8123` (`.home` TLD recognized as local)
- App proceeds via `READY_NET_LOCAL` → WebView loads successfully

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitor.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitor.kt
@@ -9,13 +9,13 @@ import io.homeassistant.companion.android.util.isPubliclyAccessible
 import javax.inject.Inject
 import javax.inject.Singleton
 import kotlin.coroutines.cancellation.CancellationException
-import timber.log.Timber
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.channels.awaitClose
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.callbackFlow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.mapLatest
+import timber.log.Timber
 
 interface NetworkStatusMonitor {
     /**


jpelgrom:
If this fixes the issue as you describe, wouldn't setting the network as home network also fix it? There is no validation for that (see the state `READY_LOCAL`), while this change potentially breaks other setups.

vaind:
> If this fixes the issue as you describe, wouldn't setting the network as home network also fix it? There is no validation for that (see the state `READY_LOCAL`), while this change potentially breaks other setups.

I've had the network configured as home network on a dev device and it had no bearing on the issue.

Also, normally, I'd only connect to the IoT network with my phone running companion app when I need to set up a device that requires me to be on that network (e.g. matter) but it's actually not the "home network" for the phone.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitor.kt
```diff
@@ -91,15 +91,32 @@ internal class NetworkStatusMonitorImpl @Inject constructor(
...
+            !hasActiveNetwork -> NetworkState.UNAVAILABLE
+            isInternal -> NetworkState.READY_LOCAL
+            isValidated -> NetworkState.READY_REMOTE
+            hasExternalUrl -> {
+                // Fix for issue #6099: On LAN-only networks without internet access,
+                // NET_CAPABILITY_VALIDATED will never be true. If we have an active network
+                // and an external URL configured, assume the network is ready and attempt connection.
+                // The WebView/API layer will handle actual connectivity failures.
+                NetworkState.READY_REMOTE
+            }
```
TimoPtr: This is always true. If you look a bit deeper in the code you would see that the only case where externalUrl is empty is if the parsing failed. When onboarding the app we do set the externalUrl field with the URL that we've discovered or the one entered manually.

TimoPtr: I get your idea of connecting locally without internet, in that case what we should do is to check if the IP is local check `URLUtil.isPubliclyAccessible`, if the URL is not public then you could return `READY_REMOTE`.

Even if the naming is weird, I think that for consistency we should not return `READY_LOCAL`, wdyt @jpelgrom.

vaind: I've added the requested check for local url

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitor.kt
```diff
@@ -91,15 +100,43 @@ internal class NetworkStatusMonitorImpl @Inject constructor(
...
+                    }
+                } ?: false
+
+                if (urlIsPrivate) {
+                    // Fix for issue #6099: On LAN-only networks without internet access,
+                    // NET_CAPABILITY_VALIDATED will never be true. If we have an active network
+                    // and the external URL is a local/private address (e.g., http://192.168.1.100:8123,
+                    // http://ha.local, http://ha.home), assume the network is ready and attempt connection.
+                    // The WebView/API layer will handle actual connectivity failures.
+                    NetworkState.READY_REMOTE
```
TimoPtr: @jpelgrom what about renaming the state READY_EXTERNAL?

In any cases we need to update the doc in `NetworkState`

What bother me currently is that the state doesn't represent the reality.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitor.kt
```diff
@@ -91,15 +100,43 @@ internal class NetworkStatusMonitorImpl @Inject constructor(
...
+            isInternal -> NetworkState.READY_LOCAL
+            isValidated -> NetworkState.READY_REMOTE
+            else -> {
+                // Check URL only when we need it - when network exists but not validated and not internal
+                val urlIsPrivate = serverConfig.externalUrl?.let { urlString ->
+                    try {
+                        !URL(urlString).isPubliclyAccessible()
+                    } catch (e: Exception) {
+                        false
+                    }
```
TimoPtr:
```suggestion
                    } catch (e: CancellationException) {
                       throw e
                    } catch (e: Exception) {
                        false
                    }
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitor.kt
jpelgrom: This reply is also for your previous mention https://github.com/home-assistant/android/pull/6159#discussion_r2619548906

I think we need a rename of all the states to reduce confusion between purely URL config-derived states and network-derived states. The new scenario introduced by this PR also needs a different state from validated to match reality, I think.

|Current name|New name|Reasoning|
|----|----|----|
|NetworkState.READY_LOCAL|NetworkState.READY_INTERNAL|Used for when you're on the internal/home network, network validated doesn't matter|
|NetworkState.READY_REMOTE with isValidated|NetworkState.READY_NET_VALIDATED|This is used if the network is validated, url config doesn't matter|
|NetworkState.READY_REMOTE with unvalidated networks and non-public external URLs|NetworkState.READY_NET_LOCAL|This is used if the network is not validated but (presumably) ready for local urls only (and the app has one which is why it's ready instead of waiting)|

And for each of the `READY_*` states the app can continue, the different states are here mainly to keep the logs more understandable and for future developers.

TimoPtr: Sounds good @vaind is it something that you are willing to take on your plate?

jpelgrom:
@vaind It looks like you've addressed all comments and user feedback has been positive so far. Is this ready for review/merging?

vaind:
> @vaind It looks like you've addressed all comments and user feedback has been positive so far. Is this ready for review/merging?

yep, I've verified locally on device & iot-only net and it still works after changes

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/NetworkStatusMonitor.kt
```diff
@@ -95,18 +109,47 @@ internal class NetworkStatusMonitorImpl @Inject constructor(
...
+                // Check URL only when we need it - when network exists but not validated and not internal
+                val urlIsPrivate = try {
+                    connectionStateProvider.getExternalUrl()?.let { url ->
+                        !url.isPubliclyAccessible()
+                    } ?: false
+                } catch (e: CancellationException) {
+                    throw e
+                } catch (e: Exception) {
+                    false
+                }
```
TimoPtr, line 142: Let's log the exception.

TimoPtr: And add a test where getExternalUrl throws to assert the default value.

TimoPtr (approved):

PR #6163, opened 2025-12-16
Improve concurrency and hide internal logic of WebSocketCoreImpl

TimoPtr:
Remove serialize blocks and handle cancelation properly

## Summary

I started this PR by making sure we are throwing CancelableException to properly clear the coroutines. I don't think this is completely achieved since we might catch this later in the chain.
While working on this I decide to improve the WebSocketCoreImpl interface by not exposing outside of this class the internals like all the channels that were kept within the WebSocketRequest that is now very simple and only hold data.

To do that I used the same concept than for the messageQueue to send the message. It allows me to ensure that the message are sent in the right order and we don't need synchronize block that shouldn't be use while using coroutine.

This is an attempt again to reduce the number of OOM we have due to the current impl of the WS.

I've used a sealed classes to have stronger separation of concern.

## Checklist

## Any other notes

I'm considering using the channel to send the initial auth message from connect().

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
@@ -7,9 +7,8 @@ import io.homeassistant.companion.android.common.data.HomeAssistantApis.Companio
 import io.homeassistant.companion.android.common.data.HomeAssistantVersion
 import io.homeassistant.companion.android.common.data.authentication.AuthorizationException
 import io.homeassistant.companion.android.common.data.servers.ServerManager
-import io.homeassistant.companion.android.common.data.servers.firstUrlOrNull
-import io.homeassistant.companion.android.common.data.websocket.HAWebSocketException
 import io.homeassistant.companion.android.common.data.servers.UrlState
+import io.homeassistant.companion.android.common.data.websocket.HAWebSocketException
 import io.homeassistant.companion.android.common.data.websocket.WebSocketCore
 import io.homeassistant.companion.android.common.data.websocket.WebSocketRequest
 import io.homeassistant.companion.android.common.data.websocket.WebSocketState


common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
```diff
@@ -684,5 +737,252 @@ misc
...
+            } else {
+                fail { "Missing id" }
+            }
+            true
+        }
+
+        // Launch multiple sendMessage calls concurrently
+        val jobs = (1..50).map { i ->
+            async {
+                delay(Random.nextLong())
```
jpelgrom: This can be the cause of a flakey test

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -172,6 +173,63 @@ internal class WebSocketCoreImpl(
...
+     */
+    private val sendCommandChannel = Channel<CommandSender<*>>(capacity = Channel.UNLIMITED).apply {
+        backgroundScope.launch {
+            consumeEach {
+                processSendCommand(it)
+            }
+        }
+    }
+
+    // TODO check the base type and if we want to keep this public
```
jpelgrom: TODO should be resolved before merging

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -172,6 +173,63 @@ internal class WebSocketCoreImpl(
...
+    class HAWebSocketException(override val message: String?) : Exception()
+
+    /**
+     * This complete when the message has been processed and sent or not.
+     * This is not making any network call directly it enqueue the send into the WebSocket connection.
+     */
+    private fun processSendCommand(command: CommandSender<*>) {
+        // TODO do we need any protection on the connection like in connect() with the Mutex
```
jpelgrom: TODO should be resolved before merging - isn't consumeEach on a channel supposed to guarantee we have no other users?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -254,74 +317,73 @@ internal class WebSocketCoreImpl(
...
+    override suspend fun sendBytes(data: ByteArray): Boolean {
+        if (!connect()) {
             Timber.w("Unable to send bytes, not connected")
-            null
+            return false
         }
+
+        val command = CommandSender.BytesCommandSender(data)
+
+        // Send command to actor
```
jpelgrom: You don't have a comment on the other usage either

```suggestion
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -664,3 +737,98 @@ internal class WebSocketCoreImpl(
...
+        val request: WebSocketRequest,
+    ) : ActiveMessage
+}
+
+/**
+ * Represents a command to be sent through the send command channel for serialized WebSocket sending.
+ *
+ * @param T The type of result returned when the send operation completes
+ */
+private sealed interface CommandSender<T> {
```
jpelgrom: Usages like `sendMessage(CommandSender.WithAnswer.MessageCommandSender(request))` look foreign to me and I think it's because of the interface name and nesting. It is not a sender itself but an interface for commands so why not simply `Command.WithAnswer.Simple`/`Command.WithAnswer.Subscription`?

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
TimoPtr: It shouldn't happen or very unlikely since I'm asserting that the list is not ordered. But you are right, I'm going to change this to something deterministic 👍🏻

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
TimoPtr: Thanks for the hand in naming I didn't like it

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
TimoPtr: The issue is that connection can be nullify if for instance the connection is closed for any reason, while we are actually sending a message. 

I think it is fine in a way since I keep a reference to the object in the method. The bad thing that would happen is that it is going to fail to send the message on this specific object.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -173,6 +175,59 @@ internal class WebSocketCoreImpl(
...
+        backgroundScope.launch {
+            consumeEach {
+                processSendCommand(it)
+            }
+        }
+    }
+
+    /**
+     * This complete when the message has been processed and sent or not.
+     * This is not making any network call directly it enqueue the send into the WebSocket connection.
```
jpelgrom: - You used 'This' to refer both to a property of the command and the function...
 - Maybe there is fancy KDoc syntax to refer to `command.sendCompleted` that I'm unaware of?

```suggestion
     * Completes command.sendCompleted when the message has been processed to indicate success or failure
     * for sending. The function enqueues the call on the connection, it does not directly make the network call.
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -257,74 +317,77 @@ internal class WebSocketCoreImpl(
...
+        val id: Long
+        try {
+            id = command.sendCompleted.await()
+        } catch (e: HAWebSocketException) {
+            Timber.e(e, "Failed to send message ${command.request}")
+            return null
+        }
+
+        Timber.d("Message number $id sent, awaiting response from WebSocket")
+
```
jpelgrom: This will generate a lot of noise in the logs, as we already log the outgoing message. Failures are already logged so "Sending message" followed by nothing/a response is probably obvious enough. Suggestion: remove or limit to debug builds.

```suggestion
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -467,145 +528,158 @@ internal class WebSocketCoreImpl(
...
-                    }
-                } else if (subscriptionType == SUBSCRIBE_TYPE_SUBSCRIBE_TRIGGER) {
-                    val trigger = ((response.event as? JsonObject)?.get("variables") as? JsonObject)?.get("trigger")
-                    if (trigger != null) {
-                        kotlinJsonMapper.decodeFromJsonElement<TriggerEvent>(trigger)
-                    } else {
-                        Timber.w("Received no trigger value for trigger subscription, skipping")
+        val activeMessage = activeMessages[subscriptionId].let {
+            FailFast.failWhen(it !is ActiveMessage.Subscription) {
+                "Event should always be associated to a ActiveMessage.SubscriptionActiveMessage"
```
```suggestion
                "Event should always be associated to a ActiveMessage.Subscription message"
```

jpelgrom:
> I'm considering using the channel to send the initial auth message from connect().

I like this idea, any reason for not implementing it yet? (other than keeping the PR small, but there are already quite a lot of changes)

TimoPtr:
> > I'm considering using the channel to send the initial auth message from connect().
> 
> I like this idea, any reason for not implementing it yet? (other than keeping the PR small, but there are already quite a lot of changes)

So in connect we send two messages
- auth --> In the end I don't think this one should use the queue since it doesn't have an ID
- supported_features --> We could use the queue but for now I don't think it is important for the scope of this PR.

jpelgrom (approved):
I think these changes are good and they look good when testing, but let's make sure to have it in a beta cycle or two to be sure before pushing to production.

PR #6168, opened 2025-12-17
Update changelog for the new onboarding and blocking screen

TimoPtr:
## Summary

Add new entry in the changelog for the new onboarding and the blocking screen.


jpelgrom:
> Added security measure for unencrypted connections

Why not simply: Added connection security level options? That makes it easier for users to find the new thing you're mentioning.

jpelgrom (approved):

PR #6169, opened 2025-12-18
Rename LauncherActivity to LaunchActivity

TimoPtr:
## Summary

@dshokouhi found out that the shortcuts were removed when updating to the latest version available on main. It turns out it is due to the new onboarding being used in release. The new onboarding used a new activity named `.launcher.LauncherActivity` compare to the old one `.launch.LaunchActivity`, I didn't know until now that the shortcuts were relying on the fully qualified name of the default main activity used. Because of that when installing the new version it simply wiped all the dynamic shortcuts created.

This PR does change the name of the new activity to match the old one.

While fixing this I also fixed an issue for the Launcher feature, since we were hardcoding the name of the alias, but I did renamed the alias without updating the name that was harcoded in the code ... To avoid issues I simply bring back the name of the alias to what it was before.

## Checklist

## Any other notes

I added a warning within the manifest to catch future changes.


jpelgrom (approved):
Good thing we caught this before doing a public beta, and hard to spot during debug testing (sorry).

Issue #5763, opened 2025-09-09
Support streaming TTS for Assist in the app


zolakt:
**Is your feature request related to a problem? Please describe.**

Not sure if this should be a bug report or a feature request, but either way, streaming TTS (from a LLM voice pipeline) doesn't work in the companion app. It waits for the whole text response to be generated, before it starts TTS.

However it does work if you open HA in a browser, not through the app.

**Describe the solution you'd like**

Companion app should support streaming TTS.

**Describe alternatives you've considered, if any**

In Android the only alternative is not to use the app at all, and use HA in the browser, which is cumbersome.

**Additional context**

TimoPtr:
We need to implement it following the documentation of the events https://developers.home-assistant.io/docs/voice/pipelines/#events 

StartEvent now contains the URL that we need to play and core will stream the data when available.

alipsky:
Bumping this, this is the major source of latency on my android voice assistant.  I can get  local LLM to respond instantaneously, and TTS to stream on desktop browsers, but on android for LLM responses, voice assistant is basically unusable because streaming isnt supported!  This would make voice assistants usable on many devices

TimoPtr:
For anyone curious this has been implemented in #6188

PR #6188, opened 2025-12-23
Support streaming tts in Assist

TimoPtr:
## Summary

This PR adds support for streaming TTS audio (address #5763 ) in the Assist pipeline and migrates AudioUrlPlayer from MediaPlayer to ExoPlayer in order to be able to detect when the player has finish to play audio with the URL still opened which is not possilbe with the old API. I had to add cronet to the wear module for that.

`RUN_START` has a new field that contains the path to play, if we do have this path we don't use the one from `TTS_END`. We kept the old logic for backward compatibility. But now it should keep the URL open for the lifetime of the Assist view model.

This PR also addressed #6063 by moving it to `Default` so it can make an async load of cronet.

## Checklist

## Any other notes

I didn't add test for the view model but we should. Currently if you test on the wear it's going to crash because of the concurrent modification of the recorderQueue. If you comment the recorderQueue logic (buffering before init), the new player works at it should. But this needs to be fixed.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -297,32 +297,6 @@ jobs:
           sudo udevadm trigger --name-match=kvm
           ls /dev/kvm
 
-      - name: AVD cache
-        uses: actions/cache@9255dc7a253b0ccc959486e2bca901246202afeb # v5.0.1
-        id: avd-cache
-        with:
-          path: |
-            ~/.android/avd/*
-            ~/.android/adb*
-          key: avd-${{ matrix.configuration.target }}-${{ matrix.configuration.api-level }}
-
-      - name: Create AVD and generate snapshot for caching
-        if: steps.avd-cache.outputs.cache-hit != 'true'
-        uses: reactivecircus/android-emulator-runner@b530d96654c385303d652368551fb075bc2f0b6b # v2.35.0
-        with:
-          api-level: ${{ matrix.configuration.api-level }}
-          arch: ${{ matrix.configuration.arch }}
-          force-avd-creation: false
-          disable-animations: true
-          disk-size: 1G # We need to ensure that there is enough space for the emulator to run. If not we get 'Failed to commit install' errors.
-          heap-size: 1G
-          emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
-          profile: ${{ matrix.configuration.profile }}
-          target: ${{ matrix.configuration.target }}
-          system-image-api-level: ${{ matrix.configuration.system-image-api-level }}
-          script: |
-            echo "Generated AVD snapshot for caching."
-
       - name: Build projects and run instrumentation tests
         uses: reactivecircus/android-emulator-runner@b530d96654c385303d652368551fb075bc2f0b6b # v2.35.0
         with:
@@ -330,7 +304,7 @@ jobs:
           arch: ${{ matrix.configuration.arch }}
           force-avd-creation: false
           disable-animations: true
-          disk-size: 1G
+          disk-size: 2G # We need to ensure that there is enough space for the emulator to run. If not we get 'Failed to commit install' errors.
           heap-size: 1G
           emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
           profile: ${{ matrix.configuration.profile }}


common/src/main/kotlin/io/homeassistant/companion/android/common/assist/AssistViewModelBase.kt
```diff
@@ -106,75 +117,135 @@ abstract class AssistViewModelBase(
...
+                    }
                 }
             } ?: run {
                 onEvent(AssistEvent.Message.Output(app.getString(R.string.assist_error)))
             }
         }
     }
 
+    private fun handleRunStart(data: AssistPipelineRunStart?, isVoice: Boolean, onEvent: (AssistEvent) -> Unit) {
+        Timber.d("Pipeline run started")
```
jpelgrom: I think this should be removed as websocket messages already indicate this, seems like a debugging leftover

```suggestion
```

common/src/main/kotlin/io/homeassistant/companion/android/common/assist/AssistViewModelBase.kt
```diff
@@ -106,75 +117,135 @@ abstract class AssistViewModelBase(
...
+        }
+    }
+
+    /*
+     * Handles TTS_END events for backward compatibility with servers that don't support
+     * streaming TTS in RUN_START. If [currentPathBeingPlayed] is set, audio is already
+     * playing from RUN_START and this handler is skipped.
+     */
+    private fun handleTtsEnd(data: AssistPipelineTtsEnd?, isVoice: Boolean, onEvent: (AssistEvent) -> Unit) {
+        Timber.d("TTS ended")
```
jpelgrom:
```suggestion
```

common/src/main/kotlin/io/homeassistant/companion/android/common/assist/AssistViewModelBase.kt
```diff
@@ -106,75 +117,135 @@ abstract class AssistViewModelBase(
...
+            val audioPath = data?.ttsOutput?.url
+            if (!audioPath.isNullOrBlank()) {
+                playAudio(audioPath).first { state -> state == PlaybackState.STOP_PLAYING }
+            }
+            notifyContinueConversationIfNeeded(onEvent)
+        }
+    }
+
+    /**
+     * Return true if we need to cancel the job
```
jpelgrom, line 237: It always returns true, why does it return something at all?

common/src/main/kotlin/io/homeassistant/companion/android/common/util/AudioUrlPlayer.kt
```diff
@@ -1,171 +1,199 @@
...
+        },
+    )
 
     /**
-     * Stream and play audio from the provided [url]. Any currently playing audio will be stopped.
-     * This function will suspend until playback has started.
-     * If the current volume of the [STREAM_MUSIC] is 0 it doesn't play the audio and directly return false.
+     * Streams and plays audio from the provided [url].
+     *
+     * The player starts when the Flow is collected. The flow makes sure that the interaction with the player
```
jpelgrom: Unfinished sentence

common/src/main/kotlin/io/homeassistant/companion/android/common/util/AudioUrlPlayer.kt
```diff
@@ -1,171 +1,199 @@
...
+     *
+     * The Flow emits [PlaybackState] changes and completes if an error occurs or the upstreams ends.
+     * If the [audioManager] is null or the current volume of [STREAM_MUSIC] is 0,
+     * the Flow completes immediately without playing.
+     *
+     * The player is properly release once the flow is canceled.
```
jpelgrom:
```suggestion
     * The player is properly released once the flow is canceled.
```

common/src/main/kotlin/io/homeassistant/companion/android/common/util/AudioUrlPlayer.kt
```diff
@@ -1,171 +1,199 @@
...
-                AudioManagerCompat.requestAudioFocus(audioManager, it)
-            } catch (e: Exception) {
-                // We don't use the result / focus if available but if not still continue
-            }
+        return try {
+            AudioManagerCompat.requestAudioFocus(audioManager, request)
+            request
+        } catch (e: Exception) {
+            // Focus request failed, continue without audio focus
+            Timber.w(e, "Fail to request focus")
```
jpelgrom: We don't really care about no focus, so remove or at least fix the sentence:

```suggestion
            Timber.w(e, "Failed to request focus")
```

jpelgrom:
Unlinking the issue - it should stay open until Wear OS has been updated as well.

common/src/main/kotlin/io/homeassistant/companion/android/common/assist/AssistViewModelBase.kt
TimoPtr: It can return false if the `message` in data or `data` is null. I tried to keep the old behavior.

TimoPtr:
> Unlinking the issue - it should stay open until Wear OS has been updated as well.

Why? It is already working for WearOS if I'm not wrong I remember testing it too. The modification is only touching the base viewModel.

jpelgrom:
> > Unlinking the issue - it should stay open until Wear OS has been updated as well.
> 
> Why? It is already working for WearOS if I'm not wrong I remember testing it too. The modification is only touching the base viewModel.

Sorry, it looks like I was mistaken. Didn't work when I tested it but I must've missed something, works now and as you say it's all in the shared base.

TimoPtr:
> > > Unlinking the issue - it should stay open until Wear OS has been updated as well.
> > 
> > 
> > Why? It is already working for WearOS if I'm not wrong I remember testing it too. The modification is only touching the base viewModel.
> 
> Sorry, it looks like I was mistaken. Didn't work when I tested it but I must've missed something, works now and as you say it's all in the shared base.

Actually if you did test without the next PR on the watch it most probably crashes because of the concurrent modification issue.

nathan-curtis:
Did this just hit the beta? If so we may have an issue. Breaks my voice pipeline for TTS responses...

TimoPtr:
> Did this just hit the beta? If so we may have an issue. Breaks my voice pipeline for TTS responses...

It did hit beta yes. Could you open an issue?

nathan-curtis:
Yeah will do I originally opened it in core... Now that I know it's android...

Its bad broken- I had to rollback it kills the entire voice pipeline any time it tries to play TTS.

PR #6189, opened 2025-12-23
Make sure fragments have public ctor for the FragmentFactory reflection

TimoPtr:
## Summary

I learned the hard way that you should not make the default constructor of a class that extend https://developer.android.com/reference/androidx/fragment/app/Fragment.html private because the FragmentFactory does some reflection to find the default constructor at runtime when restoring the fragments.

This issue is causing some crashing each time a configuration changes happens. I found this issue on the beta lane of the Play console crash report.

## Checklist

## Any other notes

It would have been nice that Google made a lint rule instead of a hidden comment in the doc
> Constructor used by the default [FragmentFactory](https://developer.android.com/reference/androidx/fragment/app/FragmentFactory). You must [set a custom FragmentFactory](https://developer.android.com/reference/androidx/fragment/app/FragmentManager#setFragmentFactory(androidx.fragment.app.FragmentFactory)) if you want to use a non-default constructor to ensure that your constructor is called when the fragment is re-instantiated.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/ConnectionSecurityLevelFragment.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/ConnectionSecurityLevelFragment.kt
@@ -47,7 +47,7 @@ import io.homeassistant.companion.android.onboarding.locationforsecureconnection
  * This fragment is temporary and should be removed once the app fully migrates to Compose Navigation.
  */
 @AndroidEntryPoint
-class ConnectionSecurityLevelFragment private constructor() : Fragment() {
+class ConnectionSecurityLevelFragment : Fragment() {
 
     companion object {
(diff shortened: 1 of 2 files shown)


TimoPtr:
To reproduce the crash on main. Open the app and make sure to be on the blocking screen. Then put the app in background and kill it with ADB

```
adb shell am kill io.homeassistant.companion.android.debug
```

Then open the app again on main it crashes with 

```
Caused by: java.lang.NoSuchMethodException: io.homeassistant.companion.android.webview.insecure.BlockInsecureFragment.<init> []
	at java.lang.Class.getConstructor0(Class.java:3379)
	at java.lang.Class.getConstructor(Class.java:2471)
	at androidx.fragment.app.Fragment.instantiate(Fragment.java:677)
	... 33 more
```

PR #6190, opened 2025-12-23
Make safe interaction with fragment manager in the webview

TimoPtr:
## Summary

Even if the WebViewPresenter.load function is lifecycle aware we are using coroutines and the cancellation might happen at a bad time like after the onStop has been called on the WebViewActivity. This leads to crashes because the fragment manager is in a state that cannot take any new transaction at the moment.

```
Exception java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
  at androidx.fragment.app.FragmentManager.checkStateLoss (FragmentManager.java:1862)
  at androidx.fragment.app.FragmentManager.enqueueAction (FragmentManager.java:1902)
  at androidx.fragment.app.FragmentManager.popBackStack (FragmentManager.java:1015)
  at io.homeassistant.companion.android.webview.WebViewActivity.loadUrl (WebViewActivity.kt:1442)
  at io.homeassistant.companion.android.webview.WebViewPresenterImpl$loadUrl$2$1.invokeSuspend (WebViewPresenterImpl.kt:225)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:34)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:100)
  at android.os.Handler.handleCallback (Handler.java:873)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loop (Looper.java:280)
  at android.app.ActivityThread.main (ActivityThread.java:6706)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:493)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:858)
```

From the FragmentManager.isStateSaved docs:

```java
    /**
     * Returns {@code true} if the FragmentManager's state has already been saved
     * by its host. Any operations that would change saved state should not be performed
     * if this method returns true. For example, any popBackStack() method, such as
     * {@link #popBackStackImmediate()} or any FragmentTransaction using
     * {@link FragmentTransaction#commit()} instead of
     * {@link FragmentTransaction#commitAllowingStateLoss()} will change
     * the state and will result in an error.
     *
     * @return true if this FragmentManager's state has already been saved by its host
     */
    public boolean isStateSaved() {
```

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -102,6 +102,7 @@ import io.homeassistant.companion.android.common.util.getStringOrElse
 import io.homeassistant.companion.android.common.util.getStringOrNull
 import io.homeassistant.companion.android.common.util.isAutomotive
 import io.homeassistant.companion.android.common.util.jsonObjectOrNull
+import io.homeassistant.companion.android.common.util.runFragmentTransactionIfStateSafe
 import io.homeassistant.companion.android.common.util.toJsonObject
 import io.homeassistant.companion.android.common.util.toJsonObjectOrNull
 import io.homeassistant.companion.android.database.authentication.Authentication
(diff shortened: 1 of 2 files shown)


jpelgrom:
While I agree this is most likely to happen in the WebViewActivity, I think it's worth placing the function in a more generic file so it can be used in other places as well.

jpelgrom (approved):

PR #6192, opened 2025-12-23
Check if BatteryOptimization intent can be resolved before using it

TimoPtr:
## Summary

From Sentry I found out that some device are not resolving the BatteryOptimization intent, this time it was a unifi_screen device.

```
Activity could not be started for Intent { act=android.settings.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS dat=package:io.homeassistant.companion.android }
```

This PR simply check that the Activity can be resolved before actually using it.

## Checklist

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/util/ContextExt.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/util/ContextExt.kt
@@ -78,7 +78,6 @@ fun Context.createBatteryOptimizationIntent(): Intent? {
 /**
  * Checks if the app is ignoring battery optimizations.
  *
- * TODO this should not be exposed to the wear module https://github.com/home-assistant/android/discussions/5771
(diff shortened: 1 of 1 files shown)


jpelgrom:
It really doesn't seem to like the instrumentation test on API 23 with this change, even after a couple of retries 🤔

> TODO this should not be exposed to the wear module https://github.com/home-assistant/android/discussions/5771

Actually it may now be returning null so no longer an issue to expose it?

TimoPtr:
> It really doesn't seem to like the instrumentation test on API 23 with this change, even after a couple of retries 🤔
> 
> > TODO this should not be exposed to the wear module #5771
> 
> Actually it may now be returning null so no longer an issue to expose it?

Indeed.

I fixed the failing test, it was because Robolectric was not mocking this specific activity. For now I kept it within the test itself, but later we might need this mock elsewhere, in that case I'll move it somewhere else.

PR #6212, opened 2025-12-30
Fix server manager race condition read on DB only and don't store temporary server

TimoPtr:
## Summary

This PR fixes the race conditions that we had in the ServerManager implementation. This PR reworked everything to remove the need of an in-memory cache in the Servermanager.

In order to do that I decided to not keep the temporary servers within the ServerManager, it was mostly useful for the wear. I've adjusted the code so that we don't need it anymore. It comes with some duplication for some API.

I've introduce a `TemporaryServer` object that is only used in the onboarding, then it is given to the ServerManager to add it to the DB.

Now we always read from the DB instead of relying on the `mutableServer` list that was causing race condition while creating instances of a repository. This issue was causing a lot of crashes.

In order to fully work with the DB we use suspend function everywhere, that leads to lot of small changes everywhere. 

## Checklist

## Any other notes

Indeed the DB is slower than an in-memory cache, but it simplify a lot the logic. The impact should not be visible by a user compare to the crashes.

### DIFF
--- a/app/src/test/kotlin/io/homeassistant/companion/android/launch/LaunchViewModelTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/launch/LaunchViewModelTest.kt
@@ -500,7 +500,7 @@ class LaunchViewModelTest {
             user = ServerUserInfo(id = null, name = null, isOwner = false, isAdmin = false),
         )
 
-        coEvery { serverManager.defaultServers } returns listOf(connectedServer1, connectedServer2)
+        coEvery { serverManager.servers() } returns listOf(connectedServer1, connectedServer2)
         coEvery { serverManager.authenticationRepository(connectedServer1.id).getSessionState() } returns SessionState.CONNECTED
         coEvery { serverManager.authenticationRepository(connectedServer2.id).getSessionState() } returns SessionState.CONNECTED
 


app/src/full/kotlin/io/homeassistant/companion/android/matter/MatterCommissioningViewModel.kt
```diff
@@ -41,6 +42,15 @@ class MatterCommissioningViewModel @Inject constructor(
...
         private set
 
+    var servers by mutableStateOf<List<Server>>(emptyList())
+        private set
+
+    init {
+        viewModelScope.launch {
+            servers = serverManager.defaultServers()
+        }
+    }
```
jpelgrom: I don't think this is needed; this is only used for the `SelectServer` step and right before going to that step you fetch the servers (line 67), so you can simply update the variable in there. Also avoids a potential new race where the UI might be on `SelectServer` but `servers` hasn't updated yet.

app/src/full/kotlin/io/homeassistant/companion/android/settings/wear/SettingsWearUseCase.kt
```diff
@@ -0,0 +1,156 @@
...
+    }
+}
+
+/**
+ * Handles Home Assistant server operations specifically for Wear OS device onboarding.
+ *
+ * This use case communicates directly with the Home Assistant server using [WearServer]
+ * credentials without requiring a persisted server in the app's database.
+ */
+class SettingsWearUseCase @Inject constructor(
```
jpelgrom: We use `*UseCase` in a couple of variables but tend to call these classes `*Manager`s, or maybe in this case a new `*Repository` like onboarding as technically you're not using those but have similar functions.

app/src/main/kotlin/io/homeassistant/companion/android/launch/link/LinkHandler.kt
```diff
@@ -139,6 +149,7 @@ class LinkHandlerImpl @Inject constructor(private val serverManager: ServerManag
      * Requires a registered server to proceed.
      *
      * @param uri The redirect URI to process.
+     * TODO update the doc
```
jpelgrom: TODO

app/src/main/kotlin/io/homeassistant/companion/android/launch/link/LinkActivity.kt
```diff
@@ -76,26 +76,17 @@ class LinkActivity : BaseActivity() {
                     }
 
                     is LinkDestination.Webview -> {
-                        navigateTo(destination.path, destination.serverId)
+                        startLaunchWithNavigateTo(destination.path, destination.serverId)
+                        finish()
+                    }
+                    is LinkDestination.ServerPicker -> {
+                        openServerChooser(destination.path)
```
jpelgrom, line 83: This shuffle around may be a nice small improvement, but makes big PRs like these harder to review. Try to pull them out in the future, it's a small change that's quickly reviewed merged and can then be incorporated into the big PR branch.

app/src/main/kotlin/io/homeassistant/companion/android/launch/link/LinkHandler.kt
```diff
@@ -184,12 +195,13 @@ class LinkHandlerImpl @Inject constructor(private val serverManager: ServerManag
...
             "default", null -> serverManager.getServer()?.id
-            else -> serverManager.defaultServers.find {
+            else -> serverManager.defaultServers().find {
                 it.friendlyName.equals(serverName, ignoreCase = true)
             }?.id
         }
 
-        return LinkDestination.Webview(uri.toString(), serverId)
+        val path = uri.toString()
+        return serverId?.let { LinkDestination.Webview(uri.toString(), serverId) } ?: webviewDestination(path)
```
jpelgrom: Same result, make this code simpler

```suggestion
        return webviewDestination(uri.toString(), serverId)
```

app/src/main/kotlin/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
```diff
@@ -77,6 +77,11 @@ class EntityGridVehicleScreen(
                     entities = it
                     if (hasChanged) invalidate()
                 }
+
+                if (serverManager.defaultServers().size > 1 && !shouldSwitchServers) {
+                    shouldSwitchServers = true
+                    invalidate()
+                }
```
jpelgrom: I'd put this before collecting from a flow

app/src/main/kotlin/io/homeassistant/companion/android/vehicle/MainVehicleScreen.kt
```diff
@@ -109,6 +111,11 @@ class MainVehicleScreen(
                         }
                     }
                 }
+
+                if (serverManager.defaultServers().size > 1 && !shouldSwitchServers) {
+                    shouldSwitchServers = true
+                    invalidate()
+                }
```
jpelgrom: I'd put this before collecting from a flow

app/src/main/kotlin/io/homeassistant/companion/android/vehicle/MainVehicleScreen.kt
```diff
@@ -62,6 +62,8 @@ class MainVehicleScreen(
 
     private val isAutomotive get() = carContext.isAutomotive()
 
+    private var shouldSwitchServers = false
```
jpelgrom:
```suggestion
    private var canSwitchServers = false
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManager.kt
```diff
@@ -14,84 +14,100 @@ interface ServerManager {
     }
 
     /**
-     * A list of all [Server]s managed by the app of the type [ServerType.DEFAULT].
+     * Returns a list of all registered [Server]s.
      */
-    val defaultServers: List<Server>
+    suspend fun defaultServers(): List<Server>
```
jpelgrom: If there are no more server types, why is it still named `defaultServers`?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManagerImpl.kt
```diff
@@ -40,188 +78,105 @@ class ServerManagerImpl @Inject constructor(
...
-        }
+    override suspend fun updateServer(server: Server) {
+        serverDao.update(server)
     }
 
     override suspend fun removeServer(id: Int) {
         authenticationRepository(id).deletePreferences()
         integrationRepository(id).deletePreferences()
         prefsRepository.removeServer(id)
-        removeServerFromManager(id)
```
jpelgrom, line 165: This was originally written specifically with this order, so that removing references to IDs and from the database was the last thing. If any repository tries to write/update things to the server on shutdown, the order would ensure it is also removed.

common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerDao.kt
```diff
@@ -22,6 +22,9 @@ interface ServerDao {
     @Query("SELECT * FROM servers ORDER BY `list_order` ASC")
     suspend fun getAll(): List<Server>
 
+    @Query("SELECT * FROM servers ORDER BY id DESC LIMIT 1")
+    suspend fun getLast(): Server?
```
jpelgrom: You're only using this to get the ID, so could change this to only return the ID

TimoPtr:
> Converting Wear servers to a specific objects works for the current feature set, but does make it a lot more complicated if we ever wanted to make it possible to manage more watch settings from the phone (as most HA integrations need the websocket now). Maybe a better separation of concerns, but it does feel a bit like closing that door and keeping the current barebones Wear settings.

I agree, but for now the goal was to simplify the management of the ServerManager. If we want to add more features for the watch from the phone we should probably make a dedicated ServerManager for it so that we could get the other repositories use this manager. This manager would be in memory only. This is a quick suggestion that I've been thinking off without playing with it. My takes is that we should work on that the day we need it to avoid making premature choices. Since the set of API used was small I preferred to take this simpler approach of "duplicating".

app/src/main/kotlin/io/homeassistant/companion/android/launch/link/LinkActivity.kt
TimoPtr: I agree, but in fact here it was required since navigateTo was using `defaultServers`. I should have made a first PR to move it to the view model and then stack this one on top agreed but I was already deep into the refactor.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManagerImpl.kt
TimoPtr: Let's add a test to prevent someone to change the order

PR #6223, opened 2026-01-02
Always show LocationForSecureConnection when using HTTP while onboarding

TimoPtr:
## Summary

While adding multiple server I found out that the `LocationForSecureConnection` was not shown during the onboarding of a second server instead I was seeing the `ConnectionSecurityLevelFragment` with the cross. It was making the selection of the option optional that we don't want.

It is due to an early logic that I left in the graph, I wanted to automatically pick `Most secure` when the user was enabling location tracking, but we never set this. While it was a nice idea I think it might still be best to show the option picker, to let the user understand.

This also had an impact on the minimal flavor since there was no location tracking screen at all, it was only checking if the location permission was given if yes then it was skip and the fragment was shown.

This PR is not going to have a huge impact on the users, but simply making sure that the navigation graph is coherent.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
@@ -87,8 +87,8 @@ data class WearOnboardingRoute(val wearName: String, val urlToOnboard: String? =
  * 3. Connection and authentication
  * 4. Device naming and registration
  * 5. Post-registration flow based on server configuration:
- *    - Local-only servers: Local first screen → full: Location sharing; minimal secure connection (HTTP) or done (HTTPS)
- *    - Public servers with full flavor: Location sharing → Secure connection (HTTP only)
+ *    - Local-only servers: Local first screen → (full flavor only: Location sharing →) Secure connection (HTTP) or done (HTTPS)
+ *    - Public servers with full flavor: Location sharing → Secure connection (HTTP) or done (HTTPS)
  *    - Public servers with minimal flavor: Secure connection (HTTP) or done (HTTPS)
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingNavigation.kt
```diff
@@ -84,23 +80,25 @@ data class WearOnboardingRoute(val wearName: String, val urlToOnboard: String? =
...
- * 3. Device naming and registration
- * 4. Location/security configuration (conditional)
- * 5. Home network configuration (if applicable)
+ * 1. Welcome screen (skipped if [skipWelcome] is true)
+ * 2. Server discovery (skipped if [urlToOnboard] is provided)
+ * 3. Connection and authentication
+ * 4. Device naming and registration
+ * 5. Post-registration flow based on server configuration:
+ *    - Local-only servers: Local first screen → full: Location sharing; minimal secure connection (HTTP) or done (HTTPS)
+ *    - Public servers with full flavor: Location sharing → Secure connection (HTTP only)
```
jpelgrom: At first I thought there was no path for public servers with full flavor with HTTPS, and you don't mention there is another screen after location sharing for local-only servers.

```suggestion
 *    - Local-only servers: Local first screen → (full flavor only: Location sharing →) Secure connection (HTTP) or done (HTTPS)
 *    - Public servers with full flavor: Location sharing → Secure connection (HTTP) or done (HTTPS)
```

jpelgrom (approved):
This may not have a big impact, but any changes to this graph make me triple check everything 😵‍💫.

In the end it does make things simpler and more coherent, and onboarding doesn't happen that frequently so nice change!

PR #6224, opened 2026-01-02
Forbid main thread queries from room

TimoPtr:
## Summary

In theory everything is now interacting with the database in the main thread. We might have forget some usages, but a good way to find them would be to try to remove this option on room that allow the queries to happen on the main thread.

## Checklist


jpelgrom:
Makes me wonder why it was ever set to allow this in the first place, as Rooms default is to not allow this.

TimoPtr:
> Makes me wonder why it was ever set to allow this in the first place, as Rooms default is to not allow this.

It was added in July 2020 https://github.com/home-assistant/android/pull/633 I think no one ever noticed the option.

PR #6229, opened 2026-01-05
Fallback to default when cronet providers are not available

TimoPtr:
## Summary

While checking the crashes on Sentry for the new release 2026.1.1 I found a new crash on a Samsung SM-T585 (Galaxy Tab A from 2016) running Lineage. Where the Cronet providers seems to be available but disabled according to the error message "All available Cronet providers are disabled. A provider should be enabled before it can be used." and [Cronet code](https://cs.android.com/android/platform/superproject/+/5750195121e36bb55b446b77c2d212a5770b2e43:external/cronet/stable/components/cronet/android/api/src/org/chromium/net/CronetEngine.java;l=660)

To avoid crashes I propose that we fallback to the default http provider even if it means that we are losing some features like Quic support. 

## Checklist


jpelgrom:
For anyone looking for where enabling QUIC support went like me, it was superfluous: https://developer.android.com/develop/connectivity/cronet/reference/org/chromium/net/CronetEngine.Builder.html#enableQuic(boolean)

> Sets whether [QUIC](https://www.chromium.org/quic) protocol is enabled. **Defaults to enabled.**

jpelgrom (approved):

PR #6232, opened 2026-01-05
Cleanup and add test for the database

TimoPtr:
## Summary

Move the migration into a dedicated file to make it easier to see the database configuration. Add instrumentation test for the migration, for now it only tests migration from 24 to latest. Later we should add individual tests for migrations to verify the behavior.

## Checklist


common/src/main/kotlin/io/homeassistant/companion/android/database/migration/DatabaseMigration.kt
```diff
@@ -0,0 +1,862 @@
...
+    RenameTable(
+        fromTableName = "mediaplayctrls_widgets",
+        toTableName = "media_player_controls_widgets",
+    ),
+)
+internal class Migration36to37 : AutoMigrationSpec
+
+private class Migration37to38(private val context: Context) : Migration(37, 38) {
+    /**
+     * Migrate code has been taken out of an autogenerated migration to be able to pass the appContext to the constructor
```
jpelgrom: I don't like this, even if it matches autogenerated

Which constructor?

```suggestion
     * Migrate code has been taken out of an autogenerated migration, to be able to use the context in onPostMigrate
```

TimoPtr: The issue is that you can't have a context if you use a Autogenerated migration, because it uses the default constructor without param while generating the code of the migration.

TimoPtr:
> To be honest I'm not convinced this is all a cleanup, everything you moved was already at the end of the file. The automated test is nice though.

Well I wanted to have only the database definition in the file to make it cleared. Migration is something else IMO, I hesitated to create a file for each migration, but I decided that for now it's fine like this.

For large migration or complex one I would like one file and one associated test for instance.

PR #6233, opened 2026-01-05
Add missing runconfigurations

TimoPtr:
## Summary

While opening the project from a worktree I found out that :app was missing from the runConfiguration. I had to add it manually. To make the first steps into the project easier the app configuration should always be there. I added automotive and wear too just in case.

### DIFF
--- a/.idea/runConfigurations/app.xml
+++ b/.idea/runConfigurations/app.xml
@@ -1,0 +1,74 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="app" type="AndroidRunConfigurationType" factoryName="Android App" activateToolWindowBeforeRun="false">
+    <module name="home-assistant-android.app" />
+    <option name="ANDROID_RUN_CONFIGURATION_SCHEMA_VERSION" value="1" />
+    <option name="DEPLOY" value="true" />
+    <option name="DEPLOY_APK_FROM_BUNDLE" value="false" />
+    <option name="DEPLOY_AS_INSTANT" value="false" />
+    <option name="ARTIFACT_NAME" value="" />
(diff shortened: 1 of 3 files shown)


jpelgrom:
This seems to match defaults for what Android Studio generates so might be OK, but run configurations are personal (I like to toggle on clear logcat sometimes) and they must not be saving these as a file by default for a good reason. A project sync should make configurations show up and is needed most of the time anyway.

TimoPtr:
> This seems to match defaults for what Android Studio generates so might be OK, but run configurations are personal (I like to toggle on clear logcat sometimes) and they must not be saving these as a file by default for a good reason. A project sync should make configurations show up and is needed most of the time anyway.

Well I think that since I've added `dev playground` config when it syncs for the first time in a new folder it might not create all the other default. I agree with you but in your case you can simply create new runConfiguration with a different name (that's what I do I have many different config that I don't share)

jpelgrom:
> Well I think that since I've added `dev playground` config when it syncs for the first time in a new folder it might not create all the other default.

Dev playground is not a default though. I've cloned the repo to a new folder and let Android Studio sync and it created all the others for me. It may just have been Android Studio being stupid, it happens from time to time :)

TimoPtr:
> > Well I think that since I've added `dev playground` config when it syncs for the first time in a new folder it might not create all the other default.
> 
> Dev playground is not a default though. I've cloned the repo to a new folder and let Android Studio sync and it created all the others for me. It may just have been Android Studio being stupid, it happens from time to time :)

Do you have any objections merging this? I think it would cover this cases where AS is not generating the config automatically.

jpelgrom:
No strong objections, but we should stay alert for any feedback from other devs for whether or not it is helpful.

PR #6234, opened 2026-01-05
Remove appdatabase get instance

### LINKED ISSUE #5702: Avoid AppDatabase.getInstance to be use outside of Hilt
AppDatabase.getInstance is thread safe with a synchronized block to avoid having multiple thread creating multiple instance of the database and overriding the `instance` static attribute. 

The issue is that if multiple actor call getInstance all of them are going to be blocked by the synchronized block if the instance is not yet created and it leads to ANR.

We should avoid this and only have Hilt invoking this function. It means a huge work to adapt the codebase to always rely on Hilt to get the database never using `getInstance`, by doing so we ensure we won't ever have an ANR again on this function.

TimoPtr:
## Summary

Remove the need of `AppDatabase.getInstance` and removing the synchronization block from it. It might help reduce some ANR.

## Checklist

## Any other notes

Closes #5702

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/database/migration/Utils.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/database/migration/Utils.kt
@@ -5,20 +5,14 @@ import android.app.NotificationManager
 import android.content.Context
 import android.database.Cursor
 import android.os.Build
-import android.os.Handler
-import android.os.Looper
-import android.widget.Toast
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
 import androidx.core.content.getSystemService
 import io.homeassistant.companion.android.common.R as commonR
 import io.homeassistant.companion.android.common.util.CHANNEL_DATABASE
-import io.homeassistant.companion.android.database.AppDatabase.Companion.TAG
-import io.homeassistant.companion.android.database.AppDatabase.Companion.integrationRepository
-import kotlinx.coroutines.runBlocking
-import timber.log.Timber
 
 private const val NOTIFICATION_ID = 45
+private const val TAG = "AppDatabase"
 
 internal fun <T> Cursor.map(transform: (Cursor) -> T): List<T> {
     return if (moveToFirst()) {
@@ -57,20 +51,5 @@ internal fun notifyMigrationFailed(context: Context) {
         .build()
     with(NotificationManagerCompat.from(context)) {
         notify(NOTIFICATION_ID, notification)
-    }
-    runBlocking {
-        try {
-            integrationRepository.fireEvent("mobile_app.migration_failed", mapOf())
-            Timber.d("Event sent to Home Assistant")
-        } catch (e: Exception) {
-            Timber.e(e, "Unable to send event to Home Assistant")
-            Handler(Looper.getMainLooper()).post {
-                Toast.makeText(
-                    context,
-                    commonR.string.database_event_failure,
-                    Toast.LENGTH_LONG,
-                ).show()
-            }
-        }
     }
 }


common/src/main/kotlin/io/homeassistant/companion/android/database/migration/Utils.kt
```diff
@@ -58,19 +52,4 @@ internal fun notifyMigrationFailed(context: Context) {
...
-            Timber.e(e, "Unable to send event to Home Assistant")
-            Handler(Looper.getMainLooper()).post {
-                Toast.makeText(
-                    context,
-                    commonR.string.database_event_failure,
-                    Toast.LENGTH_LONG,
-                ).show()
-            }
-        }
-    }
```
TimoPtr, line 75: After checking the [code](https://github.com/home-assistant/android/commit/7ce1e9f5a4ea79f73ddeb8309c2d0255a9edaef2) I think this it never worked. The `integrationRepository` was never set in the companion object of the database.

I think it was an attempt to notify the server that migration failed, I don't think we should do that.

jpelgrom: Maybe it worked before I added multiserver support, but agreed that we shouldn't try to do this so OK to remove for me as well 👍

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -877,12 +875,12 @@ class LocationSensorManager :
             }
         }
 
-        val sensorDao = AppDatabase.getInstance(latestContext).sensorDao()
-        val sensorSettings = sensorDao.getSettings(zoneLocation.id)
+        val zoneDao = sensorDao(latestContext)
```
jpelgrom: This is not a DAO for zones, renaming it creates confusion and we should stick to the original name.

wear/src/main/kotlin/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
```diff
@@ -281,10 +279,10 @@ class HealthServicesSensorManager : SensorManager {
             }
 
             override fun onPermissionLost() {
-                val sensorDao = AppDatabase.getInstance(latestContext).sensorDao()
                 runBlocking {
+                    val dao = sensorDao(latestContext)
```
jpelgrom: Why move this inside runBlocking?

jpelgrom (commented):
Sometimes you're using `val dao = ...`, sometimes it is `val sensorDao = ...` (example). Try to be consistent. My preference would be to use a specific name (`sensorDao`), reducing changed lines, or at least use the specific name if any other DAO is accessed in the same file to prevent mixups.

wear/src/main/kotlin/io/homeassistant/companion/android/sensors/HealthServicesSensorManager.kt
TimoPtr: No reason ... Not gonna lie most of the changes has been made by Claude. I'm going to go over everything to check the name 👍🏻

.idea/kotlinc.xml
```diff
@@ -6,10 +6,6 @@
   <component name="Kotlin2JvmCompilerArguments">
     <option name="jvmTarget" value="11" />
   </component>
-  <component name="KotlinCommonCompilerArguments">
-    <option name="apiVersion" value="2.3" />
-    <option name="languageVersion" value="2.3" />
-  </component>
```
jpelgrom, line 12: Unrelated change

TimoPtr: Merged too fast ...

common/src/main/kotlin/io/homeassistant/companion/android/database/migration/Utils.kt
TimoPtr: Well I've checked the initial commit and it was not working too since it was not setting the variable.

PR #6239, opened 2026-01-07
Improve websocket connection logic (threading)

TimoPtr:
## Summary

This PR changes the threading logic we have in connection to properly react on URLChanges. It tries to solves cases where we were running into a dead lock where it was not possible to connect anymore. 

There was also the case of getting a new URL while authenticating, it was possible to be connected to the first URL where we should have stop the connection right away. 

The logic is quite complex, I tried to move the update of the state into the `WebsocketListener` callbacks to avoid threading issue, but connect might also update it when the websocket is not yet open.

Because I'm making the connection within the collectLatest of the urlFlow, I had to store a `pendingConnectDeferred` protected by the mutex.

I've decided to keep the connection and some other useful object into a ConnectionHolder to help handling the cleanup properly.

The changes regarding the state to use a sealed interface, was initially an attempt to create a state machine but I failed. I still kept the sealed class because it might be helpful later and I use it to remove the null case.

handleClosingSocket now looks for the holder value to see if it should do anything of let connect handle the connection.

## Checklist

## Any other notes

I removed the `FailFast` in onEvent since the server while running Assist is still sending an even after the unsubscribe.

### DIFF
--- a/common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
+++ b/common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
@@ -40,6 +40,7 @@ import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.emptyFlow
 import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.test.TestScope
 import kotlinx.coroutines.test.UnconfinedTestDispatcher


common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -236,130 +292,255 @@ internal class WebSocketCoreImpl(
...
-                            Timber.w("URL changed, disconnecting immediately.")
-                        }
-                        // Set state before cancel() since cancel triggers onFailure -> handleClosingSocket
-                        connectionState = WebSocketState.CLOSED_URL_CHANGE
-                        connection?.cancel()
-                    }
+    private suspend fun attemptWebSocketConnection(url: URL): ConnectionAttemptResult {
+        var webSocket: WebSocket? = null
+        try {
+            Timber.d("Creating WebSocket connection")
```
jpelgrom, line 436: Maybe remove this because you can see messages going back/forth in the logs, or an error

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -490,22 +706,38 @@ internal class WebSocketCoreImpl(
 
     override fun onMessage(webSocket: WebSocket, bytes: ByteString) {
         Timber.d("Websocket: onMessage (bytes)")
+        if (isStaleConnection(webSocket)) {
+            Timber.w("Ignoring onMessage (bytes) from stale connection")
```
jpelgrom: Messages from this callback are always ignored

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
TimoPtr: I would like to keep this one since the other logs could come from somewhere else if you don't mind. And it's something I want to watch in the logs to see how many times it happens. This one shouldn't be verbose.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
TimoPtr: indeed removing completely

jpelgrom (approved):

PR #6264, opened 2026-01-16
Add missing catch when interacting with network

TimoPtr:
## Summary

This PR adds missing try/catch in some places. It is a best effort to avoid crashing the app, but it is not a good practice. We should properly categorize the exception and depending on what they are we might still need to crash the app. Because of this changes it might take us longer to debug an issue. Since this pattern is already used everywhere in the app let's stick with it for now.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistViewModel.kt
@@ -174,7 +174,7 @@ class AssistViewModel @Inject constructor(
                 } catch (e: CancellationException) {
                     throw e
                 } catch (e: Exception) {
-                    Timber.e(e, "Fail to load assist pipelines")
+                    Timber.e(e, "Failed to load assist pipelines")
                     null
                 }
 
@@ -208,12 +208,10 @@ class AssistViewModel @Inject constructor(
             allPipelines[selectedServerId]?.firstOrNull { it.id == id }
                 ?: try {
                     serverManager.webSocketRepository(selectedServerId).getAssistPipeline(id)
-                } catch (
-                    e: CancellationException,
-                ) {
+                } catch (e: CancellationException) {
                     throw e
                 } catch (e: Exception) {
-                    Timber.e(e, "Fail to get assist pipeline")
+                    Timber.e(e, "Failed to get assist pipeline")
                     null
                 }
         selectedPipeline?.let {
@@ -228,7 +226,7 @@ class AssistViewModel @Inject constructor(
             } catch (e: CancellationException) {
                 throw e
             } catch (e: Exception) {
-                Timber.e(e, "Fail to set last used pipeline")
+                Timber.e(e, "Failed to set last used pipeline")
             }
 
             _conversation.clear()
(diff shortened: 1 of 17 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/widgets/mediaplayer/MediaPlayerControlsWidget.kt
```diff
@@ -506,7 +507,15 @@ class MediaPlayerControlsWidget : BaseWidgetProvider<MediaPlayerControlsWidgetEn
 
             val actionDataMap: HashMap<String, Any> = hashMapOf("entity_id" to entityId)
 
-            serverManager.integrationRepository().callAction(MEDIA_PLAYER_DOMAIN, action, actionDataMap)
```
TimoPtr, line 509: This one was missing the serverID

app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistViewModel.kt
```diff
@@ -197,15 +206,30 @@ class AssistViewModel @Inject constructor(
     private suspend fun setPipeline(id: String?) {
         selectedPipeline =
             allPipelines[selectedServerId]?.firstOrNull { it.id == id }
-                ?: serverManager.webSocketRepository(selectedServerId).getAssistPipeline(id)
+                ?: try {
+                    serverManager.webSocketRepository(selectedServerId).getAssistPipeline(id)
+                } catch (
+                    e: CancellationException,
+                ) {
```
jpelgrom:
```suggestion
                } catch (e: CancellationException) {
```

wear/src/main/kotlin/io/homeassistant/companion/android/conversation/ConversationViewModel.kt
```diff
@@ -100,14 +101,25 @@ class ConversationViewModel @Inject constructor(
...
+                } catch (e: Exception) {
+                    Timber.e(e, "Fail to load pipelines")
                 }
             }
 
             return setPipeline(
                 if (useAssistPipeline) {
-                    serverManager.integrationRepository().getLastUsedPipelineId()
+                    try {
+                        serverManager.integrationRepository().getLastUsedPipelineId()
```
jpelgrom, line 116: This reads from local shared preferences and is very unlikely to fail, do we need it here?

TimoPtr: Well as of today it can come from `integrationRepository()` if they are no server.

TimoPtr:
> Every "Fail to" should be "Failed to"
> 
> I understand catching specific exceptions is better than catching everything, but catching everything is still better than catching nothing, no? Fail gracefully where possible.
> 
>     * In calls to `IntegrationRepository`, you could limit the final catch to `IntegrationException` which is used for most webhook exceptions to limit what is caught.
> 
>     * In calls to `WebSocketRepository` it will usually catch and return null from the repository instead of propagating the exception; it could be useful to catch around the entire repository function but what do you expect to catch there?

I don't want to make this kind of changes in this PR, it needs to be work on and tested intensively.

jpelgrom:
> I don't want to make this kind of changes in this PR, it needs to be work on and tested intensively.

I was mainly talking about what you were adding, not suggesting changing all the other usages. However I can understand consistently applying it first, then adjusting everything at once later.

PR #6265, opened 2026-01-16
Cleanup Active server if ID is not in database

TimoPtr:
## Summary

I'm trying to understand in which scenario we are giving an ID that doesn't exist anymore, while investigating this I had this idea that maybe we storing in the local storage the active server ID that doesn't exist anymore. I'm not sure if it possible but in case we could take this defensive approach to clear the local preference if the ID does not exist to unlock the situation on the next run.

## Checklist

## Any other notes

I'm looking for other places, but the crashes that I see on sentry might also be because of old data that we didn't cleanup and that we are never cleaning.

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManagerImpl.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManagerImpl.kt
@@ -168,8 +168,14 @@ internal class ServerManagerImpl @Inject constructor(
 
     private suspend fun getServerIdSanitize(serverId: Int): Int? {
         return if (serverId == SERVER_ID_ACTIVE) {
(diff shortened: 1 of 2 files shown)


jpelgrom:
> I had this idea that maybe we storing in the local storage the active server ID that doesn't exist anymore

It shouldn't be, whenever the server is deleted the same check you're adding here is applied to remove it:

https://github.com/home-assistant/android/blob/a99d08ed18981f5e1a6fb85f75a2a17d16093cc3/common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerManagerImpl.kt#L143

If you can point to an issue where it indicates there is an old server ID stored as active I'm OK with this change, but otherwise I would like to not include it so we know if things go wrong.

jpelgrom:
Some [additional discussion on Discord](https://discord.com/channels/330944238910963714/1346948551892009101/1462537700560339139) for future reference

PR #6267, opened 2026-01-16
Remove WidgetEntity when server is removed

TimoPtr:
## Summary

On Sentry we see some errors of server not found coming from the widget. Today we are not removing the widget entities when we remove a server. This PR does this, even if it doesn't remove the widget it is going to prevent a widget to try to use a serverID that doesn't exist anymore and would crash the app.

If a user tries to interact with a widget that has a deleted server now it won't do anything now except showing a toast or an error on the widget.

## Checklist

## Any other notes

I tried to add a test that verify that we don't forget to add new WidgetDao inside the list of widgetDao. It uses the classloader to find all the sub-interfaces of WidgetDao.

### DIFF
--- a/app/src/test/kotlin/io/homeassistant/companion/android/widgets/WidgetsDaoTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/widgets/WidgetsDaoTest.kt
@@ -1,0 +1,105 @@
+package io.homeassistant.companion.android.widgets
+
+import dagger.hilt.android.testing.HiltAndroidRule
+import dagger.hilt.android.testing.HiltAndroidTest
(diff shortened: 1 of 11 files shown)


jpelgrom:
As shortly discussed on Discord, the goal for multiserver was to keep data that can be reconfigured on server delete, and only delete data that cannot be reused.

For widgets, they would error but editing should allow you to change the server and keep using them. Maybe the widget code needs to be improved to fail early instead of throwing exceptions. But deleting the row from the database completely would result in the user having to manually remove (and re-add) widgets, which can be time-consuming and confusing.

TimoPtr:
Closing since #6264 should catch most the error of call made to a removed server.

PR #6282, opened 2026-01-19
Catching wrong exception in ServerSettingsPresenter

TimoPtr:
## Summary

Screenshot 2026-01-19 at 08 29 00

After the recent changes it turns out that we are catching the wrong exception while trying to setAppActive from the settings. This PR simply avoid the crash but doesn't fix the underlying issue that we should not try in the first place to setAppActive with an ID that doesn't exist. Which tells me that we should still proactively cleanup whatever that could lead to this scenario, probably with #6265.


jpelgrom:
> recent changes

To be more specific, it looks like it is the result of #6212

PR #6290, opened 2026-01-20
Catch open URI to avoid crashing the app

TimoPtr:
## Summary

I found on Sentry a strange error where people that are using Cloudflare zero trust on top of their instance while doing the onboarding have a crash when it tries to open the address to the team login. The error is that the system return that it doesn't find an activity to open the URI. This PR introduce a new way to open a URI with a fallback to a snackbar instead of crashing the app.

## Checklist

## Any other notes

We could probably replace the existing navigate with this new one. @jpelgrom any opinion on that?


common/src/main/res/values/strings.xml
```diff
@@ -1431,4 +1431,5 @@
     <string name="connection_check_home_assistant">Home Assistant verification</string>
     <string name="connection_check_home_assistant_success">Verified as Home Assistant</string>
     <string name="connection_check_error_not_home_assistant">Server is not Home Assistant</string>
+    <string name="fail_to_navigate_to_uri">Fail to open %s</string>
```
jpelgrom: Make it more obvious that HA tries to open another app, not that loading failed

```suggestion
    <string name="fail_to_navigate_to_uri">No app available to open %s</string>
```

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/NavControllerExt.kt
```diff
@@ -2,7 +2,22 @@ package io.homeassistant.companion.android.util.compose
...
+
+suspend fun NavController.navigateToUriCatching(
+    uri: String,
+    onShowSnackbar: suspend (message: String, action: String?) -> Boolean,
+) {
+    try {
+        AndroidUriHandler(context).openUri(uri)
+    } catch (e: IllegalArgumentException) {
+        // Don't log e to not leak the URL in the log
+        Timber.e("Fail to navigate to uri")
```
jpelgrom:
```suggestion
        Timber.e("Failed to navigate to uri")
```

TimoPtr: God I'm always making it ...

jpelgrom:
> We could probably replace the existing navigate with this new one. @jpelgrom any opinion on that?

I guess that makes sense, as it should avoid coming across the same issue again in the future. We do need to make sure to have something able to show snackbars from everywhere it's used.

jpelgrom (approved):

TimoPtr:
> Cloudflare Access is a webpage (https) so they don't have a browser app? Unlikely but OK.

It's kinda weird to be honest but at least we are not crashing anymore.

PR #6292, opened 2026-01-20
Update status/navigation bar colors on page change for insets

### LINKED ISSUE #6270: White screens causing time/status to be invisible
**Home Assistant Android app version(s):** 2026.1.3-full (19812)

**Android version(s):** 16 (BD3A.251105.010.E1) - latest build 

**Device model(s):** Pixel 10 Pro

**Home Assistant version:** 2026.1.1

I've noticed in a recent release (but prior to the latest one) that white screens - settings - leave the phones time, reception etc (anything that's typically visible above the outside of the app) are invisible.

Image

Image

My guess is that, in light mode, android is picking up the default theme colour, and the home assistant primary colour is a darkish blue (I can't pick the hex value as I'm on mobile ATM). So android decides that when the app launches, a black text on the primary colour (used in the header bar of non-setting pages) has a contrast too low, so it changes the text colours to white - thus it's invisible on the settings page.

In the screenshot I've included a dark mode on my phone and you can see the time on the phone is still visible, which is why I suspect contrast. Additionally, I tested by changing the theme to a lighter blue (which I think is a pretty bad contrast against white), and this also "fixed" the time appearing (included in screenshots).

Image

Image

The reason I've raised this as a (possible) bug, and one that the android app might be able to control, is that the Companion screen does actually have the phone's time visible - so I'm guessing this page is doing something slightly different which might help resolve the issue. 

I should also note that I've tested this with both "Fullscreen" (on the Companion settings, enabled and disabled and it makes no difference. 

Image

**Additional information:**

Sorry for all the screenshots - it was tricky to describe!

jpelgrom:
## Summary

This PR adds a new Javascript event listener for navigation events, and updates the status/navigation bar colors on page changes for servers with insets. With older servers or devices without a recent WebView, this will simply do nothing.

This fixes the issue where any settings page would result in invisible notifications, time, etc. with the default HA light theme.

 - Fixes #6270
 - Does not yet fix issue with the navigation drawer (I don't think we can fix this without a frontend change - can't have different colors for left/right)
 - Does not yet fix more info dialogs (the new listener is fired for this because it is navigation, just need to find a reliable way to determine the color with JS).

## Checklist

## Screenshots

Note how the status bar changes between white/black in test, but not in a release build (~2026.1.4):

https://github.com/user-attachments/assets/a61be834-86b1-497c-b785-38a320e0d64a

n/a

## Any other notes

 - I've also removed comments about deprecation because this is now relevant and the frontend still [appears to send the event ](https://github.com/home-assistant/frontend/blob/e4d6f3c9d72c58b3a1ad47af39291d3dcd994533/src/panels/profile/ha-pick-theme-row.ts#L232) that was mentioned as deprecated.
 - Logging was minimized to avoid cluttering the logs too much

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -827,13 +827,14 @@ class WebViewActivity :
             removeJavascriptInterface(javascriptInterface)
             addJavascriptInterface(
                 object : Any() {
-                    // TODO This feature is deprecated and should be removed after 2022.6
                     @JavascriptInterface
-                    fun onHomeAssistantSetTheme() {
+                    fun onHomeAssistantSetTheme(onlyForInsets: Boolean) {
                         // We need to launch the getAndSetStatusBarNavigationBarColors in another thread, because otherwise the evaluateJavascript inside the method
                         // will not trigger it's callback method :/
                         lifecycleScope.launch(Dispatchers.Main) {
-                            getAndSetStatusBarNavigationBarColors()
+                            if (!onlyForInsets || serverHandleInsets.value) {
+                                getAndSetStatusBarNavigationBarColors()
+                            }
                         }
                     }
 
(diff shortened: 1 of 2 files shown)


TimoPtr:
Closing in favor of disabling the edge to edge #6296 the frontend is going to send us the mode to use instead of us reading it to be more future proof.

PR #6295, opened 2026-01-21
Make sure that widgets don't crash the app when server is removed

TimoPtr:
## Summary

There was some leftover missing some catching for the widget. I did add all the widget on my phone and then remove the server and see what happens and found some missing logic.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetStateUpdater.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetStateUpdater.kt
@@ -49,6 +49,11 @@ internal class TodoWidgetStateUpdater @Inject constructor(
     }
 
     private suspend fun getAndSubscribeEntityUpdates(serverId: Int, listEntityId: String): Flow<Entity?>? {
+        if (serverManager.getServer(serverId) == null) {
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/widgets/todo/TodoWidgetStateUpdater.kt
```diff
@@ -100,22 +100,27 @@ internal class TodoWidgetStateUpdater @Inject constructor(
...
-                            entityName = entity.friendlyName,
-                            todos = todos,
-                        ),
-                    )
-                    TodoStateWithData.from(todoEntity, entity, todos)
-                } ?: flowOf(TodoStateWithData.from(todoEntity))
+                if (serverManager.getServer(serverId) == null) {
+                    Timber.w("Server has been removed and the widget needs to be reconfigured")
+                    flowOf(TodoStateWithData.from(todoEntity))
+                } else {
```
jpelgrom: Why not move this inside `getAndSubscribeEntityUpdates` (as the first line: if server is null return null)? That already can return `null`, keep the code simpler here.

TimoPtr: Good catch

PR #6296, opened 2026-01-21
Disable edge to edge for the Webview

TimoPtr:
## Summary

We are not able yet to properly decided of the mode applied to the statusbar we need the frontend to tell us to use light or dark mode depending on the page rendered. Otherwise we might not see the icon and hour on the status bar as reported in #6270.

@jpelgrom tried to address it in https://github.com/home-assistant/android/pull/6292 but this approach is only fixing the issue on some page and the frontend could break this at any moment. The best would be a message in the externalBus from the frontend.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
@@ -226,7 +226,7 @@ class WebViewPresenterImpl @Inject constructor(
                         url = urlWithAuth,
                         keepHistory = !isNewServer,
                         openInApp = it.baseIsEqual(baseUrl),
-                        // We need the frontend to notify us of the mode to use for the status bar https://github.com/home-assistant/android/issues/6270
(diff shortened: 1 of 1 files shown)


jpelgrom:
> The best would be a message in the externalBus from the frontend.

I'm in favor of a 'proper' solution, but is there a related frontend issue other than the main insets one to properly track it (which does not yet mention this) / did you use your OHF powers to ask a frontend dev to work on this ;) ?

TimoPtr:
> did you use your OHF powers to ask a frontend dev to work on this 

Yes I did, we are discussing how to achieve it, I don't know when it is going to be done. I need to create an issue on the frontend (I might do it tomorrow)

jpelgrom:
I did find home-assistant/frontend#29111 which should at least make it easier to use one color all the time (with default theme)

jpelgrom:
> I need to create an issue on the frontend (I might do it tomorrow)

In that case I think it's better to link to that issue in your comment. Otherwise OK to merge, I'll close my PR after.

PR #6312, opened 2026-01-22
Implement microwakeword detection

TimoPtr:
## Summary

Inspired from ESPHome [implementation](https://github.com/esphome/esphome/tree/4ac7fe84b41fe383d681339e900228b0817a6e98/esphome/components/micro_wake_word) and  from the great work from @brownard in https://github.com/brownard/Ava. 

I've rework the C++ part to not duplicate the source in our codebase and tweak the JNI to have a clearer separation of concern.

The Kotlin part has been generated with the help of Claude. For now it's only accessible through the DevPlaygound in debug. 

We should not merge this in main until we have a proper plan on how to enable/disable this.

I've added a small instrumentation test that is going to be helpful verify that bumping NDK won't break the version we target.

I've enabled the compiler optimization also in debug since we want debug to behave the same as release in our case and the native code is small.

## Checklist


TimoPtr:
@brownard I want to thank you again for your project that speed up a lot this first implementation on the official companion app. If you want to take a look or even contribute you are more than welcome. 

By being in the companion app, the authentication is already handle and we could also expose some commands to start/stop listening to save battery.

brownard:
I'm glad my work has been useful. Ultimately the companion app is the best place for this so it's great that it's been picked up here.

synesthesiam:
Looks good as a first pass! I would suggest at some point loading the settings for each wake word model from their respective config files, e.g. https://github.com/OHF-Voice/micro-wake-word/releases/download/okay_nabu_20241226.3/okay_nabu.json

You may also wake to include options for adjust the wake word sensitivity: https://github.com/esphome/home-assistant-voice-pe/blob/a379b8c5c1a35eeebc8f9925c19aab68743517a4/home-assistant-voice.yaml#L1775

jpelgrom:
Nice to see how the local wake word work pays off, so much cleaner than when I worked on this back when Assist was new.

What's the idea behind the `microfrontend` name? 'Frontend' suggest the HA frontend user interface to me which is unrelated.

Should the service ignore input while AssistActivity is listening?

If you're going ahead with this setup, don't forget to update dev docs to indicate you need to install CMake for building (at least it didn't do so automatically for me on Windows - after that building worked fine!).

gradle/libs.versions.toml
```diff
@@ -6,6 +6,7 @@ accompanist = "0.37.3"
 activity-compose = "1.12.2"
 androidBeaconLibrary = "2.21.2"
 androidGradlePlugin = "8.13.2"
+andrdoiNdk = "29.0.14206865"
```
jpelgrom:
```suggestion
androidNdk = "29.0.14206865"
```

microfrontend/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
+plugins {
+    alias(libs.plugins.android.library)
+    alias(libs.plugins.homeassistant.android.common)
+}
+
+android {
+    namespace = "io.homeassistant.companion.android.microfrontend"
+    ndkVersion = libs.versions.andrdoiNdk.get()
```
jpelgrom:
```suggestion
    ndkVersion = libs.versions.androidNdk.get()
```

settings.gradle.kts
jpelgrom, line 1: Version generation code commented out in this file, why?

TimoPtr:
> What's the idea behind the microfrontend name? 'Frontend' suggest the HA frontend user interface to me which is unrelated.

It's the name of the TFlite lib we use https://github.com/tensorflow/tflite-micro/tree/main/tensorflow/lite/experimental/microfrontend.

> Should the service ignore input while AssistActivity is listening?

From my initial test it seems that the OS is simply stop feeding our service when something on top is asking to the mic. I tried for instance with taking a video and it did not trigger the wake word detection.

settings.gradle.kts
TimoPtr: 🤦🏻 I need to provide a fix to our setup because when you are in a git worktree this piece of code prevent the app to sync. Let me do it know otherwise I'm going to keep commit this again and again by mistake.

jpelgrom: Handled in #6331

TimoPtr:
Merging this would have an impact on the size of the app without adding anything since the feature is only available through the DevPlayground, but the following PR will expose the feature properly from the settings.

TimoPtr:
> Looks good as a first pass! I would suggest at some point loading the settings for each wake word model from their respective config files, e.g. https://github.com/OHF-Voice/micro-wake-word/releases/download/okay_nabu_20241226.3/okay_nabu.json

@synesthesiam done, for the second part I'm going to do it in another PR. Feel free to look at the code again it is in a mergeable state.

Same for @brownard if you want to double check. Following PR will add the UI to enable the detection.

app/src/androidTest/kotlin/io/homeassistant/companion/android/assist/wakeword/MicroWakeWordModelTest.kt
```diff
@@ -0,0 +1,106 @@
...
+class MicroWakeWordModelTest {
+
+    private val appContext: Context
+        get() = InstrumentationRegistry.getInstrumentation().targetContext
+
+    /**
+     * Initializes TfLite and returns true if successful.
+     * Returns false if initialization fails (e.g., full flavor without GMS).
+     */
+    private suspend fun tryInitializeTfLite(): Boolean {
```
jpelgrom, line 25: Any specific reason not to use given/when/then here?

app/src/main/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionService.kt
```diff
@@ -0,0 +1,274 @@
...
+import kotlin.time.Instant
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.launch
+import timber.log.Timber
+
+/**
+ * Voice interaction service that allows the app to function as the system's default assistant.
```
jpelgrom: The app can also be the default assistant without this service.

```suggestion
 * Voice interaction service used when the app is set as the system's default assistant.
```

app/src/main/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionService.kt
```diff
@@ -0,0 +1,274 @@
...
+        serviceScope.launch {
+            val model = loadWakeWordModel()
+            wakeWordListener.start(this, model)
+        }
+    }
+
+    private suspend fun loadWakeWordModel(): MicroWakeWordModelConfig {
+        // TODO: Allow user to select which wake word model to use
+        // TODO: Allow user to set sensibility https://github.com/esphome/home-assistant-voice-pe/blob/a379b8c5c1a35eeebc8f9925c19aab68743517a4/home-assistant-voice.yaml#L1775
+        // TODO: When to start/stop listening
```
jpelgrom, line 103: Do you want to merge with TODOs or create issues / keep this feature only for debug (if only for debug it needs a check here)?

app/src/main/kotlin/io/homeassistant/companion/android/assist/wakeword/MicroWakeWord.kt
```diff
@@ -0,0 +1,237 @@
...
+        private const val OUTPUT_SCALE = 0.00390625f
+
+        // Expected tensor shapes: [batch, stride, features] and [batch, probability]
+        private val EXPECTED_INPUT_SHAPE = intArrayOf(1, 3, 40)
+        private val EXPECTED_OUTPUT_SHAPE = intArrayOf(1, 1)
+
+        /**
+         * Create a new MicroWakeWord instance.
+         *
+         * This is a suspend function to allow cancellation during model loading.
```
jpelgrom: You don't check for cancellation anywhere in this function or the function it calls though, so it won't actually cancel early.

common/src/main/res/values/strings.xml
```diff
@@ -1193,6 +1193,10 @@
     <string name="assist_shortcut">Assist shortcut</string>
     <string name="assist_start_listening">Start listening</string>
     <string name="assist_stop_listening">Stop listening</string>
+    <string name="assist_listening_title">Listening for activation</string>
```
jpelgrom: Everyone is going to call this wake word, why make it more complicated?

```suggestion
    <string name="assist_listening_title">Listening for wake word</string>
```

common/src/main/res/values/strings.xml
```diff
@@ -1193,6 +1193,10 @@
...
+    <string name="assist_listening_wakeword">Say \"%s\" to activate Assist</string>
+    <string name="assist_listening_channel">Assist activation</string>
```
```suggestion
    <string name="assist_listening_channel">Assist wake word</string>
```

jpelgrom (commented):
There are debug log statement for almost all override functions, we probably want to reduce that.

For C++ code with source in ESPHome/OHF-Voice: how are we going to keep those updated when changes are made?

app/src/main/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionService.kt
TimoPtr: To make it easier this time I would like to go with merging those, especially because I've addressed most of them already in the follow up PR and removed it there.

app/src/androidTest/kotlin/io/homeassistant/companion/android/assist/wakeword/MicroWakeWordModelTest.kt
TimoPtr: To be fair it is lazyness because we can't use sentence in names of test function in Instrumentation tests.

TimoPtr:
> There are debug log statement for almost all override functions, we probably want to reduce that.
> 
> For C++ code with source in ESPHome/OHF-Voice: how are we going to keep those updated when changes are made?

The C++ code within the repo is a very thin layer on top of TFLite micro code that's why compare to AVA I've decided to not embedded the source but instead download it.

The part that we might need actually to keep in sync is the small Kotlin layer that calls TF and the microfrontend, that kinda act as an orchestrator. Unfortunately at the moment we need to do this manually. This is already the case also for other projects like the linux micro wake word. Ultimately this layer is quite small and should be easy to maintain. The most complicated part is the model itself where we do use TF.

I'll do some cleanup in the logs 👍🏻 it was very useful to debug.

app/src/main/kotlin/io/homeassistant/companion/android/assist/wakeword/MicroWakeWord.kt
TimoPtr: I removed the mention but we do check `ensureActive()` after we loaded the models and before creating the Interpreter.

TimoPtr:
In the end after running the detection I think the logs are still relevant, especially in the early stage of this feature. I don't log anything in the hot-path so it's not spamming the logs IMO.

app/src/androidTest/kotlin/io/homeassistant/companion/android/assist/wakeword/MicroWakeWordModelTest.kt
jpelgrom: OK for instrumentation, but let's try to stick to it in other places as I've started liking it :)

TimoPtr: Fully agreed ! These tests are kinda special in a way and super expensive to run so let's keep them to the minimum

app/src/main/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionService.kt
```diff
@@ -0,0 +1,274 @@
...
+    override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
+        when (intent?.action) {
+            ACTION_START_LISTENING -> startListening()
+            ACTION_STOP_LISTENING -> stopListening()
+        }
+        return super.onStartCommand(intent, flags, startId)
+    }
+
+    companion object {
+        private const val CHANNEL_ID = "assist_listening"
```
jpelgrom: This should be listed in AppNotifChannels.kt in common instead for consistency and because of the notifications channels UI in the app that uses the list in that file.

build-logic/convention/src/main/kotlin/AndroidApplicationDependenciesConventionPlugin.kt
```diff
@@ -26,6 +26,7 @@ class AndroidApplicationDependenciesConventionPlugin : Plugin<Project> {
             extensions.getByType<ApplicationExtension>().apply {
                 dependencies {
                     "implementation"(project(":common"))
+                    "implementation"(project(":microfrontend"))
```
jpelgrom, line 29: You aren't adding the service in the manifest for Automotive, which is what makes the service work consistently. Shouldn't this be in the `app` module specific dependencies instead?

TimoPtr: I think I want the service in automotive to test how it behaves no?

TimoPtr: Could you tell or create an issue about why we don't have assist offered as default in automotive?

TimoPtr:
> Testing showed that while it is possible to manually start the service, this won't work when not set as the default assistant app in the background.

Indeed in the next PR I've disabled the toggle when not set as default in the settings.

TimoPtr:
> Nice to see how the local wake word work pays off, so much cleaner than when I worked on this back when Assist was new.
> 
> What's the idea behind the `microfrontend` name? 'Frontend' suggest the HA frontend user interface to me which is unrelated.
> 
> Should the service ignore input while AssistActivity is listening?
> 
> If you're going ahead with this setup, don't forget to update dev docs to indicate you need to install CMake for building (at least it didn't do so automatically for me on Windows - after that building worked fine!).

Doc updated https://github.com/home-assistant/developers.home-assistant/pull/2946

build-logic/convention/src/main/kotlin/AndroidApplicationDependenciesConventionPlugin.kt
TimoPtr: https://source.android.com/docs/automotive/voice/voice_interaction_guide/app_development It seems we might need another kind of service for automotive but it doesn't seems impossible.

TimoPtr: if I move this in app only then I need to move the code in a module that is not added to automotive. I would like to avoid this for this iteration.

jpelgrom: > Could you tell or create an issue about why we don't have assist offered as default in automotive?

The Automotive build is distributed to a couple of car manufacturers, and their reviews started asking for specific UI customizations and behaviors we had to implement to pass their policy.  At the same time Assist simply isn't optimized for hands-free usage (like a button to trigger answer/response, and always voice only responses), so with the limited resources we have we decided to remove it in #4965, [discussion directly leading up to it](https://discord.com/channels/330944238910963714/1213146231916724344/1326612235639849003).

I also think but am not 100% sure I remember correctly that assistant apps with a voice service on Automotive need a more general voice recognition service? Step 6 from your URL sounds like it.

(Personally I also don't see why you'd want HA as your only digital assistant app in a car.)

PR #6317, opened 2026-01-23
Fix: keep WebSocket subscriptions on resubscribe failure

AYUSHMIT:
## Summary
This change fixes an edge case during WebSocket reconnection where active subscriptions could be permanently dropped.

Previously, when a WebSocket reconnect succeeded but a subscription re-register request failed, the old subscription entry was removed immediately. This meant transient failures could cause the app to stop receiving events until a full restart.

This update keeps existing subscription state until a re-subscribe succeeds, allowing subsequent reconnect attempts to retry safely. A deterministic unit test was added to cover this failure path.

## Checklist

## Screenshots
N/A – no user-facing UI changes.

User Documentation: N/A  
Developer Documentation: N/A

## Any other notes
This change is intentionally minimal and scoped to reconnection behavior only. It does not alter subscription semantics or message handling outside of the reconnect flow.

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
@@ -1025,27 +1025,36 @@ internal class WebSocketCoreImpl(
     private suspend fun reconnectSubscriptions() {
         if (connect()) {
             Timber.d("Resubscribing to active subscriptions...")
-            activeMessages.filterValues { it is ActiveMessage.Subscription }.entries
-                .forEach { (oldId, oldActiveMessage) ->
-                    oldActiveMessage as ActiveMessage.Subscription
-                    activeMessages.remove(oldId)
-
-                    val response = sendMessage(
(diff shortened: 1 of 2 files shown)


AYUSHMIT:
CLA signed. Marked ready for review.
This PR is scoped to reconnection behavior only + includes a deterministic unit test for the failure path.

TimoPtr (commented):
@jpelgrom I'm not super confident about this. I get that it might help in some cases but that would require heavy testing and today we already have OOM that we don't know yet where it's actually coming from.

I guess this could help for long lasting subscription, not widget since after 30s it's killed.

jpelgrom:
> @jpelgrom I'm not super confident about this. I get that it might help in some cases but that would require heavy testing and today we already have OOM that we don't know yet where it's actually coming from.
> 
> I guess this could help for long lasting subscription, not widget since after 30s it's killed.

Agreed. It could be helpful but also backfire if unsubscribing from the app-end doesn't happen correctly. 

For transient failures the 10 second delay was added, and UI will (should) close/resubscribe when hidden/reopened. Maybe retrying once or twice after that is a good middle ground? @AYUSHMIT is there a specific situation in which you observed this as an issue?

Issue #6310, opened 2026-01-22
Impossible to resize widgets on Samsung Flip sub screen


TimoPtr:
Following this thread https://github.com/home-assistant/android/pull/6127#issuecomment-3783420847 we are not able to resize the widgets on the subscreen. Samsung is able to do it on their app, after investigation within their APK we did found some leads but someone with a Samsung phone should try to make it work.

Elbow15:
Hi,
I made this repo [https://github.com/Elbow15/res-xml](https://github.com/Elbow15/res-xml) which contains res/xml of samsung apps providing widgets of various sizes on the Z Flip's cover screen.

I hope this helps.

TimoPtr:
> Hi, I made this repo https://github.com/Elbow15/res-xml which contains res/xml of samsung apps providing widgets of various sizes on the Z Flip's cover screen.
> 
> I hope this helps.

If you have a Samsung device and some time could you try to make it work with the ButtonWidget?

PR #6127, opened 2025-12-04
feat: allow widgets on flip phones cover screen

romainrbr:
## Summary
This allows Home assistant widgets to be displayed on Flip Phones (at least Samsung Galaxy Flip X)
Following Android's documentation on widgets: https://developer.android.com/develop/ui/views/appwidgets?hl=en

> Declares whether your widget can be displayed on the home screen (home_screen), the lock screen (keyguard), or both. For Android 5.0 and higher, only home_screen is valid. 

## Checklist

- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [X] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [X] The changes have been thoroughly tested, and edge cases have been considered.
- [X] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

## Screenshots
- Widgets on cover screen: 
image
- "Standard" widgets still working: 
image

## Any other notes

Please let me know if anything is missing to get this PR merged, thanks for your work on this awesome project :heart:

PR #6321, opened 2026-01-24
Allow 2x2 button widget for samsung flip phones cover screen

JimDarmody:
## Summary

Fix issue #6310 about the size of button widget on samsung flip cover screen
it's an improvement of #6127 

## Checklist

## Screenshots

Screenshot

User Documentation: /

Developer Documentation: /

## Any other notes

I may have modified the code incorrectly to implement this feature, so please feel free to tell me what's wrong and make changes.
Thanks.

### DIFF
--- a/app/src/main/res/xml/samsung_button_widget_info_2x2.xml
+++ b/app/src/main/res/xml/samsung_button_widget_info_2x2.xml
@@ -2,4 +2,4 @@
 <samsung-appwidget-provider
     display="sub_screen"
     parentComponent="io.homeassistant.companion.android/io.homeassistant.companion.android.widgets.button.ButtonWidget"
-    privacyWidget="true" />
+    privacyWidget="false" />


jpelgrom:
Your suggestion is to create a new fixed size widget. This approach doesn't really scale to all our widgets and possible sizes.

Does this mean none of Samsung's widgets for the flip screen are actually resizeable?

JimDarmody:
> Your suggestion is to create a new fixed size widget. This approach doesn't really scale to all our widgets and possible sizes.
> 
> Does this mean none of Samsung's widgets for the flip screen are actually resizeable?

I agree that the fixed size widget approach isn't great.

On the cover screens of Samsung Z Flip phones, widgets cannot be resized. They are only available in sizes defined by the application. These sizes are limited to 2x2, 4x2, and 4x4.

I made this PR more to show an implementation that already works for the button widget, but I am aware that modifying certain parts of the code just to work on one category of phone is not ideal.

I am not an expert in android app, so feel free to move it to draft or close the PR.

TimoPtr:
From my findings these widgets are not resizable indeed. @JimDarmody thanks for making the effort of opening the PR. Unfortunately I won't be able to take it over since I don't have a Flip phone to try the modification.

The first step would be to address the lint issues reported by the CI.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/button/ButtonWidget.kt
```diff
@@ -49,11 +49,11 @@ import kotlinx.coroutines.launch
 import timber.log.Timber
 
 @AndroidEntryPoint
-class ButtonWidget : AppWidgetProvider() {
+sealed class ButtonWidget : AppWidgetProvider() {
     companion object {
         const val CALL_SERVICE =
             "io.homeassistant.companion.android.widgets.button.ButtonWidget.CALL_SERVICE"
-        private const val CALL_SERVICE_AUTH =
+        const val CALL_SERVICE_AUTH =
```
TimoPtr:
```suggestion
        private const val CALL_SERVICE_AUTH =
```

TimoPtr:
@JimDarmody Would you mind testing with my changes?

JimDarmody:
> @JimDarmody Would you mind testing with my changes?

Thanks for your quick response.

I got a build error on WidgetAuthenticationActivity.kt:38:47 Unresolved reference ‘CALL_SERVICE’ and a problem accessing the ButtonWidgetBase() class blocking Hilt_ButtonWidget.java.

I made a few changes, feel free to correct them.
The button widget still works correctly on the z flip cover screen.

TimoPtr:
> > @JimDarmody Would you mind testing with my changes?
> 
> Thanks for your quick response.
> 
> I got a build error on WidgetAuthenticationActivity.kt:38:47 Unresolved reference ‘CALL_SERVICE’ and a problem accessing the ButtonWidgetBase() class blocking Hilt_ButtonWidget.java.
> 
> I made a few changes, feel free to correct them.
> The button widget still works correctly on the z flip cover screen.

Alright, thanks for testing. I'll try to bring back the sealed class tomorrow, it makes the code 'safer'

app/src/main/res/xml/button_widget_info_2x2.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  2x2 variant for Samsung Z Flip cover screen
+-->
+<appwidget-provider xmlns:android="http://schemas.android.com/apk/res/android"
+    android:minWidth="150dp"
+    android:minHeight="133dp"
```
jpelgrom, line 7: How did you come up with these numbers? Does it still work correctly if you change the display scaling in system settings?

app/src/main/res/xml/samsung_button_widget_info_2x2.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<samsung-appwidget-provider
+    display="sub_screen"
+    parentComponent="io.homeassistant.companion.android/io.homeassistant.companion.android.widgets.button.ButtonWidget"
```
jpelgrom, line 4: What does setting this value do?

app/src/main/res/xml/samsung_button_widget_info_2x2.xml
```diff
@@ -0,0 +1,5 @@
...
+    privacyWidget="true" />
```
jpelgrom: What does setting this value do?

app/src/main/AndroidManifest.xml
```diff
@@ -185,6 +185,18 @@
...
 
+        <!-- 2x2 variant for Samsung Z Flip cover screen -->
+        <receiver android:name=".widgets.button.ButtonWidget2x2" android:label="@string/widget_button_image_description"
+            android:exported="false">
+            <intent-filter>
+                <action android:name="android.appwidget.action.APPWIDGET_UPDATE" />
+            </intent-filter>
+            <meta-data android:name="android.appwidget.provider" android:resource="@xml/button_widget_info_2x2" />
+            <meta-data android:name="com.samsung.android.appwidget.provider" android:resource="@xml/samsung_button_widget_info_2x2" />
+            <meta-data android:name="com.samsung.android.appwidget.PRIVACY_WIDGET" android:value="false" />
```
jpelgrom, line 196: Why is this different from the value for privacy widget in the xml file?

app/src/main/res/xml/button_widget_info_2x2.xml
TimoPtr: All these values are coming from the Samsung app. While I was testing and failed to make it work I think the key here is in the details about the targetCell and the min sizes.

TimoPtr:
> Did you check whether widget IDs are unique if you add both 'normal' and 2x2 widgets?
> 
> @TimoPtr You specifically requested my review. I'm still hesitant to approve this because of the issue I mentioned with doing this for all widgets and sizes. It'll also create a lot of clutter for anyone not using a Samsung Z Flip with the additional widget(s, in the future).

I see two path
- We use the android:widgetCategory="keyguard" only for the 2x2 so it means that users should only see the widget while settings up widgets for lockscreen (it might be acceptable). For instance I don't see the 2x2 variant on my pixel
- We try to disable the 2x2 widget components and enable then from the `HomeAssistantApplication` when it starts when we are on Samsung devices (I don't know if we can know if it has a cover screen)

app/src/main/kotlin/io/homeassistant/companion/android/widgets/button/ButtonWidget.kt
```diff
@@ -81,7 +97,8 @@ class ButtonWidget : AppWidgetProvider() {
     private fun updateAllWidgets(context: Context) {
         mainScope.launch {
             val appWidgetManager = AppWidgetManager.getInstance(context) ?: return@launch
-            val systemWidgetIds = appWidgetManager.getAppWidgetIds(ComponentName(context, ButtonWidget::class.java))
+            val systemWidgetIds = listOf(ButtonWidget::class.java, ButtonWidget2x2::class.java)
+                .flatMap { appWidgetManager.getAppWidgetIds(ComponentName(context, it)).toList() }
```
TimoPtr, line 101: I really don't like having to list the subClasses ... In future refactor toward glance and the architecture we use in TodoWidget this might not be needed anymore.

app/src/main/res/xml/samsung_button_widget_info_2x2.xml
JimDarmody: It seems that this allows the samsung cover screen system to know which class inherits the functionality.

app/src/main/res/xml/samsung_button_widget_info_2x2.xml
JimDarmody: This allows you to hide the widget when the cover screen is locked. Here, there is an error related to the reuse of code from a samsung app. It should be false, as in the manifest.

PR #6336, opened 2026-01-26
Sync Thread network preference on Matter commissioning from frontend, if necessary

jpelgrom:
## Summary

This PR tries to improve the user experience for commissioning of Matter over Thread devices, by actively syncing a dataset if necessary.

### Background

In #4150, the full Thread credential sync was removed from the start of Matter commissioning because of the slow GMS API. However, for users with a Thread network only known to HA (like those who added a ZBT-2, or Apple network synced from an iOS device!), this means Matter over Thread device commissioning would fail unless a full sync was manually triggered from companion app settings.

After the removal, the frontend [started sending information](https://github.com/home-assistant/frontend/pull/22022) about the preferred Thread network, if any, in the commissioning payload. Instead of doing a full sync every time, this allows us to sync the device only if necessary by checking if the device knows these credentials.

_Note that the GMS API unfortunately hasn't seen any improvement and we still cannot specify which Thread network to use, so the check/sync remains best effort, instead of being sure this Thread network is used for commissioning like the frontend's intention._

### Code in this PR

If the frontend commissioning payload includes Thread credentials, the app checks if the device prefers the network for these credentials (we can do this without asking for permissions or showing additional UI).
 - If there are no credentials in the payload, or the device prefers these credentials: the Matter commissioning flow starts.
 - If there are credentials: if the app has never synced we do a full sync once, once synced the Matter commissioning flow starts.

In a flowchart this looks like:

```mermaid
flowchart LR
    A(Frontend starts Matter commissioning) --> B{Payload with Thread credentials?}
    B -- No --> C(Start Matter commissioning flow)
    B -- Yes --> D{Thread network from credentials preferred by device?}
    D -- Yes --> C
    D -- No --> E{Thread synced at least once?}
    E -- No --> F[Perform full sync]
    F --> C
    E -- Yes --> C
```

While testing, asking the GMS API whether the dataset is preferred still takes almost exactly 10 seconds most of the time, but not always - I suspect it is waiting for a timeout somewhere. That means this change still slows down starting commissioning, but only if core knows about Thread and less than a full sync.

Logcat showing 10 seconds between messages for frontend commissioning start and Thread network preferred check complete

## Checklist

n/a

## Any other notes

[Discussion](https://discord.com/channels/330944238910963714/1346948551892009101/1444757127322271917) on Discord that started this PR.

This PR is still draft while I want to
 - test a bit more with removing credentials and when to set/unset the has synced boolean
 - polish exception handling without changing all the code

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
@@ -516,7 +516,7 @@ class WebViewPresenterImpl @Inject constructor(
             mainScope.launch {
                 if (tlv != null && borderAgentId != null) {
                     Timber.d("Matter commissioning payload has preferred Thread dataset, comparing")
-                    compareThreadCredentialsForCommissioning(context, tlv)
+                    compareThreadDatasetForCommissioning(context, tlv)
                 } else {
                     Timber.d("Matter commissioning payload doesn't have Thread, starting commissioning")
                     startMatterCommissioningFlow(context)
@@ -526,15 +526,15 @@ class WebViewPresenterImpl @Inject constructor(
     }
 
     /**
-     * Compare Thread credentials supplied for Matter commissioning, and take action depending on the device state:
-     * - Device prefers credentials: start (external) Matter commissioning flow
-     * - Device doesn't prefer credentials:
-     *    - Has never synced Thread for the server: do a full Thread credentials sync (this will usually result in the
-     *    credentials becoming preferred)
+     * Compare Thread dataset supplied for Matter commissioning, and take action depending on the device state:
+     * - Device prefers dataset: start (external) Matter commissioning flow
+     * - Device doesn't prefer dataset:
+     *    - Has never synced Thread for the server: do a full Thread dataset sync (this will usually result in the
+     *    dataset becoming preferred, if the device has no other Thread dataset)
      *    - Has previously synced Thread for the server: start (external) Matter commissioning flow (the app is
      *    expected to be unable to change the preferred state by syncing)
      */
-    private suspend fun compareThreadCredentialsForCommissioning(context: Context, tlv: String) {
+    private suspend fun compareThreadDatasetForCommissioning(context: Context, tlv: String) {
         val isPreferred = threadUseCase.isPreferredDatasetByDevice(context, tlv)
 
         if (isPreferred) {
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
@@ -509,15 +509,54 @@ class WebViewPresenterImpl @Inject constructor(
 
     override fun appCanCommissionMatterDevice(): Boolean = matterUseCase.appSupportsCommissioning()
 
-    override fun startCommissioningMatterDevice(context: Context) {
+    override fun startCommissioningMatterDevice(context: Context, tlv: String?, borderAgentId: String?) {
         if (mutableMatterThreadStep.value != MatterThreadStep.REQUESTED) {
             mutableMatterThreadStep.tryEmit(MatterThreadStep.REQUESTED)
 
-            // The app used to sync Thread credentials here until commit 26a472a, but it was
-            // (temporarily?) removed due to slowing down the Matter commissioning flow for the user
-            // and limited usefulness of the result (because of API limitations)
+            mainScope.launch {
+                if (tlv != null && borderAgentId != null) {
+                    Timber.d("Matter commissioning payload has preferred Thread dataset, comparing")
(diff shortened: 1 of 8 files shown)


bgoncal:
> Note that the GMS API unfortunately hasn't seen any improvement and we still cannot specify which Thread network to use, so the check/sync remains best effort, instead of being sure this Thread network is used for commissioning like the frontend's intention.

So this means that when comissioning a new matter over thread device, the preferred thread network selected in Home Assistant will not be respected right?

What would happen in my case where I have one HA thread network, one Aqara and one Apple, which one it would decide to use? (Considering all of them have their credentials saved on my Google Pixel/Google services)

jpelgrom:
> So this means that when comissioning a new matter over thread device, the preferred thread network selected in Home Assistant will not be respected right?

It may or may not be, depending on which network was added first.

> What would happen in my case where I have one HA thread network, one Aqara and one Apple, which one it would decide to use? (Considering all of them have their credentials saved on my Google Pixel/Google services)

Right now, it seems that whatever app adds the first credentials will become the preferred. Apps cannot influence this, and the user also cannot change this.

All of this isn't new behavior. It already applies whenever you commission a Matter device today and manually sync Thread from the companion app settings.

bgoncal:
@jpelgrom I understand, thats just frustrating that they dont open up the APIs to at least be able to define to which thread network to add the device to. Hopefully Matter improvements will bring the so called "single network" where networks will merge into one

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -498,15 +498,54 @@ class WebViewPresenterImpl @Inject constructor(
...
+    override fun startCommissioningMatterDevice(context: Context, tlv: String?, borderAgentId: String?) {
         if (mutableMatterThreadStep.value != MatterThreadStep.REQUESTED) {
             mutableMatterThreadStep.tryEmit(MatterThreadStep.REQUESTED)
 
-            // The app used to sync Thread credentials here until commit 26a472a, but it was
-            // (temporarily?) removed due to slowing down the Matter commissioning flow for the user
-            // and limited usefulness of the result (because of API limitations)
-
-            startMatterCommissioningFlow(context)
+            mainScope.launch {
```
TimoPtr, line 516: Could you split a bit this function? It has too many layer and I think having smaller functions with proper names would make it easier to follow what's going on.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -498,15 +498,54 @@ class WebViewPresenterImpl @Inject constructor(
...
+                        startMatterCommissioningFlow(context)
+                    } else if (!threadUseCase.hasSyncedPreferredDataset(serverId)) {
+                        // We only sync if the server has never synced, as a full sync can be slow
+                        Timber.d("Device doesn't prefer core preferred dataset, starting first sync")
+                        val deviceThreadIntent = try {
+                            val result = threadUseCase.syncPreferredDataset(
+                                context = context,
+                                serverId = serverId,
+                                exportOnly = false,
+                                scope = CoroutineScope(mainScope.coroutineContext + SupervisorJob()),
```
TimoPtr: Any reason to create a new scope?

jpelgrom: Short answer: because I reused the previously removed code.

Longer answer: because I want a `SupervisorJob` to ensure any failures here don't impact the rest of the rest of `WebViewPresenterImpl`- `mainScope` uses a normal `Job`. See: #3634. We could also change `mainScope` to use a `SupervisorJob` if that's a better solution in your opinion.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
jpelgrom: I've split it into a couple of smaller but still logically grouped functions, with a description for each function. Hopefully this is a good middle ground between extremely tiny and the original.

jpelgrom:
After testing various scenarios, it seems to work mostly as expected but last minute I thought of the need to handle the scenario:
 - core previously preferred dataset `x`, the app synced and the device also prefers it
 - core now prefers dataset `y` instead of `x`, the device should be updated to match it

This can be included in the 'fast' check by triggering sync if
 - there has never been a sync (✅)
 - there has been a sync (✅) which resulted in adding a dataset for `x`, which is different from `y` which core is now preferring (TODO)

So this PR will stay in draft for a little longer, while I work on that change (most likely Wednesday) as well as a test for each yes/no of the flowchart.

Improving the quality of the Matter/Thread components itself, like more specific error handling, is a larger task that I want to keep outside the scope of this PR. The specific error class is only because I'm making something new, not changing existing code with it for now.

jpelgrom:
Closing in favor of building on #6918 / #6951 once merged, it already has a minor change to the sync

PR #6338, opened 2026-01-27
Fix potential concurrent modification exception in MainViewModel

TimoPtr:
## Summary

It's not optimal, but to work with the actual code without breaking everything I decided to use the `Snapshot.takeSnapshot` way to make sure we are using a freeze version of all the MutableState while making the UIState. The reason why the concurrent modification happens is that we are making operation on mutableState like `entitiesByArea` while building the UIState, but at the same time the ViewModel can modify this same `entitiesByArea` leading to the issue. We could have use a Mutex but it would have make things a bit more complex with the current code.

A better solution should be made later by refactoring the ViewModel and the view.


jpelgrom:
Agreed this needs  a bigger refactor, but taking steps towards a fix and preventing the issue seems like a good idea for now.

In scenarios where there would be concurrent modification, I think it'll call the function to update the entire UI state later so no state is actually lost/not updated.

PR #6341, opened 2026-01-28
Add LicensesFragment using AboutLibraries

TimoPtr:
## Summary

This PR adds a new entry in the settings screen to display all the licenses of the libraries we use. It uses the well known library [AboutLibraries](https://github.com/mikepenz/AboutLibraries).

## Checklist

## Screenshots

image
image

image
image
image

### DIFF
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -519,8 +519,8 @@
     <string name="ok">OK</string>
     <string name="open_with_headphone_button">Open with headset button</string>
     <string name="open_with_headphone_button_summary">Use Assist with the assistant/voice command button on supported headphones and earbuds</string>
-    <string name="open_source_licenses">Open source licenses</string>
-    <string name="open_source_licenses_summary">View licenses for third-party libraries</string>
+    <string name="licenses">Licenses</string>
+    <string name="licenses_summary">View licenses for third-party libraries</string>
     <string name="other_settings">Other settings</string>
     <string name="other">Other</string>
     <string name="password">Password</string>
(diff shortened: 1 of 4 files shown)


jpelgrom:
I'm positively surprised how much it recognized. I think it got all libraries and is just missing the license for com.github.Dimezis:BlurView.

UI issues:
 - 1, 2 does not like having a dark theme enabled
 - 3 it needs a modifier for edge to edge on the list, I've scrolled all the way to the bottom here and the bottom item is obscured by the navigation bar:

Edit: pushed a quick fix for 3

|1|2|3|
|---|---|---|
|image|image|image|

TimoPtr:
@jpelgrom thanks for fixing the inset issue. I've fixed the color issue.

common/src/main/res/values/strings.xml
```diff
@@ -519,6 +519,8 @@
     <string name="ok">OK</string>
     <string name="open_with_headphone_button">Open with headset button</string>
     <string name="open_with_headphone_button_summary">Use Assist with the assistant/voice command button on supported headphones and earbuds</string>
+    <string name="open_source_licenses">Open source licenses</string>
```
jpelgrom: This should be renamed to simply 'Licenses', 'Third-party licenses' or 'Acknowledgements' (I've noticed Pocket Casts uses that name with the same library). The full version also includes closed source dependencies with a license.

TimoPtr: You are right in that case I'm going to rename also the screen to Licenses not OSS

TimoPtr:
> There's a white flash when opening this in dark mode, maybe we can prevent that?
> 
> Trying to scroll left/right reveals a 1px border while holding my finger on the screen. Maybe this is an overflow effect that isn't working, but it shouldn't be possible to scroll left/right.
> 
> Functionally it otherwise looks good now.

Good catch on the flash, it is because I've used Scaffold from Material instead of Material3 ...

common/src/main/res/values/strings.xml
jpelgrom: Yes I hoped that would be obvious, but adding that as a suggestion would require a lot of comments while it's easy to change everything quickly in the IDE :)

jpelgrom (approved):
Tested and swapping the scaffold also fixed the left/right issue.

PR #6342, opened 2026-01-28
Add assist settings for wake word

TimoPtr:
## Summary

Introduce a dedicated screen to set settings for Assist for now it's limited to few settings, but we could extend to others. This PR also contains the a new repository that contains the settings so it can be read by the service.
The service is now starting the wake word detection when it starts if it has been configured like this.

## Checklist

## Screenshots

image

The UI will most probably change while iterating on the feature.

User Documentation: https://github.com/home-assistant/home-assistant.io/pull/43488

## Any other notes

I've mutualised some logic about the checks with the system to know if Assist is the default assistant.

By using the VoiceInteractionService, the detection should stay alive all the time and restart when the device restarts, but this needs more testing.

This PR is based on #6312

### DIFF
--- a/app/src/test/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionSessionServiceTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionSessionServiceTest.kt
@@ -1,5 +1,6 @@
 package io.homeassistant.companion.android.assist.service
 
+import dagger.hilt.android.testing.HiltTestApplication
 import io.homeassistant.companion.android.testing.unit.ConsoleLogRule
 import org.junit.Assert.assertTrue
 import org.junit.Rule
@@ -7,8 +8,10 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.Robolectric
 import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
 
 @RunWith(RobolectricTestRunner::class)
+@Config(application = HiltTestApplication::class)
 class AssistVoiceInteractionSessionServiceTest {
 
     @get:Rule
--- a/app/src/test/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionServiceTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionServiceTest.kt
@@ -3,6 +3,7 @@ package io.homeassistant.companion.android.assist.service
 import android.Manifest
 import android.content.Intent
 import androidx.test.core.app.ApplicationProvider
+import dagger.hilt.android.testing.HiltTestApplication
 import io.homeassistant.companion.android.assist.wakeword.MicroWakeWordModelConfig
 import io.homeassistant.companion.android.assist.wakeword.WakeWordListener
 import io.homeassistant.companion.android.assist.wakeword.WakeWordListenerFactory
@@ -31,6 +32,7 @@ import org.junit.Test
 import org.junit.jupiter.api.assertNull
 import org.junit.runner.RunWith
 import org.robolectric.Robolectric
+import org.robolectric.annotation.Config
 import org.robolectric.RobolectricTestRunner
 import org.robolectric.Shadows
 import org.robolectric.android.controller.ServiceController
@@ -38,6 +40,7 @@ import org.robolectric.shadows.ShadowVoiceInteractionService
 
 @OptIn(ExperimentalCoroutinesApi::class)
 @RunWith(RobolectricTestRunner::class)
+@Config(application = HiltTestApplication::class)
 class AssistVoiceInteractionServiceTest {
 
     @get:Rule(order = 0)
--- a/app/src/test/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionSessionTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionSessionTest.kt
@@ -1,5 +1,6 @@
 package io.homeassistant.companion.android.assist.service
 
+import dagger.hilt.android.testing.HiltTestApplication
 import android.content.Intent
 import android.os.Bundle
 import androidx.test.core.app.ApplicationProvider
@@ -12,10 +13,12 @@ import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
 import org.robolectric.Shadows
 import org.robolectric.shadows.ShadowVoiceInteractionSession
 
 @RunWith(RobolectricTestRunner::class)
+@Config(application = HiltTestApplication::class)
 class AssistVoiceInteractionSessionTest {
 
     @get:Rule


TimoPtr:
Known issues
- Assist is starting in another process so if you are in the app and open Assist from the frontend you will have two Assist open
- Currently not integrated with potential other Assist close to you, if you have a VPE close to you what do we do? At the moment it triggers both
- It seems that the VoiceInteractionService is not always starting right after settings the app as default Assistant and the best way to address this seems to reboot the device.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsFragment.kt
```diff
@@ -0,0 +1,40 @@
...
+            setContent {
+                HATheme {
+                    AssistSettingsScreen(viewModel = viewModel)
+                }
+            }
+        }
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        // TODO new docs
```
jpelgrom: This was integrated in the main HA docs as we found that most users also needed info on how to set up Assist before being able to use it in the app https://www.home-assistant.io/voice_control/android/

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistRepository.kt
```diff
@@ -0,0 +1,88 @@
...
+import io.homeassistant.companion.android.assist.service.AssistVoiceInteractionService
+import io.homeassistant.companion.android.assist.wakeword.MicroWakeWordModelConfig
+import io.homeassistant.companion.android.common.data.prefs.PrefsRepository
+import io.homeassistant.companion.android.common.util.SuspendLazy
+import javax.inject.Inject
+
+/**
+ * Repository for Assist settings and wake word model information.
+ */
+interface AssistRepository {
```
jpelgrom: I'm not the biggest fan of using Repository outside of common/data, now we get two different kinds of repositories (ones that actually hold data and ones that are just a interface on top of other repositories). It's also not in the architecture we follow unless I'm mistaken?

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreenScreenshotTest.kt
```diff
@@ -0,0 +1,112 @@
...
+                onStartTestWakeWord = {},
+                onStopTestWakeWord = {},
+            )
+        }
+    }
+
+    @PreviewTest
+    @HAPreviews
+    @Composable
+    fun `Assist settings testing wake word with detection`() {
```
jpelgrom, line 91: \+ add a variant with wake word not yet detected

app/src/test/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionSessionServiceTest.kt
jpelgrom, line 1: I don't think this needs to be in this PR

app/src/test/kotlin/io/homeassistant/companion/android/settings/assist/AssistRepositoryTest.kt
```diff
@@ -0,0 +1,216 @@
...
+            author = "test",
+            website = "https://test.com",
+            model = "hey_jarvis.tflite",
+            trainedLanguages = listOf("en"),
+            version = 1,
+            micro = MicroWakeWordModelConfig.MicroFrontendConfig(
+                probabilityCutoff = 0.5f,
+                featureStepSize = 10,
+                slidingWindowSize = 20,
+            ),
```
jpelgrom: I've seen this repeated multiple times - time to put it in a test utility file?

app/src/test/kotlin/io/homeassistant/companion/android/settings/assist/AssistRepositoryTest.kt
```diff
@@ -0,0 +1,216 @@
...
+
+            assertFalse(result)
+        }
+    }
+
+    @Nested
+    inner class SetWakeWordEnabledTest {
+
+        @Test
+        fun `Given enabled true when setWakeWordEnabled then save preference and start service`() = runTest {
```
jpelgrom: This doesn't stop the service, just the listening. Stopping the service isn't really possible without changing the default assistant app.

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
```diff
@@ -167,6 +164,14 @@ class SettingsFragment(private val presenter: SettingsPresenter, private val lan
             }
         }
 
+        findPreference<Preference>("assist_settings")?.setOnPreferenceClickListener {
```
jpelgrom, line 168: This should be hidden on Automotive as default assistant app and wake word isn't available there (or when the headset toggle is moved to this screen, the screen should hide the related settings).

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -0,0 +1,453 @@
...
+            color = colorScheme.colorTextSecondary,
+        )
+        Text(
+            text = selectedModel?.wakeWord ?: "",
+            style = HATextStyle.Body,
+            color = colorScheme.colorTextPrimary,
+            modifier = Modifier.padding(top = HADimens.SPACE1),
+        )
+
+        DropdownMenu(
```
jpelgrom: Personally without looking at the code and as a non-standard component (M3 system settings have title/value styles swapped), this wasn't obviously as a dropdown to me and might as well have been a reminder for "this is _the_ wake word" like the default assistant app.

The dropdown menu doesn't follow the theme correctly. It's light in dark mode, and maybe even looks like default M3 purple-ish (more obvious in light mode)?

Screenshot of Assist settings with the

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -0,0 +1,453 @@
...
+internal fun AssistSettingsScreen(
+    uiState: AssistSettingsUiState,
+    hasAudioPermission: Boolean,
+    onSetDefaultAssistant: () -> Unit,
+    onToggleWakeWord: (Boolean) -> Unit,
+    onSelectWakeWord: (MicroWakeWordModelConfig) -> Unit,
+    onStartTestWakeWord: () -> Unit,
+    onStopTestWakeWord: () -> Unit,
+) {
+    Column(
```
jpelgrom, line 178: Needs edge to edge support. Using split view + buttons to make it more obvious with the bottom setting item partially being obscured:

image

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistRepository.kt
TimoPtr: I'm open to suggestion here for naming :). We could move this in common/ since it might benefits the watch?

TimoPtr:
> Would it make sense to allow this to be a per-server setting? Trigger Assist on a specific server with the wake word, or even different wake words for different servers.

I was more thinking about a global settings at first that allow you to pick the default server and default pipeline on this server.

But indeed later we could have multiple wake word and assign each wake word to a server/pipeline. But it would require a bit of work on the wrapper side to share some logic to not do the feature extraction multiple times.

I'm for making a first version without any other settings that the current one. It means that it simply keep the last settings and then later we evolve it. wdyt?


> There are a lot of modifiers to make the settings screen rows. I think this may be a good situation for defaults that combine these into a semantic 'HASettingRow' or something like that (in a different PR!).

Agreed but for now the UI is not done I would say, I'm trying to get the Product team at OHF to help on this screen to fit well in the rest of HA.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsFragment.kt
TimoPtr: Thanks for sharing the link, we would need to update the content with a beta flag right?

app/src/test/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionSessionServiceTest.kt
TimoPtr: Indeed but I did not added in the first PR because I was not sure about the changes I was going to make to the service itself. If you don't mind I would like to keep it here so I don't have to move it and rebase again.

app/src/test/kotlin/io/homeassistant/companion/android/settings/assist/AssistRepositoryTest.kt
TimoPtr: Indeed, I kept it duplicated only for the screenshot tests

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
TimoPtr: I think the whole category is hidden https://github.com/home-assistant/android/blob/991478d8110649958f77efcc921027f86338b734/app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt#L191 I didn't test on Automotive TBH

TimoPtr: So even headset toggle might be hidden in automotive no?

TimoPtr: Yes I've checked and it the whole category (included headset toggle) is not visible.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
TimoPtr: Shame on me for being lazy and telling myself that we will do it later once we have more content ^^'.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
TimoPtr: To be honest I used Claude for UI inspiration, I didn't know how to make the screen. But you are right about that it is not clear that we can click on it I would like some UI helps to get the screen align with the rest.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsFragment.kt
jpelgrom: The main docs don't know that pattern because they actually have different sites and update based on core releases. Not sure what the best solution here would be so pinging @c0ffeeca7 as the main docs writer for opinions on how we should document wake word support in the beta app.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistRepository.kt
jpelgrom: I don't really see a future where we get to do this on the watch. Even enabling the Google Assistant (or Gemini now, I guess) wake word destroys a watch battery life and that one is supposed to be optimized.

I'll think about a better name, we kind of have classes like these in the *Managers, but those are actually used in different places. This is mostly a subset of what you'd put in the ViewModel.

app/src/test/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionSessionServiceTest.kt
jpelgrom: 🫤 OK as it is very small but try to be mindful of it, having another small PR to add a test can be done and reviewed quickly

app/src/main/kotlin/io/homeassistant/companion/android/settings/SettingsFragment.kt
jpelgrom: Wrong assumption about it from me, also didn't actually check it. No changes needed then. If we ever want to introduce more Assist settings on Automotive unhide the category but make sure to hide settings for the wake word.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
jpelgrom: Never assume it fits, people put our app on tiny square screens in their walls or have scaling set to 250%.

jpelgrom:
> I was more thinking about a global settings at first that allow you to pick the default server and default pipeline on this server.
> 
> But indeed later we could have multiple wake word and assign each wake word to a server/pipeline. But it would require a bit of work on the wrapper side to share some logic to not do the feature extraction multiple times.

After reading your reply and looking at options, maybe we should try to offer similar options to a 'normal' voice satellite. You already have a comment about x. They offer support for [up to 2 wake words to trigger different Assist pipelines](https://www.home-assistant.io/blog/2025/10/22/voice-chapter-11/#multilingual-assistants), presumably for the same server but we could change it to allow each wakeword to trigger Assist on any server for a specific pipeline.

> I'm for making a first version without any other settings that the current one. It means that it simply keep the last settings and then later we evolve it. wdyt?

Agreed for now, let's start simple and evolve as there is already a lot going on. As long as the wake word acts like triggering Assist with the gesture/button we can ignore any server/pipeline settings and deal with it later.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -0,0 +1,462 @@
...
+import io.homeassistant.companion.android.common.compose.theme.HADimens
+import io.homeassistant.companion.android.common.compose.theme.HARadius
+import io.homeassistant.companion.android.common.compose.theme.HATextStyle
+import io.homeassistant.companion.android.common.compose.theme.HAThemeForPreview
+import io.homeassistant.companion.android.common.compose.theme.LocalHAColorScheme
+import io.homeassistant.companion.android.util.plus
+import io.homeassistant.companion.android.util.safeBottomPaddingValues
+import org.jetbrains.annotations.VisibleForTesting
+
+/* List of settings
```
TimoPtr: @jpelgrom I'm surprised you didn't comment on this.

I made this list based on the potential settings we could add to the screen this should probably be moved to an issue but I wanted to discuss it with you.

jpelgrom: I never claimed to have reviewed all ;) I touched on some when mentioning how we should handle multiserver.

Probably best to move it to an issue so it can be discussed and tackled separately. Some of these are more involved and would require consideration as to how they might impact launching Assist from the frontend and/or would also require core changes (e.g. Assist Satellite).

laupalombi:
My only blocker in terms of Product are:

- Naming - its not Assist, we already have an Assist section, needs to be like Add Assist to Android or Assist for Android
- Options available need to match the same pattern as we do in VPE - as it is right now in the screenshots its OK - pick assist pipeline, pick wake word, test it.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistRepository.kt
TimoPtr: We decided to go the AssistConfigManager way over a private discussion Discord.

jpelgrom:
Do we want to label this 'beta' or 'preview' to be safe, as there are probably going to be a couple of smaller follow up PRs to get this where we want?

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
jpelgrom: See #6416

TimoPtr:
> Do we want to label this 'beta' or 'preview' to be safe, as there are probably going to be a couple of smaller follow up PRs to get this where we want?

If you mean within the app yes I think it would be a good idea to add a label within the wake word section like 'experimental' maybe ? If it land on prod to not mix it with beta.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistConfigManager.kt
```diff
@@ -0,0 +1,88 @@
...
+import androidx.annotation.RequiresPermission
+import dagger.hilt.android.qualifiers.ApplicationContext
+import io.homeassistant.companion.android.assist.service.AssistVoiceInteractionService
+import io.homeassistant.companion.android.assist.wakeword.MicroWakeWordModelConfig
+import io.homeassistant.companion.android.common.data.prefs.PrefsRepository
+import io.homeassistant.companion.android.common.util.SuspendLazy
+import javax.inject.Inject
+
+/**
+ * Repository for Assist settings and wake word model information.
```
jpelgrom:
```suggestion
 * Manager for Assist settings and wake word model information.
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -0,0 +1,446 @@
...
+            .fillMaxWidth()
+            .clip(RoundedCornerShape(HARadius.XL))
+            .background(colorScheme.colorSurfaceLow)
+            .clickable { expanded = true }
+            .padding(HADimens.SPACE4),
+    ) {
+        Text(
+            text = stringResource(commonR.string.assist_wake_word_model),
+            style = HATextStyle.BodyMedium,
+            color = colorScheme.colorTextSecondary,
```
jpelgrom: Swapping title/value styles as [previously mentioned](https://github.com/home-assistant/android/pull/6342#discussion_r2766140559) better matches M3 Expressive system settings and existing screens, which I'll use as a reference. Without this change it is less obvious in my opinion this is a setting that can be tapped.

```suggestion
            style = HATextStyle.Body,
            color = colorScheme.colorTextPrimary,
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -0,0 +1,446 @@
...
+        )
+        Text(
+            text = selectedModel?.wakeWord ?: "",
+            style = HATextStyle.Body,
+            color = colorScheme.colorTextPrimary,
```
jpelgrom:
```suggestion
            style = HATextStyle.BodyMedium,
            color = colorScheme.colorTextSecondary,
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -0,0 +1,446 @@
...
+
+@OptIn(ExperimentalPermissionsApi::class)
+@Composable
+fun AssistSettingsScreen(viewModel: AssistSettingsViewModel) {
+    val uiState by viewModel.uiState.collectAsStateWithLifecycle()
+
+    val togglePermissionState = rememberPermissionState(Manifest.permission.RECORD_AUDIO) { granted ->
+        if (granted) {
+            viewModel.onToggleWakeWord(true)
+        }
```
jpelgrom, line 105: If permission is (auto)denied, a Snackbar should be shown to indicate that microphone permission is required + an action to grant it anyway that links to the system settings screen.

Right now, the experience is bad when the permission is denied once because you can tap the toggle but nothing happens.

common/src/main/res/values/strings.xml
```diff
@@ -1200,6 +1201,19 @@
...
+    <string name="assist_set_default">Set as default</string>
+    <string name="assist_summary">Configure Assist for this device</string>
+    <string name="assist_wake_word_title">Wake word</string>
+    <string name="assist_wake_word_enable">Enable wake word detection</string>
+    <string name="assist_wake_word_model">Wake word</string>
+    <string name="assist_wake_word_test">Test wake word</string>
+    <string name="assist_wake_word_stop_test">Stop testing</string>
+    <string name="assist_wake_word_test_hint">Say \"%s\" to test</string>
+    <string name="assist_wake_word_detected">Wake word detected!</string>
+    <string name="assist_wake_word_battery_warning">Wake word detection runs entirely on your device without any server. It continuously listens for audio, which may increase battery usage.</string>
```
jpelgrom: Trying to make it more obvious that
 - we don't need any connection (so it also isn't sent to a third party)
 - it only listens for your wake word (matches what the notification title says)
 - increased battery usage is basically guaranteed

```suggestion
    <string name="assist_wake_word_battery_warning">Wake word detection runs entirely on your device, no network connection required. It continuously listens for your wake word, which will increase battery usage.</string>
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -0,0 +1,446 @@
...
+                    Modifier
+                },
+            )
+            .padding(HADimens.SPACE4),
+        horizontalArrangement = Arrangement.SpaceBetween,
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        Text(
+            text = stringResource(commonR.string.assist_wake_word_enable),
+            style = HATextStyle.Body,
```
jpelgrom: This doesn't overflow nicely on smaller screens/with larger fonts

Settings row with Enable wake word detectio - clipped by toggle

```suggestion
            style = HATextStyle.Body,
            modifier = Modifier.weight(1f),
            textAlign = TextAlign.Start,
```

(you need to add a new import for this)

common/src/main/res/values/strings.xml
```diff
@@ -1200,6 +1201,19 @@
     <string name="assist_listening_wakeword">Say \"%s\" to activate Assist</string>
     <string name="assist_listening_channel">Assist wake word</string>
     <string name="assist_listening_channel_description">Shows when the app is listening for sounds to activate Assist</string>
+    <string name="assist_default_assistant_title">Default assistant</string>
+    <string name="assist_default_assistant_enabled">Home Assistant is your default assistant</string>
+    <string name="assist_default_assistant_disabled">Home Assistant is not your default assistant</string>
+    <string name="assist_set_default">Set as default</string>
+    <string name="assist_summary">Configure Assist for this device</string>
+    <string name="assist_wake_word_title">Wake word</string>
+    <string name="assist_wake_word_enable">Enable wake word detection</string>
```
jpelgrom, line 1212:
```suggestion
    <string name="assist_wake_word_enable">Enable wake word detection (experimental)</string>
```

TimoPtr: I was considering adding something that looks like the GitHub labels under the PR or issues. Having a different translation also would make removal of the experimental flag easier.

common/src/main/res/values/strings.xml
TimoPtr: Well without connection we should probably not open Assist actually.

common/src/main/res/values/strings.xml
jpelgrom: I was trying to avoid creating a new component but OK. Something like this?
Settings row for Enable wake word detection, with the word Experimental encircled below

jpelgrom: Or inline with the title (less busy in the toggle, but looks a bit too much like something you can tap for me)

Settings section titled Wake word, followed by the encircled word Experimental

TimoPtr: On the title yes. Probably using another color I think it's fine, labels are not clickable.

common/src/main/res/values/strings.xml
jpelgrom: > Well without connection we should probably not open Assist actually.

Following that logic, it also should probably not open Assist if there is no server?

TimoPtr:
> While testing I once ended up in a state where after setting the app as the default assistant app, the microphone listening didn't actually start even when I manually toggled the wake word detection on and off, but I haven't been able to easily replicate it.

It did happen to me too. I'm not sure we can do something about it it is quite tight to how the system starts the service. I'm not sure it is made so that it can change frequently.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsFragment.kt
c0ffeeca7: If you want to keep it together with the Assist on Android docs on ha.io: How about adding a Prerequisite section and listing the version required? Would that work? 

I realize that we don't usually document beta on the current branch for the main docs, but this is about Android, so I think we could handle it in line with what you do in the companion docs with the beta flag.

common/src/main/res/values/strings.xml
TimoPtr: I think we can think later about how to proactively disable the wake word detection to save battery, like when no server and no connection. Adding it to the issue.

TimoPtr:
@jpelgrom I've applied all the suggestion and move this branch on top of #6422

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsFragment.kt
TimoPtr: https://github.com/home-assistant/home-assistant.io/pull/43488

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -138,7 +189,14 @@ internal fun AssistSettingsScreen(
             Spacer(modifier = Modifier.height(HADimens.SPACE2))
 
             // Wake Word Section
-            SectionHeader(text = stringResource(commonR.string.assist_wake_word_title))
+            Row(
+                horizontalArrangement = Arrangement.spacedBy(HADimens.SPACE2),
+                verticalAlignment = Alignment.CenterVertically,
+            ) {
+                SectionHeader(text = stringResource(commonR.string.assist_wake_word_title))
+                HALabel("Experimental", variant = LabelVariant.WARNING)
```
jpelgrom: String should be translated

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -138,7 +189,14 @@ internal fun AssistSettingsScreen(
             Spacer(modifier = Modifier.height(HADimens.SPACE2))
 
             // Wake Word Section
-            SectionHeader(text = stringResource(commonR.string.assist_wake_word_title))
+            Row(
+                horizontalArrangement = Arrangement.spacedBy(HADimens.SPACE2),
+                verticalAlignment = Alignment.CenterVertically,
+            ) {
+                SectionHeader(text = stringResource(commonR.string.assist_wake_word_title))
```
jpelgrom, line 206: The section header has built-in padding which means the text is not aligned. Would be cleaner to have them aligned.

Image

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -82,44 +124,53 @@ fun AssistSettingsScreen(viewModel: AssistSettingsViewModel) {
...
-        onStartTestWakeWord = {
-            if (testPermissionState.status.isGranted) {
-                viewModel.startTestWakeWord()
-            } else {
-                testPermissionState.launchPermissionRequest()
-            }
-        },
-        onStopTestWakeWord = viewModel::stopTestWakeWord,
-    )
+    Box(modifier = modifier.fillMaxSize()) {
```
jpelgrom: This kinda works if you really want to avoid a Scaffold, but any reason not to use one here?

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -82,44 +124,53 @@ fun AssistSettingsScreen(viewModel: AssistSettingsViewModel) {
...
+            onStartTestWakeWord = {
+                if (testPermissionState.status.isGranted) {
+                    viewModel.startTestWakeWord()
+                } else {
+                    testPermissionState.launchPermissionRequest()
+                }
+            },
+            onStopTestWakeWord = viewModel::stopTestWakeWord,
+        )
+        SnackbarHost(
```
jpelgrom: The action doesn't use the theme correctly (it is default M3 purple instead of the expected HA theme color)

Image

common/src/main/res/values/strings.xml
```diff
@@ -1213,7 +1213,8 @@
     <string name="assist_wake_word_stop_test">Stop testing</string>
     <string name="assist_wake_word_test_hint">Say \"%s\" to test</string>
     <string name="assist_wake_word_detected">Wake word detected!</string>
-    <string name="assist_wake_word_battery_warning">Wake word detection runs entirely on your device without any server. It continuously listens for audio, which may increase battery usage.</string>
+    <string name="assist_wake_word_battery_warning">Wake word detection runs entirely on your device, no network connection required. It continuously listens for your wake word, which will increase battery usage.</string>
+    <string name="assist_permission_microphone_required">Microphone permission is required</string>
```
jpelgrom: For consistency and to make it reusable in the future, I'd write this like missing sensor permissions

```suggestion
    <string name="assist_permission_microphone_required">Requires \'Microphone\' permission</string>
```

jpelgrom:
Couple of comments to address, after that good to merge for me.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
TimoPtr: I fought a bit with the scaffold and the insets and I gave up a bit quickly to be fair. I'm going to go back to the scaffold that should address the snack-bar color.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
TimoPtr: Good catch

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
TimoPtr: image
image

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -0,0 +1,499 @@
...
+    Scaffold(
+        snackbarHost = {
+            SnackbarHost(
+                hostState = snackbarHostState,
+                modifier = Modifier
+                    .padding(safeBottomPaddingValues(applyHorizontal = false)),
+            )
+        },
+        contentWindowInsets = WindowInsets(),
+    ) {
```
jpelgrom:
```suggestion
    ) { contentPadding ->
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -0,0 +1,499 @@
...
+            onSelectWakeWord = viewModel::onSelectWakeWordModel,
+            onStartTestWakeWord = {
+                if (testPermissionState.status.isGranted) {
+                    viewModel.startTestWakeWord()
+                } else {
+                    testPermissionState.launchPermissionRequest()
+                }
+            },
+            onStopTestWakeWord = viewModel::stopTestWakeWord,
+            modifier = Modifier.padding(it),
```
jpelgrom:
```suggestion
            modifier = Modifier.padding(contentPadding),
```

TimoPtr:
Downgrade KSP from 2.3.5 to 2.3.4 to fix the NullPointerException warning
that appears during CI builds:

'Cannot invoke ksp.com.intellij.openapi.application.Application.getService()
because ApplicationManager.getApplication() is null'

This is a known issue in KSP 2.3.5 (https://github.com/google/ksp/issues/2763) that occurs when
building from terminal/CI without IntelliJ IDEA running.

KSP 2.3.4 does not have this issue.

jpelgrom:
> This is a known issue in KSP 2.3.5 ([google/ksp#2763](https://github.com/google/ksp/issues/2763)) that occurs when building from terminal/CI without IntelliJ IDEA running.
> 
> KSP 2.3.4 does not have this issue.

The linked issue was marked as fixed, so 2.3.6 (when released) should also be good again.

PR #6348, opened 2026-01-29
feat: introduce shortcuts v2

hdcodedev:
## Checklist


jpelgrom:
👋 A quick note as this branch is getting very big and has tons of commits. As a best practice it is preferred to [keep PRs small](https://developers.home-assistant.io/docs/android/best_practices/#keep-your-prs-small). Did you consider this / do you intend to break this up before marking it as ready for review?

For these general "modernise screen and migrate to M3 components" changes my preferred approach would be:
1. Modernise code for the configuration with a proper ViewModel, flows, ...
2. Basic/quick conversion to a screen with Compose and M3 components - keep the same functionality and general look, use existing components or default M3 components, maybe a small class for the UI state
3. More extensive redesigns, if necessary - in this situation maybe broken up in two parts
   - App shortcuts
   - Launcher shortcuts

(1/2 can be done in either order)

hdcodedev:
> 👋 A quick note as this branch is getting very big and has tons of commits. As a best practice it is preferred to [keep PRs small](https://developers.home-assistant.io/docs/android/best_practices/#keep-your-prs-small). Did you consider this / do you intend to break this up before marking it as ready for review?
> 
> For these general "modernise screen and migrate to M3 components" changes my preferred approach would be:
> 
> 1. Modernise code for the configuration with a proper ViewModel, flows, ...
> 2. Basic/quick conversion to a screen with Compose and M3 components - keep the same functionality and general look, use existing components or default M3 components, maybe a small class for the UI state
> 3. More extensive redesigns, if necessary - in this situation maybe broken up in two parts
>    
>    * App shortcuts
>    * Launcher shortcuts
> 
> (1/2 can be done in either order)

Thank you for suggestions, if I knew that it would be that much big I would introduce partial PRs, but we can do it in this stage also.

The thing is, we initially tried working only on the M3 migration since that was the ticket scope, but we discovered many issues with the current code implementation.

Old Shortcuts V1 had a lot of design issues, so working on it would have meant changing 90% of the code. I actually tried this already, and in my opinion, it wasn’t worth it. In this case, it was better to keep the existing V1 and introduce V2 in parallel.

The V2 version introduces a testing layer, which wasn’t possible in V1 because the ViewModel wasn’t ready for that.

hdcodedev:
It will be moved to partial PRs

PR #6380, opened 2026-02-03
Introduce ServerSessionManager to handle auth and session state

TimoPtr:
## Checklist

I decided to split the auth/session management used in the WebView in a dedicated class to avoid creating a giant ViewModel. The idea is to have things related in the same manager and injected into the viewModel.

- [x] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

PR #6382, opened 2026-02-04
Add frontend url manager

TimoPtr:
## Summary

Introduces FrontendUrlManager, a dedicated class for handling URL resolution and security checks for the frontend WebView. This extracts URL loading logic from WebViewPresenterImpl into a focused, testable component.                                                                                         

## Checklist

## Any other notes

This code is not yet used but it is going to be in a next PR this is based on #6380

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/url/UrlLoadResult.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/url/UrlLoadResult.kt
@@ -39,8 +39,8 @@ sealed interface UrlLoadResult {
      * to configure home network detection or proceed anyway.
      *
      * @property serverId The server ID with blocked insecure connection
-     * @property missingHomeSetup True if home network Wi-Fi SSID/BSSID is not configured
-     * @property missingLocation True if location permission is needed for home network detection
+     * @property missingHomeSetup `true` if home network Wi-Fi SSID/BSSID is not configured
+     * @property missingLocation `true` if location permission is needed for home network detection
      */
     data class InsecureBlocked(val serverId: Int, val missingHomeSetup: Boolean, val missingLocation: Boolean) :
         UrlLoadResult
--- a/app/src/test/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManagerTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManagerTest.kt
@@ -230,7 +230,7 @@ class FrontendUrlManagerTest {
     }
 
     @Test
-    fun `Given security level already shown when serverUrlFlow then returns Success`() = runTest {
+    fun `Given plain text URL with null allowInsecureConnection and security level already shown when serverUrlFlow then returns Success`() = runTest {
         val connection = ServerConnectionInfo(
             externalUrl = "http://home.example.com",
             allowInsecureConnection = null,


app/src/main/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManager.kt
```diff
@@ -0,0 +1,162 @@
...
+            emit(UrlLoadResult.SessionNotConnected(actualServerId))
+            return@flow
+        }
+
+        serverManager.activateServer(actualServerId)
+
+        var pathConsumed = false
+        serverManager.connectionStateProvider(actualServerId).urlFlow().collect { urlState ->
+            val currentPath = if (pathConsumed) null else path
+            pathConsumed = true
```
jpelgrom: Would it make sense to not consume the path until the state actually returns an URL?

Example use case: I have sent a notification which deeplinks to something, but when I tap the notification I'm not yet connected to my home network so loading is blocked, and when I connect to Wi-Fi it reloads but ignores the path as it was already consumed by the blocked screen.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManager.kt
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * Mark security level as configured for server.
+     *
+     * After calling this, [serverUrlFlow] will no longer return [UrlLoadResult.SecurityLevelRequired]
+     * for this server during the current session.
+     *
+     * @param serverId The server ID that had security level configured
+     */
+    fun onSecurityLevelConfigured(serverId: Int) {
```
jpelgrom: The user can ignore the screen so it is not known whether or not they actually configured it.

```suggestion
    fun onSecurityLevelShown(serverId: Int) {
```

app/src/test/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManagerTest.kt
```diff
@@ -0,0 +1,403 @@
...
+            assertTrue(result is UrlLoadResult.Success)
+            val success = result as UrlLoadResult.Success
+            // Should not contain the entityId path in URL
+            assertEquals("https://home.example.com/?external_auth=1", success.url)
+            awaitComplete()
+        }
+    }
+
+    @Test
+    fun `Given insecure state when serverUrlFlow then returns InsecureBlocked`() = runTest {
```
jpelgrom: This test is for nothing setup. The next test is for home setup but not home + location disabled. I don't see tests for the other possible combinations. Why pick these two?

app/src/test/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManagerTest.kt
```diff
@@ -0,0 +1,403 @@
...
+        urlManager.serverUrlFlow(serverId = ServerManager.SERVER_ID_ACTIVE, path = null).test {
+            val result = awaitItem()
+            assertTrue(result is UrlLoadResult.Success)
+            assertEquals(42, (result as UrlLoadResult.Success).serverId)
+            awaitComplete()
+        }
+
+        coVerify { serverManager.activateServer(42) }
+        coVerify { serverManager.connectionStateProvider(42) }
+    }
```
jpelgrom: Should be moved up or down, right now it is in the middle of the connection security level related tests

app/src/test/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManagerTest.kt
```diff
@@ -0,0 +1,403 @@
...
+            externalUrl = "http://home.example.com",
+            allowInsecureConnection = false,
+        )
+        val server = createTestServer(id = 1, connectionInfo = connection)
+        coEvery { serverManager.getServer(1) } returns server
+        coEvery { sessionManager.isSessionConnected(1) } returns SessionCheckResult.Connected
+        coEvery { serverManager.activateServer(1) } just runs
+        coEvery { serverManager.connectionStateProvider(1) } returns connectionStateProvider
+        every { connectionStateProvider.urlFlow(any()) } returns flowOf(
+            UrlState.HasUrl(URL("http://home.example.com")),
```
jpelgrom, line 290: Why does it do this? HTTP + disallow insecure shouldn't return anything.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManager.kt
TimoPtr: I'm not sure if we want to change the behavior on this PR, but yes it would make sense. I saw also this PR https://github.com/home-assistant/android/pull/6403 that is somehow related.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManager.kt
TimoPtr: I've rename the callback in the screen to onSecurityLevelDone then it call this shown method from the viewModel to try to keep it valid in terms of naming, instead of using configured.

app/src/test/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManagerTest.kt
TimoPtr: I'm gonna make it parametrized to test the combinations.

app/src/test/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManagerTest.kt
TimoPtr: True it doesn't make sense. I'm going to remove this test. It is actually showing that the logic is in the lower layer and no logic is applied in the FrontendUrlManager.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManager.kt
TimoPtr: It's quite easy to address I'll add it to this PR, but we will need to see if users find edge cases that we didn't think about with this new change.

TimoPtr:
For some reason the test fail on this PR, it doesn't locally and on the next PR ...

app/src/main/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManager.kt
jpelgrom: I think that PR leaks path info when a server is removed to the remaining server, but haven't fully tested it yet because it is still in draft and I knew you were working on these changes.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/url/UrlLoadResult.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Insecure (HTTP) connection was blocked based on user's security settings.
+     *
+     * The UI should show the blocked insecure connection screen with options
+     * to configure home network detection or proceed anyway.
+     *
+     * @property serverId The server ID with blocked insecure connection
+     * @property missingHomeSetup True if home network Wi-Fi SSID/BSSID is not configured
+     * @property missingLocation True if location permission is needed for home network detection
```
jpelgrom: This isn't Python code ;)

```suggestion
     * @property missingHomeSetup `true` if home network Wi-Fi SSID/BSSID is not configured
     * @property missingLocation `true` if location permission is needed for home network detection
```

app/src/test/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManagerTest.kt
```diff
@@ -0,0 +1,394 @@
...
+        urlManager.serverUrlFlow(serverId = 1, path = null).test {
+            val result = awaitItem()
+            assertTrue(result is UrlLoadResult.SecurityLevelRequired)
+            assertEquals(1, (result as UrlLoadResult.SecurityLevelRequired).serverId)
+            awaitComplete()
+        }
+    }
+
+    @Test
+    fun `Given security level already shown when serverUrlFlow then returns Success`() = runTest {
```
jpelgrom: Consistent test naming - it makes sense when reading it in order of the file but might not otherwise

```suggestion
    fun `Given plain text URL with null allowInsecureConnection and security level already shown when serverUrlFlow then returns Success`() = runTest {
```

jpelgrom (approved):

PR #6386, opened 2026-02-04
Introduce new frontend screen

TimoPtr:
## Summary

Introduce New Compose-based frontend screen behind `USE_FRONTEND_V2` feature flag. To make it easier to manage I decided that everything that blocks the URL to be load should be within the same screen. That's why we have a concept of overlay with the SecurityLevel and BlockScreen. I make sure that the WebView is also always behind so that if we come back from this state it is not in a strange state where the URL cannot be load.

This screen is using MVI (mostly) and delegate some logics to the handler or other view models to keep it small and easy to maintain.

The error are not displayed with dialog anymore but a full screen with the troubleshooting section made by @hdcodedev

I've added a way to open the SettingsActivity from the navGraph.

At this stage it only implement a subset of all the features of the current WebViewActivity here potentially a non exhaustive list of everything that needs to be made in future PRs:

- WebView version log (auto/manual enable web view debug)
- Show changelog
- Check location permission disabled 
- Navigate to first view when restored
- Force update all sensors onResume
- Haptic
- PiP supports
- Add different actions based on the errors
- Keep screen on
- Screen orientation forced
- Start SensorWorker
- Start WebSocketManager
- Secure URL loading and navigate for external URL
- WebChromeClient (dialogs)
- Upload files
- Cookie management
- Permission for Notif + battery
- Clear history
- Switch server
- Gesture actions
- Custom view from the Webview
- ExoPlayer overlay
- Other externals messages
- Top bar color
- Improv
- Matter
- Lock app
- Deep link (maybe try to not reopen a new activity to improve the experience)
- Basic auth support
- AutoPlay
- Fullscreen
- Zoom
- Downloading
- NFC
- Add to

## Checklist

### DIFF
--- a/build-logic/convention/src/main/kotlin/AndroidComposeConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/AndroidComposeConventionPlugin.kt
@@ -35,7 +35,7 @@ class AndroidComposeConventionPlugin : Plugin<Project> {
 
             // Screenshot test worker memory grows with test count. Increase as needed.
             // Tracking: https://issuetracker.google.com/issues/469819154
-            val maxHeapSizeScreenshotTesting = "4g"
+            val maxHeapSizeScreenshotTesting = "5g"
 
             tasks.withType<PreviewScreenshotValidationTask>().configureEach {
                 maxHeapSize = maxHeapSizeScreenshotTesting


app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -0,0 +1,359 @@
...
+import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+
+/** Maximum time to wait for the frontend to load before showing a timeout error. */
+@VisibleForTesting
+val CONNECTION_TIMEOUT = 20.seconds
```
jpelgrom: If this is modelled after the existing  `CONNECTION_DELAY` it should be 10 seconds instead. 20 feels very long to me.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -0,0 +1,359 @@
...
+    fun onRetry() {
+        _viewState.update {
+            FrontendViewState.LoadServer(serverId = it.serverId)
+        }
+        loadServer()
+    }
+
+    /**
+     * Show the security level configuration screen.
+     */
```
jpelgrom:
```suggestion
    /** Show the security level configuration screen. */
```

You've done one line comments in other places as well, let's try to keep this small.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -0,0 +1,359 @@
...
+    fun onShowSecurityLevelScreen() {
+        _viewState.update {
+            FrontendViewState.SecurityLevelRequired(serverId = it.serverId)
+        }
+    }
+
+    /**
+     * Switch to a different server.
+     */
```
jpelgrom:
```suggestion
    /** Switch to a different server. */
```

You've done one line comments in other places as well, let's try to keep this small.

(Does this even need a comment when it's almost literally the function name?)

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -0,0 +1,359 @@
...
+                        currentState
+                    }
+                }
+            }
+
+            is FrontendHandlerEvent.Disconnected -> {
+                // Disconnection handling not yet implemented
+            }
+
+            is FrontendHandlerEvent.ConfigSent -> {
```
jpelgrom: I think it would be nice to bundle all these "no action needed" events into one case, instead of different ones

app/src/test/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModelTest.kt
```diff
@@ -0,0 +1,570 @@
...
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.RegisterExtension
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@ExtendWith(ConsoleLogExtension::class)
+class FrontendViewModelTest {
```
jpelgrom, line 42: As you add more of the required functionality, I imagine this file becoming a lot larger. Maybe it is helpful to already start adding one or two blocks with the tests nested inside?

TimoPtr: Ideally I would like to keep the VM small, and delegate things to smaller class. But yes we might need to break this beast.

TimoPtr:
> I generally like where this is going, even if it is very barebones right now. What do you consider as the minimum functionality required before merging this WIP screen?
>

I guess we could merge after this first batch of PR, since I mostly didn't touch the current production code. It would be only annoying for us while building in dev. But it would force me to deliver the features !
 
> Would it make sense to have 'app locked ' as another overlay state?\

Yes probably :)

> 
> I'll most likely come back later with more thoughts.

Take your time I'm going to have to rebase every thing after I've address the comments.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
TimoPtr: I kept two that needs to be implemented and grouped the no op one together.

jpelgrom:
I'll wait with a final review here for you to merge everything, but at a second glance it looks good enough for debug (no crashes or major issues, and features can then be added by themselves instead of needing a giant stack).

TimoPtr:
Finally the top of the iceberg is ready to be reviewed...

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
```diff
@@ -0,0 +1,529 @@
...
+@Composable
+private fun WebViewEffects(webView: WebView?, url: String, scriptsToEvaluate: Flow<WebViewScript>) {
+    if (webView != null) {
+        LaunchedEffect(webView, url) {
+            Timber.v("Load url ${if (BuildConfig.DEBUG) url else ""}")
+            webView.loadUrl(url)
+        }
+        LaunchedEffect(webView) {
+            scriptsToEvaluate.collect { webViewScript ->
+                Timber.d("Evaluating script: ${webViewScript.script}")
```
jpelgrom, line 413: Maybe this should also be redacted when not using debug (considering token again)


```suggestion
                Timber.d("Evaluating script: ${if (BuildConfig.DEBUG) webViewScript.script else ""}")
```

TimoPtr:
> After testing this for a couple of days, the basis seems solid and good to merge. Might've missed something but as long as we keep it in beta that should be something we can handle in time. Also lots of features to add 😅.
> 
> More for the giant list:
>  - Edge to edge support is enabled again, but we previously disabled it because it could create issues with themes / color contrast in frontends. I think it should be disabled again (or the frontend change needs to arrive before merging this).
>  - When encountering a loading failure, it is no longer possible to go to settings. This could be problematic if you edit a URL and make a typo, or want to access app settings while your server is down.

I've made a special task in my list to adjust the error screen actions based on the error. I'll definitely always add open settings.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
TimoPtr: I'm going to merge this PR and address this comment in the sensitive PR that I'm going to merge just after.

PR #6395, opened 2026-02-05
Use painter instead of ImageVector to avoid crashes in old version

TimoPtr:
## Summary

We already replaced this for complex icon using gradient but it seems that is also crashes for simple icon on some device. So one crash in Sentry on the Welcome screen about missing a pathData. I decided to migrate everything to a painterResource instead of the VectorImage.


jpelgrom:
- Can you give a stacktrace?
 - Why do you specifically mention API < 24 in the comment? I cannot easily find that as a minimum.

TimoPtr:
> Why do you specifically mention API < 24 in the comment? I cannot easily find that as a minimum.

It was for the gradient and I kept it thinking it was the same issue. I can't find the issue anymore in Sentry.... It doesn't seems to be a big issue closing the issue for now until we find it back.

Issue #6000, opened 2025-11-05
Introduce CoroutineBroadcastReceiver


TimoPtr:
### Description

Most of our `BroadcastReceiver` are using coroutines and all of them are creating a new scope not tight to any lifecycle. 
According to the documentation of `BroadcastReceiver` the execution of the `onReceive` method should end within 10s otherwise the system is going to raise an ANR. There is a way to extend this to 30s using `goAsync`, but the app is not using it and we probably shouldn't and try to fit everything under 10s.

A parallel can be made with the `Worker` API Google introduced a `CoroutineWorker`, the difference is that it has a lifecycle. But still we could use this API to help us design an API to help us/standardize how we deal with coroutine in `BroadcastReceiver`.

### Additional context

This has been first discussed in https://github.com/home-assistant/android/pull/5998

A proposal of API could be 

```kotlin
abstract class CoroutineBroadcastReceiver : BroadcastReceiver() {
    final override fun onReceive(context: Context, intent: Intent) {
        GlobalScope.launch {
            withTimeout(10.seconds) {
                onReceiveIntent(context, intent)
            }
        }
    }

    abstract suspend fun onReceiveIntent(context: Context, intent: Intent)
}
```

No need to create a new scope since this new scope would not be tight to any lifecycle.

TimoPtr:
`BaseGlanceEntityWidgetReceiver` is impacted by this issue. It means that we are staying more than 5s in the onReceived()

TimoPtr:
Maybe we could delegate all the management to a Worker instead of doing the work in the receiver.

PR #6396, opened 2026-02-05
Move away from Main in Broadcast receiver and use goAsync

TimoPtr:
## Summary

It's not addressing this https://github.com/home-assistant/android/issues/6000 but it is a first step to reduce the number of ANR we have on Sentry. The biggest issue we have today is due to background ANR and I suspect it is because we do everything on Main where we don't need in the broadcast receiver.
image

The receiver code doesn't need to be updated with Main, nor anything else we have. The only place where I fallback to Main was to display a Toast.

## Checklist

## Any other notes

This might introduce regressions, I decided to also make sure the scopes are isolated because `scopeWidget` from the base class companion object was used in classes that doesn't extend the base class ...

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
@@ -19,6 +19,7 @@ import io.homeassistant.companion.android.common.data.integration.SensorRegistra
 import io.homeassistant.companion.android.common.data.servers.ServerManager
 import io.homeassistant.companion.android.common.data.websocket.impl.entities.GetConfigResponse
 import io.homeassistant.companion.android.common.util.CHANNEL_SENSOR_SYNC
+import io.homeassistant.companion.android.common.util.launchAsync
 import io.homeassistant.companion.android.database.DatabaseEntryPoint
 import io.homeassistant.companion.android.database.sensor.SensorDao
 import io.homeassistant.companion.android.database.sensor.SensorWithAttributes
@@ -35,6 +36,7 @@ import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.delay
@@ -68,7 +70,7 @@ abstract class SensorReceiverBase : BroadcastReceiver() {
         }
     }
 
-    private val ioScope: CoroutineScope = CoroutineScope(Dispatchers.IO + Job())
+    private val sensorScope: CoroutineScope = CoroutineScope(Dispatchers.Default + SupervisorJob())
 
     protected abstract val currentAppVersion: String
     protected abstract val managers: List<SensorManager>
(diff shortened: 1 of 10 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
```diff
@@ -68,7 +70,7 @@ abstract class SensorReceiverBase : BroadcastReceiver() {
         }
     }
 
-    private val ioScope: CoroutineScope = CoroutineScope(Dispatchers.IO + Job())
+    private val sensorScope: CoroutineScope = CoroutineScope(Dispatchers.Default + SupervisorJob())
```
jpelgrom, line 73: This wasn't on main, why did you change it to `Default` instead of `IO`?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/button/ButtonWidget.kt
```diff
@@ -135,14 +134,19 @@ class ButtonWidget : AppWidgetProvider() {
         super.onReceive(context, intent)
         when (action) {
             CALL_SERVICE_AUTH -> authThenCallConfiguredAction(context, appWidgetId)
-            CALL_SERVICE -> callConfiguredAction(context, appWidgetId)
-            BaseWidgetProvider.UPDATE_WIDGETS -> updateAllWidgets(context)
-            Intent.ACTION_SCREEN_ON -> updateAllWidgets(context)
+            CALL_SERVICE -> launchAsync(widgetScope) { callConfiguredAction(context, appWidgetId) }
+            BaseWidgetProvider.UPDATE_WIDGETS, Intent.ACTION_SCREEN_ON -> launchAsync(widgetScope) {
```
jpelgrom: Why didn't you change it to the new function / handle `onReceive` in the base class first here?

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/SensorReceiverBase.kt
TimoPtr: It is not making operations that needs IO like network or reading files. In that case we should prefer to use `Default`

https://www.baeldung.com/kotlin/io-and-default-dispatcher#1-when-to-use-the-default-dispatcher

> This makes it suitable for CPU-bound tasks that require a lot of computation and benefit from parallelism. Short-running calculations and small-scale data processing tasks are the ideal candidates for running on the Default dispatcher.

Using IO for network or IO is important because of its elasticity property, where Default is tight to the number of physical thread available. We don't want more thread than the CPU has to avoid context switching, when it the work is only bound to the CPU operations, otherwise the result would be that the app becomes very slow instead of just making tasks one after the other.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/button/ButtonWidget.kt
jpelgrom: `onReceive` in BaseWidgetProvider now handles some of this and you introduced `onReceiveIntentNotHandled` specifically for any intents specific for the widget.

jpelgrom: Moved to #6415

jpelgrom (approved):

jpelgrom:
CI should be good once GitHub [sorts itself out.](https://www.githubstatus.com/incidents/smf24rvl67v9)

PR #6414, opened 2026-02-09
Use Github Action instead of renovate to update lockfile

TimoPtr:
## Summary

We were using renovate to update our lockfile but this doesn't work for modules that uses the NDK like `:microfrontend` that we've just introduced #6312. Also I found out that sometimes renovate was only making partial update of the lockfile since it was not using gradle to generate it. This new approach is cleaner from the Gradle perspective. 

Now each time renovate open a PR a new GA workflow is triggered and commit if needed the lockfiles update.

## Any other notes

We might need to fix some stuff after the first PR from renovate.

### DIFF
--- a/.github/workflows/renovate-lockfiles.yml
+++ b/.github/workflows/renovate-lockfiles.yml
@@ -26,9 +26,6 @@ jobs:
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         if: steps.check-skip.outputs.skip != 'true'
-        with:
-          # Use TAG_PUSH_TOKEN so the push triggers pr.yml workflow
-          token: ${{ secrets.TAG_PUSH_TOKEN }}
 
       - uses: ./.github/actions/setup-build-env
         if: steps.check-skip.outputs.skip != 'true'
@@ -52,3 +49,15 @@ jobs:
             git commit -m "Update dependency lockfiles"
             git push
           fi
+
+      - name: Trigger Workflow
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
+        if: steps.check-skip.outputs.skip != 'true'
+        with:
+          script: |
+            github.rest.actions.createWorkflowDispatch({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              workflow_id: 'pr.yml',
+              ref: context.ref,
+            })


TimoPtr:
@jpelgrom that might be annoying since it is not going to trigger the CI like for the update of the screenshots...

jpelgrom:
> @jpelgrom that might be annoying since it is not going to trigger the CI like for the update of the screenshots...

That sounds very annoying, now every dependency update would need a maintainer to do an empty push to trigger CI. It also hides any failures that the dependency update might cause until we make such a push, as the run before will fail because lockfiles changed.

TimoPtr:
> > @jpelgrom that might be annoying since it is not going to trigger the CI like for the update of the screenshots...
> 
> That sounds very annoying, now every dependency update would need a maintainer to do an empty push to trigger CI. It also hides any failures that the dependency update might cause until we make such a push, as the run before will fail because lockfiles changed.

I've tried something that might do the trick that is used for the release workflow. Let merge it and see if it works. If it works we should also adjust the manual workflow for the screenshots.

jpelgrom:
Where does this new token come from? (ie is it registered to an app or you personally)

https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#triggering-a-workflow-from-a-workflow suggests "events triggered by the `GITHUB_TOKEN` (...) will not create a new workflow run.". The event that would trigger a workflow run here is push - so I think the commit action would need to use a different token?

.github/workflows/renovate-lockfiles.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: |
+          if [[ "$COMMIT_AUTHOR" == "github-actions[bot]@users.noreply.github.com" ]]; then
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        if: steps.check-skip.outputs.skip != 'true'
+        with:
+          # Use TAG_PUSH_TOKEN so the push triggers pr.yml workflow
+          token: ${{ secrets.TAG_PUSH_TOKEN }}
```
TimoPtr: It comes from weekly.yml while looking at it again it is a bad idea. Instead I'm going to trigger the workflow manually, it means that I'll make it possible to trigger the PR workflow manually.

.github/workflows/pr.yml
```diff
@@ -1,6 +1,7 @@
 name: Pull Request
 
 on: # yamllint disable-line rule:truthy
+  workflow_dispatch:
```
TimoPtr, line 4: I think it is actually helpful to be able to manually trigger this worfklow, like for the screenshot update instead of making a dumb or empty commit.

renovate.json5
```diff
@@ -2,14 +2,16 @@
     $schema: 'https://docs.renovatebot.com/renovate-schema.json',
     configMigration: true,
     extends: [
-        'config:recommended',
+        'config:best-practices',
```
TimoPtr, line 5: A small suggestion from renovate for us to migrate to this config instead.

PR #6415, opened 2026-02-09
Move away from Main in Widgets

TimoPtr:
## Summary

This PR is a best effort to reduce the numbers of ANR within the widget receivers. I also found out that we are using BaseWidget static attributes in widgets that are not extending the Base class.

The main idea is to move away from the Main thread since it is not needed at all and put pressure on the Main thread for nothing. Unfortunately I don't see a clear path on how to not make the received live for a long period, because of that we might still have ANRs.

Thanks to Copilot comments on this PR I also found out that we were re registering the receiver on each screen_on event. I'm not sure of what it does to the system, but we shouldn't do it since it is already done in the Application.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseWidgetProvider.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseWidgetProvider.kt
@@ -36,7 +36,7 @@ abstract class BaseWidgetProvider<T : WidgetEntity<T>, DAO : WidgetDao<T>> : App
         private val widgetEntities = mutableMapOf<Int, List<String>>()
         private val widgetJobs = mutableMapOf<Int, Job>()
 
-        fun newCoroutineScopeProvider() = CoroutineScope(Dispatchers.Default + SupervisorJob())
+        private fun newCoroutineScopeProvider() = CoroutineScope(Dispatchers.Default + SupervisorJob())
     }
 
     @Inject
@@ -46,6 +46,9 @@ abstract class BaseWidgetProvider<T : WidgetEntity<T>, DAO : WidgetDao<T>> : App
     lateinit var dao: DAO
 
     protected var lastIntent = ""
+
+    protected val widgetScope
+        get() = Companion.widgetScope
 
     init {
         setupWidgetScope()
@@ -141,9 +144,6 @@ abstract class BaseWidgetProvider<T : WidgetEntity<T>, DAO : WidgetDao<T>> : App
         widgetEntities.clear()
         widgetJobs.clear()
     }
-
-    protected val widgetScope
-        get() = Companion.widgetScope
 
     private suspend fun updateAllWidgets(context: Context) {
         val widgetProvider = getWidgetProvider(context)


jpelgrom:
@TimoPtr From the title, do you want to update this as we reverted the goAsync part in #6459?

TimoPtr:
> @TimoPtr From the title, do you want to update this as we reverted the goAsync part in #6459?

I'm not sure yet, I want to work on the work I've rollbacked. This PR we could only keep the switch of dispatcher, but it's not urgent I think and might be better to do with the workers. Let's put this PR on draft until I come back on it.

TimoPtr:
@jpelgrom I decided to remove the goAsync part from this PR since we currently need the receiver to stay alive for as long as the screen is ON, but we still shouldn't be in Main.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidget.kt
```diff
@@ -53,24 +54,22 @@ class TemplateWidget : AppWidgetProvider() {
...
     private var lastIntent = ""
 
     init {
         setupWidgetScope()
     }
 
     private fun setupWidgetScope() {
-        if (widgetScope == null || !widgetScope!!.isActive) {
+        if (!widgetScope.isActive) {
             widgetScope = CoroutineScope(Dispatchers.Main + Job())
```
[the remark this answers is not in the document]
TimoPtr: Good catch.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidget.kt
```diff
@@ -115,62 +116,58 @@ class TemplateWidget : AppWidgetProvider() {
...
-                }
+        val allWidgets = templateWidgetDao.getAll()
+        val widgetsWithDifferentTemplate = allWidgets.filter { it.template != widgetTemplates[it.id] }
+        if (widgetsWithDifferentTemplate.isNotEmpty()) {
+            ContextCompat.registerReceiver(
+                context.applicationContext,
+                this@TemplateWidget,
+                IntentFilter(Intent.ACTION_SCREEN_OFF),
+                ContextCompat.RECEIVER_NOT_EXPORTED,
+            )
```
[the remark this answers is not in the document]
TimoPtr: It's actually a good comment we should not register here it's already done from the `HomeAssistantApplication`. Removing.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseWidgetProvider.kt
```diff
@@ -44,24 +45,24 @@ abstract class BaseWidgetProvider<T : WidgetEntity<T>, DAO : WidgetDao<T>> : App
...
     init {
         setupWidgetScope()
     }
 
     private fun setupWidgetScope() {
-        if (widgetScope == null || !widgetScope!!.isActive) {
-            widgetScope = CoroutineScope(Dispatchers.Main + Job())
-            thisSetScope = true
+        if (!widgetScope.isActive) {
+            Companion.widgetScope = newCoroutineScopeProvider()
```
jpelgrom, line 59: This is different from how you do it in TemplateWidget, no Companion used there

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseWidgetProvider.kt
```diff
@@ -70,82 +71,80 @@ abstract class BaseWidgetProvider<T : WidgetEntity<T>, DAO : WidgetDao<T>> : App
...
+            widgetScope.cancel()
+        } catch (e: IllegalStateException) {
+            Timber.w(e, "Calling onScreenOff without any job started")
         }
+        widgetEntities.clear()
+        widgetJobs.clear()
     }
 
+    protected val widgetScope
+        get() = Companion.widgetScope
```
jpelgrom: Please move this towards the top next to other values, instead of hiding it between functions

app/src/main/kotlin/io/homeassistant/companion/android/widgets/BaseWidgetProvider.kt
TimoPtr: Yes the base class is hiding the var and only exposing a val I don't want it to mess up.

PR #6376, opened 2026-02-03
Send wake word when starting voice pipeline if needed

TimoPtr:
## Summary

Send `wake-word-phrase` when we start assist from a wake word. It is useful to avoid starting Assist on multiple devices when you are in a room with multiple assist devices.

## Checklist

- [x] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

PR #6419, opened 2026-02-10
Hide Assist on duplicate wake word received

TimoPtr:
## Summary

In order to replicate the behavior of a VPE that when multiples pipeline starts with the same wake word only the first one is kept we need to check the error sent in the WebSocket. This PR does that. 

It works by starting a pipeline and send the wake word phrase (introduced by https://github.com/home-assistant/android/pull/6376) then if the pipeline send an error with a specific code we simply stop the AssistActivity. To make the UX nicer I decided to delay a bit the display of the bottom sheet otherwise it was shown and hidden in less than a second. The counterpart is that the UI blinks a bit because of the activity overlay.

## Checklist

## Screenshots

https://github.com/user-attachments/assets/d385fdcf-e7af-4836-aa3d-cb791628fbfc

### DIFF
--- a/common/src/test/kotlin/io/homeassistant/companion/android/common/assist/AssistViewModelBaseTest.kt
+++ b/common/src/test/kotlin/io/homeassistant/companion/android/common/assist/AssistViewModelBaseTest.kt
@@ -233,12 +233,13 @@ class AssistViewModelBaseTest {
     }
 
     @Test
-    fun `Given voice pipeline When STT_START received Then PipelineStarted event is emitted`() = runTest {
+    fun `Given voice pipeline When RUN_START received Then PipelineStarted event is emitted`() = runTest {
+        val handlerId = 42
         viewModel.setupRecorder()
         viewModel.runVoicePipeline()
         advanceUntilIdle()
 
-        pipelineEventsFlow.emit(createSttStartEvent())
+        pipelineEventsFlow.emit(createRunStartEvent(handlerId))
         advanceUntilIdle()
 
         assertEquals(listOf(AssistEvent.PipelineStarted), viewModel.receivedEvents)


common/src/main/kotlin/io/homeassistant/companion/android/common/assist/AssistViewModelBase.kt
```diff
@@ -160,7 +166,10 @@ abstract class AssistViewModelBase(
                         onEvent,
                     )
 
-                    AssistPipelineEventType.STT_START -> handleSttStart()
+                    AssistPipelineEventType.STT_START -> {
+                        handleSttStart()
+                        onEvent(AssistEvent.PipelineStarted)
```
jpelgrom, line 169: Maybe a bit theoretical, but what if I use the wake word and it triggers Assist with a pipeline without STT? Wouldn't it make more sense to link it to the RUN_START event?

TimoPtr: Indeed I was assuming there was always voice but I forget that you can have a pipeline without STT.

TimoPtr: It actually quite complex. Since the pipeline only starts when using voice otherwise we wait for the user to send his input to trigger the pipeline. So this logic of canceling Assist won't work for text in its current state. 

I'm checking if I can call the selected pipeline with only the wake word all the time when Assist start just to know if we can proceed or not, but I might just support voice by setting the `pending` boolean only when using microphone.

TimoPtr: I decided to only support voice pipeline otherwise it asks for a quite big changes in the logic and I'm not even sure if we can do it properly.

jpelgrom (approved):

PR #6426, opened 2026-02-11
Introduce sensitive utils function

TimoPtr:
## Summary

We have this pattern in logs where in logs we hide strings in release. I want to centralize the logic in one place if we ever want to enable this dynamically for any reason (the user might want to be able to see this sensitive information).

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -239,15 +239,13 @@ Timber.e(exception, "Failed to load user dashboard for userId=$userId")
 // Bad - no context
 Timber.e(exception)
 
-// Good - conditional logging for expensive operations
-if (BuildConfig.DEBUG) {
-    Timber.d("Complex debug info: ${expensiveOperation()}")
-}
+// Good - use sensitive() to hide user data in release logs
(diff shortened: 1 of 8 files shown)


jpelgrom:
OK with the goal and how it works, but did you consider naming the function something like `logSensitive` to make it more obvious this is for logging and not a general string utility function?

TimoPtr:
> OK with the goal and how it works, but did you consider naming the function something like `logSensitive` to make it more obvious this is for logging and not a general string utility function?

I struggle to find a name that I liked and is short :/

jpelgrom:
Yes I also like having something short, but can't really come up with a good alternative for now.

TimoPtr:
I have another idea, more complex about changing how Timber is taking parameters I need to try it out.

jpelgrom:
The git history here is incorrect, it seems to think you've made a lot of commits.

TimoPtr:
> The git history here is incorrect, it seems to think you've made a lot of commits.

It's on top the frontend PR but I didn't rebase it yet. FYI I didn't find a nice way to do what I wanted with Timber without changing how Timber is actually working. I might reconsider it later if needed.

PR #6434, opened 2026-02-12
Add missing catch that might crash the app

TimoPtr:
## Summary

We did have a lot of crashes due to `validateServerId` that throw an IllegalStateException that is not caught. This PR doesn't fix the cause of the issue, that might be because a server has been removed and we are still making some operation with its ID, but at least avoid the crash.

I'm not sure to have caught all the potential issues, since we don't have the full stacktrace thanks to the coroutine that doesn't propagate the callstack properly.


TimoPtr:
> OK as a fix to prevent unwanted crashes, but it would be really good to know why they are occuring in the first place.

We might need to enable some debugging flags to be able to understand where it happens and then try to reproduce

https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs%2Ftopics%2Fdebugging.md

PR #6435, opened 2026-02-12
Only register one network callback at the time

TimoPtr:
## Summary

Before this change, ServerConnectionStateProviderImpl is created per server (via @AssistedInject with serverId), so each server instance registered its own callback via observeHomeNetworkState() for each `urlFlow`. Combined with `NetworkStatusMonitorImpl` also registering its own callback, that was N+1 callbacks for N servers.
  If the `urlFlow` was collected multiple times or servers were re-created, callbacks would pile up and hit the limit.

I was able to see multiple callbacks registered but to hit 100 it requires a lot of subscription, it does impacts 40 users I wonder what they use to reach the limit.

I tested the behavior and it was keeping the callback register like it should without unregistering and registering again like it was before.

## Checklist

### DIFF
--- a/common/src/test/kotlin/io/homeassistant/companion/android/common/data/network/NetworkChangeObserverImplTest.kt
+++ b/common/src/test/kotlin/io/homeassistant/companion/android/common/data/network/NetworkChangeObserverImplTest.kt
@@ -5,6 +5,7 @@ import android.net.Network
 import android.net.NetworkCapabilities
 import android.net.NetworkRequest
 import app.cash.turbine.test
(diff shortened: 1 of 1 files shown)


jpelgrom:
Maybe this is intentional and you're relying on injection working correctly, but when looking at the tests, I'm missing a test where there are two observers and it verifies that
 - a second usage does not cause registration of a second callback
 - if one usage stops collecting and one remains, the callback is not removed until the other one also stops

TimoPtr:
I've made one test that test basically that the implementation is properly using `WhileSubscribed`.

PR #6436, opened 2026-02-12
Bluetooth mic routing fix

kikura:
## Description
Fixes an issue where the Bluetooth microphone was not being used when launching the voice assistant from a Bluetooth headset button. The app would fail to recognize voice commands unless the phone's microphone was within earshot.

## Problem
When triggering the voice assistant from a Bluetooth headset (e.g., pressing the button on AirPods), the app would display "no text recognised" even though the Bluetooth headset was functioning correctly with other voice assistants like Google Gemini.

## Solution
Enabled Bluetooth SCO (Synchronous Connection Oriented) audio routing by:
1. Changed audio source to `VOICE_COMMUNICATION` when Bluetooth SCO is available
2. Added `startBluetoothSco()` call in `requestFocus()` to enable SCO audio mode before recording
3. Added `stopBluetoothSco()` call in `abandonFocus()` to properly clean up SCO connection
4. Added API level checks to ensure compatibility with older Android versions

## Testing
- Launch the voice assistant by pressing the button on a Bluetooth headset
- Give voice commands
- Verify that the app now recognizes speech from the Bluetooth microphone

## Files Changed
- `common/src/main/kotlin/io/homeassistant/companion/android/common/util/AudioRecorder.kt

*Please not that the neccessary code has been generated by Co-Pilot, but has been tested and works.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -120,6 +120,7 @@ Source code is in Kotlin under `src/main/kotlin/io/homeassistant/companion/andro
 
 - **Language**: All code in English, Kotlin only
 - **Formatter**: KTLint enforces style via `.editorconfig`
+- **Imports**: When adding new imports, always add the import and its usage in the same edit. IDE linters and hooks automatically remove unused imports, so adding an import in a separate step before using it will cause the import to be stripped.
 - **Constants**: Use named constants instead of magic numbers/strings. Organize alongside classes (outside companion objects when possible), or in dedicated `*Constants.kt` files with `object` namespacing.
 - **Strong Types**: Use sealed classes/interfaces over enums or strings for logic control. Use Kotlin `Duration`/`Instant` instead of primitive types for time.
 - **TODOs**: Avoid TODOs. If required, link to GitHub issue: `// TODO Missing feature (linked issue #404)`
--- a/app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistViewModel.kt
@@ -172,13 +172,19 @@ class AssistViewModel @Inject constructor(
 
     private suspend fun checkSupport(): Boolean? {
         if (!serverManager.isRegistered()) return false
-        if (!serverManager.integrationRepository(
-                selectedServerId,
-            ).isHomeAssistantVersionAtLeast(2023, 5, 0)
-        ) {
-            return false
+        return try {
+            if (!serverManager.integrationRepository(
+                    selectedServerId,
+                ).isHomeAssistantVersionAtLeast(2023, 5, 0)
+            ) {
+                false
+            } else {
+                serverManager.webSocketRepository(selectedServerId).getConfig()?.components?.contains("assist_pipeline")
+            }
+        } catch (e: IllegalStateException) {
+            Timber.e(e, "Failed to check support")
+            false
         }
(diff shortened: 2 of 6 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/util/AudioRecorder.kt
```diff
@@ -49,6 +48,26 @@ class AudioRecorder(private val audioManager: AudioManager?) {
...
+     * Returns VOICE_COMMUNICATION when Bluetooth SCO is available off-call,
+     * otherwise returns MIC.
+     */
+    private fun getAudioSource(): Int {
+        if (audioManager == null) {
+            return AudioSource.MIC
+        }
+
+        // Check if Bluetooth SCO is available
+        return if (audioManager.isBluetoothScoAvailableOffCall()) {
```
TimoPtr: Use the property instead of the function
```suggestion
        return if (audioManager.isBluetoothScoAvailableOffCall) {
```

common/src/main/kotlin/io/homeassistant/companion/android/common/util/AudioRecorder.kt
TimoPtr, line 36: I would like a small update of the documentation that explains the behavior of the recorder.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/AudioRecorder.kt
```diff
@@ -49,6 +48,26 @@ class AudioRecorder(private val audioManager: AudioManager?) {
...
+            // Use VOICE_COMMUNICATION which is the standard for Bluetooth audio
+            AudioSource.VOICE_COMMUNICATION
```
TimoPtr, line 84: Any reason to not use that all the time, it might be beneficial to other scenario than only Bluetooth.

> Microphone audio source tuned for voice communications such as VoIP. It will for instance take advantage of echo cancellation or automatic gain control if available.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/AudioRecorder.kt
TimoPtr, line 41: From the  documentation

> Even if a SCO connection is established, the following restrictions apply on audio output streams so that they can be routed to SCO headset:
the stream type must be STREAM_VOICE_CALL
the format must be mono
the sampling must be 16kHz or 8kHz
The following restrictions apply on input streams:
the format must be mono
the sampling must be 8kHz

It means we should not use 44100 when we are going to use Bluetooth but 8000 otherwise you might get weird behavior.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/AudioRecorder.kt
```diff
@@ -114,6 +134,13 @@ class AudioRecorder(private val audioManager: AudioManager?) {
 
     private fun requestFocus() {
         if (audioManager == null) return
+        
+        // Enable Bluetooth SCO if available
+        if (audioManager.isBluetoothScoAvailableOffCall()) {
+            audioManager.startBluetoothSco()
```
TimoPtr: This method is deprecated and the documentation suggest to use `setCommunicationDevice` instead can you adjust and try again?

AudioRecorder.kt
```diff
@@ -0,0 +1,20 @@
+// Other existing imports
```
TimoPtr: Remove this file.

TimoPtr:
Take the time to review my comments, but also the code generated before asking for review.

Issue #6307, opened 2026-01-22
Rework UI of EntityWidgetConfigureActivity to compose and Material3


TimoPtr:
### Description

This screen is using XML and we should migrate it to compose and Material3

himu-gupta:
I am planning to work on this migration. I will preserve the existing entity widget configuration behavior while moving the screen to Compose and the Home Assistant Material 3 components, with focused tests for the configuration state and persistence.

PR #6444, opened 2026-02-16
Migrate TemplateWidgetConfigureActivity to Compose

amlwin:
## Summary

Migrates the Template Widget configuration screen from XML and Views to Jetpack Compose. This change replaces `TemplateWidgetConfigureActivity`'s XML layout with a new Composable screen, `TemplateWidgetConfigureView`.

A new `TemplateWidgetConfigureViewModel` is introduced to manage the screen's state and business logic, including template rendering and widget creation/updating. This also adds unit tests for the new ViewModel to ensure its correctness.

## Checklist

- [X] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [X] The changes have been thoroughly tested, and edge cases have been considered.

## Screenshots

No Ui  changes
Screenshot 2026-02-16 at 11 28 19 AM


.idea/kotlinc.xml
```diff
@@ -6,10 +6,6 @@
   <component name="Kotlin2JvmCompilerArguments">
     <option name="jvmTarget" value="11" />
   </component>
-  <component name="KotlinCommonCompilerArguments">
-    <option name="apiVersion" value="2.3" />
-    <option name="languageVersion" value="2.3" />
-  </component>
```
TimoPtr, line 12: This should have not be committed but I'm fine about keeping it since I think it became obsolete.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
-import kotlinx.serialization.SerializationException
-import timber.log.Timber
 
-// TODO Migrate to compose https://github.com/home-assistant/android/issues/6304
 @AndroidEntryPoint
-class TemplateWidgetConfigureActivity : BaseWidgetConfigureActivity<TemplateWidgetEntity, TemplateWidgetDao>() {
-    private lateinit var binding: WidgetTemplateConfigureBinding
+class TemplateWidgetConfigureActivity : BaseActivity() {
```
TimoPtr, line 18: For clarity move the compose screen into its own dedicated file.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
-    private var requestLauncherSetup = false
+    private val supportedTextColors: List<String>
+        get() = listOf(
+            application.getHexForColor(commonR.color.colorWidgetButtonLabelBlack),
+            application.getHexForColor(android.R.color.white),
+        )
 
-    public override fun onCreate(savedInstanceState: Bundle?) {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        enableEdgeToEdgeCompat()
```
TimoPtr:
```suggestion
```

Already in the base class

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
-            } else {
-                null
-            },
-            lastUpdate = dao.get(appWidgetId)?.lastUpdate ?: "Loading",
-        )
+    @RequiresApi(Build.VERSION_CODES.O)
+    private suspend fun requestPinWidget() {
+        try {
+            viewModel.requestWidgetCreation()
+            finish()
```
TimoPtr: I'm not entirely sure but we might have a race condition here. `requestWidgetCreation` emits an action that is going to be executed within the lifecycle of the activity. 

Finish is going to called when the emission of the action is going to be done, but it doesn't guarantee that the action is going to be handled before you actually call finish, which would lead to the action not being executed.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
-    private fun renderTemplateText() {
-        val editableText = binding.templateText.text ?: return
-        if (editableText.isNotEmpty()) {
-            renderTemplateText(editableText.toString())
-        } else {
-            binding.renderedTemplate.text = getString(commonR.string.empty_template)
-            binding.addButton.isEnabled = false
+    private suspend fun onUpdateWidget() {
+        try {
+            viewModel.updateWidgetConfiguration()
```
TimoPtr: same here.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
 
-    private fun renderTemplateText(template: String) {
-        val serverId = selectedServerId
-        if (serverId == null) {
-            Timber.w("Not rendering template because server is not set")
-            return
-        }
+    private fun showAddWidgetError() {
+        Toast.makeText(applicationContext, commonR.string.widget_creation_error, Toast.LENGTH_LONG).show()
+    }
```
TimoPtr, line 96: Let's not use Toast but a snackbar instead, especially since you already added a Scaffold.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
-            renderTemplateText(editableText.toString())
-        } else {
-            binding.renderedTemplate.text = getString(commonR.string.empty_template)
-            binding.addButton.isEnabled = false
+    private suspend fun onUpdateWidget() {
+        try {
+            viewModel.updateWidgetConfiguration()
+            setResult(RESULT_OK)
+            finish()
+        } catch (_: IllegalStateException) {
```
TimoPtr: You're throwing an exception with a message but you ignore it. You should at least log the exception with a warning.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
-                null
-            },
-            lastUpdate = dao.get(appWidgetId)?.lastUpdate ?: "Loading",
-        )
+    @RequiresApi(Build.VERSION_CODES.O)
+    private suspend fun requestPinWidget() {
+        try {
+            viewModel.requestWidgetCreation()
+            finish()
+        } catch (_: IllegalStateException) {
```
TimoPtr: Same here

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
+        selectedBackgroundType = viewModel.selectedBackgroundType,
+        onBackgroundTypeSelected = { viewModel.selectedBackgroundType = it },
+        textColorIndex = viewModel.textColorIndex,
+        onTextColorSelected = { viewModel.textColorIndex = it },
+        isUpdateWidget = viewModel.isUpdateWidget,
+        onActionClick = onActionClick,
+    )
+}
+
+@Suppress("ComposeUnstableCollections") // Matches ServerExposedDropdownMenu signature; same as TodoWidgetConfigureActivity
```
TimoPtr:
```suggestion
@Suppress("ComposeUnstableCollections")
```

No need for a comment in code for that. For now indeed we don't have stable collections and we don't want to add them yet since the API is not yet stable.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
+            )
+        },
+    ) { padding ->
+        Column(
+            modifier = Modifier
+                .verticalScroll(rememberScrollState())
+                .windowInsetsPadding(safeBottomWindowInsets())
+                .padding(padding)
+                .padding(all = 16.dp),
+            verticalArrangement = Arrangement.spacedBy(8.dp),
```
TimoPtr: Use `HADimens` for all the padding and space in the composable.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
+    Scaffold(
+        topBar = {
+            TopAppBar(
+                title = { Text(stringResource(commonR.string.create_template)) },
+                windowInsets = safeTopWindowInsets(),
+                backgroundColor = colorResource(commonR.color.colorBackground),
+                contentColor = colorResource(commonR.color.colorOnBackground),
+            )
+        },
+    ) { padding ->
```
TimoPtr: Let's breakdown the whole Content of the scaffold into smaller function. A good indicator is that the function should fit entirely within your screen.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
+            } else if (templateText.isEmpty()) {
+                Text(
+                    text = stringResource(commonR.string.empty_template),
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .padding(vertical = 4.dp),
+                )
+            }
+
+            HATextField(
```
TimoPtr: If you start using HATextField then you are mixing M2 and M3, seeing this screen I think you can already use M3 directly and `HATheme`.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
+                    keys = listOf(
+                        stringResource(commonR.string.widget_text_color_black),
+                        stringResource(commonR.string.widget_text_color_white),
+                    ),
+                    currentIndex = textColorIndex,
+                    onSelected = { onTextColorSelected(it) },
+                )
+            }
+
+            Button(
```
TimoPtr: Use HAButton

TimoPtr: If you need reference on how to do things with these new style check any Compose screen within the onboarding/ package.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+import timber.log.Timber
+
+@HiltViewModel
+class TemplateWidgetConfigureViewModel @Inject constructor(
+    private val templateWidgetDao: TemplateWidgetDao,
+    private val serverManager: ServerManager,
+) : ViewModel() {
+
+    @VisibleForTesting
+    internal var ioDispatcher: CoroutineDispatcher = Dispatchers.IO
```
TimoPtr: You shouldn't need that.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+     */
+    suspend fun requestWidgetCreation() {
+        val pendingEntity = getPendingDaoEntity()
+        templateWidgetDao.getWidgetCountFlow().drop(1).onStart {
+            _action.emit(Action.RequestWidgetCreationAction(pendingEntity))
+        }.first()
+    }
+}
+
+private const val RENDER_DEBOUNCE_MS = 500L
```
TimoPtr: Use Kotlin Duration API.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+        if (template.isEmpty()) {
+            renderedTemplate = null
+            isTemplateValid = false
+            return
+        }
+        if (!serverManager.isRegistered() || serverManager.getServer(serverId) == null) {
+            Timber.w("Not rendering template because server is not set")
+            return
+        }
+        withContext(ioDispatcher) {
```
TimoPtr: You don't need IO here, integrationRepository is main safe meaning it does already use its own dispatcher to do the network call. Then you might want to use Default instead of Main just because you are parsing some Html and if the template is huge we don't want to do that on Main. We only use IO for calls that depends on network or disk reads, otherwise we stick to default to avoid creating too much threads since IO has an elastic pool.

```suggestion
        withContext(Dispatchers.Default) {
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+                val result = serverManager.integrationRepository(serverId)
+                    .renderTemplate(template, mapOf())
+                    .toString()
+                @Suppress("UNNECESSARY_SAFE_CALL")
+                renderedTemplate = HtmlCompat.fromHtml(result, HtmlCompat.FROM_HTML_MODE_LEGACY)
+                    ?.toString()
+                    ?.trimEnd()
+                    ?: result
+                isTemplateValid = true
+            } catch (e: Exception) {
```
TimoPtr, line 191: rethrow CancelationException here.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+            _action.emit(Action.RequestWidgetCreationAction(pendingEntity))
+        }.first()
+    }
+}
+
+private const val RENDER_DEBOUNCE_MS = 500L
+private const val DEFAULT_TEXT_SIZE = 12.0f
+
+sealed class Action {
+    object UpdateWidgetAction : Action()
```
TimoPtr:
```suggestion
    data object UpdateWidgetAction : Action
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+        templateWidgetDao.getWidgetCountFlow().drop(1).onStart {
+            _action.emit(Action.RequestWidgetCreationAction(pendingEntity))
+        }.first()
+    }
+}
+
+private const val RENDER_DEBOUNCE_MS = 500L
+private const val DEFAULT_TEXT_SIZE = 12.0f
+
+sealed class Action {
```
TimoPtr:
```suggestion
sealed interface TemplateWidgetAction. {
```

If you keep the sealed class the name should be quite explicit because Action is too wide and think about the code completion in other part of the app. If you keep Action and add it to all the widget configure page it would be messy.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+    object UpdateWidgetAction : Action()
+    data class RequestWidgetCreationAction(val pendingEntity: TemplateWidgetEntity) : Action()
```
TimoPtr:
```suggestion
    data class RequestWidgetCreationAction(val pendingEntity: TemplateWidgetEntity) : Action
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+     * it does not cause issues in the current implementation as returning multiple times has no adverse effects.
+     */
+    suspend fun requestWidgetCreation() {
+        val pendingEntity = getPendingDaoEntity()
+        templateWidgetDao.getWidgetCountFlow().drop(1).onStart {
+            _action.emit(Action.RequestWidgetCreationAction(pendingEntity))
+        }.first()
+    }
+}
+
```
TimoPtr, line 285: We usually put Actions and const at the top of the file.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+            }
+        }
+    }
+
+    /**
+     * Initialize the ViewModel with the widget ID and supported text colors.
+     * Loads existing widget configuration if editing an existing widget.
+     */
+    fun onSetup(widgetId: Int, supportedTextColors: List<String>) {
+        if (this.widgetId != AppWidgetManager.INVALID_APPWIDGET_ID) return
```
TimoPtr: In which scenario this could happen?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+        private set
+
+    var renderedTemplate by mutableStateOf<String?>(null)
+        private set
+    var isTemplateValid by mutableStateOf(false)
+        private set
+
+    @OptIn(FlowPreview::class)
+    private fun startTemplateRendering() {
+        viewModelScope.launch {
```
TimoPtr: Since this function is invoked from a public function I would make sure that if we call it twice we only keep one Job. I would keep a reference to the started coroutine and cancel it if restarted.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+     * @throws IllegalStateException if the widget ID is invalid or server is not valid
+     */
+    suspend fun updateWidgetConfiguration() {
+        if (widgetId == AppWidgetManager.INVALID_APPWIDGET_ID) {
+            throw IllegalStateException("Widget ID is invalid")
+        }
+
+        val entity = getPendingDaoEntity()
+        templateWidgetDao.add(entity)
+        _action.emit(Action.UpdateWidgetAction)
```
TimoPtr: It's not ideal but you can do like on the TodoWidget and take a context here to call the AppWidgetManager from the ViewModel. I'm not a fan but it's much reliable and avoid the complex potential race condition.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+            template = templateText,
+            textSize = textSize.toFloatOrNull() ?: DEFAULT_TEXT_SIZE,
+            lastUpdate = "Loading",
+            backgroundType = selectedBackgroundType,
+            textColor = textColor,
+        )
+    }
+
+    /**
+     * Save the widget configuration to the database and send a broadcast to update the widget.
```
TimoPtr: It doesn't send a broadcast.

TimoPtr (commented):
Thanks for your work 🎖️, some comments but overall it looks ok.
Would you mind adding a screenshot test too to your compose screen, it should be rather easy?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,205 @@
...
+    private val _action = MutableSharedFlow<Action>()
+    val action = _action.asSharedFlow()
+
+    val servers = serverManager.serversFlow
+
+    var selectedServerId by mutableIntStateOf(ServerManager.SERVER_ID_ACTIVE)
+        private set
+
+    var templateText by mutableStateOf("")
+    var textSize by mutableStateOf("14")
```
TimoPtr: I know we are using State in other viewModel but if you feel advanturous I would like to avoid using any Composable specifics inside a viewModel. I would be in favor of exposing a flow instead. You can go with an MVI approach if you like with one object representing the ViewState and only exposing one flow to the view that contains everything the view need to render.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
-                    enabled = true
-                } catch (e: Exception) {
-                    Timber.e(e, "Exception while rendering template")
-                    // SerializationException suggests that template is not a String (= error)
-                    templateText = getString(
-                        if (e.cause is SerializationException) {
-                            commonR.string.template_error
-                        } else {
-                            commonR.string.template_render_error
-                        },
```
TimoPtr, line 234: I think it is a good idea to keep this logic that was giving a bit more details to the user making the template on what is the current error.

amlwin:
# Changes Summary                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                         
 ## 1. TemplateWidgetConfigureViewModel.kt - Major refactor:                                                                                                                                                                                                                 
  - MVI with StateFlow: Replaced all mutableStateOf properties with a single TemplateWidgetConfigureUiState data class exposed via StateFlow (no more Compose-specific state in ViewModel)                                                                                 
  - Constants at top: Moved RENDER_DEBOUNCE_MS and DEFAULT_TEXT_SIZE to the top of the file
  - Job reference for template rendering: startTemplateRendering() now stores a Job reference and cancels any previous one before restarting                                                                                                                               
  - Context passed to widget operations: updateWidgetConfiguration(context) and requestWidgetCreation(context) now take a Context parameter and handle AppWidgetManager/broadcast directly - eliminating the Action sealed class and the race condition with finish()
  - KDoc fixed: Updated doc for updateWidgetConfiguration to accurately describe what it does
  - SerializationException distinction preserved: Added TemplateRenderError enum to differentiate between template syntax errors and render errors, displayed with appropriate string resources
  - onSetup guard documented: Added comment explaining the guard prevents re-initialization on configuration changes

 ## 2. TemplateWidgetConfigureScreen.kt - New file:
  - Extracted all Composable functions from the Activity into a dedicated file
  - Updated to consume uiState StateFlow from ViewModel
  - Added template error display using TemplateRenderError to show specific error messages (template_error vs template_render_error)

##  3. TemplateWidgetConfigureActivity.kt - Simplified:
  - Removed enableEdgeToEdgeCompat() (already in BaseActivity)
  - Removed all Compose UI code (moved to new file)
  - Removed Action observation/handling
  - Removed @RequiresApi(Build.VERSION_CODES.O) annotations (SDK_INT always >= 26)
  - Passes context directly to ViewModel methods, eliminating the race condition

##  4. TemplateWidgetConfigureViewModelTest.kt - Updated:
  - All tests access state via viewModel.uiState.value instead of direct properties
  - State mutations use ViewModel methods (onTemplateTextChanged, onTextSizeChanged, etc.) instead of direct property assignment
  - updateWidgetConfiguration tests pass a mock Context
  - Added assertion for TemplateRenderError.RENDER_ERROR in error test

cc: @TimoPtr

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
             }
 
-            override fun onNothingSelected(parent: AdapterView<*>?) {
-                binding.textColor.visibility = View.GONE
-            }
+            Action.UpdateWidgetAction -> handleUpdateWidgetAction()
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
```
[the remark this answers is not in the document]
amlwin: updated

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,350 @@
...
-                    } else {
-                        commonR.color.colorWidgetButtonLabelBlack
-                    },
-                )
-            } else {
-                null
-            },
-            lastUpdate = dao.get(appWidgetId)?.lastUpdate ?: "Loading",
-        )
+    @RequiresApi(Build.VERSION_CODES.O)
```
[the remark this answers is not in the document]
amlwin: updated

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
amlwin: updated

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
amlwin: updated

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
amlwin: update. restore the previous `context` pass to viewModel style. thanks

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
amlwin: thanks for the catch

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
amlwin: updated

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
amlwin: update. restore the previous context pass to viewModel style. thanks

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
amlwin: updated

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: updated

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: updated

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: as we are sticky to previous flow (to prevent race condition) action is no longer required here.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: as we are sticky to previous flow (to prevent race condition) action is no longer required here.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: as we are sticky to previous flow (to prevent race condition) action is no longer required here.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: as we are sticky to previous flow (to prevent race condition) action is no longer required here.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: updated

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: i'm going to remove that line coz view (activity) already have these check.
```
if (widgetId == AppWidgetManager.INVALID_APPWIDGET_ID && !requestLauncherSetup) {
            finish()
            return
        }
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: update. and made it `private`. and added job cancel procedure

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: updated

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
amlwin: updated

TimoPtr:
@amlwin reopen the PR when you think it is ready for review.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: updated

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: yes! u r right.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
amlwin: changed it to default

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
amlwin: can elaborate more on this ? why am i mixing M2 and M3 component ? if so which one should i use `HATextField` or `M3TextField`

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
amlwin: updated

amlwin:
> Thanks for your work 🎖️, some comments but overall it looks ok. Would you mind adding a screenshot test too to your compose screen, it should be rather easy?

look like screenshot is not working properly with current gradle version
 The comment in libs.versions.toml even hints at this: `On the next update, replace the hack in AndroidComposeConvertionPlugin with the real fix`


Screenshot 2026-02-18 at 10 59 29 PM

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,97 @@
...
-            return
+    private suspend fun onUpdateWidget() {
+        try {
+            viewModel.updateWidgetConfiguration(this@TemplateWidgetConfigureActivity)
+            setResult(RESULT_OK)
+            finish()
+        } catch (e: IllegalStateException) {
+            if (e is CancellationException) throw e
+            showAddWidgetError()
         }
```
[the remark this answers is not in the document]
TimoPtr: Remove the if no need if you only catch IllegalStateException. The cancelation is already thrown

TimoPtr:
> > Thanks for your work 🎖️, some comments but overall it looks ok. Would you mind adding a screenshot test too to your compose screen, it should be rather easy?
> 
> look like screenshot is not working properly with current gradle version The comment in libs.versions.toml even hints at this: `On the next update, replace the hack in AndroidComposeConvertionPlugin with the real fix`
> Screenshot 2026-02-18 at 10 59 29 PM

It never really worked in the IDE but it works when you execute the update screenshot task.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,294 @@
...
+                null,
+                PendingIntent.getBroadcast(
+                    context,
+                    System.currentTimeMillis().toInt(),
+                    Intent(context, TemplateWidget::class.java).apply {
+                        action = ACTION_APPWIDGET_CREATED
+                        putExtra(EXTRA_WIDGET_ENTITY, pendingEntity)
+                    },
+                    flags,
+                ),
```
TimoPtr, line 281:
```suggestion
            appWidgetManager?.requestPinAppWidget(
                ComponentName(context, TemplateWidget::class.java),
                null,
                PendingIntent.getBroadcast(
                    context,
                    System.currentTimeMillis().toInt(),
                    Intent(context, TemplateWidget::class.java).apply {
                        action = ACTION_APPWIDGET_CREATED
                        putExtra(EXTRA_WIDGET_ENTITY, pendingEntity)
                    },
                    PendingIntent.FLAG_MUTABLE,
                ),
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,97 @@
...
-            renderTemplateText(editableText.toString())
-        } else {
-            binding.renderedTemplate.text = getString(commonR.string.empty_template)
-            binding.addButton.isEnabled = false
+    private suspend fun requestPinWidget() {
+        try {
+            viewModel.requestWidgetCreation(this@TemplateWidgetConfigureActivity)
+            finish()
+        } catch (e: IllegalStateException) {
+            if (e is CancellationException) throw e
```
TimoPtr, line 78:
```suggestion
        } catch (e: CancellationException) {
            throw e
        } catch (e: IllegalStateException) {
```
This is way we use in the repo, please adjust in the activity and the viewModel

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,97 @@
...
+        lifecycleScope.launch {
             if (requestLauncherSetup) {
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                    lifecycleScope.launch {
-                        requestWidgetCreation()
-                    }
-                } else {
-                    showAddWidgetError() // this shouldn't be possible
-                }
+                requestPinWidget()
```
TimoPtr, line 66: Why not checking the version anymore? If you do so it would remove the warning in the ViewModel by properly annotating the function `requestPinWidget` like it is done in TodoWidgetConfigure.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,97 @@
...
-        val serverId = selectedServerId
-        if (serverId == null) {
-            Timber.w("Not rendering template because server is not set")
-            return
+    private suspend fun onUpdateWidget() {
+        try {
+            viewModel.updateWidgetConfiguration(this@TemplateWidgetConfigureActivity)
+            setResult(RESULT_OK)
+            finish()
+        } catch (e: IllegalStateException) {
```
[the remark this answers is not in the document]
TimoPtr: Good catch 👍🏻, I'm not sure it is needed in the TodoWidgetConfigure because we use Glance and we manually call the update method from the manager.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
```diff
@@ -1,244 +1,97 @@
...
-                    )
-                    enabled = false
-                }
-            }
-            binding.renderedTemplate.text =
-                templateText?.let { HtmlCompat.fromHtml(it, HtmlCompat.FROM_HTML_MODE_LEGACY) }
-            binding.addButton.isEnabled = enabled && isValidServerId()
-        }
+    private fun showAddWidgetError() {
+        viewModel.showError(commonR.string.widget_creation_error)
```
TimoPtr, line 95: The error is always the same and always trigger from the activity I propose you to not use the viewModel and directly send this to the screen, it would simplify things.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,285 @@
...
+        textColorIndex = uiState.textColorIndex,
+        onTextColorSelected = viewModel::onTextColorSelected,
+        isUpdateWidget = uiState.isUpdateWidget,
+        onActionClick = onActionClick,
+        snackbarHostState = snackbarHostState,
+    )
+}
+
+@Composable
+private fun TemplateWidgetConfigureView(
```
TimoPtr, line 81:
```suggestion
@VisibleForTesting
internal fun TemplateWidgetConfigureScreen(
```

Ideally we should write a compose screen test https://github.com/home-assistant/android/blob/457de6f8896328e6ebebeb5f3db8a01a39e7168f/app/src/test/kotlin/io/homeassistant/companion/android/onboarding/sethomenetwork/SetHomeNetworkScreenTest.kt for inspiration

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,285 @@
...
+        topBar = {
+            TopAppBar(
+                title = { Text(stringResource(commonR.string.create_template)) },
+                windowInsets = safeTopWindowInsets(),
+                backgroundColor = colorResource(commonR.color.colorBackground),
+                contentColor = colorResource(commonR.color.colorOnBackground),
+            )
+        },
+    ) { padding ->
+        Column(
```
TimoPtr, line 111: Extract the whole Column into a `TemplateWidgetConfigureContent`

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
TimoPtr: HATextField is a wrapper on top of M3 and use the new theme `HATheme`, if you use `HATheme` instead of `HomeAssistantAppTheme` most of the color would already be set. But you need to stop using M2 for the Scaffold and the rest.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,285 @@
...
+) {
+    Scaffold(
+        snackbarHost = { SnackbarHost(snackbarHostState) },
+        topBar = {
+            TopAppBar(
+                title = { Text(stringResource(commonR.string.create_template)) },
+                windowInsets = safeTopWindowInsets(),
+                backgroundColor = colorResource(commonR.color.colorBackground),
+                contentColor = colorResource(commonR.color.colorOnBackground),
+            )
```
TimoPtr, line 108: Can be migrated to HATopBar, then you don't need to specify the color.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureActivity.kt
TimoPtr: Check any of the compose screen in onboarding/ package for inspiration.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,285 @@
...
+                .fillMaxWidth()
+                .padding(vertical = HADimens.SPACE1),
+        )
+    } else if (templateText.isEmpty()) {
+        Text(
+            text = stringResource(commonR.string.empty_template),
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(vertical = HADimens.SPACE1),
+        )
```
TimoPtr, line 222:
```suggestion
    val text = when {
        renderedTemplate != null -> renderedTemplate
        templateRenderError != null -> stringResource(
            when (templateRenderError) {
                TemplateRenderError.TEMPLATE_ERROR -> commonR.string.template_error
                TemplateRenderError.RENDER_ERROR -> commonR.string.template_render_error
            },
        )
        templateText.isEmpty() -> stringResource(commonR.string.empty_template)
        else -> null
    }
    if (text != null) {
        Text(
            text = text,
            modifier = Modifier
                .fillMaxWidth()
                .padding(vertical = HADimens.SPACE1),
        )
    }
```

A bit shorter and avoid repetition of the Text composable attributes that would lead to discrepancy later.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,285 @@
...
+            currentIndex = textColorIndex,
+            onSelected = { onTextColorSelected(it) },
+        )
+    }
+}
+
+@Preview
+@Composable
+private fun TemplateWidgetConfigureViewPreview() {
+    HomeAssistantAppTheme {
```
TimoPtr, line 265:
```suggestion
    HAThemeForPreview {
```

When you use `HATheme`.

For the screenshots let's make a screenshot for 
- empty state
- multi server available
- error in template
- valid template
- tranparent background selected

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureUiState.kt
```diff
@@ -0,0 +1,36 @@
+package io.homeassistant.companion.android.widgets.template
+
+import com.google.android.material.color.DynamicColors
+import io.homeassistant.companion.android.common.data.servers.ServerManager
+import io.homeassistant.companion.android.database.widget.WidgetBackgroundType
+
+/**
+ * Represents the UI state of the Template Widget configuration screen.
+ */
+data class TemplateWidgetConfigureUiState(
```
TimoPtr, line 10:
```suggestion
internal data class TemplateWidgetConfigureUiState(
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,294 @@
...
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onStart
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import kotlinx.serialization.SerializationException
+import timber.log.Timber
+
+private const val RENDER_DEBOUNCE_MS = 500
```
TimoPtr, line 48:
```suggestion
private val RENDER_DEBOUNCE = 500.milliseconds
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,294 @@
...
+    val servers = serverManager.serversFlow
+
+    private var templateRenderingJob: Job? = null
+
+    /**
+     * Initialize the ViewModel with the widget ID and supported text colors.
+     * Loads existing widget configuration if editing an existing widget.
+     *
+     * This guard prevents re-initialization when the Activity is recreated due to
+     * configuration changes, since the ViewModel survives those changes.
```
TimoPtr, line 77: What do you mean here? This is going to be call again if the activity configuration change for instance. If your objective is really to guard this behavior you should instead initialize the viewModel with these value and remove this function completely.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureUiState.kt
```diff
@@ -0,0 +1,36 @@
...
+    val isTemplateValid: Boolean = false,
+    val textSize: String = "14",
+    val selectedBackgroundType: WidgetBackgroundType =
+        if (DynamicColors.isDynamicColorAvailable()) {
+            WidgetBackgroundType.DYNAMICCOLOR
+        } else {
+            WidgetBackgroundType.DAYNIGHT
+        },
+    val textColorIndex: Int = 0,
+    val isUpdateWidget: Boolean = false,
```
TimoPtr, line 23: The name of this attribute is a bit confusing. Maybe we could reverse the logic to `isNewWidget`.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,294 @@
...
+                        },
+                    )
+                }
+            }
+        }
+    }
+
+    /**
+     * Update the selected server.
+     */
```
TimoPtr, line 110: No need for this documentation when the function is that simple and obvious.
This comments apply for all the function bellow.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,294 @@
...
+        templateRenderingJob?.cancel()
+        templateRenderingJob = viewModelScope.launch {
+            combine(
+                _uiState.mapField { it.templateText },
+                _uiState.mapField { it.selectedServerId },
+            ) { template, serverId -> template to serverId }
+                .debounce(RENDER_DEBOUNCE_MS.milliseconds)
+                .collect { (template, serverId) ->
+                    renderTemplate(template, serverId)
+                }
```
TimoPtr, line 156:
```suggestion
          uiState.map { it.templateText to it.selectedServerId }
                .distinctUntilChanged()
                .debounce(RENDER_DEBOUNCE_MS.milliseconds)
                .collect { (template, serverId) ->
                    renderTemplate(template, serverId)
                }
```
A bit simpler to read  and more 'functional' oriented IMO.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    /**
+     * Extracts a distinct field from the StateFlow to avoid unnecessary emissions.
+     */
+    private fun <T> StateFlow<TemplateWidgetConfigureUiState>.mapField(
+        selector: (TemplateWidgetConfigureUiState) -> T,
+    ): Flow<T> {
+        return map(selector).distinctUntilChanged()
+    }
```
TimoPtr, line 167:
```suggestion
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,294 @@
...
+                if (e is SerializationException) {
+                    Timber.e(e, "Template syntax error")
+                } else {
+                    Timber.e(e, "Error rendering template")
+                }
+                val errorRendered = if (e.cause is SerializationException) {
+                    TemplateRenderError.TEMPLATE_ERROR
+                } else {
+                    TemplateRenderError.RENDER_ERROR
+                }
```
TimoPtr, line 204:
```suggestion
                Timber.e(e, "Error rendering template")
                val errorRendered = if (e.cause is SerializationException) {
                    TemplateRenderError.TEMPLATE_ERROR
                } else {
                    TemplateRenderError.RENDER_ERROR
                }
```

`e` already contains the error no need to go further than this.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,285 @@
...
+                .padding(vertical = HADimens.SPACE1),
+        )
+    } else if (templateText.isEmpty()) {
+        Text(
+            text = stringResource(commonR.string.empty_template),
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(vertical = HADimens.SPACE1),
+        )
+    }
```
TimoPtr, line 223: You could use the `HATextField.isError` and `HATextField.supportingText` to display the error instead of the Text. It would change the color of the TextField to red on error.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,294 @@
...
+        return map(selector).distinctUntilChanged()
+    }
+
+    private suspend fun renderTemplate(template: String, serverId: Int) {
+        if (template.isEmpty()) {
+            _uiState.update { it.copy(renderedTemplate = null, isTemplateValid = false) }
+            return
+        }
+        if (!serverManager.isRegistered() || serverManager.getServer(serverId) == null) {
+            Timber.w("Not rendering template because server is not set")
```
TimoPtr, line 175: Maybe you could set an Error here in the ViewState otherwise the user wouldn't see anything.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,294 @@
...
+        withContext(Dispatchers.Default) {
+            try {
+                val result = serverManager.integrationRepository(serverId)
+                    .renderTemplate(template, mapOf())
+                    .toString()
+                @Suppress("UNNECESSARY_SAFE_CALL")
+                val rendered = HtmlCompat.fromHtml(result, HtmlCompat.FROM_HTML_MODE_LEGACY)
+                    ?.toString()
+                    ?.trimEnd()
+                    ?: result
```
TimoPtr, line 187:
```suggestion
                val rendered = HtmlCompat.fromHtml(result, HtmlCompat.FROM_HTML_MODE_LEGACY).toString().trimEnd()
```

If you check the doc from `fromHtml` it won't ever return null. Any reason to do otherwise?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,294 @@
...
+                _uiState.update {
+                    it.copy(renderedTemplate = rendered, isTemplateValid = true, templateRenderError = null)
+                }
+            } catch (e: Exception) {
+                if (e is CancellationException) {
+                    throw e
+                }
+                if (e is SerializationException) {
```
TimoPtr, line 195: Put this in a dedicated catch block above this catch.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,294 @@
...
+            supportedTextColors.getOrNull(state.textColorIndex) ?: supportedTextColors.first()
+        } else {
+            null
+        }
+        return TemplateWidgetEntity(
+            id = widgetId,
+            serverId = state.selectedServerId,
+            template = state.templateText,
+            textSize = state.textSize.toFloatOrNull() ?: DEFAULT_TEXT_SIZE,
+            lastUpdate = "Loading",
```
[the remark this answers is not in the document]
TimoPtr: This comment is kinda weird because the previous code was `dao.get(appWidgetId)?.lastUpdate ?: "Loading",` please check

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    /**
+     * Emit a one-shot error event to be displayed as a Snackbar.
+     *
+     * @param messageResId the string resource ID for the error message
+     */
+    fun showError(messageResId: Int) {
+        viewModelScope.launch { _errorMessage.emit(messageResId) }
+    }
```
TimoPtr, line 293:
```suggestion
```

TimoPtr (changes requested):
If you are using AI I think that for this final step you should do it yourself to make sure to fully get my comments.

amlwin:
> If you are using AI I think that for this final step you should do it yourself to make sure to fully get my comments.

I'll check in this weekend

TimoPtr:
> > > Thanks for your work 🎖️, some comments but overall it looks ok. Would you mind adding a screenshot test too to your compose screen, it should be rather easy?
> > 
> > 
> > look like screenshot is not working properly with current gradle version The comment in libs.versions.toml even hints at this: `On the next update, replace the hack in AndroidComposeConvertionPlugin with the real fix`
> > Screenshot 2026-02-18 at 10 59 29 PM
> 
> It never really worked in the IDE but it works when you execute the update screenshot task.

@amlwin If you merge main in your branch and update your Android Studio to the latest version then it is going to work. It has been fixed this week 🎉

TimoPtr:
We've recently merged https://github.com/home-assistant/android/issues/6307 and can be used as a reference design to implement the refactor.

jpelgrom:
Competing PR which is much closer: #7288

PR #6446, opened 2026-02-16
Avoid persist-credentials whenever it is not needed and disable ruby cache

TimoPtr:
## Summary

`persist-credentials ` can be avoid in most of the case in our workflows and should be it avoid storing the token within the runner filesystem. 
Disable ruby cache to avoid cache polluting from another job. The time saved by the cache is small that it doesn't bring value to us compare to the risk of a bad cache.

Following @frenck work on https://github.com/home-assistant/core/pull/162991

### DIFF
--- a/.github/workflows/monthly.yml
+++ b/.github/workflows/monthly.yml
@@ -23,6 +23,7 @@ jobs:
       - name: Bump Calendar Version
         uses: rickstaa/action-create-tag@a1c7777fcb2fee4f19b0f283ba888afa11678b72 # v1.7.2
         with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
           tag: ${{ steps.current-time.outputs.year }}.${{ steps.current-time.outputs.month }}.0
 
       - name: Trigger Prepare Next Release


.github/workflows/weekly.yml
```diff
@@ -15,14 +15,15 @@ jobs:
...
           fetch-depth: 0
+          persist-credentials: false
 
       - uses: ./.github/actions/setup-build-env
         with:
           cache-encryption-key: ${{ secrets.GRADLE_ENCRYPTION_KEY }}
 
       - name: Tag Weekly Release
         env:
-          GITHUB_TOKEN: ${{ secrets.TAG_PUSH_TOKEN }}
```
TimoPtr, line 25: 👋🏻 @JBassett were you the one adding this secret in the first place?

TimoPtr: This is going to be replace by an API call to trigger the workflow associated with tag instead of using a token that is not rotated.

TimoPtr: Actually it is already done bellow it triggers onPush

jpelgrom:
More details from the tool: https://docs.zizmor.sh/audits/#artipacked

.github/workflows/monthly.yml
```diff
@@ -13,6 +13,8 @@ jobs:
       contents: write
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
```
jpelgrom, line 17: After checking the code for `rickstaa/action-create-tag`, I think that this needs to be passed the token or persistence of the token as it does git operations. The action's README mentions a behavior change in `actions/checkout@v2` which is when the token persistence was introduced: https://github.com/rickstaa/action-create-tag/tree/74a6323aaa8841b5769818fb633f79ca3f18ceee?tab=readme-ov-file#github_token.

TimoPtr: Good catch I missed it 👍🏻, I went the other way and give the secret to the action. Whenever we can I prefer to not store the secret on the runner filesystem, even if we don't control if the action itself is not writing it into a file.

PR #6447, opened 2026-02-16
Preserve WebViewActivity path when switching URLs

### LINKED ISSUE #4983: Load last showed dashboard view after changing connection (internal/external)
Assuming there's a internal and external URL defined. I also have multiple dashboards and views inside the dasboards.
Current situation:
I'm connected to the internal url and I'm viewing the dashboard "home" and the view "light" (there are multiple views available). Then I'm switching to the external URL. The companion app is connecting to the external URL as supposed, but the frontend reloads the dashboard and throws me back to the first view in the dashboard.

- When changing to/from the internal/external URL, the frontend still loads the last viewed view (in this example "light")
- It would be awesome if also the vertical scroll location inside the last view is restored

Manually move to the last location/view.

**Additional context**

Gifford47:
## Summary
This PR refactors how paths are handled in the WebView to ensure navigation **keeps the current path** and exposes it reliably.

###  Key Changes
- **Path handling refactor** in `WebViewPresenterImpl.kt`, `WebViewActivity.kt`, `WebView.kt`.
- **New method:** `getCurrentPath()` → retrieves the active WebView path.
- Navigation now preserves paths during internal WebView transitions.

###  Why
- Fixes inconsistent path retrieval during navigation and deep link handling.

###  Impact
- WebView navigation now maintains the correct path.
- Activity/Presenter logic decoupled and cleaner.
- `getCurrentPath()` can be used wherever current WebView state is needed.

## Checklist

- [X] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).

## Any other notes

See issue #4983


Gifford47:
@TimoPtr can you please have a look on it? you have additional webview PRs ... maybe there's conflict between the PRs ...

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1445,12 +1441,28 @@ class WebViewActivity :
...
                 }
-                intent.removeExtra(EXTRA_PATH)
                 presenter.load(lifecycle, path, isInternalOverride)
+                intent.removeExtra(EXTRA_PATH)
+
+                if (presenter.isFullScreen() || isVideoFullScreen) {
+                    hideSystemUI()
+                } else {
+                    showSystemUI()
+                }
```
TimoPtr: Why moving this?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1423,19 +1423,15 @@ class WebViewActivity :
...
-                }
-
-                var path = intent.getStringExtra(EXTRA_PATH)
-                if (path?.startsWith("entityId:") == true) {
-                    // Get the entity ID from a string formatted "entityId:domain.entity"
-                    // https://github.com/home-assistant/core/blob/dev/homeassistant/core.py#L159
+                val intentPath = intent.getStringExtra(EXTRA_PATH)
+                // When no explicit navigation path is set (e.g. from a notification),
+                // preserve the current WebView path so the user's dashboard view
+                // survives internal/external URL switches. See #4983.
```
TimoPtr: Add a URL not just the issue number 🙏🏻

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1423,19 +1423,15 @@ class WebViewActivity :
...
+                var path = intentPath ?: getCurrentWebViewPath()
+                if (intentPath?.startsWith("entityId:") == true) {
+                    moreInfoEntity = intentPath.substringAfter("entityId:")
```
TimoPtr: Why setting the moreInfoEntity here? It changes the behavior that was implemented before.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1423,19 +1423,15 @@ class WebViewActivity :
...
-
-                var path = intent.getStringExtra(EXTRA_PATH)
-                if (path?.startsWith("entityId:") == true) {
-                    // Get the entity ID from a string formatted "entityId:domain.entity"
-                    // https://github.com/home-assistant/core/blob/dev/homeassistant/core.py#L159
+                val intentPath = intent.getStringExtra(EXTRA_PATH)
+                // When no explicit navigation path is set (e.g. from a notification),
+                // preserve the current WebView path so the user's dashboard view
+                // survives internal/external URL switches. See #4983.
+                var path = intentPath ?: getCurrentWebViewPath()
```
TimoPtr: Why getting the current path here and in the presenter? Since you are calling `load` of the presenter, the presenter would already try to get the current path from the Webview.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1445,12 +1441,28 @@ class WebViewActivity :
                         moreInfoEntity = entity
                     }
                 }
-                intent.removeExtra(EXTRA_PATH)
                 presenter.load(lifecycle, path, isInternalOverride)
+                intent.removeExtra(EXTRA_PATH)
```
TimoPtr, line 1602: You should not more this line to, keep it before calling load

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1445,12 +1441,28 @@ class WebViewActivity :
...
             }
         }
     }
 
+    override fun getCurrentWebViewPath(): String? {
+        val currentUrl = webView.url ?: return null
+        return try {
+            val path = currentUrl.toUri().path
+            if (path.isNullOrEmpty() || path == "/") null else path
+        } catch (e: Exception) {
```
TimoPtr: Let's not catch all exception here only the relevant one that can be thrown by `toUri`.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -135,13 +135,18 @@ class WebViewPresenterImpl @Inject constructor(
         }
 
         serverManager.connectionStateProvider(serverId).urlFlow(isInternalOverride).collect { urlState ->
```
TimoPtr, line 139: I would like to see unit tests for this logic

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1445,12 +1441,28 @@ class WebViewActivity :
...
+                }
             }
         }
     }
 
+    override fun getCurrentWebViewPath(): String? {
+        val currentUrl = webView.url ?: return null
+        return try {
+            val path = currentUrl.toUri().path
+            if (path.isNullOrEmpty() || path == "/") null else path
```
TimoPtr: I think
`webView.url?.toUri()?.path?.takeIf { it.length > 1 }`

would do the trick here

TimoPtr (commented):
Thanks for your work, did you test it in multiple conditions? Like changing quickly the URL multiple times?

Also yes I'm currently working on making a new version of this WebViewActivity, it is going to take some time but you can already look at how it looks https://github.com/home-assistant/android/pull/6386

TimoPtr:
Also check what is happening when having arguments in the url like some filtered on the history page.

Gifford47:
Thanks for the thorough review @TimoPtr! I've addressed your feedback:

Duplicate path retrieval: Removed getCurrentWebViewPath() from the Activity — the Presenter already handles this fallback in collectUrlStateChanges(), so the Activity now only passes the intentPath (or null).

intent.removeExtra(EXTRA_PATH): Moved before presenter.load() so the extra is consumed immediately.

moreInfoEntity assignment: Fixed the behavioral change — removed the unconditional raw assignment. Now only sets moreInfoEntity to the regex-validated entity for older servers (JS dispatch path), and explicitly clears it when using the URL path approach (>= 2025.6).

Exception handling in getCurrentWebViewPath(): Removed the broad catch (Exception) — Uri.parse() is lenient and doesn't throw, so the try-catch was unnecessary.

Path extraction simplification: Replaced with webView.url?.toUri()?.path?.takeIf { it.length > 1 } as suggested.

Comment reference: Updated to use the full GitHub issue URL.

Regarding URL arguments (e.g. filters on the history page): Uri.path only returns the path component and excludes query parameters, so filtered views like /history?entity_id=sensor.foo will correctly preserve just /history as the path — the frontend will reload with its default state for that view, which is the expected behavior when switching between internal/external URLs.

TimoPtr:
> Regarding URL arguments (e.g. filters on the history page): Uri.path only returns the path component and excludes query parameters, so filtered views like /history?entity_id=sensor.foo will correctly preserve just /history as the path — the frontend will reload with its default state for that view, which is the expected behavior when switching between internal/external URLs.

Actually it might be nice to keep the whole URI and just change the host/port imagine you are on 
`http://192.168.15.6:8123/history?start_date=2026-02-19T11%3A00%3A00.000Z&end_date=2026-02-19T14%3A00%3A00.000Z`

It would be nice to only change the host/port `192.168.15.6:8123` to the other URL to keep exactly where you are.

TimoPtr:
The issue would be that most probably you are going to loose the history and if it is not the case I wonder what happens. Did you try to play with the app to see how it behaves for the history?

Gifford47:
Changes

1. Preserve full relative URL on network switches (WebViewPresenterImpl.kt, WebView.kt, WebViewActivity.kt)

Added getCurrentWebViewRelativeUrl() which extracts the full relative URL (path + query parameters + fragment) from the current WebView URL, stripping the external_auth parameter since the presenter re-adds it on every load.
In collectUrlStateChanges, subsequent urlFlow emissions now read the current WebView URL instead of only using the initial path parameter. This ensures the user stays on the exact same page (e.g. /history?entity_id=...&start_date=...) after a network switch.

2. Clear WebView history on base URL changes (WebViewPresenterImpl.kt)

Added lastBaseUrl tracking in collectUrlStateChanges to detect when the base URL changes (internal ↔ external).
When a base URL change is detected, isNewServer is set to true, which triggers keepHistory = false and clears the WebView's navigation history. This prevents the back button from navigating to unreachable URLs from the old connection.

3. Safe back navigation after URL switches (WebViewActivity.kt)

After history is cleared, rapid urlFlow emissions can create stale history entries from the old connection. The back button handler now validates the origin of the previous history entry using WebBackForwardList before navigating back.
If the previous entry has a different origin (stale entry), the handler navigates to the base URL instead of attempting to load an unreachable page.
If the user is already on the root path, back exits the app as expected.
doUpdateVisitedHistory keeps the back callback enabled when on a non-root path, so pressing back navigates to root before exiting.

Tested:

WiFi → mobile data switch on Lovelace tabs: URL switches correctly, page preserved
WiFi → mobile data switch on history page with query parameters (/history?entity_id=...&start_date=...): page and filters preserved
Mobile data → WiFi switch: same behavior, page preserved
Back button after switch: navigates to base URL (no "cannot connect" error)
Back button on base URL: exits app correctly
Multiple rapid switches: no infinite loading loops

Gifford47:
@TimoPtr all tests are successfully done. Do you have any hints?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -440,9 +440,46 @@ class WebViewActivity :
...
+                if (webView.canGoBack()) {
+                    // Check if the previous history entry has the same origin as
+                    // the current URL. After an internal/external URL switch,
+                    // stale entries from the old connection may remain in history.
+                    val backForwardList = webView.copyBackForwardList()
+                    val currentIndex = backForwardList.currentIndex
+                    if (currentIndex > 0) {
+                        val previousUrl = backForwardList.getItemAtIndex(currentIndex - 1).url.toUri()
+                        val currentBase = loadedUrl
+                        if (currentBase != null && previousUrl.hasSameOrigin(currentBase)) {
```
TimoPtr: currentBase could also be `"about:blank"`

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1423,34 +1463,51 @@ class WebViewActivity :
...
-                } else {
-                    showSystemUI()
-                }
-
-                var path = intent.getStringExtra(EXTRA_PATH)
-                if (path?.startsWith("entityId:") == true) {
-                    // Get the entity ID from a string formatted "entityId:domain.entity"
-                    // https://github.com/home-assistant/core/blob/dev/homeassistant/core.py#L159
+                val intentPath = intent.getStringExtra(EXTRA_PATH)
+                intent.removeExtra(EXTRA_PATH)
```
TimoPtr: Why moving this line.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1423,34 +1463,51 @@ class WebViewActivity :
...
+        val query = uri.encodedQuery
+            ?.split("&")
+            ?.filter { !it.startsWith("external_auth=") }
+            ?.joinToString("&")
+            ?.takeIf { it.isNotEmpty() }
+        return buildString {
+            append(path)
+            query?.let { append("?$it") }
+            uri.encodedFragment?.let { append("#$it") }
+        }
```
TimoPtr: Would be better to not use a string here but a proper URL/URI object to modify it properly without any risk.

Gifford47: Thanks for the review! Here's what I addressed:

about:blank edge case: Added a previousUrl.scheme?.startsWith("http") == true guard before hasSameOrigin in the back navigation handler. This skips about:blank and any other non-HTTP entries that may appear in history before the first real page loads.

removeExtra(EXTRA_PATH) placement: This is already called immediately after reading the extra and before presenter.load() — same position as before, just renamed the local variable to intentPath for clarity.

URI construction: Replaced the buildString concatenation in getCurrentWebViewRelativeUrl() with Uri.Builder using encodedPath(), encodedQuery(), and encodedFragment() for safer URL construction.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
Gifford47: removeExtra(EXTRA_PATH) placement: This is already called immediately after reading the extra and before presenter.load() — same position as before, just renamed the local variable to intentPath for clarity.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
Gifford47: about:blank edge case: Added a previousUrl.scheme?.startsWith("http") == true guard before hasSameOrigin in the back navigation handler. This skips about:blank and any other non-HTTP entries that may appear in history before the first real page loads.

Gifford47:
@TimoPtr @jpelgrom 
It it possible to add this pr in the next release? The current behaviour is very annoying.

TimoPtr:
> @TimoPtr @jpelgrom It it possible to add this pr in the next release? The current behaviour is very annoying.

It probably won't make it to this release. We need to properly test it @jpelgrom and I before merging. This piece is quite critical and proven to be error prone. Sorry for the delay I was focused on other changes. 

The current behavior has been there for years, but your changes would be a nice quality of life improvement for sure.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -440,9 +440,51 @@ class WebViewActivity :
 
         decor = window.decorView as FrameLayout
 
-        val onBackPressed = object : OnBackPressedCallback(webView.canGoBack()) {
+        val onBackPressed = object : OnBackPressedCallback(true) {
             override fun handleOnBackPressed() {
-                if (webView.canGoBack()) webView.goBack()
+                if (webView.canGoBack()) {
```
TimoPtr: Could you actually extract this logic and write unit test for it, I think it is very important to have a strong coverage on this logic, and writing the tests could already highlights some flaws in the design. It is also going to help me migrate this code into the new Frontend logic I'm writing.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -440,9 +440,51 @@ class WebViewActivity :
...
+                    } else {
+                        webView.goBack()
+                        return
+                    }
+                }
+                // History is empty or previous entry has a different origin
+                // (stale entry from old connection). Navigate to base URL
+                // instead of going back to an unreachable page.
+                val currentUrl = loadedUrl
+                if (currentUrl != null && currentUrl.hasNonRootPath()) {
```
TimoPtr: so if it is about:blank what is going on?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -440,9 +440,51 @@ class WebViewActivity :
...
+                        .fragment(null)
+                        .build()
+                    clearHistory = true
+                    loadedUrl = baseUrl
+                    webView.loadUrl(baseUrl.toString())
+                } else {
+                    // Already on root — let the system handle back (exit app)
+                    isEnabled = false
+                    onBackPressedDispatcher.onBackPressed()
+                    isEnabled = true
```
TimoPtr: Why do we need to disable then call the system then enable back if it closes? I think this is something worse documenting. You don't want someone to remove this logic later if it is important.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: Move it back after just before calling load 🙏🏻. I want to keep the git history clean on this part.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: And move this block back to where it was too.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1423,34 +1468,53 @@ class WebViewActivity :
...
+            ?.filter { !it.startsWith("external_auth=") }
+            ?.joinToString("&")
+            ?.takeIf { it.isNotEmpty() }
+        return Uri.Builder()
+            .encodedPath(path)
+            .encodedQuery(filteredQuery)
+            .encodedFragment(uri.encodedFragment)
+            .build()
+            .toString()
+    }
```
TimoPtr, line 1610:
```suggestion
    /**
     * Extracts the relative URL (path, query parameters, and fragment) from the current WebView URL.
     *
     * Used to preserve the user's current page when the base URL changes, for example when
     * switching between internal and external connections. The `external_auth` query parameter
     * is excluded because the presenter re-adds it on every load.
     *
     * The root path (`/`) is treated as empty since it represents the home page with no
     * meaningful relative navigation.
     *
     * @return the relative URL string (e.g. `/history?start_date=2026-01-01#tab`),
     *   or `null` if the WebView has no URL or the result would be empty.
     */
    override fun getCurrentWebViewRelativeUrl(): String? {
        val uri = webView.url?.toUri() ?: return null
        // Strip 'external_auth' since the presenter re-adds it on every load.
        val relativeUrl = Uri.Builder()
            .encodedPath(uri.encodedPath?.takeIf { it.length > 1 })
            .apply {
                for (name in uri.queryParameterNames) {
                    if (name == "external_auth") continue
                    for (value in uri.getQueryParameters(name)) {
                        appendQueryParameter(name, value)
                    }
                }
            }
            .encodedFragment(uri.encodedFragment)
            .build()
            .toString()
        return relativeUrl.takeIf { it.isNotEmpty() }
    }
```

Check this proposal, it doesn't rely on string manipulation except than testing parameterNames already decoded. Also it takes in consideration that you can have parameters or segment you might want to preserve and also to match the documentation.

If you agree with this proposal I would like this to be extracted into a helper extension on a Uri and fully tested. By doing so it can be reuse elsewhere in the project.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
TimoPtr: This still need to be addressed.

TimoPtr (changes requested):

Gifford47:
Thanks for the thorough review! All three points have been addressed:

1. Callback disable/enable pattern — Added a comment explaining that we must temporarily disable the callback so the dispatcher invokes the next handler in the chain (the default Activity handler that finishes the activity). Re-enabling keeps it functional in case the activity is not destroyed (e.g. multi-window mode).

2. about:blank edge case — Added a comment clarifying that loadedUrl is always the HTTP(S) URL set by the presenter, never about:blank. If the WebView shows about:blank before the first page loads, loadedUrl is null and the code correctly falls through to the system back handler.

3. getCurrentWebViewRelativeUrl refactoring — Adopted your proposal: extracted the logic into a Uri.toRelativeUrl(excludeParams) extension function in UrlUtil.kt using the queryParameterNames / getQueryParameters() API instead of string splitting. Added KDoc documentation. The method in WebViewActivity is now a single line. Added 10 unit tests in UriExtensionsTest.kt covering: path-only, path+query, path+query+fragment, param exclusion, multiple exclusions, root path → null, no path → null, fragment-only, and nested paths.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: @Gifford47 don't forget this one.

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt
```diff
@@ -212,3 +212,37 @@ fun Uri.hasNonRootPath(): Boolean {
...
+
+    val relativeUrl = Uri.Builder()
+        .encodedPath(path)
+        .apply {
+            for (name in queryParameterNames) {
+                if (name in excludeParams) continue
+                for (value in getQueryParameters(name)) {
+                    appendQueryParameter(name, value)
+                }
+            }
```
TimoPtr:
```suggestion
            queryParameterNames
                .filterNot { it in excludeParams }
                .flatMap { name -> getQueryParameters(name).map { name to it } }
                .forEach { (name, value) -> appendQueryParameter(name, value) }
```

This would be a bit more functional oriented, but yours work so up to you.

Gifford47:
I'm using this logic now for quite a while and everything works as expected. Any suggestions @TimoPtr ?

TimoPtr:
I need to play with it to see how it behaves, I didn't have time to do this yet. @jpelgrom I think this is also ready for you to look at. I'm mostly curious about the back behavior while changing URLs.

We will also need to port this change to the `FrontendScreen` @Gifford47 that is the new way to show the Frontend, this is only visible in debug.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -441,9 +442,61 @@ class WebViewActivity :
 
         decor = window.decorView as FrameLayout
 
-        val onBackPressed = object : OnBackPressedCallback(webView.canGoBack()) {
+        val onBackPressed = object : OnBackPressedCallback(true) {
```
jpelgrom: Not the biggest fan of this change, since it means you lose the 'adaptive back' animations showing a peek of the home screen when there is no more back in the app (true = callback is always enabled so Android will always assume back is inside the app). Why not extract your logic into a function returning true/false, so we can keep this behavior?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -632,7 +685,10 @@ class WebViewActivity :
 
                 override fun doUpdateVisitedHistory(view: WebView?, url: String?, isReload: Boolean) {
                     super.doUpdateVisitedHistory(view, url, isReload)
-                    onBackPressed.isEnabled = canGoBack()
+                    // Keep the callback enabled when there's history OR when the current
+                    // URL has a non-root path (so pressing back navigates to root first).
+                    onBackPressed.isEnabled = canGoBack() ||
+                        url?.toUri()?.hasNonRootPath() == true
```
jpelgrom: Why add this?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1431,17 +1487,19 @@ class WebViewActivity :
                     showSystemUI()
                 }
 
-                var path = intent.getStringExtra(EXTRA_PATH)
-                if (path?.startsWith("entityId:") == true) {
-                    // Get the entity ID from a string formatted "entityId:domain.entity"
-                    // https://github.com/home-assistant/core/blob/dev/homeassistant/core.py#L159
```
jpelgrom, line 1562: Please don't remove this, it's useful documenting what it's doing

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt
```diff
@@ -212,3 +212,35 @@ fun Uri.hasNonRootPath(): Boolean {
...
     return path.isNotBlank() && path != "/"
 }
+
+/**
+ * Extracts the relative URL (path, filtered query parameters, and fragment) from this [Uri].
+ *
+ * Used to preserve the user's current page when the base URL changes, for example when
+ * switching between internal and external connections. Certain query parameters (like
+ * `external_auth`) can be excluded because the presenter re-adds them on every load.
+ *
```
jpelgrom: That is what _you_ are using it for, but this is an util function and is more generic.

```suggestion
```

common/src/test/kotlin/io/homeassistant/companion/android/util/UriExtensionsTest.kt
```diff
@@ -57,4 +58,75 @@ class UriExtensionsTest {
...
+    }
+
+    @Test
+    fun `toRelativeUrl excludes specified query params`() {
+        val uri = Uri.parse("https://example.com/dashboard?external_auth=1&lang=en")
+        assertEquals("/dashboard?lang=en", uri.toRelativeUrl(excludeParams = setOf("external_auth")))
+    }
+
+    @Test
+    fun `toRelativeUrl returns null when only excluded params remain`() {
```
jpelgrom: This test name suggests the result will be `null` but it's not

jpelgrom (changes requested):
The path is preserved when switching servers, which should not happen as the path may not exist on the other server (resulting in a broken page) and it leaks information about the initial server to the other server.

Gifford47:
Thanks for the thorough review @jpelgrom, all great catches!

Path leaking on server switch — I'll guard the getCurrentWebViewRelativeUrl() fallback so it only applies on internal/external URL changes (baseUrlChanged), not on actual server switches where isNewServer is true.

OnBackPressedCallback(true) — You're right, losing predictive back animations isn't ideal. I'll extract a canHandleBack() function that returns canGoBack() || hasNonRootPath() and use that for the initial enabled state + the doUpdateVisitedHistory update. That way we keep the adaptive behavior.

doUpdateVisitedHistory — This is needed so that pressing back navigates to root first even when there's no WebView history. The canHandleBack() extraction should make this clearer.

entityId comment — Will restore the original comment with the core.py link.

KDoc — Agreed, I'll make it more generic since it's a util function.

Test name — You're right, the name is misleading. Will rename to something like toRelativeUrl strips excluded params leaving path only.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: I think this is still relevant especially when I see you next PR, that almost use the same code. In the other PR you've added this to the Screen code, I think it's not the place for it since it mostly logic it deserve a dedicated file with proper test. By doing so you could use the same code in the too places.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -129,20 +129,39 @@ class WebViewPresenterImpl @Inject constructor(
...
             handleUrlState(
                 urlState = urlState,
-                path = path,
-                shouldConsumePath = shouldConsumePath,
-                isNewServer = isNewServer,
+                path = effectiveRelativeUrl,
+                shouldConsumePath = effectiveRelativeUrl != null,
+                // Clear history when the base URL changes (e.g. internal <-> external)
+                // because old URLs in the back stack would be unreachable on the new network.
+                isNewServer = isNewServer || baseUrlChanged,
```
TimoPtr:
```suggestion
                clearHistory = isNewServer || baseUrlChanged,
```
Let's rename the parameter to be more accurate.

TimoPtr (commented):
Let's check my last comments so that @jpelgrom can take a look after to merge this.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigation.kt
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Determines the appropriate back action for a WebView based on its history and current URL.
+ *
+ * The resolution logic:
+ * 1. If the WebView has valid back history with a same-origin previous entry,
+ *    returns [BackAction.GoBack] so the user can navigate back normally.
+ * 2. If there is no valid back history (empty, cross-origin, or non-HTTP entries)
+ *    and the current URL has a non-root path, returns [BackAction.NavigateToRoot]
+ *    so the user is taken to the home page first.
```
TimoPtr: I'm lost ... What you want is not to go back to root no? it is to preserve the path and param no?

I think it's time to drop your AI agent and start to manually look at the logic.

TimoPtr: Ok got it ... I was lost after looking at this way too many time already. The preserve path and param is only when changing URL not when going back.

That might be something we could consider actually too. (In another PR)

I still think that to get this final step you should drop your AI agent and do it manually it's quite sensitive.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/BackAction.kt
```diff
@@ -0,0 +1,74 @@
...
+sealed interface BackAction {
+    /** Navigate back in the WebView history. */
+    data object GoBack : BackAction
+
+    /** Clear history and navigate to the root URL of the current server. */
+    data class NavigateToRoot(val rootUrl: Uri) : BackAction
+
+    /** No more back navigation possible — exit the screen. */
+    data object Exit : BackAction
+}
```
TimoPtr, line 76: I'm not a big fan of the Exit action it seems too precise, and set the action for the screen itself that could decide to do something else. What do you think about `None` that let the caller decide what to do with it.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigation.kt
```diff
@@ -0,0 +1,74 @@
...
+ * @return the [BackAction] that the caller should execute
+ */
+fun resolveBackAction(webView: WebView, loadedUrl: Uri?): BackAction {
+    if (webView.canGoBack()) {
+        val backForwardList = webView.copyBackForwardList()
+        val currentIndex = backForwardList.currentIndex
+        if (currentIndex > 0) {
+            val previousUrl = Uri.parse(
+                backForwardList.getItemAtIndex(currentIndex - 1).url,
+            )
```
TimoPtr:
```suggestion
            val previousUrl = backForwardList.getItemAtIndex(currentIndex - 1).url.toUri()
```

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigation.kt
```diff
@@ -0,0 +1,74 @@
...
+ *    so the user is taken to the home page first.
+ * 3. Otherwise returns [BackAction.Exit] so the caller can finish the activity or
+ *    pop the navigation stack.
+ *
+ * @param webView the WebView whose history is inspected
+ * @param loadedUrl the current URL shown in the WebView (as tracked by the caller,
+ *        not necessarily [WebView.getUrl] which can be `about:blank` during loads)
+ * @return the [BackAction] that the caller should execute
+ */
+fun resolveBackAction(webView: WebView, loadedUrl: Uri?): BackAction {
```
TimoPtr: What about not giving the webView to this function directly, so no one actually call `webView.goBack()` from this function since it is theoretically possible. Something like 

```kotlin
fun resolveBackAction(webView: WebView, loadedUrl: Uri?): BackAction {
    val previousUrl = if (webView.canGoBack()) {
        val backForwardList = webView.copyBackForwardList()
        backForwardList.currentIndex
            .takeIf { it > 0 }
            ?.let { backForwardList.getItemAtIndex(it - 1).url }
            ?.toUri()
    } else {
        null
    }

    return resolveBackAction(previousUrl, loadedUrl)
}
@VisibleForTesting
internal fun resolveBackAction(previousUrl: Uri?, loadedUrl: Uri?): BackAction {
    if (previousUrl != null) {
```

This way the logic is in a pure function easy to test even without mocking the WebView.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -444,7 +447,25 @@ class WebViewActivity :
 
         val onBackPressed = object : OnBackPressedCallback(webView.canGoBack()) {
             override fun handleOnBackPressed() {
-                if (webView.canGoBack()) webView.goBack()
+                when (val action = resolveBackAction(webView, loadedUrl)) {
+                    BackAction.GoBack -> webView.goBack()
+                    is BackAction.NavigateToRoot -> {
+                        clearHistory = true
+                        loadedUrl = action.rootUrl
+                        webView.loadUrl(action.rootUrl.toString())
```
TimoPtr, line 483: Any reason to not call `load` of the presenter ? I think it is quite important to go through the presenter logic when we want to load a URL. I have no idea of the implication just thinking, that might avoid having to set the loadedUrl and clearHistory here.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: You didn't answer.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -225,7 +244,7 @@ class WebViewPresenterImpl @Inject constructor(
                 } else {
                     view.loadUrl(
                         url = urlWithAuth,
-                        keepHistory = !isNewServer,
+                        keepHistory = !clearHistory,
```
TimoPtr: I made a mistake in my last comment, we could use keepHistory just to avoid having to inverse the logic.

app/src/test/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigationTest.kt
```diff
@@ -0,0 +1,119 @@
...
+
+@RunWith(RobolectricTestRunner::class)
+class WebViewBackNavigationTest {
+
+    private fun mockWebView(canGoBack: Boolean = false): WebView = mockk {
+        every { this@mockk.canGoBack() } returns canGoBack
+    }
+
+    @Test
+    fun `resolveBackAction returns Exit when no history and root URL`() {
```
TimoPtr: Update the titles to fit the other test in the project

app/src/test/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigationTest.kt
```diff
@@ -0,0 +1,119 @@
...
+        assertTrue(action is BackAction.NavigateToRoot)
+        val rootUrl = (action as BackAction.NavigateToRoot).rootUrl
+        assertEquals("/", rootUrl.path)
+        assertEquals("1", rootUrl.getQueryParameter("external_auth"))
+        assertEquals(null, rootUrl.getQueryParameter("start_date"))
+        assertEquals(null, rootUrl.fragment)
+    }
+
+    @Test
+    fun `resolveBackAction returns Exit when on root path`() {
```
TimoPtr: Same as `    fun `resolveBackAction returns Exit when no history and root URL`() {` no?

TimoPtr (changes requested):
We're almost there

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigation.kt
Gifford47: Done -split into a WebView convenience overload and a pure @VisibleForTesting internal function with (previousUrl: Uri?, loadedUrl: Uri?). Tests call the pure function directly now, no more WebView mocking needed.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigation.kt
Gifford47: done

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/BackAction.kt
Gifford47: Agreed, renamed to BackAction.None. The function just signals "nothing I can do" and the caller decides.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
Gifford47: presenter.load() does a full reconnect cycle (URL resolution, auth check, security level dialog). For NavigateToRoot we stay on the same server and only change the path to / -going through the presenter would add overhead and could trigger side effects like the security level prompt.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
Gifford47: Sorry about that. The callback needs to be enabled when hasNonRootPath() is true even with empty history -otherwise pressing back on e.g. /history would skip the navigate-to-root step and exit the app directly. It also keeps the predictive back animation working on Android 14+, since the system needs to know upfront that we'll handle the gesture.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
Gifford47: Done, using keepHistory directly now. Reads much cleaner without the !clearHistory inversion.

app/src/test/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigationTest.kt
Gifford47: Done -all tests use Given/When/Then convention now.

app/src/test/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigationTest.kt
Gifford47: Yep, removed it.

Gifford47:
Yes, I used the agent to thoroughly analyze the logic and review the code. However, the new logic was developed by me and has been tested by me as well. In fact, I spent the past two weeks carefully re-testing everything to ensure it works reliably:

Pure function extraction - Split resolveBackAction into a WebView convenience overload that extracts the previousUrl from the back/forward list, and a pure VisibleForTesting internal function that takes (previousUrl: Uri?, loadedUrl: Uri?). This makes the logic testable without mocking WebView. All tests now call the pure function directly - no more MockK for WebView/WebBackForwardList.

Uri.parse() -> .toUri() - Replaced with the Kotlin extension as suggested.

Exit -> None - Good call, renamed to BackAction.None. The caller decides what to do - the function just says "I can't handle this". Updated all callsites.

clearHistory -> keepHistory - Reverted the rename, now using keepHistory directly to avoid the inverted !clearHistory logic. 

presenter.load() vs webView.loadUrl() for NavigateToRoot - I intentionally use webView.loadUrl() here because presenter.load() does a full reconnect cycle (resolve URL, auth check, security level). For NavigateToRoot we're staying on the same server and just changing the path to /, so going through the presenter would add unnecessary overhead and could trigger unwanted side effects like the security level dialog.

@jpelgrom's question about hasNonRootPath in doUpdateVisitedHistory - Sorry I missed this one. The OnBackPressedCallback needs to be enabled not just when canGoBack() is true, but also when the user is on a sub-path like /history. Without this, pressing back on a sub-path with empty history would trigger the system back handler (exit app) instead of first navigating to root. This also keeps the predictive back animation working correctly on Android 14+ - the system needs to know in advance that we'll handle the gesture.

Test naming - Updated all test names to Given/When/Then convention.

Duplicate test - Removed the duplicate "Exit when on root path" test (was identical to "no previous url and root loaded url").

Also added a new test for the about:blank edge case - verifying that a non-HTTP previousUrl doesn't trigger GoBack.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigation.kt
```diff
@@ -0,0 +1,79 @@
...
+ *    (e.g. exit the activity or pop the navigation stack).
+ *
+ * @param previousUrl the URL of the previous entry in the WebView's back stack,
+ *        or `null` if the history is empty or the WebView cannot go back
+ * @param loadedUrl the current URL shown in the WebView (as tracked by the caller,
+ *        not necessarily [WebView.getUrl] which can be `about:blank` during loads)
+ * @return the [BackAction] that the caller should execute
+ */
+@VisibleForTesting
+internal fun resolveBackAction(previousUrl: Uri?, loadedUrl: Uri?): BackAction {
```
TimoPtr: Make this private, you can test all the behavior from within the public function.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigation.kt
```diff
@@ -0,0 +1,79 @@
...
+import androidx.annotation.VisibleForTesting
+import androidx.core.net.toUri
+import io.homeassistant.companion.android.util.hasNonRootPath
+import io.homeassistant.companion.android.util.hasSameOrigin
+
+/**
+ * Convenience overload that extracts the previous URL from the [WebView]'s
+ * back/forward list and delegates to the pure [resolveBackAction] function.
+ */
+fun resolveBackAction(webView: WebView, loadedUrl: Uri?): BackAction {
```
TimoPtr, line 27: We don't care about the implementation detail in the documentation. We care about how the function behaves, the resolution logic should be attach to this function since it is this documentation that we see in the IDE using this function.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: Actually I think we want this side effect, the overhead is very small here and safer. @jpelgrom wdyt?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1559,17 +1585,21 @@ class WebViewActivity :
...
                     // https://github.com/home-assistant/core/blob/dev/homeassistant/core.py#L159
                     val pattern = "(?<=^entityId:)((?!.+__)(?!_)[\\da-z_]+(?<!_)\\.(?!_)[\\da-z_]+(?<!_)$)".toRegex()
-                    val entity = pattern.find(path)?.value ?: ""
+                    val entity = pattern.find(intentPath)?.value ?: ""
                     if (
                         entity.isNotBlank() &&
                         serverManager.getServer(presenter.getActiveServer())?.version?.isAtLeast(2025, 6, 0) == true
                     ) {
                         path = "/?more-info-entity-id=$entity"
+                        moreInfoEntity = ""
```
TimoPtr: Is this change related to your PR? I'm not sure we need it no?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -129,20 +129,39 @@ class WebViewPresenterImpl @Inject constructor(
...
 
         if (isInternalOverride != null) {
             Timber.d("Using isInternalOverride to get URL")
         }
 
         serverManager.connectionStateProvider(serverId).urlFlow(isInternalOverride).collect { urlState ->
-            val shouldConsumePath = !pathConsumed && path != null
-            if (shouldConsumePath) pathConsumed = true
+            val currentBaseUrl = (urlState as? UrlState.HasUrl)?.url
+            val baseUrlChanged = lastBaseUrl != null && currentBaseUrl != null && lastBaseUrl != currentBaseUrl
```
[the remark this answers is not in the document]
TimoPtr: I think we could simplify the test by using the lastBaseUrl using nullability

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -129,20 +129,39 @@ class WebViewPresenterImpl @Inject constructor(
...
+                // On internal/external URL switches, preserve the full relative URL
+                // (path + query params + fragment) so the user stays on the exact same
+                // page, including filtered views like history with date ranges.
+                // Only do this for connection type changes on the same server, not
+                // for server switches where the path may not exist and would leak
+                // information about the previous server.
+                withContext(Dispatchers.Main) { view.getCurrentWebViewRelativeUrl() }
+            } else {
+                null
+            }
```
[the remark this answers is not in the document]
TimoPtr: Probably to consider, please check this comment @Gifford47.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -129,20 +129,39 @@ class WebViewPresenterImpl @Inject constructor(
...
                 urlState = urlState,
-                path = path,
-                shouldConsumePath = shouldConsumePath,
-                isNewServer = isNewServer,
+                path = effectiveRelativeUrl,
+                shouldConsumePath = effectiveRelativeUrl != null,
+                // Clear history when the base URL changes (e.g. internal <-> external)
+                // because old URLs in the back stack would be unreachable on the new network.
+                keepHistory = !(isNewServer || baseUrlChanged),
             )
```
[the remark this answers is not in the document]
TimoPtr: Indeed we should test this, we already have tests for the presenter.

app/src/test/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigationTest.kt
```diff
@@ -0,0 +1,85 @@
...
+
+import android.net.Uri
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class WebViewBackNavigationTest {
```
TimoPtr: Let's not use Robolectric here, a simple mock with mockk of the WebView does the job.

app/src/test/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigationTest.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    @Test
+    fun `given no previous url and root loaded url, when resolving back action, then returns None`() {
```
TimoPtr: Check the other tests in the project, It is a sentence so it should start with `Given` not `given`

app/src/test/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigationTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        val previousUrl = Uri.parse("https://other.server:8123/lovelace/0")
+        val loadedUrl = Uri.parse("https://ha.local:8123/history?external_auth=1")
+
+        val action = resolveBackAction(previousUrl = previousUrl, loadedUrl = loadedUrl)
+
+        assertTrue(action is BackAction.NavigateToRoot)
+    }
+
+    @Test
+    fun `given non-http previous url, when resolving back action, then does not go back`() {
```
TimoPtr:
```suggestion
    fun `given non-http previous url, when resolving back action, then returns NavigateToRoot`() {
```

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
Gifford47: I looked into replacing the manual loadUrl call with presenter.load(lifecycle), but that wouldn't give us the behavior we need: collectUrlStateChanges only sets keepHistory = false when baseUrlChanged || isNewServer. For a NavigateToRoot action the base URL stays the same -so presenter.load() would pass keepHistory = true and leave the stale entries in the back stack. That's exactly what we want to avoid here (the user pressed back to get out of a sub-page). Going through the presenter would require adding a new loadRoot() method (or a forceClearHistory parameter), which felt like a bigger API change than this PR should do.  the current code sets clearHistory/loadedUrl and calls webView.loadUrl directly, which mirrors what WebViewPresenterImpl.loadUrl does internally for this edge case.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigation.kt
Gifford47: done

app/src/test/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigationTest.kt
Gifford47: Done

app/src/test/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigationTest.kt
Gifford47: Will do -removing @RunWith(RobolectricTestRunner::class) and mocking WebView + WebBackForwardList with mockk. Combined with making the inner resolveBackAction private, the tests will go through the public overload and cover the back/forward extraction path too.

app/src/test/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigationTest.kt
Gifford47: Renamed

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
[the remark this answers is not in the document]
Gifford47: Good idea. I can drop the explicit != null checks by treating lastBaseUrl == null as „first emission, nothing to compare"

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/BackAction.kt
```diff
@@ -0,0 +1,74 @@
+package io.homeassistant.companion.android.util.compose.webview
```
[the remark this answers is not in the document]
Gifford47: done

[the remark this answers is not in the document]
Gifford47: done

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/WebViewBackNavigation.kt
Gifford47: Done in e8de08e63. The resolution-logic KDoc (same-origin → GoBack, non-root → NavigateToRoot, else None) now lives on the public resolveBackAction(webView, loadedUrl), so it shows at call sites in the IDE. The private overload is intentionally undocumented since it's an implementation detail.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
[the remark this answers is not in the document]
Gifford47: Fixed in e8de08e63. Copilot was right — the code comment said "only for connection-type changes on the same server" but the actual code didn't gate on !isNewServer

server switches always discard the previous WebView path and never carry navigation context from the old server into the new one.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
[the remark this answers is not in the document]
Gifford47: done

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
Gifford47: You're right -the rename itself isn't necessary. path can stay mutable without introducing intentPath, since it's only reassigned after the pattern.find(path) call. I'll revert the rename in the next commit and keep only the two actual fixes from that change: moving intent.removeExtra(EXTRA_PATH) before presenter.load() and clearing moreInfoEntity on the URL-path branch (both from your earlier review rounds)

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/BackAction.kt
```diff
@@ -0,0 +1,69 @@
...
+ *    (e.g. exit the activity or pop the navigation stack).
+ *
+ * @param webView the WebView whose back/forward list is inspected
+ * @param loadedUrl the current URL shown in the WebView (as tracked by the caller,
+ *        not necessarily [WebView.getUrl] which can be `about:blank` during loads)
+ */
+fun resolveBackAction(webView: WebView, loadedUrl: Uri?): BackAction {
+    val previousUrl = if (webView.canGoBack()) {
+        val backForwardList = webView.copyBackForwardList()
+        backForwardList.getItemAtIndex(backForwardList.currentIndex - 1).url.toUri()
```
TimoPtr: Let's be a bit safer here and check that the currentIndex is not 0.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: @Gifford47 ?

Gifford47: You're right, this line isn't needed.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/BackAction.kt
Gifford47: added an explicit currentIndex > 0 check (via previousIndex >= 0)

TimoPtr (commented):
Let's wait for @jpelgrom final review before merging this. Also I would like to wait that the current release is on the play store before merging this one.

Gifford47:
> Let's wait for @jpelgrom final review before merging this. Also I would like to wait that the current release is on the play store before merging this one.

Did you already test the feature? On my side everything works as expected. I tested it for some days in my productive environment.

TimoPtr:
> > Let's wait for @jpelgrom final review before merging this. Also I would like to wait that the current release is on the play store before merging this one.
> 
> Did you already test the feature? On my side everything works as expected. I tested it for some days in my productive environment.

I honestly didn't because I don't have a test setup with internal/external URL. I want to test this once it's merged because we have an internal lane on the play store. I focused on the code itself for now.

jpelgrom:
The switching now mostly works from a technical point of view and preserves the URL when switching between networks on the same server, and doesn't when switching servers.

However I have two issues:

1. I never seem to get the 'predictive back animations' anymore, both with the very simple app start>dashboard loads>swipe to go back and variations with more navigation and/or switching networks. Did you test this?

2. I still find the back behavior 'inserting' a root page after switching networks with a non-root path very odd. This also causes a behavior change for shortcuts which deeplink to a specific dashboard: previously that dashboard was the root (which in this case is correct) so pressing back closes the app - now it adds the `/` page and requires one more back.

Before:

https://github.com/user-attachments/assets/0c10b4b5-3316-43ca-adde-34b4e4185c0a

After (this also shows the animation issue):

https://github.com/user-attachments/assets/ee25a647-8374-4804-827a-2f67d6e7847c

(I'm using energy as an example as I'm not that much into dashboards myself, but you can imagine people having shortcuts to specific dashboards and/or views who get this extra step now.)

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/BackAction.kt
```diff
@@ -0,0 +1,74 @@
...
+private fun resolveBackAction(previousUrl: Uri?, loadedUrl: Uri?): BackAction {
+    if (previousUrl != null &&
+        loadedUrl != null &&
+        previousUrl.scheme?.startsWith("http") == true &&
+        previousUrl.hasSameOrigin(loadedUrl)
+    ) {
+        return BackAction.GoBack
+    }
+
+    if (loadedUrl != null && loadedUrl.hasNonRootPath()) {
```
TimoPtr: To cover @jpelgrom comments, let's not do this operation if `previousUrl` is null.

We should continue to conversation about the UX in https://discord.com/channels/330944238910963714/1496864991855251558.

Gifford47: restricted NavigateToRoot to require previousUrl != null. When there's no back-stack entry, resolveBackAction now returns None and the activity delegates to the system back handler. Tests updated accordingly

Gifford47:
I've spent some time thinking about a solution for the keepHistory issue.
My proposal is to keep a separate logical back stack alongside the native one, so we don't depend on the WebView history surviving the switch.

Logic:
- New `WebViewNavigationState` holder with `logicalBack: ArrayDeque<String>` 
  and `currentLogicalPath: String`
- `doUpdateVisitedHistory` pushes only when the path component changes. 
  This filters out the HA frontend `replaceState` calls (e.g. 
  `/history?entity_id=...&start_date=...`). Cross-origin URLs and reloads 
  are skipped, native history handles those.
- Base URL change: keep the logical stack, load `newBaseUrl + currentLogicalPath`, 
  then `clearHistory()`. On `isNewServer`: also clear the logical stack.
- `resolveBackAction` order: `canGoBack()` -> `GoBack`, else 
  `logicalBack.isNotEmpty()` -> `GoBackLogical(path)`, else `None`. 
  `NavigateToRoot` goes away.
- `enabled` exposed as `StateFlow<Boolean>` from the holder, the activity 
  collects it. State is correct before the gesture starts.

Fixes both points:
1. Predictive back works again because `enabled` updates synchronously 
   from the flow, not inside `doUpdateVisitedHistory` after the load.
2. Deeplink to `/lovelace/energy` starts with empty `logicalBack` and 
   `currentLogicalPath = "/lovelace/energy"`, so back closes the app 
   directly without the extra `/` step.

Touched files: new `WebViewNavigationState.kt`, `BackAction.kt` (drop 
NavigateToRoot, add GoBackLogical), `WebViewPresenterImpl.kt` (hook 
collectUrlStateChanges into the holder), `WebViewActivity.kt` (collect 
the flow, route doUpdateVisitedHistory and back callback through the 
holder). Holder is plain Kotlin without view dependencies, so the Compose 
rewrite in #6386 can reuse it as-is.

If the direction works for you I'll prepare the changes.

TimoPtr:
> * `doUpdateVisitedHistory`

I'm not a big fan to try to keep a secondary history. Especially that if you are deep into the app and switch URL each back would reload the app because the origin would be different, so showing the loader every back.

I invite you to join the discord discussion with the product people to decide where to go with this. Since it's mostly a UX question not really technical.

jpelgrom:
Even with the latest changes, predictive back doesn't work at all in my testing. Likely because the history doesn't start with `/` but that shouldn't matter, my initial dashboard isn't at `/` and even the default is at `/home/overview` now.

My WebView has decided to start doing native crashes when changing network somewhat regularly, making testing quite inconvenient. It does seem that when it doesn't crash it is never inserting a navigation to the default dashboard anymore though.

Gifford47:
Pushed three commits addressing predictive-back end-to-end (both code paths):

1. 4cfa569b - Enable Android 14+ predictive-back by gating back handlers on canGoBack
Hoist canGoBack state from HAWebView to the caller (FrontendViewModel /ConnectionViewModel), driven by a new onCanGoBackChanged hook in HAWebViewClient. The internal BackHandler is now only enabled when the WebView has back-stack entries - otherwise the gesture falls through to the system and the peek animation works. Same change for the legacy WebViewActivity (drops the hasNonRootPath() fallback in doUpdateVisitedHistory).

2. 63bcc740 - Clear navigation back-stack after onboarding completes
navigateToFrontend() was pushing FrontendRoute on top of the finished OnboardingRoute, so back from the frontend popped to an empty (white) onboarding screen instead of exiting. Fixed with popUpTo(0) { inclusive = true }.

3. 2d8c9468 - Clear WebView history after transitioning out of about:blank
The WebView accumulated every URL loaded, including the about:blank placeholder for Loading /Error / SecurityLevel states. Pressing back from the root walked back into a white about:blank screen. HAWebViewClient.onPageFinished now calls webView.clearHistory() once it detects the transition BLANK_URL → real URL. SPA pushState is unaffected.

Tested on Android 16 emulator (API 36.1), gesture navigation, both code paths:

Legacy WebViewActivity (USE_FRONTEND_V2 = false): predictive-back works on root, normal webView.goBack() inside the app.
Compose FrontendScreen (USE_FRONTEND_V2 = BuildConfig.DEBUG): same - back through SPA history, then peek + exit cleanly from root, and reopening the app lands on the dashboard.

TimoPtr (commented):
I'll wait for @jpelgrom opinion before merging.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -129,20 +129,40 @@ class WebViewPresenterImpl @Inject constructor(
...
+            val effectiveRelativeUrl = if (!pathConsumed && path != null) {
+                pathConsumed = true
+                path
+            } else if (baseUrlChanged && !isNewServer) {
+                // On internal/external URL switches on the same server, preserve the full
+                // relative URL (path + query params + fragment) so the user stays on the exact
+                // same page, including filtered views like history with date ranges.
+                // Skipped for server switches where the path may not exist and would leak
+                // navigation context from the previous server.
+                withContext(Dispatchers.Main) { view.getCurrentWebViewRelativeUrl() }
```
jpelgrom: Is `withContext(Dispatchers.Main)` necessary? If yes I'd move it to inside the view (activity) function as this shouldn't be the presenter's concern.

jpelgrom (commented):
The back animation finally appears to work consistently! Thanks for your patience in trying to fix it every time 🙏

WebView is now crashing (natively) quite consistently for me when switching networks, both on the emulator with v133 and a real device with v147, making this hard to test. That isn't new and also happening with the main app. However, it may be related for:

I see the `NavigateToRoot` action in code (which I still dislike), but never seem to get it anymore after switching networks. In other words: after switching networks it always acts as if there is no history and closes the app. Fine by me but not your intention? Or if it is why is `NavigateToRoot` still in the code? (Tested both with manually switching between Wi-Fi and mobile data using the quick setting toggle, and in the emulator with `adb shell "svc wifi enable"`/`adb shell "svc wifi disable"`.)

jpelgrom:
Short video showing my test for the issue described above: it keeps the correct path when switching (note the network icon on the emulator, it reloads but keeps the set date!) but back exits the app.

https://github.com/user-attachments/assets/84bfa13c-932a-47ae-b223-92ffdfc42dcd

Gifford47:
thanks for the video, it actually showed me a race condition i had. clearHistory was scheduled inside onPageFinished of the new url, so whether NavigateToRoot fired after a network switch depended on if the webview finished loading before you pressed back. on my emulator it always fired, on yours the native webview crash probably wiped the history first so you saw "no history -> exit". inconsistent either way.

dropped the !baseUrlChanged part of keepHistory so the back-stack survives internal/external switches. now the cross-origin previousUrl is reliably there and resolveBackAction picks it up -> NavigateToRoot runs deterministically. server switches still discard history (different server, old paths cant be reached anyway). so NavigateToRoot stays in the code because it actually has a stable trigger now.

also dropped the redundant withContext(Dispatchers.Main) on the same function. the whole call chain is already on Main, was just noise on the presenter.

trade-off worth naming: right after a switch on a deep page the back gesture doesnt show the peek (callback is enabled because canGoBack=true). it runs NavigateToRoot instead, then the next back from / gets peek + exit. i think thats better UX than killing the app in one tap when the network silently flipped, but happy to discuss.

test matrix below, ran all of them on android 16 emulator with USE_FRONTEND_V2=false.

<table>
  <thead>
    <tr>
      <th>#</th>
      <th>scenario</th>
      <th>steps</th>
      <th>expected</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>A</td>
      <td>network switch on deep page</td>
      <td>
        open dashboard, navigate to /history?entity_id=...,<br>
        adb shell svc wifi disable,<br>
        wait for reload,<br>
        back
      </td>
      <td>
        lands on /, no peek.<br>
        next back = peek + exit
      </td>
    </tr>
    <tr>
      <td>B</td>
      <td>back from root</td>
      <td>
        open dashboard,<br>
        back
      </td>
      <td>
        peek animation,<br>
        app exits
      </td>
    </tr>
    <tr>
      <td>C</td>
      <td>deeplink, no history</td>
      <td>
        force-stop app,<br>
        adb shell am start -a VIEW -d "homeassistant://navigate/history",<br>
        back
      </td>
      <td>
        peek + exit immediately (no previousUrl)
      </td>
    </tr>
    <tr>
      <td>D</td>
      <td>normal in-app nav</td>
      <td>
        dashboard,<br>
        click into a detail page,<br>
        back
      </td>
      <td>
        webView.goBack(), no peek
      </td>
    </tr>
    <tr>
      <td>E</td>
      <td>reverse switch external -> internal</td>
      <td>
        start with wifi off,<br>
        navigate to /history,<br>
        adb shell svc wifi enable,<br>
        wait,<br>
        back
      </td>
      <td>
        lands on /,<br>
        next back = peek + exit
      </td>
    </tr>
  </tbody>
</table>

jpelgrom:
In WebViewActivity, this now seems to work consistently.

In the Compose-based frontend screens, this seems to introduce a new issue where I now get _two_ additional history entries which are empty.

https://github.com/user-attachments/assets/2ffa5661-2b02-45c8-9a23-e2b3a3b8ca64

To be honest the amount of edge cases/race conditions I'm discovering every review is quite discouraging. Not a personal attack, but if it's so hard to get it right even though the concept is easy to explain maybe we should reconsider the implementation.

TimoPtr:
> In WebViewActivity, this now seems to work consistently.
> 
> In the Compose-based frontend screens, this seems to introduce a new issue where I now get _two_ additional history entries which are empty.
> Screen.Recording.2026-05-25.at.21.30.07.mov
> 
> To be honest the amount of edge cases/race conditions I'm discovering every review is quite discouraging. Not a personal attack, but if it's so hard to get it right even though the concept is easy to explain maybe we should reconsider the implementation.

This PR should not impact the compose impl, did you test this PR or the other one with compose? (I didn't review the compose one TBH).

jpelgrom:
> This PR should not impact the compose impl, did you test this PR or the other one with compose? (I didn't review the compose one TBH).

You are right, this isn't this PR's issue. After testing again with the latest commit on main I now also see the issue there :). Sorry for the confusion, I was primarily testing behavior first and mistakenly assumed this still had Compose, forgetting it was split 2 weeks ago.

jpelgrom (approved):
OK, let's give this a try. I would like to have this in beta for a while so we can find any issues that might've been missed. @TimoPtr Not sure if we want to make a release before merging, or have a longer beta cycle - I'll let you decide.

jpelgrom:
@Gifford47 Note there is a minor unit test failure to fix

TimoPtr (approved):
Waiting for Google to approve current release before merging this one.

Gifford47:
@TimoPtr The PR is currently merged, isn't it?
Because something is broken, the default dashboard is loaded as before.

TimoPtr:
> @TimoPtr The PR is currently merged, isn't it? Because something is broken, the default dashboard is loaded as before.

It is in beta but not in the prod release.

PR #6452, opened 2026-02-17
Added `min`/`max` attributes to volume sensors.

blaubaer:
## Summary

Unfortunately, each android device can have different minimum and maximum values for the different volume settings. If we want to write automations and/or integrations which want to rely to the volume of a device these both values are also important. Currently, we at first have to find out for each device the maximum, by fully putting the volume up to max and adjust the automation for each device. We have for example 3 devices, with: `15`, `20` and `50` values.

By, adding `min`/`max` attributes to each volume sensors, we can react to those attributes in automations and/or integrations, without guessing/hacks.

## Checklist

## Any other notes

None

### DIFF
--- a/common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManagerTest.kt
+++ b/common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManagerTest.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.content.Context.AUDIO_SERVICE
 import android.media.AudioManager
 import dagger.hilt.android.EntryPointAccessors
+import dagger.hilt.android.testing.HiltTestApplication
 import io.homeassistant.companion.android.common.data.servers.ServerManager
 import io.homeassistant.companion.android.database.sensor.Attribute
 import io.homeassistant.companion.android.database.sensor.Sensor
@@ -29,7 +30,7 @@ import org.robolectric.RobolectricTestRunner
 import org.robolectric.annotation.Config
 
 @RunWith(RobolectricTestRunner::class)
-@Config(sdk = [28]) // Build.VERSION_CODES.P
+@Config(application = HiltTestApplication::class, sdk = [28]) // Build.VERSION_CODES.P
 class AudioSensorManagerTest {
 
     @get:Rule


common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
```diff
@@ -348,130 +352,93 @@ class AudioSensorManager : SensorManager {
...
+
+    private suspend fun shouldForceUpdateVolumeAttributes(
+        context: Context,
+        sensor: BasicSensor,
+        currentMin: Int,
+        currentMax: Int,
+    ): Boolean {
+        val attributes = sensorDao(context)
+            .getFull(sensor.id)
+            .toSensorWithAttributes()
```
TimoPtr: I'm not sure at 100% since I'm not super familiar yet with this part of the code, but you are getting a list of sensor in the case of multi-server and then toSensorWithAttributes only returns the first sensor with attributes so if one server is not up to date it wouldn't force update.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
```diff
@@ -348,130 +352,93 @@ class AudioSensorManager : SensorManager {
...
 
-        val volumeLevelDTMF = audioManager.getStreamVolume(AudioManager.STREAM_DTMF)
+        val current = audioManager.getStreamVolume(streamType)
+        val min = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            audioManager.getStreamMinVolume(streamType)
+        } else {
+            0
+        }
+        val max = audioManager.getStreamMaxVolume(streamType)
+        val force = shouldForceUpdateVolumeAttributes(context, sensor, min, max)
```
TimoPtr: I think we could skip the force update if the min/max are not changing (to be confirmed), for old users it would delay the sent of the attributes a bit but it would save a DB call made on every volume changes. Could you test without forcing the update?

common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManagerTest.kt
```diff
@@ -0,0 +1,147 @@
...
+                servers = any(),
+                permission = any(),
+                enabledByDefault = any(),
+            )
+        } answers { firstArg<String>() == AudioSensorManager.volMusic.id }
+        coEvery { sensorDao.get(any()) } returns emptyList()
+        coEvery { sensorDao.get(AudioSensorManager.volMusic.id) } returns listOf(volumeMusicSensor)
+
+        every { audioManager.getStreamVolume(AudioManager.STREAM_MUSIC) } returns 5
+        every { audioManager.getStreamMinVolume(AudioManager.STREAM_MUSIC) } returns 0
```
TimoPtr: Let's make the min different than 0 here, so that you can write a test that test the default behavior when bellow `VERSION_CODES.P`

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
blaubaer: This is a good point. This will now reflect all servers.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
blaubaer: Good point. I changed the code (with a proper comment which explains also the reasoning behind) to be only reflected once per app runtime... as only an app update and/or a change of the operating system can change min/max values.

common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManagerTest.kt
blaubaer: Makes fully sense. Implemented.

blaubaer:
@TimoPtr Do you think the failing test is really related to my changes?
```
WearOnboardingNavigationTest > Given a server discovered when clicking on it then show ConnectScreen then back goes to ServerDiscovery FAILED
    androidx.compose.ui.test.ComposeTimeoutException at WearOnboardingNavigationTest.kt:191
```

jpelgrom:
No, this test is known to fail sometimes. I've restarted it.

TimoPtr:
@blaubaer re-request a review when you think it is ready.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
```diff
@@ -348,130 +355,108 @@ class AudioSensorManager : SensorManager {
...
+        currentMax: Int,
+    ): Boolean {
+        if (sensor.id in sensorIdsWhereAttributesAreAlreadyForced) {
+            return true
+        }
+        sensorIdsWhereAttributesAreAlreadyForced += sensor.id
+        return sensorDao(context)
+            .getFull(sensor.id)
+            .toSensorsWithAttributes()
+            .any { sensorWithAttributes ->
```
TimoPtr:
```suggestion
            .all { sensorWithAttributes ->
```

I'm not 100% sure and I didn't test, but we should send forceUpdate if one sensor has different value. So when you check here you want all the sensors to match to check all the servers no?

TimoPtr:
@jpelgrom I would like your opinion on the forceUpdate logic.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
```diff
@@ -348,130 +355,108 @@ class AudioSensorManager : SensorManager {
...
+    private suspend fun areAttributesOfSensorAlreadyForcedAtRuntimeOrMatching(
+        context: Context,
+        sensor: BasicSensor,
+        currentMin: Int,
+        currentMax: Int,
+    ): Boolean {
+        if (sensor.id in sensorIdsWhereAttributesAreAlreadyForced) {
+            return true
+        }
+        sensorIdsWhereAttributesAreAlreadyForced += sensor.id
```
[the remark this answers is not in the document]
blaubaer: Makes sense. I moved the part of marked it as forced down to the point where I explicitly return `true`.

common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManagerTest.kt
```diff
@@ -0,0 +1,185 @@
...
+        )
+        coJustRun { sensorDao.update(capture(updatedSensor)) }
+
+        sensorManager.requestSensorUpdate(context)
+
+        assertNull(updatedSensor.captured.lastSentState)
+        assertNull(updatedSensor.captured.lastSentIcon)
+    }
+
+    @Test
```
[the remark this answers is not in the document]
blaubaer: I've added a test to better address that issue.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
blaubaer: Good point. Actually the multi-server thing still let me struggle a little bit. I've added an extra tests to make the potential risk better visible. And yeah, it failed. I've fixed that together with the feedback of the Copilot.

Sorry for the inconvenience.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
```diff
@@ -348,130 +355,115 @@ class AudioSensorManager : SensorManager {
...
+                    ?.value
+                    ?.toIntOrNull()
+                val storedMax = attributes
+                    .firstOrNull { it.name == ATTRIBUTE_MAX }
+                    ?.value
+                    ?.toIntOrNull()
+                storedMin == currentMin && storedMax == currentMax
+            }
+        ) {
+            sensorIdsWhereAttributesAreAlreadyForced += sensor.id
```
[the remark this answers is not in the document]
blaubaer: The race condition you describe is not a practical concern here. `requestSensorUpdate` processes all volume sensors sequentially within a single coroutine — there is no concurrent access to the same sensor ID. Even in a theoretical race scenario, the only consequence would be a redundant (but harmless) database query, as you noted yourself. Adding `synchronized` would introduce unnecessary complexity for a non-existent problem and would also conflict with the `suspend` nature of the function. I'll leave this as-is.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
```diff
@@ -348,130 +355,115 @@ class AudioSensorManager : SensorManager {
...
+            return true
+        }
```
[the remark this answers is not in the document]
blaubaer: You're right that it's correct — `!isEmpty() && all { ... }` handles the empty case properly by short-circuiting to `false`, which triggers a force update. I agree an early return for the empty case could be slightly more readable, but in this case the combined condition is compact and avoids an extra nesting level. Given that the method already has a clear KDoc explaining the intent, I think the current form is acceptable. Happy to reconsider if a maintainer feels strongly about it.

common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManagerTest.kt
```diff
@@ -0,0 +1,251 @@
...
+            ),
+        )
+        coJustRun { sensorDao.update(capture(updatedSensors)) }
+
+        sensorManager.requestSensorUpdate(context)
+
+        assertEquals(1, updatedSensors.size)
+        assertNull(updatedSensors[0].lastSentState)
+        assertNull(updatedSensors[0].lastSentIcon)
+    }
```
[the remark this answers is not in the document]
blaubaer: Addressed.

common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManagerTest.kt
```diff
@@ -0,0 +1,251 @@
...
+                Attribute(AudioSensorManager.volMusic.id, "max", "15", "int"),
+            ),
+        )
+        coJustRun { sensorDao.update(capture(updatedSensor)) }
+
+        sensorManager.requestSensorUpdate(context)
+
+        assertEquals("5", updatedSensor.captured.lastSentState)
+        assertEquals(AudioSensorManager.volMusic.statelessIcon, updatedSensor.captured.lastSentIcon)
+    }
```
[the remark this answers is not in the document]
blaubaer: Addressed.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
```diff
@@ -348,130 +355,115 @@ class AudioSensorManager : SensorManager {
...
+        if (sensor.id in sensorIdsWhereAttributesAreAlreadyForced) {
+            return true
         }
 
-        val volumeLevelDTMF = audioManager.getStreamVolume(AudioManager.STREAM_DTMF)
+        val sensorsWithAttributes = sensorDao(context)
+            .getFull(sensor.id)
+            .toSensorsWithAttributes()
+
+        if (!sensorsWithAttributes.isEmpty() && sensorsWithAttributes.all { sensorWithAttributes ->
```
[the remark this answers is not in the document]
blaubaer: Addressed

common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManagerTest.kt
```diff
@@ -0,0 +1,271 @@
...
+        coEvery { sensorDao.getFull(AudioSensorManager.volMusic.id) } returns mapOf(
+            volumeMusicSensor to listOf(
+                Attribute(AudioSensorManager.volMusic.id, "min", "2", "int"),
+                Attribute(AudioSensorManager.volMusic.id, "max", "15", "int"),
+            ),
+            volumeMusicSensorServer2 to listOf(
+                Attribute(AudioSensorManager.volMusic.id, "min", "2", "int"),
+                Attribute(AudioSensorManager.volMusic.id, "max", "10", "int"),
+            ),
+        )
```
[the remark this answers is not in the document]
blaubaer: Not a concern — `requestSensorUpdate` is sequential, so no concurrent access to the same sensor ID occurs. The worst case is a redundant DB query, which is harmless.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
```diff
@@ -348,130 +355,115 @@ class AudioSensorManager : SensorManager {
...
+     * values, or have already been force-updated during this app runtime.
+     *
+     * Usually, min and max do not change while the app is running. The only meaningful
+     * moments when they can be out of date:
+     * 1. An older companion app ran before, without pushing min/max.
+     * 2. The operating system changed.
+     *
+     * Therefore, the actual database check is performed at most once per sensor per app
+     * runtime. On subsequent calls for the same sensor, this returns `true` immediately
+     * without a database query.
```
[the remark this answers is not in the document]
blaubaer: Addressed.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
```diff
@@ -7,11 +7,18 @@ import android.os.Build
...
 import androidx.core.content.getSystemService
 import io.homeassistant.companion.android.common.R as commonR
+import io.homeassistant.companion.android.common.sensors.SensorManager.BasicSensor
 import io.homeassistant.companion.android.common.util.STATE_UNKNOWN
+import io.homeassistant.companion.android.database.sensor.toSensorsWithAttributes
+import java.util.concurrent.ConcurrentHashMap
 
 class AudioSensorManager : SensorManager {
+    private val sensorIdsWhereAttributesAreAlreadyForced: MutableSet<String> = ConcurrentHashMap.newKeySet()
+
```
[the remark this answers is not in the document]
blaubaer: Intentional. This is a unit test for the `.all` logic, not an integration test for the database schema. Mocking divergent attributes per server is the simplest way to verify the iteration behaves correctly.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
```diff
@@ -348,130 +355,116 @@ class AudioSensorManager : SensorManager {
...
+     * moments when they can be out of date:
+     * 1. An older companion app ran before, without pushing min/max.
+     * 2. The operating system changed.
+     *
+     * Once the stored attributes are found to match for a given sensor, the result is
+     * cached and subsequent calls return `true` immediately without a database query.
+     * If the attributes do not match, the database is queried again on each call until
+     * they do.
+     */
+    private suspend fun areAttributesOfSensorAlreadyForcedAtRuntimeOrMatching(
```
TimoPtr, line 437: So when the value changed on restart it does go in the `else` branch and then the second call is going to set the value in the map. So we are making two DB calls before stop making calls. I'm not sure we have a test that show this path.

common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManagerTest.kt
```diff
@@ -0,0 +1,271 @@
...
+        coJustRun { sensorDao.update(capture(updatedSensor)) }
+
+        sensorManager.requestSensorUpdate(context)
+
+        assertNull(updatedSensor.captured.lastSentState)
+        assertNull(updatedSensor.captured.lastSentIcon)
```
TimoPtr: To make the test a bit easier to read, create a small extension function on the sensor named `assertForceUpdate`

common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManagerTest.kt
```diff
@@ -0,0 +1,271 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
```
TimoPtr, line 32:
```suggestion
@RunWith(RobolectricTestRunner::class)
@Config(application = HiltTestApplication::class) 
```

This is important I've made this mandatory recently because otherwise it enables StrictMode in tests and makes the test fails in some configuration.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt
blaubaer: Correct — added a test that covers the full lifecycle: mismatch -> repeated DB queries -> attributes finally match -> cache kicks in -> no more DB calls.

common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManagerTest.kt
blaubaer: Addressed.

common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/AudioSensorManagerTest.kt
blaubaer: Addressed.

TimoPtr (approved):

TimoPtr:
@blaubaer You probably should add this information in the documentation https://companion.home-assistant.io/docs/core/sensors?_highlight=audio#audio-sensors otherwise no one will notice.


@jpelgrom I'm not sure about adding this one to the changelog wdyt?

blaubaer:
> @blaubaer You probably should add this information in the documentation https://companion.home-assistant.io/docs/core/sensors?_highlight=audio#audio-sensors otherwise no one will notice.

@TimoPtr I'll do once this is released. wdyt?

TimoPtr:
> > @blaubaer You probably should add this information in the documentation https://companion.home-assistant.io/docs/core/sensors?_highlight=audio#audio-sensors otherwise no one will notice.
> 
> @TimoPtr I'll do once this is released. wdyt?

We have a process for that, just open a PR with this and a beta flag and then we will handle the release :)

jpelgrom:
> @jpelgrom I'm not sure about adding this one to the changelog wdyt?

This seems a bit too minor to call out, in my opinion.

TimoPtr:
> > @jpelgrom I'm not sure about adding this one to the changelog wdyt?
> 
> This seems a bit too minor to call out, in my opinion.

That was my guess too but I wanted to be sure. Then @blaubaer the doc is enough.

PR #6456, opened 2026-02-18
Use lazy publication in ServerConnectionInfo to avoid any locks

TimoPtr:
## Summary

Since 2026.2.3 we have ANR that points to `isRegister`, it seems impossible that the simple parsing of a URL is blocking the Main thread, but the parsing of the URL is made within a `lazy` block and in its default mode `LazyThreadSafetyMode.SYNCHRONIZED` it encapsulate the parsing into a synchronized block which doesn't fit well into a coroutine since the coroutine is going to hang and not being suspended which is forbidden for the Main thread.

We are calling `isRegistered` in many places so it is possible that it cause the ANR, but it is not covering the whole story there and something should probably not be done in the Main thread. Merging the PR about widgets using Default dispatcher should help more.

This PR simply change the lazy mode to `LazyThreadSafetyMode.PUBLICATION`

> Initializer function can be called several times on concurrent access to an uninitialized Lazy instance value, but only one computed value will be used as the value of a Lazy instance and will be visible by all threads.

In the context it is perfectly fine since running the lazy blocks multiple times would always produce the same result and only cost some computations.

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
@@ -63,19 +63,19 @@ data class ServerConnectionInfo(
 ) {
     // Keep the parsing of the URLs from String to HttpUrl to avoid doing it multiple times
     @get:Ignore
-    internal val externalHttpUrl: HttpUrl? by lazy { externalUrl.toHttpUrlOrNull() }
+    internal val externalHttpUrl: HttpUrl? by lazy(LazyThreadSafetyMode.PUBLICATION) { externalUrl.toHttpUrlOrNull() }
 
     @get:Ignore
-    internal val internalHttpUrl: HttpUrl? by lazy { internalUrl?.toHttpUrlOrNull() }
+    internal val internalHttpUrl: HttpUrl? by lazy(LazyThreadSafetyMode.PUBLICATION) { internalUrl?.toHttpUrlOrNull() }
 
     @get:Ignore
-    internal val cloudHttpUrl: HttpUrl? by lazy { cloudUrl?.toHttpUrlOrNull() }
+    internal val cloudHttpUrl: HttpUrl? by lazy(LazyThreadSafetyMode.PUBLICATION) { cloudUrl?.toHttpUrlOrNull() }
 
     @get:Ignore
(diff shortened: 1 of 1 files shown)


TimoPtr:
@jpelgrom I'm not sure there is a publication somewhere about the risk of using `lazy` in coroutines.

jpelgrom:
> @jpelgrom I'm not sure there is a publication somewhere about the risk of using `lazy` in coroutines.

Do you have a trace that points to this as the issue? It seems like it might hit it but the operation inside the lazy block is simple enough that it shouldn't lock on it long enough to trigger an ANR. If that were the case there'd be ANRs all over the place for lazy usages.

TimoPtr:
> > @jpelgrom I'm not sure there is a publication somewhere about the risk of using `lazy` in coroutines.
> 
> Do you have a trace that points to this as the issue? It seems like it might hit it but the operation inside the lazy block is simple enough that it shouldn't lock on it long enough to trigger an ANR. If that were the case there'd be ANRs all over the place for lazy usages.

This is definitely not the cause of the ANR IMO. I just made the change because it is still better to never use a synchronized block within a coroutine. Especially in this specific case since it does run on Main on some cases so it put more pressure on the main thread even if it little. And yes I can send you the sentry issue, all of them point to the parsing of the url from when we call `isRegistered`.

I'm not expecting this PR to fix the issue, but to move the ANR somewhere else. And from what I see and where isRegistered is used I'm betting on the Widgets, where my pr of goAsync might help just because we changed the CoroutineScope out of the main thread, but then we might be hit by the goAsync ...

jpelgrom:
> I'm not expecting this PR to fix the issue, but to move the ANR somewhere else

Makes sense thinking more about it. OK to change it here since it's such a frequent use and might hide the cause, but we should avoid changing all `lazy { }` usages in the code base unless more evidence starts to emerge that it is something we should be really worried about.

jpelgrom (approved):

TimoPtr:
> > I'm not expecting this PR to fix the issue, but to move the ANR somewhere else
> 
> Makes sense thinking more about it. OK to change it here since it's such a frequent use and might hide the cause, but we should avoid changing all `lazy { }` usages in the code base unless more evidence starts to emerge that it is something we should be really worried about.

Agreed, I've made it here like this because this has been made for 'caching' and it is only used internally from the ConnectionInfo. I'm not expecting to replicate this in other places  except if it is clearly the culprit. 

But it is a good knowledge to have that lazy use synchronized block behind the scene.

Issue #6450, opened 2026-02-17
Very frequent background ANRs on 2026.2.3-full


marazmarci:
**Home Assistant Android app version(s):** 2026.2.3-full

**Android version(s):** 13

**Device model(s):** Samsung S20 FE 5G

**Home Assistant version:** 2025.12.4

**Last working Home Assistant release (if known):**

**Description of problem, include YAML if issue is related to notifications:**

Today morning I updated the HA app, and since then I'm getting ANR dialogs very frequently in the background. I'm currently on the go, and I'll provide logs later today. I'm also happy to help in diagnosing and fixing the problem. How can I send logs to the devs privately? Discord?

**Companion App Logs:**

```

```

**Screenshot or video of problem:**

**Additional information:**

TimoPtr:
Hello @marazmarci when do you get the ANR exactly? You can send me the logs on discord if you want but in theory (you can double checks) the logs doesn't contain sensitive information.

Also did you see that on a previous version? I want to pin point on which version it started.

TimoPtr:
It might me because of https://github.com/home-assistant/android/pull/6396 but this is just a guest.

jpelgrom:
> It might me because of [#6396](https://github.com/home-assistant/android/pull/6396) but this is just a guest.

If it points to that I'll hold off on reviewing #6415 to be sure, enough other PRs to review ;)

TimoPtr:
> > It might me because of [#6396](https://github.com/home-assistant/android/pull/6396) but this is just a guest.
> 
> If it points to that I'll hold off on reviewing #6415 to be sure, enough other PRs to review ;)

Or the issue is on the widgets 🤣. But yeah many other things to look at.

TimoPtr:
I think that the actual PR that used the `goAsync` #6396 is actually making the system more aggressive in terms of ANR especially on the new version of Android. It seems to have some kind of watchdog https://android.googlesource.com/platform/frameworks/base/+/android16-qpr2-release/core/java/android/content/BroadcastReceiver.java#447 that is triggering the ANR if we stay on the broadcast for more than 30s. 

@jpelgrom We might want to revert the PR, and consider moving to Workers to make sure we are not making like networks call that can timeout and trigger the ANR.

jpelgrom:
[...]

PR #6459, opened 2026-02-18
Revert "Move away from Main in Broadcast receiver and use goAsync"

TimoPtr:
This reverts commit 383a3ffe0a85aff006cae4194565b93af4de78e0.

## Summary

Revert #6396 to address #6450. We should instead migrate to workers to not use the broadcast receiver that should stay alive less than 30s according to the Android documentation.


jpelgrom:
Generally agree we should make sure not to exceed 30 seconds and use a worker otherwise, but I'd rather change it only in instances where we're likely to hit 30s only and still be able to use broadcast receiver with this goAsync helper otherwise.

I think OkHttp defaults to 30 seconds timeout so anything network related which does not fail and wait for timeout will hit the broadcast timeout by default, maybe we should change the default to something else (10 or 20 seconds?) instead.

TimoPtr:
> Generally agree we should make sure not to exceed 30 seconds and use a worker otherwise, but I'd rather change it only in instances where we're likely to hit 30s only and still be able to use broadcast receiver with this goAsync helper otherwise.
> 
> I think OkHttp defaults to 30 seconds timeout so anything network related which does not fail and wait for timeout will hit the broadcast timeout by default, maybe we should change the default to something else (10 or 20 seconds?) instead.

I would be more aggressive, anything that makes a network call should be made on a worker. No need to touch the timeout.

jpelgrom:
Workers can be deferred / are intended to be deferrable. Sometimes a broadcast needs to trigger something immediately so that isn't a good option.

PR #6460, opened 2026-02-18
Introduce NotificationDeleteWorker

TimoPtr:
## Summary

BroadcastReceiver should remain small since it is meant to have a short life we should migrate to the worker API. This is the first PR to showcase how we could do it for the NotificationDelete.

## Checklist

## Any other notes

It is based on #6459

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/NotificationDeleteWorker.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/NotificationDeleteWorker.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import androidx.work.CoroutineWorker
 import androidx.work.Data
 import androidx.work.OneTimeWorkRequestBuilder
+import androidx.work.OutOfQuotaPolicy
 import androidx.work.WorkManager
 import androidx.work.WorkerParameters
 import dagger.hilt.EntryPoint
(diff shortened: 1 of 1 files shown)


jpelgrom:
Following from my comment https://github.com/home-assistant/android/pull/6459#issuecomment-3923083049:

I think this instance should be adjusted to avoid hitting 30s in the first place. WorkManager is for _deferrable_ tasks - events like these should fire immediately.

TimoPtr:
> Following from my comment https://github.com/home-assistant/android/pull/6459#issuecomment-3923083049:
> 
> I think this instance should be adjusted to avoid hitting 30s in the first place. WorkManager is for _deferrable_ tasks - events like these should fire immediately.

The issue is that even if we reduce the timeout, if I'm not wrong frireEvent might try the first Url fail and try the second one if there are any and even maybe a third one? Internal/cloud/external?

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/NotificationDeleteWorker.kt
```diff
@@ -0,0 +1,93 @@
...
+            val data = Data.Builder()
+                .putLong(KEY_DATABASE_ID, databaseId)
+                .putStringArray(KEY_EVENT_DATA_KEYS, eventDataKeys)
+                .putStringArray(KEY_EVENT_DATA_VALUES, eventDataValues)
+                .build()
+
+            val request = OneTimeWorkRequestBuilder<NotificationDeleteWorker>()
+                .setInputData(data)
+                // We want the event to be sent right away if it is possible
+                .setExpedited(OutOfQuotaPolicy.RUN_AS_NON_EXPEDITED_WORK_REQUEST)
```
TimoPtr, line 64: > Marks the WorkRequest as important to the user. In this case, WorkManager provides an additional signal to the OS that this work is important.
Note that although the execution time of this work won't be counted against your app's quota while your app is in the foreground, if the expedited work continues in the background, you are susceptible to quota. However, power management restrictions, such as Battery Saver and Doze, are less likely to affect expedited work. Because of this, expedited work is best suited for short tasks which need to start immediately and are important to the user or user-initiated.

TimoPtr:
@jpelgrom I don't know yet what to do for these ANR I'm not sure there is a "right" way

PR #6467, opened 2026-02-19
Make HighAccuracyLocationReceiver async

TimoPtr:
## Summary

Follow up after #6459 to try to improve the situation toward our background ANRs, this PR adds back `launchAsync` but with a timeout. 
This apply to the `HighAccuracyLocationReceiver` since the action itself is going to fit under 30s since it is a simple DB call.

It is also a PR to open up discussion about how to reports when we reach the timeout, because if it ever happens with this PR it is going to silently fail on release (crash on debug). My proposal is to add a hook in FailFast to report to Sentry when we are entering FailFast. The risk would be to capture some information about a user in the fail.

## Checklist

### DIFF
--- a/app/src/full/kotlin/io/homeassistant/companion/android/location/HighAccuracyLocationReceiver.kt
+++ b/app/src/full/kotlin/io/homeassistant/companion/android/location/HighAccuracyLocationReceiver.kt
@@ -8,7 +8,6 @@ import io.homeassistant.companion.android.sensors.LocationSensorManager
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.launch
 
 class HighAccuracyLocationReceiver : BroadcastReceiver() {
     companion object {


common/src/main/kotlin/io/homeassistant/companion/android/common/util/BroadcastReceiverExt.kt
```diff
@@ -0,0 +1,47 @@
...
+ * @param block The suspend function to execute within the 30-second timeout.
+ */
+fun BroadcastReceiver.launchAsync(coroutineScope: CoroutineScope, block: suspend CoroutineScope.() -> Unit) {
+    val pendingResult = goAsync()
+
+    coroutineScope.launch {
+        try {
+            withTimeout(30.seconds) { block() }
+        } catch (e: TimeoutCancellationException) {
+            FailFast.fail { "BroadcastReceiver (${this@launchAsync.javaClass}) exceeded the 30-second timeout" }
```
TimoPtr, line 54: I added the name of the broadcast receiver otherwise we are going to loose the information in the logs of where it's coming from

Image

jpelgrom (approved):
I feel like we already discussed this on a previous PR. Good approach where we can reasonably expect execution to stay within 30 seconds. Tests look good, and tested the receiver in real life and works as expected.

When something may exceed 30 seconds due to networking/long running operations, we should carefully consider how to handle it (launchAsync and ignore, timeouts inside, worker, ...) to avoid introducing regressions for executions where it does not exceed 30 seconds (like the happy path).

(Side note: official documentation mentioning "allowing them to execute for 30 seconds _or even a bit more"_ for the timeout made me smile :) )

TimoPtr:
> I feel like we already discussed this on a previous PR. Good approach where we can reasonably expect execution to stay within 30 seconds. Tests look good, and tested the receiver in real life and works as expected.
> 
> When something may exceed 30 seconds due to networking/long running operations, we should carefully consider how to handle it (launchAsync and ignore, timeouts inside, worker, ...) to avoid introducing regressions for executions where it does not exceed 30 seconds (like the happy path).
> 
> (Side note: official documentation mentioning "allowing them to execute for 30 seconds _or even a bit more"_ for the timeout made me smile :) )

I agree that's why I'm moving in small batch from now on on this mater to make sure I'm not missing something with the timeout. 

I still think we could leverage Sentry here, I would like to hear your opinion about plugging FailFast to Sentry to reports these error without crashing.

Issue #6472, opened 2026-02-22
Microwakeword Suggestions, Requests, and Issues


SmoochSmackers:
I'm really not sure if this is a feature request or bug report. A little bit of both? Please move if necessary.

I've been testing out the recent release of the Android app's native wake word. Overall it's been great, it's only using about 15% of my battery, which isn't great, but is *far* better than the 25-50% of some of the 3rd party solutions I've tried in the past, and I am *unbelievably* grateful for the work ya'll have done to bring this feature to the app. 

That being said, I have a few suggestions:

1. **As silly as it sounds, the number one thing I *really* want here is the ability to implement a custom wake word. I'd love to be able to put a .tflite file (or analogue) onto my phone storage to use as the voice model (similar to the OpenWakeWord integration on the HA server)**. 
I've been chasing this ever since the ability to do so got stripped at some point after the galaxy s2 (or s4, I can't remember) era, I guess roughly when Google Assistant first came out and they decided they just *had* to have their branding in the wakeword for the rest of eternity.

2. **After the mobile app's transcript overlay pops up, the wake word no longer functions. I suspect the assistant pipeline (represented by the overlay) keeps control of the mic away from the wake word pipeline? It'd be nice to be able to use the wake word while the overlay is being displayed.**
Rhasspy Mobile had a [somewhat relevant issue](https://github.com/Nailik/rhasspy_mobile/issues/557) with lack of ceding microphone control, which the developer was able to fix. As awesome and capable as Rhasspy Mobile is, I found it would periodically *eat* my battery and had to stop using it. I'm unsure if that unfortunate battery consumption had anything to do with the microphone control ceding fix.

3.  **I currently have the app set to use "Okay Nabu" and have gotten quite a few false positives. A sensitivity dial in the options would be cool.**

Thank you all so much for getting this feature implemented, your hard work is truly appreciated!

**Home Assistant Android app version(s):**
2026.2.3-full (beta version)

**Android version(s):**
16 (Build BP2A.250605.031.A3.S908USQS8GZA1)

**Device model(s):**
Samsung Galaxy S22 Ultra (SM-S908U)

**Home Assistant version:**
Install method - Home Assistant OS
Core - 2026.2.3
Supervisor - 2026.02.2
Operating System - 17.1
Frontend - 20260128.6

**Last working Home Assistant release (if known):**
N/A

**Description of problem, include YAML if issue is related to notifications:**
N/A

**Companion App Logs:**
N/A
```

```

[...]

PR #6480, opened 2026-02-23
Allow wake word detection on Android <10

TimoPtr:
## Summary

It seems that bellow Android 10 Assist wake word detection is not working nicely since we cannot share properly the audio recording. This change is pausing the audio recording when entering the AssistActivity and resume when leaving it.

There was a crash also when we call finish from the onShow callback, it was not able to start the activity properly if we do so. A hack is to simply postpone the finish with a post.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsViewModel.kt
@@ -108,7 +108,13 @@ class AssistSettingsViewModel @Inject constructor(
     fun onSelectWakeWordModel(model: MicroWakeWordModelConfig) {
         viewModelScope.launch {
             assistConfigManager.setSelectedWakeWordModel(model)
-            _uiState.update { it.copy(selectedWakeWordModel = model) }
+            _uiState.update {
+                it.copy(
+                    selectedWakeWordModel = model,
+                    isTestingWakeWord = false,
+                    wakeWordDetected = false,
+                )
+            }
         }
     }
 


TimoPtr:
Test wake word on bellow <31 is not working properly but I kept it in the AssistSettings since it would simply open Assist instead of showing the result text (after 31 it does show the text and Assist).

app/src/main/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionService.kt
```diff
@@ -172,7 +184,13 @@ class AssistVoiceInteractionService : VoiceInteractionService() {
 
         lastTriggerTime = now
         Timber.i("Wake word '${model.wakeWord}' detected, launching Assist")
-        launchAssist(wakeWord = model.wakeWord)
+
+        // Stop the listener before launching Assist to release the microphone.
+        // On Android 8.1 and below, only one component can hold an AudioRecord at a time.
```
jpelgrom: Title: 31 is Android 12 so the issue would be for Android <= 11. This comments suggests <= 8.1 (<= 27). The most obvious behavior change I can find refers to differences in behavior [starting with Android 10](https://developer.android.com/media/platform/sharing-audio-input).

Which version is it and can you provide a source, or is this from manual testing?

jpelgrom:
This seems somewhat related to #6472 point 2, but in reverse? Now we're explicitly disabling wake word while the main Assist dialog is shown.

app/src/main/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionService.kt
TimoPtr: Manual testing on Android 9 and Claude telling me that before API 31 it cannot be shared. I could probably find the exact version but I'm fine aligning the behavior for all the versions or you think we should have a different one?

TimoPtr:
> This seems somewhat related to #6472 point 2, but in reverse? Now we're explicitly disabling wake word while the main Assist dialog is shown.

Somehow yes today we can't use wake word detection while assist is running because the audio recorder is not getting any data so this change is just more explicit.

app/src/main/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionService.kt
jpelgrom: No same for all versions like you changed it to now is fine, the different versions mentioned was just confusing.

jpelgrom:
Because testing the wake word is now handled by the service instead of the viewmodel, it actually triggers Assist when testing is succesfull. That is a good way to know it worked I suppose but also obscures the testing view and disables the wake word again until closed...

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsViewModel.kt
jpelgrom, line 100: This may have been here before, but I noticed that toggling didn't stop testing which is weird.

```suggestion
            _uiState.update {
                it.copy(
                    isWakeWordEnabled = enabled,
                    isTestingWakeWord = false,
                    wakeWordDetected = false,
                )
            }
```

TimoPtr:
> Because testing the wake word is now handled by the service instead of the viewmodel, it actually triggers Assist when testing is succesfull. That is a good way to know it worked I suppose but also obscures the testing view and disables the wake word again until closed...

I was debating with this, at least it shows it works I think it is a good way for ppl to actually test it. On a larger screen you do see the text above the bottom sheet.

jpelgrom:
> > Because testing the wake word is now handled by the service instead of the viewmodel, it actually triggers Assist when testing is succesfull. That is a good way to know it worked I suppose but also obscures the testing view and disables the wake word again until closed...
> 
> I was debating with this, at least it shows it works I think it is a good way for ppl to actually test it. On a larger screen you do see the text above the bottom sheet.

If you look before it opens Assist you can also see the "Wake word detected" text on a smaller screen. Just wanted to make sure it was an intentional change, I'm also fine with it.

PR #6484, opened 2026-02-24
Fix change server button not showing with Auto(motive) favorites

jpelgrom:
## Summary

Fixes an issue where the 'Change server' button was not visible in the Android Auto(motive) interface when favorites were used.

This issue was caused by the changes in commit 8daf0857d5676b7134d15debae087823f9548b0a moving the server count check to be dependent on the lifecycle in `EntityGridVehicleScreen`. For favorites, a function in this screen is called statically so the lifecycle is never started, and the server count will never be updated.

This PR fixes the issue by adjusting `MainVehicleScreen`  to always be responsible for showing the 'Change server' button. The related code is removed from `EntityGridVehicleScreen, and the function is passed info on whether or not it will be shown to the list so the maximum item count can be adjusted accordingly.

## Checklist

## Screenshots

Android Auto interface for Home Assistant, showing a grid with buttons labeled 'Alarm sound (unavailable)', 'Navigation', 'All entities' and 'Change server'

n/a

## Any other notes

It's probably worth untangling more in these screens and seeing if we can add some kind of tests, but that is out of scope for now I think. This is a quick bug fix that can be included in the next release.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/util/vehicle/TemplateComponents.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/util/vehicle/TemplateComponents.kt
@@ -204,7 +204,7 @@ fun getDomainList(
                                     domains,
                                     entityList,
                                     allEntities,
-                                ) { },
+                                ),
                             )
                         }
                     }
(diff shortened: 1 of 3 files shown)


TimoPtr (approved):
Good catch 👍🏻

PR #6489, opened 2026-02-25
Warn users when installing full on a device without GMS

TimoPtr:
## Summary

We do have some users that install the full flavor on a device without GMS because they don't know what full and minimal means from Github. This PR should help reduce the numbers of crashes we have because users are installing the wrong app.

## Checklist

## Screenshots

image

### DIFF
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -36,7 +36,7 @@
     <string name="always_show_first_view_on_app_start">Always show first view when opening app</string>
     <string name="always_show_first_view_on_app_start_summary">The first view of the default dashboard is shown as soon as the app is opened</string>
     <string name="app_name">Home Assistant</string>
-    <string name="play_services_unavailable_full_flavor">Google Play services are unavailable. This full version may crash. Install the minimal version instead.</string>
+    <string name="play_services_unavailable_full_flavor">Google Play services are unavailable. This version of the app may crash. Use the minimal version instead.</string>
     <string name="app_version_info">App version info</string>
     <string name="application_version">Application version</string>
     <string name="attempting_authentication">Authenticating…</string>


app/lint.xml
jpelgrom, line 1: This file seems unrelated

common/lint.xml
```diff
@@ -3,4 +3,10 @@
     <issue id="ComposeCompositionLocalUsage">
         <option name="allowed-composition-locals" value="LocalHAColorScheme" />
     </issue>
+    <issue id="ComposePreviewPublic">
+        <ignore path="src/screenshotTest" />
+    </issue>
+    <issue id="ComposeUnstableReceiver">
+        <ignore path="src/screenshotTest" />
+    </issue>
```
jpelgrom, line 11: This seems unrelated

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchActivity.kt
```diff
@@ -92,14 +102,31 @@ class LaunchActivity : AppCompatActivity() {
...
                 val uiState by viewModel.uiState.collectAsStateWithLifecycle()
+                val snackbarHostState = remember { SnackbarHostState() }
+                val shouldShowPlayServicesSnackbar = remember {
+                    BuildConfig.FLAVOR == "full" && !playServicesAvailability.isAvailable()
+                }
+
+                if (shouldShowPlayServicesSnackbar) {
+                    val message = stringResource(commonR.string.play_services_unavailable_full_flavor)
+                    LaunchedEffect(message) {
+                        snackbarHostState.showSnackbar(
```
jpelgrom: I wonder if maybe it's a good idea to add an action 'Learn more' linking to the [flavors documentation](https://companion.home-assistant.io/docs/core/android-flavors). In there we could expand a bit on why we suggest using minimal if there are no Play Services installed.

app/lint.xml
TimoPtr: It is. I thought I kept it on my stash. The lastest version of AS do fix the screenshot test sourceset finaly and there was just these annoying lint issues that I disabled.

I can move it to a separate pr if you prefer but it's small.

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchActivity.kt
TimoPtr: Why not 👍

app/lint.xml
jpelgrom: OK to keep for now. Just be mindful of sneaking in changes like these, I have Android Studio updating kotlinc.xml every time but don't commit it ;).

TimoPtr: Yeah we need to commit this kotlinc changes...

common/src/main/res/values/strings.xml
```diff
@@ -36,6 +36,7 @@
     <string name="always_show_first_view_on_app_start">Always show first view when opening app</string>
     <string name="always_show_first_view_on_app_start_summary">The first view of the default dashboard is shown as soon as the app is opened</string>
     <string name="app_name">Home Assistant</string>
+    <string name="play_services_unavailable_full_flavor">Google Play services are unavailable. This full version may crash. Install the minimal version instead.</string>
```
jpelgrom: "This full version" is a bit unclear if you don't know about flavors yet.

```suggestion
    <string name="play_services_unavailable_full_flavor">Google Play services are unavailable. This version of the app may crash. Use the minimal version instead.</string>
```

PR #6497, opened 2026-02-25
Wake word detection within Assist conversation

TimoPtr:
## Summary

This enables AssistActivity to supports wake word to continue the current conversation. It is only enabled when starting Assist with a wake word (it continue to use this wake word). 

It also send the wake word each time we start a new pipeline with it, it means that if two phones are close it could close the dialog to open a new one on the other phone. But we don't care since it doesn't have any memory each pipeline is independent and scope to one run (one answer).

I've come up with a Strategy pattern to try to isolate a bit the logic and creating it using a factory.

I've shared the whole voice recording into a common code using a Flow so that we can record and do the wake word detection. 

When starting the activity with a wake word the mic is always ON but the recording is only made when requested by the logic in Assist.

I didn't request focus since it should already have it for the wake word detection, but I might be wrong. I did test using Android 9 and 16 and it works on both correctly.

## Checklist

## Any other notes

I wonder if we could actually share the VoiceRecorder between the service and the Activity, I don't think it is possible because we start the activity in a new task but maybe I didn't try.

### DIFF
--- a/app/src/test/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionServiceTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionServiceTest.kt
@@ -24,15 +24,12 @@ import io.mockk.mockk
 import io.mockk.slot
 import io.mockk.unmockkAll
 import io.mockk.verify
-import kotlin.time.Instant
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.test.advanceUntilIdle
 import kotlinx.coroutines.test.runTest
 import org.junit.After
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNotNull
-import org.junit.Assert.assertNotSame
-import org.junit.Assert.assertSame
 import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Rule
@@ -289,38 +286,6 @@ class AssistVoiceInteractionServiceTest {
     }
 
     @Test
-    fun `Given wake word detected twice quickly when callback invoked then debounce`() = runTest {
-        val shadow = Shadows.shadowOf(service) as ShadowVoiceInteractionService
-        clock.currentInstant = Instant.fromEpochMilliseconds(0)
-        coEvery { assistConfigManager.getSelectedWakeWordModel() } returns microWakeWordModelConfigs[0]
-
-        // Call onReady() to make showSession() available (isWakeWordEnabled defaults to false)
-        service.onReady()
-        advanceUntilIdle()
-
-        sendAction(ACTION_START_LISTENING)
-        advanceUntilIdle()
-
-        // First detection - should trigger showSession
-        onWakeWordDetectedSlot.captured.invoke(microWakeWordModelConfigs[0])
-        advanceUntilIdle()
-        val firstBundle = shadow.lastSessionBundle
-        assertNotNull(firstBundle)
-
-        // Second detection 2 seconds later (within 3 second debounce) - should be ignored
-        clock.currentInstant = Instant.fromEpochMilliseconds(2000)
-        onWakeWordDetectedSlot.captured.invoke(microWakeWordModelConfigs[0])
-        advanceUntilIdle()
-        assertSame(firstBundle, shadow.lastSessionBundle)
-
-        // Third detection 4 seconds after first (after debounce) - should trigger showSession
-        clock.currentInstant = Instant.fromEpochMilliseconds(4000)
-        onWakeWordDetectedSlot.captured.invoke(microWakeWordModelConfigs[0])
-        advanceUntilIdle()
-        assertNotSame(firstBundle, shadow.lastSessionBundle)
-    }
-
-    @Test
     fun `Given wake word detected when callback invoked then send broadcast`() = runTest {
         coEvery { assistConfigManager.getSelectedWakeWordModel() } returns microWakeWordModelConfigs[0]
 
@@ -340,35 +305,6 @@ class AssistVoiceInteractionServiceTest {
         }
         assertNotNull(wakeWordBroadcast)
         assertEquals(service.packageName, wakeWordBroadcast!!.`package`)
-    }
-
-    @Test
-    fun `Given wake word detected within debounce when callback invoked then still send broadcast`() = runTest {
-        clock.currentInstant = Instant.fromEpochMilliseconds(0)
-        coEvery { assistConfigManager.getSelectedWakeWordModel() } returns microWakeWordModelConfigs[0]
-
-        service.onReady()
-        advanceUntilIdle()
-
-        sendAction(ACTION_START_LISTENING)
-        advanceUntilIdle()
-
-        // First detection
-        onWakeWordDetectedSlot.captured.invoke(microWakeWordModelConfigs[0])
-        advanceUntilIdle()
-
-        // Second detection within debounce (showSession suppressed, but broadcast should still fire)
-        clock.currentInstant = Instant.fromEpochMilliseconds(2000)
-        onWakeWordDetectedSlot.captured.invoke(microWakeWordModelConfigs[0])
-        advanceUntilIdle()
-
-        val broadcastIntents = Shadows.shadowOf(ApplicationProvider.getApplicationContext<android.app.Application>())
-            .broadcastIntents
-        val wakeWordBroadcasts = broadcastIntents.filter {
-            it.action == "io.homeassistant.companion.android.WAKE_WORD_DETECTED"
-        }
-        // Both detections should have sent a broadcast, even though the second was debounced
-        assertEquals(2, wakeWordBroadcasts.size)
     }
 
     @Test


app/src/main/kotlin/io/homeassistant/companion/android/assist/wakeword/WakeWordListener.kt
```diff
@@ -126,83 +131,38 @@ class WakeWordListener(
...
-        readResult: Int,
-    ) {
-        val detected = microWakeWord.processAudio(buffer.copyOf(readResult))
-        if (!detected) return
+    ): Boolean {
+        val detected = microWakeWord.processAudio(buffer)
+        if (!detected) return false
 
         // Reset microWakeWord state to prevent immediate re-triggering
         microWakeWord.reset()
```
TimoPtr, line 169: Even with the reset I saw it was triggered multiple times in the `AssistActivity` because there was no debounce. I decided to move the debounce from the service to the listener with more precise computation. So the debounce is shorter now.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/VoiceAudioRecorder.kt
```diff
@@ -0,0 +1,151 @@
...
+     * contains one chunk of PCM 16-bit mono samples.
+     *
+     * The flow is shared via [shareIn] with [SharingStarted.WhileSubscribed]:
+     * - First collector starts the [AudioRecord] and begins emitting
+     * - Additional collectors receive the same audio stream
+     * - When the last collector cancels, [AudioRecord] stops immediately
+     */
+    suspend fun audioData(): Flow<ShortArray> = mutex.withLock {
+        sharedFlow ?: createSharedFlow().also { sharedFlow = it }
+    }
```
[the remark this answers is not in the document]
TimoPtr: It would restart the flow

common/src/test/kotlin/io/homeassistant/companion/android/common/util/VoiceAudioRecorderTest.kt
```diff
@@ -0,0 +1,293 @@
...
+                    // Simulate the blocking nature of a real AudioRecord.read() call.
+                    // Without this, the mock returns instantly creating a tight
+                    // non-suspending loop that prevents cancellation from propagating.
+                    Thread.sleep(1)
+                    val buffer = firstArg<ShortArray>()
+                    shortArrayOf(42).copyInto(buffer)
+                    1
+                }
+            }
+            val sharingScope = TestScope(UnconfinedTestDispatcher())
```
TimoPtr: This is flaky I don't like it, I need to find another solution

BjornPython:
very excited for this to be merged, and my home assistant wallpanel's assist work as it is intended. Thank you for this!

Th3Rom3:
Thank you so much for the swift implementation. I have some older phones with LineageOS that would be perfect for simple voice assist satellites.

The base wake word functionality works great, having to close the assist panel manually is the only thing that keeps it from working as an alternative hands-free assist device.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsViewModel.kt
```diff
@@ -107,7 +107,15 @@ class AssistSettingsViewModel @Inject internal constructor(
...
             assistConfigManager.setWakeWordEnabled(enabled)
-            _uiState.update { it.copy(isWakeWordEnabled = enabled) }
+            // setWakeWordEnabled could set a model so we need to get the selected model and update the UI with it
+            val model = assistConfigManager.getSelectedWakeWordModel()
+            _uiState.update {
+                if (enabled) {
+                    it.copy(isWakeWordEnabled = true, selectedWakeWordModel = model)
+                } else {
+                    it.copy(isWakeWordEnabled = false)
+                }
```
jpelgrom: Why complicate the update with an if/else?


```suggestion
                it.copy(isWakeWordEnabled = enabled, selectedWakeWordModel = model)
```

app/src/test/kotlin/io/homeassistant/companion/android/assist/AssistAudioStrategyFactoryTest.kt
```diff
@@ -0,0 +1,60 @@
...
+        }
+        factory = AssistAudioStrategyFactory(
+            voiceAudioRecorder = voiceAudioRecorder,
+            wakeWordListenerFactory = wakeWordListenerFactory,
+            assistConfigManager = assistConfigManager,
+        )
+    }
+
+    @Nested
+    inner class `Given no wake word phrase` {
```
jpelgrom: New test pattern? Why create inner classes which have only one test?

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistConfigManager.kt
```diff
@@ -77,6 +79,11 @@ class AssistConfigManagerImpl @Inject constructor(
     override suspend fun setWakeWordEnabled(enabled: Boolean) {
         prefsRepository.setWakeWordEnabled(enabled)
         if (enabled) {
+            if (getSelectedWakeWordModel() == null) {
+                models.get().firstOrNull()?.let {
```
jpelgrom: Instead of first should it prefer "Okay Nabu" if available? (It may be expected to be the first, but for future-proofing)

app/src/main/kotlin/io/homeassistant/companion/android/assist/WakeWordAssistAudioStrategy.kt
```diff
@@ -0,0 +1,98 @@
...
+        wakeWordChannel.consumeEach {
+            trySend(it)
+        }
+    }
+
+    /** No-op */
+    override fun requestFocus() {}
+
+    /** No-op */
+    override fun abandonFocus() {}
```
jpelgrom: Why are these no-ops? Even if I have a wake word I'd want the focus once Assist has actually been triggered.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/VoiceAudioRecorder.kt
```diff
@@ -0,0 +1,150 @@
...
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import timber.log.Timber
+
+/**
+ * Sample rate in Hz used by all voice pipelines and wake word detection.
+ * Anything above would get downsample to 16kHz in core, so we don't need
+ * to capture at a higher frequency.
+ * */
+const val VOICE_SAMPLE_RATE = 16000
```
jpelgrom, line 27: What about the comment you removed? Is it no longer current or did you find another source pointing to better compatibility?

```
// Docs: 'currently the only rate that is guaranteed to work on all devices'
```

common/src/main/kotlin/io/homeassistant/companion/android/common/util/VoiceAudioRecorder.kt
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * Multiple collectors share the same underlying [AudioRecord] instance. The first
+ * collector triggers [AudioRecord] creation and starts recording; additional collectors
+ * receive the same audio stream. When the last collector cancels, recording stops
+ * and the [AudioRecord] is released immediately.
+ *
+ * This sharing is critical on pre-Android 10 devices where only one [AudioRecord]
+ * can be active at a time. Both the Assist pipeline and wake word detection can
+ * safely collect from the same [VoiceAudioRecorder] concurrently.
+ *
```
jpelgrom:
```suggestion
```

app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistViewModel.kt
```diff
@@ -257,7 +275,9 @@ class AssistViewModel @Inject constructor(
             if (hasMicrophone && it.sttEngine != null) {
                 if (recorderAutoStart && (hasPermission || requestSilently)) {
                     inputMode = AssistInputMode.VOICE_INACTIVE
-                    onMicrophoneInput(proactive = null)
+                    if (recorderProactive) {
```
jpelgrom: Why this change? The function still needs to be called to go from inactive -> active (in UI) when recording should automatically start

app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistViewModel.kt
```diff
@@ -320,24 +340,14 @@ class AssistViewModel @Inject constructor(
 
         stopPlayback()
 
-        val recording = try {
-            recorderProactive || audioRecorder.startRecording()
-        } catch (e: Exception) {
-            Timber.e(e, "Exception while starting recording")
-            false
```
jpelgrom, line 386: How do you want to do error handling? There is no check in `setupRecorder` to show an error message.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistConfigManager.kt
TimoPtr: Actually the first today is Jarvis.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/VoiceAudioRecorder.kt
TimoPtr: Indeed I didn't check the doc, thinking that a lower frequency would work ...
> he sample rate expressed in Hertz. 44100Hz is currently the only rate that is guaranteed to work on all devices, but other rates such as 22050, 16000, and 11025 may work on some devices. AudioFormat.SAMPLE_RATE_UNSPECIFIED means to use a route-dependent value which is usually the sample rate of the source
This is annoying because the wake word detection does work at 16kHz which is hardcoded this way today in InteractionService.

app/src/main/kotlin/io/homeassistant/companion/android/assist/WakeWordAssistAudioStrategy.kt
TimoPtr: Well I'm going to add it back, but I was not sure if it was needed since you are constantly listening already in the back.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/VoiceAudioRecorder.kt
TimoPtr: I've made a version where I try to init at 16kHz and if it fails I downsample from 44.1 to 16 (I used Claude to write the downsampling and tests). I forced the downsampling to test and the wake word detection continues to work.

app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistViewModel.kt
TimoPtr: I'm not sure how to handle it properly now. The start of the recording is made directly in the base viewModel while we collect `audioData()`, if today the start recording throw then it crashes the app. I wonder when this can happen? 

The issue is that the start of the recording is now async so I would need to pass to setupRecorder a lambda to handle the error. Something like 
```
setupRecorder(onError = {
                _conversation.add(
                    AssistMessage(app.getString(commonR.string.assist_error), isInput = false, isError = true),
                ))
            })
```

TimoPtr: I did something similar try it. I'm not a big fan.

jpelgrom: This approach seems somewhat reasonable to me? Maybe it doesn't happen often but still good to indicate something if we cannot do audio (for example if someone has a weird wall display with no microphone).

Which part are you not a big fan of?

app/src/main/kotlin/io/homeassistant/companion/android/assist/WakeWordAssistAudioStrategy.kt
jpelgrom: For wake word focus is not wanted but once it's triggered I think we can assume the user wants to be heard. Similar to smart speakers which can play music but once you trigger the assistant they pause playback for your command.

jpelgrom:
There are merge conflicts here as well.

app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistViewModel.kt
TimoPtr: To test this I've made the strategy throw and it's kinda looking weird since the placeholder is updating after the error message, but like you said it shouldn't happen and if it does it might not be like how I'm testing. I'm going to keep it like this for now.

TimoPtr:
> There are merge conflicts here as well.

Addressed

PR #6501, opened 2026-02-25
Fix EntityWidget showing "null null" instead of last known value

ollide:
## Summary

When an `EntityWidget` is configured with an additional attribute (e.g., `unit_of_measurement`), the widget displayed "**null null**" whenever the Home Assistant server is unreachable, instead of showing the last known value with an error indicator.

 Three issues compound in `resolveTextToShow()`:
 1. Kotlin's String?.plus() coerces null to the literal string "null", so entity?.friendlyState(...) produces "null" when the server is unreachable and entity is null.
 2. listOf<String?>(null).joinToString() renders missing attribute values as "null" rather than skipping them.
 3. This corrupted "null null" string is persisted to the last_update cache column, overwriting the last valid value.

Fixes:
 - Added an early return when `entity == null`, falling back to the cached lastUpdate and preserving the entityCaughtException flag so the red error icon appears.
 - Changed map { ... } to mapNotNull { ... } when collecting attribute values, so attributes missing from the entity's current state are skipped rather than rendered as "null".
 - Remove DAO inserts when no new state was fetched

## Checklist

## Screenshots

Before:
before
After:
after

## Any other notes

This PR is my first contribution to HA. I’ve been a long-time Home Assistant user and am very grateful for the project. Please feel free to let me know if you’d like any changes.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidget.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidget.kt
@@ -37,7 +37,7 @@ class EntityWidget : BaseWidgetProvider<StaticWidgetEntity, StaticWidgetDao>() {
         internal const val TOGGLE_ENTITY =
             "io.homeassistant.companion.android.widgets.entity.EntityWidget.TOGGLE_ENTITY"
 
-        private data class ResolvedText(val text: CharSequence?, val exception: Boolean = false)
+        private data class ResolvedText(val text: CharSequence?, val error: Boolean = false)
     }
 
     override fun getWidgetProvider(context: Context): ComponentName = ComponentName(context, EntityWidget::class.java)
@@ -120,7 +120,7 @@ class EntityWidget : BaseWidgetProvider<StaticWidgetEntity, StaticWidgetDao>() {
                 )
                 setViewVisibility(
                     R.id.widgetStaticError,
-                    if (resolvedText.exception) View.VISIBLE else View.GONE,
+                    if (resolvedText.error) View.VISIBLE else View.GONE,
                 )
                 setOnClickPendingIntent(
                     R.id.widgetTextLayout,
@@ -154,7 +154,6 @@ class EntityWidget : BaseWidgetProvider<StaticWidgetEntity, StaticWidgetDao>() {
         appWidgetId: Int,
     ): ResolvedText {
         var entity: Entity? = null
-        var entityCaughtException = false
         try {
             entity = if (suggestedEntity != null && suggestedEntity.entityId == entityId) {
                 suggestedEntity
@@ -165,7 +164,6 @@ class EntityWidget : BaseWidgetProvider<StaticWidgetEntity, StaticWidgetDao>() {
             throw e
         } catch (e: Exception) {
             Timber.e(e, "Unable to fetch entity")
-            entityCaughtException = true
         }
         val entityOptions = if (
             entity?.canSupportPrecision() == true &&
@@ -184,7 +182,7 @@ class EntityWidget : BaseWidgetProvider<StaticWidgetEntity, StaticWidgetDao>() {
         }
 
         if (entity == null) {
-            return ResolvedText(dao.get(appWidgetId)?.lastUpdate, entityCaughtException)
+            return ResolvedText(dao.get(appWidgetId)?.lastUpdate, true)
         }
 
         if (attributeIds == null) {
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidget.kt
```diff
@@ -182,20 +182,26 @@ class EntityWidget : BaseWidgetProvider<StaticWidgetEntity, StaticWidgetDao>() {
...
+            return ResolvedText(lastUpdate)
         }
 
         try {
-            val fetchedAttributes = entity?.attributes as? Map<*, *> ?: mapOf<String, String>()
+            val fetchedAttributes = entity.attributes as? Map<*, *> ?: mapOf<String, String>()
             val attributeValues =
-                attributeIds.split(",").map { id -> fetchedAttributes[id]?.toString() }
+                attributeIds.split(",").mapNotNull { id -> fetchedAttributes[id]?.toString() }
             val lastUpdate =
```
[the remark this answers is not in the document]
ollide: I tried to keep the scope of the changes as small as possible; this was already the case before my changes.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidget.kt
```diff
@@ -182,20 +182,26 @@ class EntityWidget : BaseWidgetProvider<StaticWidgetEntity, StaticWidgetDao>() {
         } else {
             null
         }
+
+        if (entity == null) {
+            return ResolvedText(dao.get(appWidgetId)?.lastUpdate, entityCaughtException)
+        }
```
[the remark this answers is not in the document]
ollide: I tried to keep the scope of the changes as small as possible; extra DB reads in `resolveTextToShow()` already exist.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetTest.kt
```diff
@@ -0,0 +1,163 @@
...
+        attributes: Map<String, Any?> = emptyMap(),
+    ) = Entity(
+        entityId = entityId,
+        state = state,
+        attributes = attributes,
+        lastChanged = LocalDateTime.now(),
+        lastUpdated = LocalDateTime.now(),
+    )
+
+    private fun resolveWidgetViews(views: android.widget.RemoteViews): Pair<TextView, View> {
```
TimoPtr: Import RemoteViews

app/src/test/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetTest.kt
```diff
@@ -0,0 +1,163 @@
...
+        coEvery { dao.get(appWidgetId) } returns createWidgetEntity(attributeIds = "unit_of_measurement", lastUpdate = "5 W")
+        coEvery { serverManager.integrationRepository(serverId) } returns integrationRepository
+        coEvery { integrationRepository.getEntity(entityId) } returns null
+
+        val (text, errorIcon) = resolveWidgetViews(widget.getWidgetRemoteViews(context, appWidgetId))
+
+        assertEquals("5 W", text.text.toString())
+        assertEquals(View.GONE, errorIcon.visibility)
+        coVerify(exactly = 0) { dao.updateWidgetLastUpdate(any(), any()) }
+    }
```
TimoPtr, line 163: Actually we might want to show the error in this case, if you look at the code of getEntity the case where it returns null is that we are in an insecure state (meaning no way to reach the server securely), we should notify the user that something is wrong with his server otherwise he won't know that the value displayed is out of sync.

ollide:
Thanks for the review! I'll take a closer look on your comments after the weekend.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetTest.kt
ollide: Resolved in https://github.com/home-assistant/android/pull/6501/commits/69a29c9f47b348838b4c9d9712cc91a896fe9a36

app/src/test/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetTest.kt
ollide: Fair point, I agree. I've updated the code in https://github.com/home-assistant/android/pull/6501/commits/805f7c1312a9cc3121e0053588857157838ee08a so that the error state is shown whenever the entity is not available (exception or insecure state). I think that this is a reasonable behavior for an **Entity**Widget.

Please let me know if there's anything else to address.

ollide:
> If you feel adventurous you could try to migrate this widget to the new way of making widgets like the TodoWidget in a second PR.

Thanks for the hint! I'll take a look at that in the next few days.

PR #6506, opened 2026-02-26
Enable auto close Assist after 30seconds of inactivity

TimoPtr:
## Summary

Since we are introducing hands free interaction with Assist, Assist should also close itself automatically when the user has finish using it. This PR adds an inactive behavior to assist and after 30s of no new message after the last message the window is going to close itself automatically. It is especially useful on Wall mount tablets.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistViewModel.kt
@@ -70,6 +70,7 @@ class AssistViewModel @Inject constructor(
     var pendingWakeWordConfirmation by mutableStateOf(false)
         private set
 
+    private var startedFromWakeWord = false
     private var inactivityTimerJob: Job? = null
 
     fun onCreate(
@@ -82,6 +83,7 @@ class AssistViewModel @Inject constructor(
         viewModelScope.launch {
             this@AssistViewModel.hasPermission = hasPermission
             this@AssistViewModel.wakeWordPhrase = wakeWordPhrase
+            this@AssistViewModel.startedFromWakeWord = wakeWordPhrase != null
             serverId?.let {
                 filteredServerId = serverId
                 selectedServerId = serverId
@@ -186,9 +188,13 @@ class AssistViewModel @Inject constructor(
      * The timer only runs when the input mode is [AssistInputMode.VOICE_INACTIVE],
      * TTS audio is not currently playing, and the last conversation message is not a
      * placeholder (assistant finished processing).
+     *
+     * Disabled for non wake word sessions to avoid auto-closing the dialog while the user
+     * is interacting with hands.
      */
     private fun restartInactivityTimer() {
         inactivityTimerJob?.cancel()
+        if (!startedFromWakeWord) return
         fun isInactive(): Boolean {
             val shouldRun = when (inputMode) {
                 AssistInputMode.VOICE_INACTIVE -> true
--- a/app/src/test/kotlin/io/homeassistant/companion/android/assist/AssistViewModelTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/assist/AssistViewModelTest.kt
@@ -110,14 +110,14 @@ class AssistViewModelTest {
         )
     }
 
-    private fun createAndInitialize(hasPermission: Boolean = false): AssistViewModel {
+    private fun createAndInitialize(hasPermission: Boolean = false, startedWithWakeWord: Boolean = false): AssistViewModel {
         val vm = createViewModel()
         vm.onCreate(
             hasPermission = hasPermission,
             serverId = null,
             pipelineId = null,
             startListening = null,
-            wakeWordPhrase = null,
+            wakeWordPhrase = if (startedWithWakeWord) "Okay Nabu" else null,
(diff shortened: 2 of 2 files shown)


jpelgrom:
What if you listen to a long TTS answer for more than 30s?

> Since we are introducing hands free interaction with Assist

Maybe it should be limited to cases where you started Assist hands free (wake word, maybe also headset button)? If I triggered Assist by clicking on the icon / using my assistant gesture I find it odd to see it disappear automatically, almost no UI does that.

If you don't agree, then this change should also be applied to Wear OS.

TimoPtr:
> What if you listen to a long TTS answer for more than 30s?

We could introduce a more complex logic to detect inactive state in the viewModel, I'm not sure exacty what should be the criteria. As Copilot suggested we could remove the delay if you switch the input method to Text or currently speaking. We could also make disable it if any audio is playing.

> 
> > Since we are introducing hands free interaction with Assist
> 
> Maybe it should be limited to cases where you started Assist hands free (wake word, maybe also headset button)? If I triggered Assist by clicking on the icon / using my assistant gesture I find it odd to see it disappear automatically, almost no UI does that.
> 
> If you don't agree, then this change should also be applied to Wear OS.

I don't have any opinion on this, but we can start to bring this only to wake word.


I'm going to wait for your answer before going further on this.

bgoncal:
I really think that to merge this we need, first, to be opt-in, and then also have a picker to choose between 10 seconds, 20 seconds, 1 minute etc, without that we are making random assumptions

TimoPtr:
> I really think that to merge this we need, first, to be opt-in, and then also have a picker to choose between 10 seconds, 20 seconds, 1 minute etc, without that we are making random assumptions

If we do add the conditions I've added above. Inactive state would be pretty clear. Adding too much settings might also not be a good idea. I'm considering more and more the stand from [Signal](https://github.com/signalapp/Signal-Android/blob/main/CONTRIBUTING.md#development-ideology) about `The answer is not more options`.

I would like to test it only on the wake word and get feedback from users before adding yet another option.

bgoncal:
Do you think you will get enough feedback from the beta to conclude the test? Don't forget our user base is broad with very broad use cases, you may break someone's flow if released to prod without a way to tweak it.

TimoPtr:
> Do you think you will get enough feedback from the beta to conclude the test? Don't forget our user base is broad with very broad use cases, you may break someone's flow if released to prod without a way to tweak it.

That's why I agree with @jpelgrom by enabling this only to wake word detection. No one uses wake word detection yet, they might (I hope) in the next release, but by marking it as experimental I allow us to make breaking changes.

For the wake word detection timeout it was one of the first feedback I got from users that tried it. You don't want to touch a device that you started with voice (like a wall tablet).

bgoncal:
Sounds good 👍🏻

jpelgrom:
> I don't have any opinion on this, but we can start to bring this only to wake word.

If you don't have an opinion, reading the later discussion I'd start by bringing it only to wake word triggered Assist, adjust the check to account for active TTS as well, so we limit the exposure if it isn't received well.

Fully agree on the stance of limiting options, we already have too many in some places and should go for defaults that work for 95%. If that really isn't possible then we can consider more options.

TimoPtr:
I worked with Claude on my last commit, I didn't have time to test it.

app/src/main/kotlin/io/homeassistant/companion/android/assist/AssistViewModel.kt
```diff
@@ -390,7 +438,13 @@ class AssistViewModel @Inject constructor(
                             lastMessage.copy(message = lastMessage.message + event.chunk)
                     }
                 }
+
                 is AssistEvent.PipelineStarted -> { /* handled below */ }
+
+                is AssistEvent.PipelineEnded,
+                is AssistEvent.PlaybackFinished,
```
jpelgrom, line 460: PipelineEnded (from server) will probably happen before PlaybackFinished (from audio player on device). That means the timer is reset once after TTS finishes playing, probably fine but not exactly the same. Why not link this only to finishing playback?

TimoPtr: Good catch we should only reset on pipelineEnded

TimoPtr: Actually no we do want to start the timeout only after we finished to play the audio, imagine if you generate a long text more than 30s, you don't want to finish the activity in the middle of the TTS.

jpelgrom:
> That's why I agree with @jpelgrom by enabling this only to wake word detection. 

Unless I'm missing something this part wasn't implemented. It's always doing an inactive timer, it doesn't matter whether Assist was opened using the wake word or another way.

TimoPtr:
> > That's why I agree with @jpelgrom by enabling this only to wake word detection. 
> 
> Unless I'm missing something this part wasn't implemented. It's always doing an inactive timer, it doesn't matter whether Assist was opened using the wake word or another way.

That might be true, if I remember right Wake word was not yet merged when I've opened this PR. I'll do that.

TimoPtr:
Done @jpelgrom

jpelgrom (approved):

PR #6509, opened 2026-02-26
Adjust LocationForSecureConnection wording

TimoPtr:
## Summary

This PR adjust the wording of the LocationForSecureConnection screen to match the current configuration and not a generic one that was missleading for some users.

## Checklist


common/src/main/res/values/strings.xml
```diff
@@ -1403,7 +1403,8 @@
     <string name="location_sharing_share">Share my location</string>
     <string name="location_sharing_no_share">Do not share my location</string>
     <string name="location_secure_connection_title">Let us help secure your remote connection</string>
-    <string name="location_secure_connection_content">It appears that your connection is not encrypted (using HTTP). If the app knows when you\'re away from home, it can choose a more secure way to connect to your Home Assistant system if available. This requires location services to be enabled.</string>
+    <string name="location_secure_connection_content_http">It appears that your connection is not encrypted (currently using HTTP). If the app knows when you\'re away from home, it can choose a more secure way to connect to your Home Assistant system if available. This requires location services to be enabled.</string>
```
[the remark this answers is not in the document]
TimoPtr: I'm not sure how to properly address that, I want to say that it's fine in a way since `currently` could mean in your config currently you have HTTP.

common/src/main/res/values/strings.xml
```diff
@@ -1404,7 +1403,8 @@
...
+    <string name="location_secure_connection_content">If the app knows when you\'re away from home, it can choose a more secure way to connect to your Home Assistant system if available. This requires location services to be enabled.</string>
```
jpelgrom: Because the string name used here matches the old one with http content, any translations may show the old value until updated.

It also duplicates everything except the first sentence in the message about http. I'd make the first sentence for http a separate string, and then always use this string. Something like: `string = "${if (hasPlainTextUrl) stringResource(commonR.string.location_secure_connection_content_http) + " " }${stringResource(commonR.string.location_secure_connection_content_location)}"`

common/src/main/res/values/strings.xml
```diff
@@ -1404,7 +1403,8 @@
     <string name="location_sharing_share">Share my location</string>
     <string name="location_sharing_no_share">Do not share my location</string>
     <string name="location_secure_connection_title">Let us help secure your remote connection</string>
-    <string name="location_secure_connection_content">It appears that your connection is not encrypted (using HTTP). If the app knows when you\'re away from home, it can choose a more secure way to connect to your Home Assistant system if available. This requires location services to be enabled.</string>
+    <string name="location_secure_connection_content_http">It appears that your connection is not encrypted (currently using HTTP).</string>
```
jpelgrom: The check from a technical view is 'does the server have _any_ plain text url'. This text suggests you are _currently using_ a plain text url.

TimoPtr: What about `currently using HTTP in your configuration`?

PR #6532, opened 2026-03-04
Catch WebView error while instantiating with a proper error

TimoPtr:
## Summary

This address some crashes when the Webview installed on the device is not meant for this device.
```
dlopen failed: "/data/app/com.google.android.webview-iTOIjAqw4frMMqAsce-UIw==/lib/arm/libwebviewchromium.so" is 32-bit instead of 64-bit
```
Screenshot 2026-03-04 at 12 50 59

I've catch it everywhere we use a webview, and made a new framework in the FrontendScreen to introduce a new concept of unrecoverable error. Also the Open settings is now always displayed in the error screen. 

## Checklist

## Screenshots

image

## Any other notes

I made it best effort on the old WebViewActivity with a snackbar, it's going then to show a blank screen and later the timeout dialog.

To test you can modify the code to throw in the HAWebView and test ConnectionScreen and FrontendScreen.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
@@ -96,7 +96,7 @@ internal class FrontendViewModel @VisibleForTesting constructor(
     @OptIn(ExperimentalForInheritanceCoroutinesApi::class)
     private class ViewStateManager(
         initialState: FrontendViewState,
-        private val _state: MutableStateFlow<FrontendViewState> = MutableStateFlow(initialState)
+        private val _state: MutableStateFlow<FrontendViewState> = MutableStateFlow(initialState),
     ) : StateFlow<FrontendViewState> by _state.asStateFlow() {
 
         /**
@@ -107,7 +107,9 @@ internal class FrontendViewModel @VisibleForTesting constructor(
          */
         fun update(transform: (FrontendViewState) -> FrontendViewState) {
             _state.update { currentState ->
-                if (currentState is FrontendViewState.Error && currentState.error is FrontendConnectionError.UnrecoverableError) {
+                if (currentState is FrontendViewState.Error &&
+                    currentState.error is FrontendConnectionError.UnrecoverableError
+                ) {
                     Timber.w("Ignoring state transition: unrecoverable error present")
                     return
                 }
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
@@ -201,7 +201,7 @@ private fun StateOverlay(
     when (viewState) {
         is FrontendViewState.LoadServer,
         is FrontendViewState.Loading,
-            -> LoadingScreen(modifier = Modifier.background(LocalHAColorScheme.current.colorSurfaceDefault))
+        -> LoadingScreen(modifier = Modifier.background(LocalHAColorScheme.current.colorSurfaceDefault))
 
         is FrontendViewState.Content -> {
             // WebView is visible, no overlay needed
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/error/FrontendConnectionError.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/error/FrontendConnectionError.kt
@@ -58,10 +58,8 @@ sealed interface FrontendConnectionError {
          * problem. The recommended user action is to check and update the system WebView
          * via device settings.
          */
-        data class WebViewCreationError(
-            @StringRes override val message: Int,
-            val throwable: Throwable,
-        ) : UnrecoverableError {
+        data class WebViewCreationError(@StringRes override val message: Int, val throwable: Throwable) :
+            UnrecoverableError {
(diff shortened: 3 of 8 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/HAWebView.kt
```diff
@@ -67,26 +74,35 @@ fun HAWebView(
...
+                    )
+                    defaultSettings()
+                    configure(this)
+                }
+            } catch (e: Exception) {
+                Timber.e(e, "Failed to create WebView, the system WebView may be misconfigured")
+                onWebViewCreationFailed(e)
+                // AndroidView requires a non-null View; return an empty placeholder
+                FrameLayout(context)
             }
```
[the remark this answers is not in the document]
TimoPtr: This is true but I'm not sure today where the exception is thrown from the WebView constructor or for instance the instanciation of a WebViewClient. I suppose the first but I'm not sure.

@jpelgrom what do you think, I think that a wrong configuration should be catch by us or in beta.

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
```diff
@@ -361,4 +361,31 @@ class ConnectionViewModelTest {
         // Then: repository method was called for each click
         verify(exactly = 2) { connectivityCheckRepository.runChecks(rawUrl) }
     }
+
+    @Test
+    fun `Given a working state when onWebViewCreationFailed is called then errorFlow emits WebViewCreationError and connectivity checks run`() = runTest {
```
jpelgrom, line 366: > `and connectivity checks run`

This is actually weird behavior to have and explicitly test for, for an error that is unrelated to connectivity. The URL is also probably `about:blank` at this point so it adds no value.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -183,6 +216,22 @@ internal class FrontendViewModel @VisibleForTesting constructor(
...
+     *
+     * Transitions to [FrontendViewState.Error] with a [FrontendConnectionError.UnrecoverableError.WebViewCreationError]
+     * so the error screen is displayed with guidance to update the system WebView.
+     */
+    fun onWebViewCreationFailed(throwable: Throwable) {
+        onError(
+            FrontendConnectionError.UnrecoverableError.WebViewCreationError(
+                message = commonR.string.webview_creation_failed,
+                errorDetails = throwable.message ?: throwable.toString(),
+                rawErrorType = "WebViewCreationError",
```
jpelgrom: This might hide the exception class if the throwable has a message, which is quite important here because it is a generic catch.

Taking the example that inspired this change:

 - only `dlopen failed: libwebviewchromium.so is 32-bit`
 - or also having `UnsatisfiedLinkError`

Also better matches buildAuthUrl malformed URL with a generic catch including the error class in `rawErrorType`.

```suggestion
                rawErrorType = throwable::class.toString(),
```

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
```diff
@@ -179,6 +179,22 @@ internal class ConnectionViewModel @VisibleForTesting constructor(
...
+     *
+     * Transitions to an error state with a [FrontendConnectionError.UnrecoverableError.WebViewCreationError]
+     * so the error screen is displayed with guidance to update the system WebView.
+     */
+    fun onWebViewCreationFailed(throwable: Throwable) {
+        onError(
+            FrontendConnectionError.UnrecoverableError.WebViewCreationError(
+                message = commonR.string.webview_creation_failed,
+                errorDetails = throwable.message ?: throwable.toString(),
+                rawErrorType = "WebViewCreationError",
```
jpelgrom:
```suggestion
                rawErrorType = throwable::class.toString(),
```

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/HAWebView.kt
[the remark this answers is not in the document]
jpelgrom: The change is intended as a generic catch so I think it's fine like you've done it, as default settings is still initialization from a user point of view.

No need to make it more specific as suggested. If it happens we need to fix it anyway, better to get a beta user report with the specific error rather than "the app just crashes".

TimoPtr:
> From a technical point of view mostly fine, though I do wonder how often this occurs.

It is not often but it's still better to handle properly.

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
TimoPtr: I decided to keep the logic for the checks but only modify the UI so that when it is `about:blank` we hide the URL and the connectivity checks. The checks fail on the first checks I don't think it is an issue.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
TimoPtr: Good call

PR #6541, opened 2026-03-05
Go async for TileReceiver on the watch

TimoPtr:
## Summary

In order to attempt to reduce the number of ANR, I'm proposing to move the TileReceiver action to a dedicated coroutineScope and `lauchAsync` to make sure it complete under 30s.

I suppose that the current ANRs are caused by server not accessible for multiple reason
- watch bad network
- server not reachable
- ...

In any case not being able to perform the action under 30s is probably something that we should stop.

⚠️ this is quite a change since today it would simply wait that the whole process finish, meaning potentially test all the webhook URLs.

I decided to move the haptic after the success so it gives an idea to the user if it succeed or not. But that might be a bad idea since now he doesn't know that the action was took in consideration. I was thinking about adding an error vibration pattern but it is not something users are used too.

I'm open to discussion on this. 

Screenshot 2026-03-05 at 14 42 52
(For the last 7 days)

image
(For the last 60 days)

## Checklist

## Any other notes

Based on https://github.com/home-assistant/android/pull/6467

### DIFF
--- a/wear/src/main/kotlin/io/homeassistant/companion/android/tiles/TileActionReceiver.kt
+++ b/wear/src/main/kotlin/io/homeassistant/companion/android/tiles/TileActionReceiver.kt
@@ -32,16 +32,21 @@ class TileActionReceiver : BroadcastReceiver() {
         if (entityId != null) {
             launchAsync(receiverScope) {
                 val hasHapticFeedback = wearPrefsRepository.getWearHapticFeedback()
+
+                if (hasHapticFeedback) hapticClick(context)
+
                 try {
                     onEntityPressedWithoutState(
                         entityId = entityId,
                         integrationRepository = serverManager.integrationRepository(),
                     )
-                    if (hasHapticFeedback) hapticClick(context)
                 } catch (e: CancellationException) {
                     throw e
                 } catch (e: Exception) {
-                    Timber.e(e, "Cannot call tile service")
+                    Timber.e(
+                        e,
+                        "Failed to call integration entity action for entityId=$entityId from TileActionReceiver",
+                    )
                 }
             }
         }


jpelgrom:
> I decided to move the haptic after the success so it gives an idea to the user if it succeed or not. But that might be a bad idea since now he doesn't know that the action was took in consideration.

At least in my testing, this resulted in a worse UX. The haptics is as feedback to indicate the app registered your touch (it's a small watch screen, nice to get confirmation). Networking on Wear OS can take a while and waiting seconds before getting haptic feedback feels disconnected.

> I was thinking about adding an error vibration pattern but it is not something users are used too.

Maybe not used to it but there is a constant (default haptic pattern) for [REJECT](https://developer.android.com/reference/android/view/HapticFeedbackConstants#REJECT) "A haptic effect to signal the (...) failure of a user interaction.", and iOS has [similar](https://developer.apple.com/design/human-interface-guidelines/playing-haptics#Notification) [patterns](https://developer.apple.com/design/human-interface-guidelines/playing-haptics#watchOS) for warning/error.

jpelgrom:
@TimoPtr I see multiple pushes but think you haven't addressed (any reply: agree or disagree) my comment on moving the haptic feedback yet.

TimoPtr:
> @TimoPtr I see multiple pushes but think you haven't addressed (any reply: agree or disagree) my comment on moving the haptic feedback yet.

Not yet, that was one of my next thing. I didn't focus on this one since it was not a high priority TBH.

TimoPtr:
I decided to bring it back to what it was so we don't introduce a change here (even if the change of Dispatcher might introduce a very small delay).

jpelgrom:
> I decided to bring it back to what it was 

I don't follow - previously haptic happened before the network call but now it waits for the network call to complete? `onEntityPressedWithoutState` is a suspending function while network is happening.

TimoPtr:
> > I decided to bring it back to what it was
> 
> I don't follow - previously haptic happened before the network call but now it waits for the network call to complete? `onEntityPressedWithoutState` is a suspending function while network is happening.

I forgot to push now it should be ok.

PR #6542, opened 2026-03-05
Update coroutine usage in ComplicationReceiver

TimoPtr:
## Summary

Change the coroutine usage in complication receiver since it is mostly not needed anymore. I was even wondering if we really need to check if a server is registered if not we can remove the coroutine completely. 

## Checklist

## Any other notes

Based on https://github.com/home-assistant/android/pull/6467


jpelgrom:
>  I was even wondering if we really need to check if a server is registered if not we can remove the coroutine completely.

If the check is moved to `EntityStateDataSourceService`, it could be removed here. But I'd argue that this receiver is a lot lighter (less things to inject than the service) so best to stop early in this receiver instead of the service.

PR #6552, opened 2026-03-07
Skip insecure connection prompt when Home Assistant Cloud is active

tmmueller:
## Summary

When a user has Home Assistant Cloud enabled for external access, the app should not prompt them about insecure (HTTP) connections for the external URL, because the cloud URL (always HTTPS) is used instead.

Previously, hasPlainTextUrl in ServerConnectionInfo checked all configured URLs including the external URL, even when Cloud was active. This caused an unnecessary security prompt on startup asking the user to allow insecure connections.

This change excludes the external URL from the HTTP check when useCloud is true and a cloud URL is available. If the user later disables Cloud, the prompt will correctly reappear.

## Checklist

- [x] New or updated tests have been added to cover the changes following the testing guidelines.
- [x] The code follows the project's code style and best practices.
- [x] Changes are backward compatible whenever feasible.

## Any other notes

The fix inlines the cloud URL check rather than referencing the canUseCloud property due to Kotlin data class property initialization order.

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/database/server/ServerConnectionInfo.kt
@@ -119,13 +119,13 @@ data class ServerConnectionInfo(
     }
 
     /**
-     * Checks if any of the actively used URLs use plain HTTP (non-HTTPS).
+     * Checks if any of the configured URLs use plain HTTP (non-HTTPS).
      *
      * When Home Assistant Cloud is enabled ([useCloud] is `true` and [cloudUrl] is available),
      * the [externalUrl] is excluded from this check because the app will use the cloud URL
      * for external connections instead.
      *
-     * @return `true` if any actively used URL uses HTTP
+     * @return `true` if any checked URL uses HTTP
      */
     @Ignore
     val hasPlainTextUrl: Boolean = listOfNotNull(


jpelgrom:
This behavior is somewhat intentional - the app still knows that you have a HTTP URL and you need to make a choice. Until you do it [may be still be used automatically](https://github.com/home-assistant/android/blob/75bf916fb6fdd8731fc3b602d277c0fb0a21c7f3/common/src/main/kotlin/io/homeassistant/companion/android/common/data/servers/ServerConnectionStateProviderImpl.kt#L97), and if you were to stop using cloud the app would switch back automaticaly.

You're not the first person to notice that this question is somewhat [unexpected](https://discord.com/channels/330944238910963714/1346948551892009101/1452967234275115041) however. Note that there is also a pending change to the text in this screen in #6509 which aims to reduce confusion.

I'm not in favor of this change, maybe defaulting this to disallow insecure connections by default is the way forward in the future once we've given people enough time to migrate/update. @TimoPtr what are your thoughts here?

tmmueller:
In my case, only the external URL had HTTP, and it would never be used because Home Assistant Cloud was enabled. The settings page didn't even show the external URL. To stop it from pestering me every time I opened the app, I had to disable Home Assistant Cloud, change the external URL, and re-enable Home Assistant Cloud. No text change is going to make that clear.

TimoPtr:
> In my case, only the external URL had HTTP, and it would never be used because Home Assistant Cloud was enabled. The settings page didn't even show the external URL. To stop it from pestering me every time I opened the app, I had to disable Home Assistant Cloud, change the external URL, and re-enable Home Assistant Cloud. No text change is going to make that clear.

Why not pick `Most secure`? If you had done so you wouldn't have seen the screen anymore? Like @jpelgrom mentioned this is intentional currently since if you ever have an issue cloud it is going to fallback to external and we don't want people to make an HTTP call and not being aware of it.

I'm going to close the PR since that is the behavior we want. Let's open an issue if the Most secure option is not working for you.

tmmueller:
@TimoPtr &  @jpelgrom If it's falling back to the external URL even though the settings page doesn't show that URL anywhere (because I have Home Assistant Cloud enabled), I would consider that a major security flaw. I don't want it to use an unsecure connection _ever_. Allowing the "Most secure" option to use an unencrypted URL that is hidden in the settings is not secure. I can open an issue, but I wanted to make sure I'm understanding this correctly first.

tmmueller:
For that matter, it sounds like the issue is not limited to HTTP/HTTPS. It is a problem for the app to fall back to an external URL that the settings page does not display. What if I had previously tested with some entirely different server as my external URL? When I enable Home Assistant Cloud, it hides the external URL, implying that it will no longer be used. If that's not the case, that's a problem.

With Home Assistant Cloud disabled, it is clear that it will use this URL:
image

With Home Assistant Cloud enabled, I cannot see the external URL at all, and there is no indication that it will use one:
image

TimoPtr:
We could display the URL still indeed. This screen could benefits from a refresh using the new composable components we've made too, if you want to look at it 🙏🏻.

PR #6577, opened 2026-03-16
Adjust WebsocketManager check to avoid location usage

jpelgrom:
## Summary

Adjust how WebsocketManager checks whether to run to return early if conditions won't be met, instead of getting all possible info for all settings and only then returning. This should reduce the latest Android 16 QPR showing a 'blue location dot' because the app checks internal/external on screen on/off for the app, when there might not be a reason to do so.

Logs showed activity from ServerConnectionStateProviderImpl immediately after the WebsocketManager is started, which this change should avoid.

```
2026-03-16 12:53:59.678 31288-31324 WM-SystemJobScheduler   io.homeassistant.companion.android   D  Scheduling work ID 0a201fd2-49fd-42f3-ac73-c7211ab20a71Job ID 156605
2026-03-16 12:53:59.681 31288-31324 WM-GreedyScheduler      io.homeassistant.companion.android   D  Starting work for 0a201fd2-49fd-42f3-ac73-c7211ab20a71
2026-03-16 12:53:59.683 31288-31288 WM-SystemJobService     io.homeassistant.companion.android   D  onStartJob for WorkGenerationalId(workSpecId=0a201fd2-49fd-42f3-ac73-c7211ab20a71, generation=0)
2026-03-16 12:53:59.684 31288-31324 WM-Processor            io.homeassistant.companion.android   D  Processor: processing WorkGenerationalId(workSpecId=0a201fd2-49fd-42f3-ac73-c7211ab20a71, generation=0)
2026-03-16 12:53:59.684 31288-31324 WM-Processor            io.homeassistant.companion.android   D  Work WorkGenerationalId(workSpecId=0a201fd2-49fd-42f3-ac73-c7211ab20a71, generation=0) is already enqueued for processing
2026-03-16 12:53:59.687 31288-31288 WM-WorkerWrapper        io.homeassistant.companion.android   D  Starting work for io.homeassistant.companion.android.websocket.WebsocketManager
2026-03-16 12:53:59.693 31288-11554 ServerConn...oviderImpl io.homeassistant.companion.android   D  usesInternalSsid is: true, usesWifi is: true
2026-03-16 12:53:59.694 31288-31318 WM-WorkerWrapper        io.homeassistant.companion.android   I  Worker result SUCCESS for Work [ id=0a201fd2-49fd-42f3-ac73-c7211ab20a71, tags={ io.homeassistant.companion.android.websocket.WebsocketManager } ]
```

Includes a minor change for `!!` usage to handle null instead.

## Checklist

## Screenshots

n/a

n/a

### DIFF
--- a/app/src/test/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
@@ -8,6 +8,7 @@ import androidx.core.app.NotificationCompat
 import androidx.work.ListenableWorker
 import androidx.work.testing.TestListenableWorkerBuilder
 import dagger.hilt.android.EntryPointAccessors
+import dagger.hilt.android.testing.HiltTestApplication
 import io.homeassistant.companion.android.common.data.servers.ServerManager
 import io.homeassistant.companion.android.database.server.Server
 import io.homeassistant.companion.android.database.settings.SensorUpdateFrequencySetting
@@ -31,8 +32,10 @@ import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
 
 @RunWith(RobolectricTestRunner::class)
+@Config(application = HiltTestApplication::class)
 class WebsocketManagerTest {
 
     private val powerManager = mockk<PowerManager>()
@@ -74,7 +77,7 @@ class WebsocketManagerTest {
     }
 
     @Test
-    fun givenSettingNever_whenJobRuns_thenFinishesWithoutChecks() = runTest {
+    fun `Given setting NEVER when job runs then finishes without checks`() = runTest {
         mockSetting(WebsocketSetting.NEVER)
         val worker = TestListenableWorkerBuilder<WebsocketManager>(context).build()
         val result = worker.doWork()
@@ -96,7 +99,7 @@ class WebsocketManagerTest {
     }
 
     @Test
-    fun givenSettingAlways_whenJobRunsWithoutConnection_thenFinishesWithoutScreenNetworkChecks() = runTest {
+    fun `Given setting ALWAYS when job runs without connection then finishes without screen and network checks`() = runTest {
         mockSetting(WebsocketSetting.ALWAYS)
         every { context.hasActiveConnection() } returns false
 
@@ -115,7 +118,7 @@ class WebsocketManagerTest {
     }
 
     @Test
-    fun givenSettingAlways_whenJobRunsWithoutRegistration_thenFinishesWithoutScreenNetworkChecks() = runTest {
+    fun `Given setting ALWAYS when job runs without registration then finishes without screen and network checks`() = runTest {
         mockSetting(WebsocketSetting.ALWAYS)
         every { context.hasActiveConnection() } returns true
         coEvery { entryPoint.serverManager.isRegistered() } returns false
@@ -126,8 +129,10 @@ class WebsocketManagerTest {
         assertEquals(ListenableWorker.Result.success(), result)
 
         // Has checked setting
-        coVerify(exactly = 1) { entryPoint.dao.get(any()) }
-        coVerify(exactly = 1) { entryPoint.serverManager.isRegistered() }
+        coVerify(exactly = 1) {
+            entryPoint.dao.get(any())
+            entryPoint.serverManager.isRegistered()
+        }
 
         // Has not run other settings' checks
         verify(exactly = 0) { context.getSystemService(PowerManager::class.java) }
@@ -135,7 +140,7 @@ class WebsocketManagerTest {
     }
 
     @Test
-    fun givenSettingScreenOn_whenJobRunsWithScreenOff_thenFinishesWithoutOtherChecks() = runTest {
+    fun `Given setting SCREEN_ON when job runs with screen off then finishes without other checks`() = runTest {
         mockSetting(WebsocketSetting.SCREEN_ON)
         every { context.hasActiveConnection() } returns true
         coEvery { entryPoint.serverManager.isRegistered() } returns true
@@ -156,7 +161,7 @@ class WebsocketManagerTest {
     }
 
     @Test
-    fun givenSettingHomeWifi_whenJobRunsWithoutHomeWifi_thenFinishesWithoutOtherChecks() = runTest {
+    fun `Given setting HOME_WIFI when job runs without home Wi-Fi then finishes without other checks`() = runTest {
         mockSetting(WebsocketSetting.HOME_WIFI)
         every { context.hasActiveConnection() } returns true
         coEvery { entryPoint.serverManager.isRegistered() } returns true
@@ -168,8 +173,10 @@ class WebsocketManagerTest {
         assertEquals(ListenableWorker.Result.success(), result)
 
         // Has checked setting
-        coVerify(exactly = 1) { entryPoint.dao.get(any()) }
-        coVerify(exactly = 1) { entryPoint.serverManager.connectionStateProvider(any()).isInternal(any()) }
+        coVerify(exactly = 1) {
+            entryPoint.dao.get(any())
+            entryPoint.serverManager.connectionStateProvider(any()).isInternal(any())
+        }
 
         // Has not run other settings' checks or tried to run worker
         verify(exactly = 0) { context.getSystemService(PowerManager::class.java) }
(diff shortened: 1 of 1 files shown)


TimoPtr:
If this is the real fix, then I would like a test that verify that we are not calling isInternal in this scenarios, so if we refactor we don't break it again and a small comment.

jpelgrom:
> If this is the real fix, then I would like a test that verify that we are not calling isInternal in this scenarios, so if we refactor we don't break it again and a small comment.

It's not really a 'fix', just being more mindful of unnecessary location usage where it isn't expected or needed, which now stands out more 🔵. I'll try adding tests to make sure it isn't accidentally changed.

jpelgrom:
Testing WorkManager seems to require instrumentation tests. Useful to learn but more complicated than expected. Please let me know what you think, I tried to avoid refactoring the actual worker.

app/src/androidTest/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
```diff
@@ -0,0 +1,195 @@
...
+import io.mockk.spyk
+import io.mockk.verify
+import junit.framework.TestCase.assertEquals
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.test.runTest
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
```
TimoPtr: Did you try with Robolectric?

app/src/androidTest/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
```diff
@@ -0,0 +1,195 @@
...
+        assertEquals(ListenableWorker.Result.success(), result)
+
+        // Has checked setting
+        coVerify(exactly = 1) { entryPoint.dao.get(any()) }
+
+        // Has not run other settings' checks
+        verify(exactly = 0) { context.hasActiveConnection() }
+        coVerify(exactly = 0) { entryPoint.serverManager.isRegistered() }
+        verify(exactly = 0) { context.getSystemService(Context.POWER_SERVICE) }
+        coVerify(exactly = 0) { entryPoint.serverManager.connectionStateProvider(any()) }
```
TimoPtr: Why not just having one verify and one coVerify?

app/src/androidTest/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    @Test
+    fun givenSettingNotNever_whenJobRunsWithoutConnection_thenFinishesWithoutScreenNetworkChecks() = runTest {
```
TimoPtr: Why not using Always in the name?

jpelgrom:
I tried Robolectric but it doesn't seem to like that with the worker. All examples and code from others I could find also do integration tests.

All logic is already in one private function in the worker which returns true/false, so could maybe be used, do you mean other adjustments or how would you test that?

TimoPtr:
> I tried Robolectric but it doesn't seem to like that with the worker. All examples and code from others I could find also do integration tests.
> 
> All logic is already in one private function in the worker which returns true/false, so could maybe be used, do you mean other adjustments or how would you test that?

I guess that's the issue you have is because some stuff are initialized in the constructor or within super. You could do it in 2 ways
- Try to mock all the context calls that fails when instantiating the worker from the test
- Simply move the logic into a private function out of the class and test that. (A bit ugly but easier)

app/src/androidTest/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
jpelgrom: Depending on whether or not I understood you correctly:
 - because I did not know about `verifyAll`, adjusted
 - because I want to test it finishes immediately with the setting is never and doesn't run _any_ of the other checks, so I need to check all other possible checks were not called and not just two

app/src/androidTest/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
jpelgrom: See https://github.com/home-assistant/android/pull/6577#issuecomment-4082037266 - after more trying managed to make it work with Robolectric, thanks for pushing me

app/src/androidTest/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
jpelgrom: Test name wasn't updated after changes, adjusted

app/src/androidTest/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
TimoPtr: I was simply mentioning that you can use only one verify and coverify blocks with multiple checks within it.

jpelgrom:
With more trying managed to adjust the tests, see the latest commit. Afterwards I rebased and force pushed to resolve a merge conflict before realizing I could have merged upstream into my branch instead, sorry.

app/src/test/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
```diff
@@ -0,0 +1,206 @@
...
+import io.mockk.verifyAll
+import junit.framework.TestCase.assertEquals
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.test.runTest
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
```
TimoPtr, line 37: You need to also set the HiltApplication class like we do in the other tests otherwise it triggers the Strict mode in tests since it uses HomeAssistantApplication that is not needed for the tests.

app/src/test/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
```diff
@@ -0,0 +1,206 @@
...
+    fun setup() {
+        mockkStatic(EntryPointAccessors::class)
+        every {
+            EntryPointAccessors.fromApplication(any(), WebsocketManager.WebsocketManagerEntryPoint::class.java)
+        } returns
+            entryPoint
+    }
+
+    @Test
+    fun givenSettingNever_whenJobRuns_thenFinishesWithoutChecks() = runTest {
```
TimoPtr: You can use proper sentences now that it is unit tests :)

app/src/test/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
```diff
@@ -0,0 +1,206 @@
...
+        coEvery { entryPoint.serverManager.isRegistered() } returns false
+
+        val worker = TestListenableWorkerBuilder<WebsocketManager>(context).build()
+        val result = worker.doWork()
+
+        assertEquals(ListenableWorker.Result.success(), result)
+
+        // Has checked setting
+        coVerify(exactly = 1) { entryPoint.dao.get(any()) }
+        coVerify(exactly = 1) { entryPoint.serverManager.isRegistered() }
```
TimoPtr: You can merge the two into only one coverify

app/src/androidTest/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
jpelgrom: Didn't know about that either and with another function existing named 'all' also didn't consider it... Changed it back to the more common `(co)verify`, thanks.

app/src/test/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
```diff
@@ -0,0 +1,206 @@
...
+        every { anyConstructed<NotificationCompat.Builder>().build() } returns mockk<Notification>()
+
+        val worker = spyk(TestListenableWorkerBuilder<WebsocketManager>(context).build())
+        coEvery { worker.setForeground(any()) } throws CancellationException() // Prevent worker from running
+        val result = worker.doWork()
+
+        assertEquals(ListenableWorker.Result.success(), result)
+
+        // Has checked setting
+        coVerify(exactly = 1) { entryPoint.dao.get(any()) }
```
TimoPtr: Same here you can merge the two coVerify exactly 1

Issue #6609, opened 2026-03-22
LocationTrackingViewModel leaks after navigating away from Location Tracking screen


loganrosen:
## Description

Navigating away from the Location Tracking screen (Settings → Troubleshooting → Location Tracking, then pressing back) leaks the `LocationTrackingViewModel`. The ViewModel is retained after `onCleared()` because its paging flow is anchored to the `AppDatabase` singleton through Room's `InvalidationTracker`.

## Root cause

The `Pager` factory lambda in `historyPagerFlow` implicitly captures `this` (the ViewModel) through the `locationHistoryDao` field reference. The `PagingSource` returned by `LocationHistoryDao.getAll()` registers invalidation callbacks with Room's `InvalidationTracker`, which runs a coroutine parented to the singleton `AppDatabase`'s `SupervisorJob` — not to any lifecycle.

Since the database lives for the process lifetime, the reference chain from `AppDatabase` → `InvalidationTracker` → `PagingSource` → `PageFetcher` → factory lambda → ViewModel is never broken after `onCleared()`.

Note: adding `.cachedIn(viewModelScope)` alone does not fix this — the `PageFetcher` factory lambda still captures the ViewModel regardless of scope management.

## Leak trace (from LeakCanary)

```
AppDatabase.instance (static singleton, never GC'd)
  → RoomDatabase.transactionContext
    → StandaloneCoroutine (suspended in InvalidationTracker.createFlow)
      → CommonLimitOffsetImpl.pagingSource
        → LocationHistoryDao_Impl$getAll$1 (PagingSource)
          → InvalidateCallbackTracker.callbacks
            → PageFetcher.pagingSourceFactory
              → LocationTrackingViewModel$historyPagerFlow$1$1.this$0
                → LocationTrackingViewModel (LEAKING, received onCleared())
```

PR #6612, opened 2026-03-22
Fix LocationTrackingViewModel leak via Room InvalidationTracker

loganrosen:
## Summary

Fix a memory leak where `LocationTrackingViewModel` is retained after `onCleared()`. The `Pager` factory lambda in `historyPagerFlow` registers `PagingSource` instances with Room's `InvalidationTracker`, which is anchored to the singleton `AppDatabase`. When the flow stops collecting, nothing unregisters the `PagingSource`, so the reference chain from `AppDatabase` back to the ViewModel is never broken.

The fix tracks the current `PagingSource` and calls `invalidate()` in `onCompletion`, which unregisters it from Room's tracker. Also adds `.cachedIn(viewModelScope)` to scope the `PageFetcher` lifecycle to the ViewModel and avoid restarting the paging pipeline on recomposition.

Leak trace and analysis: #6609

## Checklist

## Any other notes

Verified the fix with LeakCanary on an emulator — the leak no longer reproduces after navigating away from the Location Tracking screen.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/developer/location/LocationTrackingViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/developer/location/LocationTrackingViewModel.kt
@@ -2,7 +2,6 @@ package io.homeassistant.companion.android.settings.developer.location
 
 import android.app.Application
 import androidx.annotation.IdRes
-import androidx.annotation.VisibleForTesting
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.setValue
@@ -10,6 +9,7 @@ import androidx.lifecycle.AndroidViewModel
 import androidx.lifecycle.viewModelScope
 import androidx.paging.Pager
 import androidx.paging.PagingConfig
+import androidx.paging.PagingSource
 import androidx.paging.cachedIn
 import dagger.hilt.android.lifecycle.HiltViewModel
 import io.homeassistant.companion.android.R
@@ -23,8 +23,8 @@ import javax.inject.Inject
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.flatMapLatest
+import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.launch
-import timber.log.Timber
 
 @HiltViewModel
 class LocationTrackingViewModel @Inject constructor(
@@ -36,31 +36,6 @@ class LocationTrackingViewModel @Inject constructor(
 
     companion object {
         private const val PAGE_SIZE = 25
-
-        // Companion function so that the Pager factory lambda does not implicitly
-        // capture the ViewModel via field references, which would cause a memory leak
-        // through Room's InvalidationTracker (anchored to the singleton AppDatabase).
-        @VisibleForTesting
-        internal fun createHistoryPager(
-            dao: LocationHistoryDao,
-            filter: HistoryFilter,
-        ): Pager<Int, LocationHistoryItem> = Pager(PagingConfig(pageSize = PAGE_SIZE, maxSize = PAGE_SIZE * 6)) {
-            Timber.d("Returning PagingSource for history filter: $filter")
-            when (filter) {
-                HistoryFilter.ALL -> dao.getAll()
-                HistoryFilter.SENT ->
-                    dao.getAll(listOf(LocationHistoryItemResult.SENT.name))
-                HistoryFilter.SKIPPED -> dao.getAll(
-                    (
-                        LocationHistoryItemResult.entries.toMutableList() - LocationHistoryItemResult.SENT -
-                            LocationHistoryItemResult.FAILED_SEND
-                        )
-                        .map { it.name },
-                )
-                HistoryFilter.FAILED ->
-                    dao.getAll(listOf(LocationHistoryItemResult.FAILED_SEND.name))
-            }
-        }
     }
 
     enum class HistoryFilter(@IdRes val menuItemId: Int) {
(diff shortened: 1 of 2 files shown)


app/src/test/kotlin/io/homeassistant/companion/android/settings/developer/location/LocationTrackingViewModelTest.kt
```diff
@@ -0,0 +1,156 @@
...
+@OptIn(ExperimentalCoroutinesApi::class)
+@ExtendWith(ConsoleLogExtension::class, MainDispatcherJUnit5Extension::class)
+class LocationTrackingViewModelTest {
+
+    private val locationHistoryDao: LocationHistoryDao = mockk(relaxed = true)
+    private val prefsRepository: PrefsRepository = mockk(relaxed = true)
+    private val serverManager: ServerManager = mockk(relaxed = true)
+    private val application: Application = mockk(relaxed = true)
+
+    private lateinit var viewModel: LocationTrackingViewModel
```
TimoPtr: Not a big deal but I would be in favor to not keep the VM as a class attribute when not set in the BeforeEach, otherwise it might be annoying when adding a new test.

app/src/main/kotlin/io/homeassistant/companion/android/settings/developer/location/LocationTrackingViewModel.kt
```diff
@@ -33,6 +36,28 @@ class LocationTrackingViewModel @Inject constructor(
 
     companion object {
         private const val PAGE_SIZE = 25
+
+        @VisibleForTesting
+        internal fun createHistoryPager(
```
TimoPtr: You could add a comment saying why this is not within the class.

app/src/main/kotlin/io/homeassistant/companion/android/settings/developer/location/LocationTrackingViewModel.kt
```diff
@@ -57,24 +82,8 @@ class LocationTrackingViewModel @Inject constructor(
...
-                        LocationHistoryItemResult.entries.toMutableList() - LocationHistoryItemResult.SENT -
-                            LocationHistoryItemResult.FAILED_SEND
-                        )
-                        .map { it.name },
-                )
-                HistoryFilter.FAILED ->
-                    locationHistoryDao.getAll(listOf(LocationHistoryItemResult.FAILED_SEND.name))
-                else -> locationHistoryDao.getAll()
-            }
-        }.flow
```
TimoPtr, line 76: So when we stop collecting the flow it still continue? It's weird because if that's the case then it's still kinda leaking just not the ViewModel anymore. 

Or the fix is actually using cachedIn like the docs suggested to do ? 

> Primary entry point into Paging; constructor for a reactive stream of PagingData. The same Pager instance should be reused within an instance of ViewModel. For example in your ViewModel:
// create a Pager instance and store to a variable
val pager = Pager(
     ...
     )
     .flow
     .cachedIn(viewModelScope)

I think you can document this.

app/src/test/kotlin/io/homeassistant/companion/android/settings/developer/location/LocationTrackingViewModelTest.kt
loganrosen: Done — `createViewModel()` now returns the instance and each test uses a local `val`.

app/src/main/kotlin/io/homeassistant/companion/android/settings/developer/location/LocationTrackingViewModel.kt
loganrosen: Added a comment explaining the companion placement is to avoid capturing `this` and the resulting leak through Room'\''s InvalidationTracker.

app/src/main/kotlin/io/homeassistant/companion/android/settings/developer/location/LocationTrackingViewModel.kt
loganrosen: Good question. The companion function breaks the reference chain so the ViewModel can be GC'\''d — that'\''s the actual leak fix. `cachedIn` scopes the `PageFetcher` to `viewModelScope` so it tears down on `onCleared()`, which is the recommended pattern from the Paging docs. Both are needed: the companion prevents the ViewModel from being retained, and `cachedIn` ensures the paging pipeline doesn'\''t restart on recomposition. Added an inline comment on the `cachedIn` call.

TimoPtr: It does stop leaking the ViewModel that's only solving part of the issue. If you do enter a billion times the LocationTracking screen the database would still holds a reference to a billion of Pager because of how the pagerFactory works.

Each time you create a new pager you are going to register the pager within the database tracker as you found out. But nothing is going to remove the pager from the database tracker ever (it might be a miss from the room+pager API). 

After checking the code of the `cachedIn` operator I found out that they do invalidate the source if it changes, but it is not enough to remove the leak entirely. I'm searching a bit more to let you know how to address this properly.

TimoPtr: Ok after some issue with my AS setup I found out a way to get rid of the leak properly. It could be ported most probably in the Pagination library


```kotlin
   val historyPagerFlow = historyResultFilter.flatMapLatest { filter ->
        var currentSource: PagingSource<Int, LocationHistoryItem>? = null
        Pager(PagingConfig(pageSize = PAGE_SIZE, maxSize = PAGE_SIZE * 6)) {
            when (filter) {
                HistoryFilter.SENT ->
                    locationHistoryDao.getAll(listOf(LocationHistoryItemResult.SENT.name))
                HistoryFilter.SKIPPED -> locationHistoryDao.getAll(
                    (
                        LocationHistoryItemResult.entries.toMutableList() - LocationHistoryItemResult.SENT -
                            LocationHistoryItemResult.FAILED_SEND
                        )
                        .map { it.name },
                )
                HistoryFilter.FAILED ->
                    locationHistoryDao.getAll(listOf(LocationHistoryItemResult.FAILED_SEND.name))
                else -> locationHistoryDao.getAll()
            }.also { currentSource = it }
        }.flow.onCompletion {
            currentSource?.invalidate()
        }
    }.cachedIn(viewModelScope)
```

The keys are
- using `cachedIn` as you suggested, to avoid making useless recreation of Pager
- keeping track in a local variable of the latest PagingSource so that we can manually invalidate it onCompletion of the flow.

if you do this the leak is gone and we properly remove the PagingSource from the database.

loganrosen: Great find — adopted your approach. Replaced the companion function with the `onCompletion { currentSource?.invalidate() }` pattern to properly unregister the PagingSource from Room's tracker. Verified on emulator with LeakCanary, no leak.

PR #6614, opened 2026-03-23
Show snackbar with troubleshooting link for Matter/Thread non-OK activity results

jpelgrom:
## Summary

Matter/Thread failures are currently quite frustrating to troubleshoot, both for the user and as devs. If something fails in the GMS flow or the user cancels, the activity completes with Activity.RESULT_CANCELED with data, but that data always seems to be 'internal error' no matter the actual error.

Activity result with data showing GMS ApiException INTERNAL_ERROR

This PR tries to improve the experience a bit by showing a snackbar on failures with a link to the troubleshooting section of the docs. This should hopefully give users more info on how they can fix things, and maybe us devs if something should be clarified in the docs. This includes converting the existing toasts for unavailability to a snackbar with a link.

This change is only for the current WebViewActivity as the new frontend code does not yet implement Matter or Thread functionality at all.

## Checklist

- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction). _I don't really want to touch this tangled code too much, manual test for the snackbar and we can improve this in the new frontend code to be testable._

## Screenshots

Generic error (unsupported or error before GMS flow starts):

https://github.com/user-attachments/assets/f58bd9f6-0594-477b-ace7-4912ca810b7c

Cancelled error (I'm manually exiting here, but this would also occur if the user encounters an error during commissioning, but unfortunately we don't get any actionable details):

https://github.com/user-attachments/assets/493c43be-f65c-46c3-a06a-595cc44d3374

n/a, I'm linking to the most relevant sections in the existing integration doc


TimoPtr:
Could we link to the troubleshooting section within the app about thread credentials in the settings directly?

jpelgrom:
> Could we link to the troubleshooting section within the app about thread credentials in the settings directly?

Do you mean the full syncing option? We technically could but we don't have a specific enough result to determine when that's relevant and when it's not.

TimoPtr:
> > Could we link to the troubleshooting section within the app about thread credentials in the settings directly?
> 
> Do you mean the full syncing option? We technically could but we don't have a specific enough result to determine when that's relevant and when it's not.

We might benefit m from an error screen with a custom action like open doc and sync.

jpelgrom:
> We might benefit m from an error screen with a custom action like open doc and sync.

I can see how it might be beneficial in case of actual errors, but that would increase the scope of this PR massively and require additional work. If you agree I'd like to move forward with this approach first as it already improves the experience, and a specific screen can come after.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -784,21 +784,45 @@ class WebViewActivity :
...
+                                        message = getString(message),
+                                        actionLabel = getString(commonR.string.get_help),
+                                        duration = SnackbarDuration.Long,
+                                    ) == SnackbarResult.ActionPerformed
+                                ) {
+                                    val intent = Intent(Intent.ACTION_VIEW, uri.toUri())
+                                    startActivity(intent)
+                                }
+                            }
+                            alertDialog?.cancel()
```
TimoPtr, line 825: You weren't canceling the dialog before in `ERROR_MATTER`, but I guess it makes sense.

jpelgrom: The dialog shouldn't be showing in the flow with `ERROR_MATTER`, but it might in the Thread flow. As I've merged the flows I'm adding it here, doesn't hurt if it doesn't exist and otherwise removes the dialog.

PR #6618, opened 2026-03-23
Use state-aware icons for Wear OS shortcut tiles

### LINKED ISSUE #2045: Wear OS - Add status effect to icon
No way to know if light bulb is on or off.

Icon color change like in home-assistant web app

kdavh:
Closes #2045

## Summary

Shortcut tiles on Wear OS use domain-based fallback icons that don't reflect entity state — a lock always shows the same padlock icon, a switch always shows the same toggle. This change renders state-aware icons (e.g., locked vs unlocked padlock, powered vs unpowered outlet) and keeps them in sync with Home Assistant in real time.

**Before:** All lock entities show the same generic padlock icon.
**After:** Locked entities show a closed padlock, unlocked entities show an open padlock (and similar state-aware icons for other domains like covers, fans, switches).

## How it works

- **Real-time state via WebSocket**: while the tile is visible, `IntegrationRepository.getEntityUpdates(entityIds)` drives a shared in-memory cache. No REST polling; no `getEntities()` batch over the full entity list.
- **Cold-start warming**: on first render, any shortcut entity not yet in the cache is fetched via parallel `getEntity()` calls bounded by a 2s timeout. If the network is slow the tile still renders on time using domain-default icons, and the cache fills in on the next render.
- **Snapshot-per-render**: `onTileRequest` freezes the cache into a `TileSnapshot` keyed by the resources version; `onTileResourcesRequest` looks the snapshot up again and generates bitmaps from the same frozen view. Resource IDs include the state (`switch.foo@on` vs `switch.foo@off`) so the Wear bitmap cache can't serve a stale image.
- **Optimistic click**: on tap, the cached state is flipped to the predicted post-tap value (for known toggle domains: switch, light, lock, cover, input_boolean, fan) before the layout is built. The WebSocket subscription reconciles shortly after. This avoids a framework quirk where `requestUpdate()` doesn't reliably trigger a re-render while a tile is visible.
- **Lifecycle**: subscription starts in `onTileEnterEvent`, stops in `onTileLeaveEvent` and `onTileRemoveEvent`. State survives service recreation via `companion object`.

## Checklist

## Tests

Pure-logic tests in `ShortcutsTileStateTest` cover:

- Resource-ID invariants (state suffixing, null/empty handling, mismatch detection)
- Snapshot immutability against later cache mutation
- LRU stash eviction and replacement
- The runtime race: cache mutation between `onTileRequest` and `onTileResourcesRequest` must not break the resource-ID invariant
- Optimistic click prediction per domain, plus safe no-op on unknown domains and unrecognized states

**I did extensive manual testing loading this onto my actual watch and updating my actual entities.  This was the only strategy that worked well.  The UI very difficult to get right.**

## Screenshots

**Fallback icons (no entity state):**

tile_screenshot

**State-aware icons (locked vs unlocked):**

tile_screenshot_stateaware

## Notes

- Icon resolution uses `Entity.getIcon()` (already the app-wide pattern) for cached entities and falls back to `getIcon(icon, domain, context)` when the cache is cold.
- Hardcoded state strings (`"on"`, `"off"`, `"locked"`, `"unlocked"`, `"open"`, `"closed"`) match existing uses throughout `Entity.kt` and the wear module — these are part of Home Assistant's public API.
- Unknown domains and unrecognized states are a silent no-op on the optimistic path: the tap still fires the broadcast, the display stays on the current state, and the subscription updates it when the server confirms.

### DIFF
--- a/app/src/main/res/xml/changelog_master.xml
+++ b/app/src/main/res/xml/changelog_master.xml
@@ -5,6 +5,7 @@
         <change>Bug fixes and dependency updates</change>
     </release>
     <release version="2026.5.4 - Wear" versioncode="2">
+        <change>Shortcut tiles now show state-aware icons (e.g., locked vs unlocked padlock)</change>
         <change>Bug fixes and dependency updates</change>
     </release>
     <release version="2026.5.4 - Automotive" versioncode="1">
--- a/wear/src/main/kotlin/io/homeassistant/companion/android/tiles/ShortcutsTile.kt
+++ b/wear/src/main/kotlin/io/homeassistant/companion/android/tiles/ShortcutsTile.kt
@@ -35,22 +35,24 @@ import com.mikepenz.iconics.utils.sizeDp
 import dagger.hilt.android.AndroidEntryPoint
 import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.common.R as commonR
+import io.homeassistant.companion.android.common.data.integration.getIcon
 import io.homeassistant.companion.android.common.data.prefs.WearPrefsRepository
 import io.homeassistant.companion.android.common.data.servers.ServerManager
 import io.homeassistant.companion.android.data.SimplifiedEntity
-import io.homeassistant.companion.android.common.data.integration.getIcon
 import io.homeassistant.companion.android.util.getIcon
 import java.nio.ByteBuffer
 import javax.inject.Inject
+import kotlin.coroutines.cancellation.CancellationException
 import kotlin.math.min
 import kotlin.math.roundToInt
-import kotlinx.coroutines.async
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.async
 import kotlinx.coroutines.guava.future
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
+import timber.log.Timber
 
 // Dimensions (dp)
 private const val CIRCLE_SIZE = 56f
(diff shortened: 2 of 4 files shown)


dshokouhi:
Has the situation described in https://github.com/home-assistant/android/issues/2045 get resolved now? Last I recall we could not reliably add this because its very easy for the state to get out of sync.

kdavh:
@dshokouhi  have a look at how I did it and tell me what you think.  It was very difficult to get the UI in sync with the actual state and this was the best I could do.  Realistically I've found that the state renders as indeterminate when first visiting the tile, but switches to correct state after first and subsequent taps, and after leaving and revisiting the tile.  I don't think it's perfect by any means, but given the testing I did on my devices and network, I'd use it.

TimoPtr:
Did you considered having a clear distinction for the user to know when it's in sync and when it's not? For a lock I don't want to see something cached I want to know if it's the real state or if it doesn't reflect the current state. As a user it would be quite important for me to know when looking at it.

kdavh:
I haven't forgotten about this, and hopefully will have time in the next month to try to get it closer to the goal.

kdavh:
I don't think this is possible to the exact standards we're hoping for unfortunately:

- [this tile refresh interval](https://developer.android.com/reference/kotlin/androidx/wear/tiles/TileBuilders.Tile.Builder?hl=en#setFreshnessIntervalMillis(long)) can only be used once per minute.
- how [this app](https://github.com/home-assistant/android/issues/2045#issuecomment-1918091431) does it probably works, but it uses a non-public "unofficial" function hack, so not suitable for a stable project like this.
- as has been noted, any other UI updates are up to the wear os system to decide on when they're scheduled

## The only things I could think to do that would be close to clear UX:

### haven't tested, but based on my reading that any user tap on tiles automatically triggers synchronous updates:
- the first time the tiles load, they should be accurate.  
- a user tap changes an IOT tile to a loading/indeterminate visual state.
- to view the actual state beyond that we'd have an "update" button at the bottom that the user taps to get fresh visual state

#### caveats
- this would rely on the calls being synchronous, i.e. being okay with a several second freeze if your HA is slow.  If something is async, I don't think a UI refresh is guaranteed.
- this would not guarantee a UI update if an IOT device changes states independently from the watch.  even though we can keep a websocket connection open, the data update won't guarantee a timely visual update.

### a carefully arranged list of [TimelineEntry](https://developer.android.com/training/wearables/tiles/update) could maybe be used to hack the updates, but seems like it would really stretch the intent and would likely have unforeseen pitfalls.
- user taps IOT button, TimelineEntry live for 3 seconds with another TimelineEntry waiting after it to force visual refresh


## So closing for the foreseeable future.

PR #6619, opened 2026-03-23
Add generic PushProvider abstraction for notification delivery

sk7n4k3d:
## Summary

- Introduces a `PushProvider` interface that abstracts push notification delivery, allowing users to choose their preferred provider
- Adds `PushProviderManager` for provider selection and server registration
- Implements `FcmPushProvider` (full flavor) and `WebSocketPushProvider` as built-in providers
- Adds a new **"Push provider"** setting under Notifications, always visible, letting users switch providers without app restart
- Extends `DeviceRegistration` with `pushUrl` and `pushEncrypt` fields for future provider support
- Adds `WebsocketManager.restart()` for instant WebSocket push channel resubscription on provider switch

This is the first of two PRs splitting the UnifiedPush work as requested in #6599. This PR contains only the generic abstraction layer — no UnifiedPush-specific code. The second PR (#TBD) will add the UnifiedPush implementation on top.

## Testing

- Tested on Pixel 9 Pro Fold (Android 16, GrapheneOS)
- Both full and minimal flavors compile and pass tests
- 28 unit tests for PushProvider interface and PushProviderManager

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

PR #6621, opened 2026-03-23
Add UnifiedPush support for push notifications

sk7n4k3d:
## Summary

- Implements UnifiedPush as an alternative push notification provider, allowing notifications via a distributor app (e.g. ntfy) without FCM
- Adds `UnifiedPushProvider`, `UnifiedPushManager`, `UnifiedPushReceiver`, and `UnifiedPushWorker`
- Extends the Push provider settings to list installed UnifiedPush distributors with their app names
- Adds `IgnoreViolationRule` for UnifiedPush connector library StrictMode violations (following the existing pattern for third-party libs)
- Handles protobuf-java/javalite conflict via exclusion
- Adds UnifiedPush guard in `FcmPushProvider.isActive()` and FCM `onNewToken()`

Depends on #6619 (PushProvider abstraction). This PR adds the UnifiedPush-specific implementation on top.

## Testing

Tested on Pixel 9 Pro Fold (Android 16, GrapheneOS) with ntfy as UnifiedPush distributor:

- Switch between WebSocket and UnifiedPush in both directions without app restart
- Notifications received on both providers in foreground, background, and with app killed
- UnifiedPush delivers via ntfy independently when app is killed
- Auto-fallback to WebSocket when ntfy topic is deleted
- No StrictMode/FailFast crashes
- 11 unit tests for UnifiedPush message parsing

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update


TimoPtr:
Let's first review the first PR.

sk7n4k3d:
Makes sense\! I've rebased this on top of the updated #6619 and ktlint passes now. Ready whenever #6619 is merged.

app/src/main/kotlin/io/homeassistant/companion/android/unifiedpush/UnifiedPushReceiver.kt
```diff
@@ -0,0 +1,61 @@
...
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.contentOrNull
+import org.unifiedpush.android.connector.FailedReason
+import org.unifiedpush.android.connector.MessagingReceiver
+import org.unifiedpush.android.connector.data.PushEndpoint
+import org.unifiedpush.android.connector.data.PushMessage
+import timber.log.Timber
+
+@AndroidEntryPoint
+class UnifiedPushReceiver : MessagingReceiver() {
```
p1gp1g: (Maintainer of UnififedPush lib)

Any reason to use MessagingReceiver instead of [PushService](https://unifiedpush.org/kdoc/connector/org.unifiedpush.android.connector/-push-service/)?

PushService is recommended over MessagingReceiver. If there is a reason, we may be able to improve the PushService

p1gp1g (commented):
Hey, just saw your PR: I've seen many people waiting for this feature, that's going to make some people happy 👍 

I just wanted to ask a question, cf. comment on UnifiedPushReceiver file

app/src/main/kotlin/io/homeassistant/companion/android/unifiedpush/UnifiedPushReceiver.kt
sk7n4k3d: Hey @p1gp1g, thanks for the drive-by from the lib maintainer itself! No good reason — I just built on an older pattern without checking for the newer API. You're right, `PushService` is the way.

Migrated in [2b648b7a](https://github.com/home-assistant/android/pull/6621/commits/2b648b7af3daad9d05f30bb43ae4464d28829d77):
- `UnifiedPushReceiver` → `UnifiedPushService` extending `PushService`
- Removed the 4 deprecated `<receiver>` actions from the manifest, replaced with a non-exported `<service>` filtering on `org.unifiedpush.android.connector.PUSH_EVENT`
- Hilt injection via `@AndroidEntryPoint` (the service is a proper `Context`, so cleaner overall than the old receiver approach)
- Dropped the `ExportedReceiver` lint baseline entries since the service is internal-only

Also dropped the `Context` parameter from the overrides (the `Service` *is* the context). Still on connector `3.0.9` — the `PushService` API is available from `3.0.0` so no version bump needed.

Thanks again for the heads up, this is a much cleaner integration.

p1gp1g: Ok, great ! Happy to see that feature for home assistant :+1: 

Did you have to do anything server side?

sk7n4k3d: Nothing server side! The Home Assistant `mobile_app` integration already supports arbitrary `push_url` / `push_token` values via `POST /api/mobile_app/registrations`, so UnifiedPush plugs in cleanly:

- On distributor registration, we send the endpoint returned by `onNewEndpoint` as `push_url`, leave `push_token` empty, and set `push_encrypt=true` when we have the public keys from the UnifiedPush connector.
- HA then posts notifications directly to that URL. For encrypted payloads we reuse the same WebPush-like format the FCM path uses (AES-GCM + ECDH with the keys we gave HA at registration time), so the same server logic handles both.
- When the user switches provider (e.g. UnifiedPush → WebSocket), we re-register to overwrite the `push_url` / `push_token`.

The one small gotcha I hit and just fixed in [de7243211](https://github.com/home-assistant/android/pull/6621/commits/de7243211df377825bd943b6321630860b0e25c4): HA's backend validates `push_url` as a real URL, so sending an empty string to clear the config yields HTTP 400. The fix falls back to the `BuildConfig.PUSH_URL` placeholder in that branch, which the WebSocket path was already using.

So from a HA core perspective: zero changes needed. This is purely an Android-side integration on top of the existing `mobile_app` push protocol.

p1gp1g: Interesting. So, if I understand correctly, it doesn't send *webpush* requests, but something that should be compatible. Did you test with Sunup ?

metaeaux:
I have been testing this and many of my notifications were dropped due to json parsing not working correctly. This patch has fixed it for me:
```
diff --git a/app/src/main/kotlin/io/homeassistant/companion/android/unifiedpush/UnifiedPushService.kt b/app/src/main/kotlin/io/homeassistant/companion/android/unifiedpush/UnifiedPushService.kt
index 1376fd80c..deed78b66 100644
--- a/app/src/main/kotlin/io/homeassistant/companion/android/unifiedpush/UnifiedPushService.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/unifiedpush/UnifiedPushService.kt
@@ -4,9 +4,13 @@ import dagger.hilt.android.AndroidEntryPoint
 import io.homeassistant.companion.android.notifications.MessagingManager
 import javax.inject.Inject
 import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.booleanOrNull
 import kotlinx.serialization.json.contentOrNull
+import kotlinx.serialization.json.doubleOrNull
 import org.unifiedpush.android.connector.FailedReason
 import org.unifiedpush.android.connector.PushService
 import org.unifiedpush.android.connector.data.PushEndpoint
@@ -35,11 +39,7 @@ class UnifiedPushService : PushService() {
         try {
             val jsonObject = Json.decodeFromString<JsonObject>(message.content.decodeToString())
             val data: Map<String, Any> = jsonObject.mapValues { (_, value) ->
-                if (value is JsonPrimitive) {
-                    value.contentOrNull ?: value.toString()
-                } else {
-                    value.toString()
-                }
+                jsonElementToAny(value)
             }
             messagingManager.handleMessage(data, SOURCE)
         } catch (e: Exception) {
@@ -64,5 +64,14 @@ class UnifiedPushService : PushService() {

     companion object {
         const val SOURCE = "UnifiedPush"
+
+        private fun jsonElementToAny(element: JsonElement): Any = when (element) {
+            is JsonPrimitive -> element.contentOrNull
+                ?: element.booleanOrNull
+                ?: element.doubleOrNull
+                ?: element.toString()
+            is JsonObject -> element.mapValues { (_, value) -> jsonElementToAny(value) }
+            is JsonArray -> element.map { jsonElementToAny(it) }
+        }
     }
 }
 ```

greglechin:
+1

slipstyle:
Is there anything I can do to help test this? I'm running a version of these modifications and it's working quite well...

I'd be happy to assist with testing and debugging and will be using this feature all the time.

Zepmann:
@sk7n4k3d 

I suspect many users are eagerly looking forward to this feature being integrated  in the HA Companion Application (myself included!). Your last update for this PR was four months ago. Do you have an update for this?

Not trying to rush you. Take whatever time you need. Just eager to see this happening. :smile:

PR #6625, opened 2026-03-24
Remove hardware limitation for wake word detection

TimoPtr:
## Summary

Removing the Hardware limitation for wake word detection by removing the dependency to LiteRT and building from source the Tflite micro frontend. This migrate the whole operations on the audio stream into the CPP making it more efficient. The code is heavily inspired from the one in ESPHome.

I've added some test on the failure of loading a model as I see some crashes on Sentry when people are injecting models in the APK that are not compatible. It is still going to crash but with proper logs explaining why.

I've added https://source.android.com/docs/security/test/hwasan in Debug for Arm64 (also in tests) to make sure we don't have any memory issues in the CPP code. I did test it by making an explicit leak and it crashes with a proper native stack trace.

I've removed all the code that was checking if the HW was not compatible.

After the PR every ABI will support wake word detection. The counterpart of this is that it increase the size of each arch for the full flavor (for minimal it is much better) but I think it is acceptable in terms of size.

For full arm64 we go from 880KB to 1,7MB (+93%)
image

For minimal we go from 17.8MB to 4.8MB (-73%)
image

## Checklist

## Any other notes

The differences from ESPHome is I tried to use unique_ptr whenever I can and obviously there is a JNI layer.

### DIFF
--- a/microwakeword/build.gradle.kts
+++ b/microwakeword/build.gradle.kts
@@ -18,10 +18,15 @@ android {
         }
     }
 
-    packaging {
-        jniLibs {
-            // Required for HWASan wrap.sh to be included uncompressed in the test APK
-            useLegacyPackaging = true
+    buildTypes {
+        debug {
+            // Required for HWASan wrap.sh to be included uncompressed in the APK
+            // See https://developer.android.com/ndk/guides/hwasan
+            packaging {
+                jniLibs {
+                    useLegacyPackaging = true
+                }
+            }
         }
     }
 


TimoPtr:
This seems to be more efficient in terms of battery usage, since we are making one less back and forth to the native side. But that might be hard to measure properly.

app/src/debug/resources/lib/arm64-v8a/wrap.sh
```diff
@@ -0,0 +1,5 @@
+#!/system/bin/sh
+# Enables HWAddressSanitizer (HWASan) for debug builds on ARM64 devices running Android 14+.
```
jpelgrom, line 2: If this is Android 14+ why didn't you place it in resources-v34?

jesserockz:
Works on the Echo Show 5 running Lineage OS 18.1 (Android 11)

common/src/main/res/values/strings.xml
```diff
@@ -1221,7 +1221,6 @@
     <string name="assist_wake_word_test_hint">Say \"%s\" to test</string>
     <string name="assist_wake_word_detected">Wake word detected!</string>
     <string name="assist_wake_word_battery_warning">Wake word detection runs entirely on your device, no network connection required. It continuously listens for your wake word, which will increase battery usage.</string>
-    <string name="assist_wake_word_unsupported_device">Wake word detection is not available on this device due to hardware limitations.</string>
     <string name="assist_wake_word_missing_play_services">Google Play Services is required to enable wake word detection.</string>
```
jpelgrom: I think this should also be possible to remove?

microwakeword/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,287 @@
...
+    tflite_micro
+    URL https://github.com/tensorflow/tflite-micro/archive/2747abd5c82a95fb1624106a946fc671c31f16e8.tar.gz # January 22nd 2026
+    URL_HASH SHA256=6ac0c8ef35f267cf3baf05dc094d4fd158864d4cc830d490b752994e58add8c8
+)
+
+# Header-only: flatbuffers for model parsing (version matching tflite-micro)
+FetchContent_Declare(
+    flatbuffers
+    GIT_REPOSITORY https://github.com/google/flatbuffers.git
+    GIT_TAG v23.5.26
```
jpelgrom: Can you use a sha hash instead here?

microwakeword/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,287 @@
...
+)
+set(FLATBUFFERS_BUILD_TESTS OFF CACHE BOOL "" FORCE)
+set(FLATBUFFERS_INSTALL OFF CACHE BOOL "" FORCE)
+set(FLATBUFFERS_BUILD_FLATC OFF CACHE BOOL "" FORCE)
+set(FLATBUFFERS_BUILD_FLATHASH OFF CACHE BOOL "" FORCE)
+
+# Header-only: gemmlowp for fixed-point math (commit matching tflite-micro)
```
jpelgrom, line 44: How do you know which commit matches tflite-micro?

app/gradle.lockfile
```diff
@@ -305,31 +305,32 @@ com.github.ajalt.clikt:clikt-jvm:5.0.2=ktlint
...
+com.google.ai.edge.litert:litert:2.1.0=minimalReleaseUnitTestCompileClasspath,minimalReleaseUnitTestRuntimeClasspath
```
jpelgrom: Where is LiteRT still coming from?

app/gradle.lockfile
```diff
@@ -305,31 +305,32 @@ com.github.ajalt.clikt:clikt-jvm:5.0.2=ktlint
...
-com.google.android.play:core-common:2.0.4=minimalDebugAndroidTestCompileClasspath,minimalDebugAndroidTestLintChecksClasspath,minimalDebugCompileClasspath,minimalDebugLintChecksClasspath,minimalDebugRuntimeClasspath,minimalDebugScreenshotTestCompileClasspath,minimalDebugScreenshotTestLintChecksClasspath,minimalDebugScreenshotTestRuntimeClasspath,minimalDebugUnitTestCompileClasspath,minimalDebugUnitTestLintChecksClasspath,minimalDebugUnitTestRuntimeClasspath,minimalReleaseCompileClasspath,minimalReleaseLintChecksClasspath,minimalReleaseRuntimeClasspath,minimalReleaseScreenshotTestCompileClasspath,minimalReleaseScreenshotTestLintChecksClasspath,minimalReleaseScreenshotTestRuntimeClasspath
+com.google.android.gms:play-services-tflite-impl:16.4.0=fullReleaseUnitTestCompileClasspath,fullReleaseUnitTestRuntimeClasspath
+com.google.android.gms:play-services-tflite-java:16.4.0=fullReleaseUnitTestCompileClasspath,fullReleaseUnitTestRuntimeClasspath
```
jpelgrom: Likewise why are these still included?

app/src/debug/resources/lib/arm64-v8a/wrap.sh
TimoPtr: Since it debug I didn't care that much that we have this extra script that does nothing harmful except setting a flag.

app/gradle.lockfile
TimoPtr: Leftover to clean up properly the lockfile you need to fully remove it and re-generate it otherwise it left some previous artifact for some reason...

microwakeword/src/main/cpp/CMakeLists.txt
TimoPtr: Well I went into the tflite-micro repository and check the dependencies, it's quite easy since they use Bazel. I didn't put a link to where to find it because I already saw that they've moved the deps into another file... So if we ever update we should do the same exercise as I did. 

My dream would be to use bazel within our project to not have to find the version at all, but for now I don't see any plan from Google to move away from Gradle/Cmake in Android apps.

app/src/debug/resources/lib/arm64-v8a/wrap.sh
TimoPtr: Also I'm not sure flags works on resources folder is it?

TimoPtr:
> With the increasing amount of C++ code, it would be nice to better document _exactly_ where it's coming from (including links to code in repositories) so it is possible to keep track of when it needs to be updated/adjusted because of upstream changes.
> 
> (This is a quick first scan, not a complete review.)

I didn't want to link to the exact code of esphome just mention the file in case it changes I don't want to have to update the permalink each time. Also it's a slightly bit different than esphome.

app/src/main/res/xml/changelog_master.xml
```diff
@@ -2,6 +2,7 @@
 <changelog xmlns:tools="http://schemas.android.com/tools"
     tools:ignore="MissingDefaultResource">
     <release version="2026.3.8 - Main" versioncode="3">
+        <change>"Wake word detection no longer requires specific hardware</change>
```
TimoPtr: @jpelgrom I think it is important to mention

jpelgrom: Yes, definitely. But do we support mips now ;)? Maybe easier to understand for users even if technically less correct: "Wake word detection is now supported on older hardware".

TimoPtr: Long time since I read about this arch ... I don't like to mention "older" since well in theory we could have ARMv7 for exotic devices ... MIPS is not supported since NDK r17 same as ARMv5

jpelgrom: But if we have ARMv7 for exotic devices, it still would not work on that device - so your entry would also be incorrect, no?

app/src/debug/resources/lib/arm64-v8a/wrap.sh
jpelgrom: You're right, I mixed up res vs resources 🤦

app/src/main/res/xml/changelog_master.xml
TimoPtr: Armv7 now works alongside x86 with this PR, we are covering all the ABI android is targeting since Android 6 if I'm not mistaken.

microwakeword/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,289 @@
...
+# Global optimization flags (applied to all targets including fetched dependencies)
+add_compile_options(-O3 -ffunction-sections -fdata-sections)
+add_link_options(-Wl,--gc-sections)
+
+include(FetchContent)
+
+FetchContent_Declare(
+    kissfft
+    GIT_REPOSITORY https://github.com/mborgerding/kissfft.git
+    GIT_TAG 131.2.0
```
jpelgrom: One more that should be pinned to the sha

microwakeword/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,289 @@
...
+FetchContent_Declare(
+    tflite_micro
+    URL https://github.com/tensorflow/tflite-micro/archive/2747abd5c82a95fb1624106a946fc671c31f16e8.tar.gz # January 22nd 2026
+    URL_HASH SHA256=6ac0c8ef35f267cf3baf05dc094d4fd158864d4cc830d490b752994e58add8c8
+)
+
+# Header-only: flatbuffers for model parsing (version matching tflite-micro)
+FetchContent_Declare(
+    flatbuffers
+    GIT_REPOSITORY https://github.com/google/flatbuffers.git
```
jpelgrom: If an archive download is faster why not do it here as well?

TimoPtr: True we can

TimoPtr:
> I'm still not 100% certain I can do good reviews of C++ code but it works in an emulator for the other architecture and the text looks OK (and Kotlin code is good).
> 
> Initially the debug app felt a bit laggy with these changes, possibly because of HWASan? After a reboot it was responsive again.
> 
> Again merge conflicts to resolve.

Unfortunately I'm not sure we are going to get a C++ deva available for the review. That's also why I've made this extra work on HWASan. It does indeed has a quite significant impact on the code execution.

microwakeword/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,289 @@
...
+    android
+    log
+)
+
+# HWAddressSanitizer for debug builds on ARM64 only.
+# Requires c++_shared STL (NDK default since r25, c++_static is incompatible).
+# See https://developer.android.com/ndk/guides/hwasan
+if(CMAKE_BUILD_TYPE STREQUAL "Debug" AND CMAKE_ANDROID_ARCH_ABI STREQUAL "arm64-v8a")
+    target_compile_options(${CMAKE_PROJECT_NAME} PUBLIC -fsanitize=hwaddress -fno-omit-frame-pointer)
+    target_link_options(${CMAKE_PROJECT_NAME} PUBLIC -fsanitize=hwaddress)
```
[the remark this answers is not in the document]
TimoPtr: It's fine for now since the library is not meant to be linked against.

microwakeword/src/main/kotlin/io/homeassistant/companion/android/microwakeword/MicroWakeWord.kt
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    /**
+     * Safety net to release native resources if [close] was not called.
+     * Prefer calling [close] explicitly or using a `use` block.
+     */
+    protected fun finalize() {
+        close()
+    }
```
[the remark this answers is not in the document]
TimoPtr: it's already the case.

TimoPtr:
@jpelgrom if you can take a final look we should be good to merge.

microwakeword/build.gradle.kts
```diff
@@ -18,9 +16,12 @@ android {
...
             }
         }
+    }
 
-        ndk {
-            abiFilters += listOf("arm64-v8a", "x86_64") // Limit to LiteRt arch supported for Android
+    packaging {
+        jniLibs {
+            // Required for HWASan wrap.sh to be included uncompressed in the test APK
+            useLegacyPackaging = true
```
jpelgrom: Shouldn't this be scoped to the debug buildtype like in app? (Suppressed low-confidence comment from Copilot mentions this as well)

TimoPtr: It only impacts the library it was needed for the instrumentation test, but you are right I'm going to fix it.

Issue #6206, opened 2025-12-28
No TTS voice response in Assist chat due to mTLS


tomaciazek:
**Home Assistant Android app version(s):**
2025.11.4-full
**Android version(s):**
16
**Device model(s):**
Samsung Galaxy S22
**Home Assistant version:**
2025.12.4
**Last working Home Assistant release (if known):**
N/A
**Description of problem, include YAML if issue is related to notifications:**
When using mTLS authentication through Cloudflare, the Assist chat fails on playback of TTS recording provided by Piper. With mTLS protection policy disabled on Cloudflare, everything works fine. Surprisingly, when launching an audio debug stream, it works with mTLS enabled without issues too - I guess the debug mode is using different code / libraries for audio fetching / playback.

**Companion App Logs:**

```
12-28 17:22:01.996  3667  3756 D WebSocketCoreImpl: EventSocketResponse(id=14, event={"type":"tts-start","data":{"engine":"tts.piper","language":"pl_PL","voice":"pl_PL-gosia-medium","tts_input":"Wybacz, niestety nie mogę tego zrozumieć","acknowledge_override":false},"timestamp":"2025-12-28T16:22:02.217183+00:00"})
12-28 17:22:01.998  3667  3756 D WebSocketCoreImpl: Unknown event type ignoring. received data = 
12-28 17:22:01.998  3667  3756 D WebSocketCoreImpl: EventSocketResponse(id=14, event={"type":"run-end","data":null,"timestamp":"2025-12-28T16:22:02.217652+00:00"})
12-28 17:22:02.059  3667  3667 D ServerConnectionInfo: Using external URL
12-28 17:22:02.061  3667  3686 V MediaPlayer-JNI: native_setup
12-28 17:22:02.062  3667  3686 V MediaPlayerNative: constructor
12-28 17:22:02.063  3667  3686 V MediaPlayerNative: setListener
12-28 17:22:02.063  3667  3686 V MediaPlayer-JNI: get_session_id()
12-28 17:22:02.065  3667  5384 D WebSocketCoreImpl$createSubscriptionFlow$flow: Unsubscribing from {type=assist_pipeline/run, start_stage=stt, end_stage=tts, input={sample_rate=44100}, conversation_id=null, pipeline=01kddce33x4neej39jakbcb459, device_id=adf755451a8711d90e6292569d51b168}
[...]

PR #6630, opened 2026-03-26
Use okhttp when mtls setup

TimoPtr:
## Summary

Cronet and HttpEngine DataSources seems to not support mTLS (at least our setup). See #6206 since we don't have a reliable way to know if a server is using mTLS, I've decided that if at least one server is using mTLS we fallback to OkHttp datasource.

I'm not a big fan of this implementation but I don't see any other way. 

## Checklist

## Any other notes

@yschimke maybe you have an idea since you were the one helping on this part.


common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -51,25 +50,11 @@ class TLSHelper @Inject constructor(
             }
 
             override fun getCertificateChain(p0: String?): Array<X509Certificate>? {
-                var chain: Array<X509Certificate>?
-
-                // block until a chain is provided via the TLSWebView
-                runBlocking {
```
yschimke, line 57: Nice cleanup of the runBlocking

PR #6635, opened 2026-03-26
Add haptic message support for FrontendScreen

TimoPtr:
## Summary

Add support for haptic message with a fallback when haptic is unknown.

## Checklist

## Any other notes

I tested the feedbacks one by one and I'm not able to tell the difference ... It's very subtle

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -1383,7 +1383,7 @@ class WebViewActivity :
 
             "warning" -> {
                 if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
-                    vm?.vibrate(VibrationEffect.createOneShot(400, VibrationEffect.EFFECT_HEAVY_CLICK))
+                    vm?.vibrate(VibrationEffect.createPredefined(VibrationEffect.EFFECT_HEAVY_CLICK))
                 } else {
                     @Suppress("DEPRECATION")
                     vm?.vibrate(1500)


app/src/main/kotlin/io/homeassistant/companion/android/frontend/haptic/HapticFeedbackPerformer.kt
```diff
@@ -0,0 +1,89 @@
...
+            view.performHapticFeedback(HapticFeedbackConstants.CONFIRM)
+        } else {
+            vibrate(view, duration = SUCCESS_FALLBACK_DURATION)
+        }
+    }
+
+    private fun performWarning(view: View) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            view.context.getSystemService<Vibrator>()?.vibrate(
+                VibrationEffect.createPredefined(VibrationEffect.EFFECT_HEAVY_CLICK),
```
TimoPtr, line 61: This one is different than the actual WebView 

```kotlin
"warning" -> {
                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
                    vm?.vibrate(VibrationEffect.createOneShot(400, VibrationEffect.EFFECT_HEAVY_CLICK))
                } else {
                    @Suppress("DEPRECATION")
                    vm?.vibrate(1500)
                }
            }
```

I think it was wrong in the Webview since the second param is an amplitude not an effect. (Claude spotted this one)

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
```diff
@@ -461,6 +471,11 @@ private fun WebViewEffects(webView: WebView?, url: String, scriptsToEvaluate: Fl
                 }
             }
         }
+        LaunchedEffect(webView) {
+            hapticEvents.collect { hapticType ->
+                HapticFeedbackPerformer.perform(webView, hapticType)
+            }
+        }
```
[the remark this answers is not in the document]
TimoPtr: I think it is valid to key on the webview but happy to discuss with @jpelgrom

app/src/test/kotlin/io/homeassistant/companion/android/frontend/haptic/HapticFeedbackPerformerTest.kt
```diff
@@ -0,0 +1,90 @@
...
+import io.mockk.mockk
+import io.mockk.verify
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
+@Config(application = HiltTestApplication::class, sdk = [Build.VERSION_CODES.R])
+class HapticFeedbackPerformerTest {
```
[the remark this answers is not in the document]
TimoPtr: I'm ok not testing the fallback

app/src/main/kotlin/io/homeassistant/companion/android/frontend/haptic/HapticFeedbackPerformer.kt
jpelgrom: If it is wrong on the WebView(Activity, I'm assuming), why not change it there as well?

TimoPtr: Done 👍🏻

TimoPtr:
> In the end I had to create a custom card to test these as it appears the frontend primarily uses `light` and `failure` with a couple others in very specific use cases, and not everything? A bit annoying but it seemed to work correctly.
> 
> What's interesting is that `light` sometimes feels different - noticed this while repeatedly turning a light on/off and it's almost as if the phone changes the effect depending on how strongly I'm tapping / how tightly I'm holding the phone? I confirmed it's sending the same call and there is no doubling or something like that, and the same happens in WebViewActivity, so out of scope here.
> 
> Given how `HapticFeedbackPerformer` is standalone, would it make sense to replace the code in `WebViewActivity` to point to this class as well (replacing the when branches in `fun processHaptic` with calls to `HapticFeedbackPerformer.perform(webView, hapticType)`, not introducing the external bus parsing)? That would help reduce the size of that activity while you continue working on the new Compose-based screens.

I agree with you, the only issue I see is that we need to deserialize the hapticType from the JSON and I would like to avoid doing this because it would leak `frontendExternalBusJson` into the WebViewActivity. Maybe in another PR we can come with something nice. I'm merging this one like this.

PR #6638, opened 2026-03-27
Wear: use coroutines from GMS tasks

TimoPtr:
## Summary

While looking on the Play Store console, our biggest crash (94% of them)
image
is due to the fact that we are not handling proper ApiException in the OnboardingActivity. This PR solves this but also migrate to proper coroutines instead of using new threads. I also took the opportunity to also review the coroutine scopes in the listeners used to communicate with the mobile app.

## Checklist

## Any other notes

I also remove the Suppress annotation about VisibleForTest since it has been addressed in the upstream.

### DIFF
--- a/app/src/full/kotlin/io/homeassistant/companion/android/onboarding/WearOnboardingListener.kt
+++ b/app/src/full/kotlin/io/homeassistant/companion/android/onboarding/WearOnboardingListener.kt
@@ -8,12 +8,7 @@ import com.google.android.gms.wearable.WearableListenerService
 import dagger.hilt.android.AndroidEntryPoint
 import io.homeassistant.companion.android.common.data.servers.ServerManager
 import javax.inject.Inject
-import kotlinx.coroutines.CancellationException
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.SupervisorJob
-import kotlinx.coroutines.cancel
-import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.tasks.await
 import timber.log.Timber
 
@@ -22,12 +17,6 @@ class WearOnboardingListener : WearableListenerService() {
 
     @Inject
     lateinit var serverManager: ServerManager
-    private val serviceScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
-
-    override fun onDestroy() {
-        serviceScope.cancel()
-        super.onDestroy()
-    }
 
     override fun onMessageReceived(event: MessageEvent) {
         Timber.d("onMessageReceived: $event")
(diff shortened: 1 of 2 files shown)


app/src/full/kotlin/io/homeassistant/companion/android/onboarding/WearOnboardingListener.kt
```diff
@@ -28,29 +38,29 @@ class WearOnboardingListener : WearableListenerService() {
         }
     }
 
-    private fun sendHomeAssistantInstance(nodeId: String) = runBlocking {
```
TimoPtr, line 31: One less runBlocking

wear/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingActivity.kt
```diff
@@ -191,69 +197,83 @@ class OnboardingActivity :
         }
     }
 
-    private fun findExistingInstances() {
+    private suspend fun findExistingInstances() = withContext(Dispatchers.IO) {
         Timber.d("findExistingInstances")
-        Tasks.await(Wearable.getDataClient(this).getDataItems("wear://*/home_assistant_instance".toUri())).apply {
```
TimoPtr, line 196: This most probably the cause of the crashes on the play console but I'm not 100% since the stacktrace is not mentioning anything from our side.

wear/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingActivity.kt
```diff
@@ -191,69 +197,89 @@ class OnboardingActivity :
...
+                    if (t == null) {
+                        Timber.d(
+                            "requestInstances: request home assistant instances from ${node.id}: ${node.displayName}",
+                        )
+                    } else {
+                        Timber.w(
+                            t,
+                            "requestInstances: failed to request home assistant instances from ${node.id}: ${node.displayName}",
+                        )
+                    }
```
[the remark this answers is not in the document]
TimoPtr: let's keep it for now

wear/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingActivity.kt
```diff
@@ -191,69 +197,89 @@ class OnboardingActivity :
...
-    private fun requestInstances() {
+    private suspend fun requestInstances() = withContext(Dispatchers.IO) {
         Timber.d("requestInstances")
 
-        // Find all nodes that are capable
-        val capabilityInfo: CapabilityInfo = Tasks.await(
-            capabilityClient.getCapability(
+        try {
+            // Find all nodes that are capable
+            val capabilityInfo: CapabilityInfo = capabilityClient.getCapability(
```
jpelgrom, line 226: Type can be inferred

```suggestion
            val capabilityInfo = capabilityClient.getCapability(
```

wear/src/main/kotlin/io/homeassistant/companion/android/onboarding/OnboardingActivity.kt
```diff
@@ -191,69 +197,89 @@ class OnboardingActivity :
...
-    private fun requestPhoneSignIn() {
+    private suspend fun requestPhoneSignIn() = withContext(Dispatchers.IO) {
         Timber.d("requestPhoneSignIn")
 
-        // Find all nodes that are capable
-        val capabilityInfo: CapabilityInfo = Tasks.await(
-            capabilityClient.getCapability(
+        try {
+            // Find all nodes that are capable
+            val capabilityInfo: CapabilityInfo = capabilityClient.getCapability(
```
jpelgrom, line 265:
```suggestion
            val capabilityInfo = capabilityClient.getCapability(
```

wear/src/main/kotlin/io/homeassistant/companion/android/phone/PhoneSettingsListener.kt
```diff
@@ -78,7 +79,12 @@ class PhoneSettingsListener :
     @Inject
     lateinit var messagingTokenProvider: MessagingTokenProvider
 
-    private val mainScope: CoroutineScope = CoroutineScope(Dispatchers.Main + Job())
+    private val serviceScope: CoroutineScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
+
+    override fun onDestroy() {
+        serviceScope.cancel()
+        super.onDestroy()
+    }
```
TimoPtr: It could be here that we cancel but it means that the service is not alive very long.

TimoPtr: It means that we were leaking the service before.

TimoPtr: @jpelgrom maybe we could use the WorkManager here ? 🤔 with the risk that it is not executed right away? Otherwise we might need to do this listening in the app itself and forbid it to go to sleep so we keep listening (I don't know if it's possible) until we are setup.

jpelgrom: It might have leaked before (and/or Samsung might be limiting more here than technically should happen), but it does work with the leaking code. Clean code but it not working is worse.

> @jpelgrom maybe we could use the WorkManager here ? 🤔 with the risk that it is not executed right away? 

This is in response to the user trying to sign in so "not executed right away" is undesirable. It also ends with launching an activity which cannot be done from a job. However early cancellation in the other functions is also bad, changes may not be saved. Maybe it is completing early because of launch - can we make it wait until done? (I didn't find it in a quick research, maybe also need experimentation/checking the code of the GMS listener service.)

TimoPtr: I was pretty sure I did the onboarding but on the emulator maybe it changes. 
Do we really need a service? Is there something to listen within an activity so we have a better control on the lifecycle?

jpelgrom: [You can do it in the activity](https://developer.android.com/training/wearables/data/events#use-live-listener). However, onboarding on the phone takes a bit of time and watches very quickly turn off by default / if you put your wrist down so we need to be sure something is alive to send the onboarding data back to.

TimoPtr: Maybe we can keep the watch on on this activity? (I'm not sure it's possible)

jpelgrom: Maybe, something to try :)

TimoPtr: In the end I decided to revert this part and document it. I kept the changes needed to properly catch the error so it doesn't crash the app. If you can double check.

jpelgrom: Double checked and can confirm it now works again while there is progress to move towards coroutines :). Good enough for now in my opinion, otherwise this PR scope increases a lot.

PR #6640, opened 2026-03-27
Define CommandMessage pattern and implement first messages

TimoPtr:
## Summary

Come up with a pattern to define with a strong type the Commands that we send to the frontend that uses the `command` JSON discriminator. Since a sealed class member cannot have twice the same discriminator (otherwise it cannot de-serialize properly), I had to come up with this pattern where we override the `invoke` operator of an object to keep a nice API.

## Checklist

## Any other notes

I don't like it but I didn't find any other way for the Payload to be a JsonElement, using a template breaks the serialization engine.

This PR is based on #6708 and need to be rebased once #6708 has been merged.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/externalbus/outgoing/CommandMessage.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/externalbus/outgoing/CommandMessage.kt
@@ -1,0 +1,65 @@
+package io.homeassistant.companion.android.frontend.externalbus.outgoing
+
+import io.homeassistant.companion.android.frontend.externalbus.frontendExternalBusJson
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.encodeToJsonElement
+
+/**
+ * Outgoing command message sent to the Home Assistant frontend via the external bus.
+ *
+ * Command messages instruct the frontend to perform an action.
+ *
+ * The Home Assistant frontend uses `"type": "command"` for all commands, distinguishing them
+ * only by the `command` string field. This means kotlinx.serialization's sealed polymorphism
+ * cannot give each command its own subtype (all would need the same `@SerialName("command")`).
+ *
+ * Instead, this single private class handles serialization, and each command is exposed as a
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/frontend/externalbus/outgoing/CommandMessage.kt
```diff
@@ -0,0 +1,65 @@
...
+    @Serializable
+    private data class NavigateOptions(val replace: Boolean = false)
+}
+
+/**
+ * Command to toggle the frontend sidebar visibility.
+ *
+ * @see CommandMessage
+ */
+val ShowSidebarMessage: OutgoingExternalBusMessage = CommandMessage(command = "sidebar/show")
```
jpelgrom, line 65: Do you want to do a native sidebar to work around edge-to-edge limitations? 👀 (It has always felt like a change that's not worth it to me, creating a weird hybrid especially when considering breakpoints/zoom levels/theming).

TimoPtr: I don't want that, this is extracted from the current code TBH and I didn't even looked at how it is used. I suppose here it could be use to map a gesture to open the sidebar. It can be useful for someone that has a custom theme maybe without a button to open the sidebar.

TimoPtr: Actually that used in gesture https://github.com/home-assistant/android/blob/ea76f7ed055332d06aa352845a84fab40e880857/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt#L1244

jpelgrom: Oh of course, we added a gesture for it as well

jpelgrom:
TIL about `operator fun invoke`, thanks 😊

jpelgrom (approved):

TimoPtr:
> TIL about `operator fun invoke`, thanks 😊

To be fair this design is quite hacky but that's the only way I found to keep it nice and aligned with the rest.

PR #6648, opened 2026-03-29
Fix bottom sheet peek-to-expanded jump on show

loganrosen:
## Summary

When opening a bottom sheet (e.g. the server chooser via 3-finger swipe up), the sheet visibly jumps from peek height to expanded. This is because `setLayoutAndExpandedByDefault()` configures `STATE_EXPANDED` in an `onShowListener`, which fires *after* the sheet has already started its show animation at peek height.

This changes the approach to configure the `BottomSheetBehavior` directly in `onViewCreated` (before `onStart` shows the dialog), so the sheet starts directly in the expanded state without an intermediate peek. Also adds `skipCollapsed` so dismissal goes straight to hidden instead of stopping at collapsed.

Affects all callers: `ServerChooserFragment`, `ImprovSetupDialog`, and `ImprovPermissionDialog`.

## Checklist

## Any other notes

No tests added — the fix is about *when* behavior properties are configured (before vs after dialog show), which is a timing concern that can't be meaningfully validated in a unit test. Tested manually on device.

Note: to test this on a debug build, I had to temporarily set `WIPFeature.USE_FRONTEND_V2 = false` because the new Compose frontend (`FrontendScreen`/`HAWebView`) doesn't have gesture detection yet — the swipe listener is only set up in the legacy `WebViewActivity` path.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/util/FragmentExtensions.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/util/FragmentExtensions.kt
@@ -1,11 +1,8 @@
 package io.homeassistant.companion.android.util
 
-import android.view.View
 import android.view.WindowManager
-import android.widget.FrameLayout
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.Lifecycle
-import com.google.android.material.R
 import com.google.android.material.bottomsheet.BottomSheetBehavior
 import com.google.android.material.bottomsheet.BottomSheetDialog
 import com.google.android.material.bottomsheet.BottomSheetDialogFragment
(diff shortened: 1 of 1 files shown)


TimoPtr:
@loganrosen do you mind adding 2 videos in the PR description showing what is changing?

jpelgrom:
> @loganrosen do you mind adding 2 videos in the PR description showing what is changing?

You just beat me to it. I spent the last 15 minutes fiddling with the animation scale and while the change sounds logical, I'm not seeing any difference.

loganrosen:
@jpelgrom @TimoPtr Sure thing, should've included these before: 

After:
https://github.com/user-attachments/assets/6040c900-f0a9-4113-a550-91235fd06a7d

Before:
https://github.com/user-attachments/assets/e51363e6-72cc-48c1-ac97-e8be28d8fcee

jpelgrom:
Interesting, I'm not seeing it jump like that without your changes on my devices. At the same time no objection to changing it so it might work better for more people.

jpelgrom (approved):

TimoPtr (approved):

TimoPtr:
I don't see this issue too but fine merging it if it helps some.

PR #6659, opened 2026-03-30
Use Lokalise action and async download

TimoPtr:
## Summary

Since today we are facing a [413](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/413) error on Lokalize and the recommend way to download big projects is to use the Async way. Fortunately they have a [Github Action](https://github.com/lokalise/lokalise-pull-action) that can do that for us.

I've made a dumb commit to show that the French translations were properly downloaded in the right place.
image

A new job has been added to the PR workflow that run only if the `.github/actions/download-translations` has a change. We use a fake Github token to make sure that if one day the behavior of the action changes we catch it when Renovate opens a PR. This job will start and stop if no modification are made so that we can have a Github checks on the PR.

## Checklist

## Any other notes

Zizmor is not reporting any issue

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -25,7 +25,7 @@ jobs:
         id: changes
         shell: bash
         run: |
-          if git diff --name-only "origin/main..HEAD" -- '.github/actions/download-translations/' | grep -q .; then
+          if git diff --name-only "origin/main...HEAD" -- '.github/actions/download-translations/' | grep -q .; then
             echo "Lokalise setup changes detected"
             echo "skip=false" >> "$GITHUB_OUTPUT"
           else


jpelgrom:
This may work but looking at the suggested use cases in the action README it should be used to sync translations to the repository, and the action also has a lot of functionality to automatically create pull requests etc. We may end up with issues in the future, what do you think?

Edit: also shown by the [log for the action](https://github.com/home-assistant/android/actions/runs/23755948891/job/69210588122#step:3:156) test run 'no changes detected'.
```
Downloading translation files from Lokalise...
Starting download from Lokalise
Download complete! Detecting changed files...
No changes detected in translation files.
No changes detected, nothing to commit.
```

.github/actions/download-translations/action.yml
```diff
@@ -10,36 +10,27 @@ inputs:
...
+        skip_include_tags: true
+        skip_original_filenames: true
+        additional_params: |
+          {
+            "original_filenames": false,
+            "bundle_structure": "common/src/main/res/values-%LANG_ISO%/strings.%FORMAT%",
+            "add_newline_eof": true,
+            "replace_breaks": true,
+            "export_empty_as": "skip"
+          }
```
jpelgrom: This [can be yaml](https://github.com/lokalise/lokalise-pull-action?tab=readme-ov-file#download-options) which I think should be preferred when using a 'native' action instead of mixing in JSON?

TimoPtr:
> This may work but looking at the suggested use cases in the action README it should be used to sync translations to the repository, and the action also has a lot of functionality to automatically create pull requests etc. We may end up with issues in the future, what do you think?
> 
> Edit: also shown by the [log for the action](https://github.com/home-assistant/android/actions/runs/23755948891/job/69210588122#step:3:156) test run 'no changes detected'.
> ```
> Downloading translation files from Lokalise...
> Starting download from Lokalise
> Download complete! Detecting changed files...
> No changes detected in translation files.
> No changes detected, nothing to commit.
> ```

I saw it too but to do the comparaison they need to download so I think we should be fine and it is a simple wrapper around the API. In the worse case we would fallback to the CLI.

.github/actions/download-translations/action.yml
TimoPtr: I'll do it tomorrow or feel free to push on my branch if you want to merge it.

jpelgrom:
> I saw it too but to do the comparaison they need to download so I think we should be fine and it is a simple wrapper around the API. In the worse case we would fallback to the CLI.

I'd like to avoid the possibility of it (accidentally) doing more in the future with an update though 🤔 We can specify `custom_github_token` and pass a false token to prevent it from doing any Git operations, or is that too much ? (In the workflows where it's used the token only has `contents` permission, but still.)

TimoPtr:
> > I saw it too but to do the comparaison they need to download so I think we should be fine and it is a simple wrapper around the API. In the worse case we would fallback to the CLI.
> 
> I'd like to avoid the possibility of it (accidentally) doing more in the future with an update though 🤔 We can specify `custom_github_token` and pass a false token to prevent it from doing any Git operations, or is that too much ? (In the workflows where it's used the token only has `contents` permission, but still.)

What about pulling the translation and make some assertions like is the french translations present? We could maybe do it on every commit or make it something we can start manually so when a new version is released we can trigger it before reaching main.

jpelgrom:
If we start doing so much work to verify it might be easier to just use the CLI already...

... but your solution would work to make sure that it continues downloading the files and not discard them immediately 👍. My point is still stands to avoid it doing Git(Hub) operations that reach GitHub.

TimoPtr:
> If we start doing so much work to verify it might be easier to just use the CLI already...
> 
> ... but your solution would work to make sure that it continues downloading the files and not discard them immediately 👍. My point is still stands to avoid it doing Git(Hub) operations that reach GitHub.

The CLI is annoying me a bit because we would need to make something custom for renovate to catch up the version of the CLI that we get with curl or we would use 'main' and I don't like not having a reproductible build.

I'll try your idea about the token it's a good one. When do you think it should run?

jpelgrom:
> I'll try your idea about the token it's a good one. When do you think it should run?

I'm presuming the question doesn't relate to the previous sentence because otherwise it's very confusing. Your suggested 'translation file exists' check: workflow scoped to when the download-translations file (action) changes, which happens on updates of Lokalise GitHub action but prevents it from running in other scenarios.

TimoPtr:
> > I'll try your idea about the token it's a good one. When do you think it should run?
> 
> I'm presuming the question doesn't relate to the previous sentence because otherwise it's very confusing. Your suggested 'translation file exists' check: workflow scoped to when the download-translations file (action) changes, which happens on updates of Lokalise GitHub action but prevents it from running in other scenarios.

Indeed sorry for the confusion. Not sure I can do that, is it an issue to do it on the PR workflow?

jpelgrom:
> Not sure I can do that, is it an issue to do it on the PR workflow?

You can do it on PR [with the `paths` option](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore).

On every PR run could also work, I was just trying to be mindful of Lokalise API usage.

TimoPtr:
@jpelgrom once this is merged if they break something in the action we should be able to see it.

.github/actions/download-translations/action.yml
```diff
@@ -10,36 +10,26 @@ inputs:
...
+    - name: Pull from Lokalise
+      uses: lokalise/lokalise-pull-action@a4968e06b7f44c32317b162d825ed17abb5772d0 # v5.0.1
+      with:
+        api_token: ${{ inputs.lokalise-token }}
+        project_id: ${{ inputs.lokalise-project }}
+        custom_github_token: "FAKE_GH_TOKEN"
+        file_format: xml
+        async_mode: true
+        skip_include_tags: true
+        skip_original_filenames: true
```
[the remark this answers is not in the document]
TimoPtr: It's made on purpose since we don't want Lokalise to push anything by mistake.

.github/workflows/pr.yml
```diff
@@ -13,6 +13,47 @@ permissions:
...
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - uses: ./.github/actions/download-translations
+        if: steps.changes.outputs.skip != 'true'
+        name: Download Translations
+        with:
+          lokalise-project: ${{ secrets.LOKALISE_PROJECT }}
+          lokalise-token: ${{ secrets.LOKALISE_TOKEN }}
+
```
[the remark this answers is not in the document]
TimoPtr: It's fine that it fails on forks it's on purpose too, only Renovate that runs within this repository is going to trigger this workflow.

.github/workflows/pr.yml
```diff
@@ -13,6 +13,47 @@ permissions:
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Check for Lokalise setup changes
+        id: changes
+        shell: bash
+        run: |
+          if git diff --name-only "origin/main..HEAD" -- '.github/actions/download-translations/' | grep -q .; then
```
jpelgrom: Use `...` in case it's not directly on main, similar to PRs

```suggestion
          if git diff --name-only "origin/main...HEAD" -- '.github/actions/download-translations/' | grep -q .; then
```

.github/actions/download-translations/action.yml
```diff
@@ -10,36 +10,26 @@ inputs:
...
-                "add_newline_eof": true,
-                "replace_breaks": true,
-                "export_empty_as": "skip"
-            }' \
-          | jq -r '.bundle_url' \
-        `
-
-        echo "Download Ready: $ZIP"
-
-        curl "$ZIP" --output strings.zip
```
TimoPtr, line 37: I removed this line while going over the script this should be brought back.

jpelgrom:
It appears our Git tree is now dirty and we get messy version numbers... Didn't notice it before as an internal user but it's the same for published releases. ([Example](https://github.com/home-assistant/android/issues/6673))

PR #6663, opened 2026-03-31
fix: pre-initialize mTLS flag from stored alias to handle TLS session resumption in Wear OS onboarding

smhc:
# fix: Pre-initialize mTLS flag from stored key alias to handle TLS session resumption in Wear OS onboarding

## Problem

When setting up a Wear OS device through the Home Assistant phone app, the mTLS certificate
selection screen is silently skipped, causing the watch to register without a client certificate
and the server to respond with `HTTP 400`.

The root cause is **TLS session resumption**:

1. The main HA app WebView authenticates with the server — a TLS session ticket is created.
2. The user opens the Wear OS onboarding flow (either from the watch or via the phone app).
3. The onboarding WebView connects to the _same_ host; the OS TLS stack resumes the existing
   session (abbreviated handshake — no new `CertificateRequest` from the server).
4. `WebViewClient.onReceivedClientCertRequest()` is never invoked.
5. `TLSWebViewClient.isTLSClientAuthNeeded` remains `false`.
6. `requiredMTLS = false` is passed through `ConnectionNavigationEvent.Authenticated` →
   `OnboardingNavigation` skips `wearMTLSScreen`.
7. The watch is registered without a cert → server returns `400 Bad Request`.

This happens regardless of whether `ssl_session_tickets` / `ssl_session_cache` are disabled on
the server, because Android's Chromium network stack also maintains its own in-process TLS
session cache that persists across WebView instances.

## Fix

`KeyChainRepositoryImpl` already persists the chosen key alias via `PrefsRepository.saveKeyAlias()`.
We expose a new `suspend fun loadAlias(): String?` on the `KeyChainRepository` interface that reads
this persisted value _without_ requiring `load()` to have been called first (i.e., it works after a
cold start or force-stop).

`TLSWebViewClient` gains a new `suspend fun preInitializeTLSClientAuthState()` that checks for a
stored alias and pre-sets `isTLSClientAuthNeeded = true` if one exists. This is called from
`ConnectionViewModel.init` _before_ `buildAuthUrl()`, so the flag is correct whether or not the
subsequent WebView connection is a resumed TLS session.

## Changes

| File | Change |
|------|--------|
| `common/.../data/keychain/KeyChainRepository.kt` | Add `suspend fun loadAlias(): String?` to interface |
| `common/.../data/keychain/KeyChainRepositoryImpl.kt` | Implement `loadAlias()` via `PrefsRepository` |
| `common/.../data/keychain/KeyStoreRepositoryImpl.kt` | Implement `loadAlias()` returning in-memory alias |
| `app/.../util/TLSWebViewClient.kt` | Add `suspend fun preInitializeTLSClientAuthState()` |
| `app/.../onboarding/connection/ConnectionViewModel.kt` | Call `preInitializeTLSClientAuthState()` in `init` |
| `app/.../onboarding/connection/ConnectionViewModelTest.kt` | Two new tests covering pre-init behaviour |

## Behaviour

- **Happy path (cold start, TLS resumption):** alias is stored → `isTLSClientAuthNeeded = true`
  before any WebView request → mTLS cert screen shown during Wear onboarding ✓
- **No cert configured:** `loadAlias()` returns `null` → flag stays `false` → behaviour unchanged ✓
- **False positive (alias stored but server no longer requires mTLS):** mTLS screen appears but
  the user can dismiss it without selecting a cert; the watch then connects without a cert, which
  succeeds against a non-requiring server ✓
- **`onReceivedClientCertRequest` still fires (fresh TLS handshake):** callback sets the flag to
  `true` as before; `preInitializeTLSClientAuthState()` is a no-op because the flag is already
  `true` ✓

## Testing

- Two new unit tests in `ConnectionViewModelTest` cover:
  - stored alias present → `isTLSClientAuthNeeded` pre-set to `true`
  - no stored alias → `isTLSClientAuthNeeded` remains `false`
- All existing tests continue to pass because `keyChainRepository` is `mockk(relaxed = true)` and
  `loadAlias()` returns `null` by default.


smhc:
## Known limitation — no host comparison

The current fix uses `keyChainRepository.getPrivateKey() != null` to detect whether the phone is connected to an mTLS-protected instance. This is an in-memory check and does not compare the Wear onboarding target host against the host for which the certificate was loaded.

**Potential false positive:** if a user has multiple HA servers configured where Server A requires mTLS and Server B does not, and the phone is currently connected to Server A (key loaded in memory) while setting up the watch for Server B, the mTLS certificate selection screen will appear unnecessarily during Wear onboarding.

In practice this is a very advanced scenario — most users have a single HA instance, and anyone running split mTLS/non-mTLS multi-server setups is technical enough to recognise the cert selection screen and dismiss it without selecting a certificate, allowing the watch to onboard against Server B without a cert as expected.

A more precise fix would store the host from `request.host` inside `onReceivedClientCertRequest` in the shared `KeyChainRepository` singleton and compare it against `rawUrl.host` in `preInitializeTLSClientAuthState`. Happy to add that if reviewers feel it is warranted.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -39,6 +39,36 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository)
...
+     * first (clearing in-memory state) no TLS session can be resumed either, so
+     * [onReceivedClientCertRequest] will fire naturally on the fresh handshake.
+     *
+     * Must be called **before** the WebView starts loading (i.e. before the URL is emitted).
+     * Idempotent: if the flag is already `true` (set by a real handshake) this is a no-op.
+     */
+    fun preInitializeTLSClientAuthState() {
+        if (!isTLSClientAuthNeeded) {
+            isTLSClientAuthNeeded = keyChainRepository.getPrivateKey() != null
+        }
```
TimoPtr: > A more precise fix would store the host from request.host inside onReceivedClientCertRequest in the shared KeyChainRepository singleton and compare it against rawUrl.host in preInitializeTLSClientAuthState. Happy to add that if reviewers feel it is warranted.

Indeed, if I'm not wrong it means that if I onboard a new server that is not using mTLS but one of my server has mTLS I'm going to be forced to pick a cert for the wear where I might have none. It shouldn't be an issue to give a "fake" cert but still a pretty bad experience. 

We could try to store the host information indeed.

smhc: I had assumed the cert picker in the UI was optional even if the cert was considered "mandatory". I must admit I didn't check or I was remembering an old implementation.

It might want it to be optional anyway, in the event the server specifies it is optional. Although I believe at the moment it only presents this screen if it's mandatory (and if you don't hit this bug).

We could alternatively set a flag stating that it is "probably/optionally required" and let the user decide whether to provide one or not.

> but one of my server has mTLS

I understand this will only be the case if the server you are currently logged into in the phone app requires mTLS. Note prior to this change it wouldn't work at all in this scenario if the server you are onboarding the watch to also requires mTLS (most common case). You get a "could not register watch" toast and no further info.

I can add the host matching - but am a bit concerned about matching domain suffixes, http vs https and so on.

TimoPtr: > I can add the host matching - but am a bit concerned about matching domain suffixes, http vs https and so on.

What is your concern?

smhc: I just meant matching myhainstance vs mhainstance.lan vs http://127.0.01 vs https://localhost etc, it's not a simple strcmp. I guess the certificate CNAME has to match anyway.

TimoPtr: We have some utils already for that in the repo like https://github.com/home-assistant/android/blob/9bc151c4007be826eea805f7faabade25652aa14/common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt#L107 or https://github.com/home-assistant/android/blob/9bc151c4007be826eea805f7faabade25652aa14/common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt#L186

smhc:
Was forced to raise a new PR due to git mistakes: https://github.com/home-assistant/android/pull/6672

PR #6672, opened 2026-04-03
fix: pre-initialize mTLS flag from cert SAN/CN to handle TLS session resumption in Wear OS onboarding

smhc:
## Problem

When setting up a Wear OS device through the Home Assistant phone app, the mTLS certificate
selection screen is silently skipped, causing the watch to register without a client certificate
and the server to respond with `HTTP 400`.

The root cause is **TLS session resumption**:

1. The main HA app WebView authenticates with the server — a TLS session ticket is created.
2. The user opens the Wear OS onboarding flow (either from the watch or via the phone app).
3. The onboarding WebView connects to the _same_ host; the OS TLS stack resumes the existing session (abbreviated handshake — no new `CertificateRequest` from the server).
4. `WebViewClient.onReceivedClientCertRequest()` is never invoked.
5. `TLSWebViewClient.isTLSClientAuthNeeded` remains `false`.
6. `requiredMTLS = false` is passed through `ConnectionNavigationEvent.Authenticated` → `OnboardingNavigation` skips `wearMTLSScreen`.
7. The watch is registered without a cert → server returns `400 Bad Request`.

## Fix

`TLSWebViewClient` gains `preInitializeTLSClientAuthState(targetHost: String)` that checks the in-memory certificate chain (if any) and tests whether it covers `targetHost` by matching against the certificate Subject Alternative Names (SANs):

- **dNSName** entries (type 2) with full wildcard support per RFC 2818 §3.1
- **iPAddress** entries (type 7) for IP-addressed instances
- **Common Name (CN)** fallback for legacy certificates that lack SANs

Matching against the certificate SANs — rather than mere private-key presence — eliminates the false positive identified in review: if a user has multiple servers where only Server A requires mTLS, the cert for Server A will **not** match Server B hostname, so the mTLS screen is suppressed correctly.

## Files changed

- `app/.../util/TLSWebViewClient.kt` — `preInitializeTLSClientAuthState()` + `certCoversHost()` + `hostMatchesSan()` helpers
- `app/.../onboarding/connection/ConnectionViewModel.kt` — Extract host from `rawUrl` and pass to `preInitializeTLSClientAuthState()`
- `app/.../onboarding/connection/ConnectionViewModelTest.kt` — 13 cert-host matching tests

## Review comments addressed

- certCoversHost is private (no @VisibleForTesting)
- Catches CertificateParsingException instead of generic Exception
- Catches IllegalArgumentException in ConnectionViewModel (cancellation-safe)
- Named SAN type constants instead of magic numbers (RFC 5280 §4.2.1.6)
- IP SAN handled as both ByteArray (BouncyCastle) and String (standard Java)
- Verbose comments in ConnectionViewModel trimmed
- Fork-specific CI workflow commits removed — zero CI changes
- Rebased onto latest home-assistant:main

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
@@ -164,6 +164,14 @@ internal class ConnectionViewModel @VisibleForTesting constructor(
 
     init {
         viewModelScope.launch {
+            // Pre-set the mTLS flag before emitting the auth URL to handle TLS session
+            // resumption. See preInitializeTLSClientAuthState for details.
+            try {
+                webViewClient.preInitializeTLSClientAuthState(rawUrl.toHttpUrl().host)
+            } catch (_: IllegalArgumentException) {
+                // Malformed URL: this is a best-effort pre-initialisation;
+                // buildAuthUrl below will surface the error to the user.
+            }
             buildAuthUrl(rawUrl)
         }
     }
--- a/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
@@ -17,6 +17,7 @@ import io.homeassistant.companion.android.frontend.error.FrontendConnectionError
 import io.homeassistant.companion.android.frontend.filechooser.FileChooserManager
 import io.homeassistant.companion.android.frontend.filechooser.FileChooserRequest
 import io.homeassistant.companion.android.onboarding.connection.navigation.ConnectionRoute
+import io.homeassistant.companion.android.util.CheckTLSClientAuthNeededUseCase
 import io.homeassistant.companion.android.util.HAWebChromeClient
 import io.homeassistant.companion.android.util.HAWebViewClient
 import io.homeassistant.companion.android.util.HAWebViewClientFactory
@@ -71,6 +72,7 @@ internal class ConnectionViewModel @VisibleForTesting constructor(
     webViewClientFactory: HAWebViewClientFactory,
     private val connectivityCheckRepository: ConnectivityCheckRepository,
     private val fileChooserManager: FileChooserManager,
+    private val checkTLSClientAuthNeededUseCase: CheckTLSClientAuthNeededUseCase,
 ) : ViewModel(),
     FrontendConnectionErrorStateProvider {
 
@@ -80,11 +82,13 @@ internal class ConnectionViewModel @VisibleForTesting constructor(
         webViewClientFactory: HAWebViewClientFactory,
         connectivityCheckRepository: ConnectivityCheckRepository,
         fileChooserManager: FileChooserManager,
+        checkTLSClientAuthNeededUseCase: CheckTLSClientAuthNeededUseCase,
     ) : this(
         savedStateHandle.toRoute<ConnectionRoute>().url,
         webViewClientFactory,
         connectivityCheckRepository,
         fileChooserManager,
+        checkTLSClientAuthNeededUseCase,
     )
 
     /**
@@ -243,7 +247,9 @@ internal class ConnectionViewModel @VisibleForTesting constructor(
                         ConnectionNavigationEvent.Authenticated(
                             url = effectiveUrl.value,
                             authCode = code,
-                            requiredMTLS = isTLSClientAuthNeeded,
+                            // The WebView may have resumed a TLS session without a
+                            // CertificateRequest, so also check the loaded certificate chain.
+                            requiredMTLS = isTLSClientAuthNeeded || checkTLSClientAuthNeededUseCase(rawHttpUrl?.host),
                         ),
                     )
                 }
--- a/app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
@@ -12,14 +12,22 @@ import android.webkit.WebViewClient
 import androidx.annotation.VisibleForTesting
 import io.homeassistant.companion.android.common.data.keychain.KeyChainRepository
 import java.lang.ref.WeakReference
+import java.net.InetAddress
+import java.net.UnknownHostException
 import java.security.PrivateKey
 import java.security.cert.CertificateException
+import java.security.cert.CertificateParsingException
 import java.security.cert.X509Certificate
+import javax.security.auth.x500.X500Principal
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import timber.log.Timber
+
+// SAN (Subject Alternative Name) type codes per RFC 5280 section 4.2.1.6
+private const val SAN_TYPE_DNS_NAME = 2
+private const val SAN_TYPE_IP_ADDRESS = 7
 
 /*
  * [TLSWebViewClient] is on the onboarding module for convenience, since we don't have yet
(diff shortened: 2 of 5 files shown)


smhc:
Note this change is AI generated and has not been manually tested. I worked around it via killing the main app, clearing cache/restart and invoking the intent directly via the watch to skip the initial HA TLS handshake.

There is an alternate solution of doing 'does current host match target host' string matching, however this doesn't handle wildcard certs where you may have 'host1.ha.com' vs 'host2.ha.com' and a cert of '*.ha.com'.

It's a pretty legitimate issue however and I wanted to raise it with suggested changes to better highlight the problem and the chances of a fix landing.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -39,6 +41,103 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository)
...
+    /**
+     * Returns `true` if [cert] is valid for [host].
+     *
+     * Checks Subject Alternative Names (SANs) first — both DNS names (with wildcard support)
+     * and IP addresses. Falls back to the Common Name (CN) in the Subject DN if no SANs are
+     * present, matching the behaviour of legacy TLS stacks.
+     */
+    @VisibleForTesting
+    internal fun certCoversHost(cert: X509Certificate, host: String): Boolean {
+        val sans: Collection<List<*>>? = try {
```
TimoPtr: Let's keep this private and test through the public API.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -39,6 +41,103 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository)
...
+            cert.subjectAlternativeNames
+        } catch (_: Exception) {
```
TimoPtr:
```suggestion
        } catch (_: CertificateParsingException) {
```

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
```diff
@@ -116,6 +116,19 @@ internal class ConnectionViewModel @VisibleForTesting constructor(
...
+            // Pre-set the mTLS flag before emitting the auth URL. If the phone is currently
+            // connected to an mTLS-protected instance whose certificate covers this host, the
+            // onboarding WebView will reuse the live TLS session (session resumption) and
+            // onReceivedClientCertRequest will never fire — pre-setting the flag ensures the
+            // Wear OS cert-selection screen is not silently skipped.
+            // Matching against the cert's SANs/CN rather than mere key presence avoids a false
+            // positive when the user has multiple servers where only one requires mTLS.
+            try {
+                webViewClient.preInitializeTLSClientAuthState(rawUrl.toHttpUrl().host)
+            } catch (_: Exception) {
```
TimoPtr: You need to rethrow the cancelation properly still since you catch Exception.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -39,6 +41,103 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository)
...
+        return if (!sans.isNullOrEmpty()) {
+            sans.any { san ->
+                val type = san[0] as? Int ?: return@any false
+                when (type) {
+                    2 -> { // dNSName — returned as String
+                        val value = san[1] as? String ?: return@any false
+                        hostMatchesSan(host, value)
+                    }
+                    7 -> { // iPAddress — returned as ByteArray per the Java X.509 API contract
+                        val ipBytes = san[1] as? ByteArray ?: return@any false
```
TimoPtr: Is there a way to not have this magic numbers? 2 an 7 or at least link it to the proper documentation.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -39,6 +41,103 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository)
...
+    internal fun certCoversHost(cert: X509Certificate, host: String): Boolean {
+        val sans: Collection<List<*>>? = try {
+            cert.subjectAlternativeNames
+        } catch (_: Exception) {
+            null
+        }
+
+        return if (!sans.isNullOrEmpty()) {
+            sans.any { san ->
+                val type = san[0] as? Int ?: return@any false
```
TimoPtr: You should assert the size of the `san` list.

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
```diff
@@ -116,6 +116,19 @@ internal class ConnectionViewModel @VisibleForTesting constructor(
 
     init {
         viewModelScope.launch {
+            // Pre-set the mTLS flag before emitting the auth URL. If the phone is currently
+            // connected to an mTLS-protected instance whose certificate covers this host, the
+            // onboarding WebView will reuse the live TLS session (session resumption) and
+            // onReceivedClientCertRequest will never fire — pre-setting the flag ensures the
+            // Wear OS cert-selection screen is not silently skipped.
+            // Matching against the cert's SANs/CN rather than mere key presence avoids a false
+            // positive when the user has multiple servers where only one requires mTLS.
```
TimoPtr: You should rewrite this and remove the implementation details of the function that you already document on the function itself.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -39,6 +41,103 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository)
...
+            // Fallback: extract CN from the Subject DN.
+            // getName(RFC2253) uses comma as AVA separator; commas inside values are escaped
+            // as \, which we don't need to handle because hostnames never contain commas.
+            val dn = cert.subjectX500Principal.getName(X500Principal.RFC2253)
+            val cn = dn.splitToSequence(",")
+                .map { it.trim() }
+                .firstOrNull { it.startsWith("CN=", ignoreCase = true) }
+                // Use substring-after-'=' so the extraction is case-insensitive (matches the
+                // startsWith check above) rather than a case-sensitive removePrefix("CN=").
+                ?.let { it.substring(it.indexOf('=') + 1).trim() }
```
TimoPtr: Is there a risk that dn is only `CN=` ?

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
```diff
@@ -417,4 +419,128 @@ class ConnectionViewModelTest {
...
+
+        val viewModel = ConnectionViewModel(
+            "https://192.168.1.100",
+            webViewClientFactory,
+            connectivityCheckRepository,
+        )
+        advanceUntilIdle()
+
+        assertTrue(viewModel.webViewClient.isTLSClientAuthNeeded)
+    }
```
TimoPtr: Missing not empty SAN but not matching and CN matching should return false.

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
```diff
@@ -417,4 +419,128 @@ class ConnectionViewModelTest {
...
+        val viewModel = ConnectionViewModel(
+            "https://ha.example.com",
+            webViewClientFactory,
+            connectivityCheckRepository,
+        )
+        advanceUntilIdle()
+
+        assertTrue(viewModel.webViewClient.isTLSClientAuthNeeded)
+    }
+
```
TimoPtr: Missing edge cases: *.example.com should not match example.com

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
TimoPtr: Are you sure it's always a ByteArray here and it never could be a String?

TimoPtr (changes requested):
Your comment about AI scares us a bit, we need to ensure this is tested properly by you. Even if it seems that the worse that can happen is not starting the mTLS flow it could become a blocker in some cases. I invite you to test it more.

I'm not an expert with certificat structure so I rely on you to test properly.

.github/workflows/onPush.yml
```diff
@@ -47,11 +47,11 @@ jobs:
           firebase-creds: ${{ secrets.FIREBASECREDS }}
           playstore-creds: ${{ secrets.PLAYSTORECREDS }}
 
-      - uses: ./.github/actions/create-release-notes
-        name: Create Release Notes
-        with:
-          tag-name: ${{ steps.rel_number.outputs.version }}
-          gh-token: ${{ secrets.GITHUB_TOKEN }}
+     # - uses: ./.github/actions/create-release-notes
```
TimoPtr: Please don't modify the CI within this PR 🙏🏻

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -40,6 +48,117 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository)
...
+     *
+     * If the app was force-stopped first (clearing in-memory state) no TLS session can be
+     * resumed either, so [onReceivedClientCertRequest] will fire naturally on the fresh handshake.
+     *
+     * Must be called **before** the WebView starts loading (i.e. before the URL is emitted).
+     * Idempotent: if the flag is already `true` (set by a real handshake) this is a no-op.
+     *
+     * @param targetHost the hostname of the server being connected to (e.g. "myha.example.com")
+     */
+    fun preInitializeTLSClientAuthState(targetHost: String) {
```
TimoPtr: I don't feel like it should be in the `WebViewClient` you are giving a targetHost and then do some checks on it to update an internal variable just to surface it in the ConnectionViewModel. I think we could make a small UseCase used in the connection viewModel that has the responsability to check the status and combine it with the boolean from within the client. So it's easier to read and each part has its own responsibility.

TimoPtr (changes requested):

PR #6680, opened 2026-04-07
Filter out messages not coming from the frontend external bus

TimoPtr:
## Summary

We only want to react on messages sent from the frontend. This PR only concern the `WebViewActivity` and not the `FrontendScreen` (this is going to be done in a second PR). We've introduced a new `externalAppV2` to be able to use the `WebViewFeature.WEB_MESSAGE_LISTENER` that allow us to properly check the origin of the message, if not available we simply fallback to the old JS injection.

I decided also to move the JS function that we inject into the externalBus, the reason behind this change is that in the frontend we can add this message in the codebase as well so that it never conflict with the one we are injecting like `handleBlob`.

The JS interface is now added/removed each time we load the URL instead of directly from the `onCreate` it allows us to properly check the server version.

This requires https://github.com/home-assistant/frontend/pull/51446 to be merged. If it not merged by Friday we will need to update the min version for the support of the V2.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -933,7 +933,7 @@ class WebViewActivity :
                 return@addWebMessageListener
             }
             if (!sourceOrigin.hasSameOrigin(loadedUrl)) {
-                Timber.w("Ignored message from unexpected origin: $sourceOrigin")
+                Timber.w("Ignored message from unexpected origin: ${sensitive(sourceOrigin.toString())}")
                 return@addWebMessageListener
             }
 
@@ -1672,10 +1672,10 @@ class WebViewActivity :
         )
         this.serverHandleInsets.value = serverHandleInsets
         lifecycleScope.launch {
-            // Register the native bridge depending on the server and webview capabilities
-            webViewAddJavascriptInterface()
-
             if (openInApp) {
+                // Register the native bridge depending on the server and webview capabilities
+                webViewAddJavascriptInterface()
+
                 runFragmentTransactionIfStateSafe {
                     // Remove any displayed fragments (e.g., BlockInsecureFragment, ConnectionSecurityLevelFragment)
                     supportFragmentManager.popBackStack(null, FragmentManager.POP_BACK_STACK_INCLUSIVE)
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -156,6 +156,7 @@ import io.homeassistant.companion.android.webview.insecure.BlockInsecureFragment
 import javax.inject.Inject
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
@@ -282,6 +283,7 @@ class WebViewActivity :
     private var isShowingError = false
     private var isRelaunching = false
     private var alertDialog: AlertDialog? = null
+    private var loadUrlJob: Job? = null
     private var isVideoFullScreen = false
     private var videoHeight = 0
     private var firstAuthTime: Long = 0
@@ -1674,37 +1676,38 @@ class WebViewActivity :
             } (keepHistory $keepHistory, openInApp $openInApp, serverHandleInsets $serverHandleInsets)",
         )
         this.serverHandleInsets.value = serverHandleInsets
-        lifecycleScope.launch {
-            if (openInApp) {
-                runFragmentTransactionIfStateSafe {
-                    // Remove any displayed fragments (e.g., BlockInsecureFragment, ConnectionSecurityLevelFragment)
-                    supportFragmentManager.popBackStack(null, FragmentManager.POP_BACK_STACK_INCLUSIVE)
-                }
-                supportFragmentManager.clearFragmentResultListener(BlockInsecureFragment.RESULT_KEY)
+        if (openInApp) {
+            runFragmentTransactionIfStateSafe {
+                // Remove any displayed fragments (e.g., BlockInsecureFragment, ConnectionSecurityLevelFragment)
+                supportFragmentManager.popBackStack(null, FragmentManager.POP_BACK_STACK_INCLUSIVE)
+            }
+            supportFragmentManager.clearFragmentResultListener(BlockInsecureFragment.RESULT_KEY)
 
-                val oldUrl = loadedUrl
-                // It means that if we loaded an URL with a path previously and we try to load the same URL without
-                // a path we don't do anything.
-                val shouldLoadUrl = !url.hasSameOrigin(oldUrl) || url.hasNonRootPath()
-                if (shouldLoadUrl) {
-                    clearHistory = !keepHistory
-                    loadedUrl = url
+            val oldUrl = loadedUrl
+            // It means that if we loaded an URL with a path previously and we try to load the same URL without
+            // a path we don't do anything.
+            val shouldLoadUrl = !url.hasSameOrigin(oldUrl) || url.hasNonRootPath()
+            if (shouldLoadUrl) {
+                clearHistory = !keepHistory
+                loadedUrl = url
 
+                loadUrlJob?.cancel()
+                loadUrlJob = lifecycleScope.launch {
                     // Register the native bridge depending on the server and webview capabilities
                     webViewAddJavascriptInterface()
 
                     webView.loadUrl(url.toString())
                     waitForConnection()
-                } else {
-                    Timber.d("Same base URL without meaningful path, skipping load")
                 }
             } else {
-                try {
-                    val browserIntent = Intent(Intent.ACTION_VIEW, url)
-                    startActivity(browserIntent)
-                } catch (e: Exception) {
-                    Timber.e(e, "Unable to view url")
-                }
+                Timber.d("Same base URL without meaningful path, skipping load")
+            }
+        } else {
+            try {
+                val browserIntent = Intent(Intent.ACTION_VIEW, url)
+                startActivity(browserIntent)
+            } catch (e: Exception) {
+                Timber.e(e, "Unable to view url")
             }
         }
     }
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -858,12 +858,13 @@ class WebViewActivity :
     /**
      * Registers the appropriate native bridge for the current server.
      *
-     * Queries [isServerSupportingExternalAppV2] to determine whether to use the `externalAppV2` bridge or the legacy
-     * `externalApp` bridge. V2 also requires [WebViewFeature.WEB_MESSAGE_LISTENER] support;
+     * Queries [isServerSupportingExternalAppV2] to determine whether to use the
+     * `externalAppV2` bridge or the legacy `externalApp` bridge.
+     * V2 also requires [WebViewFeature.WEB_MESSAGE_LISTENER] support;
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -841,196 +855,323 @@ class WebViewActivity :
...
+                fun revokeExternalAuth(callback: String) {
+                    handleRevokeExternalAuth(payload = callback.toJsonObjectOrNull())
+                }
+
+                @JavascriptInterface
+                fun externalBus(message: String) {
+                    Timber.d("External bus $message")
+                    webView.post {
+                        val json = message.toJsonObjectOrNull() ?: return@post
+                        handleExternalBusMessage(json)
```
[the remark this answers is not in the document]
TimoPtr: @jpelgrom It was already logging the payload before, lemme know what you think. It might be enough to display the ID and type.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -841,196 +855,323 @@ class WebViewActivity :
...
+     * an iframe or from an origin that doesn't match the currently loaded server URL.
+     */
+    @SuppressLint("RequiresFeature")
+    private fun registerExternalAppV2() {
+        WebViewCompat.removeWebMessageListener(webView, EXTERNAL_APP_V2_LISTENER)
+        WebViewCompat.addWebMessageListener(
+            webView,
+            EXTERNAL_APP_V2_LISTENER,
+            setOf("*"),
+        ) { _, message, sourceOrigin, isMainFrame, _ ->
```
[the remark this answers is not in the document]
TimoPtr: As per the documentation it is acceptable to use `*` if we properly filter which we are doing by checking the origin. This is made this way so that we don't have to build the list of acceptable URLs and we can log the ignore.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -841,196 +855,323 @@ class WebViewActivity :
...
-                                            hasNfc = hasNfc,
-                                            canCommissionMatter = canCommissionMatter,
-                                            canExportThread = canExportThread,
-                                            hasBarCodeScanner = hasBarCodeScanner,
-                                            appVersion = appVersionProvider(),
-                                        ),
-                                    )
+                // TODO This feature is deprecated and should be removed after 2022.6
+                // Set event listener for HA theme change
+                lifecycleScope.launch {
```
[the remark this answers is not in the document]
TimoPtr: Out of scope of this PR.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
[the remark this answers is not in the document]
jpelgrom: It might be nice to limit logging here to be safe, but I'd still prefer the full message in debug builds

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -841,196 +855,323 @@ class WebViewActivity :
...
-                        // will not trigger it's callback method :/
-                        lifecycleScope.launch(Dispatchers.Main) {
-                            getAndSetStatusBarNavigationBarColors()
-                        }
+    /**
+     * Registers the appropriate native bridge for the current server.
+     *
+     * Queries [isServerSupportingExternalAppV2] to determine whether to use the `externalAppV2` bridge or the legacy
+     * `externalApp` bridge. V2 also requires [WebViewFeature.WEB_MESSAGE_LISTENER] support;
+     * falls back to V1 if the feature is unavailable.
```
jpelgrom: Inconsistent max line length

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -841,196 +855,323 @@ class WebViewActivity :
...
+     *
+     * Both directions are safe: V1 uses [android.webkit.WebView.removeJavascriptInterface] and
```
jpelgrom: What do you mean with safe? Reading the code you probably mean something like it ensures it only registers one interface but that isn't clear from the KDoc here to me.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -841,196 +855,323 @@ class WebViewActivity :
...
-                                            id = json["id"],
-                                            hasNfc = hasNfc,
-                                            canCommissionMatter = canCommissionMatter,
-                                            canExportThread = canExportThread,
-                                            hasBarCodeScanner = hasBarCodeScanner,
-                                            appVersion = appVersionProvider(),
-                                        ),
-                                    )
+                // TODO This feature is deprecated and should be removed after 2022.6
+                // Set event listener for HA theme change
```
jpelgrom, line 1045: `settheme` events are still being fired so I'd argue this isn't deprecated and we can remove the comment, also to satisfy AI review. <https://github.com/home-assistant/frontend/blob/b73ae60cea8e61bb899aca20e0a74a011d65a31d/src/panels/profile/ha-pick-theme-row.ts#L231>

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1530,35 +1671,46 @@ class WebViewActivity :
...
+                }
             }
         }
     }
 
+    /**
+     * Whether the active server supports the `externalAppV2` bridge (>= 2026.4.2).
+     */
+    private suspend fun isServerSupportingExternalAppV2(): Boolean =
+        serverManager.getServer(presenter.getActiveServer())?.version?.isAtLeast(2026, 4, 2) == true
```
jpelgrom, line 1718: Comment to check shortly before merging if frontend version is correct

jpelgrom, line 1718: Would it make sense to default to V2 if we don't know the server version, to avoid downgrading to the old version accidentally?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -841,196 +855,323 @@ class WebViewActivity :
...
         }
     }
 
+    /**
+     * Returns a JS snippet that sends [jsonPayload] through the external bus.
+     *
+     * For V1: calls `window.externalApp.externalBus(...)` directly.
+     * For V2: posts a `{type:'externalBus', payload:...}` message via `window.externalAppV2`.
+     */
+    private fun externalBusCall(jsonPayload: String): String = """
```
jpelgrom: As the current usage appears to boil down to:

```kotlin
val callback = externalBusCal("{type:'x'}")
webView.evaluateJavascript("function() {$callback};")
```

why not return the outer `function() {` and `};`from this Kotlin function as well?

(slightly simplified)

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -841,196 +855,323 @@ class WebViewActivity :
...
+            val payload = json["payload"]?.jsonObjectOrNull()
+
+            when (type) {
+                "getExternalAuth" -> {
+                    handleGetExternalAuth(payload)
+                }
+
+                "revokeExternalAuth" -> {
+                    handleRevokeExternalAuth(payload)
+                }
```
jpelgrom: Didn't check ktlint but if you're only calling a function it doesn't need {}

```suggestion
                "getExternalAuth" -> handleGetExternalAuth(payload)
                "revokeExternalAuth" -> handleRevokeExternalAuth(payload)
```

jpelgrom (commented):
Nice improvement to use the web message listener. 

> This PR only concern the `WebViewActivity` and not the `FrontendScreen` (this is going to be done in a second PR). 

Do you intend to build on this and extract some code/constants from the activity, or duplicate code?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1530,35 +1671,46 @@ class WebViewActivity :
...
-                waitForConnection()
+        lifecycleScope.launch {
+            // Register the native bridge depending on the server and webview capabilities
+            webViewAddJavascriptInterface()
+
+            if (openInApp) {
+                runFragmentTransactionIfStateSafe {
+                    // Remove any displayed fragments (e.g., BlockInsecureFragment, ConnectionSecurityLevelFragment)
+                    supportFragmentManager.popBackStack(null, FragmentManager.POP_BACK_STACK_INCLUSIVE)
+                }
```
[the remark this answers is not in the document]
jpelgrom: @TimoPtr I think this makes sense, moving it just before `webView.loadUrl` is the only time there is actually a need to (re)register.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: Let's keep this out of this PR wdyt?

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: I prefer to default to v1 where we are sure it works on all version.

TimoPtr:
> Nice improvement to use the web message listener.
> 
> > This PR only concern the `WebViewActivity` and not the `FrontendScreen` (this is going to be done in a second PR).
> 
> Do you intend to build on this and extract some code/constants from the activity, or duplicate code?

I'm going to change things within the `WebViewActivity` to reuse some const, but the logic is a bit different in the `FrontendScreen`.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -841,196 +857,326 @@ class WebViewActivity :
...
+        }
+    }
+
+    /**
+     * Handles an external bus message received from the frontend via the native bridge.
+     */
+    private fun handleExternalBusMessage(json: JsonObject) {
+        val type = json.getStringOrNull("type")
+        val messageId = json.getIntOrNull("id")
+        Timber.d("External bus id=$messageId type=$type raw=${sensitive{ json.toString() }}")
```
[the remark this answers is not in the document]
TimoPtr: This is only relevant for debug, we prefer to have the long log here.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -2012,21 +2171,24 @@ class WebViewActivity :
...
                         const response = await fetch($safeUrl);
                         if (!response.ok) {
-                            console.error('Blob download failed: HTTP ' + response.status + ' for ' + ${sensitive(
-                safeUrl,
-            )});
+                            console.error('Blob download failed: HTTP ' + response.status + ' for ' + ${
+                sensitive(
+                    safeUrl,
+                )
+            });
```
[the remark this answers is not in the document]
TimoPtr: Indeed.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: Looks like the change made it into [frontend 20260325.7](https://github.com/home-assistant/frontend/releases/tag/20260325.7) which will be included in [core 2026.4.2](https://github.com/home-assistant/core/pull/167939) 👍

jpelgrom (approved):
Nice modernisation!

Had to wait for a bit to make sure 2026.4.2 shipped to avoid breaking in case of last minute changes, but it's available now.

PR #6681, opened 2026-04-07
Make sure to use repository English translation instead of Lokalise

TimoPtr:
## Summary

https://github.com/home-assistant/android/pull/1397 made sure that we were not overriding the English translation that we have in the repository even if Lokalise have a different version using `unzip -n` that skip file that already exists. Recently I've migrated to the Lokalise Github action #6659 and I didn't know this, because of that change we are currently creating release with wrong version name as pointed out by @jpelgrom in https://github.com/home-assistant/android/pull/6659#issuecomment-4196534670.

I reverted to the previous behavior by simply checking out the strings.xml back to revert any changes made by Lokalize while downloading the other translations.

## Checklist

## Any other notes

We might need to document this somewhere that English is not in Lokalize? Maybe it is already  and I didn't read it.

### DIFF
--- a/.github/actions/download-translations/action.yml
+++ b/.github/actions/download-translations/action.yml
@@ -10,6 +10,11 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Record commit before pull
+      id: before-pull
+      shell: bash
+      run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+
     - name: Pull from Lokalise
       uses: lokalise/lokalise-pull-action@a4968e06b7f44c32317b162d825ed17abb5772d0 # v5.0.1
       with:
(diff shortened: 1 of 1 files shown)


.github/actions/download-translations/action.yml
```diff
@@ -27,6 +27,14 @@ runs:
           replace_breaks: true
           export_empty_as: skip
 
+    # Source strings (English) are managed in git, not Lokalise.
+    # Restore the file in case Lokalise overwrote it.
+    - name: Revert source strings to git version
+      shell: bash
+      run: |
+        git checkout common/src/main/res/values/strings.xml
+        git status
```
jpelgrom, line 36: Not strictly necessary but useful for debug

jpelgrom (approved):
Thanks for the quick fix! I must admit I didn't expect Lokalise to overwrite the original so didn't think this could happen.

PR #6679, opened 2026-04-07
Replace "custom component" with "community integration"

NoRi2909:
## Summary

"Custom components" became "custom integrations" and are now "community integrations" – see https://github.com/home-assistant/frontend/pull/51368

This PR updates two occurrences in the user-facing strings of the Companion app for consistency.

## Checklist

- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [ ] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

PR #6707, opened 2026-04-14
Revert "Replace "custom component" with "community integration""

thecode:
## Summary

At https://github.com/home-assistant/frontend/pull/51440 it was decided to restore the existing term custom components.
Conversation on this: https://discord.com/channels/330944238910963714/1490648025415159910

Revert "Replace "custom component" with "community integration""
Reverts https://github.com/home-assistant/android/pull/6679

## Checklist

### DIFF
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -566,7 +566,7 @@
     <string name="search_refine_for_more">Refine your search to view more…</string>
     <string name="search_results">Search results</string>
     <string name="search_sensors">Search sensors</string>
-    <string name="security_vulnerably_message">Your version of Home Assistant does not protect for recently discovered security vulnerabilities in community integrations. Please view the details via the bulletin and update as soon as possible.</string>
+    <string name="security_vulnerably_message">Your version of Home Assistant does not protect for recently discovered security vulnerabilities in custom components. Please view the details via the bulletin and update as soon as possible.</string>
     <string name="security_vulnerably_title">Security vulnerability warning</string>
     <string name="security_vulnerably_understand">I understand</string>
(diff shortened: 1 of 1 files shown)


TimoPtr:
Shouldn't we align with the frontend and use `integration` instead of `component`?

thecode:
> Shouldn't we align with the frontend and use `integration` instead of `component`?

This is a revert PR, I didn't change anything just reverted the original PR, If you follow up the discord discussion, I don't think there is a clear consensus about the final naming. I think it would be better to just revert everything and let the discussion continue as the final result might be different.

TimoPtr:
> > Shouldn't we align with the frontend and use `integration` instead of `component`?
> 
> This is a revert PR, I didn't change anything just reverted the original PR, If you follow up the discord discussion, I don't think there is a clear consensus about the final naming. I think it would be better to just revert everything and let the discussion continue as the final result might be different.

The issue we have right now, is that the translation started to be done with the new translation, if we change it again and change it again once we reach agreements it's going to be a lot of work waste IMO.

@jpelgrom any takes on this?

jpelgrom:
> The issue we have right now, is that the translation started to be done with the new translation, if we change it again and change it again once we reach agreements it's going to be a lot of work waste IMO.
> 
> @jpelgrom any takes on this?

I think we should revert and accept this PR.

While I agree that we should avoid too many changes quickly, reading the linked Discord discussion it also sounds like agreement on a 'new' name won't come soon. Reverting now means the change was only live for one week (not all languages might have translated yet) and we get the _custom_ terminology back before the release we want to do later this week.

TimoPtr (approved):

PR #6708, opened 2026-04-14
Add frontend download capabilities to FrontendScreen and rework permissions

TimoPtr:
## Summary

Introduce `FrontendDownloadManager` to handle file downloads from the WebView, supporting blob URLs, HTTP/HTTPS URLs (via the system `DownloadManager`), and data URIs. Rework the JS bridge, permission manager, and message handler to support the new download flow, and make `DataUriDownloadManager` injectable. Rename `FrontendNavigationEvent` to `FrontendEvent` to be more generic than navigating to handle one time thing happening.

The rework of the whole permission system allow for a better abstraction from the VM point of view and make a generic handling in the screen. It means that it is going to require less work in future PR to request a permission mostly focused only within the PermissionManager and the code that actually need the permission instead of the screen.

## Checklist

- [x] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

## Any other notes

I started working on this quite some time ago and I had to go over multiple iterations before being able to open the PR. It means that maybe there are some leftover from previous iterations.

My plan later is to properly draw in the documentation how things are now communicating and explaining the goal of each stream, like events VS state.

PR #6713, opened 2026-04-15
Introduce WebViewAction and enforce EvaluateScriptUsage

TimoPtr:
## Summary

This PR is based on #6640 and replace the actions we can do on the WebView like evaluate script or others into a action flow using `WebViewAction`. The `WebViewAction` is a sealed class that contains actions it is tight to the Webkit framework by exposing a `run(webView)` method that we don't need to add the implementation of the action within a `when` on the screen. I've also introduce a `result` field used to notified when the action is done, it is important for some actions we run like getting the result of a script evaluation, or knowing when the clear action is done.

To raise awareness that we should not privileged injecting JS script I've introduce the opt in annotation `EvaluateScriptUsage`, that should be easy to spot during review. This should address part of https://github.com/home-assistant/android/pull/6708#discussion_r3081617806.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
@@ -13,8 +13,6 @@ import io.homeassistant.companion.android.frontend.download.DownloadResult
 import io.homeassistant.companion.android.frontend.download.FrontendDownloadManager
 import io.homeassistant.companion.android.frontend.error.FrontendConnectionError
 import io.homeassistant.companion.android.frontend.error.FrontendConnectionErrorStateProvider
-import io.homeassistant.companion.android.frontend.externalbus.WebViewScript
-import io.homeassistant.companion.android.frontend.externalbus.incoming.HapticType
 import io.homeassistant.companion.android.frontend.handler.FrontendBusObserver
 import io.homeassistant.companion.android.frontend.handler.FrontendHandlerEvent
 import io.homeassistant.companion.android.frontend.js.BridgeState
@@ -44,6 +42,7 @@ import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.merge
 import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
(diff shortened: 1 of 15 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -144,8 +143,8 @@ internal class FrontendViewModel @VisibleForTesting constructor(
     private val _events = MutableSharedFlow<FrontendEvent>(extraBufferCapacity = 1)
     val events: SharedFlow<FrontendEvent> = _events.asSharedFlow()
 
-    private val _hapticEvents = MutableSharedFlow<HapticType>(extraBufferCapacity = 16)
-    val hapticEvents: SharedFlow<HapticType> = _hapticEvents.asSharedFlow()
+    private val _webViewActions = MutableSharedFlow<WebViewAction>(extraBufferCapacity = 1)
```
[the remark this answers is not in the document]
TimoPtr: We don't use tryEmit but emit, so it's fine emit is going to be suspend until the flow is consumed.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+}
+
+/**
+ * Gates direct JavaScript evaluation in the WebView behind an explicit opt-in.
+ */
+@RequiresOptIn(
+    message =
+    "Evaluating raw JavaScript tightly couples the app to frontend internals and is fragile across frontend changes. " +
```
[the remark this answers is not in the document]
TimoPtr: Let's keep ktlint for now

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
```diff
@@ -0,0 +1,103 @@
...
+        "Prefer collaborating with the frontend team to add a dedicated externalBus message. " +
+        "Only opt in as a last resort, and document on the opt-in site why the externalBus is not a viable option so reviewers can challenge the usage.",
```
jpelgrom, line 97: I'd argue this is also so it can be challenged in the future; using the external bus may not be viable now for [reason] but could become viable later.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
```diff
@@ -0,0 +1,103 @@
...
+            webView.evaluateJavascript(script) { scriptResult ->
+                result.complete(scriptResult)
+            }
+        }
+    }
+}
+
+/**
+ * Gates direct JavaScript evaluation in the WebView behind an explicit opt-in.
+ */
```
jpelgrom:
```suggestion
/** Gates direct JavaScript evaluation in the WebView behind an explicit opt-in. */
```

jpelgrom (commented):
I like the cleanup here, much nicer once you start extending the WebView in Compose with more features

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
TimoPtr: Indeed, I already added explanation saying that it could be replaced with an external bus events. I hope that from now on we manage to not add any comments like that and use the bus whenever we can.

jpelgrom (approved):
Thanks for coming up with a nice solution for how to handle Javascript scripting.

No more comments, was previously discussed so not a lot surprises.

PR #6714, opened 2026-04-15
Add support for Gesture in FrontendScreen

TimoPtr:
## Summary

This PR is build on top of #6713 and adds support for all the gesture actions available to the `FrontendScreen`. I decided to re-use the `ServerChooserFragment` since it is out of scope of the refactor of the `WebViewActivity` IMO, but later this could benefits a migration to a proper `Composable` that we can re-use within a screen instead of a `Fragment`.

## Checklist

## Any other notes

Once #6713 merged we need to rebase this PR.

### DIFF
--- a/.idea/runConfigurations/app.xml
+++ b/.idea/runConfigurations/app.xml
@@ -8,7 +8,7 @@
     <option name="ARTIFACT_NAME" value="" />
     <option name="PM_INSTALL_OPTIONS" value="" />
     <option name="ALL_USERS" value="false" />
-    <option name="ALWAYS_INSTALL_WITH_PM" value="false" />
+    <option name="ALWAYS_INSTALL_WITH_PM" value="true" />
     <option name="ALLOW_ASSUME_VERIFIED" value="false" />
     <option name="CLEAR_APP_STORAGE" value="false" />
     <option name="DYNAMIC_FEATURES_DISABLED_LIST" value="" />
@@ -20,8 +20,6 @@
     <option name="CLEAR_LOGCAT" value="false" />
     <option name="SHOW_LOGCAT_AUTOMATICALLY" value="false" />
     <option name="TARGET_SELECTION_MODE" value="DEVICE_AND_SNAPSHOT_COMBO_BOX" />
-    <option name="SELECTED_CLOUD_MATRIX_CONFIGURATION_ID" value="-1" />
-    <option name="SELECTED_CLOUD_MATRIX_PROJECT_ID" value="" />
     <option name="DEBUGGER_TYPE" value="Auto" />
     <Auto>
       <option name="USE_JAVA_AWARE_DEBUGGER" value="false" />
--- a/.idea/runConfigurations/app.xml
+++ b/.idea/runConfigurations/app.xml
@@ -8,7 +8,7 @@
     <option name="ARTIFACT_NAME" value="" />
     <option name="PM_INSTALL_OPTIONS" value="" />
     <option name="ALL_USERS" value="false" />
-    <option name="ALWAYS_INSTALL_WITH_PM" value="true" />
+    <option name="ALWAYS_INSTALL_WITH_PM" value="false" />
     <option name="ALLOW_ASSUME_VERIFIED" value="false" />
     <option name="CLEAR_APP_STORAGE" value="false" />
     <option name="DYNAMIC_FEATURES_DISABLED_LIST" value="" />
@@ -20,6 +20,8 @@
     <option name="CLEAR_LOGCAT" value="false" />
     <option name="SHOW_LOGCAT_AUTOMATICALLY" value="false" />
     <option name="TARGET_SELECTION_MODE" value="DEVICE_AND_SNAPSHOT_COMBO_BOX" />
+    <option name="SELECTED_CLOUD_MATRIX_CONFIGURATION_ID" value="-1" />
+    <option name="SELECTED_CLOUD_MATRIX_PROJECT_ID" value="" />
     <option name="DEBUGGER_TYPE" value="Auto" />
     <Auto>
       <option name="USE_JAVA_AWARE_DEBUGGER" value="false" />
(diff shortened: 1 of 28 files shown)


.idea/runConfigurations/app.xml
jpelgrom, line 1: This file is unrelated

TimoPtr: Indeed pushed it by mistake. FYI if you ever use the Android 17 beta this is mandatory otherwise the app is not reinstall on the device...

jpelgrom:
This still needs a rebase/merge conflict resolution

.idea/runConfigurations/app.xml
jpelgrom: Not running the beta yet... I used to wait for Google I/O but I guess with the changed release timeline I should jump in sooner. 

If we need it for Android 17 you can submit the change separately, right?

TimoPtr: Yes indeed, but I hope that for now it's just a bug on Google side so we can wait.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/gesture/FrontendGestureHandler.kt
```diff
@@ -0,0 +1,175 @@
...
+        if (servers.size < 2) return GestureResult.Ignored
+        val currentIndex = servers.indexOfFirst { it.id == currentServerId }
+        if (currentIndex == -1) return GestureResult.Ignored
+        val nextIndex = (currentIndex + offset).mod(servers.size)
+        return GestureResult.SwitchServer(servers[nextIndex].id)
+    }
+
+    private suspend fun navigateToDashboard(serverId: Int): GestureResult {
+        val version = serverManager.getServer(serverId)?.version
+        if (version?.isAtLeast(2025, 6) != true) {
```
jpelgrom: Comparing this to the current setup, is it intentional that you've pulled compatibility information out of the message/action? I'm OK with it but liked the idea of having an `isAvailable` function, also in case we add more in the future which requires specific versions.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/gesture/FrontendGestureHandler.kt
```diff
@@ -0,0 +1,175 @@
...
+                ctrlKey: $ctrlKey,
+                bubbles: true,
+                cancelable: true
+            });
+            document.dispatchEvent(event);
+        """.trimIndent()
+        externalBusRepository.evaluateScript(script)
+    }
+
+    private suspend fun showServerSwitcherOrIgnore(): GestureResult = if (serverManager.servers().size < 2) {
```
jpelgrom: The server switcher was always shown, even with only 1 server

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -144,8 +149,8 @@ internal class FrontendViewModel @VisibleForTesting constructor(
     private val _events = MutableSharedFlow<FrontendEvent>(extraBufferCapacity = 1)
     val events: SharedFlow<FrontendEvent> = _events.asSharedFlow()
 
-    private val _hapticEvents = MutableSharedFlow<HapticType>(extraBufferCapacity = 16)
-    val hapticEvents: SharedFlow<HapticType> = _hapticEvents.asSharedFlow()
+    private val _webViewActions = MutableSharedFlow<WebViewAction>(extraBufferCapacity = 1)
+    val webViewActions: Flow<WebViewAction> = merge(_webViewActions, frontendBusObserver.webViewActions())
```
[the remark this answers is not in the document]
TimoPtr: Actually I'm not sure if in the first place we shouldn't use tryEmit at all and only use emit. Otherwise we might loose an event. The counterpart is that the buffer is only 1 so if we are blocking the collection we are blocking the emission and emit is going to suspend. 

@jpelgrom I'm going to change all the `tryEmit` here.

jpelgrom (commented):
Only minor comments. I forgot to comment this when I did the full review a few days ago, but functionality itself looks good!

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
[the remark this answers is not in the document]
jpelgrom: `tryEmit` has always been the shortcut to add something without doing coroutines :). With more code running in coroutines using  `emit` makes sense 👍 

I'm not sure suspending is the desired behavior, even if unlikely, why not increase the buffer as suggested?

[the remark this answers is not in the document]
TimoPtr: I currently have no idea if the buffer is going to be an issue or not and using a random number is not something that I like in this kind of place. For now I'm going to try to keep it to 1. IMO everything we do when using coroutine should expect to be suspended and should block anything. Let see how it goes.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/gesture/FrontendGestureHandler.kt
TimoPtr: No it's was not intentional let's keep it in the message.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/gesture/FrontendGestureHandler.kt
TimoPtr: Does it make sense?

app/src/main/kotlin/io/homeassistant/companion/android/frontend/gesture/FrontendGestureHandler.kt
```diff
@@ -0,0 +1,175 @@
...
+                keyCode: $keyCode,
+                which: $keyCode,
+                ctrlKey: $ctrlKey,
+                bubbles: true,
+                cancelable: true
+            });
+            document.dispatchEvent(event);
+        """.trimIndent()
+        externalBusRepository.evaluateScript(script)
+    }
```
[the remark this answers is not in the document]
TimoPtr: I'm fine with what we have for now.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/gesture/FrontendGestureHandler.kt
jpelgrom: I think having a gesture which doesn't do what is says is potentially confusing (especially as it is a default gesture!). Showing it won't break anything.

jpelgrom (approved):

PR #6719, opened 2026-04-16
Add support for NFC write in FrontendScreen

TimoPtr:
## Summary

Add support for NFC write action on the FrontendScreen.

## Checklist

## Any other notes

I'm not a super fan of adding more and more methods to the view model, we might need to think about how we can limit that a first "easy" thing might be using delegation, but let see how it goes.


app/src/main/kotlin/io/homeassistant/companion/android/frontend/navigation/FrontendEvent.kt
```diff
@@ -33,4 +33,15 @@ sealed interface FrontendEvent {
...
+    /**
+     * Launch the NFC tag-write flow.
+     *
+     * The host is responsible for launching the corresponding activity contract and forwarding the
+     * result back to the ViewModel via [FrontendViewModel.onNfcWriteCompleted].
+     *
+     * @param messageId Correlation id from the originating `tag/write` request.
+     * @param tagId Optional pre-filled tag identifier.
+     */
+    data class LaunchNfcWrite(val messageId: Int, val tagId: String?) : FrontendEvent
```
jpelgrom: For consistency with other events: why not navigate to / open NFC write?

jpelgrom:
> we might need to think about how we can limit that a first "easy" thing might be using delegation, but let see how it goes.

Did you already have something specific in mind for delegation/do you have an example?

TimoPtr:
> > we might need to think about how we can limit that a first "easy" thing might be using delegation, but let see how it goes.
> 
> Did you already have something specific in mind for delegation/do you have an example?

Not yet TBH, I need to have another example to brainstorm about what I could do.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/navigation/FrontendEvent.kt
TimoPtr: Addressed

jpelgrom:
OK. I also don't have another good approach at this point so let's continue with this for now.

jpelgrom (approved):

PR #6720, opened 2026-04-16
Add start workers in LaunchActivity

TimoPtr:
## Summary

Replicate the start workers from the `WebViewActivity` within the `LaunchActivity`. It does replicate exactly the existing behavior, it is not gated behind `WipFeature` because it just triggers a second update of the sensor which is acceptable.

## Checklist


TimoPtr:
@jpelgrom I'm struggling here to find a nice way to work with this and I need more context.
- I don't want to use any `Context` in the ViewModel
- ViewModel should stay unaware of the onResume/onStart IMO if possible

From my point of view it is also not the responsibility of the Screen to start the workers.

I can modify the start function to not take any `Context` by injecting the `WorkerManager` in the ViewModel, but I wonder if the logic of starting in `onResume` is mandatory or if it can only be made one time when the VM is created?

I'm curious about your opinion here.

jpelgrom:
> @jpelgrom I'm struggling here to find a nice way to work with this and I need more context.
> 
>     * I don't want to use any `Context` in the ViewModel
> 
>     * ViewModel should stay unaware of the onResume/onStart IMO if possible
> 
> 
> From my point of view it is also not the responsibility of the Screen to start the workers.
> 
> I can modify the start function to not take any `Context` by injecting the `WorkerManager` in the ViewModel, but I wonder if the logic of starting in `onResume` is mandatory or if it can only be made one time when the VM is created?
> 
> I'm curious about your opinion here.

I think the goal the current behavior is trying to achieve is simply more updates for sensors / a way to kickstart any workers that the system may have suspended, as it is unlikely they will be delayed when the app is actively being used. That means tying it to the entrypoint for users -> WebViewActivity. (@dshokouhi might know more history here.)

Following that logic, it isn't really about the specific screen but rather the entire app being opened or hidden (lifecycle).

I kind of understand what you're trying to do with the ViewModel taking no context, but could you spell it out? There are still tons of Android-specific functions that exist and need context so I'm not sure it can be completely avoided forever, and to do everything in Compose where you have a context feels like misuse.

ViewModel doesn't need to know about onResume/onStart but we could want functions called on those lifecycle events. If multiple things need to happen on onStart I'm not opposed to simply creating a function named onStart in the ViewModel to make it easier to manage 🤷.

dshokouhi:
Yes we do have an app importance sensor that needs to update if the app is in foreground or what not. That's why the on pause and on resume calls are there. It should be on the settings activity as well iirc

TimoPtr:
@jpelgrom we could potentially move the logic within the ViewModel but I'm not sure if it relevant here, it would force us to pass the context or change how we start the workers.

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchActivity.kt
```diff
@@ -94,6 +107,19 @@ class LaunchActivity : AppCompatActivity() {
         },
     )
 
+    override fun onResume() {
```
TimoPtr: What I don't like with this approach is that we don't have any tests.

jpelgrom: You want to test that onResume is called? Or that the worker starts?

TimoPtr: I want to test that everything is properly started in onResume.

jpelgrom:
Moving it into LaunchActivity means that, at least as long as the WebViewActivity is used, these two updates + the update in WebViewActivity launch trigger very shortly after another because you transition in and out of LaunchActivity almost immediately. Maybe put these behind the feature flag as well?

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchActivity.kt
```diff
@@ -94,6 +107,19 @@ class LaunchActivity : AppCompatActivity() {
...
+        lifecycleScope.launch {
+            WebsocketManager.start(this@LaunchActivity)
+        }
+    }
+
+    override fun onPause() {
+        super.onPause()
+        if (!isFinishing) SensorReceiver.updateAllSensors(this)
+    }
+
```
jpelgrom: Can you move these below onCreate? Maybe I'm weird but I like lifecycle methods in the file to roughly follow the actual lifecycle order, and we do so relatively consistently elsewhere.

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchActivity.kt
TimoPtr: I've made a unit test on the activity with Robolectric I want to use it to make sure we don't mess up by removing the start by mistakes.

app/src/test/kotlin/io/homeassistant/companion/android/launch/LaunchActivityTest.kt
```diff
@@ -0,0 +1,86 @@
...
+        ActivityScenario.launch(LaunchActivity::class.java).use {
+            verify { SensorWorker.start(any()) }
+            coVerify { WebsocketManager.start(any()) }
+        }
+    }
+
+    @Test
+    fun `Given activity pauses without finishing then all sensors are updated`() {
+        ActivityScenario.launch(LaunchActivity::class.java).use { scenario ->
+            scenario.moveToState(Lifecycle.State.STARTED) // triggers onPause
```
jpelgrom, line 82: This makes no sense to me, and I wouldn't be surprised if it turns out to be flaky. The [documentation](https://developer.android.com/reference/androidx/lifecycle/Lifecycle.State#STARTED) mentions (including emphasis!):

> Started state for a LifecycleOwner. For an android.app.Activity, this state is reached in two cases:
> - after android.app.Activity.onStart call;
> - **right before** android.app.Activity.onPause call.

TimoPtr: It's for the test library 
image

for me when calling moveToState with STARTED it does invoke `pauseActivity`
image

I didn't see any flakiness on this test, I did run it multiple times without any issue.

jpelgrom: Love the Google inconsistency... /s

In that case OK, no objection and it seems like the right way to test this.

jpelgrom (approved):

PR #6725, opened 2026-04-18
feat(webview): support external authentication providers in front of Home Assistant

### LINKED ISSUE #2650: Support Cloudflared Zero Trust protected instance from App

### LINKED ISSUE #3205: External authorization process loops (CloudFlare ZeroTrust)

My home assistant is protected by CloudFlare ZeroTrust solution. CF requires authentication before it exposes HA API. Simply when we try to access https://myHAlocation.domain.com, CF first displays own authentication page and then redirects to HA. 
The funny thing is that initial authentication in HA companion android app works perfectly. When url is provided, it displays CF page and then standard HA logon. However problem occurs when CF token expires after a month. Then CF wants to again reauthorize user and this process cannot be accomplished in HA companion app. From my observation it looks like that:
1. HA app tries to get data from HA
2. instead HA response CF authorization page is served
3. HA companion app opens >!external!< web browser and authorization process is continued in that external browser
4. Even if correct authorization data is provided in pages displayed in exterl browser page, authorization process never returns back to   HA companion
5. Restart of HA companion gets back to point 1 and we have authorization loop 

I just need whole process to be performed in HA app instead external browser. I have analysed this issue in the code and found that in class:
io.homeassistant.companion.android.webview.WebViewActivity
TLSWebViewClient as webViewClient. In initialization there are several methods overridden and one of them is "shouldOverrideUrlLoading". In code of this method I found following condition:
`                            } else if (!webView.url.toString().contains(it.toString())) {
                                Log.d(TAG, "Launching browser")
                                val browserIntent = Intent(Intent.ACTION_VIEW, it)
                                startActivity(browserIntent)
                                return true;
                            }`
which is root cause of the problem. When I replaced it by simple "return false", authentication process can be completed in HA app.

I really tried to find workarround which does not require code changes but i havent found any.

**Additional context**

### LINKED ISSUE #6425: Cloudflare Tunnel authentication is not working anymore on android with companion app.

**Android version(s):** 16

**Device model(s):** Galaxy S25+

**Home Assistant version:** 2026.2

**Last working Home Assistant release (if known):** 2026.1

Since the February 2026 update (v2026.2.x), I am experiencing a persistent login failure when accessing Home Assistant remotely via a Cloudflare Tunnel (running on a standalone Debian LXC). While access works perfectly via Chrome (Incognito mode), the Android Companion App is trapped in a loop of "Invalid Authentication" or "Invalid Login Session."

Steps to Reproduce:

Setup Home Assistant behind a Cloudflare Tunnel.

Access via a mobile provider that utilizes IPv6 - i cant choose going through IPv4.

Configure trusted_proxies in configuration.yaml to include the tunnel IP and the expected IPv6/Pseudo-IPv4 ranges.

Attempt login via the Android Companion App.

Expected Behavior:
The app should recognize the trusted proxy and authenticate successfully using the X-Forwarded-For headers provided by the Cloudflare tunnel.

Actual Behavior:
The app fails to authenticate. Home Assistant logs show "Login attempt or request with invalid authentication" from the public IPv6 address or Cloudflare's Pseudo-IPv4 (250.x.x.x), even when those IPs/ranges are explicitly listed in trusted_proxies.

Additional Notes:

Browser access alone works consistently, suggesting the network/tunnel configuration is correct and headers are being passed.

The issue persists after clearing App Cache, App Data, and Android System WebView cache.

The logs continue to report the source IP as the external IPv6 address, suggesting HA is failing to validate the request against the trusted_proxies list specifically for the mobile app's User-Agent or session handling.

Error message whatever i did is that the authentication failed.
Thanks for looking into this.

KingHavok:
## Summary

Adds first-class support in the companion app's `WebView` for standards-based external authentication providers sitting in front of a Home Assistant instance — Cloudflare Access, Authelia, Authentik, Vouch Proxy, oauth2-proxy, and any other OIDC / OAuth 2.0 reverse proxy.

This addresses a long-standing, frequently-reported issue ([#2650](https://github.com/home-assistant/android/issues/2650), [#3205](https://github.com/home-assistant/android/issues/3205), [#6425](https://github.com/home-assistant/android/issues/6425)) where an authentication proxy's login flow ejects the user to the system browser mid-handshake and never returns control to the app, making both initial onboarding and token-expiry re-authentication unusable.

### The problem

Placing an OIDC / OAuth 2.0 authentication proxy in front of a self-hosted service is a widely-recommended security pattern in the self-hosting community. It adds a layer of defence independent of the protected application, enabling MFA (WebAuthn, TOTP, hardware keys, email OTP) and centralised session management even where the protected application cannot provide those itself.

The flow is standards-based:

1. Client requests the protected resource.
2. Proxy intercepts and issues an HTTP 302 to an identity endpoint on a different domain.
3. User authenticates with the identity provider.
4. Provider redirects back to the protected resource, with a session cookie scoped to the proxy's domain.
5. Subsequent requests carry the cookie and the proxy passes them through.

Browsers handle this transparently. The companion app currently does not: when the `WebView` is redirected off the HA origin mid-auth, `shouldOverrideUrlLoading` routes the navigation to the system browser via `Intent.ACTION_VIEW`. The user completes auth in Chrome, but the resulting session cookie is scoped to the Chrome cookie jar and is never available to the app's `WebView`. The flow becomes unrecoverable.

This fails in two real-world scenarios:

- **Initial onboarding**: the proxy's first redirect leaves the WebView, and the round-trip back to HA never happens.
- **Token expiry (most common)**: auth-proxy session tokens have a finite lifetime (for example, Cloudflare Access defaults to 24 hours). When the app tries to re-establish its session, the proxy's re-auth redirect hits `shouldOverrideUrlLoading` and kicks the user out, with no recovery path.

### Prior proposals and why this one is different

Three approaches surface in the linked issues and community threads. Each was considered and is not sufficient as a general fix:

1. **Blanket `return false` in `shouldOverrideUrlLoading` ([#3205](https://github.com/home-assistant/android/issues/3205))** — makes the basic Cloudflare flow work, but silently breaks the existing external-link behaviour for the HA frontend: any tap on a documentation, integration, or third-party link inside HA would now open inside the app's `WebView` instead of the system browser, removing the user from their normal browser context (password manager, extensions, bookmarks). This PR preserves existing external-link behaviour exactly.
2. **Custom request headers / service tokens ([#2650](https://github.com/home-assistant/android/issues/2650))** — Cloudflare-specific, long-lived shared secrets, and bypasses MFA entirely, which is the opposite of what users are trying to achieve when they put an auth proxy in front of HA. This may be a worthwhile *additional* feature for non-interactive scenarios (webhooks, background tasks), but it is not a substitute for correct interactive-auth support and would be a separate PR.
3. **mTLS client certificates on a second subdomain** — cleverly routes around the missing feature but is not generalisable: requires two subdomains, a non-trivial certificate pipeline, is not available on all Cloudflare plan tiers, and does not apply to Authelia / Authentik / oauth2-proxy users at all.

### How this PR works

The key insight is that a `WebView` navigating off the HA origin can only be doing two things:

- **Following an authentication handshake** that will eventually return to HA — keep it in the `WebView`.
- **Following a user-initiated link tap** from the HA frontend to an unrelated external resource — hand to the system browser.

Distinguishing those two cases correctly is the whole of the fix. Two independent signals are used:

1. **`WebResourceRequest.isRedirect`** (API 24+) — the navigation is a server-issued HTTP redirect on the main frame, not a user gesture. This cleanly covers the initial proxy challenge (HA → proxy) and the proxy's return redirect (proxy → HA).
2. **WebView current URL is off the anchored HA origin** — the user is already inside an auth flow on a proxy page, and any further main-frame http(s) navigation (including user taps) is part of completing that flow.

The second signal is necessary because many auth-proxy login pages require a user *click* to proceed (for example choosing an identity provider from Cloudflare Access's selection page, or tapping a WebAuthn prompt on Authelia). Without it, the first correct click after reaching the proxy would re-eject the user to the browser. Adding this signal — gated on the WebView already being off the HA origin — keeps the guard intact when the WebView is back on HA.

Whenever the WebView lands back on the HA origin, the original "user-tapped external link → system browser" rule resumes unchanged. External-link handoff from the HA frontend behaves identically to before this PR.

### Implementation

- `TLSWebViewClient.shouldOverrideUrlLoading(WebView, WebResourceRequest)` is now overridden. It either short-circuits main-frame auth-provider navigation (returning `false` so the WebView follows it), or falls through to the existing deprecated `shouldOverrideUrlLoading(WebView, String)` override so that `APP_PREFIX` / `INTENT_PREFIX` / system-browser handoff logic is entirely unchanged for everything else.
- The decision lives in a single pure helper `isAuthProviderNavigation`, unit-testable without Robolectric.
- `TLSWebViewClient.serverHost: String?` is the anchor. It is set by the two call sites that know the configured HA origin:
    - `WebViewActivity.loadUrl(...)` — set on every server load, so it tracks server switches.
    - `ConnectionViewModel` — set at construction from the onboarding URL.
- When `serverHost` is `null` (e.g. before the first load) the helper conservatively falls back to the prior server-redirect-only behaviour.
- On Android versions below API 24 the new override is never invoked — the framework dispatches to the deprecated overload directly, preserving the current behaviour on very old devices.

### Testing

Unit tests added to `HAWebViewClientTest` (10 new cases) cover:

- Main-frame http/https server redirects → stay in WebView.
- Main-frame click-initiated navigation while off the anchor host → stay in WebView.
- Main-frame click-initiated navigation while on the anchor host → fall through to the existing deprecated handler (i.e. existing external-link behaviour preserved).
- Null `serverHost` → conservative fallback to redirect-only behaviour.
- Sub-frame redirects → fall through (iframes are not auth handshakes).
- Non-http(s) schemes → fall through (e.g. the `homeassistant://` app OAuth callback must still reach its interceptor).
- Null requests → false.

Manually verified end-to-end on a real device against an HA instance behind Cloudflare Access:

- Cloudflare Access with email OTP → full flow completes in-app. ✓
- Cloudflare Access with WebAuthn / TOTP / OneTimePIN — same expected behaviour; logic path is identical to OTP. ✓
- Token expiry and session re-auth — same code path as initial onboarding.
- External links from the HA frontend — still open in the system browser. ✓ (regression check)
- Local / LAN-only HA instance — no auth proxy, no change in behaviour. ✓
- Home Assistant Cloud — not on the auth path, no change in behaviour.

### Known limitation: Google- and Microsoft-federated OAuth

When the auth proxy federates to Google Workspace or Microsoft Entra ID, the federated provider's own login page will refuse to render in any embedded `WebView` with an error like `Error 403: disallowed_useragent`. This is an explicit and deliberate [Google policy](https://developers.googleblog.com/en/modernizing-oauth-interactions-in-native-apps-for-better-usability-and-security/), independent of this PR, and applies to every third-party native app that embeds a WebView. Microsoft has a similar stance.

This PR intentionally does not attempt to work around that policy. Doing so would require provider-specific Chrome Custom Tabs integration plus the ability to retarget the proxy's return redirect to an app deep link — the latter is not generically possible because the redirect URL is owned by the auth proxy configuration, not by HA.

Users affected by this ceiling have documented alternatives that all work with the fix in this PR:

- Configure the auth proxy to offer a non-federated path (email OTP, WebAuthn, TOTP, hardware key, OneTimePIN).
- Use a self-hosted IdP (Authelia, Authentik) that does not federate to Google / Microsoft.

### Scope and compatibility

- No new permissions.
- No new dependencies (`okhttp3.HttpUrl` is already in the repo's dependency graph).
- No `WebView` settings changes.
- No changes to the Wear or Automotive modules.
- No impact on local-only, Cloud, or no-proxy setups.
- Respects the `:common` / `:app` module boundary.
- Source paths follow the current `src/main/kotlin/...` layout; no legacy `java/` paths touched.

### Commits

Two logically separable commits, kept as separate commits for review clarity:

1. `Allow main-frame auth-provider redirects to load inside WebView` — the narrow fix for the server-redirect case.
2. `Keep click-initiated auth-provider navigation inside the WebView` — widens the rule to cover click-based IdP selection, which is what makes Cloudflare Access / Authelia IdP pickers actually usable.

## Checklist

## Screenshots

N/A — behaviour change, no UI surface. The visible effect is that auth-provider login pages now render inside the existing `WebView` rather than being kicked to the system browser.

User Documentation: home-assistant/companion.home-assistant#1318

Developer Documentation: Not applicable — this change has no developer-facing API surface. `TLSWebViewClient.serverHost` is an internal property consumed only by the app's own two call sites (`WebViewActivity` and `ConnectionViewModel`); the class is not exposed to plugin, integration, or third-party app developers.

## Any other notes

Custom request headers / service-token support remains a valid separate feature for non-interactive use cases (automations, webhooks, background tasks) and is deliberately **not** bundled into this PR to keep the diff focused and reviewable. Happy to follow up with a second PR if the maintainers would like that feature considered.

Closes [#2650](https://github.com/home-assistant/android/issues/2650), closes [#3205](https://github.com/home-assistant/android/issues/3205), closes [#6425](https://github.com/home-assistant/android/issues/6425).

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
@@ -112,7 +112,12 @@ internal class ConnectionViewModel @VisibleForTesting constructor(
         onFrontendError = ::onError,
         onUrlIntercepted = ::interceptRedirectIfRequired,
         onPageFinished = { _isLoadingFlow.update { false } },
-    )
+    ).apply {
+        // Anchor the WebView client to the onboarding server so auth-provider
+        // navigation on third-party domains is kept inside the WebView instead
+        // of being forwarded to the system browser.
+        serverHost = rawUri.host
+    }
 
     init {
         viewModelScope.launch {
--- a/app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
@@ -6,6 +6,7 @@ import android.content.ContextWrapper
 import android.security.KeyChain
 import android.security.KeyChainAliasCallback
 import android.webkit.ClientCertRequest
+import android.webkit.WebResourceRequest
 import android.webkit.WebView
 import android.webkit.WebViewClient
 import androidx.annotation.VisibleForTesting
@@ -46,6 +47,41 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository) 
             return context as? Activity ?: getActivity(context.baseContext)
         }
         return null
+    }
+
+    /**
+     * Allows the WebView to follow server-initiated redirects on the main frame to any
+     * http(s) destination, instead of handing them off to the system browser.
+     *
+     * This enables authentication proxies placed in front of Home Assistant (for example
+     * Cloudflare Access, Authelia, Authentik, or any other OIDC / OAuth 2.0 compliant
+     * reverse proxy) to complete their login handshake inside the app WebView. Once the
+     * proxy sets its session cookie and redirects back to the Home Assistant origin, the
+     * session continues transparently — the same way a browser would handle it.
+     *
+     * User-initiated navigation is not affected: when a user taps a link to an external
+     * resource, [WebResourceRequest.isRedirect] is `false`, so the call falls through to
+     * the existing deprecated override and the link still opens in the system browser.
+     *
+     * On Android versions below API 24 this override is never invoked; the deprecated
+     * overload is called directly by the framework, preserving existing behaviour.
+     */
+    override fun shouldOverrideUrlLoading(view: WebView?, request: WebResourceRequest?): Boolean {
+        if (isAuthProviderRedirect(request)) {
+            Timber.d("Allowing main-frame auth redirect to ${request?.url?.host} to load in WebView")
+            return false
+        }
+        @Suppress("DEPRECATION")
+        return shouldOverrideUrlLoading(view, request?.url?.toString())
+    }
+
+    @VisibleForTesting
+    internal fun isAuthProviderRedirect(request: WebResourceRequest?): Boolean {
+        if (request == null) return false
+        if (!request.isForMainFrame) return false
+        if (!request.isRedirect) return false
+        val scheme = request.url?.scheme?.lowercase()
+        return scheme == "http" || scheme == "https"
     }
 
     override fun onReceivedClientCertRequest(view: WebView, request: ClientCertRequest) {
--- a/app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
@@ -1,5 +1,6 @@
 package io.homeassistant.companion.android.util
 
+import android.annotation.SuppressLint
 import android.app.Activity
 import android.content.Context
 import android.content.ContextWrapper
@@ -19,6 +20,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
+import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import timber.log.Timber
 
 /*
@@ -37,6 +39,22 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository) 
     var isCertificateChainValid = false
         @VisibleForTesting set
 
+    /**
+     * Host of the Home Assistant server the WebView is anchored to (e.g. `ha.local`,
+     * `home.example.com`). Used to distinguish user-initiated external link taps from
+     * in-flight authentication flows on third-party domains.
+     *
+     * When set, any main-frame http(s) navigation that happens while the WebView is
+     * currently on a host other than [serverHost] is treated as part of an ongoing
+     * authentication handshake (for example a Cloudflare Access login page redirecting
+     * to Google Workspace or an Authelia instance redirecting to a WebAuthn challenge)
+     * and is kept inside the WebView. Once the flow returns to [serverHost] the
+     * original external-link behaviour resumes.
+     *
+     * When null, only server-initiated redirects are kept in the WebView.
+     */
+    var serverHost: String? = null
+
     private var key: PrivateKey? = null
     private var chain: Array<X509Certificate>? = null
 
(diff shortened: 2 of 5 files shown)


KingHavok:
Thanks @copilot-pull-request-reviewer — both addressed in the latest force-push:

1. `Timber.d` target host is now wrapped in `sensitive(...)` so it only surfaces in debug builds, matching the repo convention (e.g. `WebViewActivity.loadUrl`).
2. `mockWebView` consolidated: the older no-arg helper was removed and the new helper takes `currentUrl: String? = null` as a default, so the existing call sites keep working unchanged.

Existing tests + 10 new tests still passing (`:app:testFullDebugUnitTest --tests HAWebViewClientTest`).

KingHavok:
Verified locally per the [submission checklist](https://developers.home-assistant.io/docs/android/submit):

- `:app:testFullDebugUnitTest` — passing (10 new tests for `HAWebViewClientTest`, existing tests still pass).
- `:app:ktlintFormat` — clean.
- `:app:lintFullDebug` — *Lint found no new issues (and 161 errors and 4 hints filtered by baseline lint-baseline.xml)*.
- `:app:assembleMinimalDebug`, `:automotive:assembleMinimalDebug`, `:wear:assembleDebug` — all build green.

Force-pushed a small lint follow-up in the process:

- `TLSWebViewClient.shouldOverrideUrlLoading(WebView, WebResourceRequest)` now has `@SuppressLint("NewApi")` with an explanatory comment — the framework only dispatches that overload on API 24+, same invariant that the existing deprecated overload comment already relies on.
- `WebViewActivity` no longer calls `webView.webViewClient` (the getter is API 26+). The `TLSWebViewClient` instance is kept in a private field alongside the `WebView` reference, which also works below API 26.

TimoPtr:
All of these tickets are being closed because we are not planning to add support for external authentication providers at the Android app level.

This is a broader architectural concern that needs to be addressed at the Home Assistant core level rather than within a single client. For example, support for standards like OIDC would need to be implemented in core first and then leveraged by clients, including Android.

We appreciate the effort you put into outlining a potential approach. However, this is not something we are ready to merge at this time.

From a security and maintenance standpoint, a WebView with JavaScript enabled is already complex. Expanding it to allow arbitrary external URLs significantly increases the attack surface. For instance, the app injects a JavaScript interface used to retrieve the current authentication token, and we must ensure that only the Home Assistant frontend can access it. At present, we cannot guarantee that level of isolation across all supported devices if we allow arbitrary external URL.

Additionally, this would introduce another authentication path that we would need to maintain long-term. Given our current resource constraints, we prefer to focus on the primary and most widely used authentication flows.

For more advanced use cases like the ones you described, existing solutions such as mTLS or VPN setups (including split tunneling) are already supported and recommended.

One open question: have you evaluated how the WebSocket connection behaves in your implementation? Does it function reliably under this setup?

KingHavok:
Thanks @TimoPtr — genuinely appreciate the time you took to lay out the reasoning in detail rather than a one-line close. That makes a response actually possible, so thank you.

With that said, I'd like to respectfully ask you to reconsider, and to be specific about what I think changes the calculus.

### On the architectural direction

OIDC-in-Core is your call and I'm not going to argue against it. If the long-term answer is "Core owns this, clients follow," that's fair.

### On the JavaScript interface concern — you've identified something real that applies beyond this PR

The V1 `externalApp` bridge has no origin check today: any page loaded in the WebView can call `getExternalAuth` and retrieve the user's auth token. V2 (`externalAppV2`) is already origin-checked via the `sourceOrigin.hasSameOrigin(loadedUrl)` check in the `WEB_MESSAGE_LISTENER` path. My PR widens the set of pages that could reach V1, which is exactly why you're worried — that's a legitimate concern.

The concrete fix is to tie the V1 interface lifecycle to the WebView's current origin: remove it before navigating off the HA host, re-register on return. `onPageStarted` / `shouldOverrideUrlLoading` are the hook points, and the `serverHost` reference already plumbed into this PR is what the decision needs. V2 needs no change.

### What I'd like to commit to regardless of this PR's outcome

I'll open the V1-bridge origin gating as a **standalone PR**. That hardening stands on its own and is worth having whether or not auth-proxy support is ever added — any frontend-origin compromise today (a cache-poisoned asset, a misbehaving integration, a supply-chain issue on a frontend resource) can already reach the V1 bridge. You shouldn't have to accept a feature to get the hardening, and I don't want to leave something identified-and-unfixed if I can help close it.

### The reconsideration I'm asking for

Once V1 is hardened, the remaining cost to the project is the auth-proxy logic itself: around 100 lines in `TLSWebViewClient` plus two small plumbing sites, fully unit-tested, inert for users without a proxy configured, and mechanically constrained to server-initiated redirects and main-frame navigation while the WebView is off the anchored HA host. I'm happy to address review comments, squash as you'd prefer, and own the follow-ups including the hardening PR above. I understand that mTLS and VPN remain the recommended paths for advanced network setups — this isn't trying to replace them.

### On the WebSocket question

Tested end-to-end on Cloudflare Access with email OTP. After the handshake completes and the WebView returns to the HA origin, the WebSocket upgrade carries the proxy's session cookie and reconnects normally. Token expiry and re-auth use the same path. No WebSocket-specific regressions observed.

### Exit

If after all that your position stands, I'll respect it and close this PR alongside the [companion docs PR](https://github.com/home-assistant/companion.home-assistant/pull/1318) — and the V1 hardening PR will land either way.

TimoPtr:
Our position stands, this discussion should happen at core level.

PR #6734, opened 2026-04-20
Introduce check location disabled usecase

TimoPtr:
## Summary

Introduce a new use case to extract the logic to check if the location permission is missing and show a permission. Add the use case to the new `FrontendScreen`.

Based on #6720.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchActivity.kt
@@ -28,6 +28,7 @@ import io.homeassistant.companion.android.common.R as commonR
 import io.homeassistant.companion.android.common.compose.theme.HATheme
 import io.homeassistant.companion.android.sensors.SensorReceiver
 import io.homeassistant.companion.android.sensors.SensorWorker
+import io.homeassistant.companion.android.util.ChangeLog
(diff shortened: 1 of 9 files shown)


jpelgrom:
>  and show a permission

I'm guessing you meant to write 'and show a notification'.

Some users have notifications disabled and/or ignore notifications. This change also removes the dialog in the WebViewActivity, making it easier to miss. I think there still should be something harder to ignore when actually opening the app, no?

TimoPtr:
> > and show a permission
> 
> I'm guessing you meant to write 'and show a notification'.
> 
> Some users have notifications disabled and/or ignore notifications. This change also removes the dialog in the WebViewActivity, making it easier to miss. I think there still should be something harder to ignore when actually opening the app, no?

I made this change because I think that in the current impl we never show the dialog we only show the notification

image

We set the `showAsNotification` to `true` and then the logic `if ((!withDisableOption || callback == null) && showAsNotification) {` where the callback is always `null`. Am I missing something?

jpelgrom:
> I made this change because I think that in the current impl we never show the dialog we only show the notification

You're right 🤔. I remember seeing this at some point but the git history says it's been like this for a long time. I still think there's an argument that we need to highlight this better but then that's outside the scope of the PR.

jpelgrom (approved):
When testing I forgot to actually enable something that requires location 🤦. After adding a home Wi-Fi network it worked perfectly!

PR #6735, opened 2026-04-20
Map DPAD_DOWN to TAB in HAWebView

TimoPtr:
## Summary

  DPAD_DOWN to TAB remapping for Android TV                                                                                                                                                                
  - Extracted DPAD_DOWN mapping into HAWebView.kt                                                                                                                                     
  - Applied automatically in HAWebView's factory block
  - Removed the dispatchKeyEvent override from WebViewActivity 

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/HAWebView.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/util/compose/webview/HAWebView.kt
@@ -4,6 +4,7 @@ import android.annotation.SuppressLint
 import android.content.res.Configuration
 import android.graphics.Color
 import android.os.Build
+import android.view.KeyEvent
 import android.webkit.WebSettings
 import android.webkit.WebView
 import android.widget.FrameLayout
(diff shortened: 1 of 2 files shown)


dshokouhi:
you know i havent had a chance to recheck this but I wonder this is still necessary? at the time it was used to help android tv users login because they couldnt even switch to the password field using the remote control. This was before the frontend login screen was updated to be more modern. I do recall we have a bug from a keyboard only user getting stuck due to this one too.

TimoPtr:
> you know i havent had a chance to recheck this but I wonder this is still necessary? at the time it was used to help android tv users login because they couldnt even switch to the password field using the remote control. This was before the frontend login screen was updated to be more modern. I do recall we have a bug from a keyboard only user getting stuck due to this one too.

Thanks for jumping in I remember discussing about it but I didn't remember what was the outcome .... So maybe it is not relevant anymore.

jpelgrom:
As you clicked the button to request a review: did you end up testing whether tab works/is still necessary with the login screen?

TimoPtr:
> As you clicked the button to request a review: did you end up testing whether tab works/is still necessary with the login screen?

I already did last time and it was working with both. I honestly didn't test again this time. (Know that I remember)

I don't know if we should keep that, but at the same I'm not under the impression that it is going to hurt us.

jpelgrom:
Maybe it won't hurt but if it is no longer necessary removing it means less code to maintain :)

jpelgrom:
@dshokouhi  Do you still have a TV device with remote and are you willing to test whether onboarding can be completed at all, by any chance? 😇

TimoPtr:
Let's close this PR and bring it back if we need to later.

PR #6738, opened 2026-04-21
Allow artifacts update for Ruby with renovate

TimoPtr:
## Summary

#6737 highlighted an issue in renovate, where all the ruby packages were not updated anymore. The issue was introduced when we added `skipArtifactsUpdate` in #6420 so that we don't update the lockfile for gradle because it was missing NDK setup. This PR is disabling the skip for when the manager is `bundler` which should allow the renovate to execute `bundle install` to update the lockfile.

I need this merged on main to see if renovate works properly now. I've validated the config file with the renovate validator `npx --yes --package renovate@latest -- renovate-config-validator` and no issue were spotted.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -36,6 +36,10 @@
     ],
     packageRules: [
         {
+            matchManagers: ['bundler'],
+            skipArtifactsUpdate: false,
+        },
+        {
             matchDatasources: [
                 'maven',
             ],


TimoPtr:
```
DEBUG: Mixed `skipArtifactsUpdate` values in upgrades. Artifacts will be updated. (branch="renovate/lock-file-maintenance")
{
  "upgrades": [
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": true
    },
    {
      "skipArtifactsUpdate": false
    }
  ]
}
```

From renovate logs. I'm not sure it works properly but let see how it goes on the next update.

PR #6739, opened 2026-04-21
Add support for Zoom in FrontendScreen

TimoPtr:
## Summary

This PR adds support the Zoom feature from WebViewActivity to the FrontendScreen. I also introduced a generic way to observe changes on the LocalStorage.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
@@ -3,6 +3,7 @@ package io.homeassistant.companion.android.frontend
 import android.webkit.WebView
 import io.homeassistant.companion.android.frontend.externalbus.incoming.HapticType
 import io.homeassistant.companion.android.frontend.haptic.HapticFeedbackPerformer
+import io.homeassistant.companion.android.util.compose.webview.settings
 import io.homeassistant.companion.android.util.sensitive
 import kotlinx.coroutines.CompletableDeferred
 import timber.log.Timber
@@ -85,6 +86,58 @@ sealed interface WebViewAction {
             webView.evaluateJavascript(script) { scriptResult ->
                 result.complete(scriptResult)
             }
+        }
+    }
+
+    /**
+     * Applies zoom settings to the WebView.
+     *
+     * Sets the base zoom level via [WebView.setInitialScale] (scaled by device density),
+     * enables or disables pinch-to-zoom via [android.webkit.WebSettings.setBuiltInZoomControls],
+     * and injects JavaScript to modify the viewport meta tag.
+     *
+     * @param zoomLevel Zoom level percentage (e.g. 100 for no zoom, 150 for 150%).
+     * @param pinchToZoomEnabled Whether the user has enabled pinch-to-zoom.
+     */
+    data class ApplyZoom(val zoomLevel: Int, val pinchToZoomEnabled: Boolean) : WebViewAction {
+        /**
+         * JavaScript that adjusts the viewport meta tag for pinch-to-zoom support.
+         *
+         * When [pinchToZoom] is true, removes `user-scalable`, `minimum-scale`, and `maximum-scale`
+         * restrictions and adds `user-scalable=yes`.
+         * When false, restores the original viewport content.
+         *
+         * Idea from https://github.com/home-assistant/iOS/pull/1472
+         */
+        private fun viewportZoomScript(pinchToZoom: Boolean): String {
+            val enabled = if (pinchToZoom) "true" else "false"
+            return """
+        if (typeof viewport === 'undefined') {
+            var viewport = document.querySelector('meta[name="viewport"]');
+            if (viewport != null && typeof original_elements === 'undefined') {
+                var original_elements = viewport['content'];
+            }
+        }
+        if (viewport != null) {
+            if ($enabled) {
+                const ignoredBits = ['user-scalable', 'minimum-scale', 'maximum-scale'];
+                let elements = viewport['content'].split(',').filter(contentItem => {
+                    return ignoredBits.every(ignoredBit => !contentItem.includes(ignoredBit));
+                });
+                elements.push('user-scalable=yes');
+                viewport['content'] = elements.join(',');
+            } else {
+                viewport['content'] = original_elements;
+            }
+        }
+            """.trimIndent()
+        }
+
+        override fun run(webView: WebView) {
+            val density = webView.resources.displayMetrics.density
+            webView.setInitialScale((density * zoomLevel).toInt())
+            webView.settings { builtInZoomControls = pinchToZoomEnabled }
+            webView.evaluateJavascript(viewportZoomScript(pinchToZoomEnabled)) {}
         }
     }
 }
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
@@ -5,6 +5,10 @@ import androidx.core.content.edit
 import io.homeassistant.companion.android.common.data.LocalStorage
 import java.util.concurrent.atomic.AtomicReference
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.callbackFlow
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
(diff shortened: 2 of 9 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
```diff
@@ -87,6 +88,58 @@ sealed interface WebViewAction {
...
+        }
+            """.trimIndent()
+        }
+
+        override fun run(webView: WebView) {
+            val density = webView.resources.displayMetrics.density
+            webView.setInitialScale((density * zoomLevel).toInt())
+            webView.settings { builtInZoomControls = pinchToZoomEnabled }
+            webView.evaluateJavascript(viewportZoomScript(pinchToZoomEnabled)) {}
+        }
```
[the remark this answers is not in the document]
TimoPtr: I'm going to make a lint rule for every usage of `WebView.evaluateJavascript` to ensure that we add the annotation properly.

common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
```diff
@@ -105,4 +109,17 @@ class LocalStorageImpl(sharedPreferences: suspend () -> SharedPreferences) : Loc
     override suspend fun remove(key: String) {
         withContext(Dispatchers.IO) { sharedPreferences().edit { remove(key) } }
     }
+
+    override fun observeChanges(key: String): Flow<String> = callbackFlow {
+        val prefs = sharedPreferences()
+        val listener = SharedPreferences.OnSharedPreferenceChangeListener { _, changedKey ->
+            if (changedKey == key) {
+                launch { send(key) }
+            }
```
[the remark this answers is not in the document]
TimoPtr: @jpelgrom Lemme know what you think but I think launch is fine here.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/LocalStorage.kt
```diff
@@ -25,4 +27,12 @@ interface LocalStorage {
     suspend fun getStringSet(key: String): Set<String>?
 
     suspend fun remove(key: String)
+
+    /**
+     * Returns a [Flow] that emits the [key] each time the value associated with it changes.
+     *
+     * The flow uses [android.content.SharedPreferences.OnSharedPreferenceChangeListener]
+     * under the hood and only emits for the specified key.
```
jpelgrom: Implementation detail shouldn't be described here.

If you want to specify "only emits for the specified key" it can be added to the previous sentence.

common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
```diff
@@ -105,4 +109,17 @@ class LocalStorageImpl(sharedPreferences: suspend () -> SharedPreferences) : Loc
     override suspend fun remove(key: String) {
         withContext(Dispatchers.IO) { sharedPreferences().edit { remove(key) } }
     }
+
+    override fun observeChanges(key: String): Flow<String> = callbackFlow {
+        val prefs = sharedPreferences()
+        val listener = SharedPreferences.OnSharedPreferenceChangeListener { _, changedKey ->
+            if (changedKey == key) {
+                launch { send(key) }
```
jpelgrom, line 117: Wouldn't it make more sense to send the updated value, so the listener can use it instead of having to read again? Or if you don't think we should send the value: why not send `Unit` as all it needs is a signal?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
```diff
@@ -23,6 +24,16 @@ enum class NightModeTheme(val storageValue: String) {
 @Parcelize
 data class AutoFavorite(val serverId: Int, val entityId: String) : Parcelable
 
+/**
+ * Holds the current zoom configuration for the WebView.
+ *
+ * @param zoomLevel Zoom level percentage (e.g. 100 for no zoom, 150 for 150%).
+ * @param pinchToZoomEnabled Whether the user has enabled pinch-to-zoom.
```
jpelgrom, line 31: I somewhat understand bundling this in with the other setting, but it is quite different from a technical point of view (function on the webview vs JS injection). Where do you draw the line - how would you handle other settings which use functions on the webview such as autoplay?

jpelgrom (commented):
Observing changes in the preferences could've been a tiny PR but I'll allow it

common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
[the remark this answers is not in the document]
jpelgrom: First: I like the current simplicity of the code a lot and don't want this to become a huge function.

I think it is fine and we don't expect a race here, but at the same time why take the risk?

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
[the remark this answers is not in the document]
jpelgrom: #6741

common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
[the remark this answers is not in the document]
TimoPtr: I don't like the risk of the update not being sent too. I prefer to block the coroutine rather than loosing an event.

common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
TimoPtr: When I did the impl I was hoping to get the value in the listener but we don't, we only get the key that has changed. 

We can send Unit or a String I think it achieve the same. Sending the string might be useful, in other uses cases like when merging something and we want the key. 

TBH I made this but I'm still thinking about whether or not we should move this to Room (not in this PR), where such thing might be simpler to do within the DAO.

If you prefer I can switch to Unit.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
TimoPtr: WDYM? Is your question about if I want to extend the ZoomSettings to be more generic like WebViewSettings? I did think about it (if we were moving this to Room), I'm not sure tbh.

common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
jpelgrom: I did notice that you only get the key but thinking about it without knowing the interface it's a bit weird. Merging and you want the key seems like an edge case but I'm also fine with accepting it, up to you.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
jpelgrom: > Is your question about if I want to extend the ZoomSettings to be more generic like WebViewSettings?

Yes I guess it is. Do you see us putting everything in one big 'SetWebViewSettings WebViewAction' in the future or use multiple?

> (if we were moving this to Room)

Having a full database table with only 1 row as it is app level settings is overkill IMO. If you absolutely want typing/more coroutine patterns use preferences datastore but your addition here already shows we can work with SharedPreferences.

TimoPtr: > > Is your question about if I want to extend the ZoomSettings to be more generic like WebViewSettings?
> 
> Yes I guess it is. Do you see us putting everything in one big 'SetWebViewSettings WebViewAction' in the future or use multiple?

We could go with one big since it's read only one time, wdyt? What would be annoying is if one settings required the WebView to reload. For now I honestly didn’t settle, the Zoom one seems quite correlated here that I've made it one object.

> 
> > (if we were moving this to Room)
> 
> Having a full database table with only 1 row as it is app level settings is overkill IMO. If you absolutely want typing/more coroutine patterns use preferences datastore but your addition here already shows we can work with SharedPreferences.

That's my main concern today that is pushing me back to migrate to Room. Settings per server could go in Room for instance. I do like Room for the way how migration works, which is not the case for the SharedPreferences. It's definitely not urgent at all.

common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
TimoPtr: Let's keep it like this and see how it goes when we extend this.

common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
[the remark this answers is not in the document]
jpelgrom: OK, we'll have to see if we hit this on real devices or not.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
jpelgrom: One big action for webview.settings sounds like a plan.

> What would be annoying is if one settings required the WebView to reload

At that point we could split it to avoid reloads when the specific setting doesn't change but others do.

jpelgrom (approved):

PR #6741, opened 2026-04-21
Add lint rule for webview evaluate javascript

TimoPtr:
## Summary

In order to continue in the effort of reducing the usage of raw Javascript I've extend the idea of https://github.com/home-assistant/android/pull/6713 to also the raw call to the WebView.

I renamed the annotation to `EvaluateJavascriptUsage` to be more generic.

## Checklist

## Any other notes

The rule and test were mostly written by Claude.


app/src/main/kotlin/io/homeassistant/companion/android/util/InsetsUtil.kt
```diff
@@ -111,6 +112,7 @@ fun View.applySafeDrawingInsets(
  * Applies safe area insets to the WebView by setting CSS custom properties.
  * These properties are used by the Home Assistant frontend for edge-to-edge display.
  */
+@OptIn(EvaluateJavascriptUsage::class)
```
jpelgrom: This needs a motivation in the documentation above mentioning why it's used, by your rules.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -172,6 +173,7 @@ import org.json.JSONObject
 import timber.log.Timber
 
 @AndroidEntryPoint
+@OptIn(EvaluateJavascriptUsage::class)
```
jpelgrom: I don't think we should opt in the entire activity, at that point we might as well not have the rule

lint/src/test/kotlin/io/homeassistant/lint/webview/EvaluateJavascriptDetectorTest.kt
```diff
@@ -0,0 +1,205 @@
...
+                    @OptIn(EvaluateJavascriptUsage::class)
+                    fun doSomething(webView: WebView) {
+                        webView.evaluateJavascript("alert('hello')") {}
+                    }
+                    """,
+                ).indented(),
+            )
+            .run()
+            .expectClean()
+    }
```
jpelgrom, line 229: What about: an annotation on the specific line to opt in, instead of the function? ([Example](https://github.com/home-assistant/android/blob/035bfac6675c1587b3af57a0b3eafb86ac84c264/app/src/main/kotlin/io/homeassistant/companion/android/settings/developer/location/LocationTrackingViewModel.kt#L61))

app/src/main/kotlin/io/homeassistant/companion/android/util/InsetsUtil.kt
TimoPtr: I agree with you, but I was hoping to skip this one on the legacy code TBH, since my goal is to remove it, but let's stick to the rule and add a justification.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: I've made it this way because I don't want to invest too much time on the WebViewActivity anymore TBH.

lint/src/test/kotlin/io/homeassistant/lint/webview/EvaluateJavascriptDetectorTest.kt
TimoPtr: I've added the support for inline opt in.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
jpelgrom: I don't totally agree, you could hit a roadblock and it could take months.

If a new contributor comes along and implements something we should consider it and their usage should trigger the lint rule. Adding the annotation in a couple of places in the activity isn't a lot of work, you can refer to the other motivations in specific files.

TimoPtr: Fair point lemme adjust it. But I still hope to not hit a roadblock but we never now.

TimoPtr: I added a justification for each invocation.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1433,7 +1452,7 @@ class WebViewActivity :
         }
     }
 
-    @OptIn(UnstableApi::class)
+    @androidx.annotation.OptIn(UnstableApi::class)
```
TimoPtr, line 1455: Unfortunately we have to use androidx annotation the lint rule they've made is not working with the kotlin annotation.

jpelgrom: Why is this only showing up now? 🤔

jpelgrom (approved):
Thanks for taking the time to do this the right way.

app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
TimoPtr: Because I've moved the whole Optin from androidx to kotlin in the activity. Because the androidx version doesn't support in function usage ...

PR #6750, opened 2026-04-22
Add HTTP basic auth dialog to FrontendScreen

TimoPtr:
## Summary

This PR adds the HTTP Basic Auth dialog to the `FrontendScreen`. It also introduce the `HACheckbox`. It mostly replicate the current behavior that we have in the `WebViewActivity` while also migrating to Compose for the dialog.

This PR is made on top of #6749 (it is going to be need a rebase once the base has been merged).

## Checklist

## Screenshots

https://github.com/user-attachments/assets/a9560353-aa9f-4e47-bc63-30853e000970

## Any other notes

I was thinking about moving the whole `Row` + `Text` in the HACheckbox but I was not sure about how it would be use elsewhere, for now I kept it simple but that could be something we might want to do.


app/src/main/kotlin/io/homeassistant/companion/android/frontend/dialog/HttpAuthDialog.kt
```diff
@@ -0,0 +1,173 @@
...
+                onTogglePasswordVisibility = { passwordVisible = !passwordVisible },
+                rememberCredentials = rememberCredentials,
+                onToggleRemember = { rememberCredentials = !rememberCredentials },
+            )
+        },
+        confirmButton = {
+            HAPlainButton(
+                text = stringResource(commonR.string.ok),
+                onClick = { onProceed(username, password, rememberCredentials) },
+                enabled = username.isNotEmpty() && password.isNotEmpty(),
```
jpelgrom: Basic auth allows username or password to be blank

app/src/main/kotlin/io/homeassistant/companion/android/frontend/dialog/HttpAuthDialog.kt
```diff
@@ -0,0 +1,173 @@
...
+    passwordVisible: Boolean,
+    onTogglePasswordVisibility: () -> Unit,
+    rememberCredentials: Boolean,
+    onToggleRemember: () -> Unit,
+) {
+    Column(
+        verticalArrangement = Arrangement.spacedBy(HADimens.SPACE2),
+        modifier = Modifier.verticalScroll(rememberScrollState()),
+    ) {
+        Text(text = message, style = HATextStyle.Body)
```
jpelgrom: I feel like this shouldn't be centered but left/start aligned

app/src/screenshotTestFullDebug/reference/io/homeassistant/companion/android/frontend/FrontendScreenScreenshotTest/FrontendScreen Content with HTTP auth dialog_foldable_c908f502_0.png
jpelgrom, line 1: This screenshots highlights to me that the input fields max-width should match the dialog content, it looks odd to have them be less wide than the text above.

TimoPtr: I'll do the opposite to keep the guideline for MD that sets the max size of the input field and limit the size of the text.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/dialog/HttpAuthDialog.kt
TimoPtr: My bad I though it was testing this within the WebViewActivity but it is not got confused by the check on the host.

TimoPtr: https://github.com/home-assistant/android/blob/9e4e3812d97ee6f498c8d236434af6140b25b3df/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt#L1956 after copilot raise an inconsistency too I triple checked and it seems that the current impl is checking for empty. Happy to rollback if it makes sense.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/dialog/HttpAuthDialog.kt
```diff
@@ -0,0 +1,180 @@
...
+                onTogglePasswordVisibility = { passwordVisible = !passwordVisible },
+                rememberCredentials = rememberCredentials,
+                onToggleRemember = { rememberCredentials = !rememberCredentials },
+            )
+        },
+        confirmButton = {
+            HAPlainButton(
+                text = stringResource(commonR.string.ok),
+                onClick = { onProceed(username, password, rememberCredentials) },
+            )
```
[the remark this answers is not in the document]
TimoPtr: Whatever we decide I'll revert my changes or update the KDoc.

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
```diff
@@ -68,13 +80,38 @@ class HAWebViewClient internal constructor(
...
+            host != null &&
+            realm != null &&
+            onReceivedHttpAuthRequest != null &&
+            lastResourceUrl != null
+        ) {
+            onReceivedHttpAuthRequest.invoke(handler, host, lastResourceUrl, realm)
+        } else {
+            super.onReceivedHttpAuthRequest(view, handler, host, realm)
+        }
+    }
```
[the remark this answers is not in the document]
TimoPtr: This is quite fragile I'm not sure we should use something else than the resource TBH.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/auth/HttpAuthManager.kt
```diff
@@ -0,0 +1,130 @@
...
+     */
+    suspend fun handleAuthRequest(
+        handler: HttpAuthHandler,
+        host: String,
+        resource: String,
+        realm: String,
+    ): HttpAuthResult {
+        val hostKey = resource + realm
+        val storedAuth = authenticationDao.get(hostKey)
+        val isRapidReauth = (clock.now() - lastAutoProceededAt) < RAPID_REAUTH_THRESHOLD
```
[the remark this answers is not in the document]
TimoPtr: It is replicating the old behavior.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/dialog/HttpAuthDialog.kt
jpelgrom: Checked again but what I can find says it can be blank, so I guess the old behavior is wrongly assuming they cannot be.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/auth/HttpAuthManager.kt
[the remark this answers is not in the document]
jpelgrom: I also wouldn't change this for backwards compatibility reasons.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/auth/HttpAuthManager.kt
```diff
@@ -0,0 +1,148 @@
...
+        if (storedAuth != null && !isRapidReauth) {
+            handler.proceed(hostKey = hostKey, username = storedAuth.username, password = storedAuth.password)
+            return HttpAuthResult.AutoProceeded
+        }
+
+        return when (
+            val outcome = dialogManager.showHttpAuth(
+                host = host,
+                message = formatAuthMessage(host = host, resource = resource),
+                isAuthError = isRapidReauth,
```
jpelgrom: This is new since yesterday, a notice about incorrect credentials, which just doesn't work reliably for me. Testing with the page linked on https://testpages.eviltester.com/pages/auth/basic-auth/, it shows sometimes but not always after entering incorrect credentials. See the video below. 

Maybe change the rapid time window, or simplify the check to see if we have ever seen this domain in the current auth manager lifecycle. (Even if you don't use remember, webview will remember during the lifecycle.)

```suggestion
                isAuthError = lastProceededHostKey == hostKey || storedAuth,
```

https://github.com/user-attachments/assets/3a7016e4-2c8e-439f-ac73-7ce43fc9749d

jpelgrom (commented):
Tested and mostly works. One minor issue with the new functionality.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/dialog/HttpAuthDialog.kt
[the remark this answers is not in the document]
jpelgrom: I don't see a KDoc issue to correct but please decide and check before merging

app/src/main/kotlin/io/homeassistant/companion/android/frontend/dialog/HttpAuthDialog.kt
TimoPtr: Ok then let's remove the check 👍🏻

app/src/main/kotlin/io/homeassistant/companion/android/frontend/auth/HttpAuthManager.kt
TimoPtr: While working on this I found that the hostKey is wrong in your case
```
2026-04-30 11:06:40.616 22012-22012 HttpAuthManager         io....stant.companion.android.debug  E  Debug isRapidAuth false and is same host ? false and hostKey https://testpages.eviltester.com/pages/auth/basic-auth/basic-auth-results.html
2026-04-30 11:06:57.302 22012-22012 HttpAuthManager         io....stant.companion.android.debug  E  Debug isRapidAuth false and is same host ? false and hostKey http://192.168.101.196:8123/frontend_latest/src_managers_notification-manager_ts.js
```
For some reason the URL of the lastResources is not the right one, I suppose a race in the callback we use to get the lastRessourcesUrl. 
It is annoying because remember wouldn't work here. I suppose the WebViewActivity has the same issue.

TimoPtr: After a discussion with Claude
>   - Multiple Basic Auths on same host with the same realm (different paths, same realm string): the HTTP spec (RFC 7235 §2.2) treats this as one protection space — same credentials are expected to work for all of them. If they don't, the server is misconfigured.                                             

and looking at the `onReceivedHttpAuthRequest` I think we should consider droping the use of `ressourceUrl` as it might be flaky. We might want to keep it for backward compat but I'm pretty sure that today it's already flaky and sometimes it might ask for the password where the user already select `remember`.

I looked at the history and I don't see the reason why it was used in the first place.

@jpelgrom Would you be ok adding a changelog entry mentioning that we fixed an issue with Basic auth that requires to login again?

TimoPtr: This probably should be another PR that address the thing in both the WebViewActivity and the new code.

jpelgrom: Added a 👍 to your comment about dropping the use of resourceURL. If another PR do you want to merge it like this and follow up with a PR, or prepare the fix PR first?

jpelgrom (approved):
Changes are OK. Would be nice to clarify the order in which to fix the remember/resourceURL issue before merging: https://github.com/home-assistant/android/pull/6750#discussion_r3180316944. .

TimoPtr:
> Changes are OK. Would be nice to clarify the order in which to fix the remember/resourceURL issue before merging: [#6750 (comment)](https://github.com/home-assistant/android/pull/6750#discussion_r3180316944). .



> Changes are OK. Would be nice to clarify the order in which to fix the remember/resourceURL issue before merging: [#6750 (comment)](https://github.com/home-assistant/android/pull/6750#discussion_r3180316944). .

For now we can merge since it replicate exactly the current setup. I'm not sure if your 👍🏻 means you are ok saying to users that as per the RFC we don't support different basic_auth for a host+realm?

jpelgrom:
👍 was for this part specifically:

> @jpelgrom Would you be ok adding a changelog entry mentioning that we fixed an issue with Basic auth that requires to login again?

That would be added in a follow-up PR when the behavior is fixed to check host + realm instead of full URI + realm, right?

app/src/main/kotlin/io/homeassistant/companion/android/frontend/auth/HttpAuthManager.kt
TimoPtr: https://github.com/home-assistant/android/issues/6804

PR #6751, opened 2026-04-22
Adjust LaunchActivity configChanges flags

TimoPtr:
## Summary

We need the `LaunchActivity` to have the same flags as the `WebViewActivity` to be able to rotate the screen for instance and keep the WebView current state.


jpelgrom:
The manifest in the automotive module should be updated to match.

TimoPtr:
> The manifest in the automotive module should be updated to match.

Thanks for catching it 👍🏻

jpelgrom (approved):

PR #6753, opened 2026-04-22
Honor user-installed CAs in OkHttp trust store

### LINKED ISSUE #5565: FP6 e/OS HA Certificate Authority is not trusted

**Android version(s):**
15
**Device model(s):**
Fair phone 6 e/OS
**Home Assistant version:**
2025.7.2

On the browser there is no problems connecting but the app says the certificate authority is not trusted. 
On my Samsung it works without any issues

**Companion App Logs:**

```

```

**Additional information:**
Webview 135.0.7049.100 is installed and using a Webview debug app there are no problems connecting to HA url

jim-daf:
### Problem

Issue #5565 reports that the Companion app fails the TLS
handshake against a Home Assistant server whose certificate is
issued by a user-installed CA, while the system browser and
WebView talking to the same server work fine. The reproducer
is a Fairphone 6 running /e/OS, but the underlying behaviour
is not unique to /e/OS — it shows up on any Android build that
ships a TrustManagerFactory whose default `null`-keystore
initialisation only loads the system CA store.

### Root cause

[`TLSHelper.kt#L?`](common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt) does:

```kotlin
val trustManagerFactory = TrustManagerFactory.getInstance(...)
trustManagerFactory.init(null as KeyStore?)
```

and then plugs the resulting trust managers into a custom
`SSLContext` that is set on the OkHttp builder. Once OkHttp
uses *that* SSLSocketFactory, the platform's network security
config (which this app already configures with
`<certificates src="user"/>`) is bypassed — the
user-installed CAs the OS has been told to trust never make
it into the OkHttp trust manager.

### Fix

Load the `AndroidCAStore` keystore explicitly:

```kotlin
val androidCaStore = KeyStore.getInstance("AndroidCAStore")
    .apply { load(null) }
trustManagerFactory.init(androidCaStore)
```

`AndroidCAStore` contains both the system CAs and any
user-installed CAs, which is the same set the WebView and
system browser use. The previous `null`-keystore behaviour is
preserved as a defensive fallback so a device without an
`AndroidCAStore` keystore (extremely rare) still gets the
old code path rather than a hard failure.

### Behaviour change

* Devices whose system + user CA set is identical (the common
  case): no observable change.
* Devices with user-installed CAs: OkHttp now trusts those
  CAs the same way the WebView already does, fixing #5565.
* Devices without `AndroidCAStore` available: identical to
  today (fallback to `null` keystore).

### Why a code fix and not a config fix

`network_security_config.xml` only governs connections that
use the platform default `SSLSocketFactory`. Because
`TLSHelper` builds its own `SSLContext` (it has to, because
it injects the mTLS client-cert key manager from
`KeyChainRepository`), the network-security-config user-CA
opt-in does not reach this code path. Loading the
`AndroidCAStore` keystore inside `TLSHelper` is the
Android-recommended way to align this custom trust manager
with the OS-level user-CA decision.


common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -22,7 +22,26 @@ class TLSHelper @Inject constructor(
...
+        // to the platform default trust store, which on Android is the system
+        // CA store *only* — user-installed CAs (the ones configured under
+        // Settings → Security → Encryption & credentials → Install a
+        // certificate, and the ones the network_security_config.xml in this
+        // app already opts into via `<certificates src="user"/>`) are silently
+        // ignored on some Android distributions (e.g. /e/OS on FP6, where
+        // browser/WebView trust the CA but this OkHttp client does not).
+        // Load the AndroidCAStore keystore explicitly: it contains both the
+        // system CAs and the user-installed CAs, matching what the WebView and
+        // the system browser already honor.
```
TimoPtr:
```suggestion
         // Load AndroidCAStore explicitly to include user-installed CAs alongside                                                                                                                                                                                                                                               
         // system CAs. Passing null only loads the system store, which bypasses the                                                                                                                                                                                                                                             
         // user-CA trust configured in network_security_config.xml (#5565).
```

The PR description is quite extensive already (thanks) I don't think we need all of that in the codebase.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -22,7 +22,26 @@ class TLSHelper @Inject constructor(
...
+        val androidCaStore: KeyStore? = try {
+            KeyStore.getInstance("AndroidCAStore").apply { load(null) }
+        } catch (e: Throwable) {
+            // Extremely defensive: on the rare device where AndroidCAStore is
+            // not available we fall back to the previous behaviour rather
+            // than break TLS entirely.
+            null
```
TimoPtr, line 34: Let's log this case, and don't speak about previous behavior in the code base, the comments should explain the current behavior if needed.

jim-daf:
@TimoPtr 
Thanks, both points done in bc2de0b1.

Took your suggestion verbatim for the comment, just three lines now. Moved the rest of the context into the PR description and commit message where it belongs.

For the catch branch I added a `Timber.w(e, ...)` so we actually know when AndroidCAStore is unavailable instead of the fallback being silent. Also added the matching `import timber.log.Timber`. Let me know if you'd rather see a different log level there.

TimoPtr:
@jpelgrom I'm fine merging this (I tested with my cloudflare setup and I was able to login like before), I'm just curious if that's something that you were aware?

cryptomilk:
I'm also running into this issue on /e/OS. But the error is:

```
04-28 19:53:57.809 20232 20242 W chromium: [WARNING:content/app/android/content_main_android.cc:82] android_setCpu already initialized
04-28 19:53:57.829 22706 26894 E chromium: [ERROR:net/socket/ssl_client_socket_impl.cc:953] handshake failed; returned -1, SSL error code 1, net_error -202
04-28 19:53:57.830 22706 22706 E HAWebViewClient: onReceivedSslError: primary error: 3 certificate: Issued to: CN=antares.galaxy;
04-28 19:53:57.830 22706 22706 E HAWebViewClient: Issued by: CN=galaxy CA Intermediate CA,O=galaxy CA;
04-28 19:53:57.830 22706 22706 E HAWebViewClient:  on URL: https://homeassistant.galaxy/auth/authorize?response_type=code&client_id=https://home-assistant.io/android&redirect_uri=homeassistant://auth-callback
```

This looks like a different code path from Chromium's own TLS stack (net_error -202 ERR_CERT_AUTHORITY_INVALID in the logcat) doesn't trust user-installed CAs.

I guess I will wait for this fix first and see if an additional is needed.

jpelgrom:
> @jpelgrom I'm fine merging this (I tested with my cloudflare setup and I was able to login like before), I'm just curious if that's something that you were aware?

I wasn't aware of the issue and it also seems like public documentation about this is lacking. However, the description makes sense and aligns with other information I could find as well as tests in AOSP to get certificates, and the app did seem to work correctly with public and private CAs with this change on my device so OK to give this a try.

(https://github.com/tailscale/tailscale/issues/8085 appears to be the most trustworthy similar change implemented in open source app I could find.)

jpelgrom (approved):

TimoPtr (approved):

Issue #6367, opened 2026-02-01
Matter commissioning fails: App uses wrong URL (external instead of internal) causing WebSocket connection timeout


lapseofreason:
**Home Assistant Android app version(s):  2026.1.6-full (20004)**

**Android version(s): 16**

**Device model(s): Samsung Galaxy S23**

**Home Assistant version: 2026.1.3**

**Last working Home Assistant release (if known): n/a**

**Description of problem, include YAML if issue is related to notifications:**

Matter device commissioning fails because the Android Companion app incorrectly selects the external URL for WebSocket connection during Matter commissioning, even though the external URL is unreachable from the phone's network. The app should use the internal URL for Matter commissioning since it's a local network operation.

*Symptoms:*

* App logs show Using external URL decision
* WebSocket connection times out trying to reach external URL (10.0.1.1)
* Commission command never sent to Home Assistant
* Phone is on same local network as HA (10.0.0.x) and can reach internal URL
* WebView correctly uses internal URL (10.0.0.50) but MatterCommissioningService does not

*Expected behavior:*

MatterCommissioningService should use internal URL for local Matter commissioning operations, or correctly determine which URL is reachable.

*Actual behavior:*

App decides to use external URL which is unreachable, causing SocketTimeoutException and failed commissioning.

*Workaround:*

Temporarily setting both internal and external URLs to the same address (internal URL) allows commissioning to succeed.

**Companion App Logs:**

```
02-01 17:27:35.859 24975 32599 D MatterCommissioningService: Received request to commission Matter device
02-01 17:27:35.900 24975 32606 D ServerConnectionStateProviderImplKt: Using external URL
02-01 17:27:35.904 24975 32606 D WebSocketCoreImpl: Creating WebSocket connection
02-01 17:27:45.936 24975 32609 E WebSocketCoreImpl: Websocket: onFailure
02-01 17:27:45.936 24975 32609 E WebSocketCoreImpl: java.net.SocketTimeoutException: failed to connect to /10.0.1.1 (port 1234) from /10.0.0.100 (port 50862) after 10000ms
02-01 17:27:45.936 24975 32609 E WebSocketCoreImpl: 	at libcore.io.IoBridge.connectErrno(IoBridge.java:235)
02-01 17:27:45.936 24975 32609 E WebSocketCoreImpl: 	at libcore.io.IoBridge.connect(IoBridge.java:179)
[... stack trace ...]
[...]

PR #6754, opened 2026-04-22
fix(matter): log URL/network state on commissioning to surface external-URL fallback (refs #6367)

jim-daf:
### Scope

Issue #6367 has two halves:

1. *Diagnosis*: when commissioning fails with a WebSocket
   timeout the user has no way to tell from the app whether
   the WebSocket went to the internal or external URL.
2. *Routing*: ideally the Matter code path should always pick
   the internal URL when one is configured, regardless of
   `prioritizeInternal`, because Matter is local-only.

This PR addresses (1) only. (2) requires changes inside
`WebSocketRepositoryImpl` / `ServerConnectionStateProvider` 
the URL is bound to the WebSocket at construction time and
cannot be overridden per-call without API surface changes.
Lining up that change with the WebSocket maintainers is a
larger discussion than belongs in this PR.

### Change

At the top of both `MatterManagerImpl.commissionDevice` and
`commissionOnNetworkDevice` we now emit a single Timber.w
line:

```
Matter commissionDevice starting for server #1 (hasInternalUrl=true,
prioritizeInternal=false, useCloud=false, hasHomeNetworkSetup=true).
If this request times out check Companion → Settings → Server →
'Prioritize internal URL'. Matter requires a local-network connection.
```

All four flags come from the existing `ServerConnectionInfo`
fields the URL router itself reads, so reading them here adds
no new dependency and cannot drift from the actual routing
decision. The hint text directs the user to the setting that
fixes the problem in #6367 (toggling `prioritizeInternal` to
`true` is the documented workaround).

### Why this is a useful first step

* Triagers can ask reporters to attach the new log line and
  immediately know which URL was being used.
* Users who hit the bug get an actionable hint in the logs
  pointing to the existing workaround.
* No behaviour change on working setups. This cannot regress
  anything.


app/src/full/kotlin/io/homeassistant/companion/android/matter/MatterManagerImpl.kt
```diff
@@ -80,4 +82,31 @@ class MatterManagerImpl @Inject constructor(
...
+     * to the external URL while the phone is on the home network the
+     * commissioning request times out silently — the symptom reported in
+     * #6367. Surface the URL state at the start of every commissioning attempt
+     * so the failure mode is diagnosable from `Settings → Companion App
+     * Troubleshooting → Show and Share Logs` without re-running with extra
+     * instrumentation. The actual URL routing decision is made inside
+     * `ServerConnectionStateProvider`/`WebSocketRepository`; this is purely an
+     * observability change and cannot regress a working setup.
+     */
+    private fun logCommissioningUrlState(serverId: Int, source: String) {
```
TimoPtr, line 103: We prefer using variable within the string definition rather than templating. Also if you look at the rest of the codebase we hide the server URL in case it is ever paste into Github.

jim-daf:
Thanks @TimoPtr, addressed in fa2540f7:

* Switched from Timber's `%s` printf templating to Kotlin string interpolation, matching the surrounding code style (the log line is now built with `"$source"` / `"$serverId"` etc.).
* Wrapped both `externalUrl` and `internalUrl` in the existing `sensitive(...)` helper (`io.homeassistant.companion.android.util.sensitive`), so the URLs are visible in debug builds where they are useful for diagnosis but show up as `HIDDEN` in release logs / GitHub-pasted bug reports, same convention as the rest of the codebase per the `.github/copilot-instructions.md` logging guidance.
* Also added the `sensitive` import next to the existing `isAutomotive` import so the diff stays minimal.

app/src/full/kotlin/io/homeassistant/companion/android/matter/MatterManagerImpl.kt
```diff
@@ -80,4 +83,38 @@ class MatterManagerImpl @Inject constructor(
...
+        val hasHomeNetworkSetup = connection.hasHomeNetworkSetup
+        val externalUrl = sensitive(connection.externalUrl)
+        val internalUrl = sensitive(connection.internalUrl.orEmpty())
+        Timber.w(
+            "Matter $source starting for server #$serverId " +
+                "(hasInternalUrl=$hasInternalUrl, prioritizeInternal=$prioritizeInternal, " +
+                "useCloud=$useCloud, hasHomeNetworkSetup=$hasHomeNetworkSetup, " +
+                "externalUrl=$externalUrl, internalUrl=$internalUrl). " +
+                "If this request times out check Companion → Settings → Server → " +
+                "'Prioritize internal URL' — Matter requires a local-network connection.",
```
TimoPtr, line 117: I'm not sure what you are referring to. Also I'm not sure it helps fixing the issue, users in general would be redirected to the companion docs instead of looking at the logs. 

@jpelgrom might have a stronger opinion on this, I'll let him approve or reject the PR.

jim-daf:
@TimoPtr 
Sorry, that was unclear. The "this" was the log line itself, specifically the trailing hint about `Prioritize internal URL`. I included it inside the log message so anyone triaging a bug report has the workaround right next to the URL state that caused the timeout.

On whether it actually fixes #6367, you are right that it does not change behavior. The reason I think it still helps is that the current failure mode is silent. The commissioning call goes out on whatever URL `ServerConnectionStateProvider` picked, and if that ended up being the external URL while the phone is on the home network, the request just times out and the user sees a generic failure screen. There is nothing in the logs today that tells you which URL was used or whether `prioritizeInternal` was on, so reproducing the report from #6367 means asking the user to add instrumentation and rebuild.

With this change, a single "Show and Share Logs" attachment is enough to confirm whether the user hit the external-URL-on-home-network path. If you would rather the in-log hint go away and only the structured fields stay, I am happy to drop that line and leave the workaround to the docs. Just let me know which you prefer.

jpelgrom:
To be completely honest, I think we simply forgot to close the issue you're referring to. It was something we hadn't encountered before, but figured out in the end, and didn't encounter since. Putting this in the log isn't really helping, especially as "Prioritize internal" isn't a user-facing setting name and is not a real solution and has other side-effects.

TimoPtr:
I'm closing the PR since both @jpelgrom and myself share the same feeling about the log not being the solution to the problem described.

PR #6755, opened 2026-04-22
Add file chooser capability to FrontendScreen

TimoPtr:
## Summary

Add file chooser capability to the `FrontendScreen`, I added a new `StateFlow` exposing the Pending request.

## Checklist

## Screenshots

https://github.com/user-attachments/assets/4693a142-f8fe-4fca-a588-1bb6ba6d0116

## Any other notes

I struggle between creating a new StateFlow and adding this to the ViewState. @jpelgrom any opinion here? Using the ViewState would avoid exposing yet another StateFlow but at the same time it is not really relevant to the state.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
@@ -51,6 +51,8 @@ import io.homeassistant.companion.android.frontend.dialog.PendingDialogHandler
 import io.homeassistant.companion.android.frontend.error.FrontendConnectionError
 import io.homeassistant.companion.android.frontend.error.FrontendConnectionErrorScreen
 import io.homeassistant.companion.android.frontend.error.FrontendConnectionErrorStateProvider
+import io.homeassistant.companion.android.frontend.filechooser.FileChooserEffect
+import io.homeassistant.companion.android.frontend.filechooser.FileChooserRequest
 import io.homeassistant.companion.android.frontend.js.FrontendJsBridge
 import io.homeassistant.companion.android.frontend.js.FrontendJsCallback
 import io.homeassistant.companion.android.frontend.permissions.MultiplePermissionsEffect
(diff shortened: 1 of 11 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/frontend/filechooser/FileChooserEffect.kt
```diff
@@ -0,0 +1,39 @@
...
+            currentRequest?.onResult(uris)
+            currentRequest = null
+        },
+    )
+
+    if (pendingRequest != null) {
+        LaunchedEffect(pendingRequest) {
+            currentRequest = pendingRequest
+            launcher.launch(pendingRequest.fileChooserParams)
+        }
```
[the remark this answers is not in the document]
TimoPtr: If it throws today it crashes.

jpelgrom:
> I struggle between creating a new StateFlow and adding this to the ViewState. @jpelgrom any opinion here? Using the ViewState would avoid exposing yet another StateFlow but at the same time it is not really relevant to the state.

It isn't a state for the view as you pointed out, and fits in nicely with the other pending things from the webview triggering actions/effects outside the view 👍

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
```diff
@@ -193,6 +198,10 @@ internal fun FrontendScreenContent(
         pendingDialog = pendingDialog,
     )
 
+    FileChooserEffect(
```
jpelgrom, line 201: If this is an effect -> why isn't it in WebViewEffects?

If this isn't an effect -> why isn't it named like the others PendingFileChooserHandler or similar?

TimoPtr: It's an effect and I ddin't add it to WebViewEffects since it doesn't require a WebView.

jpelgrom: OK that makes sense, a 3rd type.

jpelgrom (approved):
Tested and confirmed, approach seems good also combined with the other PR stacked on top.

One step closer to completion...

PR #6757, opened 2026-04-23
Initialize Timber before Application.onCreate

TimoPtr:
## Summary

While looking at https://github.com/home-assistant/android/pull/6753 I found out that we were missing the logs that we might have during Hilt initialization happening in `super.onCreate` because we are planting the Tree after it. We should plant the Tree before anything else.


jpelgrom (approved):

jpelgrom:
Maybe also update `wear` to match? I've never seen this race before but if it happens with `app` it can also happen with `wear`.

TimoPtr:
> Maybe also update `wear` to match? I've never seen this race before but if it happens with `app` it can also happen with `wear`.

Indeed, thanks for always catching the things like that I tend to forget.

PR #6764, opened 2026-04-24
Configure Composer diagnostic stacktrace

TimoPtr:
## Summary

We were already enabling the compose stacktrace in debug but the recent version of Compose introduced a new way of setting it up that also allow getting more information without performance impacts on release (less precise).
Also `setDiagnosticStackTraceEnabled` has been deprecated

image

I decided to enable it in Release too with the `Groupkeys` mode and make common function for both app and wear.

image


jpelgrom (approved):

jpelgrom:
We should monitor performance with this change in production. Google says no impact/overhead but Compose performance on low-end devices can be bad so it should really have no impact, otherwise revert to `None`.

PR #6766, opened 2026-04-24
Disable microwakeword test on wearOS

TimoPtr:
## Summary

For now we don't have any reason to run the microwakeword instrumentation test on the wear and it seems that is it causing the CI to be flaky. Probably because it is just too much for the host to run the emulator and build the C++ at the same time (just a guess).


jpelgrom (approved):
Indeed we don't need this for Wear as is clear from the other Wear test not including it...

PR #6767, opened 2026-04-24
Fix resize cast for ExoPlayer in WebViewActivity causing blank player

TimoPtr:
## Summary

The `exoplayer/resize` message sent by the frontend is a Float not a Int because of that today it always return 0 and the resize doesn't nothing and is not shown on the screen. This PR make sure we cast the strings to Floats.

It is a bit hard to test, to make it easier I modified the `ha-hlp-player.ts` in the frontend to always use exo player.

I used Float and not Double because the whole Compose framework is working with Float precision no need to use Doubles.

## Checklist

## Screenshots

https://github.com/user-attachments/assets/2d91b023-4f5c-482c-b060-9ca450baa3c0


jpelgrom:
`exoplayer/resize` returns the value from [`getBoundingClientRect()`](https://github.com/home-assistant/frontend/blob/907687c16dc9efaac3bf927ecb88e94362104f07/src/components/ha-hls-player.ts#L252) which should be doubles. Floats will probably work but I'd use doubles to be safe.

[`getBoundingClientRect`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) -> [`DOMRect`](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect) -> [bottom](https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly/bottom), [left](https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly/left), [right](https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly/right), [top](https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly/top)

TimoPtr:
Your right I was lazy and didn't check the fact given by Claude when it told me no that there was no conversion from Double to Dp ... Which is false. I'll also adjust in the FrontendScreen PR where I initially got this issue. 

I'm going to keep the Float json helper function even if we don't use it since it's quite standard.

jpelgrom (approved):

PR #6768, opened 2026-04-24
Setup emulator WTF

TimoPtr:
## Summary

This PR integrates [emulator.wtf](https://emulator.wtf/) suing the Github Action to run instrumentation tests across all supported Android versions in CI. The token is injected using the Github environment `ui-test`.

Test execution is significantly faster: our 4 instrumentation tests now run across 14 emulators in under 30 seconds (excluding the build that is still on GHA).

As part of this change, Android phones emulators have been removed from GitHub Actions since they are now handled by emulator.wtf. Wear OS and Automotive emulators are still executed in CI to ensure the app launches correctly on those platforms.

Thanks to the emulator.wtf team for providing such a useful tool.

We will monitor this setup closely. If it introduces issues either on our side or theirs we can revert this PR and fall back to the previous configuration.

In case of failures, results will continue to be reported in PR comments as before, with the addition of a link to emulator.wtf containing the full logcat output.

I've added a Gradle flag `abis` to select the ABI we want to build to limit the build to specific ABIs.

## Checklist

Developer Documentation: home-assistant/developers.home-assistant#3072

## Any other notes

The next step would be to make a full 2e2 test by starting HA in GHA and then using the `egress-tunnel` of the Action to access HA from the emulator.


TimoPtr:
This is a comment added to the PR when the test on the emulator are failing, this disappear once we fix the issue.

image

It doesn't show minimal or full but a simple click on the link does.

.github/workflows/pr.yml
```diff
@@ -319,6 +319,85 @@ jobs:
         with:
           test-name: Unit test
 
+  instrumentation_test_wtf:
+    name: "Instrumentation Test WTF"
+    needs: [lint, lockfiles, ktlint]
+    environment: ui-test
+    runs-on: ubuntu-latest
+    permissions:
```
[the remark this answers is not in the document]
TimoPtr: It passes so I won't add it.

.github/workflows/pr.yml
```diff
@@ -319,6 +319,85 @@ jobs:
...
+        with:
+          cache-encryption-key: ${{ secrets.GRADLE_ENCRYPTION_KEY }}
+          mock-google-services: "true"
+
+      - name: Run test
+        # Tee stdout/stderr so a follow-up step can extract the emulator.wtf run URL.
+        # `set -o pipefail` preserves gradle's exit code through the pipe.
+        run: |
+          set -o pipefail
+          ./gradlew :app:testFullDebugWithEmulatorWtf 2>&1 | tee emulator-wtf.log
```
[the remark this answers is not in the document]
TimoPtr: I'm not sure how to handle this nicely to avoid spanning again 14 new devices. If we can reuse the existing it would be nice.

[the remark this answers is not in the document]
TimoPtr: See with folks at emulator.wtf they are ok for now to run both. If it becomes an issue we can fallback to only run one flavor.

jpelgrom:
Why did you decide to use a Gradle plugin + custom comment logic here? They offer a GitHub Action + standard junit test results we can upload using other well known actions, even mentioned in [their guide](https://docs.emulator.wtf/integrations/github-action/).

I'm not a big fan of using a Gradle plugin which then downloads their CLI, if we can also use a lightweight action which downloads their CLI but doesn't impact local dev environments.

TimoPtr:
> Why did you decide to use a Gradle plugin + custom comment logic here? They offer a GitHub Action + standard junit test results we can upload using other well known actions, even mentioned in [their guide](https://docs.emulator.wtf/integrations/github-action/).
> 
> I'm not a big fan of using a Gradle plugin which then downloads their CLI, if we can also use a lightweight action which downloads their CLI but doesn't impact local dev environments.

I used the plugin since it was easier for me to test on my setup. But indeed we can use the Action.

TimoPtr:
@jpelgrom I've migrated to the action so we don't have a new deps within Gradle locally. The only downside for now is that I'm not able to send a message with the URL within the PR but I think it is acceptable since a simple click on the task would show the link within the logs.

.github/workflows/pr.yml
```diff
@@ -319,6 +319,62 @@ jobs:
         with:
           test-name: Unit test
 
+  instrumentation_test_wtf:
+    name: "Instrumentation Test WTF"
```
jpelgrom: I'd argue we don't need partial brand names in here, and "WTF" isn't very descriptive for someone who doesn't know the website is https://emulator.wtf

```suggestion
  instrumentation_test_app:
    name: "Instrumentation Test app"
```

.github/workflows/pr.yml
```diff
@@ -319,6 +319,62 @@ jobs:
...
+          api-token: ${{ secrets.EMULATOR_WTF_TOKEN }}
+          devices: ${{ steps.devices.outputs.list }}
+          app: app/build/outputs/apk/minimal/debug/app-minimal-debug.apk
+          test: app/build/outputs/apk/androidTest/minimal/debug/app-minimal-debug-androidTest.apk
+          outputs-dir: build/test-results/minimal
+
+      - uses: ./.github/actions/upload-test-results
+        if: always()
+        with:
+          test-name: Instrumentation Test WTF
```
jpelgrom:
```suggestion
          test-name: Instrumentation Test app
```

.github/workflows/pr.yml
```diff
@@ -406,7 +452,7 @@ jobs:
 
   publish_test_results:
     name: "Publish Tests Results"
-    needs: [instrumentation_test, unit_tests, screenshot_test]
+    needs: [instrumentation_test, unit_tests, screenshot_test, instrumentation_test_wtf]
```
jpelgrom:
```suggestion
    needs: [instrumentation_test, unit_tests, screenshot_test, instrumentation_test_app]
```

.github/workflows/pr.yml
```diff
@@ -427,3 +473,4 @@ jobs:
           files: |
             **/androidTest-results/**/TEST-*.xml
             **/build/test-results/**/TEST-*.xml
+            **/build/build-results/**/results.xml
```
jpelgrom: This path is no longer relevant I think, you're saving in `build/test-results`?

jpelgrom (approved):
A lot cleaner and more consistent now <3

Minor comments to check for you but approach look good. To merge you'll need to change the required CI checks.

(Would be nice if their action also [pinned dependencies using the hash instead of tag](https://github.com/emulator-wtf/run-tests/blob/7abcd461ef5ce48b945667043987080c55aa130c/action.yml#L139). Do you have a direct line to pass the feedback or should we submit a PR?)

TimoPtr:
I've opened a PR https://github.com/emulator-wtf/run-tests/pull/19 on their repo for pinning the versions.

.github/workflows/pr.yml
TimoPtr: Indeed it need to be `**/build/test-results/**/results.xml` I'll keep it alongside the one above just to avoid a more complex regex or catching unwanted files.

TimoPtr:
```suggestion
            **/build/test-results/**/results.xml
```

.github/actions/upload-test-results/action.yml
```diff
@@ -14,6 +14,7 @@ runs:
         name: ${{ inputs.test-name }} results
         path: |
           **/build/test-results/**/TEST-*.xml
+          **/build/build-results/**
```
TimoPtr:
```suggestion
          **/build/test-results/**/results.xml
```

jpelgrom:
It looks like we forgot to consider PRs from forks... 🤦

Those don't have access to secrets so the action will fail by default. Do you know if it'll work if we check (there's no malicious intent/exfiltration of the secret) and retry?

PR #6785, opened 2026-04-27
Fix compose unit test and kotlin configuration after update

TimoPtr:
## Summary

After updating compose we started to see unit test failure. It is because of the new API made that switch the execution from `UnconfinedTestDispatcher` to `StandardTestDispatcher` https://developer.android.com/develop/ui/compose/testing/migrate-v2.

I've made a small helper `TestSharedFlow` because we can't use a `MutableSharedFlow.emit` anymore it requires a small buffer otherwise it hangs forever since we can't advance the consumer while producing the event, instead we use `tryEmit` but we check that the emission succeed.

## Checklist

## Any other notes

I've also update the Kotlin configuration that was missing after the bump of Kotlin.

### DIFF
--- a/app/src/test/kotlin/io/homeassistant/companion/android/onboarding/locationsharing/navigation/LocationSharingNavigationTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/onboarding/locationsharing/navigation/LocationSharingNavigationTest.kt
@@ -40,6 +40,7 @@ internal class LocationSharingNavigationTest : BaseOnboardingNavigationTest() {
             assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<LocationForSecureConnectionRoute>() == true)
 
             composeTestRule.activity.onBackPressedDispatcher.onBackPressed()
+
             waitForIdle()
 
             // In the test scenario since we never opened NameYourDevice the stack still contains Welcome
--- a/app/src/test/kotlin/io/homeassistant/companion/android/onboarding/locationsharing/navigation/LocationSharingNavigationTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/onboarding/locationsharing/navigation/LocationSharingNavigationTest.kt
@@ -40,6 +40,7 @@ internal class LocationSharingNavigationTest : BaseOnboardingNavigationTest() {
             assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<LocationForSecureConnectionRoute>() == true)
 
             composeTestRule.activity.onBackPressedDispatcher.onBackPressed()
+            waitForIdle()
 
             // In the test scenario since we never opened NameYourDevice the stack still contains Welcome
             assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<WelcomeRoute>() == true)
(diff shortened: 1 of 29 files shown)


app/src/test/kotlin/io/homeassistant/companion/android/onboarding/locationsharing/navigation/LocationSharingNavigationTest.kt
```diff
@@ -40,6 +40,7 @@ internal class LocationSharingNavigationTest : BaseOnboardingNavigationTest() {
             assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<LocationForSecureConnectionRoute>() == true)
 
             composeTestRule.activity.onBackPressedDispatcher.onBackPressed()
+            waitForIdle()
```
jpelgrom, line 44: Minor tweak for consistency, almost all other `waitForIdle()` after `onBackPressed()` have a newline
```suggestion

            waitForIdle()
```

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/nameyourdevice/navigation/NameYourDeviceNavigationTest.kt
```diff
@@ -83,6 +83,7 @@ internal class NameYourDeviceNavigationTest : BaseOnboardingNavigationTest() {
             assertTrue(navController.currentBackStackEntry?.destination?.hasRoute<LocalFirstRoute>() == true)
 
             composeTestRule.activity.onBackPressedDispatcher.onBackPressed()
+            waitForIdle()
```
jpelgrom, line 87:
```suggestion

            waitForIdle()
```

jpelgrom (approved):
> After updating compose we started to see unit test failure. It is because of the new API made that switch the execution from `UnconfinedTestDispatcher` to `StandardTestDispatcher`

v1 is only deprecated and didn't change behavior. Still, OK to migrate to the new one.

> I've made a small helper `TestSharedFlow` because we can't use a `MutableSharedFlow.emit` anymore it requires a small buffer otherwise it hangs forever since we can't advance the consumer while producing the event, instead we use `tryEmit` but we check that the emission succeed.

After replicating the issue this makes sense, but it is odd that Google didn't consider this. Might be good to provide feedback if you didn't already - I searched for an issue but couldn't find one quickly.

TimoPtr:
> > After updating compose we started to see unit test failure. It is because of the new API made that switch the execution from `UnconfinedTestDispatcher` to `StandardTestDispatcher`
> 
> v1 is only deprecated and didn't change behavior. Still, OK to migrate to the new one.

Well they've changed something in v1 ... I didn't have the courage to open an issue I simply migrated and adjust to the fact that we use StandardTesDispatcher now so we need to advance the time manually.

> 
> > I've made a small helper `TestSharedFlow` because we can't use a `MutableSharedFlow.emit` anymore it requires a small buffer otherwise it hangs forever since we can't advance the consumer while producing the event, instead we use `tryEmit` but we check that the emission succeed.
> 
> After replicating the issue this makes sense, but it is odd that Google didn't consider this. Might be good to provide feedback if you didn't already - I searched for an issue but couldn't find one quickly.

PR #6771, opened 2026-04-24
Add ExoPlayer support for FrontendScreen from frontend request

TimoPtr:
## Summary

This PR adds support for showing exoPlayer from a frontend command on the `FrontendScreen`. 
I've introduced a `ExoPlayerOverlay` to manage the player, once the player is initialized for the session we keep it so it can be reuse. I've also introduced a way to request full screen to the activity, it is also use to request leaving full screen but if the settings is already set to full screen then it stays on full screen.

## Checklist

- [x] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

## Screenshots

https://github.com/user-attachments/assets/c48a6fb6-6c48-4074-a13b-f318ff6b89a0

PR #6790, opened 2026-04-30
Add support for custom view in FrontendScreen

TimoPtr:
## Summary

Add support for custom view coming from the `WebView` into the `FrontendScreen`. It is built on top of the ExoPlayer support PR #6771.

## Checklist


app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewState.kt
```diff
@@ -56,6 +57,7 @@ sealed interface FrontendViewState {
         val statusBarColor: Color? = null,
         val backgroundColor: Color? = null,
         val exoPlayerState: ExoPlayerUiState? = null,
+        val customView: View? = null,
     ) : FrontendViewState
```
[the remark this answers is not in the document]
TimoPtr: @jpelgrom what's your take here, I don't like this too but I'm not sure about coming with a remember + a flag in the UiState either.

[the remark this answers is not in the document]
TimoPtr: After playing with Claude it suggested something that I like that is actually creating the Client from Compose by calling the viewModel so we can keep the view scoped to the UI only.

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebChromeClient.kt
```diff
@@ -46,4 +52,13 @@ class HAWebChromeClient(
         }
         return super.onShowFileChooser(webView, filePathCallback, fileChooserParams)
     }
+
+    override fun onShowCustomView(view: View?, callback: CustomViewCallback?) {
```
jpelgrom, line 56: We don't do anything with the callback in WebViewActivity either, but maybe we should?

> The application may explicitly exit fullscreen mode by invoking callback (ex. when the user presses the back button).

TimoPtr: Not sure we need to, what do you have in mind? I wonder about the leaving the activity but this might break PiP then. I propose to not touch this here to replicate the current behavior wdyt?

jpelgrom: The note about pressing the back button triggered me - does the back button currently not work? - but I haven't tested. OK to not touch it to replicate current behavior. Do you think we need an issue to check?

jpelgrom (approved):

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebChromeClient.kt
TimoPtr: Yes it does go back. I don't think we need an issue at the moment, until we find a use case that fails.

PR #6791, opened 2026-04-30
Fix WebViewActivity leak via Handler.postDelayed in waitForConnection (#5453)

### LINKED ISSUE #5453: Leak for Message(Queue) in WebViewActivity
**Home Assistant Android app version(s):** observed with commits from #5445, which is based on the recent main branch

**Android version(s):** 16

**Device model(s):** Google Pixel 7a

**Home Assistant version:** unrelated

**Last working Home Assistant release (if known):** first observed in #5419, so maybe before that (2025.6.5) but I'm not sure

I'm seeing this leak popping up more frequently now, but do not have a good way to reproduce it. It seems to happen most frequently when the WebViewActivity is recreated (for example, by changing the device theme).

**Companion App Logs:**

```
┬───
│ GC Root: System class
│
├─ android.app.ActivityThread class
│    Leaking: NO (MessageQueue↓ is not leaking and a class is never leaking)
│    ↓ static ActivityThread.sMainThreadHandler
├─ android.app.ActivityThread$H instance
│    Leaking: NO (MessageQueue↓ is not leaking)
│    ↓ Handler.mQueue
├─ [android.os.MessageQueue](http://android.os.messagequeue/) instance
│    Leaking: NO (MessageQueue#mQuitting is false)
│    HandlerThread: "main"
│    ↓ MessageQueue[0]
│                  ~~~
├─ [android.os.Message](http://android.os.message/) instance
│    Leaking: UNKNOWN
│    Retaining 134 B in 4 objects
│    Message.what = 0
│    Message.when = 93097284 (687 ms after heap dump)
│    Message.obj = null
│    Message.callback = instance @43948936 of io.homeassistant.companion.
│    android.webview.WebViewActivity$$ExternalSyntheticLambda23
│    Message.target = instance @43948904 of [android.os.Handler](http://android.os.handler/)
│    ↓ Message.callback
│              ~~~~~~~~
├─ io.homeassistant.companion.android.webview.
│  WebViewActivity$$ExternalSyntheticLambda23 instance
│    Leaking: UNKNOWN
│    Retaining 12 B in 1 objects
│    f$0 instance of io.homeassistant.companion.android.webview.WebViewActivity
│    with mDestroyed = true
│    ↓ WebViewActivity$$ExternalSyntheticLambda23.f$0
│                                                 ~~~
╰→ io.homeassistant.companion.android.webview.WebViewActivity instance
     Leaking: YES (ObjectWatcher was watching this because io.homeassistant.
     companion.android.webview.WebViewActivity received Activity#onDestroy()
     callback and Activity#mDestroyed is true)
     Retaining 103.2 kB in 2188 objects
     key = 8535a6ba-c005-4a2d-a428-362779c1c85d
     watchDurationMillis = 5613
     retainedDurationMillis = 608
     mApplication instance of io.homeassistant.companion.android.
     HomeAssistantApplication
     mBase instance of androidx.appcompat.view.ContextThemeWrapper

METADATA

Build.VERSION.SDK_INT: 36
Build.MANUFACTURER: Google
LeakCanary version: 2.14
App process name: io.homeassistant.companion.android.debug
Class count: 45043
Instance count: 325260
Primitive array count: 223337
Object array count: 44041
Thread count: 83
Heap total bytes: 42440944
Bitmap count: 4
Bitmap total bytes: 30568
Large bitmap count: 0
Large bitmap total bytes: 0
Db 1: open /data/user/0/io.homeassistant.companion.android.
debug/no_backup/androidx.work.workdb
Db 2: open /data/user/0/io.homeassistant.companion.android.
debug/databases/HomeAssistantDB
Count of retained yet cleared: 5 KeyedWeakReference instances
Stats: LruCache[maxSize=3000,hits=134992,misses=275288,hitRate=32%]
RandomAccess[bytes=13583713,reads=275288,travel=139399383506,range=50877810,size
=63850549]
Analysis duration: 22919 ms
```

**Screenshot or video of problem:** n/a

**Additional information:** n/a

jim-daf:
## Summary

Fixes #5453 — `WebViewActivity` was being retained in the main thread `MessageQueue` after `onDestroy`, matching the LeakCanary trace in the issue.

## Root cause

`WebViewActivity.waitForConnection()` posted a 10-second delayed `Runnable` directly on the main thread `Handler`:

```kotlin
Handler(Looper.getMainLooper()).postDelayed(
    {
        val firstSegment = loadedUrl?.pathSegments?.firstOrNull().orEmpty()
        if (
            !isConnected &&
            !firstSegment.contains("api") &&
            !firstSegment.contains("local")
        ) {
            showError(errorType = ErrorType.TIMEOUT_EXTERNAL_BUS)
        }
    },
    CONNECTION_DELAY,
)
```

The lambda captures `this` (the `WebViewActivity`). If the activity is destroyed before `CONNECTION_DELAY` (10 s) elapses — e.g. on the configuration change that motivated the report (theme switch recreates the activity) — the queued `Message` keeps a strong reference to the destroyed activity until the delay fires. This matches the leak trace exactly:

```
android.os.MessageQueue
  ↓ MessageQueue[0]
android.os.Message
  Message.callback = instance @… of WebViewActivity$$ExternalSyntheticLambda23
  ↓ Message.callback
WebViewActivity$$ExternalSyntheticLambda23
  f$0 instance of WebViewActivity (with mDestroyed = true)
  ↓ ExternalSyntheticLambda23.f$0
WebViewActivity (Leaking: YES)
```

`waitForConnection()` is called from `onResume`, `onDismissListener`s of error dialogs, and the timeout-external-bus "wait" button — all paths that can fire shortly before a recreation.

## Fix

Replace the raw `Handler.postDelayed` with `lifecycleScope.launch { delay(CONNECTION_DELAY); … }`. The activity-scoped coroutine is automatically cancelled when the `Lifecycle` reaches `DESTROYED`, so the captured `WebViewActivity` is released immediately on `onDestroy` instead of lingering in the main `MessageQueue`. Behaviour is otherwise unchanged: the timeout still runs on the main dispatcher and still calls `showError(ErrorType.TIMEOUT_EXTERNAL_BUS)` after the same 10 s delay when the connection has not been established.

The now-unused `android.os.Handler` and `android.os.Looper` imports are removed.

## Testing

- Verified that the only `Handler` / `Looper` references in `WebViewActivity` were the ones removed by this change (`HttpAuthHandler`, `SslErrorHandler`, `RenderProcessGoneDetail` `handler` parameters are unrelated identifiers).
- Manual reproduction (recreating the activity via a theme switch within the 10 s window) no longer reports the leak in LeakCanary.

## Notes for reviewers

This is a minimal, behaviour-preserving fix targeting only the leak in #5453. There are other `webView.post { … }` and `runOnUiThread { … }` call sites in this file, but those are dispatched immediately and don't sit in the queue across `onDestroy`, so they are out of scope here.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -1996,11 +1996,6 @@ class WebViewActivity :
             Timber.i("Fragments ${supportFragmentManager.fragments} displayed, skipping connection wait")
         } else {
             Timber.d("Waiting for loadedUrl ${sensitive(loadedUrl.toString())}")
-            // Use lifecycleScope so the delayed check is automatically cancelled when the
-            // activity is destroyed. Previously this used Handler(Looper.getMainLooper())
-            // .postDelayed { ... } with a lambda capturing `this`, which kept the destroyed
-            // WebViewActivity retained in the main MessageQueue for up to CONNECTION_DELAY ms
-            // (e.g. when the activity was recreated due to a theme change). See issue #5453.
             lifecycleScope.launch {
                 delay(CONNECTION_DELAY)
                 val firstSegment = loadedUrl?.pathSegments?.firstOrNull().orEmpty()
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
@@ -14,8 +14,6 @@ import android.net.http.SslError
 import android.os.Build
 import android.os.Bundle
 import android.os.Environment
-import android.os.Handler
-import android.os.Looper
 import android.text.method.HideReturnsTransformationMethod
 import android.text.method.PasswordTransformationMethod
 import android.util.DisplayMetrics
@@ -1998,19 +1996,22 @@ class WebViewActivity :
             Timber.i("Fragments ${supportFragmentManager.fragments} displayed, skipping connection wait")
         } else {
             Timber.d("Waiting for loadedUrl ${sensitive(loadedUrl.toString())}")
-            Handler(Looper.getMainLooper()).postDelayed(
-                {
-                    val firstSegment = loadedUrl?.pathSegments?.firstOrNull().orEmpty()
-                    if (
-                        !isConnected &&
-                        !firstSegment.contains("api") &&
-                        !firstSegment.contains("local")
-                    ) {
-                        showError(errorType = ErrorType.TIMEOUT_EXTERNAL_BUS)
-                    }
-                },
-                CONNECTION_DELAY,
-            )
+            // Use lifecycleScope so the delayed check is automatically cancelled when the
+            // activity is destroyed. Previously this used Handler(Looper.getMainLooper())
+            // .postDelayed { ... } with a lambda capturing `this`, which kept the destroyed
+            // WebViewActivity retained in the main MessageQueue for up to CONNECTION_DELAY ms
+            // (e.g. when the activity was recreated due to a theme change). See issue #5453.
+            lifecycleScope.launch {
+                delay(CONNECTION_DELAY)
+                val firstSegment = loadedUrl?.pathSegments?.firstOrNull().orEmpty()
+                if (
+                    !isConnected &&
+                    !firstSegment.contains("api") &&
+                    !firstSegment.contains("local")
+                ) {
+                    showError(errorType = ErrorType.TIMEOUT_EXTERNAL_BUS)
+                }
+            }
         }
     }
 


app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewActivity.kt
```diff
@@ -1998,19 +1996,22 @@ class WebViewActivity :
...
-                        showError(errorType = ErrorType.TIMEOUT_EXTERNAL_BUS)
-                    }
-                },
-                CONNECTION_DELAY,
-            )
+            // Use lifecycleScope so the delayed check is automatically cancelled when the
+            // activity is destroyed. Previously this used Handler(Looper.getMainLooper())
+            // .postDelayed { ... } with a lambda capturing `this`, which kept the destroyed
+            // WebViewActivity retained in the main MessageQueue for up to CONNECTION_DELAY ms
+            // (e.g. when the activity was recreated due to a theme change). See issue #5453.
```
TimoPtr:
```suggestion
```
We only describe the current behavior not the previous one. Using a coroutine is valid and doesn't need to be explained.

FYI this is going to be removed once we use the FrontendScreen and this is already handled properly but thanks for addressing the current issue.

TimoPtr (approved):

Issue #6055, opened 2025-11-16
Camera image upload to file input (frontend)


Mofef:
**Home Assistant Android app version(s):**
2025.8.7-full

**Android version(s):**
16  / 1st of July 2025

**Device model(s):**
Samsung Galaxy S25 Ultra a.k.a. SM-S938B/DS

**Home Assistant version:**
Installation method  Home Assistant Container
Core  2025.11.2
Frontend  20251105.0

**Last working Home Assistant release (if known):**
//

**Description of problem, include YAML if issue is related to notifications:**

I am working on a custom dashboard card for the [bparasite plant sensor](https://github.com/rbaron/b-parasite), which offers users to upload a photo of their plant in the config page of the card.

I could not make that work with a prompt to directly take a camera image *in the app* it did work when i accessed HA via the chrome browser on my android though. Since this feature seems to be known to be flaky on Android, I asked Gemini to create test-cases, which you can find here: https://mofef.github.io/android_camera_test.html
They **work as intended via android chrome but not if I integrate the exact same page as an iframe into my card.** It only ever shows a file picker, not offering the camera option.

The HA app does have camera permission in case that is relevant. 

Since I'm not familiar with android development i asked Gemini for a hypothesis:
Currently, the Home Assistant Android app's `WebView` implementation of `onShowFileChooser` appears to only trigger the standard file picker (`ACTION_GET_CONTENT`). On modern Android versions (14+), this often defaults to the new Photo Picker or a system file picker that **omits the Camera tile**.

This makes it impossible for a user to "Take a Photo" directly from an HTML `<input type="file">` element within the HA app. They are forced to background the app, open the Camera app, take a photo, return to HA, and select the file from the gallery.

Here is some more of what Gemini has to say about it, but I don't have the competence to review that sadly. Please ignore unless you are interested: **trigger warning: potential AI slop**
<details>
### Testing & Diagnosis
I have tested the following HTML5 input configurations inside the Home Assistant Android App (latest version) on Android 14/15.

| HTML Input Configuration | Result in HA App | Result in Chrome (Ref) |
| :--- | :--- | :--- |
| `<input type="file" accept="image/*">` | Opens Gallery (Photo Picker). No Camera option. | Gallery (Photo Picker). |
| `<input type="file" accept="image/*" capture="environment">` | Opens Gallery. No Camera option. | **Opens Camera directly.** |
[...]

PR #6794, opened 2026-04-30
fix(WebView): offer camera capture in <input type=file> chooser (refs #6055)

jim-daf:
### Problem

Issue [#6055](https://github.com/home-assistant/android/issues/6055) reports that an HTML
`<input type="file" accept="image/*" capture="environment">` element inside a
Lovelace card (the user's b-parasite plant-sensor card) works correctly in
Chrome for Android but inside the Companion app the chooser only ever shows
the system file picker — never a Camera tile. The user has to background
the app, open the Camera, take a photo, return to HA, and then pick the
file from the gallery.

### Root cause

[`ShowWebFileChooser.kt`](https://github.com/home-assistant/android/blob/main/app/src/main/kotlin/io/homeassistant/companion/android/webview/ShowWebFileChooser.kt)
has two issues that combine to break the camera path:

1. `createIntent` calls `input.createIntent()` (which builds an
   `ACTION_GET_CONTENT` carrying the page's `accept=` filter and `multiple`
   flag) and then overwrites `type = "*/*"`, throwing the filter away.
   On Android 14+ this is what makes the system fall back to the generic
   Photo Picker / Documents UI, which by design never exposes a Camera
   shortcut.
2. No `MediaStore.ACTION_IMAGE_CAPTURE` intent is ever added to the
   chooser, so even with the correct MIME filter the user could not pick
   the camera — Chromium's `FileChooserParams.createIntent()` does not
   add capture intents either; that is the embedder's job.

### Fix

* Use the framework-provided intent **verbatim** so the page's `accept=`
  filter and `multiple` flag survive.
* When the accept filter allows images and/or video, wrap the chooser in
  `Intent.ACTION_CHOOSER` with `EXTRA_INITIAL_INTENTS` containing
  `MediaStore.ACTION_IMAGE_CAPTURE` and/or `MediaStore.ACTION_VIDEO_CAPTURE`.
  This is the standard pattern Android documents for camera-aware file
  inputs in WebView embedders.
* Generate a temp file under `context.getExternalFilesDir(DIRECTORY_PICTURES)`
  for `EXTRA_OUTPUT` and share it via the existing
  `${'$'}{applicationId}.provider` FileProvider. This directory belongs to the
  app's scoped external storage so **no new runtime permission is required**
  (in particular, no `CAMERA` permission for the embedder — the system
  Camera app handles that on its own behalf).
* In `parseResult`, fall back to the recorded capture URI when the camera
  app returns with a `null` data Intent — that is the documented contract
  of `ACTION_IMAGE_CAPTURE` when `EXTRA_OUTPUT` is set.
* Reset the captured-URI member at the start of every `createIntent` call
  so a stale URI from a previous, cancelled chooser cannot leak into the
  next result.
* Add an `external-files-path` entry to `provider_paths.xml` so the
  FileProvider authority can actually hand the captured file back to the
  chooser caller.

### Files changed

* `app/src/main/kotlin/io/homeassistant/companion/android/webview/ShowWebFileChooser.kt`
* `app/src/main/res/xml/provider_paths.xml` (one new `<external-files-path>` line)

### Behaviour matrix

| Page input | Before | After |
|---|---|---|
| `<input type="file">` | system picker, all types | unchanged |
| `<input type="file" accept="image/*">` | Photo Picker only | Photo Picker **and Camera tile** |
| `<input type="file" accept="image/*" capture="environment">` | Photo Picker only | Photo Picker **and Camera tile** |
| `<input type="file" accept="video/*">` | system picker | system picker **and Camcorder tile** |
| `<input type="file" accept=".pdf">` | system picker for PDFs | unchanged (no capture intents added) |
| User cancels chooser | callback receives `null` | unchanged |
| User picks file | callback receives picked URIs | unchanged |
| User taps Camera tile | n/a | callback receives the captured photo URI |

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/ShowWebFileChooser.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/ShowWebFileChooser.kt
@@ -1,20 +1,119 @@
 package io.homeassistant.companion.android.webview
 
+import android.app.Activity
 import android.content.Context
 import android.content.Intent
 import android.net.Uri
+import android.os.Environment
+import android.provider.MediaStore
 import android.webkit.WebChromeClient
 import androidx.activity.result.contract.ActivityResultContract
+import androidx.core.content.FileProvider
+import java.io.File
+import timber.log.Timber
 
+/**
+ * Bridges Chromium's `<input type="file">` chooser request into an Android
+ * activity result.
+ *
+ * Issue [#6055](https://github.com/home-assistant/android/issues/6055):
+ * the previous implementation forced `type = "*/*"` on the framework-provided
+ * intent and did not expose any camera capture entry point, so a page using
+ * `<input type="file" accept="image/*" capture="environment">` would only
+ * ever see the system file picker (no Camera tile). On Android 14+ this
+ * defaulted to the Photo Picker, which intentionally never offers a Camera
+ * shortcut for capture.
+ *
+ * The new behaviour:
+ *
+ * * The intent built by [WebChromeClient.FileChooserParams.createIntent] is
+ *   used verbatim — it already encodes the page's `accept` filter and the
+ *   `multiple` flag, so the system can route to the Photo Picker /
+ *   Documents UI as appropriate.
+ * * When the page's accept filter allows images and/or video the chooser is
+ *   wrapped with `EXTRA_INITIAL_INTENTS` containing
+ *   [MediaStore.ACTION_IMAGE_CAPTURE] and/or
+ *   [MediaStore.ACTION_VIDEO_CAPTURE] so the user sees a Camera / Camcorder
+ *   tile alongside the file picker. The captured media is written to a
+ *   FileProvider URI under `getExternalFilesDir(DIRECTORY_PICTURES)` —
+ *   no extra runtime permissions are required (the directory belongs to
+ *   this app).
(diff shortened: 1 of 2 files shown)


Chreece:
I can reproduce #6055 on current Companion Android with another custom card (`<input type="file" accept="image/*" capture="environment">`): Chrome opens the camera, while Home Assistant Companion only opens the existing photo/file picker.

The relevant code has moved since this PR was opened. On current `main` the chooser now lives in `app/src/main/kotlin/io/homeassistant/companion/android/frontend/filechooser/FileChooserEffect.kt`, and the same underlying behavior is still present:

```kotlin
override fun createIntent(context: Context, input: WebChromeClient.FileChooserParams): Intent {
    return input.createIntent().apply {
        type = "*/*"
    }
}
```

So the issue is still current, but this PR needs rebasing onto the new file/Compose flow. A rebased version should preserve the page-provided accept types/multiple flag, honor image capture requests with an `ACTION_IMAGE_CAPTURE` intent, and include the capture URI in `clipData` as noted in the existing review comment. The app already has the `${applicationId}.provider` FileProvider; current `provider_paths.xml` also has an `external-path` mapping that covers app external files.

This is affecting a real HA dashboard camera workflow today, not only the original b-parasite case. I can test a rebased build against the current VeSync food-scanner card once available.

Issue #5650, opened 2025-08-15
Android Health Connect - Write


LoganJFisher:
**Is your feature request related to a problem? Please describe.**

Currently the Health Connect functionality of the Android companion app only supports reading. I would like to be able to write to Health Connect from Home Assistant as well. As my bathroom scale is bluetooth connected to Home Assistant utilizing the HACS BodyMiScale integration, this is the only way its data could automatically be exported to then be accessible by other services like Sparky Fitness.

**Describe the solution you'd like**

I would like if, instead of the simple "read" toggle already present, if each Health Connect function had a drop-down menu with options "disabled", "read", "write", and "read & write".

**Describe alternatives you've considered, if any**

Other than manually exporting this data, I'm not aware of any other alternatives.

**Additional context**

N/A

dshokouhi:
I am not sure if HA will qualify to write data to Health Connect as those devices that are connected should send its own data, all the manufacturers have their own apps and approval process. There is even a mention about not syncing with incompatible devices which can easily be seen as devices connected through HA that are not health connect approved. 

https://support.google.com/googleplay/android-developer/answer/12991134?hl=en

`Do not use health and fitness data APIs with apps that sync data between incompatible devices or platforms.`

anasmadrhar:
Hey @dshokouhi, @LoganJFisher. 
Building on this issue since it's been sitting unclaimed. i put together a working impl on a fork to see if writes were feasible without breaking the existing read path.

branch: https://github.com/anasmadrhar/hass-android/tree/feature/health-connect-sync ([PR #1 there](https://github.com/anasmadrhar/hass-android/pull/1))

what's in it:

- writes for all 25 currently-supported HC types via a new `command_health_connect_write` push command. payload is a flat map (`data_type`, `value` or `systolic`/`diastolic`, `time`/`start_time`/`end_time`, optional `client_record_id`, plus json arrays for heart rate samples / sleep stages). dispatched from `MessagingManager` to a new `HealthConnectWriteCommandHandler` -> `HealthConnectWriteRepository` (one typed `writeX` per data type).
- HC -> HA real-time reads via the changes API, opt-in. periodic worker (~15min repeat / 5min flex) per data type, persisted token in SharedPreferences, calls `SensorReceiver.updateAllSensors` when upsertions or deletions show up. handles `changesTokenExpired` by re-minting and forcing a refresh. paginates `hasMore`.
[...]

PR #6799, opened 2026-05-02
Add Health Connect write commands and Changes API real-time sync

anasmadrhar:
## Summary

builds on #5650. wires up two-way sync between Home Assistant and Health Connect so HA can write back into HC, plus picks up changes from other apps faster than the existing 15-min poll.

what's in it:

1. `command_health_connect_write` push command that writes a record into HC, dispatched via the existing MessagingManager. covers all 26 existing data types plus 4 new ones (ExerciseSession + Speed / Power / CyclingPedalingCadence). payload is a flat FCM data map (data_type, value or systolic+diastolic, optional unit, ISO-8601 time / start_time+end_time, optional client_record_id), plus JSON arrays for series records (heart rate samples, sleep stages, speed/power/cadence).
2. HC to HA real-time read path via the Changes API. opt-in toggle on the new Health Connect settings screen, periodic worker per data type, drains hasMore, handles changesTokenExpired by re-minting, triggers a sensor refresh on upsertions/deletions. catches writes from Samsung Health / Google Fit / smart-scale apps within a few minutes instead of waiting for the next 15-min cycle.
3. per-sensor "Allow writes from HA" toggle on each HC sensor's detail screen. enabling it adds the matching WRITE_* to the sensor's required permissions and reuses the existing perm-grant flow.
4. new Settings -> Sensors -> Health Connect screen with the real-time sync toggle and a bulk-enable row that flips on every sensor, persists the write toggles, fires a single combined permission request. live "X / Y sensors enabled" status driven by a SensorDao flow.
4. new `health_connect_exercise_session` sensor surfaces the latest HC workout with rich attributes computed over the session window: avg/min/max HR, distance, derived avg pace, recorded speed, power, cadence, calories, steps, segments and laps (segment reps double as swim stroke counts when the source app populates them).
5. unit conversion delegates to HC's typed unit classes, no hardcoded factors on the app side. accepts kg/g/lb/oz, m/cm/km/mi/ft/in, C/F, mmol/L vs mg/dL, L/mL/fl_oz, kcal/cal/J/kJ, W vs kcal/day, m/s vs km/h vs mph.
6. ISO-8601 timestamps accept both ...Z and ...+HH:MM offset (HA's now().isoformat() defaults to the latter), blank time field falls back to now.

motivation: #5650 has been sitting unclaimed for a while. the missing piece is reliably pushing values from HA into HC for hardware that doesn't speak HC natively (bluetooth scales without a Play Store app, dumb treadmills with vibration sensors, off-brand pulse oximeters that only expose MQTT, anything ESPHome). also the inverse: faster pickup of writes from other apps so dashboards reflect a watch run before the 15-min cycle catches up.

ran on my own setup before opening this. weight readings from a smart scale push into HC via the new command end-to-end including idempotency by client_record_id. simulate-run script wraps Distance + Steps + ActiveCalories + HeartRate + Speed + Power + CyclingPedalingCadence inside an ExerciseSession and HC shows it as one workout entry with full stats.

a heads-up on Play declaration: the new WRITE_* permissions (and three new READ_* ones for Speed / Power / CyclingPedalingCadence) need a refreshed Health Apps declaration plus a Data Safety update. that's an org-owner step that a contributor PR can't ship on its own. happy to scope down to a smaller initial subset if that's easier to defend, or split into waves: weight + hydration + blood pressure + body temperature first, then activity types, then everything else. let me know which works.

honest gaps: the enriched session sensor's aggregation reads aren't unit-tested (would need ~50 lines of mocks for aggregate(...) against six metric sets, all thin wrappers over the SDK). validated end-to-end on device. HealthConnectChangesWorker itself is glue calling the (tested) repository, no Robolectric test. no screenshot tests for the new settings UI. personal-best sensors (longest run, fastest 5K) are flagged as out-of-scope here, HC has no PB API so it's an iterate-history-and-track problem worth its own change.

a side note on the proxy: while running this on my own deployment of the open-source FCM proxy I hit two pre-existing bugs unrelated to HC that prevented HC payloads (and any object/array data values) from getting through. fixes are up at [home-assistant/mobile-apps-fcm-push#294](https://github.com/home-assistant/mobile-apps-fcm-push/pull/294) — independent of this PR but worth landing alongside if HC payloads are going to be exercised in production. the matching whitelist update for the new `command_health_connect_write` payload keys + ignore-list entry is at [home-assistant/mobile-apps-fcm-push#295](https://github.com/home-assistant/mobile-apps-fcm-push/pull/295) (depends on this PR for the wire format; primitives work standalone, the array-typed fields need #294 too).
## Checklist

## Screenshots

Screenshot_20260502_173042_Home Assistant
Screenshot_20260502_173045_Home Assistant
Screenshot_20260502_173049_Home Assistant
Screenshot_20260502_173111_Home Assistant
Screenshot_20260502_173105_Home Assistant

## Link to pull 

request in documentation repositories

User Documentation: home-assistant/companion.home-assistant#1321 (draft, will mark ready once this PR is on track to merge)

## Any other notes

closes (or unblocks, depending on Play declaration outcome) [#5650](https://github.com/anasmadrhar/hass-android/issues/5650).


anasmadrhar:
addressed the review feedback in 50c8863f:

- **MessagingManager.kt** — fallback notification now rewrites `title`/`message` with the actual failure reason instead of reposting `command_health_connect_write` as the body. permission errors point at the settings screen, validation errors carry the exact rejection reason, internal failures carry the exception message.
- **HealthConnectUnitConversion.kt** — added `cm` / `centimeter(s)` / `centimetre(s)` for length, plus a regression test in `HealthConnectUnitConversionTest`.
- **HealthConnectWriteCommandHandler.kt** — `RestingHeartRate` and `Steps` now reject fractional input via a shared `requireIntegral` helper instead of silently truncating with `toLong()`. payloads like `1234.9` come back as `InvalidPayload` with a message naming the field. regression tests added for both.
- **HealthConnectModule.kt** — dropped `@Singleton` from the provider so each `provider.get()` re-queries SDK status. `HealthConnectClient.getOrCreate` is itself idempotent so this is cheap, and a user who installs/updates Health Connect mid-session no longer gets stuck seeing `null` until the next cold start.

TimoPtr:
Adding this will cause the battle with Google to get approved. For now we are not ready to have this fight closing.

PR #6801, opened 2026-05-03
Stream active notification content attributes in recorder-safe buckets

Komzpa:
## Summary

- Keep Active notification count content attributes enabled by default.
- Split active notification contents into recorder-safe buckets when the payload would be too large for Home Assistant recorder.
- Preserve existing installs by removing the migration that disabled the content-attributes setting.

## Why

When one device has many active notifications, sending every notification as attributes on a single state update can exceed Home Assistant recorder's state-attribute size limit. In that case recorder warns and does not store the oversized attributes.

This keeps the numeric notification count stable while sending the content attributes as multiple forced updates. Each update has the same total count and a small part marker, so recorder can store the full snapshot across history rows. Buckets are sent oldest first; the last/current bucket contains the newest notifications.

Bucket metadata uses short keys:

- `total`: total active notifications
- `part`: current bucket number, starting at 1
- `parts`: total bucket count
- `last`: whether this is the final bucket
- `snapshot`: shared id for all buckets from one update
- `cut` / `cut_count`: present only if oversized content had to be trimmed

## Testing

- `git diff --check`
- `./gradlew :app:ktlintCheck :common:ktlintCheck`
- `./gradlew :common:compileDebugAndroidTestKotlin`
- `./gradlew :app:testFullDebugUnitTest --tests 'io.homeassistant.companion.android.sensors.NotificationSensorManagerTest'`


app/src/main/kotlin/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
```diff
@@ -257,7 +257,7 @@ class NotificationSensorManager :
                         applicationContext,
                         activeNotificationCount,
                         SETTING_INCLUDE_CONTENTS_AS_ATTRS,
-                        default = true,
```
TimoPtr, line 260: That's a breaking change, @jpelgrom how do you handle such thing in the past? Users relying on this at the moment will have to go in the settings and enable it back. It might break some automations.

common/src/main/kotlin/io/homeassistant/companion/android/database/migration/DatabaseMigration.kt
```diff
@@ -860,3 +861,31 @@ private class Migration40to41(private val context: Context) : Migration(40, 41)
...
+private val MIGRATION_51_52 = object : Migration(51, 52) {
+    override fun migrate(db: SupportSQLiteDatabase) {
+        db.execSQL(
+            """
+            UPDATE `sensor_settings`
+            SET `value` = 'false'
+            WHERE `sensor_id` = 'active_notification_count'
+                AND `name` = 'active_notification_count_content_attrs'
+                AND `value` = 'true'
+            """.trimIndent(),
```
TimoPtr: We could potentially maybe keep to true existing installation and false any new installation so new users would have to enable the flag where current installation that could rely on it keep their old behavior.

app/src/main/kotlin/io/homeassistant/companion/android/sensors/NotificationSensorManager.kt
Komzpa: Adjusted this so it no longer disables the setting for existing users or changes the default to off.

The current version keeps `active_notification_count_content_attrs` enabled by default and removes the 51 → 52 migration entirely. Instead, when the notification content payload is too large for recorder, the app sends the same count across multiple forced bucket updates with short `part` / `parts` / `snapshot` metadata. That keeps existing automations from losing the content attributes while keeping each recorded state payload below the recorder size limit.

jpelgrom: > That's a breaking change, @jpelgrom how do you handle such thing in the past? Users relying on this at the moment will have to go in the settings and enable it back. It might break some automations.

Generally we don't touch settings for sensors once created. When changing it in code it won't change it in the database by default, unless you add a migration as was done here before the entire overhaul.

jpelgrom:
Personally speaking, this bucket behavior is very convoluted and once again highlights how this shouldn't be accommodated in a sensor. It also duplicates sending logic in the actual sensor which is unwanted.

PR #6806, opened 2026-05-04
Support for auto play video in FrontendScreen

TimoPtr:
## Summary

This Pr add support for auto play in `FrontendScreen`. It exposes the settings as a flow too to react without having to check the value again when the screen is already loaded. 

I've used `reload` instead of re-creating the activity.

## Checklist

## Any other notes

I took the opportunity to mutualize a bit the logic of observing changes in the pref repo.


common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
```diff
@@ -411,4 +403,12 @@ internal class PrefsRepositoryImpl @Inject constructor(
...
         localStorage().putString(PREF_SELECTED_WAKE_WORD, wakeWord)
     }
+
+    private suspend fun <T> observeChanges(vararg keys: String, mapper: suspend () -> T): Flow<T> {
+        val localStorage = localStorage()
+        // Seed an initial emission so collectors read the current value immediately
+        return (keys.map { localStorage.observeChanges(it) } + flowOf(""))
+            .merge()
+            .map { mapper() }
+    }
```
jpelgrom: Why is this specific to this repository and not part of the `LocalStorage` interface?

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -248,6 +250,19 @@ internal class FrontendViewModel @VisibleForTesting constructor(
...
     private var zoomObserverJob: Job? = null
 
+    /**
+     * The user's "Autoplay video" preference.
+     *
+     * Lives outside [FrontendViewState] because autoplay is orthogonal to the screen state
+     * machine the WebView is rendered during `Loading`, `Content`, and `Error`, and all three
+     * need the value. Exposed as a [StateFlow] so the screen can read the current value
+     * synchronously when configuring the WebView at creation time (avoiding a one-shot reload
+     * once the persisted value lands) and react to subsequent changes via collection.
```
jpelgrom: I have no idea what orthogonal means here. You can explain it but better to avoid it for future readers who might also be confused.

```suggestion
     * Lives outside [FrontendViewState] because the WebView is rendered during `Loading`,
     * `Content`, and `Error`states , and all three states need the value. Exposed as a [StateFlow]
     * so the screen can read the current value synchronously when configuring the WebView at
     * creation time (avoiding a one-shot reload once the persisted value lands) and react to
     * subsequent changes via collection.
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
TimoPtr: You're right it make more sense.

TimoPtr: Actually I've moved only the vararg part I'm not sure to want the merge and mapper within the localStorage.

jpelgrom: Why not have merge/mapper within the localStorage? I can see this pattern being useful in other repositories and at that point I don't want to duplicate it.

TimoPtr: I didn't like having the mapper logic in the base class it felt too specific but I'll move it and see how it behaves in the future usage 👍🏻

jpelgrom (approved):

PR #6812, opened 2026-05-05
Add support for PiP for FrontendScreen

TimoPtr:
## Summary

Based on #6790 to add support for PiP in `FrontendScreen` using the LauncherActivity.

## Checklist


automotive/lint-baseline.xml
jpelgrom, line 1: No changes in the app lint baseline file? The manifest is also changed in app...

jpelgrom (approved):
One very minor thing to check, all good for everything else! I don't have a lot of experience with PIP personally so don't guarantee nothing has been missed.

There is a Jetpack Core-PiP library in alpha for Compose but fine to use the 'normal' APIs to get more control here.

TimoPtr:
> There is a Jetpack Core-PiP library in alpha for Compose but fine to use the 'normal' APIs to get more control here.

The 'normal' API was simple enough, there was also some other API to auto enable PiP but since we need it for the previous version there is simply no point to use it.

automotive/lint-baseline.xml
TimoPtr: I only updated the baseline because it was failing on the API check within the LaunchActivity, let's not update the baseline when it's passing for now.

PR #6813, opened 2026-05-06
Move the debug flag for WebView into the Application and Settings

TimoPtr:
## Summary

Move the set of the debug capability of the WebView at the Application level and also in the settings level. The idea is to enable/disable it globally so that it apply right away when changing and also when the app starts.

There is a risk where maybe the debug flag is set after the webview load since it uses the ioScope in the Application but I'm not sure.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/HomeAssistantApplication.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/HomeAssistantApplication.kt
@@ -100,12 +100,12 @@ open class HomeAssistantApplication :
         registerActivityLifecycleCallbacks(LifecycleHandler)
 
         ioScope.launch {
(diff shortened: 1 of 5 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/HomeAssistantApplication.kt
```diff
@@ -99,6 +100,7 @@ open class HomeAssistantApplication :
         registerActivityLifecycleCallbacks(LifecycleHandler)
 
         ioScope.launch {
+            WebView.setWebContentsDebuggingEnabled(BuildConfig.DEBUG || prefsRepository.isWebViewDebugEnabled())
```
jpelgrom: You may want to consider moving this down in this block so if it crashes we actually have crash reporting enabled

jpelgrom (approved):
Verified. I don't see a new race happening here as this should result in it being set earlier in the lifecycle, but good to stay alert to reports of it not working.

PR #6814, opened 2026-05-07
Migrate TagReaderActivity to use BottomSheet and add confirmation buttons

TimoPtr:
## Summary

This PR is replacing the full screen processing of a tag, with a bottom sheet with confirmation (Allow Once and Allow Always). It moves it to its own taskAfinity and made transparent so it can be shown on top of anything. 

The design is inspired from the Android 17 biometrics (for the close button). 

I've added in the settings a way to clear the allowed tags. 

I've added support for `next.home-assistant.io` links in debug which is useful while working with a local dev frontend. 

The previous implementation was blinking when scanning since it can go quite fast, I decided to show the bottom sheet for at least 1.5s to make it less blinky and still fast.

## Checklist

## Screenshots

image
image


app/src/main/kotlin/io/homeassistant/companion/android/nfc/TagReaderActivity.kt
```diff
@@ -1,87 +1,48 @@
...
-                    }
-                }
-            }.awaitAll()
+        if (isNfcTag || isQrTag) {
+            val url = if (isNfcTag) NFCUtil.extractUrlFromNFCIntent(intent) else intent.data
+            viewModel.onIntentReceived(url = url, isNfcTag = isNfcTag)
         } else {
-            showProcessingError(isNfcTag)
+            finish()
         }
```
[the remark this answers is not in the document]
TimoPtr: I'm not sure we need that, I didn't have any issue during my tests testing multiples.

jpelgrom:
> The previous implementation was blinking when scanning since it can go quite fast, I decided to show the bottom sheet for at least 1.5s to make it less blinky and still fast.

I feel like this was intentional, at least when scanning NFC tags where you already have feedback that something was scanned. Maybe out of scope, but did you consider changing it to show a checkmark confirmation for a short while (like 1s) after the network request completed instead? Then the dialog offers real feedback.

app/src/debug/AndroidManifest.xml
```diff
@@ -28,6 +28,21 @@
             android:theme="@android:style/Theme.Material.Light.NoActionBar"
             android:name="io.homeassistant.companion.android.HiltComponentActivity"
             />
+
+        <activity
```
jpelgrom, line 35:
```suggestion
        <!--
        Support for next.home-assistant.io links when developing (merged with non-next links in main manifest).
        -->
        <activity
```

app/src/main/kotlin/io/homeassistant/companion/android/nfc/views/TagReaderScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+private fun TagApprovalSheetHeader(onClose: () -> Unit, modifier: Modifier = Modifier) {
+    Box(
+        modifier = modifier
+            .fillMaxWidth()
+            .padding(bottom = HADimens.SPACE2),
+    ) {
+        Column(modifier = Modifier.align(Alignment.TopCenter)) {
+            Image(
+                painter = adaptiveIconPainterResource(R.mipmap.ic_launcher_round),
+                contentDescription = stringResource(commonR.string.home_assistant_branding_icon_content_description),
```
jpelgrom: As the app name is repeated below isn't this just for decoration and can we remove the description?

app/src/main/AndroidManifest.xml
```diff
@@ -469,7 +469,12 @@
         <activity
             android:name=".nfc.TagReaderActivity"
             android:label="@string/tag_reader_title"
-            android:exported="true">
+            android:launchMode="singleTask"
+            android:taskAffinity="io.homeassistant.companion.android.tag.reader"
+            android:autoRemoveFromRecents="true"
+            android:showWhenLocked="true"
```
jpelgrom: This might be nice for some use cases but also removes the security of having to unlock the device to actually scan a tag. Now anyone with your device can trigger tag automations.

automotive/src/debug/AndroidManifest.xml
```diff
@@ -0,0 +1,21 @@
+<manifest xmlns:tools="http://schemas.android.com/tools"
+    xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application tools:ignore="MissingApplicationIcon">
+        <activity
```
jpelgrom, line 8:
```suggestion
        <!--
        Support for next.home-assistant.io links when developing (merged with non-next links in main manifest).
        -->
        <activity
```

automotive/src/main/AndroidManifest.xml
```diff
@@ -371,7 +371,12 @@
         <activity
             android:name=".nfc.TagReaderActivity"
             android:label="@string/tag_reader_title"
-            android:exported="true">
+            android:launchMode="singleTask"
+            android:taskAffinity="io.homeassistant.companion.android.tag.reader"
+            android:autoRemoveFromRecents="true"
+            android:showWhenLocked="true"
```
jpelgrom: See https://github.com/home-assistant/android/pull/6814/changes#r3203981076 + are cars ever locked?

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/util/PainterResourceUtil.kt
```diff
@@ -0,0 +1,48 @@
...
+ * wraps it in a [BitmapPainter]. For every other resource type and on pre-O devices where
+ * adaptive icons don't exist it delegates to [painterResource] unchanged.
+ *
+ * Typical use: rendering the app launcher icon (e.g. `R.mipmap.ic_launcher_round`) inside Compose.
+ *
+ * Note: callers that want a circular icon (matching launcher behaviour on API 26+) should clip
+ * the resulting [androidx.compose.foundation.Image] with `Modifier.clip(CircleShape)` this
+ * helper returns the unmasked bitmap.
+ *
+ * Got from https://gist.github.com/tkuenneth/ddf598663f041dc79960cda503d14448.
```
jpelgrom: There appear to be some changes

```suggestion
 * Based on https://gist.github.com/tkuenneth/ddf598663f041dc79960cda503d14448.
```

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/util/PainterResourceUtil.kt
```diff
@@ -0,0 +1,48 @@
...
+ * rasterized assets (PNG / JPG / WebP); it throws on adaptive icons (`<adaptive-icon>` XML in
+ * `mipmap-anydpi-v26`), which is the format launcher icons use on API 26+. This helper detects
+ * that case and rasterises the adaptive icon to a [android.graphics.Bitmap] via [toBitmap], then
+ * wraps it in a [BitmapPainter]. For every other resource type and on pre-O devices where
+ * adaptive icons don't exist it delegates to [painterResource] unchanged.
+ *
+ * Typical use: rendering the app launcher icon (e.g. `R.mipmap.ic_launcher_round`) inside Compose.
+ *
+ * Note: callers that want a circular icon (matching launcher behaviour on API 26+) should clip
+ * the resulting [androidx.compose.foundation.Image] with `Modifier.clip(CircleShape)` this
```
jpelgrom:
```suggestion
 * the resulting [androidx.compose.foundation.Image] with `Modifier.clip(CircleShape)`; this
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
```diff
@@ -160,4 +160,10 @@ interface PrefsRepository {
     suspend fun getSelectedWakeWord(): String?
 
     suspend fun setSelectedWakeWord(wakeWord: String)
+
+    suspend fun addAllowedTag(tag: String)
+
+    suspend fun allowedTags(): Set<String>
```
jpelgrom:
```suggestion
    suspend fun getAllowedTags(): Set<String>
```

Prefer consistency, using `()` already makes it not look like Kotlin property access syntax

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
```diff
@@ -47,6 +47,7 @@ private const val PREF_CHANGE_LOG_POPUP_ENABLED = "change_log_popup_enabled"
 private const val PREF_SHOW_PRIVACY_HINT = "show_privacy_hint"
 private const val PREF_WAKE_WORD_ENABLED = "wake_word_enabled"
 private const val PREF_SELECTED_WAKE_WORD = "selected_wake_word"
+private const val PREF_ALLOWED_TAGS = "approved_tags"
```
jpelgrom: Why not match the key and variable?


```suggestion
private const val PREF_ALLOWED_TAGS = "allowed_tags"
```

common/src/main/res/values/strings.xml
```diff
@@ -490,6 +492,10 @@
     <string name="nfc_trigger_device">Scanned by this device:</string>
     <string name="nfc_invalid_tag">This tag does not contain Home Assistant data</string>
     <string name="nfc_processing_tag_error">Error while processing NFC tag</string>
+    <string name="tag_approval_title">Allow sending?</string>
+    <string name="tag_approval_description">Sending the tag %1$s to perform automations attach to it.</string>
+    <string name="tag_allow_once">Allow once</string>
+    <string name="tag_allow_always">Allow always</string>
```
jpelgrom: Are the strings already decided?
 - "Allow sending" sounds weird to me, from the user perspective you're scanning a tag and the fact that scanning happens outside the app and the app only sends data is an implementation detail. (+ description should be adjusted if the title is adjusted)
 - Maybe surround the tag ID with quotes to make it more obvious as user data like: `the tag 'abcdefgh' to perform`
 - "to perform automations attach to it" .should be _attached_, but to better match the HA [terminology](https://www.home-assistant.io/docs/automation/basics/) I'd use something like "to trigger automations linked to it".

app/src/main/kotlin/io/homeassistant/companion/android/nfc/views/TagReaderScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+            Box(
+                modifier = Modifier
+                    .size(CLOSE_BUTTON_VISIBLE_SIZE)
+                    .clip(CircleShape)
+                    .background(LocalHAColorScheme.current.colorSurfaceLow),
+                contentAlignment = Alignment.Center,
+            ) {
+                Icon(
+                    imageVector = Icons.Default.Close,
+                    contentDescription = stringResource(commonR.string.cancel),
```
jpelgrom: This should be set on the outer Box clickable modifier

app/src/main/kotlin/io/homeassistant/companion/android/nfc/views/TagReaderScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+                .padding(HADimens.SPACE2)
+                .size(CLOSE_BUTTON_TAP_SIZE)
+                .clickable(
+                    interactionSource = remember { MutableInteractionSource() },
+                    indication = null,
+                    onClick = onClose,
+                ),
+            contentAlignment = Alignment.Center,
+        ) {
+            Box(
```
jpelgrom, line 215: Why is there a nested Box structure here? Can't you make the inner one clickable? Also for the touch ripple to properly be a circle.

app/src/main/res/xml/preferences_developer.xml
```diff
@@ -20,8 +20,12 @@
         android:icon="@drawable/ic_thread"
         android:title="@string/thread_debug"
         android:summary="@string/thread_debug_summary"/>
+    <Preference
+        android:key="tag_clear_allowed"
+        android:icon="@drawable/ic_delete"
```
jpelgrom: I think [`mdi:tag-remove-outline`](https://pictogrammers.com/library/mdi/icon/tag-remove-outline/) would be a good candidate here.

app/src/main/kotlin/io/homeassistant/companion/android/nfc/views/TagReaderScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+        HAFilledButton(
+            text = stringResource(commonR.string.tag_allow_always),
+            onClick = onAllowAlways,
+            modifier = Modifier.fillMaxWidth(),
+        )
+        HAPlainButton(
+            text = stringResource(commonR.string.tag_allow_once),
+            onClick = onAllowOnce,
+            modifier = Modifier.fillMaxWidth(),
+        )
```
jpelgrom, line 265: I don't think there are strong Material guidelines on once vs always buttons, so I'd align the order + which button is primary with iOS. Right now they're different orders and different primary buttons - I think iOS with once as primary + first option makes more sense, or both should receive equal importance.

common/src/main/res/values/strings.xml
jpelgrom: cc @bgoncal

bgoncal: @TimoPtr Do we have an agreement on the copy then? I am fine improving it a bit like @jpelgrom said.
Sharing again how iOS looks like, in iOS we don't have any UI out of the main app (I understood that android has right?)
588943071-b7411e42-c2ec-473b-8241-c3280531737f

Then in iOS `surround the tag ID with quotes` does not make sense anymore, since it is separately from the main text now.

`"to trigger automations linked to it".` I agree with that


How about the title? What makes more sense in your opinion? (To replace the "Allow sending")

TimoPtr:
> > The previous implementation was blinking when scanning since it can go quite fast, I decided to show the bottom sheet for at least 1.5s to make it less blinky and still fast.
> 
> I feel like this was intentional, at least when scanning NFC tags where you already have feedback that something was scanned. Maybe out of scope, but did you consider changing it to show a checkmark confirmation for a short while (like 1s) after the network request completed instead? Then the dialog offers real feedback.

I was indeed considering this, I don't know if it's really improving the UI and we would also have to properly handle the error. We also need to support multi-server, so what do show when one server fail and the other one succeed. It is something that I think we already lack today.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepositoryImpl.kt
TimoPtr: Leftover from my first impl

app/src/main/AndroidManifest.xml
TimoPtr: You're right let's remove this and see later if it is a real feature. To be honest I copy past it from Assist thinking it would be a nice addition.

app/src/main/kotlin/io/homeassistant/companion/android/nfc/views/TagReaderScreen.kt
TimoPtr: I replicated the behavior from Android 17 where there are no ripple effect on the close button of the biometric bottom sheet, to do that I had to play with these nested boxes. There might be other ways of doing it most probably.

TimoPtr: The 2 box is to be able to keep a clickable area that is acceptable while keeping the visual nice.

app/src/main/kotlin/io/homeassistant/companion/android/nfc/views/TagReaderScreen.kt
TimoPtr: Aligned

app/src/main/res/xml/preferences_developer.xml
```diff
@@ -20,8 +20,12 @@
...
+        android:title="@string/clear_allowed_tags"/>
     <Preference
         android:key="webview_clear_cache"
-        android:icon="@drawable/ic_clear"
+        android:icon="@drawable/ic_tag_remove"
```
jpelgrom: You changed the icon on the wrong preference, this should be for the new one above

app/src/main/kotlin/io/homeassistant/companion/android/nfc/views/TagReaderScreen.kt
```diff
@@ -0,0 +1,354 @@
...
+                text = stringResource(commonR.string.tag_id),
+                style = HATextStyle.BodyMedium.copy(textAlign = TextAlign.Start),
+                modifier = Modifier.padding(bottom = HADimens.SPACE1),
+            )
+            Text(
+                text = tagId,
+                style = HATextStyle.Body.copy(
+                    textAlign = TextAlign.Start,
+                    color = LocalHAColorScheme.current.colorTextPrimary,
+                ),
```
jpelgrom, line 290: If content is too long this is truncated without indication. I'd use ellipsis. (Don't apply suggestion, it also needs an import.)

For example if font is bigger / screen is smaller:

Image

```suggestion
                ),
                overflow = TextOverflow.Ellipsis,
```

app/src/main/kotlin/io/homeassistant/companion/android/nfc/views/TagReaderScreen.kt
```diff
@@ -0,0 +1,354 @@
...
+                maxLines = 1,
+            )
+        }
+        IconButton(
```
jpelgrom, line 295: The padding needs some adjustment on this button and/or the container: there is a lot more on the right than on the left and it looks off-balance. Compare with the system share sheet for example:

Image

Image

(Not sure why the icon is flipped around, but you used the one from Material so let's not worry about that)

common/src/main/res/values/strings.xml
```diff
@@ -490,6 +492,12 @@
     <string name="nfc_trigger_device">Scanned by this device:</string>
     <string name="nfc_invalid_tag">This tag does not contain Home Assistant data</string>
     <string name="nfc_processing_tag_error">Error while processing NFC tag</string>
+    <string name="tag_approval_title">Allow sending?</string>
+    <string name="tag_approval_description">Sending this tag to trigger automations linked to it.</string>
+    <string name="tag_allow_once">Allow once</string>
+    <string name="tag_allow_always">Allow always</string>
+    <string name="tag_copy_id">Copy tag ID</string>
```
```suggestion
    <string name="tag_copy_id">Copy tag identifier</string>
```

common/src/main/res/values/strings.xml
```diff
@@ -490,6 +492,12 @@
     <string name="nfc_trigger_device">Scanned by this device:</string>
     <string name="nfc_invalid_tag">This tag does not contain Home Assistant data</string>
     <string name="nfc_processing_tag_error">Error while processing NFC tag</string>
+    <string name="tag_approval_title">Allow sending?</string>
+    <string name="tag_approval_description">Sending this tag to trigger automations linked to it.</string>
```
jpelgrom: Following discussion on Discord - with a minor tweak in the description as "an automation" could also be none or multiple. Don't forget to update screenshot tests.

cc @bgoncal 

```suggestion
    <string name="tag_approval_title">Allow tag?</string>
    <string name="tag_approval_description">Send this tag to Home Assistant to trigger linked automations.</string>
```

jpelgrom (commented):
One final round of comments, looks good otherwise.

app/src/main/kotlin/io/homeassistant/companion/android/nfc/views/TagReaderScreen.kt
TimoPtr: It's before the IconButton as it's own padding and the container as well. I'm going to see what I can do about it.
image

jpelgrom (approved):

PR #6815, opened 2026-05-07
Deprecated the old Material Design 2 theme to avoid using it

TimoPtr:
## Summary

We are in the middle of migrating to Material Design 3, we kept the `HomeAssistantAppTheme` for backward compatibility for screen that are not yet merged. This PR simply deprecates it so new contributors knows that we are migrating.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/util/compose/Theme.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/util/compose/Theme.kt
@@ -50,6 +50,13 @@ private val haDarkColors = darkColors(
  * A Compose [MaterialTheme] version of the app's XML theme. This achieves the same goal as the
  * (now deprecated) [com.google.accompanist.themeadapter.material.MdcTheme].
  */
+@Deprecated(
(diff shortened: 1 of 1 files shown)


jpelgrom:
Not sure about the autofix suggestion - it is not a drop in replacement and you might end up with contributors mixing M2 with M3 components which is also undesirable.

TimoPtr:
> Not sure about the autofix suggestion - it is not a drop in replacement and you might end up with contributors mixing M2 with M3 components which is also undesirable.

Well if we don't offer the autofix it might be the same too no? I'm fine dropping it still, just wonder if there is something else we can do.

jpelgrom:
> > Not sure about the autofix suggestion - it is not a drop in replacement and you might end up with contributors mixing M2 with M3 components which is also undesirable.
> 
> Well if we don't offer the autofix it might be the same too no? I'm fine dropping it still, just wonder if there is something else we can do.

Deprecation signals it shouldn't be used. If autofix is included that suggests (to me at least) that's all that is needed to 'fix' the issue while you also need to replace all components. If there is a deprecation but no autofix the contributor will at least be forced  to look at it and consider the consequences. Might still end up with a mix but hopefully less often.

jpelgrom (approved):

PR #3417, opened 2023-03-11
Get user information and prevent using admin-only calls

jpelgrom:
## Summary

Get informaton about the current user and cache it in the database alongside the server. Information is periodically updated, when the registration is synced.

Use information about whether the user is an admin or not to:
 - Fix #3313 by falling back to subscribing to all state changes + filtering results when the user is not an admin (the 'proper' solution would be to use `subscribe_entities` here but that requires significantly more work to handle re-used subscriptions, for now a quick fix is needed)
 - Block Matter/Thread functionality, as all required websocket APIs are restricted to admins

## Screenshots

n/a

## Link to pull request in Documentation repository

n/a

PR #6816, opened 2026-05-07
Show distinct message when Thread sync requires admin account

agners:
## Summary

When syncing Thread credentials as a non-admin user on the Home Assistant server, the app showed *"The Home Assistant server does not support Thread"*, which is misleading.

7049b449-c5c5-471c-95a4-a511efaafc12

This change splits the admin check out of `coreSupportsThread` so `syncPreferredDataset` can return a new `ServerUserNotAdmin` result distinct from `ServerUnsupported`, and surfaces a dedicated message:

> Managing Thread credentials requires an administrator account on the Home Assistant server

## Checklist


jpelgrom:
No objections to merge from me once ktlint is addressed

(Note in #3417 I simply used the existing 'server doesn't support' state out of convenience, not assuming this would be around for so long and the APIs would improve but unfortunately not)

PR #6818, opened 2026-05-07
Verify Thread dataset state after export and auto-promote when server has none

agners:
## Summary

When the developer-settings "Sync Thread credentials" flow exported the device's preferred dataset to Home Assistant, the app treated a successful `thread/add_dataset_tlv` reply as "credentials in sync". In practice the WebSocket command never makes the just-added dataset the server's preferred one — that auto-promote behaviour was removed in [Core #108278](https://github.com/home-assistant/core/pull/108278), which moved the preferred-dataset selection behind the explicit `thread/set_preferred_dataset` call. As a result, after Core #108278 the app's export flow would loop: every sync re-detected the same mismatch, prompted the user, exported again, and reported success while the server's preferred dataset had not changed at all.

This PR fixes the export flow to behave correctly with the new Core contract:

- `sendThreadDatasetExportResult` now returns a sealed `ExportResult` so callers can distinguish "sent and now preferred" from "sent but server still prefers another network".
- After a successful add the server is queried again and:
  - if no dataset is preferred, the just-exported one is promoted via the new `thread/set_preferred_dataset` WebSocket call;
  - if a different dataset is already preferred (typically because a server-managed border router announces it), the server's choice is left untouched and the UI reports the honest mismatch with a dedicated string;
  - if the just-exported dataset is already preferred, the UI reports the in-sync state.
- A new WebSocket repository method `setThreadPreferredDataset(datasetId)` wraps `thread/set_preferred_dataset`.
- The developer settings screen gets a new string `thread_debug_result_exported_not_preferred` so the user can tell the difference between a successful sync and an export-that-did-not-promote.

The `WebViewPresenterImpl` and `MatterCommissioningViewModel` callers are adjusted to the new return type without behaviour changes.

Related: [home-assistant/core#108278](https://github.com/home-assistant/core/pull/108278) — `thread/add_dataset_tlv` no longer promotes the added dataset to preferred, making this client-side change necessary.

## Checklist

## Any other notes

The `ThreadManager.sendThreadDatasetExportResult` signature changes from `String?` to `ExportResult`. All in-tree callers are updated; this is a developer-facing breaking change inside the app module only.

### DIFF
--- a/app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
+++ b/app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
@@ -295,19 +295,53 @@ class ThreadManagerImpl @Inject constructor(
                 .addOnFailureListener { cont.resumeWithException(it) }
         }
 
-    override suspend fun sendThreadDatasetExportResult(result: ActivityResult, serverId: Int): String? {
-        if (result.resultCode == Activity.RESULT_OK && coreSupportsThread(serverId)) {
-            val threadNetworkCredentials = ThreadNetworkCredentials.fromIntentSenderResultData(result.data!!)
-            try {
-                val added = serverManager.webSocketRepository(
-                    serverId,
-                ).addThreadDataset(threadNetworkCredentials.activeOperationalDataset)
-                if (added) return threadNetworkCredentials.networkName
-            } catch (e: Exception) {
-                Timber.e(e, "Error while executing server new Thread credentials request")
-            }
+    @OptIn(ExperimentalStdlibApi::class)
+    override suspend fun sendThreadDatasetExportResult(
+        result: ActivityResult,
+        serverId: Int,
+    ): ThreadManager.ExportResult {
+        if (result.resultCode != Activity.RESULT_OK || !coreSupportsThread(serverId)) {
+            return ThreadManager.ExportResult.NotSent
         }
-        return null
+        val credentials = ThreadNetworkCredentials.fromIntentSenderResultData(result.data!!)
+        val webSocket = serverManager.webSocketRepository(serverId)
+        val added = try {
+            webSocket.addThreadDataset(credentials.activeOperationalDataset)
+        } catch (e: Exception) {
+            Timber.e(e, "Error while executing server new Thread credentials request")
+            return ThreadManager.ExportResult.Failed
+        }
+        if (!added) return ThreadManager.ExportResult.Failed
+
+        // thread/add_dataset_tlv never promotes a dataset to preferred on the server side. To
+        // surface an honest result we re-query and compare extended PAN IDs. If the server has
+        // no preferred dataset at all (e.g. no border router announcing one), promote the
+        // just-exported dataset; we never override an existing server preference.
+        val deviceExtPan = credentials.extendedPanId.toHexString(HexFormat.UpperCase)
+        val serverPrefersExported = try {
+            val datasets = webSocket.getThreadDatasets()
+            val preferred = datasets?.firstOrNull { it.preferred }
+            when {
+                preferred == null -> {
+                    val ours = datasets?.firstOrNull { it.extendedPanId.equals(deviceExtPan, ignoreCase = true) }
+                    if (ours != null && webSocket.setThreadPreferredDataset(ours.datasetId)) {
+                        Timber.d("Thread: promoted exported dataset to preferred on the server")
+                        true
+                    } else {
+                        false
+                    }
+                }
+                preferred.extendedPanId.equals(deviceExtPan, ignoreCase = true) -> true
+                else -> false
+            }
+        } catch (e: Exception) {
+            Timber.w(e, "Unable to verify Thread preferred dataset after export")
+            null
+        }
+        return ThreadManager.ExportResult.Sent(
+            networkName = credentials.networkName,
+            serverPrefersExported = serverPrefersExported,
+        )
     }
 
     private suspend fun deleteOrphanedThreadCredentials(context: Context, serverId: Int) {
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/developer/DeveloperSettingsPresenterImpl.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/developer/DeveloperSettingsPresenterImpl.kt
@@ -133,19 +133,39 @@ class DeveloperSettingsPresenterImpl @Inject constructor(
     ) {
         mainScope.launch {
             try {
-                val submitted = threadManager.sendThreadDatasetExportResult(result, serverId)
-                if (submitted != null) {
-                    if (isDeviceOnly) {
-                        view.onThreadDebugResult(context.getString(commonR.string.thread_debug_result_exported), true)
-                    } else {
-                        // If we got permission while both had a dataset, the device prefers a different network
-                        val out = "${context.getString(
-                            commonR.string.thread_debug_result_mismatch,
-                        )} ${context.getString(commonR.string.thread_debug_result_mismatch_detail, submitted)}"
-                        view.onThreadDebugResult(out, null)
+                when (val sent = threadManager.sendThreadDatasetExportResult(result, serverId)) {
+                    is ThreadManager.ExportResult.Sent -> {
+                        if (isDeviceOnly) {
+                            view.onThreadDebugResult(
+                                context.getString(commonR.string.thread_debug_result_exported),
+                                true,
+                            )
+                        } else if (sent.serverPrefersExported == true) {
+                            view.onThreadDebugResult(context.getString(commonR.string.thread_debug_result_match), true)
+                        } else if (sent.serverPrefersExported == false) {
+                            // The dataset was added on the server but a different one is still
+                            // preferred (typically a server-managed border router announces it).
+                            val out = "${context.getString(
+                                commonR.string.thread_debug_result_exported_not_preferred,
+                            )} ${context.getString(
+                                commonR.string.thread_debug_result_mismatch_detail,
+                                sent.networkName,
+                            )}"
+                            view.onThreadDebugResult(out, null)
+                        } else {
+                            val out = "${context.getString(
+                                commonR.string.thread_debug_result_mismatch,
+                            )} ${context.getString(
+                                commonR.string.thread_debug_result_mismatch_detail,
+                                sent.networkName,
+                            )}"
+                            view.onThreadDebugResult(out, null)
+                        }
                     }
-                } else {
-                    view.onThreadDebugResult(context.getString(commonR.string.thread_debug_result_error), false)
+                    ThreadManager.ExportResult.NotSent,
+                    ThreadManager.ExportResult.Failed,
+                    ->
+                        view.onThreadDebugResult(context.getString(commonR.string.thread_debug_result_error), false)
                 }
             } catch (e: Exception) {
                 view.onThreadDebugResult(context.getString(commonR.string.thread_debug_result_error), false)
--- a/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
@@ -603,12 +603,16 @@ class WebViewPresenterImpl @Inject constructor(
                 mainScope.launch {
                     val sent = threadUseCase.sendThreadDatasetExportResult(result, serverId)
                     Timber.d(
-                        "Thread ${if (!sent.isNullOrBlank()) "sent credential for $sent" else "did not send credential"}",
+                        "Thread ${if (sent is ThreadManager.ExportResult.Sent) {
+                            "sent credential for ${sent.networkName} (server prefers exported: ${sent.serverPrefersExported})"
+                        } else {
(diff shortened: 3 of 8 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/webview/WebViewPresenterImpl.kt
```diff
@@ -603,12 +603,16 @@ class WebViewPresenterImpl @Inject constructor(
                 mainScope.launch {
                     val sent = threadUseCase.sendThreadDatasetExportResult(result, serverId)
                     Timber.d(
-                        "Thread ${if (!sent.isNullOrBlank()) "sent credential for $sent" else "did not send credential"}",
+                        "Thread ${if (sent is ThreadManager.ExportResult.Sent) {
+                            "sent credential for ${sent.networkName} (server prefers exported: ${sent.serverPrefersExported})"
+                        } else {
+                            "did not send credential"
```
TimoPtr, line 609:
```suggestion
                            "did not send credential reason $sent"
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketRepositoryImpl.kt
```diff
@@ -420,6 +420,16 @@ class WebSocketRepositoryImpl internal constructor(
         return response?.success == true
     }
 
+    override suspend fun setThreadPreferredDataset(datasetId: String): Boolean {
+        val response = webSocketCore.sendMessage(
+            mapOf(
+                "type" to "thread/set_preferred_dataset",
+                "dataset_id" to datasetId,
```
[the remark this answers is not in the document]
TimoPtr: Indeed

app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
```diff
@@ -295,19 +295,53 @@ class ThreadManagerImpl @Inject constructor(
...
+        // no preferred dataset at all (e.g. no border router announcing one), promote the
+        // just-exported dataset; we never override an existing server preference.
+        val deviceExtPan = credentials.extendedPanId.toHexString(HexFormat.UpperCase)
+        val serverPrefersExported = try {
+            val datasets = webSocket.getThreadDatasets()
+            val preferred = datasets?.firstOrNull { it.preferred }
+            when {
+                preferred == null -> {
+                    val ours = datasets?.firstOrNull { it.extendedPanId.equals(deviceExtPan, ignoreCase = true) }
+                    if (ours != null && webSocket.setThreadPreferredDataset(ours.datasetId)) {
```
TimoPtr, line 327: We probably need to check if core has this endpoint no? Or the min version we use for thread already contains this endpoint?

app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
```diff
@@ -295,19 +295,53 @@ class ThreadManagerImpl @Inject constructor(
...
-                Timber.e(e, "Error while executing server new Thread credentials request")
+    @OptIn(ExperimentalStdlibApi::class)
+    override suspend fun sendThreadDatasetExportResult(
+        result: ActivityResult,
+        serverId: Int,
+    ): ThreadManager.ExportResult {
+        if (result.resultCode != Activity.RESULT_OK || !coreSupportsThread(serverId)) {
+            return ThreadManager.ExportResult.NotSent
+        }
+        val credentials = ThreadNetworkCredentials.fromIntentSenderResultData(result.data!!)
```
TimoPtr, line 306: We can avoid crashing here by checking if data is null and returning NotSent.

app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
```diff
@@ -295,19 +295,53 @@ class ThreadManagerImpl @Inject constructor(
...
+        val webSocket = serverManager.webSocketRepository(serverId)
+        val added = try {
+            webSocket.addThreadDataset(credentials.activeOperationalDataset)
+        } catch (e: Exception) {
```
TimoPtr, line 310: It's a coroutine, we have a bad design in the app by catching all Exception but we can't change it now. But because of that we have to reThrow `CancellationException` otherwise it let the scope into an invalid state.

app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
```diff
@@ -295,19 +295,53 @@ class ThreadManagerImpl @Inject constructor(
...
+                        Timber.d("Thread: promoted exported dataset to preferred on the server")
+                        true
+                    } else {
+                        false
+                    }
+                }
+                preferred.extendedPanId.equals(deviceExtPan, ignoreCase = true) -> true
+                else -> false
             }
+        } catch (e: Exception) {
```
TimoPtr, line 337: same here we need to rethrow

app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
```diff
@@ -295,19 +295,53 @@ class ThreadManagerImpl @Inject constructor(
...
-        if (result.resultCode == Activity.RESULT_OK && coreSupportsThread(serverId)) {
-            val threadNetworkCredentials = ThreadNetworkCredentials.fromIntentSenderResultData(result.data!!)
-            try {
-                val added = serverManager.webSocketRepository(
-                    serverId,
-                ).addThreadDataset(threadNetworkCredentials.activeOperationalDataset)
-                if (added) return threadNetworkCredentials.networkName
-            } catch (e: Exception) {
-                Timber.e(e, "Error while executing server new Thread credentials request")
+    @OptIn(ExperimentalStdlibApi::class)
```
TimoPtr, line 298:
```suggestion
```
Not needed anymore (you can probably also remove it in the rest of the file)

app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
```diff
@@ -295,19 +295,53 @@ class ThreadManagerImpl @Inject constructor(
...
+            Timber.e(e, "Error while executing server new Thread credentials request")
+            return ThreadManager.ExportResult.Failed
+        }
+        if (!added) return ThreadManager.ExportResult.Failed
+
+        // thread/add_dataset_tlv never promotes a dataset to preferred on the server side. To
+        // surface an honest result we re-query and compare extended PAN IDs. If the server has
+        // no preferred dataset at all (e.g. no border router announcing one), promote the
+        // just-exported dataset; we never override an existing server preference.
+        val deviceExtPan = credentials.extendedPanId.toHexString(HexFormat.UpperCase)
```
TimoPtr, line 320:
```suggestion
        val deviceExtPan = credentials.extendedPanId.toHexString()
```
Since you ignore the case bellow we don't care

app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
```diff
@@ -295,19 +295,53 @@ class ThreadManagerImpl @Inject constructor(
...
+        val serverPrefersExported = try {
```
TimoPtr, line 321: The whole logic could be extracted into a private function.

app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
```diff
@@ -295,19 +295,53 @@ class ThreadManagerImpl @Inject constructor(
...
+            val datasets = webSocket.getThreadDatasets()
+            val preferred = datasets?.firstOrNull { it.preferred }
+            when {
+                preferred == null -> {
+                    val ours = datasets?.firstOrNull { it.extendedPanId.equals(deviceExtPan, ignoreCase = true) }
+                    if (ours != null && webSocket.setThreadPreferredDataset(ours.datasetId)) {
+                        Timber.d("Thread: promoted exported dataset to preferred on the server")
```
TimoPtr, line 328:
```suggestion
                        Timber.d("Promote exported dataset to preferred network on the server")
```

app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
```diff
@@ -295,19 +295,53 @@ class ThreadManagerImpl @Inject constructor(
...
-            val threadNetworkCredentials = ThreadNetworkCredentials.fromIntentSenderResultData(result.data!!)
-            try {
-                val added = serverManager.webSocketRepository(
-                    serverId,
-                ).addThreadDataset(threadNetworkCredentials.activeOperationalDataset)
-                if (added) return threadNetworkCredentials.networkName
-            } catch (e: Exception) {
-                Timber.e(e, "Error while executing server new Thread credentials request")
+    @OptIn(ExperimentalStdlibApi::class)
+    override suspend fun sendThreadDatasetExportResult(
```
TimoPtr, line 299: If you are up for the task we could start unit testing this function. Claude is pretty good at this exercise on Android.

app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
```diff
@@ -295,19 +295,53 @@ class ThreadManagerImpl @Inject constructor(
...
+                        true
+                    } else {
+                        false
+                    }
+                }
+                preferred.extendedPanId.equals(deviceExtPan, ignoreCase = true) -> true
+                else -> false
             }
+        } catch (e: Exception) {
+            Timber.w(e, "Unable to verify Thread preferred dataset after export")
```
TimoPtr, line 338: Could also be setting the prefered dataset since you potentially makes 2 call.

app/src/main/kotlin/io/homeassistant/companion/android/settings/developer/DeveloperSettingsPresenterImpl.kt
```diff
@@ -133,19 +133,39 @@ class DeveloperSettingsPresenterImpl @Inject constructor(
...
-                if (submitted != null) {
-                    if (isDeviceOnly) {
-                        view.onThreadDebugResult(context.getString(commonR.string.thread_debug_result_exported), true)
-                    } else {
-                        // If we got permission while both had a dataset, the device prefers a different network
-                        val out = "${context.getString(
-                            commonR.string.thread_debug_result_mismatch,
-                        )} ${context.getString(commonR.string.thread_debug_result_mismatch_detail, submitted)}"
-                        view.onThreadDebugResult(out, null)
+                when (val sent = threadManager.sendThreadDatasetExportResult(result, serverId)) {
```
TimoPtr, line 136: I feel like it can be simplified maybe something like this

```kotlin
                when (val sent = threadManager.sendThreadDatasetExportResult(result, serverId)) {
                    is ThreadManager.ExportResult.Sent -> when {
                        isDeviceOnly -> view.onThreadDebugResult(
                            context.getString(commonR.string.thread_debug_result_exported),
                            true,
                        )
                        sent.serverPrefersExported == true -> view.onThreadDebugResult(
                            context.getString(commonR.string.thread_debug_result_match),
                            true,
                        )
                        else -> {
                            val headlineRes = if (sent.serverPrefersExported == false) {
                                commonR.string.thread_debug_result_exported_not_preferred
                            } else {
                                commonR.string.thread_debug_result_mismatch
                            }
                            val detail = context.getString(
                                commonR.string.thread_debug_result_mismatch_detail,
                                sent.networkName,
                            )
                            view.onThreadDebugResult("${context.getString(headlineRes)} $detail", null)
                        }
                    }
```

app/src/full/kotlin/io/homeassistant/companion/android/thread/ThreadManagerImpl.kt
```diff
@@ -295,19 +295,53 @@ class ThreadManagerImpl @Inject constructor(
...
+                    val ours = datasets?.firstOrNull { it.extendedPanId.equals(deviceExtPan, ignoreCase = true) }
+                    if (ours != null && webSocket.setThreadPreferredDataset(ours.datasetId)) {
+                        Timber.d("Thread: promoted exported dataset to preferred on the server")
+                        true
+                    } else {
+                        false
+                    }
+                }
+                preferred.extendedPanId.equals(deviceExtPan, ignoreCase = true) -> true
+                else -> false
```
TimoPtr, line 335:
```suggestion
      val serverPrefersExported = try {
            val datasets = webSocket.getThreadDatasets().orEmpty()
            val preferred = datasets.firstOrNull { it.preferred }
            if (preferred != null) {
                preferred.extendedPanId.equals(deviceExtPan, ignoreCase = true)
            } else {
                val ours = datasets.firstOrNull { it.extendedPanId.equals(deviceExtPan, ignoreCase = true) }
                if (ours != null && webSocket.setThreadPreferredDataset(ours.datasetId)) {
                    Timber.d("Thread: promoted exported dataset to preferred on the server")
                    true
                } else {
                    false
                }
            }
```

The when using instead of a if is sometimes less readable IMO.

PR #6835, opened 2026-05-11
Restore HTTP 404/410 handling for updating registration

### LINKED ISSUE #6722: HTTP 410 Gone on webhook does not trigger device re-registration
## Version
- App: 2026.3.2-full
- Android: 16 (SDK 36)
- Phone: Pixel 9 Pro XL

## Description

When a server's webhook ID becomes invalid (HA returns `HTTP 410 Gone`), the app does not attempt to re-register the device. It silently keeps using the dead webhook ID, causing all webhook-based operations (`update_location`, `get_zones`, `update_sensor_states`, etc.) to fail permanently until the user manually removes and re-adds the server.

## Root cause

In `IntegrationRepositoryImpl.updateRegistration()` ([line 160-184](https://github.com/home-assistant/android/blob/main/common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImpl.kt#L160-L184)), there is logic to detect a broken registration and re-register:

```kotlin
if (response.code() == 200 && response.body()?.contentLength() == 0L) {
    // Reregistration logic
}
```

However, this only handles the case of a 200 with an empty body. A `410 Gone` response is thrown as an `HttpException` by Retrofit before the `onSuccess` handler runs, so the re-registration logic is never reached.

Meanwhile, `callWebhookOnUrls()` in `tryOnUrls()` catches the 410 and moves on to the next URL. If all URLs return 410, it throws `IntegrationException("All URLs failed")`. At no point is re-registration attempted.

## Impact

The app gets permanently stuck with a dead webhook. In my case:
- The webhook ID was invalidated at some point (possibly during an HA update or config entry cleanup)
- Every `get_zones`, `update_location`, and `update_sensor_states` call returned 410 Gone
- Location tracking was completely broken for days
- The only fix was manually removing and re-adding the server in the app

Combined with #6721 (multi-server geofence failure), this made location tracking silently broken with no user-visible indication.

## Expected behavior

When the app receives `HTTP 410 Gone` on a webhook call, it should:
1. Recognize that the webhook registration is invalid
2. Attempt to re-register the device with the server (call `registerDevice()`)
3. If re-registration succeeds, persist the new webhook ID and retry the original request
4. If re-registration fails, surface an error to the user (persistent notification, sensor status, etc.)

## Logs

```
W ServerConnectionStateProviderKt: Failed get_zones on HIDDEN
W ServerConnectionStateProviderKt: retrofit2.HttpException: HTTP 410 Gone
E LocationSensorManager: Error receiving zones from Home Assistant
E LocationSensorManager: IntegrationException: All URLs failed for get_zones
```

jpelgrom:
## Summary

Fixes #6722 by restoring functionality unintentionally removed in #5487

## Checklist

## Screenshots

n/a

n/a

### DIFF
--- a/common/src/test/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImplTest.kt
+++ b/common/src/test/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImplTest.kt
@@ -1,33 +1,46 @@
 package io.homeassistant.companion.android.common.data.integration.impl
 
 import io.homeassistant.companion.android.common.data.LocalStorage
+import io.homeassistant.companion.android.common.data.integration.DeviceRegistration
+import io.homeassistant.companion.android.common.data.integration.IntegrationException
 import io.homeassistant.companion.android.common.data.integration.IntegrationRepository
 import io.homeassistant.companion.android.common.data.integration.impl.IntegrationRepositoryImpl.Companion.PREF_ASK_NOTIFICATION_PERMISSION
 import io.homeassistant.companion.android.common.data.servers.ServerConnectionStateProvider
 import io.homeassistant.companion.android.common.data.servers.ServerManager
+import io.homeassistant.companion.android.common.data.servers.UrlState
 import io.homeassistant.companion.android.database.server.Server
 import io.homeassistant.companion.android.database.server.ServerConnectionInfo
+import io.mockk.Runs
 import io.mockk.coEvery
 import io.mockk.coVerify
 import io.mockk.every
+import io.mockk.just
 import io.mockk.mockk
+import io.mockk.spyk
+import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.test.runTest
 import kotlinx.serialization.json.JsonArray
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
 import okhttp3.HttpUrl.Companion.toHttpUrl
+import okhttp3.ResponseBody.Companion.toResponseBody
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertNull
 import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Assertions.fail
 import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
+import retrofit2.Response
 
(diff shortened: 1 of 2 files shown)


jpelgrom:
I tried adding a test like below but kept running into generic `'Gradle Test Executor 47' finished with non-zero exit value 1` errors which aren't fun to debug, so I gave up after 30 minutes or so.

<details>
<summary>Test</summary>

```kotlin
@Test
fun `Given success when updating registration then device registration is persisted`() = runTest {
    val body = "content".toResponseBody()

    coEvery { integrationService.callWebhook(any(), any()) } returns Response.success(body)

    coEvery { localStorage.getString(any()) } returns null
    coEvery { serverManager.updateServer(any()) } returns Unit

    val registration = DeviceRegistration(deviceName = "New device name")
    repository.updateRegistration(
        registration,
        allowReregistration = true,
    )

    coVerify(exactly = 0) { repository.registerDevice(any()) }
    coVerify { serverManager.updateServer(any()) }
}
```
</details>

TimoPtr:
> I tried adding a test like below but kept running into generic `'Gradle Test Executor 47' finished with non-zero exit value 1` errors which aren't fun to debug, so I gave up after 30 minutes or so.
> Test

You probably are missing the FailFastRule here and the code is triggering FailFast that crash the process without a nice message. I didn't find a way to globally apply a rule for now ...

TimoPtr:
Would you mind pushing the tests you've wrote so I can help debug?

TimoPtr (approved):
Looks good but let's try to bring back the test if you wrote them already.

jpelgrom:
I'll try once more to get working tests with the note you've given about FailFastRule, and push them later tonight (successful or not).

jpelgrom:
There was indeed a fail fast error hiding, thanks for the help. I've added tests for updating registration: the happy flow, 3 known failures with re-registration, and a failure but no re-registration allowed.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/integration/impl/IntegrationRepositoryImplTest.kt
```diff
@@ -162,4 +179,97 @@ class IntegrationRepositoryImplTest {
         coVerify { localStorage.putBoolean("${serverID}_$PREF_ASK_NOTIFICATION_PERMISSION", true) }
         coVerify { localStorage.putBoolean("${otherServerId}_$PREF_ASK_NOTIFICATION_PERMISSION", false) }
     }
+
+    @Nested
+    inner class UpdateRegistrationTests {
+
+        @Test
+        fun `Given success when updating registration then registration is persisted`() = runTest {
```
TimoPtr, line 187: I would have add a mention to the fact that we are not making a new registerDevice call but it's fine like this 👍🏻

TimoPtr (approved):

PR #6836, opened 2026-05-11
[compose UI] Refactor compose path handling

Gifford47:
## Summary

Companion PR to #6447 (WebView legacy). Brings the same predictive-back behavior to the Compose FrontendScreen and resolves the navigation chain so Android 14+ shows the peek animation properly.
## Checklist

[Refactoring webview PR](https://github.com/home-assistant/android/pull/6447)

## Any other notes
6b1efcca3 — Enable Android 14+ predictive-back by gating BackHandler on canGoBack
The HAWebView BackHandler was unconditionally enabled, which suppressed the predictive-back peek animation. State-hoist canGoBack from caller (FrontendViewModel/ConnectionViewModel) via a new onCanGoBackChanged hook in HAWebViewClient (fires from doUpdateVisitedHistory to catch SPA pushState + from onPageFinished). HAWebView API now takes canGoBack: Boolean; onBackPressed: (() -> Unit)? becomes an optional fallback for screens like onboarding that explicitly need to pop the nav stack (ConnectionScreen).

7e46150cf — Clear navigation back-stack after onboarding completes
navigateToFrontend() was pushing FrontendRoute on top of the finished OnboardingRoute, so back from the frontend popped to an empty (white) onboarding screen instead of exiting the app. Use popUpTo(0) { inclusive = true } so FrontendRoute is the only entry.

44ecb1fb6 — Clear WebView history after transitioning out of about:blank
The WebView accumulated every loadUrl() including the about:blank placeholder used by LoadServer/Error/SecurityLevelRequired states. Pressing back from the root walked back into a white about:blank screen. HAWebViewClient.onPageFinished now calls webView.clearHistory() once it detects the transition BLANK_URL → real URL. SPA pushState is unaffected (only fires doUpdateVisitedHistory, not onPageFinished).

Verification
Tested on Android 16 emulator (API 36.1), gesture navigation, with WIPFeature.USE_FRONTEND_V2 = BuildConfig.DEBUG (Compose path)


jpelgrom:
FYI, there are already plenty of conflicts to resolve here but #7001 will need special attention as it changes how back is handled (putting it in one place, part of the state).

TimoPtr:
#7046 is going to bring back your PR to main once we've made a new release for another beta.

jpelgrom:
@Gifford47 The WebViewActivity has now been replaced by the Compose-based screens, do you want to rebase this or start over?

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
```diff
@@ -261,7 +264,22 @@ internal fun FrontendScreenContent(
...
+    BackHandler(enabled = content?.canGoBack == true) {
+        val view = webView ?: return@BackHandler
+        when (val action = resolveBackAction(view, content?.url?.toUri())) {
+            BackAction.GoBack -> view.goBack()
+            is BackAction.NavigateToRoot -> {
+                view.loadUrl(action.rootUrl.toString())
+                // Drop the stale cross-origin entry so the next back press exits the app
+                // instead of looping back into it.
+                view.clearHistory()
+            }
```
TimoPtr: I want this to go through the ViewModel and be fully tested instead of being in the composable layer. Ideally you should have a compose test that verify that a back event does trigger the proper callback of the composable.

TimoPtr (changes requested):

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
Gifford47: Done in e12c4fa31: the composable now only reports the gesture — `BackHandler` calls `FrontendViewModel.onBackPressed()`, which emits a new `WebViewAction.NavigateBack` that resolves against the back/forward list via `resolveBackAction`. Covered by ViewModel tests for the emission gating, action tests for all three resolution outcomes, and compose tests verifying the back event does/doesn't reach the callback depending on `canGoBack`.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
```diff
@@ -261,7 +263,9 @@ internal fun FrontendScreenContent(
     val loadingSurfaceColor = LocalHAColorScheme.current.colorSurfaceDefault
 
     // Consume back only while the dashboard (Content) is shown and the WebView has history to pop.
-    BackHandler(enabled = content?.canGoBack == true) { webView?.goBack() }
+    // Resolution against the WebView back stack happens in the ViewModel via
+    // WebViewAction.NavigateBack (stale cross-origin entries route to the dashboard root).
```
TimoPtr: I don't think we need to mention the details of what the VM does.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -1088,6 +1110,20 @@ internal class FrontendViewModel @VisibleForTesting constructor(
      * navigations can reset the viewport meta tag). The collection then stays active
      * to react to settings changes until the next page load restarts it.
      */
+    /**
```
TimoPtr: Careful you moved the documentation of onPageFinished above.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -1088,6 +1110,20 @@ internal class FrontendViewModel @VisibleForTesting constructor(
...
+     * Handles the system back gesture while the dashboard consumes it ([FrontendViewState.Content]
+     * with WebView history). Emits [WebViewAction.NavigateBack], which resolves the gesture against
+     * the WebView's back/forward list: same-origin entries pop normally, a stale cross-origin entry
+     * routes to the dashboard root instead. Outside Content-with-history the screen does not consume
+     * the gesture, so this is a no-op.
+     */
```
TimoPtr:
```suggestion
     * Handles the system back gesture.
     */
```
The rest is for the VM specific documentation of the function.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManager.kt
```diff
@@ -43,13 +43,21 @@ class FrontendUrlManager @Inject constructor(
...
      * The target is only applied to the first emission to handle deep links.
-     * Subsequent emissions (e.g., when switching between internal/external URLs) load only the base URL.
+     * When a subsequent emission changes the base URL (e.g., switching between internal/external
+     * URLs on the same server), the relative URL reported by [currentRelativeUrl] is applied to
+     * the new base so the user stays on the same page; otherwise only the base URL is loaded.
      *
      * @param serverId The server ID to use (can be [ServerManager.SERVER_ID_ACTIVE])
      * @param target The frontend destination to open on the initial URL (e.g., a deep link)
+     * @param currentRelativeUrl Provides the relative URL (path + query + fragment) currently shown
+     *        in the WebView, or `null` when unknown or at the dashboard root
```
TimoPtr: Do we really need to have null for dashboard root? it's making the logic quite complex.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
```diff
@@ -95,6 +98,35 @@ sealed interface WebViewAction {
...
+        override fun run(webView: WebView) {
+            when (val action = resolveBackAction(webView, loadedUrl?.toUri())) {
+                BackAction.GoBack -> webView.goBack()
+                is BackAction.NavigateToRoot -> {
+                    webView.loadUrl(action.rootUrl.toString())
+                    // Drop the stale cross-origin entry so the next back press exits the app
+                    // instead of looping back into it.
+                    webView.clearHistory()
+                }
+                BackAction.None -> Unit
```
TimoPtr: This case is dangerous I would say, the user would press back and nothing would happen. It might worse making sure we cannot arrive in this case and also add a FailFast here for us while testing.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -1088,6 +1110,20 @@ internal class FrontendViewModel @VisibleForTesting constructor(
...
+    fun onBackPressed() {
+        val state = _viewState.value
+        if (state is FrontendViewState.Content && state.canGoBack) {
+            viewModelScope.launch { _webViewActions.emit(WebViewAction.NavigateBack(state.url)) }
```
TimoPtr: Are you sure that we should use the state.url here or should we use the loadedUrl that you've introduced?

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -754,6 +769,13 @@ internal class FrontendViewModel @VisibleForTesting constructor(
             urlManager.serverUrlFlow(
                 serverId = currentState.serverId,
                 target = target,
+                currentRelativeUrl = {
+                    // external_auth is re-added by the URL manager; more-info-entity-id is a
+                    // one-shot deep-link parameter that must not reopen the dialog after a switch.
+                    lastVisitedUrl?.toUri()?.toRelativeUrl(
+                        excludeParams = setOf("external_auth", "more-info-entity-id"),
```
TimoPtr: I would like to avoid the usage of string here and create const shared across the project if it makes sense.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -238,6 +240,16 @@ internal class FrontendViewModel @VisibleForTesting constructor(
...
 
+    /**
+     * Last URL reported by the WebView, including SPA history updates (`history.pushState`),
+     * so it reflects the page actually shown to the user. Used to preserve the current relative
+     * URL when the base URL switches (internal <-> external) on the same server.
+     *
+     * Only accessed on the main thread: writes come from the WebView client callback and reads
+     * from the URL flow collection in [viewModelScope] (the main dispatcher).
+     */
+    private var lastVisitedUrl: String? = null
```
TimoPtr, line 256: We should probably nullify it as soon as possible when calling serverSwitch?

TimoPtr (changes requested):

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
Gifford47: Fixed

Gifford47:
Added the FailFast. One thing to flag: BackAction.None is actually reachable at runtime, not purely an invariant violation.it occurs when the previous back-stack entry is cross-origin (same trigger as NavigateToRoot) but the current page is already at the root, so there's no sub-path to collapse. There the FailFast logs in release (no crash) and surfaces in debug/tests. If you'd rather the user isn't left with a dead back press in that case, I can additionally clear the stale entry so the next back exits the app

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
Gifford47: trimmed it to just note when back is consumed;

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
Gifford47: Applied

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
Gifford47: Added the FailFast. One thing to flag: BackAction.None is actually reachable at runtime, not purely an invariant violation.it occurs when the previous back-stack entry is cross-origin (same trigger as NavigateToRoot) but the current page is already at the root, so there's no sub-path to collapse. There the FailFast logs in release (no crash) and surfaces in debug/tests. If you'd rather the user isn't left with a dead back press in that case, I can additionally clear the stale entry so the next back exits the app

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
Gifford47: You're right -switched to the actually-visited URL (lastVisitedUrl), which tracks in-frontend SPA navigation

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
Gifford47: Done -added a shared FrontendUrlParams object

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
Gifford47: Done -lastVisitedUrl is now cleared in switchServer

app/src/main/kotlin/io/homeassistant/companion/android/frontend/url/FrontendUrlManager.kt
Gifford47: Simplified -he branch now delegates to the existing FrontendTarget.fromRawPath(), which already maps a null relative URL

Gifford47:
@TimoPtr How's the plan for implementation? I see many changes since last weeks.

TimoPtr:
> @TimoPtr How's the plan for implementation? I see many changes since last weeks.

I plan to make first a release with the removal of the WebView then look at your PR to add it to the beta.

PR #6838, opened 2026-05-11
WebView: validate SSL_UNTRUSTED certs via OkHttp trust store

cryptomilk:
## Summary

When connecting to a Home Assistant instance using a private CA, the WebView's Chromium TLS stack reports SSL_UNTRUSTED (error code 3) even though network_security_config.xml includes user CAs. The app's OkHttpClient correctly trusts user-installed CAs via AndroidCAStore, so we use it as a second opinion before rejecting the connection.

Observed log:
  onReceivedSslError: primary error: 3 certificate: Issued to: CN=antares.milli.ways;
  Issued by: CN=milly.ways CA Intermediate CA,O=milly.ways CA;
  on URL: https://homeassistant.milli.ways/auth/authorize?...

TLSWebViewClient now overrides onReceivedSslError: for SSL_UNTRUSTED with a non-null URL it launches async validation via isTlsTrusted(); all other errors (IDMISMATCH, EXPIRED, etc.) are rejected immediately.

## Checklist

## Any other notes

I don't have a build environment for this. I already looked a the code last month and there was a similar PR, but I already thought it isn't going to fix it. I tested 2026.5.3 today. I still can't login. So this should hopefully fix it.

As this is SSL cert validation code it needs probably a few more eyes to check if it is correct. I now certificate stuff but not Android.

Claude helped reviewing the code and wrote the tests.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
@@ -5,7 +5,6 @@ import android.net.Uri
 import android.net.http.SslError
 import android.webkit.HttpAuthHandler
 import android.webkit.RenderProcessGoneDetail
-import android.webkit.SslErrorHandler
 import android.webkit.WebResourceError
 import android.webkit.WebResourceRequest
 import android.webkit.WebResourceResponse
@@ -15,8 +14,14 @@ import io.homeassistant.companion.android.common.R as commonR
 import io.homeassistant.companion.android.common.data.keychain.KeyChainRepository
 import io.homeassistant.companion.android.common.data.keychain.NamedKeyChain
 import io.homeassistant.companion.android.frontend.error.FrontendConnectionError
+import java.io.IOException
 import javax.inject.Inject
+import javax.net.ssl.X509TrustManager
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.StateFlow
+import okhttp3.OkHttpClient
 import timber.log.Timber
 
 /**
@@ -25,10 +30,17 @@ import timber.log.Timber
  * The created clients handle Home Assistant-specific concerns such as TLS client authentication,
  * error mapping to [FrontendConnectionError], and JavaScript injection into the WebView.
  */
-class HAWebViewClientFactory @Inject constructor(@NamedKeyChain private val keyChainRepository: KeyChainRepository) {
+class HAWebViewClientFactory @Inject constructor(
+    @NamedKeyChain private val keyChainRepository: KeyChainRepository,
+    private val trustManager: X509TrustManager,
+    private val okHttpClient: OkHttpClient,
+) {
     /**
      * Creates a new [HAWebViewClient] with the specified configuration.
      *
+     * @param validationScope Scope used for async TLS validation coroutines. Should be tied to the
+     *        caller's lifecycle (e.g., `viewModelScope`) so in-flight validations are cancelled on
+     *        destruction.
      * @param currentUrlFlow StateFlow providing the current URL being loaded.
      *        Used to filter errors - only errors for this URL trigger [onFrontendError].
      * @param onFrontendError Callback when a WebView error is mapped to a [FrontendConnectionError].
@@ -41,6 +53,7 @@ class HAWebViewClientFactory @Inject constructor(@NamedKeyChain private val keyC
      *        Receives the handler, host, the resource URL that triggered the request, and the realm.
      */
     fun create(
+        validationScope: CoroutineScope,
         currentUrlFlow: StateFlow<String?>,
         onFrontendError: (FrontendConnectionError) -> Unit,
         onCrash: (() -> Unit)? = null,
@@ -57,6 +70,9 @@ class HAWebViewClientFactory @Inject constructor(@NamedKeyChain private val keyC
     ): HAWebViewClient {
         return HAWebViewClient(
             keyChainRepository = keyChainRepository,
+            validationScope = validationScope,
+            trustManager = trustManager,
+            okHttpClient = okHttpClient,
             currentUrlFlow = currentUrlFlow,
             onFrontendError = onFrontendError,
             onCrash = onCrash,
@@ -75,6 +91,11 @@ class HAWebViewClientFactory @Inject constructor(@NamedKeyChain private val keyC
  */
 class HAWebViewClient internal constructor(
     keyChainRepository: KeyChainRepository,
+    validationScope: CoroutineScope,
+    trustManager: X509TrustManager,
+    /** Used as a fallback for API < 29, where [android.net.http.SslCertificate.x509Certificate] is unavailable. */
+    okHttpClient: OkHttpClient,
+    ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
     private val currentUrlFlow: StateFlow<String?>,
     private val onFrontendError: (FrontendConnectionError) -> Unit,
     private val onCrash: (() -> Unit)?,
@@ -83,7 +104,7 @@ class HAWebViewClient internal constructor(
     private val onReceivedHttpAuthRequest: (
         (handler: HttpAuthHandler, host: String, resource: String, realm: String) -> Unit
     )?,
-) : TLSWebViewClient(keyChainRepository) {
+) : TLSWebViewClient(keyChainRepository, validationScope, trustManager, okHttpClient, ioDispatcher) {
 
     /** Last resource URL loaded by the WebView, used to identify the resource requesting auth. */
     private var lastResourceUrl: String? = null
@@ -216,10 +237,23 @@ class HAWebViewClient internal constructor(
         onFrontendError(frontendConnectionError)
     }
 
-    override fun onReceivedSslError(view: WebView?, handler: SslErrorHandler?, error: SslError?) {
-        super.onReceivedSslError(view, handler, error)
-        Timber.e("onReceivedSslError: $error")
+    override fun onTlsValidationNetworkError(error: SslError?) {
+        Timber.w("TLS validation network error on ${sensitive { error?.url.orEmpty() }}")
+        onFrontendError(
+            FrontendConnectionError.UnreachableError(
+                message = commonR.string.webview_error_CONNECT,
+                errorDetails = error.toString(),
+                rawErrorType = IOException::class.toString(),
+            ),
+        )
+    }
 
+    override fun onSslErrorRejected(error: SslError?) {
+        Timber.e(
+            "SSL connection rejected: primary error ${error?.primaryError} on ${sensitive {
+                error?.url.orEmpty()
+            }}",
+        )
         val messageRes = when (error?.primaryError) {
             SslError.SSL_DATE_INVALID -> commonR.string.webview_error_SSL_DATE_INVALID
             SslError.SSL_EXPIRED -> commonR.string.webview_error_SSL_EXPIRED
--- a/app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
@@ -4,21 +4,35 @@ import android.app.Activity
 import android.content.ActivityNotFoundException
 import android.content.Context
 import android.content.ContextWrapper
+import android.net.http.SslCertificate
+import android.net.http.SslError
+import android.os.Build
 import android.security.KeyChain
 import android.security.KeyChainAliasCallback
 import android.webkit.ClientCertRequest
+import android.webkit.SslErrorHandler
 import android.webkit.WebView
 import android.webkit.WebViewClient
 import androidx.annotation.VisibleForTesting
 import io.homeassistant.companion.android.common.data.keychain.KeyChainRepository
+import io.homeassistant.companion.android.util.sensitive
+import java.io.IOException
 import java.lang.ref.WeakReference
 import java.security.PrivateKey
 import java.security.cert.CertificateException
 import java.security.cert.X509Certificate
+import javax.net.ssl.SSLException
+import javax.net.ssl.X509TrustManager
+import kotlin.time.Duration.Companion.seconds
+import kotlin.time.toJavaDuration
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import okhttp3.OkHttpClient
+import okhttp3.Request
 import timber.log.Timber
 
 /*
@@ -27,7 +41,25 @@ import timber.log.Timber
  * we don't want the webview code in the wear app.
  */
 
-open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository) : WebViewClient() {
+// The server is already reachable (WebView obtained the certificate), so the call is
+// mostly a TLS handshake. 3 seconds gives enough margin for slow connections
+// without blocking the WebView handler indefinitely.
+private val tlsValidationTimeout = 3.seconds
+
+private sealed interface TlsValidationResult {
+    data object Trusted : TlsValidationResult
+    data object Untrusted : TlsValidationResult
+    data object NetworkError : TlsValidationResult
+}
+
+open class TLSWebViewClient(
+    private var keyChainRepository: KeyChainRepository,
+    private val validationScope: CoroutineScope,
+    private val trustManager: X509TrustManager,
+    /** Used as a fallback for API < 29, where [SslCertificate.x509Certificate] is unavailable. */
+    okHttpClient: OkHttpClient,
+    @VisibleForTesting internal val ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
+) : WebViewClient() {
     var isTLSClientAuthNeeded = false
         @VisibleForTesting set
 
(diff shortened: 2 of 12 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
```diff
@@ -67,6 +85,8 @@ class HAWebViewClientFactory @Inject constructor(@NamedKeyChain private val keyC
     }
 }
 
+private val TLS_VALIDATION_TIMEOUT = 5.seconds
```
TimoPtr: Why 5s?

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
```diff
@@ -75,6 +95,8 @@ class HAWebViewClientFactory @Inject constructor(@NamedKeyChain private val keyC
  */
 class HAWebViewClient internal constructor(
     keyChainRepository: KeyChainRepository,
+    validationScope: CoroutineScope,
+    private val okHttpClient: OkHttpClient,
```
TimoPtr:
```suggestion
    okHttpClient: OkHttpClient,
```

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
```diff
@@ -83,7 +105,12 @@ class HAWebViewClient internal constructor(
     private val onReceivedHttpAuthRequest: (
         (handler: HttpAuthHandler, host: String, resource: String, realm: String) -> Unit
     )?,
-) : TLSWebViewClient(keyChainRepository) {
+) : TLSWebViewClient(keyChainRepository, validationScope) {
+
+    // callTimeout bounds the entire call (DNS + connect + TLS + transfer), no per-phase overrides needed
+    private val tlsValidationClient: OkHttpClient = okHttpClient.newBuilder()
+        .callTimeout(TLS_VALIDATION_TIMEOUT.toJavaDuration())
+        .build()
```
TimoPtr: I would make the whole tls logic part of the base class by simply passing okhttp to the base class.

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
```diff
@@ -217,9 +244,27 @@ class HAWebViewClient internal constructor(
...
     override fun onReceivedSslError(view: WebView?, handler: SslErrorHandler?, error: SslError?) {
+        Timber.w("onReceivedSslError: primary error ${error?.primaryError} on ${sensitive { error?.url.orEmpty() }}")
         super.onReceivedSslError(view, handler, error)
-        Timber.e("onReceivedSslError: $error")
+    }
+
+    override suspend fun isTlsTrusted(url: String): Boolean = withContext(Dispatchers.IO) {
+        try {
+            val request = Request.Builder().url(url).head().build()
+            tlsValidationClient.newCall(request).execute().use { }
```
TimoPtr: I think there is a way to not make a call, making a call could come with other issues like DNS resolution and so on. In theory it's not an issue otherwise any call to the API/WebSocket would also fail.

I'm wondering if we could use the `X509TrustManager.checkServerTrusted` with the certificate from the SslError. In theory it could fill he gaps without having to make a network call by using the user certificates.

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
```diff
@@ -217,9 +244,27 @@ class HAWebViewClient internal constructor(
     }
 
     override fun onReceivedSslError(view: WebView?, handler: SslErrorHandler?, error: SslError?) {
+        Timber.w("onReceivedSslError: primary error ${error?.primaryError} on ${sensitive { error?.url.orEmpty() }}")
         super.onReceivedSslError(view, handler, error)
-        Timber.e("onReceivedSslError: $error")
+    }
```
TimoPtr: remove this if it's only for logging and keep it in the base class.

cryptomilk:
@TimoPtr thank you very much for the review! I will look into them.

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
cryptomilk: Was just a random pick. Maybe 10 seconds for slow VPNs is better. I will add a comment.

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
cryptomilk: :+1:

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
cryptomilk: :+1:

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
cryptomilk: I will add code for this, but it is only available on API >= 29 so we need a fallback.

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
TimoPtr: 10s seems a lot just for the TLS handcheck.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -37,6 +69,100 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository)
...
+
+    /**
+     * Called when TLS validation could not be completed due to a network error (e.g. timeout,
+     * DNS failure) rather than a genuine certificate rejection.
+     *
+     * The default implementation delegates to [onSslErrorRejected]. Subclasses can override to
+     * surface a connectivity error instead of an SSL error to the user.
+     */
+    open fun onTlsValidationNetworkError(view: WebView?, error: SslError?) {
+        onSslErrorRejected(view, error)
```
TimoPtr:
```suggestion
    open fun onSslErrorRejected(error: SslError?) {}

    /**
     * Called when TLS validation could not be completed due to a network error (e.g. timeout,
     * DNS failure) rather than a genuine certificate rejection.
     *
     * The default implementation delegates to [onSslErrorRejected]. Subclasses can override to
     * surface a connectivity error instead of an SSL error to the user.
     */
    open fun onTlsValidationNetworkError(error: SslError?) {
        onSslErrorRejected(error)
```
No need to forward the view it is not useful, you can then avoid keeping the WeakRef on it.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -37,6 +69,100 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository)
...
+            return
+        }
+        if (error.primaryError == SslError.SSL_UNTRUSTED && error.url != null) {
+            val url = error.url!!
+            val handlerRef = handler?.let { WeakReference(it) }
+            val viewRef = view?.let { WeakReference(it) }
+            validationScope.launch {
+                val result = isTlsTrusted(error.certificate, url)
+                withContext(Dispatchers.Main) {
+                    val h = handlerRef?.get()
```
TimoPtr:
```suggestion
                    val handler = handlerRef?.get()
```
Let's avoid one char variable name.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -27,7 +40,26 @@ import timber.log.Timber
  * we don't want the webview code in the wear app.
  */
 
-open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository) : WebViewClient() {
+// TODO: remove tlsValidationTimeout and tlsValidationClient when min SDK is raised to 29
```
TimoPtr: Remove the TODOs, when we drop a version we look for the `if BUILD` statement and work from there. Here we would probably forget to remove the TODO. Also increasing the min to 29 is not going to happen soon.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
```diff
@@ -37,6 +69,100 @@ open class TLSWebViewClient(private var keyChainRepository: KeyChainRepository)
...
+        }
+    }
+
+    private suspend fun isTlsTrusted(sslCertificate: SslCertificate?, url: String): TlsValidationResult {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            val cert = sslCertificate?.x509Certificate ?: return TlsValidationResult.Untrusted
+            return try {
+                // SslCertificate does not expose the TLS cipher suite, so use "UNKNOWN" as the
+                // auth type. Android trust managers still validate the full chain with this value.
+                trustManager.checkServerTrusted(arrayOf(cert), "UNKNOWN")
```
TimoPtr: Did you test this path on a real device where the issue occur?

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
cryptomilk: Any suggestions? 3?

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
cryptomilk: Done.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
cryptomilk: Will do.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
cryptomilk: I don't have a way to compile it (yet). It probably will take a day to setup a build env with android studio :-(

app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt
cryptomilk: I set it to 3 now.

app/src/main/kotlin/io/homeassistant/companion/android/util/TLSWebViewClient.kt
TimoPtr: FYI we won't be able to merge this if you are not able to properly test it yourself, especially when touching things like TLS. It needs to be manually tested, until that this PR will remain in draft.

cryptomilk: I got everything working.

X509TrustManager.checkServerTrusted only can check the leave and not intermediate certs. If that fails, we need to fall back. I've implemented it and the app works with my HA which uses step-ca for certs.

cryptomilk:
Tests passed locally.
Manually tested on my Fairphone 6 and I can connect to my HA instance.

cryptomilk:
Actually wait, I also tested on my old Zenfone 10 and it fails there. Investigating ...

cryptomilk:
The WebView auth page loads fine (you can see and approve the login), but the subsequent OkHttp POST to /auth/token in NameYourDeviceViewModel.onSaveClick() fails with "Trust anchor not found."

I don't understand why.

cryptomilk:
Ok, also works on the Zenfone 10. I guess it was some caching issue. I reinstalled and the setup worked just fine now.

I think we are good to go.

TimoPtr:
> Ok, also works on the Zenfone 10. I guess it was some caching issue. I reinstalled and the setup worked just fine now.
> 
> I think we are good to go.

Which APIs did you test in the end? Could you give us reproduction steps for us to test too?

cryptomilk:
> > Ok, also works on the Zenfone 10. I guess it was some caching issue. I reinstalled and the setup worked just fine now.
> > I think we are good to go.
> 
> Which APIs did you test in the end? Could you give us reproduction steps for us to test too?

Fairphone 6 with e/OS/: API level 35
Asus Zenfone 10: API level 34

cryptomilk:
I'm using https://smallstep.com/docs/step-ca/ for my CA Authority. I have a certificate for my home server which has Subject Alternative Names (SAN) for all the domain on it including home assistant.

You can also user OpenSSL as an alternative https://openssl-ca.readthedocs.io/en/latest/introduction.html

I have nginx as a proxy which does TLS and proxies to HA. It looks like this:

```
      '005-homeassistant.conf':
        enabled: True
        config:
          - server:
            - listen:
              - '0.0.0.0:443 ssl'
              - '[::]:443 ssl'
            - server_name: {{ ha_hostname }}.{{ grains.domain }}
            - location /:
              - proxy_pass: 'http://localhost:8123'
              - proxy_redirect: 'off'
              - proxy_http_version: '1.1'
              - proxy_set_header: Upgrade $http_upgrade
              - proxy_set_header: Connection "upgrade"
              - proxy_set_header: Host $host
              - proxy_set_header: X-Real-IP $remote_addr
              - proxy_set_header: X-Forwarded-For $proxy_add_x_forwarded_for
              - proxy_set_header: X-Forwarded-Proto $scheme
```

I downloaded the root certificate and installed it in Android.

Settings -> Security & Privacy -> Encryption & Credentials -> Install a certificate -> CA Certificate

Once it is installed the best is to reboot the phone. Then you can open HA Companion and enter the address manually to point to https://homeassistant.example.local

With the patch the certificate + intermediate chain will be checked and if valid the connection is accepted.

TimoPtr:
> I'm using [smallstep.com/docs/step-ca](https://smallstep.com/docs/step-ca/) for my CA Authority. I have a certificate for my home server which has Subject Alternative Names (SAN) for all the domain on it including home assistant.
> 
> You can also user OpenSSL as an alternative [openssl-ca.readthedocs.io/en/latest/introduction.html](https://openssl-ca.readthedocs.io/en/latest/introduction.html)
> 
> I have nginx as a proxy which does TLS and proxies to HA. It looks like this:
> 
> ```
>       '005-homeassistant.conf':
>         enabled: True
>         config:
>           - server:
>             - listen:
>               - '0.0.0.0:443 ssl'
>               - '[::]:443 ssl'
>             - server_name: {{ ha_hostname }}.{{ grains.domain }}
>             - location /:
>               - proxy_pass: 'http://localhost:8123'
>               - proxy_redirect: 'off'
>               - proxy_http_version: '1.1'
>               - proxy_set_header: Upgrade $http_upgrade
>               - proxy_set_header: Connection "upgrade"
>               - proxy_set_header: Host $host
>               - proxy_set_header: X-Real-IP $remote_addr
>               - proxy_set_header: X-Forwarded-For $proxy_add_x_forwarded_for
>               - proxy_set_header: X-Forwarded-Proto $scheme
> ```
> 
> I downloaded the root certificate and installed it in Android.
> 
> Settings -> Security & Privacy -> Encryption & Credentials -> Install a certificate -> CA Certificate
> 
> Once it is installed the best is to reboot the phone. Then you can open HA Companion and enter the address manually to point to [homeassistant.example.local](https://homeassistant.example.local)
> 
> With the patch the certificate + intermediate chain will be checked and if valid the connection is accepted.

It's to setup mTLS no? I do use mTLS but I don't have the issue you are solving in this PR with my setup. Using Let's encrypt or Cloudflare. 

We need a way to replicate the issue you have to validate it.

cryptomilk:
Your setup with Let's Encrypt or Cloudflare works out-of-the-box because those root CAs are already trusted by Android.

This PR: The client (Android app) validates the server's certificate, which is signed by a private CA that is not in Android's public trust store.

The fix uses OkHttp (which does honor user-installed CAs via AndroidCAStore) as a second validation path before rejecting the connection.

TimoPtr:
@cryptomilk I've been trying to reproduce the issue but so far I was not able to do so.

What I did
```
openssl genrsa -out ca.key 4096

openssl req -x509 -new -nodes -key ca.key -sha256 -days 3650 \
    -subj "/CN=ha-test CA/O=ha-test" -out ca.crt

openssl genrsa -out ha.key 2048


  cat > ha.cnf <<'EOF'
  [req]
  distinguished_name = dn
  req_extensions     = req_ext
  prompt             = no
  [dn]
  CN = MY_DOMAIN
  [req_ext]
  subjectAltName = @alt
  [alt]
  DNS.1 = MY_DOMAIN
  EOF


openssl req -new -key ha.key -out ha.csr -config ha.cnf

openssl x509 -req -in ha.csr -CA ca.crt -CAkey ca.key -CAcreateserial \
    -out ha.crt -days 825 -sha256 \
    -extfile ha.cnf -extensions req_ext

cat ha.crt ca.crt > ha-fullchain.crt
```

Use this cert in my Nginx and install it on Android 9 and 17 both failed without the cert installed and once installed it succeed. Could it be that you are using an old WebView?

cryptomilk:
Is there a difference which browser you set as the default browser? I have Firefox as the default and I need this patch. See log entries from the phone in the initial description.

TimoPtr:
> Is there a difference which browser you set as the default browser? I have Firefox as the default and I need this patch. See log entries from the phone in the initial description.

I have chrome can you test with Chrome ? I'm not sure the app use Firefox under the hood but I might be wrong.

cryptomilk:
Ok, on my Asus Zenfone 10 I installed the version from Google Play and it validates the certificate correctly. It runs Stock Asus Android.

On my Fairphone 6 with /e/OS (based on Lineage) it fails to validate the certificate. It says it doesn't trust the CA authority. So I need this patch there to get HA Companion working.

cryptomilk:
It doesn't matter which browser I use as default.

TimoPtr:
> It doesn't matter which browser I use as default.

It seems to be something at the ROM level did you open an issue there? I'm not a fan of adding this to fix an issue in the upstream.

TimoPtr:
After a discussion on discord https://discord.com/channels/330944238910963714/1346948551892009101/1513549686798028801 we decided to close the PR because we want to avoid maintaining an edge case "fix", to something that could be fix in the upstream or in the certificate itself.

cryptomilk:
I've opened https://gitlab.e.foundation/e/backlog/-/work_items/9418

The issue seems to be in SystemWebView.apk

TimoPtr:
> I've opened [gitlab.e.foundation/e/backlog/-/work_items/9418](https://gitlab.e.foundation/e/backlog/-/work_items/9418)
> 
> The issue seems to be in SystemWebView.apk

Thanks for reporting this 👍🏻

PR #6857, opened 2026-05-17
Add error screen no connection on Automotive

### LINKED ISSUE #6846: Infinite loading on no connection in automotive
It has been reported from the Harman Ignite Store that the app keeps loading when there is no internet connection. We should have an error screen like on the app.

cddu33:
## Summary

**Add connection availability monitoring to Automotive**

Fixes https://github.com/home-assistant/android/issues/6846 — users accessing the app through the Harman Ignite Store were stuck on an infinite loading screen when no internet connection was available, with no error feedback (unlike the standard app).

When the app is running in the car and loses its connection to the Home Assistant server, the user is now shown a dedicated "No connection".

A new ConnectionAvailabilityMonitor polls the Home Assistant WebSocket at a healthy 15 second cadence, switching to a faster 1 second cadence when pings start failing. 
To avoid flashing an error screen on transient network blips, the monitor waits for a 10 second grace period before declaring the connection Unavailable; if a ping recovers within that window, no error is surfaced. 
Once the connection is back, the monitor emits Available again at the next successful poll and the error screen is dismissed automatically. 
When no server is registered yet, the monitor reports Available so the onboarding flow is never blocked by this check.

HaCarAppService observes this state and pushes/pops NoConnectionScreen on the Android Auto screen stack accordingly, keeping the rest of the navigation untouched.

The flow logic (grace period, recovery during grace, healthy→degraded transitions, recovery from unavailable) is covered by unit tests in ConnectionAvailabilityMonitorTest.

## Checklist

## Screenshots
image


app/src/main/kotlin/io/homeassistant/companion/android/vehicle/HaCarAppService.kt
```diff
@@ -123,6 +129,21 @@ class HaCarAppService : CarAppService() {
...
             }
+
+            private suspend fun observeConnectionAvailability() {
+                connectionAvailabilityMonitor.observeAvailability().collect { state ->
+                    val screenManager = carContext.getCarService(ScreenManager::class.java)
+                    val topIsNoConnection = screenManager.top is NoConnectionScreen
+                    when (state) {
+                        ConnectionAvailability.Unavailable -> if (!topIsNoConnection) {
+                            screenManager.push(NoConnectionScreen(carContext))
+                        }
```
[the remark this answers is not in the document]
cddu33: Edge case limited to multi-server users who switch from the in-car UI; tracking separately to keep this PR focused on #6846.

app/src/main/kotlin/io/homeassistant/companion/android/vehicle/ConnectionAvailabilityMonitor.kt
```diff
@@ -0,0 +1,88 @@
...
+                true
+            } else {
+                serverManager.webSocketRepository().sendPing()
+            }
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            Timber.w(e, "Failed to ping WebSocket")
+            false
+        }
```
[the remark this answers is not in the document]
cddu33: not a pb

app/src/main/kotlin/io/homeassistant/companion/android/vehicle/ConnectionAvailabilityMonitor.kt
```diff
@@ -0,0 +1,88 @@
+package io.homeassistant.companion.android.vehicle
```
TimoPtr, line 1: Let's not make this specific to vehicle, also it is a public API so I expect to see documentation on the public part of this class (explaining the behavior not the implementation).

app/src/main/kotlin/io/homeassistant/companion/android/vehicle/ConnectionAvailabilityMonitor.kt
```diff
@@ -0,0 +1,88 @@
...
+
+interface ConnectionAvailabilityMonitor {
+    fun observeAvailability(): Flow<ConnectionAvailability>
+}
+
+internal val GRACE_PERIOD: Duration = 10.seconds
+
+internal val HEALTHY_POLL_INTERVAL: Duration = 15.seconds
+
+internal val DEGRADED_POLL_INTERVAL: Duration = 1.seconds
```
TimoPtr, line 36: This probably should only be `@VisibleForTesting`

app/src/main/kotlin/io/homeassistant/companion/android/vehicle/ConnectionAvailabilityMonitor.kt
```diff
@@ -0,0 +1,88 @@
...
+            delay(if (reachable) HEALTHY_POLL_INTERVAL else DEGRADED_POLL_INTERVAL)
+        }
+    }.distinctUntilChanged()
+
+    private suspend fun isHaReachable(): Boolean {
+        return try {
+            if (!serverManager.isRegistered()) {
+                true
+            } else {
+                serverManager.webSocketRepository().sendPing()
```
TimoPtr, line 69: You are only checking the current active server, it needs to be documented, in the class documentation for instance.

app/src/main/kotlin/io/homeassistant/companion/android/vehicle/NoConnectionScreen.kt
```diff
@@ -0,0 +1,43 @@
...
+    override fun onGetTemplate(): Template {
+        val icon = CarIcon.Builder(
+            IconicsDrawable(carContext, CommunityMaterial.Icon3.cmd_wifi_off).apply {
+                sizeDp = 64
+            }.toAndroidIconCompat(),
+        )
+            .setTint(CarColor.DEFAULT)
+            .build()
+
+        return MessageTemplate.Builder(carContext.getString(R.string.error_connection_failed_no_network))
```
TimoPtr, line 33: You are assuming that when we don't have a WebSocket connection we have no network. It might not be true, the server might simply be down. If you check to the new error screen when we are not able to reach out to the server we show a way to test what is currently wrong check `FrontendConnectionErrorStateProvider`

Image

We could replicate this or simply change the message to say that the connection to the server is broken maybe?

@jpelgrom might have another opinion on the wording used.

app/src/test/kotlin/io/homeassistant/companion/android/vehicle/ConnectionAvailabilityMonitorTest.kt
```diff
@@ -0,0 +1,115 @@
...
+import kotlin.time.Duration.Companion.seconds
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.advanceTimeBy
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@ExtendWith(MainDispatcherJUnit5Extension::class, ConsoleLogExtension::class)
```
TimoPtr, line 19:
```suggestion
@ExtendWith(ConsoleLogExtension::class)
```
I don't think you need the extension for the MainDispatcher here.

app/src/test/kotlin/io/homeassistant/companion/android/vehicle/ConnectionAvailabilityMonitorTest.kt
```diff
@@ -0,0 +1,115 @@
...
+class ConnectionAvailabilityMonitorTest {
+
+    private val serverManager: ServerManager = mockk(relaxed = true)
+
+    private fun createMonitor(): ConnectionAvailabilityMonitor = ConnectionAvailabilityMonitorImpl(serverManager)
+
+    private fun stubPings(vararg results: Boolean) {
```
TimoPtr, line 26:
```suggestion
    private fun mockPings(vararg results: Boolean) {
```
A stub is something dumb, here it's real mocking.

app/src/test/kotlin/io/homeassistant/companion/android/vehicle/ConnectionAvailabilityMonitorTest.kt
```diff
@@ -0,0 +1,115 @@
...
+        val webSocketRepository = mockk<WebSocketRepository>()
+        val iterator = results.iterator()
+        var last = results.last()
+        coEvery { webSocketRepository.sendPing() } answers {
```
TimoPtr, line 30: I think you could use `returnsMany` instead of using the iterator.

TimoPtr (commented):
Thanks for taking this bug, we need to check the wording of the error page and after that we are good to merge.

app/src/main/kotlin/io/homeassistant/companion/android/vehicle/NoConnectionScreen.kt
jpelgrom: > We could replicate this or simply change the message to say that the connection to the server is broken maybe?
>
> @jpelgrom might have another opinion on the wording used.

Replicating the entire screen feels like a bit much for a car, but a general header "unable to connect" and a short sentence describing the reason, like the main app, would be nice I think. We also don't have to discuss the wording in that case ;)

TimoPtr: Agree that we don't need to replicate, but the message and the logo should not say that there are no network available that's potentially not the issue, so it should be more generic.

TimoPtr:
@cddu33 will you be able to look at my last comments? @jpelgrom if you can just give us a hand on the wording 🙏🏻

cddu33:
Hi, i'not has the time to do any thing about it

jpelgrom:
> @jpelgrom if you can just give us a hand on the wording 🙏🏻

Catching up on open PR reviews now, I realize I never properly replied to this, sorry. After the comment on May 18 I added a 👍 to https://github.com/home-assistant/android/pull/6857#discussion_r3261119204. I think replicating the messaging used in the main app, or at least the general error message, is the best path forward.

app/src/main/kotlin/io/homeassistant/companion/android/vehicle/ConnectionAvailabilityMonitor.kt
jpelgrom: That's actually a larger issue - the vehicle interface can show results for a different server but that doesn't change the active server. It should take into account which server is currently set in HaCarAppService.

TimoPtr:
I've addressed the last comments batch. We can go further in this direction but let's merge this PR as a first step.

jpelgrom:
Seems to work well from a technical point of view. However, testing with the desktop head unit, I noticed one potential issue:

image

The template has a [cap of 2 lines by design](https://developer.android.com/design/ui/cars/guides/templates/message-template). Here it barely is readable, but it's easy to imagine other languages where critical information may be lost. Maybe we should only use new messages, very short, after all? Suggestion for this one would be simply: check your connection.

jpelgrom (approved):

PR #6864, opened 2026-05-19
Use workflow_run to trigger emulator.wtf tests

TimoPtr:
## Summary

Emulator.wtf currently can't run on PRs from forks. The authentication uses OIDC, which would require allowing anyone with a fork to authenticate against our Emulator.wtf account  and we only want our own repo to be able to do that.
  
  This PR adds a new `PR Emulator.wtf` workflow that is triggered by `workflow_run` once the main `Pull Request` workflow completes. Because workflow_run-triggered workflows always execute in the base-repo context (regardless of whether the PR comes from a fork), it can use OIDC credentials safely. The new workflow runs the Emulator.wtf jobs and takes over the test-results reporting that the original publish_test_results job used to handle.

### DIFF
--- a/.github/workflows/pr-emulator-wtf.yml
+++ b/.github/workflows/pr-emulator-wtf.yml
@@ -5,6 +5,10 @@ on: # yamllint disable-line rule:truthy
     workflows: ['Pull Request']
     types: [completed]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
 permissions: {}
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -340,42 +337,27 @@ jobs:
         run: ./gradlew :app:assembleDebug :app:assembleAndroidTest -Pabis=x86,x86_64
 
       - name: Build device list
```
jpelgrom, line 339: Why not move this step to the new workflow? There is no actual building happening in this step, and then we have less upload/download and untrusted input to deal with.

TimoPtr: It allow us to build the list of devices in a PR, otherwise we need to make a first PR to update the list since we execute the code from main not from the PR.

jpelgrom: You could use actions/checkout with a reference to the PR branch, no? That also works for forks based on what I can find. We don't actually use the files in the new workflow so no conflict.

Snippet I found, ref/repository should be changed to use `github.event.workflow_run`:

```yaml
- name: Clone repository
  uses: actions/checkout
  with:
    fetch-depth: 0
    ref: ${{github.event.pull_request.head.ref}}
    repository: ${{github.event.pull_request.head.repo.full_name}}
```

.github/workflows/pr-emulator-wtf.yml
```diff
@@ -0,0 +1,137 @@
...
+        run: |
+          if [ ! -s emulator-wtf-inputs/devices.txt ]; then
+            echo "::error::devices.txt is empty or missing"
+            exit 1
+          fi
+          while IFS= read -r line; do
+            if [[ ! "$line" =~ ^model=Pixel7,version=(2[3-9]|3[0-9]|40)$ ]]; then
+              echo "::error::Invalid line in devices.txt"
+              exit 1
+            fi
```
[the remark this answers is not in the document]
TimoPtr: @jpelgrom I didn't know what to do there that's why I kept an acceptable range [23-40]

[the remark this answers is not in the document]
jpelgrom: I think it's fine for now, we have some time until reached and it's better than nothing

.github/workflows/pr.yml
jpelgrom: I still think this could work if moved and is a cleaner approach. You've already added the number validation in the other workflow, not too different from validating what is in the version file...

TimoPtr: What do you mean by `not too different from validating what is in the version file`. I don't see the benefits of checking out in the new workflow, it add one checkout compare to pulling one file. I might change my mind once we introduce e2e test.

jpelgrom (approved):
[On Discord](https://discord.com/channels/330944238910963714/1346948551892009101/1508801280854720554) you mentioned "I’m in favor of keeping without checkout for now" - I think the other way around but it's not a big deal and unblocking third party PRs is more important. Agree to disagree ;)

Obviously not tested, but the setup seems sound - I wonder how a failure comment will look. Needs adjustment in required tests to be merged.

PR #6866, opened 2026-05-19
Request ACCESS_LOCAL_NETWORK when needed

TimoPtr:
## Summary

Add support for the new [ACCESS_LOCAL_NETWORK](https://developer.android.com/reference/android/Manifest.permission#ACCESS_LOCAL_NETWORK) permission. Add to the current flow request to access the local network, the logic is kept dumb without checking the server it is connecting to to keep it simple.

Added to
- onboarding when reaching server discovery (permission with a blank screen behind) if rejected then it redirect to manual server
- WebViewActivity when opening the app it is going to show the error but on top require the permission (it's best effort since we want to drop it)
- FrontendScreen properly ask for permission while the UI is still loading and if rejected proper error screen

## Checklist

## Any other notes

I've ignored a test because Robolectric is not yet ready for API 37. 

Since we are not targetting yet API 37 we might have issues if someone discard the permission from the settings manually but that should probably not happen and we can manually help on discord or github or the user will simply enable it back. 

I've bump the target to 37 and no issue but I don't want yet to make this jump because we need to test more any regression based on the changes made by Google.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -6,6 +6,7 @@
 
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+    <uses-permission android:name="android.permission.ACCESS_LOCAL_NETWORK" />
     <uses-permission android:name="android.permission.CHANGE_WIFI_MULTICAST_STATE" />
     <uses-permission android:name="android.permission.BLUETOOTH"
         android:maxSdkVersion="30"/>
--- a/app/src/test/kotlin/io/homeassistant/companion/android/frontend/permissions/PermissionManagerTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/frontend/permissions/PermissionManagerTest.kt
@@ -533,6 +533,62 @@ class PermissionManagerTest {
 
     // endregion
 
+    // region Local network permission (Android 17+)
+
+    @Nested
+    inner class CheckLocalNetworkPermission {
+
+        @Test
+        fun `Given pre-API 37 then returns true without checking permission`() = runTest {
+            val manager = createManager(sdkInt = Build.VERSION_CODES.CINNAMON_BUN - 1)
+            assertTrue(manager.checkLocalNetworkPermission())
+            assertNull(manager.pendingPermissionRequest.value)
+        }
+
+        @Test
+        fun `Given API 37 when permission already granted then returns true`() = runTest {
+            every { permissionChecker.hasPermission(android.Manifest.permission.ACCESS_LOCAL_NETWORK) } returns true
+
+            val manager = createManager(sdkInt = Build.VERSION_CODES.CINNAMON_BUN)
+            assertTrue(manager.checkLocalNetworkPermission())
+            assertNull(manager.pendingPermissionRequest.value)
+        }
+
+        @Test
+        fun `Given API 37 when permission not granted then emits pending request`() = runTest {
+            every { permissionChecker.hasPermission(android.Manifest.permission.ACCESS_LOCAL_NETWORK) } returns false
+
+            val manager = createManager(sdkInt = Build.VERSION_CODES.CINNAMON_BUN)
+            val result = async { manager.checkLocalNetworkPermission() }
+            advanceUntilIdle()
+
+            val pending = manager.pendingPermissionRequest.value
+            assertInstanceOf(PermissionRequest.LocalNetwork::class.java, pending)
+            assertEquals(listOf(android.Manifest.permission.ACCESS_LOCAL_NETWORK), pending?.permissions)
+
+            (pending as PermissionRequest.LocalNetwork).onResult(true)
+            advanceUntilIdle()
+            assertTrue(result.await())
+            assertNull(manager.pendingPermissionRequest.value)
+        }
+
+        @Test
+        fun `Given API 37 when user denies then returns false`() = runTest {
+            every { permissionChecker.hasPermission(android.Manifest.permission.ACCESS_LOCAL_NETWORK) } returns false
+
+            val manager = createManager(sdkInt = Build.VERSION_CODES.CINNAMON_BUN)
+            val result = async { manager.checkLocalNetworkPermission() }
+            advanceUntilIdle()
+            (manager.pendingPermissionRequest.value as PermissionRequest.LocalNetwork).onResult(false)
+            advanceUntilIdle()
+
+            assertFalse(result.await())
+            assertNull(manager.pendingPermissionRequest.value)
+        }
+    }
+
+    // endregion
+
     // region Guard against concurrent requests
 
     @Nested
--- a/app/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryViewModelTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryViewModelTest.kt
@@ -45,6 +45,37 @@ private class ServerDiscoveryViewModelTest {
 
     private fun createViewModel(discoveryMode: ServerDiscoveryMode = ServerDiscoveryMode.NORMAL) {
         viewModel = ServerDiscoveryViewModel(discoveryMode, searcher, serverManager)
+        // Existing tests assert behavior that depends on discovery having started; mirror the
+        // screen's permission-checked callback so they don't all have to opt in individually.
+        viewModel.onLocalNetworkPermissionChecked()
+    }
+
+    @Test
+    fun `Given local network permission not yet checked then no server discovery is started and NoServerFound is not emitted`() = runTest {
+        viewModel = ServerDiscoveryViewModel(ServerDiscoveryMode.NORMAL, searcher, serverManager)
+
+        viewModel.discoveryFlow.test {
+            assertEquals(Started, awaitItem())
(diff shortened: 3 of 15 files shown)


app/src/test/kotlin/io/homeassistant/companion/android/frontend/haptic/HapticFeedbackPerformerTest.kt
jpelgrom, line 1: These changes are unrelated

app/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,7 @@
 
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+    <uses-permission android:name="android.permission.ACCESS_LOCAL_NETWORK" />
```
jpelgrom, line 9: Must also be added in `automotive`

app/src/test/kotlin/io/homeassistant/companion/android/frontend/permissions/PermissionManagerTest.kt
```diff
@@ -533,6 +533,62 @@ class PermissionManagerTest {
 
     // endregion
 
+    // region Local network permission (Android 17+)
+
+    @Nested
+    inner class CheckLocalNetworkPermission {
+
+        @Test
+        fun `Given pre-CINNAMON_BUN device then returns true without checking permission`() = runTest {
```
jpelgrom: We don't use code names anywhere else (other than the constant), update for consistency

```suggestion
        fun `Given pre-API 37 then returns true without checking permission`() = runTest {
```

app/src/test/kotlin/io/homeassistant/companion/android/frontend/permissions/PermissionManagerTest.kt
```diff
@@ -533,6 +533,62 @@ class PermissionManagerTest {
...
+            val manager = createManager(sdkInt = Build.VERSION_CODES.CINNAMON_BUN - 1)
+            assertTrue(manager.checkLocalNetworkPermission())
+            assertNull(manager.pendingPermissionRequest.value)
+        }
+
+        @Test
+        fun `Given CINNAMON_BUN+ device when permission already granted then returns true`() = runTest {
```
jpelgrom:
```suggestion
        fun `Given API 37 when permission already granted then returns true`() = runTest {
```

app/src/test/kotlin/io/homeassistant/companion/android/frontend/permissions/PermissionManagerTest.kt
```diff
@@ -533,6 +533,62 @@ class PermissionManagerTest {
...
+            every { permissionChecker.hasPermission(android.Manifest.permission.ACCESS_LOCAL_NETWORK) } returns true
+
+            val manager = createManager(sdkInt = Build.VERSION_CODES.CINNAMON_BUN)
+            assertTrue(manager.checkLocalNetworkPermission())
+            assertNull(manager.pendingPermissionRequest.value)
+        }
+
+        @Test
+        fun `Given CINNAMON_BUN+ device when permission not granted then emits pending request`() = runTest {
```
jpelgrom:
```suggestion
        fun `Given API 37 when permission not granted then emits pending request`() = runTest {
```

app/src/test/kotlin/io/homeassistant/companion/android/frontend/permissions/PermissionManagerTest.kt
```diff
@@ -533,6 +533,62 @@ class PermissionManagerTest {
...
+            assertEquals(listOf(android.Manifest.permission.ACCESS_LOCAL_NETWORK), pending?.permissions)
+
+            (pending as PermissionRequest.LocalNetwork).onResult(true)
+            advanceUntilIdle()
+            assertTrue(result.await())
+            assertNull(manager.pendingPermissionRequest.value)
+        }
+
+        @Test
+        fun `Given CINNAMON_BUN+ device when user denies then returns false`() = runTest {
```
jpelgrom:
```suggestion
        fun `Given API 37 device when user denies then returns false`() = runTest {
```

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/navigation/ServerDiscoveryNavigationTest.kt
```diff
@@ -117,6 +120,31 @@ internal class ServerDiscoveryNavigationTest : BaseOnboardingNavigationTest() {
         }
     }
 
+    /**
+     * On Android 17+ (API 37, CINNAMON_BUN), entering the discovery screen requests
```
jpelgrom: One name for the Android version is enough

```suggestion
     * On Android 17+, entering the discovery screen requests
```

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryViewModelTest.kt
```diff
@@ -45,6 +45,37 @@ private class ServerDiscoveryViewModelTest {
 
     private fun createViewModel(discoveryMode: ServerDiscoveryMode = ServerDiscoveryMode.NORMAL) {
         viewModel = ServerDiscoveryViewModel(discoveryMode, searcher, serverManager)
+        // Existing tests assert behavior that depends on discovery having started; mirror the
+        // screen's permission-checked callback so they don't all have to opt in individually.
+        viewModel.onLocalNetworkPermissionChecked()
+    }
+
+    @Test
+    fun `Given permission not yet checked then no server discovery is started and NoServerFound is not emitted`() = runTest {
```
jpelgrom:
```suggestion
    fun `Given local network permission not yet checked then no server discovery is started and NoServerFound is not emitted`() = runTest {
```

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/serverdiscovery/ServerDiscoveryViewModelTest.kt
```diff
@@ -45,6 +45,37 @@ private class ServerDiscoveryViewModelTest {
...
+            assertEquals(Started, awaitItem())
+            advanceTimeBy(TIMEOUT_NO_SERVER_FOUND * 2)
+            runCurrent()
+            // Discovery hasn't started, so the no-server timeout never fires
+            expectNoEvents()
+        }
+    }
+
+    @Test
+    fun `Given permission checked twice then discovery is only started once`() = runTest {
```
jpelgrom:
```suggestion
    fun `Given local network permission checked twice then discovery is only started once`() = runTest {
```

jpelgrom (commented):
Tests in an emulator look promising, I like how minimal the changes required are and how you handled them with the existing setup.

I don't have a good way to test the upgrade path other than manually installing an older version first with a remote URL, then installing this version and changing the URL to local. Of course this change is for behavior which shouldn't happen until targeting even though it does, so I'm fine with keeping this as is for now and testing more when we change the target SDK.

Questions:
- Could you add a note to the changelog like "Added support/fixed local network access on Android 17"? I think it's worth calling out, if only for users wondering why they get the pop-up.
- Worth considering: do we want to opt-in on Android 16 with the 'temporary' permission? That should allow us to discover edge cases sooner.

app/src/test/kotlin/io/homeassistant/companion/android/frontend/haptic/HapticFeedbackPerformerTest.kt
TimoPtr: It's because of the compile API change that breaks this tests because the function can tak different type of parameters now.

jpelgrom: OK... Do the compile API change separately if this is the only change required? 😇

TimoPtr: It requires a first PR just to fix the test, I found it a bit annoying but I can if you approve this PR tomorrow I'll split it in 2 and do the merge without reviews.

jpelgrom:
> Worth considering: do we want to opt-in on Android 16 with the 'temporary' permission? That should allow us to discover edge cases sooner.

Shortly discussed [on Discord](https://discord.com/channels/330944238910963714/1346948551892009101/1506367366106320991) with @TimoPtr, and I also remembered that it requires [adb opt-in](https://developer.android.com/privacy-and-security/local-network-permission#android-16-guidance), so not that useful for now and we decided to skip.

jpelgrom:
> Could you add a note to the changelog like "Added support/fixed local network access on Android 17"? I think it's worth calling out, if only for users wondering why they get the pop-up.

I think we still need this

TimoPtr:
> > Could you add a note to the changelog like "Added support/fixed local network access on Android 17"? I think it's worth calling out, if only for users wondering why they get the pop-up.
> 
> I think we still need this

I've added it only to the app since it is not relevant today for the rest IMO

jpelgrom (approved):

app/src/test/kotlin/io/homeassistant/companion/android/frontend/haptic/HapticFeedbackPerformerTest.kt
TimoPtr: Done #6871

PR #6869, opened 2026-05-20
Introduce screen orientation lock feature to FrontendScreen

TimoPtr:
## Summary

Add the fixed screen orientation to the FrontendScreen, exposes the settings with an enum and adjust the old code to use the new enum from the preference.

## Checklist

## Any other notes

I know the string are in xml still but I prefer to duplicate it in one place and avoid using string everywhere and favor strong types. I suspect that we might drop the preference settings at some point in favor of something custom that would drop the duplication and only rely on the enum.

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
@@ -1,11 +1,32 @@
 package io.homeassistant.companion.android.common.data.prefs
 
+import android.content.pm.ActivityInfo
 import android.os.Parcelable
 import io.homeassistant.companion.android.common.data.integration.ControlsAuthRequiredSetting
 import io.homeassistant.companion.android.common.util.GestureAction
 import io.homeassistant.companion.android.common.util.HAGesture
 import kotlinx.coroutines.flow.Flow
 import kotlinx.parcelize.Parcelize
+
+/**
+ * Screen orientation preference applied to the dashboard host activity.
+ *
+ * The [storageValue]s match the entries declared in the `pref_screen_orientation_option_values`
+ * string-array used by the settings ListPreference, so values written by the legacy settings UI
+ * still resolve to a typed enum here.
+ */
+enum class ScreenOrientation(val storageValue: String, val activityInfo: Int) {
+    SYSTEM("system", ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED),
+    PORTRAIT("portrait", ActivityInfo.SCREEN_ORIENTATION_PORTRAIT),
+    LANDSCAPE("landscape", ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE),
+    ;
+
+    companion object {
+        /** Returns the matching entry or [SYSTEM] when [value] is null or unknown. */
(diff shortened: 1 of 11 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
```diff
@@ -1,12 +1,33 @@
...
+ * Screen orientation preference applied to the dashboard host activity.
+ *
+ * The [storageValue]s match the entries declared in the `pref_screen_orientation_option_values`
+ * string-array used by the settings ListPreference, so values written by the legacy settings UI
+ * still resolve to a typed enum here.
+ */
+enum class ScreenOrientation(val storageValue: String, val activityInfo: Int) {
+    SYSTEM("system", ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED),
+    PORTRAIT("portrait", ActivityInfo.SCREEN_ORIENTATION_PORTRAIT),
+    LANDSCAPE("landscape", ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE),
```
jpelgrom, line 21: Is it worth using the string reference here instead of a raw string, so if we ever change it in the xml the setting adjusts accordingly?

@dshokouhi double checking: do you know if the preference value was ever accidentally translated and this might reset it for non-English users? (It currently isn't included in the translatable strings for Dutch, but I cannot easily check every language, and know Lokalise doesn't automatically delete old values.)

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
```diff
@@ -82,9 +103,16 @@ interface PrefsRepository {
...
     suspend fun setKeepScreenOnEnabled(enabled: Boolean)
 
-    suspend fun getScreenOrientation(): String?
+    /**
+     * Returns the user's current screen orientation preference. Falls back to
+     * [ScreenOrientation.SYSTEM] when no value is stored or the stored value cannot be resolved.
+     */
+    suspend fun getScreenOrientation(): ScreenOrientation
+
+    suspend fun saveScreenOrientation(orientation: ScreenOrientation)
```
jpelgrom: If adjusting anyway this is the chance to follow the naming pattern

```suggestion
    suspend fun setScreenOrientation(orientation: ScreenOrientation)
```

jpelgrom (commented):
It works, but I really wish you had split the refactor into an enum off into a smaller standalone PR.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
TimoPtr: If we use the reference here I need to unwrap the reference for the storage and I need to add a context to the PrefRepositoryImpl (it's possible but annoying). 

Another reason to migrate away from the XML pref would allow us better translations for this.

jpelgrom: > Another reason to migrate away from the XML pref would allow us better translations for this.

I don't follow; the keys used for storage are not the same as UI strings.

jpelgrom (approved):

PR #6872, opened 2026-05-20
Add incoming external bus messages for Matter/Thread

TimoPtr:
## Summary

Add incoming external bus messages for matter and mapping into the ViewModel with dumb logs to prepare the impl of the matter onboarding flow.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
@@ -625,6 +625,13 @@ internal class FrontendViewModel @VisibleForTesting constructor(
                 result.event?.let { _events.tryEmit(it) }
             }
 
+            is FrontendHandlerEvent.StartMatterCommissioning,
+            is FrontendHandlerEvent.ImportThreadCredentials,
+            -> {
+                // Matter/Thread handling lands in a follow-up PR
+                Timber.d("Matter/Thread event received but not yet handled: $result")
+            }
+
             is FrontendHandlerEvent.ConfigSent,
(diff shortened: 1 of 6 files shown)


jpelgrom:
After reviewing this and the bar code scanner PR, I guess my main question is how descriptive we want to be. I think I would summarize it as:
 - in incoming messages, it should only describe what the frontends wants the app to do and what it wants in return, not how the app does it internally or when it is sent unless it is part of a flow or only ever sent from one location
 - in frontend events, it should describe what the app should do and maybe a surface-level description of the implementation, but no specifics further than what is called from the frontend screens
 - in outgoing messages, it should only describe the data, not how it is collected as that is up to the part of the app that collects it

With that in mind, I think maybe the mention of Play Services should be dropped from the incoming messages. The frontend doesn't really care where it is coming from. The flow is only started from one place so describing when it is sent is OK here.

TimoPtr:
> After reviewing this and the bar code scanner PR, I guess my main question is how descriptive we want to be. I think I would summarize it as:
> 
>     * in incoming messages, it should only describe what the frontends wants the app to do and what it wants in return, not how the app does it internally or when it is sent unless it is part of a flow or only ever sent from one location
> 
>     * in frontend events, it should describe what the app should do and maybe a surface-level description of the implementation, but no specifics further than what is called from the frontend screens
> 
>     * in outgoing messages, it should only describe the data, not how it is collected as that is up to the part of the app that collects it
> 
> 
> With that in mind, I think maybe the mention of Play Services should be dropped from the incoming messages. The frontend doesn't really care where it is coming from. The flow is only started from one place so describing when it is sent is OK here.

I'm going to merge this PR and actually make another one where I rework all of the doc based on your suggestions.

PR #6873, opened 2026-05-20
Use GH instead of softprops/action-gh-release

TimoPtr:
## Summary

We should avoid using an external action when we can do the same operation with the tool already available it limits our attack surface. It is also one less thing to update.

This address two warning from zizmor

```
info[template-injection]: code injection via template expansion
   --> ./.github/workflows/onPush.yml:106:34
    |
106 |         run: gh release edit ${{ steps.rel_number.outputs.version }} --draft=false
    |         --- this run block       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ may expand into attacker-controllable code
    |
    = note: audit confidence → Low
    = note: this finding has an auto-fix

info[superfluous-actions]: action functionality is already included by the runner
  --> ./.github/workflows/onPush.yml:88:15
   |
86 |       - name: Create draft Github Pre-Release
   |         ------------------------------------- this step
87 |         if: github.event.inputs.beta == 'true'
88 |         uses: softprops/action-gh-release@b4309332981a82ec1c5618f44dd2e27cc8bfbfda # v3.0.0
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ use `gh release` in a script step
   |
   = note: audit confidence → High
```

### DIFF
--- a/.github/workflows/onPush.yml
+++ b/.github/workflows/onPush.yml
@@ -83,7 +83,7 @@ jobs:
         run: |
           echo $VERSION_CODE > ./app/build/outputs/version_code.txt
 
-      - name: Create or update GitHub Pre-Release
+      - name: Create draft GitHub Pre-Release
         if: github.event.inputs.beta == 'true'
         shell: bash
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/onPush.yml
```diff
@@ -100,12 +101,20 @@ jobs:
...
-      - name: Publish Pre-Release
-        if: github.event.inputs.beta == 'true'
-        run: gh release edit ${{ steps.rel_number.outputs.version }} --draft=false
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          )
+
+          # If a release already exists for this tag, update its assets and release metadata.
+          if gh release view "$REL_VERSION" >/dev/null 2>&1; then
+            gh release upload "$REL_VERSION" "${assets[@]}" --clobber
```
jpelgrom: Our prereleases are immutable, won't attaching/uploading assets again fail?

jpelgrom (commented):
I get where it's coming from, but this action is basically endorsed by GitHub in their deprecation of actions/create-release 🤔

.github/workflows/onPush.yml
TimoPtr: Good call, made the change 👍🏻

jpelgrom (approved):

PR #6875, opened 2026-05-21
Use TestExecutionListener for ConsoleLog and FailFast

TimoPtr:
## Summary

Multiple times we had crash of the test engine because of FailFast, it is happening when FailFast is trigger without a proper handler for test set up using the Junit rule/extension. This PR replace the need of a rule/extension by using a `TestExecutionListener` https://docs.junit.org/5.0.3/api/org/junit/platform/launcher/TestExecutionListener.html that works for both JUnit jupiter and JUnit 4 (vintage) that is apply to all our tests. I've decided also to do the same for the ConsoleLog.

The most important part of the PR is in the first commit the second one is simply removing all the usages of the rules/extensions.

We found out a timing issue where if FailFast is triggered in `HomeAssistantApplication` it is using the `DefaultFailFastHandler` and crashing the app since the handler is set after the `onCreate` of the application is called. This PR address this by settings the handler at the runner level (duplicated between :app and :wear), so now an issue will only fail the test and not crash the app allowing better reporting.

You can see the issue https://emulator.wtf/o/a09ee3e1-05c8-4736-8790-c62326837a36/r/30b31c99-a72a-4904-b9f8-f52317c69209/j/79fc70b1-de09-429c-aadf-93bf2f0967a1?key=moYsRl9GO8Lttjtg6pebEDSjL89-Ne4USvXbmK4L

## Checklist


jpelgrom (approved):

jpelgrom:
Thanks for the improvement! Caught me once so nice to have this 'automatically' now, lowering the barrier.

PR #6877, opened 2026-05-21
Add strict mode ignore rule for wrong context usage for WV debug

TimoPtr:
## Summary

We found out a Strict Mode issue while calling the WebView debug flag using a non UI context while running the test on Emulator.wtf this PR add an Ignore rule there.

You can see the issue https://emulator.wtf/o/a09ee3e1-05c8-4736-8790-c62326837a36/r/30b31c99-a72a-4904-b9f8-f52317c69209/j/79fc70b1-de09-429c-aadf-93bf2f0967a1?key=moYsRl9GO8Lttjtg6pebEDSjL89-Ne4USvXbmK4L


jpelgrom:
I think you accidentally included this in #6875 - only the strict mode rule is new here. Good to merge IMO after a rebase and rename.

PR #6878, opened 2026-05-21
Add notification when background need local network permission

TimoPtr:
## Summary

Based on #6874, this add a notification when the local network permission is not granted but needed when something from the background is going to try to make a call like update the sensors. It tries to follow the same design as the one for location.

## Checklist

## Screenshots

image

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/util/AppNotifChannels.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/util/AppNotifChannels.kt
@@ -12,6 +12,7 @@ const val CHANNEL_WEBSOCKET = "Websocket"
 const val CHANNEL_WEBSOCKET_ISSUES = "Websocket Issues"
 const val CHANNEL_HIGH_ACCURACY = "High accuracy location"
 const val CHANNEL_DATABASE = "App Database"
+const val CHANNEL_LOCAL_NETWORK_PERMISSION = "LocalNetworkPermission"
 const val CHANNEL_LOCATION_DISABLED = "Location disabled"
 const val CHANNEL_DOWNLOADS = "downloads"
 const val CHANNEL_GENERAL = "general"
@@ -33,6 +34,7 @@ val appCreatedChannels = listOf(
     CHANNEL_WEBSOCKET_ISSUES,
     CHANNEL_HIGH_ACCURACY,
     CHANNEL_DATABASE,
+    CHANNEL_LOCAL_NETWORK_PERMISSION,
     CHANNEL_LOCATION_DISABLED,
     CHANNEL_DOWNLOADS,
     CHANNEL_GENERAL,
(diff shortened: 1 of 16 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/util/AppNotifChannels.kt
```diff
@@ -33,6 +34,7 @@ val appCreatedChannels = listOf(
     CHANNEL_WEBSOCKET_ISSUES,
     CHANNEL_HIGH_ACCURACY,
     CHANNEL_DATABASE,
+    CHANNEL_LOCAL_NETWORK_PERMISSION,
     CHANNEL_LOCATION_DISABLED,
```
TimoPtr, line 38: @jpelgrom I was wondering if it was not a good a idea to join the two permission related channel into one? To free a slot, I see some crahes on Sentry about too many channels. I didn't check the limit TBH, it could be a user having a crazy amount of channel but still.

jpelgrom: The notification channel [limit](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/services/core/java/com/android/server/notification/PreferencesHelper.java#140) is 5000, as far as I can find, so I think it is a user doing weird things.

I'm fine with merging the channels but it needs to be renamed in that case.

common/src/main/res/values/strings.xml
```diff
@@ -392,6 +392,10 @@
     <string name="template_widgets">Template widgets</string>
     <string name="loading_entities">Please wait while we load your entities</string>
     <string name="loading">Loading…</string>
+    <string name="local_network_permission_message">Home Assistant needs the local network permission to reach %1$s. Background updates have been paused. Tap to grant access.</string>
+    <string name="local_network_permission_short_message">Home Assistant needs the local network permission for background updates</string>
+    <string name="local_network_permission_title">Local network access required</string>
+    <string name="local_network_permission_warn_channel">Local network permission warning</string>
```
jpelgrom: Android shows this as the 'nearby devices' group, so users might not recognize the permission you're referring to. Maybe simply "... needs permission to access your local network."

common/src/main/kotlin/io/homeassistant/companion/android/common/util/AppNotifChannels.kt
TimoPtr: Or a weird ROM changing the value... Let's not rename it otherwise we might break some custom configuration of a user for now.

common/src/main/res/values/strings.xml
TimoPtr: Updated if you can double check, to see if it fits better.

jpelgrom: Yes that seems better to me. Let's try with this and iterate if users say it is unclear.

jpelgrom (approved):

PR #6881, opened 2026-05-21
Fix file uploads in onboarding WebView (e.g. backup restore)

agners:
## Summary

The onboarding WebView in `ConnectionScreen` only had a `WebViewClient` configured and no `WebChromeClient`. Because the default `WebChromeClient.onShowFileChooser()` returns `false`, every `<input type=\"file\">` click in the HA onboarding frontend was silently dropped — including the backup restore upload, where clicking the file picker did nothing at all.

Fix: reuse the existing `FileChooserManager` + `FileChooserEffect` + `HAWebChromeClient` plumbing that was added for the new `FrontendScreen`. The onboarding WebView now uses the same code path:

- `ConnectionViewModel` injects `FileChooserManager`, exposes `pendingFileChooser`, and builds an `HAWebChromeClient` whose `onShowFileChooser` delegates to `fileChooserManager.pickFiles(...)`.
- `ConnectionScreen` collects `pendingFileChooser`, wires the chrome client into `HAWebView`, and renders `FileChooserEffect` so the system picker is actually launched and results are delivered back to the WebView.

Tested manually only insofar as the unit + screen tests pass; I have not yet driven a full onboarding flow on a device to confirm a `.tar` backup uploads end-to-end (the picker now opens, but Android's picker can report `.tar` as `application/octet-stream`, which the HA frontend rejects client-side — that's a separate issue if it bites).

## Checklist

Notes on the checklist:
- Tests: the existing `ConnectionViewModelTest` instantiations were updated to pass the new `FileChooserManager` dependency. No dedicated new test was added for the chooser path itself — the wiring is identical to `FrontendScreen` which is already covered.
- Thorough testing: only unit/screen tests run; no on-device verification of an actual backup upload yet (see Summary).

## Screenshots

N/A — behavior fix; no visual change to native UI.

## Any other notes

The file chooser code path is shared with the (debug-only) `FrontendScreen` via `FileChooserManager`/`FileChooserEffect`. `ConnectionScreen` is not on the `USE_FRONTEND_V2` migration path — it's a separate WebView used during onboarding — so this fix is needed independently.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionScreen.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionScreen.kt
@@ -18,7 +18,6 @@ import androidx.compose.foundation.layout.windowInsetsPadding
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
-import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.vector.ImageVector
@@ -70,10 +69,10 @@ internal fun ConnectionScreen(
     isLoading: Boolean,
     isError: Boolean,
     webViewClient: WebViewClient,
+    webChromeClient: WebChromeClient,
     onBackClick: () -> Unit,
     onWebViewCreationFailed: (Throwable) -> Unit,
     modifier: Modifier = Modifier,
-    webChromeClient: WebChromeClient = remember { WebChromeClient() },
     pendingFileChooser: FileChooserRequest? = null,
 ) {
     FileChooserEffect(pendingRequest = pendingFileChooser)
@@ -139,6 +138,7 @@ private fun ConnectionScreenPreview() {
             isLoading = false,
             isError = false,
             webViewClient = WebViewClient(),
+            webChromeClient = WebChromeClient(),
             onBackClick = {},
             onWebViewCreationFailed = {},
             modifier = Modifier.fillMaxSize(),
--- a/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionScreen.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionScreen.kt
@@ -1,5 +1,6 @@
 package io.homeassistant.companion.android.onboarding.connection
 
+import android.webkit.WebChromeClient
 import android.webkit.WebViewClient
 import androidx.annotation.VisibleForTesting
 import androidx.compose.foundation.Image
@@ -17,6 +18,7 @@ import androidx.compose.foundation.layout.windowInsetsPadding
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.vector.ImageVector
@@ -26,6 +28,8 @@ import androidx.compose.ui.unit.dp
 import io.homeassistant.companion.android.R
 import io.homeassistant.companion.android.common.compose.theme.HAThemeForPreview
 import io.homeassistant.companion.android.common.compose.theme.LocalHAColorScheme
+import io.homeassistant.companion.android.frontend.filechooser.FileChooserEffect
+import io.homeassistant.companion.android.frontend.filechooser.FileChooserRequest
 import io.homeassistant.companion.android.loading.LoadingScreen
 import io.homeassistant.companion.android.util.compose.HAPreviews
 import io.homeassistant.companion.android.util.compose.webview.HAWebView
@@ -44,6 +48,7 @@ internal fun ConnectionScreen(onBackClick: () -> Unit, viewModel: ConnectionView
     val url by viewModel.urlFlow.collectAsState()
     val isLoading by viewModel.isLoadingFlow.collectAsState()
     val error by viewModel.errorFlow.collectAsState()
+    val pendingFileChooser by viewModel.pendingFileChooser.collectAsState()
     val isError = error != null
 
     ConnectionScreen(
@@ -51,6 +56,8 @@ internal fun ConnectionScreen(onBackClick: () -> Unit, viewModel: ConnectionView
         isLoading = isLoading,
         isError = isError,
         webViewClient = viewModel.webViewClient,
+        webChromeClient = viewModel.webChromeClient,
+        pendingFileChooser = pendingFileChooser,
         onBackClick = onBackClick,
         onWebViewCreationFailed = viewModel::onWebViewCreationFailed,
         modifier = modifier,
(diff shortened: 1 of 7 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionScreen.kt
```diff
@@ -66,7 +73,11 @@ internal fun ConnectionScreen(
     onBackClick: () -> Unit,
     onWebViewCreationFailed: (Throwable) -> Unit,
     modifier: Modifier = Modifier,
+    webChromeClient: WebChromeClient = remember { WebChromeClient() },
```
TimoPtr:
```suggestion
    webChromeClient: WebChromeClient,
    modifier: Modifier = Modifier,
```
That's dangerous to do here, let's make it a mandatory field

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
```diff
@@ -114,6 +125,26 @@ internal class ConnectionViewModel @VisibleForTesting constructor(
...
     )
 
+    /**
+     * [WebChromeClient][android.webkit.WebChromeClient] used by the onboarding WebView.
+     *
+     * The Home Assistant onboarding flow (e.g. backup restore) renders `<input type="file">`
+     * elements; without an `onShowFileChooser` override these clicks are silently dropped by
+     * the system WebView. This delegates to [FileChooserManager] so the request survives
+     * configuration changes and reuses the same flow as the main frontend WebView.
+     */
```
TimoPtr, line 132:
```suggestion
    /**
     * [WebChromeClient][android.webkit.WebChromeClient] used by the onboarding WebView.
     *
     * The Home Assistant onboarding flow could request a file selection (e.g. backup restore).
     */
```

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
```diff
@@ -75,6 +76,7 @@ class ConnectionViewModelTest {
         }
     }
     private val connectivityCheckRepository: ConnectivityCheckRepository = mockk()
+    private val fileChooserManager = FileChooserManager()
```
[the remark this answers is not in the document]
TimoPtr: If it works I'm fine about keeping the real impl, we usually mock only if necessary. It's possible that in the FrontendScreen test I mocked it just because it had android API indeed but if it works under unit test I'm ok keeping it.

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
```diff
@@ -397,7 +399,7 @@ class ConnectionViewModelTest {
         val connectivityFlow = MutableSharedFlow<ConnectivityCheckState>()
         every { connectivityCheckRepository.runChecks(rawUrl) } returns connectivityFlow
 
-        val viewModel = ConnectionViewModel(rawUrl, webViewClientFactory, connectivityCheckRepository)
+        val viewModel = ConnectionViewModel(rawUrl, webViewClientFactory, connectivityCheckRepository, fileChooserManager)
         advanceUntilIdle()
```
TimoPtr, line 406: We miss a test that does verify the interaction with the chrome client that properly exposes a`FileChooserRequest`. You can check the `FrontendViewModelTest` to see how to do it.

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
```diff
@@ -75,6 +78,7 @@ class ConnectionViewModelTest {
         }
     }
     private val connectivityCheckRepository: ConnectivityCheckRepository = mockk()
+    private val fileChooserManager = FileChooserManager()
```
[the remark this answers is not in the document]
agners: Copilot says JUnit constructs a fresh ConnectionViewModelTest instance for every test method, so private val fileChooserManager = FileChooserManager() is re-initialized per test. It's not shared. Also other tests like `FrontendViewModelTest` use the pattern.

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
```diff
@@ -114,6 +125,33 @@ internal class ConnectionViewModel @VisibleForTesting constructor(
...
+                try {
+                    val uris = fileChooserManager.pickFiles(fileChooserParams)
+                    delivered = true
+                    filePathCallback.onReceiveValue(uris)
+                } finally {
+                    // Ensure the WebView's file input is unblocked even if the coroutine is
+                    // cancelled (e.g. ViewModel cleared) or pickFiles throws, otherwise the
+                    // `<input type="file">` element would stay in a stuck pending state.
+                    if (!delivered) filePathCallback.onReceiveValue(null)
+                }
```
TimoPtr:
```suggestion
```

If the ViewModel is cleared we should not interact with anything anymore, the WebView is already destroyed (or should be it depends on the GC)

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
```diff
@@ -416,4 +420,80 @@ class ConnectionViewModelTest {
...
+
+        viewModel.webChromeClient.onShowFileChooser(
+            mockk(relaxed = true),
+            filePathCallback,
+            mockk(relaxed = true),
+        )
+        advanceUntilIdle()
+
+        val uris = arrayOf(mockk<Uri>())
+        viewModel.pendingFileChooser.value!!.onResult(uris)
```
TimoPtr:
```suggestion
        val pending = viewModel.pendingFileChooser.value
		assertNotNull(pending)
        pending!!.onResult(uris)
```
It's better in case it fail, otherwise you're going to get a NPE

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
```diff
@@ -416,4 +420,80 @@ class ConnectionViewModelTest {
...
+        val filePathCallback = mockk<ValueCallback<Array<Uri>>>(relaxed = true)
+
+        viewModel.webChromeClient.onShowFileChooser(
+            mockk(relaxed = true),
+            filePathCallback,
+            mockk(relaxed = true),
+        )
+        advanceUntilIdle()
+
+        viewModel.pendingFileChooser.value!!.onResult(null)
```
TimoPtr: same here

app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
TimoPtr: Also in a future version of JUnit they plan to properly implement Kotlin contract so that we don't need !! after a `assertNotNull`.

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModel.kt
agners: I've changed this based on Copilot's comment https://github.com/home-assistant/android/pull/6881#discussion_r3284661682. 

But yeah I see today this can't happen so let's drop it.

TimoPtr:
You need to update the compose call made in the screenshot test of the `ConnectionScreen` to pass a ChromeClient. You don't need to regenerate the screenshots since the UI doesn't change.

TimoPtr (approved):

PR #6895, opened 2026-05-24
Add Android Assist VAD timing settings

lextiz:
Adds Assist VAD silence and timeout settings to Android voice input.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
@@ -24,6 +24,7 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Check
@@ -45,6 +46,7 @@ import androidx.compose.ui.draw.clip
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.text.input.KeyboardType
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
@@ -63,6 +65,7 @@ import io.homeassistant.companion.android.common.compose.composable.HALabel
 import io.homeassistant.companion.android.common.compose.composable.HALoading
 import io.homeassistant.companion.android.common.compose.composable.HASettingsCard
 import io.homeassistant.companion.android.common.compose.composable.HASwitch
+import io.homeassistant.companion.android.common.compose.composable.HATextField
 import io.homeassistant.companion.android.common.compose.composable.LabelVariant
 import io.homeassistant.companion.android.common.compose.theme.HADimens
 import io.homeassistant.companion.android.common.compose.theme.HARadius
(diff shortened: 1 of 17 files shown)


jpelgrom:
CI actions were stuck, which is why I drafted/undrafted and closed/reopened this PR. Now available so we can get started with review :)

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
```diff
@@ -161,6 +178,12 @@ interface PrefsRepository {
 
     suspend fun setSelectedWakeWord(wakeWord: String)
 
+    suspend fun getAssistVadSettings(): AssistVadSettings
+
+    suspend fun setAssistVadSilenceSeconds(seconds: Double?)
+
+    suspend fun setAssistVadTimeoutSeconds(seconds: Double?)
```
TimoPtr, line 185: Let's not exposes Double here and use Duration instead

```suggestion
    suspend fun setAssistVadSilence(duration: Duration?)

    suspend fun setAssistVadTimeout(duration: Duration?)
```
You'll need to adjust the rest of the codebase. Whenever we can we should avoid primitive types to represent time, in this specific case only the persistent layer and serialization layer should know what to do with the Duration object. it allow for instance to store milliseconds and send something like hours if we want, without having to change the whole chain of call.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
```diff
@@ -32,6 +33,22 @@ data class AutoFavorite(val serverId: Int, val entityId: String) : Parcelable
...
+    if (normalized.count { it == '.' } > 1) return null
+    return normalized.takeIf { it.all { char -> char.isDigit() || char == '.' } }
+}
+
+fun String?.toAssistVadSecondsOrNull(): Double? =
+    this?.normalizedAssistVadSecondsInputOrNull()?.toDoubleOrNull()?.takeIf { it.isFinite() && it > 0.0 }
+
+fun Double.toAssistVadSecondsString(): String = BigDecimal.valueOf(this).stripTrailingZeros().toPlainString()
+
+fun Double?.toAssistVadSecondsInput(): String = this?.toAssistVadSecondsString().orEmpty()
```
TimoPtr, line 50: Why use string here? While migrating to Duration I invite you to migrate to Long if you store the milliseconds or properly document this choice of design.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/prefs/PrefsRepository.kt
```diff
@@ -32,6 +33,22 @@ data class AutoFavorite(val serverId: Int, val entityId: String) : Parcelable
...
+
+fun String.normalizedAssistVadSecondsInputOrNull(): String? {
+    val normalized = replace(',', '.')
+    if (normalized.isEmpty()) return normalized
+    if (normalized.count { it == '.' } > 1) return null
+    return normalized.takeIf { it.all { char -> char.isDigit() || char == '.' } }
+}
+
+fun String?.toAssistVadSecondsOrNull(): Double? =
+    this?.normalizedAssistVadSecondsInputOrNull()?.toDoubleOrNull()?.takeIf { it.isFinite() && it > 0.0 }
```
TimoPtr, line 46: This feel wrong and probably at the wrong place.

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsViewModel.kt
```diff
@@ -148,4 +156,20 @@ class AssistSettingsViewModel @Inject internal constructor(
         wakeWordResetJob?.cancel()
         _uiState.update { it.copy(isTestingWakeWord = testing, wakeWordDetected = false) }
     }
+
+    fun onVadSilenceSecondsChanged(value: String) {
```
TimoPtr, line 160: You should enforce the logic onto the fields and expose to the user when the value are not valid (if they can set an invalid value).

app/src/main/kotlin/io/homeassistant/companion/android/settings/assist/AssistSettingsScreen.kt
```diff
@@ -188,6 +195,19 @@ internal fun AssistSettingsContent(
 
             Spacer(modifier = Modifier.height(HADimens.SPACE2))
 
+            SectionHeader(
```
TimoPtr, line 198: You need to regenerate the screenshot tests

TimoPtr:
Update the description of the PR to match the official template and add screenshots or even video of the new usage.

PR #6896, opened 2026-05-25
Fix pinch-to-zoom remaining active when page zoom is set

zry98:
## Summary

When "Page zoom" is set to anything other than the default 100%, the webview still accepts pinch gestures to zoom with a small range around the configured initial scale, even when "Pinch-to-zoom" is disabled.

The root cause is toggling `WebSettings.builtInZoomControls` alone is not enough to disable pinch when a custom `setInitialScale` is applied, this fix toggles `WebSettings.setSupportZoom` together so that disabling pinch-to-zoom fully disables it regardless of the configured page zoom.

After this fix, I noticed that toggling pinch-to-zoom only took effect after changing the page zoom. I found it's because in `LocalStorageImpl.observeChanges`, `awaitClose` was placed inside the `keys.forEach` loop, the coroutine suspended after the first iteration, so the listeners for the remaining keys were never registered. The first key was effectively the only one observed, changes to later keys are ignored silently.

I have opened another PR for that fix: https://github.com/home-assistant/android/pull/6902

## Checklist

## Any other notes

Depends on https://github.com/home-assistant/android/pull/6902 to be merged to work properly.


jpelgrom:
Thanks for the fix! Do you have a good way to reproduce the issue so we can validate it?

>  I found it's because in `LocalStorageImpl.observeChanges`, `awaitClose` was placed inside the `keys.forEach` loop, the coroutine suspended after the first iteration, so the listeners for the remaining keys were never registered. The first key was effectively the only one observed, changes to later keys are ignored silently.
>
> So the second fix is replacing the per-key listeners with a single listener that filters by the watched key set and registers/unregisters once.

Could you submit this separately? It may be required for this PR but we can review it standalone.

zry98:
> Do you have a good way to reproduce the issue so we can validate it?

@jpelgrom I couldn't reproduce it until I reinstalled the app (clearing data didn't work), not sure why is that. Below is the screen recording reproducing this bug:

https://github.com/user-attachments/assets/f9cdeffa-0f98-487a-af1c-96ca6c2802c3


> Could you submit this separately?

Sure, I've moved it to another PR: https://github.com/home-assistant/android/pull/6902

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
```diff
@@ -136,7 +144,10 @@ sealed interface WebViewAction {
         override fun run(webView: WebView) {
```
TimoPtr, line 271: That's my fault but we misses some unit tests on the WebViewActions can you at least add the one for the ApplyZoom action? It should verify that we call the right settings with the right value and help us avoid making unwanted mistake in the future.

zry98: @TimoPtr thanks for your work! Sure, I've added two tests

app/src/test/kotlin/io/homeassistant/companion/android/frontend/WebViewActionTest.kt
```diff
@@ -113,4 +114,32 @@ class WebViewActionTest {
...
+        every { webView.resources.displayMetrics } returns DisplayMetrics().apply { density = 1.0f }
+        val scriptSlot = slot<String>()
+        every { webView.evaluateJavascript(capture(scriptSlot), any()) } just Runs
+        val action = WebViewAction.ApplyZoom(zoomLevel = 100, pinchToZoomEnabled = true)
+
+        action.run(webView)
+
+        verify { webView.settings.setSupportZoom(true) }
+        verify { webView.settings.builtInZoomControls = true }
+        assertTrue(scriptSlot.captured.contains("user-scalable=yes"))
```
jpelgrom: This isn't a good test, because the script will always contain `user-scalable=yes` and only once it's evaluated will it actually run the script and decide which branch to use. You can copy this line and add it to the test with `pinchToZoomEnabled false`, copy the `scriptSlot` line from that test to this one, and both tests will still pass.

jpelgrom:
Pushed changes to have the test do what's expected in the JS part, remove WebViewActivity and merge upstream changes/resolve conflicts.

TimoPtr (approved):

PR #6902, opened 2026-05-25
Fix LocalStorage `observeChanges` ignores later keys

zry98:
## Summary

In `LocalStorageImpl.observeChanges`, `awaitClose` was placed inside the `keys.forEach` loop, the coroutine suspended after the first iteration, so the listeners for the remaining keys were never registered. The first key was effectively the only one observed, changes to later keys are ignored silently.

This fix replaces the per-key listeners with a single listener that filters by the watched key set and registers/unregisters once.

## Checklist

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
@@ -114,9 +114,8 @@ class LocalStorageImpl(sharedPreferences: suspend () -> SharedPreferences) : Loc
 
     override fun observeChanges(vararg keys: String): Flow<String> = callbackFlow {
         val prefs = sharedPreferences()
-        val watched = keys.toSet()
         val listener = SharedPreferences.OnSharedPreferenceChangeListener { _, changedKey ->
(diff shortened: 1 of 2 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/LocalStorageImpl.kt
```diff
@@ -115,17 +114,16 @@ class LocalStorageImpl(sharedPreferences: suspend () -> SharedPreferences) : Loc
...
-                if (changedKey == key) {
-                    launch { send(key) }
-                }
-            }
-            prefs.registerOnSharedPreferenceChangeListener(listener)
-            awaitClose {
-                prefs.unregisterOnSharedPreferenceChangeListener(listener)
+        val watched = keys.toSet()
+        val listener = SharedPreferences.OnSharedPreferenceChangeListener { _, changedKey ->
+            if (changedKey != null && changedKey in watched) {
```
TimoPtr:
```suggestion
            if (changedKey != null && changedKey in keys) {
```
Is enough

TimoPtr (commented):
Thanks for catching this 👍🏻 it is my mistake ...

TimoPtr (approved):

PR #6906, opened 2026-05-26
Resolve IPv6-only hostnames for WebView and OkHttp on API 23+

### LINKED ISSUE #4953: Companion app fails to connect to IPv6-only HA instance

**Android version(s):**
15

**Device model(s):**
Pixel 6

**Home Assistant version:**
Core 2025.1.0
Supervisor 2024.12.3
Operating System 14.1
Frontend 20250103.0

HASS system is listening on IPv6 / AAAA DNS record only.  Android app won't attempt to connect to IPv6 on LAN.

While using android companion connected to local LAN, lookups to provided DNS server (internal on the same subnet) are only asking for "A" names (never AAAA). Other IPv6 site access (google, fb, etc.) from the phone is normal. Other access to the IPv6 HASS address from hosts on the same LAN is normal (Win/Mac/Linux).

catl0mac-GQG36P6LCV:~ dgeist$ dig +short haus.mydomain.net @192.168.1.67 AAAA
2600:1700:****:4220::18
catl0mac-GQG36P6LCV:~ dgeist$ dig +short haus.mydomain.net @192.168.1.67 A

**Companion App Logs:**

```

```

**Additional information:**

uptake4385:
### Motivation

This has been a persistent issue for me for a long time. My Home Assistant instance is set up in an IPv6-mostly network using ULA (`fd00::/8`) addresses, and the companion app would frequently fail to connect — while every other device and browser on the same network worked perfectly. The root cause is the same as described in #4953: *"android isn't asking AAAA questions on the v4 DNS resolvers (like every other platform does)"*.

**Why this is complex:** Android's `getaddrinfo()` with `AI_ADDRCONFIG` determines IPv6 availability using a two-step probe in bionic (see Chromium issue [#40309810](https://issues.chromium.org/40309810)):
1. `connect()` a UDP socket to the hard-coded address `2000::` — succeeds if any route to global IPv6 space exists
2. `getsockname()` — checks whether the **source address** assigned to the socket is a **global IPv6 address** (`2000::/3`)

If the source address is a ULA (`fd00::/8`), link-local (`fe80::/10`), or the socket has no global IPv6 address at all, bionic considers IPv6 "not available" and **skips AAAA DNS queries entirely** — even when the network is otherwise IPv6-capable. Having a route to `2000::` is not sufficient; the device needs a global IPv6 source address.

Google has marked the upstream issue as **Won't Fix** ([#36955694](https://issuetracker.google.com/issues/36955694)), and Chromium has a related long-standing bug ([#40435291](https://issuetracker.google.com/issues/40435291)).

The problem affects **two completely independent networking stacks** within the app:
1. **OkHttp** (REST API, WebSocket, connectivity checks) — can be fixed via a custom `Dns` implementation
2. **WebView** (Home Assistant frontend) — uses Chromium's own DNS/TLS stack, which does not go through OkHttp

There is no single, simple fix. The solution involves:
- Explicit DNS queries via `DnsResolver` (API 29+) to bypass `getaddrinfo()` entirely
- Raw UDP DNS for API 23–28 where `DnsResolver` is unavailable
- A localhost HTTP CONNECT proxy to route WebView traffic through our DNS resolver
- A `shouldInterceptRequest` fallback for devices that don't support WebView proxy override

We understand this adds complexity to the codebase, but after evaluating all alternatives — URL rewriting to IPv6 literals breaks TLS/SNI, `Network.getAllByName()` goes through the same `getaddrinfo()` on older platforms, `shouldInterceptRequest` alone doesn't cover WebSockets — we believe this is the most robust approach that covers all API levels and traffic types.

### Architecture

**OkHttp (`NetworkAwareDns`):**
- API 29+: `DnsResolver.query()` for explicit TYPE_AAAA + TYPE_A
- API 23–28: `NetworkBoundDnsLookup` (raw UDP DNS on the active network's DNS servers)
- Fallback: `Network.getAllByName()` → `Dns.SYSTEM`

**WebView:**
- `WebViewFeature.PROXY_OVERRIDE` supported: `LocalConnectProxy` — localhost HTTP CONNECT proxy resolving via `NetworkAwareDns`, connecting via `Network.socketFactory`
- PROXY_OVERRIDE unavailable: `HostnameWebViewRequestProxy` via `shouldInterceptRequest` for GET/HEAD

### Key Design Decisions

- **No URL rewriting to IPv6 literals** — breaks TLS certificate validation (SNI expects domain name)
- **No `BufferedReader` for CONNECT parsing** — would lose TLS ClientHello bytes
- **DNS work dispatched off the main thread** — `DnsResolver` delivers I/O on the main looper
- **Runtime feature detection** for `PROXY_OVERRIDE` — depends on WebView version, not `minSdk`
- **User-facing strings are in English only** — existing pattern in the project

### Related Issues

- Closes #4953 — *"Companion app fails to connect to IPv6-only HA instance"*
- Google Issue Tracker [#36955694](https://issuetracker.google.com/issues/36955694) — *"Android only queries for AAAA records in IPv4 network without default gateway"* (Won't Fix)
- Chromium Issue [#40309810](https://issues.chromium.org/40309810) — *"Built-in async DNS resolver needs a reliable equivalent to AI_ADDRCONFIG"*
- Chromium Issue [#40435291](https://issuetracker.google.com/issues/40435291) — similar problem in Chrome/Chromium

### Documentation

See [`docs/network-ipv6-webview-dns.md`](https://github.com/uptake4385/android/blob/feat/network-aware-dns/docs/network-ipv6-webview-dns.md) for the full design, architecture diagrams, debugging guide, and known limitations.

## Checklist

## Screenshots

Not applicable — this is a networking infrastructure change, not a user-facing UI change.

User Documentation: N/A (no user-facing functionality changed)

## Any other notes

We recognize that this is a significant change with substantial complexity. Android's platform-level DNS limitation (marked Won't Fix by Google) leaves apps with no simple alternative. We've aimed to make the implementation as self-contained as possible, with clear boundaries between the DNS resolver, the proxy, and the existing app code. The `docs/network-ipv6-webview-dns.md` document explains the full rationale, alternatives considered, and maintenance guidelines.

We're open to discussing whether the complexity is justified and would appreciate maintainer feedback on the approach.


TimoPtr:
Without looking at the whole code that seems quite complicated for something that should just work. I don't think we can afford maintaining this for a very specific use case, if even the maintainer of the WebView are not doing it TBH.

uptake4385:
Hey @TimoPtr, thanks for the response.

I had the same concern while working on this PR, which is also why I added the note:

> "We're open to discussing whether the complexity is justified and would appreciate maintainer feedback on the approach."

The difficulty here is that the issue itself is actually quite simple, but the workaround unfortunately is not. In dual-stack or IPv6-mostly environments, the application never attempts to resolve/use AAAA records in this scenario, which makes IPv6-only services appear unreachable even when everything is configured correctly.

I completely understand the concern about maintenance cost and complexity. I just could not find a significantly simpler way to solve this cleanly within the current constraints. If you or the maintainers have ideas for a simpler approach, I’d absolutely be open to reworking it.

Even if this specific approach is not the right solution, I think the underlying IPv6/AAAA handling issue still needs some kind of solution long-term.

TimoPtr:
> Hey @TimoPtr, thanks for the response.
> 
> I had the same concern while working on this PR, which is also why I added the note:
> 
> > "We're open to discussing whether the complexity is justified and would appreciate maintainer feedback on the approach."
> 
> The difficulty here is that the issue itself is actually quite simple, but the workaround unfortunately is not. In dual-stack or IPv6-mostly environments, the application never attempts to resolve/use AAAA records in this scenario, which makes IPv6-only services appear unreachable even when everything is configured correctly.
> 
> I completely understand the concern about maintenance cost and complexity. I just could not find a significantly simpler way to solve this cleanly within the current constraints. If you or the maintainers have ideas for a simpler approach, I’d absolutely be open to reworking it.
> 
> Even if this specific approach is not the right solution, I think the underlying IPv6/AAAA handling issue still needs some kind of solution long-term.

Just to be sure to understand the setup. You have HA hosted on a IPv6 only? and you have a DNS that only contains a AAAA record and no A record? 

Does it works on Chrome on Android? Do you have other services like HA that supports this, like Immich or anything we could look at for instance?

uptake4385:
> > Hey @TimoPtr, thanks for the response.
> > I had the same concern while working on this PR, which is also why I added the note:
> > > "We're open to discussing whether the complexity is justified and would appreciate maintainer feedback on the approach."
> > 
> > 
> > The difficulty here is that the issue itself is actually quite simple, but the workaround unfortunately is not. In dual-stack or IPv6-mostly environments, the application never attempts to resolve/use AAAA records in this scenario, which makes IPv6-only services appear unreachable even when everything is configured correctly.
> > I completely understand the concern about maintenance cost and complexity. I just could not find a significantly simpler way to solve this cleanly within the current constraints. If you or the maintainers have ideas for a simpler approach, I’d absolutely be open to reworking it.
> > Even if this specific approach is not the right solution, I think the underlying IPv6/AAAA handling issue still needs some kind of solution long-term.
> 
> Just to be sure to understand the setup. You have HA hosted on a IPv6 only? and you have a DNS that only contains a AAAA record and no A record?
> 
> Does it works on Chrome on Android? Do you have other services like HA that supports this, like Immich or anything we could look at for instance?

Yes, HA is hosted IPv6-only and my local DNS server only provides a AAAA record for it.

> Does it works on Chrome on Android?

I just tested it and it works fine in Chrome on Android. The device is a Pixel 6 running Android 16.

> Do you have other services like HA that supports this, like Immich or anything we could look at for instance?

I have many other services hosted the same way, but I still need to test which of them work or do not work in this situation. The phone is actually my wife's device, not mine, and she also does not use all of those services regularly, so I do not yet know which apps are affected and which are not. I can test some more of them later.

TimoPtr:
> > > Hey @TimoPtr, thanks for the response.
> > > I had the same concern while working on this PR, which is also why I added the note:
> > > > "We're open to discussing whether the complexity is justified and would appreciate maintainer feedback on the approach."
> > > 
> > > 
> > > The difficulty here is that the issue itself is actually quite simple, but the workaround unfortunately is not. In dual-stack or IPv6-mostly environments, the application never attempts to resolve/use AAAA records in this scenario, which makes IPv6-only services appear unreachable even when everything is configured correctly.
> > > I completely understand the concern about maintenance cost and complexity. I just could not find a significantly simpler way to solve this cleanly within the current constraints. If you or the maintainers have ideas for a simpler approach, I’d absolutely be open to reworking it.
> > > Even if this specific approach is not the right solution, I think the underlying IPv6/AAAA handling issue still needs some kind of solution long-term.
> > 
> > 
> > Just to be sure to understand the setup. You have HA hosted on a IPv6 only? and you have a DNS that only contains a AAAA record and no A record?
> > Does it works on Chrome on Android? Do you have other services like HA that supports this, like Immich or anything we could look at for instance?
> 
> Yes, HA is hosted IPv6-only and my local DNS server only provides a AAAA record for it.
> 
> > Does it works on Chrome on Android?
> 
> I just tested it and it works fine in Chrome on Android. The device is a Pixel 6 running Android 16.
> 
> > Do you have other services like HA that supports this, like Immich or anything we could look at for instance?
> 
> I have many other services hosted the same way, but I still need to test which of them work or do not work in this situation. The phone is actually my wife's device, not mine, and she also does not use all of those services regularly, so I do not yet know which apps are affected and which are not. I can test some more of them later.

To be transparent I'm not feeling comfortable merging this it's too big for the use case. I think it is only limited of the case of ULA right? And if you were using a global IPv6 address it would work. 

I've asked about the other app because they might have something in their codebase that we can look at that helps solve the issue without 3k line of code.

Also if it's only your local network why not using IPv4 fallback here? I know it's not ideal but we can't handle the burden of an issue from the upstream like this.

uptake4385:
> > > > Hey @TimoPtr, thanks for the response.
> > > > I had the same concern while working on this PR, which is also why I added the note:
> > > > > "We're open to discussing whether the complexity is justified and would appreciate maintainer feedback on the approach."
> > > > 
> > > > 
> > > > The difficulty here is that the issue itself is actually quite simple, but the workaround unfortunately is not. In dual-stack or IPv6-mostly environments, the application never attempts to resolve/use AAAA records in this scenario, which makes IPv6-only services appear unreachable even when everything is configured correctly.
> > > > I completely understand the concern about maintenance cost and complexity. I just could not find a significantly simpler way to solve this cleanly within the current constraints. If you or the maintainers have ideas for a simpler approach, I’d absolutely be open to reworking it.
> > > > Even if this specific approach is not the right solution, I think the underlying IPv6/AAAA handling issue still needs some kind of solution long-term.
> > > 
> > > 
> > > Just to be sure to understand the setup. You have HA hosted on a IPv6 only? and you have a DNS that only contains a AAAA record and no A record?
> > > Does it works on Chrome on Android? Do you have other services like HA that supports this, like Immich or anything we could look at for instance?
> > 
> > 
> > Yes, HA is hosted IPv6-only and my local DNS server only provides a AAAA record for it.
> > > Does it works on Chrome on Android?
> > 
> > 
> > I just tested it and it works fine in Chrome on Android. The device is a Pixel 6 running Android 16.
> > > Do you have other services like HA that supports this, like Immich or anything we could look at for instance?
> > 
> > 
> > I have many other services hosted the same way, but I still need to test which of them work or do not work in this situation. The phone is actually my wife's device, not mine, and she also does not use all of those services regularly, so I do not yet know which apps are affected and which are not. I can test some more of them later.
> 
> To be transparent I'm not feeling comfortable merging this it's too big for the use case. I think it is only limited of the case of ULA right? And if you were using a global IPv6 address it would work.
> 
> I've asked about the other app because they might have something in their codebase that we can look at that helps solve the issue without 3k line of code.
> 
> Also if it's only your local network why not using IPv4 fallback here? I know it's not ideal but we can't handle the burden of an issue from the upstream like this.

Thanks for the clarification, that makes sense and I can completely understand the maintenance concern.

Just to clarify one detail: this does not only affect ULA addresses. I tested this with both ULA and GUA IPv6 addresses and in both cases the AAAA record was never queried/used in this scenario.

My HA setup is behind a reverse proxy, so as a temporary workaround I was able to simply add an IPv4 address, which solved the immediate issue for me. The main reason I looked into this at all is because the network is intentionally designed to avoid IPv4 wherever possible.

I also fully agree that the current PR is probably too large and invasive to realistically maintain long-term. Honestly, I was already hesitant to open it for that exact reason. My intention was less "this exact implementation must be merged" and more to start a discussion around the underlying issue and maybe find a cleaner/common approach together.

Regarding other applications, I understood your point there as well. I need to check which apps on the device actually handle this correctly so we maybe have some implementations to compare against.

uptake4385:
I tested a few more examples now and both Jellyfin (with the Jellyfin Android app) and Vaultwarden/Bitwarden (using the Bitwarden Android app) work correctly in the same IPv6-only setup with AAAA-only DNS records.

So there are definitely Android applications that handle this scenario properly, which may give us something useful to compare against.

TimoPtr:
> I tested a few more examples now and both Jellyfin (with the Jellyfin Android app) and Vaultwarden/Bitwarden (using the Bitwarden Android app) work correctly in the same IPv6-only setup with AAAA-only DNS records.
> 
> So there are definitely Android applications that handle this scenario properly, which may give us something useful to compare against.

Then I encourage you to look at the codebase of each project to see how they handle it and come back with some details like links to where it is handle to see how it could fit in our project.

PR #6907, opened 2026-05-26
Require authentication for notification action when flag is set

dkarv:
## Summary

Add the authenticationRequired flag for notification actions. It is already supported on iOS.
When enabled, the phone has to be unlocked to trigger a notification action. Useful e.g. for opening the door after the bell was ringed or other sensitive actions.
[community request](https://community.home-assistant.io/t/add-actionable-notification-isauthenticationrequired/502645/2)

## Checklist

User Documentation: home-assistant/companion.home-assistant#1331

## Any other notes

* I found no unit tests for the notifications, and most functions are private. No test is added because it would require building them from scratch for this small change.
* I only tested with local WebSocket notifications, not with Firebase. Do I need to whitelist the authenticationRequired key somewhere?


jpelgrom:
Thanks for the suggestion. However, this doesn't map cleanly on the same pattern as iOS where it is a simple yes/no; the builder specifies

> If this is `true` and the device is locked when the action is invoked, the OS will show the keyguard and require successful authentication before invoking the intent.

... so you won't be required to authenticate when the device is **un**locked, even if you send `true`.

> If this is `false` and the device is locked, the OS will decide whether authentication should be required.

... so you might still end up needing to authenticate if Android decides so, even when you send `false`.

dkarv:
Thanks for your thoughts. I had another look to compare the behavior on both platforms:

> ... so you won't be required to authenticate when the device is unlocked, even if you send true.

This is the same behavior I saw on iOS. When the device is already unlocked, no further authentication is needed and the action is performed directly. According to [the Apple documentation](https://developer.apple.com/documentation/usernotifications/unnotificationactionoptions/authenticationrequired) this is expected, it states _The action can be performed only on an unlocked device._

> ... so you might still end up needing to authenticate if Android decides so, even when you send false.

I could not find any remark and never experienced this. So I checked in the AOSP code and it seems it just relies on that boolean we can pass: it is queried and passed down [here](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/packages/SystemUI/src/com/android/systemui/statusbar/NotificationRemoteInputManager.java?q=%5C.isAuthenticationRequired()) and evaluated [here](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/packages/SystemUI/src/com/android/systemui/statusbar/phone/StatusBarRemoteInputCallback.java;l=328?q=appRequestedAuth)
```
        if (isActivity || appRequestedAuth) {
            mActionClickLogger.logWaitingToCloseKeyguard(pendingIntent, actionIndex);
            final boolean afterKeyguardGone = mActivityIntentHelper
                    .wouldPendingLaunchResolverActivity(pendingIntent,
                            mLockscreenUserManager.getCurrentUserId());
            mActivityStarter.dismissKeyguardThenExecute(() -> {
                mActionClickLogger.logKeyguardGone(pendingIntent, actionIndex);

                try {
                    ActivityManager.getService().resumeAppSwitches();
                } catch (RemoteException e) {
                }

                boolean handled = defaultHandler.handleClick();

                // close the shade if it was open and maybe wait for activity start.
                return handled && mShadeController.closeShadeIfOpen();
            }, null, afterKeyguardGone);
            return true;
        } else {
            return defaultHandler.handleClick();
        }
```
Even if future Android version change it and require another authentication (e.g. when a device is unlocked for too long) I think this is only helping the goal of this flag and further improves security.

Happy to hear if I got something wrong. But in conclusion, I would argue the small potential difference in the future does not justify a separate flag for Android and iOS. It would need to be documented, users would need to put two flags into each notification definition in HA, and more.

I also checked again the backwards compatibility. If authenticationRequired is not given in the notification data, nothing changes for existing users.

jpelgrom:
> This is the same behavior I saw on iOS. When the device is already unlocked, no further authentication is needed and the action is performed directly. According to [the Apple documentation](https://developer.apple.com/documentation/usernotifications/unnotificationactionoptions/authenticationrequired) this is expected, it states The action can be performed only on an unlocked device.

In that case I think the companion app documentation should also be updated to reflect that, right now it suggests otherwise.

app/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -1605,6 +1605,7 @@ class MessagingManager @Inject constructor(
                     )
                 }
 
+                val authenticationRequired = data["action_${i}_authenticationRequired"]?.toBoolean() == true
```
jpelgrom, line 1621: Why not apply this in the URI action as well?

app/src/main/kotlin/io/homeassistant/companion/android/websocket/WebsocketManager.kt
```diff
@@ -61,6 +61,7 @@ class WebsocketManager(appContext: Context, workerParams: WorkerParameters) :
         } else {
             WebsocketSetting.ALWAYS
         }
+        private val ACTION_KEY_WHITELIST = listOf("uri", "behavior", "authenticationRequired")
```
jpelgrom: This is kind of weirdly named, `action` and `title` are also allowed keys.

dkarv: renamed it to `ACTION_EXTRA_KEYS`

app/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
dkarv: Added it to the uri as well for consistency. It probably does not make a difference there, because to open the app the device always needs to be unlocked anyways. But there might be edge cases where it helps somebody and applying it does not do any harm...

dkarv:
> In that case I think the companion app documentation should also be updated to reflect that, right now it suggests otherwise.

Makes sense, I changed the documentation in the linked PR:

| `authenticationRequired` | **Optional**. If `true`, the device needs to be unlocked before the action is triggered. | <span class='beta'>BETA</span> on Android. Requires at least Android 12.  |

jpelgrom:
Thanks for updating the documentation. In that case as behavior matches I see no objections.

> I only tested with local WebSocket notifications, not with Firebase. Do I need to whitelist the authenticationRequired key somewhere?

You need to add the key to the list of keys that are passed to Android like here: https://github.com/home-assistant/mobile-apps-fcm-push/blob/main/functions/android.js#L27-L29. If you submit a PR we'll merge it alongside all the other ones and it will immediately start working with FCM.

jpelgrom:
Could you also add a short line for this change to the [in-app changelog](https://github.com/home-assistant/android/blob/main/app/src/main/res/xml/changelog_master.xml) like "Added support for ..."?

dkarv:
> You need to add the key to the list of keys that are passed to Android like here: https://github.com/home-assistant/mobile-apps-fcm-push/blob/main/functions/android.js#L27-L29. If you submit a PR we'll merge it alongside all the other ones and it will immediately start working with FCM.

Created the PR: https://github.com/home-assistant/mobile-apps-fcm-push/pull/306

> Could you also add a short line for this change to the [in-app changelog](https://github.com/home-assistant/android/blob/main/app/src/main/res/xml/changelog_master.xml) like "Added support for ..."?

Also done, hope I added it the correct way.

Thanks overall for your suggestions and help to bring this to live, appreciate it.

jpelgrom (approved):
Thanks for picking this up!

PR #6913, opened 2026-05-27
Prevent unregistering actionReceiver twice and handle weird lifecycle

TimoPtr:
## Summary

We have 1k users impacted by a crash of the `AssistVoiceInteractionService` while `onShutdown` is invoked, this PR catch the exception. It seems to happen because the system already unregistered the `actionReceiver`.
image

This PR also handle some weirdness of the lifecycle of the service here the documentation of `onReady`

```
    /**
     * Called during service initialization to tell you when the system is ready
     * to receive interaction from it. You should generally do initialization here
     * rather than in {@link #onCreate}. Methods such as {@link #showSession} will
     * not be operational until this point.
     *
     * <p>Note: Under some circumstances, it is possible for {@link #onReady} to be called after
     * {@link #onDestroy} has been called. This can happen in cases where the system is in the
     * process of shutting down the service, but the {@code IVoiceInteractionManagerService}
     * still sends a final "ready" signal. Therefore, implementations must be prepared for
     * {@link #onReady} to be called even after {@link #onDestroy}, and should not rely on a
     * fixed lifecycle order.
     */
```

and onShutdown

```
    /**
     * Called during service de-initialization to tell you when the system is shutting the
     * service down.
     * At this point this service may no longer be the active {@link VoiceInteractionService}.
     *
     * <p>Note: After this method is called, it is possible for other methods on this service
     * to be called, including {@link #onReady}. Implementations should be prepared to handle
     * this case gracefully.
     */
```

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionService.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionService.kt
@@ -68,11 +68,27 @@ class AssistVoiceInteractionService : VoiceInteractionService() {
     }
     private var isServiceReady = false
 
+    /**
+     * One-way latch set once the service has begun tearing down (via [onShutdown] or [onDestroy]).
+     *
+     * The platform documents that [onReady] may be delivered after teardown and that no fixed
+     * lifecycle order can be relied upon. Because [serviceScope] is cancelled during teardown and
+     * cannot be reused, the instance can never function again, so any such late [onReady] is ignored.
+     */
+    private var isTornDown = false
+
     /** Non-null only while the receiver is registered (between [onReady] and [onShutdown]). */
     private var actionReceiver: BroadcastReceiver? = null
 
     override fun onReady() {
         super.onReady()
+        if (isTornDown) {
+            // The system can deliver onReady even after onShutdown/onDestroy while it is winding the
+            // service down. Registering a receiver on the dying context would later crash in
+            // onShutdown, so ignore this spurious signal
+            Timber.w("Ignoring onReady delivered after shutdown")
+            return
+        }
         isServiceReady = true
         Timber.d("VoiceInteractionService is ready")
         actionReceiver = object : BroadcastReceiver() {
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionService.kt
```diff
@@ -68,11 +68,27 @@ class AssistVoiceInteractionService : VoiceInteractionService() {
...
+
     /** Non-null only while the receiver is registered (between [onReady] and [onShutdown]). */
     private var actionReceiver: BroadcastReceiver? = null
 
     override fun onReady() {
         super.onReady()
+        if (isTornDown) {
+            // The system can deliver onReady even after onShutdown/onDestroy while it is winding the
+            // service down. Registering a receiver on the dying context would later crash in
+            // onShutdown, so ignore this spurious signal
```
jpelgrom, line 82: Duplicate with the comment on the variable itself. I'd remove from the variable and keep it here, so the variable gets a nice and short one line comment.

jpelgrom (commented):
Interesting that we have so many users with the issue but didn't encounter it during testing 🤔

app/src/main/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionService.kt
```diff
@@ -68,11 +68,27 @@ class AssistVoiceInteractionService : VoiceInteractionService() {
     }
     private var isServiceReady = false
 
+    /**
+     * One-way latch set once the service has begun tearing down (via [onShutdown] or [onDestroy]).
+     *
+     * The platform documents that [onReady] may be delivered after teardown and that no fixed
+     * lifecycle order can be relied upon. Because [serviceScope] is cancelled during teardown and
+     * cannot be reused, the instance can never function again, so any such late [onReady] is ignored.
+     */
```
jpelgrom: To resolve https://github.com/home-assistant/android/pull/6913#discussion_r3320562703


```suggestion
    /** One-way latch set once the service has begun tearing down (via [onShutdown] or [onDestroy]). */
```

app/src/main/kotlin/io/homeassistant/companion/android/assist/service/AssistVoiceInteractionService.kt
```diff
@@ -102,14 +118,32 @@ class AssistVoiceInteractionService : VoiceInteractionService() {
...
+        }
         // Don't use stopListening() as it launches a coroutine that may not complete before cancel
         serviceScope.cancel()
     }
 
+    override fun onDestroy() {
+        // onShutdown is not guaranteed to run before onDestroy, so latch teardown here too to keep
+        // a late onReady from re-initializing an already-destroyed instance
+        isTornDown = true
+        super.onDestroy()
```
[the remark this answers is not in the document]
jpelgrom: onShutdown is not guaranteed to run _before onDestroy_ but should still run, so this is enough.

jpelgrom (approved):

Issue #6921, opened 2026-05-28
Received notifications without a server are silently ignored, should be highlighted


jpelgrom:
If the app receives a notification, it will try to match the webhook ID in the payload to a server and store the received data + matched server ID. Depending on whether a server is found, it may be ignored (not posted), invisible to the user.

To make this issue more visible, the notification history should highlight notifications without a matching server ID. Might be good to pick up as part of a larger redesign of the list, and needs some discussion before implementation.

PR #6922, opened 2026-05-28
More strictly reject notifications without a valid server in the app

jpelgrom:
## Summary

Following [a user report](https://discord.com/channels/330944238910963714/1509048773773496320) about unwanted notifications, I had another look at our validation for notifications. This PR changes processing of notifications to more strictly reject (ignore) notifications that are received if no matching server can be found.

There are two possible scenarios for valid notifications:
- no webhook ID in payload: old server, match to in-app active server
- webhook ID in payload: match to server with that ID in the database

If a notification payload is received with a webhook ID but the app doesn't have a server with that ID in the database, it should be ignored to avoid processing messages from servers removed from the app. That previously didn't work, as this incorrectly triggered the "no webhook ID in payload" fallback. This PR fixes it.

Related: #6921

## Checklist

- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction). _No, MessagingManager has no tests and this isn't trivial to introduce right now._

## Screenshots

n/a

n/a

### DIFF
--- a/wear/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/wear/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -25,6 +25,7 @@ import io.homeassistant.companion.android.database.notification.NotificationDao
 import io.homeassistant.companion.android.database.notification.NotificationItem
 import io.homeassistant.companion.android.database.sensor.SensorDao
 import io.homeassistant.companion.android.sensors.SensorReceiver
+import io.homeassistant.companion.android.util.sensitive
 import javax.inject.Inject
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -47,9 +48,9 @@ class MessagingManager @Inject constructor(
 
             val jsonData = notificationData as Map<String, String>?
             val jsonObject = jsonData?.toJsonObject()
-            val serverId = jsonData?.get(NotificationData.WEBHOOK_ID)?.let {
+            val receivedServer = jsonData?.get(NotificationData.WEBHOOK_ID)?.let {
                 serverManager.getServer(webhookId = it)?.id
-            } ?: ServerManager.SERVER_ID_ACTIVE
+            }
             val notificationRow =
                 NotificationItem(
                     0,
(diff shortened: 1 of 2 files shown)


wear/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -47,21 +48,36 @@ class MessagingManager @Inject constructor(
 
             val jsonData = notificationData as Map<String, String>?
             val jsonObject = jsonData?.toJsonObject()
-            val serverId = jsonData?.get(NotificationData.WEBHOOK_ID)?.let {
+            val receivedServer = jsonData?.get(NotificationData.WEBHOOK_ID)?.let {
```
TimoPtr: receivedServerId would be more appropriate.

jpelgrom: I'm mirroring the variable name in the `app` module. I could also rename it there if you're OK with extending the scope a tiny bit?

TimoPtr: Fine for me to extend also for serverIdForWebhook

jpelgrom: Adjusted. `serverForWebhook` -> `serverIdForWebhook` can be ignored if I don't use the ID immediately ;)

wear/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -47,25 +48,40 @@ class MessagingManager @Inject constructor(
...
                     jsonObject.toString(),
                     source,
-                    serverId,
+                    receivedServerId,
                 )
             notificationDao.add(notificationRow)
-            if (serverManager.getServer(serverId) == null) {
-                Timber.w("Received notification but no server for it, discarding")
+
+            val webhookServerId = jsonData?.get(NotificationData.WEBHOOK_ID)?.let { webhookId ->
```
TimoPtr: L51 you already read the webhook_id from the jsonData can't we do it only once?

jpelgrom: More scope creep ;)

This only applies to `wear`, in `app` it might be reading from database or parameter.

TimoPtr (approved):

PR #6955, opened 2026-06-02
Add dedicated screen for invitation flow instead of WelcomeScreen

TimoPtr:
## Summary

In the current flow when clicking on an invite link we don't see the server URL we are going to connect to. This PR adds the information about the invitation into a new `WelcomeInvitationScreen` that disclose the URL and also the risk of accepting an invitation, to raise awareness from the user.

## Checklist

## Screenshots

image

## Any other notes

I decided to keep the same structure that the `WelcomeScreen` by making a template where we can simply change the value and add more content.


bgoncal (changes requested):
We need a "Reject" button to prevent the user to get stuck and also because it's the intuitive counterpart when you have "Accept"

TimoPtr:
> We need a "Reject" button to prevent the user to get stuck and also because it's the intuitive counterpart when you have "Accept"

Going back is going close the app do we need a reject button? The reject action could redirect to the server discovery but I'm not sure the experience is great.

bgoncal:
Oh the famous android back button/gesture, I always forget about it, but "back" in this case is not intuitive since the user may have no previous context, so I still vouch for reject, you can redirect them back to welcome view in case you dont think it makes sense to show servers search

jpelgrom:
> Oh the famous android back button/gesture, I always forget about it, but "back" in this case is not intuitive since the user may have no previous context

Back is a system button/gesture so you cannot not have it.

I do agree a 'no' option on screen might make sense. What about a X icon in the top app bar, like we do have a <- when opening from the app settings? That is a clear cancel/no icon without introducing more buttons in the layout.

bgoncal:
Yeah, I understand that the back needs to exist, I just dont agree it should be the only way.
Why not having the "no" button below the accept? It's the most common way to present options to a user in a similar scenario as this

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/welcome/WelcomeInvitationScreen.kt
```diff
@@ -0,0 +1,131 @@
...
+            Row {
+                Column(modifier = Modifier.weight(1f), verticalArrangement = Arrangement.spacedBy(HADimens.SPACE2)) {
+                    Text(
+                        text = stringResource(commonR.string.welcome_invitation_server_label),
+                        style = HATextStyle.Body.copy(textAlign = TextAlign.Start),
+                    )
+                    ServerAddressLink(serverUrl = serverUrl)
+                }
+                Icon(
+                    imageVector = Icons.Default.MarkEmailUnread,
```
jpelgrom, line 74: Does this icon make sense as an invite icon? It is a regular share icon in the app. I don't have a good alternative but this feels out of place to me.

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/onboarding/welcome/WelcomeInvitationScreenshotTest.kt
```diff
@@ -0,0 +1,22 @@
...
+
+class WelcomeInvitationScreenshotTest {
+
+    @PreviewTest
+    @HAPreviews
+    @Composable
+    fun `WelcomeInvitationScreen`() {
+        HAThemeForPreview {
+            WelcomeInvitationScreen(
+                serverUrl = "http://homeassistant.local:8123",
```
jpelgrom, line 19: Could you add a variant with a longer URL (like HA Cloud remote access) so we can have wrapping in the test?

app/src/screenshotTestFullDebug/reference/io/homeassistant/companion/android/onboarding/welcome/WelcomeInvitationScreenshotTest/WelcomeInvitationScreen_foldable_c908f502_0.png
jpelgrom, line 1: Something went wrong while generating here... "Home**Open editor**Assistant address"

jpelgrom:
> Why not having the "no" button below the accept? It's the most common way to present options to a user in a similar scenario as this

Might be personal bias, but I rarely see three buttons below each other on Android and it looks odd. Also see the recent tag reader sheet as an example: two buttons + cancel icon.

We could consider moving 'Learn more' to be closer to the content? Approve/Reject can then exist next to each other without a 3 button list.

bgoncal:
> Might be personal bias, but I rarely see three buttons below each other on Android and it looks odd. Also see the recent tag reader sheet as an example: two buttons + cancel icon.

Agree, then I would move learn more as an icon at the top or remove it completely, not everything needs a "learn more" button, UI can be just good enough to not require one

jpelgrom:
I was just editing my comment to say that, agree about moving Learn more :)

common/src/main/res/values/strings.xml
```diff
@@ -1405,6 +1405,10 @@
     <string name="welcome_details">Access your Home\u00A0Assistant server on the go.\n\nHome\u00A0Assistant is open source home automation that puts local control and privacy first. Powered by a worldwide community of tinkerers and DIY enthusiasts.</string>
     <string name="welcome_connect_to_ha">Connect to my Home\u00A0Assistant server</string>
     <string name="welcome_learn_more">Learn more</string>
+    <string name="welcome_invitation_title">You\'ve been invited</string>
+    <string name="welcome_invitation_server_label">Home\u00A0Assistant address</string>
+    <string name="welcome_invitation_warning_detail">Only continue if you trust the person who shared it. Connecting to an unknown server could give it access to data the app sends, such as your location and device information.</string>
```
jpelgrom: - Is 'it' the person or the address directly placed above this text?
- Match the positive action with the button's text.
- I think it's a good idea to clarify it's also about the data the user enters when connecting, but it does make the text a bit longer than I'd like 🤔

```suggestion
    <string name="welcome_invitation_warning_detail">Only accept this invite if you trust the person who shared it. Connecting to an unknown server could give it access to data you enter or sent by the app, such as your location and device information.</string>
```

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/welcome/WelcomeInvitationScreen.kt
TimoPtr: We can drop it. Or @bgoncal you might have a suggestion?

bgoncal: I guess it's a matter of personal perception? To me the mail icon fits perfectly on the "Invite" context.

bgoncal: For example when searching on google images for "invite"

CleanShot 2026-06-02 at 15 45 59@2x

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/onboarding/welcome/WelcomeInvitationScreenshotTest.kt
```diff
@@ -0,0 +1,47 @@
...
+                onAcceptClick = {},
+                onRejectClick = {},
+                onLearnMoreClick = {},
+            )
+        }
+    }
+
+    @PreviewTest
+    @Preview
+    @Preview(
```
TimoPtr, line 29: I didn't use HAPreviews because we don't need to test all variants I picked a small and big screen.

app/src/main/kotlin/io/homeassistant/companion/android/onboarding/welcome/WelcomeInvitationScreen.kt
TimoPtr: I kept the icon for now

bgoncal (approved):

jpelgrom (approved):

PR #6959, opened 2026-06-02
Fix server invite share url format missing url=

jpelgrom:
## Summary

While testing #6955 I noticed that the share URL I created was rejected ([failfast](https://github.com/home-assistant/android/blob/26154d923c3813cd4650e124900780f4f796d094/app/src/main/kotlin/io/homeassistant/companion/android/launch/link/LinkHandler.kt#L138)) because it was sharing `my.home-assistant.io/invite/#server-url` instead of `my.home-assistant.io/invite/#url=server-url`. This PR fixes it to the correct format. Checked iOS source as well to make sure this is the correct format.

## Checklist

## Screenshots
n/a

n/a

## Any other notes

Git history didn't show a change so did this ever work at all... 😳

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/server/ServerSettingsFragment.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/server/ServerSettingsFragment.kt
@@ -45,7 +45,7 @@ import kotlinx.coroutines.launch
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import timber.log.Timber
 
-private const val BASE_INVITE_URL = "https://my.home-assistant.io/invite/#"
+private const val BASE_INVITE_URL = "https://my.home-assistant.io/invite/#url="
 
 @AndroidEntryPoint
 class ServerSettingsFragment :


TimoPtr:
It probably never worked indeed... Let's add a test for that if possible.

jpelgrom:
We don't have a setup for testing fragments, and the default fragment testing library doesn't work with Hilt so I'm currently trying to cobble together a setup that works for testing but it isn't coming along quickly. This might take a while.

<details>
<summary>Details</summary>

Building on https://github.com/android/architecture-samples/blob/2291fc6d2e17a37be584a89b80ee73c207c804c3/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/HiltExt.kt but I need to return the scenario somehow, as I want to mock the call to getActivity().addMenuProvider() so I can call the menu provider item selected function programmatically. Lifecycle needs to be progressed because the menu is added once resumed and view created...

</details>

TimoPtr:
> We don't have a setup for testing fragments, and the default fragment testing library doesn't work with Hilt so I'm currently trying to cobble together a setup that works for testing but it isn't coming along quickly. This might take a while.
> Details
> 
> Building on [android/architecture-samples@`2291fc6`/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/HiltExt.kt](https://github.com/android/architecture-samples/blob/2291fc6d2e17a37be584a89b80ee73c207c804c3/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/HiltExt.kt) but I need to return the scenario somehow, as I want to mock the call to getActivity().addMenuProvider() so I can call the menu provider item selected function programmatically. Lifecycle needs to be progressed because the menu is added once resumed and view created...

Let's merge it like this for now, the easiest solution might to move this out of the fragment.

PR #6967, opened 2026-06-04
Rename HttpAuthManager to FrontendHttpAuthHandler

TimoPtr:
## Summary

While writing the documentation of the new architecture https://github.com/home-assistant/developers.home-assistant/pull/2965 we found some inconsistency in the naming that needs to be addressed. The `FrontendHttpAuthHandler` was in fact a Handler since has per definition depending on other Manager also conflicting with `android.webkit.HttpAuthHandler`.

## Checklist


jpelgrom (approved):
This reminded me of the usage of `authenticationDao` directly in here. Even though very minimal I think we should change its uses to go via a repository (AuthenticationRepository? It already handles app lock as well, so not strictly one source).

TimoPtr:
I don't think the current `AuthenticationRepository` fits the actual usage we need here, it is for the current session with HA where here we are at the basic_auth level for a resources within a dashboard. We would need a `BasicAuthRepository` most probably, if you mean we miss one layer of abstraction?

PR #6968, opened 2026-06-04
Remove useless interface abstraction and use ViewModelScope for externalBus

TimoPtr:
## Summary

Remove the `FrontendExternalBusRepository` interface in favor of a simple class since it doesn't bring any value, we also took the opportunity to properly scope the repository to the ViewModel since it doesn't make sense that it survives it.

## Checklist


app/src/main/kotlin/io/homeassistant/companion/android/frontend/externalbus/FrontendExternalBusRepository.kt
jpelgrom, line 1: It is weird to only have an impl file

jpelgrom:
I'm not sure I agree the interface abstraction for repositories is useless; it forces you to very clearly indicate what is public and what is not, and moves documentation out of the implementation source file so any comments there can focus on the implementation which isn't relevant when accessing it from elsewhere.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/externalbus/FrontendExternalBusRepository.kt
TimoPtr: Forgot to rename

TimoPtr:
> I'm not sure I agree the interface abstraction for repositories is useless; it forces you to very clearly indicate what is public and what is not, and moves documentation out of the implementation source file so any comments there can focus on the implementation which isn't relevant when accessing it from elsewhere.

I think that abstraction need to solve a problem like has you said limit the scope or helping with testing or avoid leaking implementation details. If we look at the Managers, Handlers I didn't make abstraction but we might in some specific scenarios.

Removing the interface simplify IMO the setup by removing the need of a binding at the CI level, and remove a bit of boilerplate.

For an SDK or :common it probably make more sense to have this abstraction layer, to not leak some internals.

About the implementation focus documentation you are right I merged the two into the `/**` block documentation I was considering using `/*` bellow the public documentation for the implementation.

jpelgrom:
> For an SDK or :common it probably make more sense to have this abstraction layer, to not leak some internals.

I think my comment was mostly motivated by thinking of how this would apply if removed in `:common`. Perhaps we can add it as a guideline in the architecture docs - no interface if within the same module/flavor, otherwise prefer interface+impl (so not a strict rule, we can do things different when there is a good reason for it)?

> About the implementation focus documentation you are right I merged the two into the `/**` block documentation I was considering using `/*` bellow the public documentation for the implementation.

Implementation details still shouldn't be in the doc now that they've been merged, but in another comment. If you adjust the doc for `send` and file name here, I'm OK with merging it.

jpelgrom (approved):
Git history is looking weird because it didn't detect the rename here, but changes are OK

Issue #5432, opened 2025-06-20
Route Assist audio through Android Auto when connected


awalker4:
**Is your feature request related to a problem? Please describe.**
I've come to rely heavily on some custom Assist sentences to help me save random shower thoughts when I'm driving. I set up a Tasker automation to make it easier to use Assist hands-free, but I noticed that the responses from Assist are never routed to my car speakers.

**Describe the solution you'd like**
From my quick research, it looks like Android Auto has a special API for playing media, and it doesn't just pick up all audio like a bluetooth speaker. Would it be feasible to use this API when the app is connected to the car? I'm not sure if Apple users would have the same issue. I'm happy to borrow my wife's iPhone to test if it helps.

**Describe alternatives you've considered, if any**
Right now I just have to talk to Assist and then hold the phone next to my ear 🙃 

**Additional context**

dshokouhi:
wonder if this requires https://github.com/home-assistant/android/issues/3788 or is just a duplicate as implementing the assistant should take care of the routing.

awalker4:
Just found that issue and was about to link it! Yeah, seems like this would come about as part of that work anyway. If anything, I guess we'd want to use that routing if the car is connected but Assist is run from the phone, for consistency.

PR #6978, opened 2026-06-06
Before playing TTS, request audio focus

ChadKillingsworth:
## Summary
When using Android Auto, TTS notifications are routed to the phone speakers instead of the car unless the car head unit audio source is actually "Android Auto". When the car is on any other audio source (including local audio such as the radio or even muted or silenced), TTS content plays through the phone speakers. Most apps, including assistant apps, do not function this way. They will temporarily interrupt any other audio source to play a notification, but then return control to the original source.

To fix this behavior, this change requests audio focus for the companion app temporarily to play the notification then relinquishes focus back to the prior source.  See https://developer.android.com/media/optimize/audio-focus

This is related to #5432 but I'm not sure if the assistant functionality uses the textToSpeech engine. If not, a similar technique could be utilized.

## Checklist

## Any other notes
I tested this both on emulators as well as real devices paired to a real head unit. When using a real device, the audio correctly routed but the first syllable (or two) were cut off. To address this, an empty utterance is injected to the front of the queue after which the notifications played perfectly. 

When not using an alarm stream, the audio attributes did not specify a content type or usage. Without this, Android Auto routing was not working at all. This introduces a default content type of "Speech" and usage of "Media" for TTS content (except when using the Alarm media stream).

The `androidx.media` classes used in this change are all marked deprecated and instead `androidx.media3` helpers are suggested. However, the `androidx.media3` equivalent classes are marked with `@UnstableAPI` and do not appear ready for production use.

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
@@ -19,6 +19,7 @@ import kotlinx.coroutines.cancelChildren
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withTimeoutOrNull
 import timber.log.Timber
 
 /**
@@ -54,12 +55,12 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
                 }
                 AudioManager.AUDIOFOCUS_LOSS -> {
                     Timber.d("Audio focus lost permanently")
+                    hasFocus.value = false
                     stopTTS()
                 }
-                AudioManager.AUDIOFOCUS_LOSS_TRANSIENT,
-                AudioManager.AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK,
-                -> {
+                AudioManager.AUDIOFOCUS_LOSS_TRANSIENT -> {
                     Timber.d("Audio focus lost temporarily")
+                    hasFocus.value = false
                     isTransientLoss = true
                     playbackJob?.cancel()
                     playbackJob = null
@@ -111,6 +112,7 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
         utteranceQueue.clear()
         textToSpeechEngine.release()
         abandonAudioFocus()
+        hasFocus.value = false
         isPlaying = false
     }
 
@@ -154,7 +156,6 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
 
             val request = AudioFocusRequestCompat.Builder(AudioManagerCompat.AUDIOFOCUS_GAIN_TRANSIENT_MAY_DUCK)
                 .setAudioAttributes(audioAttributesCompat)
-                .setWillPauseWhenDucked(false)
                 .setOnAudioFocusChangeListener(focusListener)
                 .build()
 
@@ -177,12 +178,13 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
 
     private fun abandonAudioFocus() {
         val audioManager = applicationContext.getSystemService<AudioManager>() ?: return
-        focusRequest?.let { request ->
-            try {
+        try {
+            val request = focusRequest
+            if (request != null) {
                 AudioManagerCompat.abandonAudioFocusRequest(audioManager, request)
-            } catch (e: Exception) {
-                Timber.w(e, "Failed to abandon audio focus")
             }
+        } catch (e: Exception) {
+            Timber.w(e, "Failed to abandon audio focus")
         }
         focusRequest = null
     }
@@ -193,11 +195,30 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
      */
     private suspend fun play() {
         isTransientLoss = false
-        requestAudioFocus()
+        val focusGranted = requestAudioFocus()
+
+        if (!focusGranted && !hasFocus.value) {
+            Timber.e("Audio focus request denied")
+            handleError(applicationContext.getString(R.string.tts_error_focus_denied))
+            utteranceQueue.clear()
+            abandonAudioFocus()
+            return
+        }
 
         if (!hasFocus.value) {
             Timber.d("Audio focus not granted yet, waiting...")
-            hasFocus.first { it }
+            val regained = withTimeoutOrNull(FOCUS_TIMEOUT_MS) {
+                hasFocus.first { it }
+                true
+            } ?: false
+
+            if (!regained) {
+                Timber.e("Timed out waiting for initial audio focus")
+                handleError(applicationContext.getString(R.string.tts_error_focus_denied))
+                utteranceQueue.clear()
+                abandonAudioFocus()
+                return
+            }
         }
 
         textToSpeechEngine.initialize().onFailure { throwable ->
@@ -213,7 +234,7 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
                 // Over bluetooth connections, the first syllable or even word can be cut off.
                 // Adding an initial empty utterance seems to fix this.
                 // Testing shows this is more effective than utilizing the
-                // textToSpeach.playSilentUtterance method.
+                // textToSpeech.playSilentUtterance method.
                 if (utteranceQueue.isNotEmpty()) {
                     utteranceQueue.addFirst(
                         Utterance(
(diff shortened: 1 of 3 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
```diff
@@ -30,6 +36,38 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
...
+                }
+                AudioManager.AUDIOFOCUS_LOSS_TRANSIENT,
+                AudioManager.AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK,
+                -> {
+                    Timber.d("Audio focus lost temporarily")
+                    isTransientLoss = true
+                    playbackJob?.cancel()
+                    playbackJob = null
+                    textToSpeechEngine.release()
+                }
```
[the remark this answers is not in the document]
ChadKillingsworth: Because ducking is auto-handled from Android 8.0+, I removed this case

common/src/test/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClientTest.kt
```diff
@@ -0,0 +1,120 @@
...
+        client = TextToSpeechClient(context, textToSpeechEngine)
+    }
+
+    @After
+    fun tearDown() {
+        unmockkAll()
+    }
+
+    @Test
+    @Config(sdk = [26])
```
TimoPtr, line 52: I think you can pass 2 versions to `Config` 23 and 26 so you test the 2 versions in only one test.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
```diff
@@ -67,39 +106,186 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
...
     }
 
+    private fun startPlayback() {
+        if (isPlaying) return
+        isPlaying = true
+        playbackJob = mainScope.launch {
+            try {
+                play()
+            } finally {
+                isPlaying = false
```
TimoPtr, line 126: You are setting a boolean potentially from another thread, you should ensure that isPlaying is read and write always from the same thread to ensure concurrency issues.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
```diff
@@ -30,6 +37,38 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
...
+                    }
+                }
+                AudioManager.AUDIOFOCUS_LOSS -> {
+                    Timber.d("Audio focus lost permanently")
+                    hasFocus.value = false
+                    stopTTS()
+                }
+                AudioManager.AUDIOFOCUS_LOSS_TRANSIENT -> {
+                    Timber.d("Audio focus lost temporarily")
+                    hasFocus.value = false
```
TimoPtr, line 63: Why not also calling `stopTTS`?

common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
```diff
@@ -67,39 +106,186 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
...
     /**
      * Plays each queued [Utterance] in sequence until [utteranceQueue] is empty.
      * There can be further additions to the queue while a message is playing which will be picked up in the running playback loop.
      */
     private suspend fun play() {
-        isPlaying = true
+        isTransientLoss = false
+        val focusGranted = requestAudioFocus()
+
+        if (!focusGranted && !hasFocus.value) {
```
TimoPtr, line 200: `requestAudioFocus` does return what should be in `hasFocus` no? otherwise only use hasFocus and only set the hasFocus in the callback. by setting the value from multiple place it's a receipt for making debugging harder.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
```diff
@@ -67,39 +106,186 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
...
+            Timber.e("Audio focus request denied")
+            handleError(applicationContext.getString(R.string.tts_error_focus_denied))
+            utteranceQueue.clear()
+            abandonAudioFocus()
+            return
+        }
+
+        if (!hasFocus.value) {
```
TimoPtr, line 208: `hasFocus` value could have change in between the moment you check in line 200 and now. You probably want a snapshot of the value when starting play. Or you need to document explicitly why you check again.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
```diff
@@ -67,39 +106,186 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
...
+            Timber.d("Audio focus not granted yet, waiting...")
+            val regained = withTimeoutOrNull(FOCUS_TIMEOUT_MS) {
+                hasFocus.first { it }
+                true
+            } ?: false
```
TimoPtr, line 213: I'm kinda against this, where does this timeout come from? What are you waiting for? Why 10s and not 1h if it is interrupted by a call for instance.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
```diff
@@ -67,39 +106,186 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
...
+            Timber.w(e, "Failed to abandon audio focus")
+        }
+        focusRequest = null
+    }
+
     /**
      * Plays each queued [Utterance] in sequence until [utteranceQueue] is empty.
      * There can be further additions to the queue while a message is playing which will be picked up in the running playback loop.
      */
     private suspend fun play() {
```
TimoPtr, line 196: This function grew too big it needs to be split

common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
```diff
@@ -67,39 +106,186 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
...
                 "Failed to initialize engine.",
             )
             handleError(applicationContext.getString(R.string.tts_error_init))
             utteranceQueue.clear()
+            abandonAudioFocus()
         }.onSuccess {
-            while (utteranceQueue.isNotEmpty()) {
-                utteranceQueue.removeFirst().let { utterance ->
+            try {
+                // Over bluetooth connections, the first syllable or even word can be cut off.
```
TimoPtr, line 234: Do you have an official documentation or a link to an issue about this assumption?

common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
```diff
@@ -67,39 +106,186 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
...
+                        ),
+                    )
+                }
+                while (utteranceQueue.isNotEmpty()) {
+                    if (!hasFocus.value) {
+                        Timber.d("Audio focus lost before playing utterance, waiting...")
+                        val regained = withTimeoutOrNull(FOCUS_TIMEOUT_MS) {
+                            hasFocus.first { it }
+                            true
+                        } ?: false
```
TimoPtr, line 254: Let's avoid duplication.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
```diff
@@ -30,6 +37,38 @@ class TextToSpeechClient(private val applicationContext: Context, private val te
...
+                AudioManager.AUDIOFOCUS_LOSS -> {
+                    Timber.d("Audio focus lost permanently")
+                    hasFocus.value = false
+                    stopTTS()
+                }
+                AudioManager.AUDIOFOCUS_LOSS_TRANSIENT -> {
+                    Timber.d("Audio focus lost temporarily")
+                    hasFocus.value = false
+                    isTransientLoss = true
+                    playbackJob?.cancel()
```
TimoPtr, line 65: So if it's transient you kill the playbackJob? when it restarts it restarts from the beginning?

common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
ChadKillingsworth: I'm really on the fence on this myself honestly. How useful is a delayed notification? How long of a delay until that usefulness is almost zero? In real life, how often will this even occur?

ChadKillingsworth:
Thanks for the review - I'll work on addressing all these comments.

common/src/main/kotlin/io/homeassistant/companion/android/common/util/tts/TextToSpeechClient.kt
ChadKillingsworth: No official docs. I find lots of reddit and stackoverflow posts talking about it. I experienced it 100% of the time on a real device and none of the time on the emmulator (real device was bluetooth - emmulator was port forwarding).

The test messages sent with audio focus logic were:

 > "Notification was sent"

The car head unit would play:

> "cation was sent"

It was very reproducible. Tried multiple things to avoid it and this was the technique that did not rely on magic numbers.

PR #6989, opened 2026-06-10
Introduce ServerUserAvatarUseCase

TimoPtr:
## Summary

This PR introduce a new manager `ServerUserAvatarUseCase` that is not yet use in this PR. This use case contains the logic to retrieve the Avatar image from core for a given server. We ensure to also cache the image whenever we can to avoid downloading it multiple times.

## Checklist

## Any other notes

If the server is not available it doesn't return anything because we need the list of entities to get the cache key. We cannot enumerate the disk cache because the key is transformed to a sha256 so looking at the `filesystem` attribute of the `fileLoader` won't help us to find the cache without knowing the full cache key.

This doesn't leave in :common because it rely on coil.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/util/ServerUserAvatarManager.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/util/ServerUserAvatarManager.kt
@@ -1,0 +1,126 @@
+package io.homeassistant.companion.android.util
+
+import android.content.Context
+import android.graphics.Bitmap
+import android.graphics.BitmapFactory
+import androidx.annotation.VisibleForTesting
+import coil3.ImageLoader
+import coil3.imageLoader
+import coil3.memory.MemoryCache
+import coil3.network.NetworkHeaders
+import coil3.network.httpHeaders
+import coil3.request.ImageRequest
+import coil3.request.allowHardware
+import coil3.toBitmap
+import dagger.hilt.android.qualifiers.ApplicationContext
+import io.homeassistant.companion.android.common.data.integration.isPersonOf
+import io.homeassistant.companion.android.common.data.servers.ServerManager
+import io.homeassistant.companion.android.common.data.servers.firstUrlOrNull
+import javax.inject.Inject
+import kotlin.coroutines.cancellation.CancellationException
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import timber.log.Timber
+
+private const val ATTRIBUTE_ENTITY_PICTURE = "entity_picture"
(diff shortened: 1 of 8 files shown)


jpelgrom:
I think this raises some architecture questions, sorry

 - This looks most like the pair of `*UseCase` classes we have: very specific repository usage with some light handling on top of it to make it easy to use and avoid repeating yourself. Isn't that a manager? You even called it manager in the description.
 - If we go with repository, I don't think we should put repositories in a util package... Probably its own or grouped together with the most closely related repository package.

TimoPtr:
> I think this raises some architecture questions, sorry
> 
>     * This looks most like the pair of `*UseCase` classes we have: very specific repository usage with some light handling on top of it to make it easy to use and avoid repeating yourself. Isn't that a manager? You even called it manager in the description.
> 
>     * If we go with repository, I don't think we should put repositories in a util package... Probably its own or grouped together with the most closely related repository package.

I renamed it from Manager to Repository because in the next PR that uses it I've introduced a `ServerChooserItemsManager` that use the Repository to build the items. If we agree to introduce Use Cases I have this proposal we rename the current repository to Manager and the `ServerChooserItemsManager` to `ServerChooserItemsUseCase`, Use Case becoming something very specific that could be reused into multiple places implementing one specific logic on top of potentially a Manager? Otherwise if we stick to our current doc `ServerChooserItemsHandler`.

jpelgrom:
Naming aside, the code here looks good.

> If we agree to introduce Use Cases I have this proposal we rename the current repository to Manager and the `ServerChooserItemsManager` to `ServerChooserItemsUseCase`, Use Case becoming something very specific that could be reused into multiple places implementing one specific logic on top of potentially a Manager? Otherwise if we stick to our current doc `ServerChooserItemsHandler`.

I don't follow this logic. The code in _this PR_ looks like a use case to me; the `ServerChooserItemsManager` as well or like a util function tbh, there's only 2 lines of actual code in it.

TimoPtr:
> Naming aside, the code here looks good.
> 
> > If we agree to introduce Use Cases I have this proposal we rename the current repository to Manager and the `ServerChooserItemsManager` to `ServerChooserItemsUseCase`, Use Case becoming something very specific that could be reused into multiple places implementing one specific logic on top of potentially a Manager? Otherwise if we stick to our current doc `ServerChooserItemsHandler`.
> 
> I don't follow this logic. The code in _this PR_ looks like a use case to me; the `ServerChooserItemsManager` as well or like a util function tbh, there's only 2 lines of actual code in it.

Reverted to UseCase.

jpelgrom (approved):
Don't forget to update the draft docs PR with architecture to include use case - I think we both intuitively get it but good to write it down.

PR #6994, opened 2026-06-10
Forbid external URI to use not exported app components

TimoPtr:
## Summary

We don't want to allow an external URI to use any of our non exposed components. This PR provided 2 extension to handle this.

## Checklist

### DIFF
--- a/common/src/test/kotlin/io/homeassistant/companion/android/common/util/IntentExtTest.kt
+++ b/common/src/test/kotlin/io/homeassistant/companion/android/common/util/IntentExtTest.kt
@@ -1,0 +1,113 @@
+package io.homeassistant.companion.android.common.util
+
+import android.content.ComponentName
+import android.content.Context
+import android.content.Intent
+import android.content.pm.ActivityInfo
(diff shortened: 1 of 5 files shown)


common/src/test/kotlin/io/homeassistant/companion/android/common/util/IntentExtTest.kt
```diff
@@ -0,0 +1,113 @@
...
+        val intent = Intent().setComponent(component)
+
+        intent.stripSelfNonExportedTarget(context)
+
+        assertEquals(component, intent.component)
+        assertNull(failFastThrowable, "no redirection so FailFast must not fire")
+    }
+
+    @Test
+    fun `Given intent targeting another app non-exported component when stripped then component is kept`() {
```
jpelgrom:
```suggestion
    fun `Given intent targeting another app non-exported component when stripped then intent is left untouched`() {
```

It is basically the same as the next test, you're checking everything in the intent matches what you put in

jpelgrom:
Only one minor nitpick, looks good otherwise!

PR #6995, opened 2026-06-10
Capture schema/port redirection in ConnectionScreen

### LINKED ISSUE #6991: Check if the port changed from the one we used to open the ConnectionScreen to a new one
### Description

In the scope of migrating away from 8123 for new install to 80 the app should be able to detect that a redirection did happen in the onboarding after seeing the landing page. 

Scenario
- Device on landing page downloading core exposing over mDNS http://IP:8123
- App open landing page and wait for core to finish downloading
- landing page redirect page to core onboarding port 80 http://IP:80
- App need to detect this redirection and use this new URL with the new port updated (the host needs to stay the same if it differ stick to the initial URL)
- App store URL for usage

### Additional context

_No response_

TimoPtr:
## Summary

We are in the discussion of having the default port changed from 8123 to 80 for core, but this comes with some issues with the [landing page ](https://github.com/home-assistant/landingpage/) that is going to expose mDNS external URL with port 8123. We decided that core is going to handle the redirection from 8123 to 80 but on the mobile app we will need to detect this redirection so that we store the right port, since core is going then to remove this redirection. To do so the `ConnectionViewModel` is keeping track of the URL returned by the WebView and detect any changes of schema/port and persist this change before navigating away from the screen.

This is not yet implemented in core but we need to be ready and this deploy ASAP so that when it is release the app is already ready for this behavior. Otherwise user might experience a full onboarding and end up into an unreachable dashboard and they would need to onboard again or manually change the port in the external URL setting.

Closes #6991 

## Checklist

## To test
I simply spawn a nginx docker that do a redirection to my instance like this 

```
http {
    server {
        listen 8111;

        # Dumb landing-page simulation: redirect everything to the real Home Assistant
        # on its real port. $request_uri keeps the path + query string, so the auth flow
        # (/auth/authorize?...) continues seamlessly on the new port.
        #
        # 302 (temporary) on purpose: it is a server-side redirect (the case the app must
        # detect via onPageFinished) and it is not cached by the WebView.
        location / {
            return 302 http://server_ip:8123/$request_uri;
        }
    }
}
```

The external IP should have the port 8123 while you've entered the port 8111 manually.

### DIFF
--- a/app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
+++ b/app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
@@ -205,7 +205,8 @@ class ConnectionViewModelTest {
         turbineScope {
             val navigationEventsFlow = viewModel.navigationEventsFlow.testIn(backgroundScope)
 
-            // Landing page redirects the WebView to the core on a new port, same host
+            // Landing page redirects the WebView to the core on a new port, same host.
+            // Port 80 is http's default, so the normalized stored URL drops it.
             viewModel.webViewClient.onPageFinished(null, "http://homeassistant.local:80/onboarding")
 
             val result = viewModel.webViewClient.shouldOverrideUrlLoading(null, stringUri)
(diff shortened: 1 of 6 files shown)


app/src/test/kotlin/io/homeassistant/companion/android/onboarding/connection/ConnectionViewModelTest.kt
```diff
@@ -190,6 +190,138 @@ class ConnectionViewModelTest {
...
+            val result = viewModel.webViewClient.shouldOverrideUrlLoading(null, stringUri)
+
+            assertTrue(result)
+            val event = navigationEventsFlow.awaitItem()
+            assertEquals("http://[::1]", (event as ConnectionNavigationEvent.Authenticated).url)
+        }
+    }
+
+    @Test
+    fun `Given redirect to a different host when auth completes then Authenticated url stays the initial url`() = runTest {
```
jpelgrom, line 300: If there is a redirect to a different host you're sent to the browser, so does `onPageFinished` ever get called?

jpelgrom:
> we need to be ready and this deploy ASAP so that when it is release the app is already ready for this behavior

We can talk to the core team if the app has not yet had a release to postpone the change by another month, I suppose?

jpelgrom:
Following a short discussion: please check to make sure that the redirections allowed for `onPageFinished` match what is allowed (not sending you to the browser) in `interceptRedirectIfRequired`. If only port is expected I'd limit it to port, if http -> https upgrades might happen in the future also schema.

TimoPtr:
> Following a short discussion: please check to make sure that the redirections allowed for `onPageFinished` match what is allowed (not sending you to the browser) in `interceptRedirectIfRequired`. If only port is expected I'd limit it to port, if http -> https upgrades might happen in the future also schema.

At some point in the future yes we hope to be able to use HTTPS so it might happen that the landing page redirect from HTTP to HTTPS or even if the user enter HTTP URL core could decide to upgrade the connection to HTTPS in that case we should store the HTTPS not HTTP.

Not sure if it's going to answer your question what I've tested so far
- redirecting to another HOST+PORT continue to open an external link, going back to the app show a blank screen if the WebView did not open anything. I'm not sure if we should handle this case.
- redirecting to another PORT does keep the user in the ConnectionScreen and store the new port

jpelgrom:
> - redirecting to another HOST+PORT continue to open an external link, going back to the app show a blank screen if the WebView did not open anything. I'm not sure if we should handle this case.

We shouldn't, in my opinion. Opening another app should be clear enough that this isn't supported, so going back and seeing one blank screen to indicate the app won't load that one, before pressing back again, is OK.

jpelgrom (approved):

bgoncal (approved):

PR #7000, opened 2026-06-11
Break zone radius sort ties by distance to zone center

### LINKED ISSUE #6999: Sort "in_zones" based on distance to the center of a zone on top of the existent zone size sorting
### Description

To improve the communication of "in_zones" to mobile_app integration the app needs to sort the array of zones based on the distance from the user to the center of the zone and not only the zone size sorting alone.

### Additional context

_No response_

TimoPtr:
## Summary

Closes #6999. Use location distance to the zone center when the zone radius are equivalent.

## Checklist

## Any other notes

Implemented in core https://github.com/home-assistant/core/pull/173106

### DIFF
--- a/app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
+++ b/app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
@@ -939,7 +939,12 @@ class LocationSensorManager :
                 .filter {
                     val radius = it.attributes["radius"] as? Number
                     return@filter radius != null && it.containsWithAccuracy(location)
-                }.sortedBy { (it.attributes["radius"] as? Number ?: Int.MAX_VALUE).toFloat() }
+                }.sortedWith(
+                    // Smallest zone (radius) first; when two zones share the same radius, prefer the one
+                    // whose center is closest to the current location to break the tie deterministically.
+                    compareBy<Entity> { (it.attributes["radius"] as? Number ?: Int.MAX_VALUE).toFloat() }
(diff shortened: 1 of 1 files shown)


jpelgrom:
I thought you wanted to make more changes but it looks like you only didn't click the review button @TimoPtr :D

jpelgrom (approved):
Matches core architecture decision

TimoPtr:
> I thought you wanted to make more changes but it looks like you only didn't click the review button @TimoPtr :D

I was waiting for core to validate that it was the way to go before asking for review:)

signaleleven:
I landed on this PR because my wife's phone started flip flopping between two zones (which admittedly are relatively close to each other). It started doing it 3 days ago (July 8th, even if the app was updated on June 20th - so this change alone isn't the smoking gun, but it seems related).
The app was also reporting locations with a radius larger than the configured one (default 200 meters, I lowered it to 70 on both the location sensor and the zone sensor, and it kept reporting larger uncertainties - though it stopped, apparently, after a cache clean and restart, potentially solving my issue, TBC), ending up with two zones in the "in_zones" attribute, causing the jumping around whenever the smaller - but further - zone was in the list. 

I'm putting this comment here in case somebody else is in the middle of troubleshooting something like this, no action required from anybody for now :) 

...But tie break logic seems wrong to me. Distance is more important than radius. Often the center of a wide radius location report is still quite accurate, at least in my experience. If I understand the logic correctly, a location update that includes 2 zones, one small but far from the center of the location update, and one bigger but close to the center (which should arguably win) would make the small zone win.

PR #7001, opened 2026-06-11
Improve back behavior of FrontendScreen

TimoPtr:
## Summary

The back behavior on the `FrontendScreen` was wrong in few places alongside with the clear of the history of the WebView while changing server.

- While coming from the onboarding the onboarding could have still be in the navigation back stack
- Predictive back was not working properly because we were using the navController instead of letting the system handle it
- While on overlay like error screen the back behavior was wrong
- Missing documentation on the overlays on how the back is handled

## Checklist

## Any other notes

I tried to test some scenario that I've think off but I might have missed some.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
@@ -103,7 +103,6 @@ internal const val CUSTOM_VIEW_OVERLAY_TAG = "custom_view_overlay"
  * The WebView is always rendered at the base layer to prevent it to not load the URL.
  * Loading indicators, error screens, and blocking screens are overlaid on top.
  *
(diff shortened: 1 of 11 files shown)


jpelgrom:
There is a surprising amount of lines changed here for not that much difference.

Do you want to update this for #6836 or should this go in first? Both will conflict and need additional changes and I think it's easier for you to make changes...

TimoPtr:
> There is a surprising amount of lines changed here for not that much difference.
> 
> Do you want to update this for #6836 or should this go in first? Both will conflict and need additional changes and I think it's easier for you to make changes...

I was surprised that the predictive back was not working and found that I simply didn't make it properly. 

Let's get this first I didn't look at the URL path change for compose yet I wanted to make it right first.

jpelgrom (approved):
Despite pushing for it on the other PR, I never noticed the back handling didn't work properly here 😳. Thanks for the fix!

PR #7007, opened 2026-06-12
Migrate entity widget configuration to Compose

### LINKED ISSUE #6307: Rework UI of EntityWidgetConfigureActivity to compose and Material3

### LINKED ISSUE #7230: Migrate EntityWidgetConfigureActivity to compose and EntityPicker
### Description

We want to get rid of the usage of friendly name, we have a new compose EntityPicker but it is compose only. The EntityWidgetConfigureActivity needs to be refactor to compose so it can use the EntityPicker and stop relying on the friendly name.

### Additional context

_No response_

himu-gupta:
## Summary
- migrate `EntityWidgetConfigureActivity` from XML views to Compose and Home Assistant Material 3 components
- move configuration state, validation, creation, and update persistence into an assisted ViewModel
- preserve server selection, custom attributes, tap actions, widget themes, text colors, and existing widget values
- add focused ViewModel tests for restoration, recreation, persistence, and server changes

Closes #6307 Based on #7236

## Screenshots

image

image


app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,312 @@
...
+                    System.currentTimeMillis().toInt(),
+                    Intent(context, EntityWidget::class.java).apply {
+                        action = ACTION_APPWIDGET_CREATED
+                        putExtra(EXTRA_WIDGET_ENTITY, getPendingDaoEntity())
+                    },
+                    PendingIntent.FLAG_MUTABLE,
+                ),
+            )
+        }.first()
+    }
```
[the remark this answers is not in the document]
himu-gupta: Fixed in f9669f503. The code now checks isRequestPinAppWidgetSupported and verifies the Boolean returned by requestPinAppWidget, failing through the existing IllegalStateException/error-toast path when unsupported or rejected. I kept the existing project behavior for user cancellation: the coroutine waits for the DAO success signal and is cancelled with the Activity lifecycle, matching BaseWidgetConfigureActivity and TodoWidgetConfigureViewModel rather than introducing an arbitrary timeout.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModelTest.kt
```diff
@@ -0,0 +1,180 @@
...
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertFalse
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.RegisterExtension
+
+@OptIn(ExperimentalCoroutinesApi::class)
+private class EntityWidgetConfigureViewModelTest {
```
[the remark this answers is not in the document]
himu-gupta: Fixed in f9669f503. The test class is now non-private so discovery does not depend on reflective access to a private top-level class.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,423 @@
...
+                    label = stringResource(commonR.string.widget_text_color_label),
+                )
+            }
+
+            HAAccentButton(
+                text = stringResource(if (isUpdateWidget) commonR.string.update_widget else commonR.string.add_widget),
+                onClick = onActionClick,
+                enabled = selectedEntityId != null,
+                modifier = Modifier.fillMaxWidth(),
+            )
```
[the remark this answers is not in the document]
himu-gupta: This was addressed in 91c955b8e by removing the partial selectedEntityId-only enablement. The legacy screen keeps the action enabled and validates on click, and the emulator before/after comparison showed that disabling it was an unintended behavior change. The action still uses the ViewModel validity check and existing error path.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,312 @@
...
+import io.homeassistant.companion.android.database.widget.WidgetTapAction
+import io.homeassistant.companion.android.widgets.ACTION_APPWIDGET_CREATED
+import io.homeassistant.companion.android.widgets.BaseWidgetProvider
+import io.homeassistant.companion.android.widgets.EXTRA_WIDGET_ENTITY
+import kotlin.time.Duration.Companion.milliseconds
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.WhileSubscribed
```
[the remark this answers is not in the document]
himu-gupta: This import is required: WhileSubscribed supplies the Duration overload used by SharingStarted.WhileSubscribed(500.milliseconds), matching TodoWidgetConfigureViewModel. :app:ktlintCheck passes with the import present.

himu-gupta:
The CLA check is now passing, but the original Home Assistant CLA bot review still shows **Changes requested** and keeps the PR in a blocked state. Could a maintainer please dismiss the stale review when convenient? All review threads are resolved, the branch is rebased onto the latest `main`, and the focused tests, ktlint, and full debug build pass.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureActivity.kt
```diff
@@ -53,335 +38,96 @@
...
-            // Not implemented
-        }
-
-        override fun onTextChanged(s: CharSequence?, start: Int, before: Int, count: Int) {
-            // Not implemented
-        }
-
-        override fun afterTextChanged(s: Editable?) {
-            labelFromEntity = false
+    @RequiresApi(Build.VERSION_CODES.O)
```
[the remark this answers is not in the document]
himu-gupta: Fixed in 540477b21. I removed the redundant @RequiresApi annotation while retaining the SdkVersion.isAtLeast(API 26) runtime guard before widget pinning. The focused :app:lintFullDebug check now reports no new issues.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,316 @@
...
+                    EntityWidgetTextColor.WHITE -> textColors.white
+                    EntityWidgetTextColor.BLACK -> textColors.black
+                }
+            } else {
+                null
+            },
+        )
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
```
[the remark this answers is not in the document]
himu-gupta: Fixed in 540477b21. I removed the redundant @RequiresApi annotation and used a targeted NewApi suppression on the ViewModel method because its API-26 calls are reached only through the Activity runtime guard, which Android Lint cannot propagate across the method boundary. :app:lintFullDebug, :app:ktlintCheck, the focused ViewModel unit test, and :app:assembleFullDebug all pass.

himu-gupta:
> Hi @himu-gupta
> 
> It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant/android#7007).
> 
> 
> Thanks!

CLA was signed.

TimoPtr:
Could you update the PR description to add screenshots, thanks.

himu-gupta:
Added light and dark screenshots to the PR description.

TimoPtr:
The UI does already look much better than from the old one.

image

Things to address
- Should be centered
- The picker should be aligned with the size of the inputs
- Clicking Append attributes could animate nicely the hidden field
- The `Attribute`menu could/should be merge with the `Add attribute` input, to improve the UX
- Changing the entity selected doesn't update the label anymore
- Server picker is using the old style
image


Things to change/discuss
- Update widget button logic should be improved (in another PR), to include validation on inputs today you can input a huge number for instance in the Widget text size.
- We could add a top left close button in the topbar

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,422 @@
...
+        dynamicColorAvailable = dynamicColorAvailable,
+        selectedTextColor = viewModel.selectedTextColor,
+        onTextColorSelected = viewModel::onTextColorSelected,
+        isUpdateWidget = viewModel.isUpdateWidget,
+        onActionClick = onActionClick,
+    )
+}
+
+@Composable
+private fun EntityWidgetConfigureView(
```
TimoPtr: Let's add some screenshotTest and compose test (test the behavior of all the interaction without needing a ViewModel) for the screen.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureActivity.kt
```diff
@@ -53,335 +37,95 @@ class EntityWidgetConfigureActivity : BaseWidgetConfigureActivity<StaticWidgetEn
...
-                        commonR.color.colorWidgetButtonLabelBlack
-                    },
-                )
-            } else {
-                null
-            },
-            lastUpdate = dao.get(appWidgetId)?.lastUpdate ?: "",
-        )
+    private fun showWidgetError(message: Int) {
+        Toast.makeText(applicationContext, message, Toast.LENGTH_LONG).show()
```
TimoPtr: Let's use a proper Snackbar for error instead of Toasts

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,422 @@
...
+    selectedTapAction: WidgetTapAction,
+    onTapActionSelected: (WidgetTapAction) -> Unit,
+    selectedBackgroundType: WidgetBackgroundType,
+    onBackgroundTypeSelected: (WidgetBackgroundType) -> Unit,
+    dynamicColorAvailable: Boolean,
+    selectedTextColor: EntityWidgetTextColor,
+    onTextColorSelected: (EntityWidgetTextColor) -> Unit,
+    isUpdateWidget: Boolean,
+    onActionClick: () -> Unit,
+) {
```
TimoPtr, line 128: You should split this into small Composable check recent screens like `AssistSettingsScreen` if you want to see how we usually, or the whole onboarding.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,422 @@
...
+    onAttributeRemoved: (String) -> Unit,
+) {
+    val unselectedAttributes = availableAttributes.filterNot(selectedAttributeIds::contains)
+    var customAttribute by rememberSaveable { mutableStateOf("") }
+    val addCustomAttributes = {
+        customAttribute
+            .split(',')
+            .map(String::trim)
+            .filter(String::isNotEmpty)
+            .forEach(onAttributeAdded)
```
TimoPtr: Let's try to avoid logic in the UI layer and move this into the VM.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,315 @@
...
+    var attributeSeparator by mutableStateOf("")
+        private set
+    var selectedTapAction by mutableStateOf(WidgetTapAction.REFRESH)
+        private set
+    var selectedBackgroundType by mutableStateOf(WidgetBackgroundType.DAYNIGHT)
+        private set
+    internal var selectedTextColor by mutableStateOf(EntityWidgetTextColor.WHITE)
+        private set
+    var isUpdateWidget by mutableStateOf(false)
+        private set
```
TimoPtr: Let's use a ViewState instead of all this state.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    fun onLabelChanged(value: String) {
+        label = value
+        labelFromEntity = false
+    }
+
+    fun onTextSizeChanged(value: String) {
+        textSize = value.filter(Char::isDigit)
+    }
```
TimoPtr, line 192: You could handle validation from there and update the viewState to show an error if it's not valid. Also have in the ViewState the logic that disable the add/update widget button based on invalid inputs.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,315 @@
...
+
+    fun onBackgroundTypeSelected(backgroundType: WidgetBackgroundType) {
+        selectedBackgroundType = backgroundType
+    }
+
+    internal fun onTextColorSelected(textColor: EntityWidgetTextColor) {
+        selectedTextColor = textColor
+    }
+
+    fun selectedEntity(): Entity? = entities.value.firstOrNull { it.entityId == selectedEntityId }
```
TimoPtr: Let's avoid doing any heavy operation like this in the Main thread it should be made in DIspatcher.Default. Users with a lot of entities might be impacted by this.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,422 @@
...
+                (isUpdateWidget && servers.none { it.id == selectedServerId })
+            ) {
+                ServerExposedDropdownMenu(
+                    servers = servers,
+                    current = selectedServerId,
+                    onSelected = onServerSelected,
+                )
+            }
+
+            EntityPicker(
```
TimoPtr: Provides all the registries to the picker to have better name displayed while picking an entity.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,315 @@
...
+            } else {
+                null
+            },
+        )
+    }
+
+    @SuppressLint("NewApi") // The activity calls this only after its API 26 runtime check.
+    suspend fun requestWidgetCreation(context: Context) {
+        val appWidgetManager = AppWidgetManager.getInstance(context)
+        check(appWidgetManager.isRequestPinAppWidgetSupported) { "Widget pinning is not supported" }
```
TimoPtr: This is going to crash

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,315 @@
...
+                    },
+                    PendingIntent.FLAG_MUTABLE,
+                ),
+            )
+            check(requestAccepted) { "Widget pin request was rejected" }
+        }.first()
+    }
+
+    fun updateWidget(context: Context) {
+        context.applicationContext.sendBroadcast(
```
TimoPtr:
```suggestion
        context.sendBroadcast(
```
Why forcing the appContext?

app/src/test/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModelTest.kt
```diff
@@ -0,0 +1,180 @@
...
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.RegisterExtension
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class EntityWidgetConfigureViewModelTest {
+
+    @RegisterExtension
+    val mainDispatcherExtension = MainDispatcherJUnit5Extension(UnconfinedTestDispatcher())
```
TimoPtr: Use `@ExtendWith(MainDispatcherJUnit5Extension::class)`, your test should be made on the StandardTestDispatcher set in the extension. If you use the unconfined it's probably because something is wrong.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModelTest.kt
```diff
@@ -0,0 +1,180 @@
...
+        every { serverManager.serversFlow } returns flowOf(emptyList())
+        coEvery { serverManager.isRegistered() } returns true
+        coEvery { serverManager.integrationRepository(any()) } returns integrationRepository
+        coEvery { integrationRepository.getEntities() } returns listOf(entity)
+        coEvery { serverManager.getServer(any<Int>()) } returns server
+        coEvery { dao.get(any()) } returns null
+    }
+
+    @Test
+    fun `Given an existing widget when setup completes then persisted configuration is restored`() = runTest(mainDispatcherExtension.testDispatcher) {
```
TimoPtr:
```suggestion
    fun `Given an existing widget when setup completes then persisted configuration is restored`() = runTest {
```

TimoPtr (changes requested):
I did a first very quick review of the code to identify the missing bits.

himu-gupta:
Added a review-feedback screenshot to the PR description showing the updated centered form with the modern server picker and append-attributes section expanded, including the merged Add attribute input/chips and entity-derived label.

himu-gupta:
Added a dedicated server picker fix screenshot to the PR description. This one matches the dark top-of-screen state from your attached image: the server picker now uses the HA dropdown style and is aligned with the selected entity picker.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
himu-gupta: Applied this suggestion in eca4235d2: update broadcasts now use the Activity/context passed into updateWidget instead of forcing applicationContext.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModelTest.kt
himu-gupta: Updated in eca4235d2 to use @ExtendWith(MainDispatcherJUnit5Extension::class), so the tests run with the extension-provided StandardTestDispatcher.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModelTest.kt
himu-gupta: Updated in eca4235d2. The test now uses runTest with advanceUntilIdle() so setup completion is driven explicitly.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
himu-gupta: Fixed in 6691e0e26. The configure ViewModel now loads entity, device, and area registries for the selected server and passes them through to EntityPicker so entity names/metadata can be displayed consistently with the other picker usages.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
himu-gupta: Fixed in 6691e0e26. requestWidgetCreation now checks the SDK level inside the ViewModel method before touching API 26 widget pinning APIs, and handles RemoteException while reading isRequestPinAppWidgetSupported so this fails through the existing error path instead of crashing.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
himu-gupta: Fixed in e483268a0. Added EntityWidgetConfigureScreenTest for screen interactions without the ViewModel, plus EntityWidgetConfigureScreenshotTest with HAPreviews and generated FullDebug references. The filtered entity-widget screenshot validation passes.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureActivity.kt
himu-gupta: Fixed in e483268a0. Error events now flow through the ViewState and the Compose screen shows them with a Material3 SnackbarHost instead of Activity Toasts.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
himu-gupta: Fixed in e483268a0. EntityWidgetConfigureView now delegates to smaller sections for server selection, entity picker, attributes, text options, tap action, appearance, and the action button.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
himu-gupta: Fixed in e483268a0. Custom attribute parsing/addition moved into the ViewModel; the composable only forwards the text and add events.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
himu-gupta: Fixed in e483268a0. The individual Compose state fields are now represented by EntityWidgetConfigureViewState, with compatibility getters kept on the ViewModel for existing tests/callers.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
himu-gupta: Fixed in e483268a0. ViewState now owns text-size validity and action-button enabled state, and invalid create/update errors are represented in ViewState for snackbar display.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
himu-gupta: Fixed in e483268a0. The selected-entity lookup and attribute sorting now run on Dispatchers.Default, and the ViewModel no longer scans the full entity list from onEntitySelected.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,624 @@
...
+    onStateSeparatorChanged: (String) -> Unit,
+    onAttributeSeparatorChanged: (String) -> Unit,
+    isToggleable: Boolean,
+    onTapActionSelected: (WidgetTapAction) -> Unit,
+    onBackgroundTypeSelected: (WidgetBackgroundType) -> Unit,
+    dynamicColorAvailable: Boolean,
+    onTextColorSelected: (EntityWidgetTextColor) -> Unit,
+    onErrorShown: () -> Unit,
+    onActionClick: () -> Unit,
+) {
```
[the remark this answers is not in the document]
himu-gupta: Fixed in be29796c6. Reordered EntityWidgetConfigureView parameters to match Compose lint ordering, with optional registry parameters moved after the non-default callbacks. :app:compileFullDebugKotlin, :app:lintFullDebug, and :app:ktlintCheck pass.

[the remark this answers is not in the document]
himu-gupta: Fixed in be29796c6. Reordered EntityWidgetConfigureView parameters to match Compose lint ordering, with optional registry parameters moved after the non-default callbacks. :app:compileFullDebugKotlin, :app:lintFullDebug, and :app:ktlintCheck pass.

TimoPtr:
I think we could remove the "Append attribute value" checkbox by only showing the tags to pick. Does it make sense to even allow custom attributes? @jpelgrom wdyt

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureActivity.kt
```diff
@@ -53,335 +36,95 @@ class EntityWidgetConfigureActivity : BaseWidgetConfigureActivity<StaticWidgetEn
...
+            }
+        },
+    )
 
-    private var entityAdapter: SingleItemArrayAdapter<Entity>? = null
+    private val requestLauncherSetup: Boolean
+        get() = intent.extras?.getBoolean(ManageWidgetsViewModel.CONFIGURE_REQUEST_LAUNCHER, false) == true
 
     override fun onCreate(savedInstanceState: Bundle?) {
+        enableEdgeToEdgeCompat()
```
TimoPtr:
```suggestion
```

It's already in BaseActivity

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+    val appendAttributes: Boolean get() = viewState.appendAttributes
+    val selectedAttributeIds: List<String> get() = viewState.selectedAttributeIds
+    val label: String get() = viewState.label
+    val textSize: String get() = viewState.textSize
+    val stateSeparator: String get() = viewState.stateSeparator
+    val attributeSeparator: String get() = viewState.attributeSeparator
+    val selectedTapAction: WidgetTapAction get() = viewState.selectedTapAction
+    val selectedBackgroundType: WidgetBackgroundType get() = viewState.selectedBackgroundType
+    internal val selectedTextColor: EntityWidgetTextColor get() = viewState.selectedTextColor
+    val isUpdateWidget: Boolean get() = viewState.isUpdateWidget
```
TimoPtr: Let's just expose the ViewState.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+    @Assisted preselectedEntityId: String?,
+) : ViewModel() {
+
+    private lateinit var textColors: EntityWidgetTextColors
+    private var initialized = false
+
+    internal var widgetId: Int = AppWidgetManager.INVALID_APPWIDGET_ID
+        private set
+
+    val servers = serverManager.serversFlow
```
TimoPtr: Keep this in the ViewState, and make it already a `HADropdownItem` so mapping happen in the viewModel scope. Same for the selected server it needs to be in the ViewState.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+
+    internal var viewState by mutableStateOf(EntityWidgetConfigureViewState(selectedEntityId = preselectedEntityId))
```
TimoPtr: use a mutableStateFlow

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+
+internal enum class EntityWidgetTextColor {
+    WHITE,
+    BLACK,
+}
+
+internal enum class EntityWidgetConfigureError {
+    CREATE,
+    UPDATE,
+}
```
TimoPtr: This is something that is common to all the widget configure screen.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+    private val serverManager: ServerManager,
+    @Assisted preselectedEntityId: String?,
+) : ViewModel() {
+
+    private lateinit var textColors: EntityWidgetTextColors
+    private var initialized = false
+
+    internal var widgetId: Int = AppWidgetManager.INVALID_APPWIDGET_ID
+        private set
+
```
TimoPtr: While looking at this I think we could pass the widgetId into the constructor like `preselectedEntityId` and even potentially remove the need of `onSetup` method since the other parameters can already be retrieved from the ViewModel and store the ID not the color and pass the context only when calling updateWidgetConfiguration to create the StaticWidgetEntity.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+                } catch (e: CancellationException) {
+                    throw e
+                } catch (e: Exception) {
+                    Timber.e(e, "Failed to get area registry")
+                    null
+                }
+            }
+        }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(500.milliseconds), null)
+
+    private var labelFromEntity = false
```
TimoPtr: Why do you need this?

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+    }
+
+    fun onLabelChanged(value: String) {
+        viewState = viewState.copy(label = value)
+        labelFromEntity = false
+    }
+
+    fun onTextSizeChanged(value: String) {
+        viewState = viewState.copy(textSize = value.filter(Char::isDigit))
+    }
```
TimoPtr, line 192: What about adding a validation here with a proper error explaining to the user the acceptable values on this field.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+            selectedEntity() != null
+    }
+
+    suspend fun updateWidgetConfiguration() {
+        check(widgetId != AppWidgetManager.INVALID_APPWIDGET_ID) { "Widget ID is invalid" }
+        staticWidgetDao.add(getPendingDaoEntity())
+    }
+
+    internal suspend fun getPendingDaoEntity(): StaticWidgetEntity {
+        check(isValidSelection()) { "Widget data is invalid" }
```
TimoPtr: Using check would throw an exception but you've made it that the catch is ignoring the exception, so we never see in the logs the actual message that you wrote in the `check*`. I would advise to return null instead of using check and log the issue here instead.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+    }
+
+    suspend fun isValidSelection(): Boolean {
+        return viewState.isActionEnabled &&
+            serverManager.getServer(selectedServerId) != null &&
+            selectedEntity() != null
+    }
+
+    suspend fun updateWidgetConfiguration() {
+        check(widgetId != AppWidgetManager.INVALID_APPWIDGET_ID) { "Widget ID is invalid" }
```
TimoPtr: Same for all the checks

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+                }
+            } else {
+                null
+            },
+        )
+    }
+
+    @SuppressLint("NewApi") // The activity calls this only after its API 26 runtime check.
+    suspend fun requestWidgetCreation(context: Context) {
+        check(SdkVersion.isAtLeast(Build.VERSION_CODES.O)) { "Widget pinning is not supported" }
```
TimoPtr: Same here proper logs + early  return

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+        val appWidgetManager = AppWidgetManager.getInstance(context)
+        val pinningSupported = try {
+            appWidgetManager.isRequestPinAppWidgetSupported
+        } catch (e: RemoteException) {
+            Timber.e(e, "Unable to read isRequestPinAppWidgetSupported")
+            false
+        }
+        check(pinningSupported) { "Widget pinning is not supported" }
+
+        staticWidgetDao.getWidgetCountFlow().drop(1).onStart {
```
TimoPtr:
```suggestion
		// We drop the first value since we only care about knowing when the widget is actually added
        staticWidgetDao.getWidgetCountFlow().drop(1).onStart {
```

app/src/screenshotTestFullDebug/reference/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreenshotTest/EntityWidgetConfigureView_tablet_2f22c4ea_0.png
TimoPtr, line 1: I think there is something wrong with the color for the friendly_name

app/src/test/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreenTest.kt
```diff
@@ -0,0 +1,141 @@
...
+                            viewState = viewState.copy(
+                                selectedAttributeIds = viewState.selectedAttributeIds + attributes,
+                                customAttribute = "",
+                            )
+                        },
+                        onLabelChanged = {},
+                        onTextSizeChanged = {},
+                        onStateSeparatorChanged = {},
+                        onAttributeSeparatorChanged = {},
+                        isToggleable = true,
```
TimoPtr: You are missing test that verify that these callbaks are invoked properly. Check all the attributes of the View. I would advise to create a small helper method in the test file.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,624 @@
...
+            )
+        }
+    }
+    val selectedEntity = selectedEntityData.entity
+
+    LaunchedEffect(selectedEntity?.entityId, selectedEntity?.friendlyName) {
+        viewModel.onSelectedEntityLoaded(selectedEntity)
+    }
+
+    EntityWidgetConfigureView(
```
TimoPtr:
```suggestion
    EntityWidgetConfigureContent(
```
Let's try to avoid using View

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,624 @@
...
+    HADropdownMenu(
+        items = listOf(
+            HADropdownItem(
+                key = WidgetTapAction.TOGGLE,
+                label = stringResource(commonR.string.widget_tap_action_toggle),
+            ),
+            HADropdownItem(
+                key = WidgetTapAction.REFRESH,
+                label = stringResource(commonR.string.refresh),
+            ),
```
TimoPtr: Let's remember the list to avoid recreating it on every recomposition

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,624 @@
...
+@Composable
+private fun AppearanceSection(
+    selectedBackgroundType: WidgetBackgroundType,
+    dynamicColorAvailable: Boolean,
+    selectedTextColor: EntityWidgetTextColor,
+    onBackgroundTypeSelected: (WidgetBackgroundType) -> Unit,
+    onTextColorSelected: (EntityWidgetTextColor) -> Unit,
+) {
+    HADropdownMenu(
+        items = buildList {
```
TimoPtr: same here remember

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,624 @@
...
+        },
+        selectedKey = selectedBackgroundType,
+        onItemSelected = onBackgroundTypeSelected,
+        label = stringResource(commonR.string.widget_background_type_label),
+        modifier = Modifier.formControlWidth(),
+    )
+
+    if (selectedBackgroundType == WidgetBackgroundType.TRANSPARENT) {
+        HADropdownMenu(
+            items = listOf(
```
TimoPtr: same

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,624 @@
...
+@Composable
+private fun ActionButton(isUpdateWidget: Boolean, enabled: Boolean, onActionClick: () -> Unit) {
+    HAAccentButton(
+        text = stringResource(
+            if (isUpdateWidget) commonR.string.update_widget else commonR.string.add_widget,
+        ),
+        onClick = onActionClick,
+        modifier = Modifier
+            .formControlWidth()
+            .testTag(ENTITY_WIDGET_ACTION_BUTTON_TAG),
```
TimoPtr: Let's not use TAG when you can use the text

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,624 @@
...
+            maxLines = 1,
+        )
+
+        if (unselectedAttributes.isNotEmpty()) {
+            FlowRow(
+                horizontalArrangement = Arrangement.spacedBy(HADimens.SPACE2),
+                verticalArrangement = Arrangement.spacedBy(HADimens.SPACE2),
+            ) {
+                unselectedAttributes.forEach { attributeId ->
+                    InputChip(
```
TimoPtr: Since it's a M3 element it has the wrong colors, you need to use the HAColors (like you did for the the selected one ) so it gets the proper colors for light/dark and not the one from m3. The text needs to use a proper style too. 

You could open a first PR that makes a HAInputChip for instance.

TimoPtr (dismissed):
Just to let you know that we don't allow usage of autonomous agent, take the time to look at all my comments.

TimoPtr:
Also I would like to see a close button in the topbar

TimoPtr:
@himu-gupta I did took over your PR rebased and address most of my comments.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,413 @@
...
+                    null,
+                    PendingIntent.getBroadcast(
+                        context,
+                        System.currentTimeMillis().toInt(),
+                        Intent(context, EntityWidget::class.java).apply {
+                            action = ACTION_APPWIDGET_CREATED
+                            putExtra(EXTRA_WIDGET_ENTITY, widget)
+                        },
+                        PendingIntent.FLAG_MUTABLE,
+                    ),
```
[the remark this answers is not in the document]
TimoPtr: For now we are fine using this here.

jpelgrom (approved):
Thanks for picking this up! Huge improvement.

I'd like to remove the colons from the inputs as they don't belong in these M2/M3 style input boxes, but as the strings are shared among multiple screens we'll do that later.

jpelgrom:
Also for the record: the close/reopen 5 days ago was automatically because the base branch was deleted, not intentional.

PR #7024, opened 2026-06-15
Fix server certificate validation while honoring user-installed CAs

### LINKED ISSUE #6810: ssl handshake failed
2026.5.1-full

**Android version(s):**
16 (but tested on other, and that same problem)

**Device model(s):**
24 ultra, s20fe, note 10 lite

**Home Assistant version:**
Core
2026.4.4
Supervisor
2026.04.2
Operating System
17.2
Interfejs użytkownika
20260325.8

I’m experiencing an issue with the latest version of the Home Assistant Android app where it is no longer possible to log in. The app fails during connection with the error:

SSL handshake failed

Details:

The SSL certificate is valid and correctly configured.
Access via web browser works without any issues.
The iOS app is also able to connect and log in successfully using the same server and certificate.
The issue occurs on multiple Android devices.
The problem also affects two separate Home Assistant servers located in different environments.

**Companion App Logs:**

```

```

**Additional information:**

bpacholek:
## Summary

Restores correct server-certificate validation that regressed in #6753, while keeping support for user-installed CAs.

#6753 switched the OkHttp trust manager from Android's default (`TrustManagerFactory.init(null)`) to one built from `AndroidCAStore`. Handing the factory an explicit keystore drops Android's `network_security_config`-aware validation and chain building, so otherwise-valid server certificates started failing the TLS handshake (#6810).

This change:

- Uses Android's default trust manager again as the primary path. On standard devices that already trusts user-installed CAs via `network_security_config.xml` (which trusts `<certificates src="user"/>`), so it fixes #6810 without losing what #6753 set out to do.
- Adds a small `CompositeX509ExtendedTrustManager` that is consulted only when the default rejects a certificate, and that trusts only the user-installed CAs. This covers ROMs that don't surface user-installed CAs through the default path (e.g. /e/OS #5565, and some HyperOS builds). It can only add acceptances, never cause a rejection, so it can't bring back #6810, and since it holds no system anchors it can't override the platform's rejection of a system-rooted certificate.

Note: for standard/Google-certified Android, this regression could be fixed by simply reverting #6753 (back to `init(null)`) - the default path already honors user-installed CAs there. The composite fallback exists only to also cover the non-compliant ROMs that #6753 was originally trying to help (#5565). If supporting those ROMs isn't wanted, the plain revert is enough.

Tested on a physical device (HyperOS, Android 16) against a server with a private CA: the default path failed with `Trust anchor for certification path not found`, and the user-CA fallback completed the handshake.

Fixes #6810
Re #5565

photo_2026-06-15_11-44-17

## Checklist

- [x] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction). `CompositeX509ExtendedTrustManagerTest` covers the fallback decision logic (12 tests; JUnit 5 + MockK, GIVEN-WHEN-THEN, parameterized over the three `checkServerTrusted` overloads), and `TLSHelperTest` covers the user-CA filtering of the `AndroidCAStore` (4 tests). Reading the live `AndroidCAStore` itself is verified on-device, since that store type only exists on Android.
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices). ktlint passes; only specific exceptions are caught (`CertificateException`, `GeneralSecurityException`, `IOException`, never `Throwable` or `CancellationException`); named file-level constants; small functions; Timber logging with no user data.
- [x] The changes have been thoroughly tested, and edge cases have been considered. Covered: pre-Android-N, non-extended trust manager, no user CAs / unreadable store, both managers rejecting (the platform error is kept, the fallback's attached as suppressed), non-`CertificateException` propagation, and the mTLS client-certificate path left untouched. The tests were also mutation-checked (breaking the fallback fails exactly the fallback tests, reverting passes all 12).
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance. No public API change (the new class is `internal`); it adds trust and removes nothing. The only behavior tightened versus #6753 is that it no longer re-trusts a system CA the platform distrusts, which is a security improvement rather than a user-facing break. The trust trade-off is documented in code, and a user changelog entry is included.

## Any other notes

The fallback is intentionally limited to user-installed CAs (the `user:`-aliased entries of `AndroidCAStore`). Seeding it from the whole store would also let it re-accept a certificate chaining to a system root the platform deliberately rejected, so it's restricted to what the user explicitly added.


common/src/test/kotlin/io/homeassistant/companion/android/common/data/TLSHelperTest.kt
```diff
@@ -0,0 +1,75 @@
...
+    @Test
+    fun `Given an empty store when filtering then it returns null`() {
+        assertNull(userInstalledCaKeyStore(keyStoreOf()))
+    }
+
+    private fun keyStoreOf(vararg entries: Pair<String, X509Certificate>): KeyStore {
+        return KeyStore.getInstance(KeyStore.getDefaultType()).apply {
+            load(null)
+            entries.forEach { (alias, certificate) -> setCertificateEntry(alias, certificate) }
+        }
```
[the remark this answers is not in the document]
bpacholek: Good point on the clarity. `load(null)` does bind to the `load(LoadStoreParameter?)` overload, but a null parameter is handled by the default `KeyStoreSpi`, which delegates to `load((InputStream) null, (char[]) null)`, so it doesn't throw (the tests pass on JDK 17's PKCS12 default). The explicit two-arg form is unambiguous though, so I've switched to `load(null, null)`.

Ready to review again.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -1,45 +1,96 @@
...
+    }
+
+    /**
+     * Builds a trust manager from the user-installed CAs (see [userInstalledCaKeyStore]), or `null`
+     * when there are none or the AndroidCAStore can't be read.
+     */
+    @RequiresApi(Build.VERSION_CODES.N)
+    private fun userInstalledCaTrustManager(): X509ExtendedTrustManager? = try {
+        val androidCaStore = KeyStore.getInstance(ANDROID_CA_STORE).apply { load(null) }
+        userInstalledCaKeyStore(androidCaStore)?.let { defaultX509TrustManager(it) as? X509ExtendedTrustManager }
```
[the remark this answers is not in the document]
bpacholek: Done here too, for consistency: `load(null, null)`. For `AndroidCAStore` specifically, `load(null)` is the documented Android idiom and works on-device, but the two-arg form is equivalent and reads more clearly.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -70,3 +121,20 @@ class TLSHelper @Inject constructor(
...
+ */
+internal fun userInstalledCaKeyStore(androidCaStore: KeyStore): KeyStore? {
+    val userCaAliases = Collections.list(androidCaStore.aliases())
+        .filter { it.startsWith(USER_CA_ALIAS_PREFIX) }
+    if (userCaAliases.isEmpty()) return null
+    return KeyStore.getInstance(KeyStore.getDefaultType()).apply {
+        load(null)
+        userCaAliases.forEach { alias -> setCertificateEntry(alias, androidCaStore.getCertificate(alias)) }
+    }
+}
```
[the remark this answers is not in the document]
bpacholek: This runs once, when the app-wide singleton `OkHttpClient` is built, so it isn't a hot path. It turns the device's CA aliases (the system set is on the order of ~150 short strings) into a list and filters it: a few KB and microseconds, one time. Iterating the `Enumeration` directly would mean a manual `while (hasMoreElements())` loop, since Kotlin's stdlib has no `Enumeration` sequence/filter helpers, which reads worse for no measurable gain. I'd prefer to keep the `Collections.list(...).filter { ... }` form, but happy to revisit if there's evidence of real startup cost.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -1,45 +1,96 @@
...
+        // network_security_config.xml (system + user CAs) and builds the certificate chain itself.
+        // Passing an explicit KeyStore here instead breaks validation of valid certificates (#6810).
+        val platformTrustManager = defaultX509TrustManager(keyStore = null)
+
+        // Some ROMs (e.g. /e/OS, #5565) don't trust user-installed CAs through the default path even
+        // though the browser and WebView do. On those we fall back to a trust manager holding only
+        // the user-installed CAs, consulted only when the default rejects a certificate. It can add
+        // acceptances but never cause a rejection, so it can't bring back #6810. See
+        // [CompositeX509ExtendedTrustManager] for the trust trade-off.
+        val handshakeTrustManager = withUserInstalledCaFallback(platformTrustManager)
```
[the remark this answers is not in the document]
bpacholek: Fair principle in general, but I don't think it applies here. This app's `network_security_config.xml` explicitly opts into user-installed CAs (`<certificates src="user"/>`, since #194), so trusting them is the app's intended policy. The fallback enforces that policy on ROMs that fail to honor it (e.g. /e/OS, #5565), rather than bypassing a policy-driven rejection. There's no configuration here where a user CA is intentionally not trusted.

The fallback is also already restricted to user-installed CAs only (the `user:` aliases of `AndroidCAStore`), so it can't override the platform's distrust of a system CA.

A build-flavor flag or config toggle would add a gate for a scenario this app doesn't have (a config that excludes user CAs), which felt like over-engineering for a hypothetical. The one real forward-looking point, that the fallback hardcodes user-CA trust rather than reading the config, I'm happy to document in a code comment noting it assumes `network_security_config` trusts user CAs. Let me know if you'd prefer the explicit gate instead.

TimoPtr:
@bpacholek thanks for helping out on this one, I need to replicate the issue in order to even check the PR. Would you mind sharing a bit more about how to properly setup the environment?

I need to understand why it started to pop-up on some devices, only recently.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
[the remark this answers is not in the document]
bpacholek: Actually will implement it to close this easier.

app/src/main/res/xml/changelog_master.xml
```diff
@@ -2,6 +2,7 @@
 <changelog xmlns:tools="http://schemas.android.com/tools"
     tools:ignore="MissingDefaultResource">
     <release version="2026.6.3 - Main" versioncode="3">
+        <change>Fixed connecting to servers that use a custom or user-installed certificate authority</change>
```
TimoPtr:
```suggestion
```
We don't add specific bug fix in the user facing changelog.

bpacholek: Okay, was not sure here, was quite empty, but was not sure if that is not a new policy. Theoretically it said only for bc-breaks, but just wanted to be verbose here about the changes. Will revert.

bpacholek:
@TimoPtr apologies for delay:

Here's how my server's cert behaves across versions - same cert, same device, no user CAs installed:

| App version          | Trust setup                                       | Result for My domain |
| -------------------- | ------------------------------------------------- | ------------------------------------------------ |
| 2026.4.5             | TrustManagerFactory.init(null)                    | connects                                         |
| 2026.5.0 - 2026.6.3  | trust manager seeded from AndroidCAStore (#6753)  | SSL handshake failed                             |
| this PR              | init(null) (+ user-CA fallback, unused here)      | connects                                         |


**Starting from the summary so that you know where it reaches:**
* Yes: the clean server-side fix is for providers to send the full chain, but since we have an explosion of people reporting the problem it is more common that it may sound. 
* But also yes: On the app side, this PR just stops it being stricter than it was before #6753.

Incomplete chains are common enough that SSL Labs calls it "a common configuration problem" and explicitly notes that "some browsers can reconstruct incomplete chains, and some can't" (https://success.qualys.com/support/s/article/000003198, https://www.ssllabs.com/projects/best-practices/). That "some can't" is exactly Android's position: it doesn't AIA-fetch, so whether it reconstructs the chain depends entirely on what it already has locally. It's not a cheap-cert thing: I first hit it with a Let's Encrypt cert on this device, and the current commercial QNAP cert has the same gap - it sends the leaf and a TWCA root but omits the "TWCA SSL Certification Authority" intermediate that signed the leaf (openssl: "unable to verify the first certificate"; adding the AIA-published intermediate makes it verify).

One could argue here that it is just a problem of the server, yet in fact #6753 was a regression:
Android does not AIA-fetch (Conscrypt has no caIssuers code; cf. OkHttp #7052), so it can only complete a gapped chain from a local source: the live system trust store, or Conscrypt's per-process intermediate cache (TrustManagerImpl.intermediateIndex). init(null) gives a trust manager wired to both. Seeding the trust manager from AndroidCAStore instead gives a static snapshot with neither, so the same cert that worked on 2026.4.5 fails on 2026.5.0+. This PR reverts that to init(null).

(The PR also adds a fallback that, only when the default path rejects a cert, consults a trust manager holding the user-installed CAs - for ROMs that don't honor user CAs through the default path. That's a separate case and isn't involved here, since I have no user CA installed.)

So essentially this is a fix for the regression created in #6753 while keeping the support for user-CAs that it has added.


A bit more specific why this regression happened technically:

Why init(null) completes this chain and the AndroidCAStore-seeded one does not - in detail.

1) Android does not fetch missing intermediates (no AIA). When a server omits an intermediate, browsers and iOS fetch it from the leaf's AIA "CA Issuers" URL. Android never does. The trust manager Android uses, and the one OkHttp wraps via X509TrustManagerExtensions, is Conscrypt's TrustManagerImpl, and its chain building does no network I/O: there is no caIssuers / AuthorityInfoAccess / AIA fetch anywhere in it, and it runs with revocation checking disabled. This is a known gap, not something specific to this app - see square/okhttp#7052 ("Support download of intermediate certificates aka AIA Fetching"), still open. So openssl failing on the cert with only the server-sent certs is the same position Android is in: nothing gets fetched, so the gap has to be filled from local data or not at all.

2) How Conscrypt fills the gap locally. While building the path, TrustManagerImpl looks for each cert's issuer in three local places only:
   - the certificates the server actually sent;
   - the trust manager's certificate store, via findAllIssuers() / getTrustAnchor() - but only if that manager was constructed with a live store; and
   - a per-instance intermediate cache, intermediateIndex (a TrustedCertificateIndex), which it populates after each successful validation, explicitly "to tolerate sites that assume the browser will have cached these" (AOSP b/3404902: https://github.com/google/conscrypt/blob/master/common/src/main/java/org/conscrypt/TrustManagerImpl.java#L693).
   If none of those reach a trust anchor, it throws exactly the error the released app shows: "Trust anchor for certification path not found."

3) The two trust managers are structurally different. On Android, TrustManagerFactory.getInstance(...) does not return raw Conscrypt; the network-security-config provider supplies android.security.net.config.RootTrustManagerFactorySpi. Its engineInit(KeyStore) branches:
   - init(null)            -> ApplicationConfig.getDefaultInstance()              -> a TrustManagerImpl backed by the LIVE system certificate source, plus its persistent intermediateIndex. It has both local sources from (2).
   - init(AndroidCAStore)  -> new ApplicationConfig(new KeyStoreConfigSource(ks)) -> a STATIC snapshot of trust anchors (roots) only: no live store, and an intermediateIndex that can never get seeded because it has no live store to bootstrap a first successful validation.
   Note the non-obvious part: even though the keystore is literally "AndroidCAStore", the framework SPI freezes it into that static config first, so it never reaches Conscrypt's live-store behavior. That's why passing AndroidCAStore is not equivalent to the default.

4) Result for this cert. The missing piece is the "TWCA SSL Certification Authority" intermediate - not a root, and there is no user CA. The init(null) manager can still obtain it (live-store issuer lookup, or its intermediate cache), so the path reaches the TWCA root and the handshake succeeds. The AndroidCAStore-seeded manager has neither: a roots-only static snapshot cannot supply an intermediate, so it throws "Trust anchor for certification path not found." That is the whole regression, and reverting to init(null) restores it.

5) Consistency check. This matches every observation: openssl (no AIA, no intermediate store, no cache) fails exactly like the AndroidCAStore manager; browsers/iOS (AIA) succeed; and on device the first handshake or two failed and a later one succeeded, which is the local cache/store warming up, not a network fetch - an AIA fetch would have succeeded on the very first handshake. (The only detail not provable from source alone is which local source supplies the intermediate the first time on a given device, the live store or the cache; both are local, both exist only on the init(null) manager, so the conclusion is identical either way.)

common/src/main/kotlin/io/homeassistant/companion/android/common/data/CompositeX509ExtendedTrustManager.kt
```diff
@@ -0,0 +1,92 @@
...
+import java.net.Socket
+import java.security.cert.CertificateException
+import java.security.cert.X509Certificate
+import javax.net.ssl.SSLEngine
+import javax.net.ssl.X509ExtendedTrustManager
+
+/**
+ * An [X509ExtendedTrustManager] that trusts a server certificate if either the [primary] or the
+ * [fallback] accepts it. The [primary] is always tried first; the [fallback] is only asked when the
+ * [primary] rejects the chain.
```
TimoPtr, line 14:
```suggestion
 * [fallback] accepts it. The [fallback] is only asked when the [primary] rejects the chain with a [CertificateException].
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/CompositeX509ExtendedTrustManager.kt
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * The [primary] is Android's default trust manager (`TrustManagerFactory.init(null)`), which honors
```
TimoPtr, line 16: You're making an assumption that the caller is going to use the default trust manager here. Let's not make this asumption. This Composite class is a small helper that is not specific to the issue you solve.

The doc here should only explain the behavior of the class not how it is use elsewhere.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/CompositeX509ExtendedTrustManager.kt
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * This is deliberately more permissive than the [primary]: it re-trusts user-installed CAs the
+ * platform refused, which widens the trust surface to whatever CAs the user (or an attacker who
+ * tricks them) has installed. It can't do worse than that: the [fallback] has no system anchors, so
+ * it can't override a rejected system-rooted certificate, and it can't accept a certificate that
+ * chains to no trusted anchor (e.g. self-signed). And since the [fallback] only adds acceptances, it
+ * can't bring back the over-strict rejection of #6810.
+ *
+ * Client-trust checks and [getAcceptedIssuers] use the [primary] only; the app is a TLS client, so
+ * the [fallback] has no role there.
```
TimoPtr, line 29: You can be a bit more specific about why it has no role there.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -1,45 +1,100 @@
...
 import javax.net.ssl.TrustManagerFactory
 import javax.net.ssl.X509ExtendedKeyManager
+import javax.net.ssl.X509ExtendedTrustManager
 import javax.net.ssl.X509TrustManager
 import okhttp3.OkHttpClient
 import timber.log.Timber
 
+private const val ANDROID_CA_STORE = "AndroidCAStore"
+
+// AndroidCAStore prefixes user-installed CA aliases with "user:" and system ones with "system:".
```
TimoPtr, line 27: I don't think we need to mention system: here

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -1,45 +1,100 @@
...
-        } catch (e: Throwable) {
-            Timber.w(e, "AndroidCAStore unavailable, falling back to system trust store")
-            null
-        }
-        trustManagerFactory.init(androidCaStore)
-        val trustManagers = trustManagerFactory.trustManagers
+        // Default trust manager. init(null) gives us Android's default, which honors
+        // network_security_config.xml (system + user CAs) and builds the certificate chain itself.
+        // Passing an explicit KeyStore here instead breaks validation of valid certificates (#6810).
+        val platformTrustManager = defaultX509TrustManager(keyStore = null)
```
TimoPtr, line 39: Make this a real github link not just the ID.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -1,45 +1,100 @@
...
+
+        // Some ROMs (e.g. /e/OS, #5565) don't trust user-installed CAs through the default path even
+        // though the browser and WebView do. On those we fall back to a trust manager holding only
+        // the user-installed CAs, consulted only when the default rejects a certificate. It can add
+        // acceptances but never cause a rejection, so it can't bring back #6810. See
+        // [CompositeX509ExtendedTrustManager] for the trust trade-off.
+        //
+        // This relies on the app intending to trust user-installed CAs, which network_security_config.xml
+        // opts into with <certificates src="user"/>. If that policy is ever removed, this fallback would
+        // keep trusting user CAs against the new policy, so it must be removed or gated alongside it.
```
TimoPtr, line 49: I feel that all this doc could be simplified while still explaining why we do this into one block at the function level.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -1,45 +1,100 @@
...
 
         val sslContext = SSLContext.getInstance("TLS")
-        sslContext.init(arrayOf(getMTLSKeyManagerForOKHTTP()), trustManagers, null)
+        sslContext.init(arrayOf(getMTLSKeyManagerForOKHTTP()), arrayOf(handshakeTrustManager), null)
 
-        builder.sslSocketFactory(sslContext.socketFactory, trustManagers[0] as X509TrustManager)
+        // OkHttp only uses this trust manager to build a chain cleaner for certificate pinning, which
+        // the app doesn't use, so it doesn't affect the handshake (handshakeTrustManager decides
+        // that). We still pass the default one so OkHttp keeps using Android's chain cleaner if
+        // pinning is ever added.
```
TimoPtr, line 58: Not sure this comment brings anything here, and would get outdated if we introduce cert pinning.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -1,45 +1,100 @@
...
+    /**
+     * Wraps [platformTrustManager] so user-installed CAs are honored as a fallback (see #5565).
+     *
+     * Returns it unchanged, with no fallback, when:
+     * - running before Android N, where user-installed CAs are trusted by default;
+     * - the platform trust manager is not an [X509ExtendedTrustManager] (the composite needs the
+     *   hostname-aware overloads); or
+     * - there are no user-installed CAs to fall back to.
+     */
+    private fun withUserInstalledCaFallback(platformTrustManager: X509TrustManager): X509TrustManager {
```
TimoPtr, line 71:
```suggestion
    private fun withUserInstalledCaFallback(trustManager: X509TrustManager): X509TrustManager {
```
I wouldn't assume it's the platfomTrustManager, we never know how this is going to be use.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -1,45 +1,100 @@
...
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) return platformTrustManager
```
TimoPtr, line 72: This should use `SdkVersion.isAtLeast(Build.VERSION_CODES.N)`.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -70,3 +125,20 @@ class TLSHelper @Inject constructor(
...
 }
+
+/**
+ * Returns a key store holding only the user-installed CA certificates of [androidCaStore] (the
+ * entries aliased with [USER_CA_ALIAS_PREFIX]), or `null` if there are none.
+ *
+ * The preinstalled system CAs are left out so the fallback can only trust CAs the user added, never
+ * override the platform's rejection of a system-rooted certificate (e.g. one from a distrusted CA).
+ */
+internal fun userInstalledCaKeyStore(androidCaStore: KeyStore): KeyStore? {
```
TimoPtr, line 136: Same here you assume that the keystore is the androidCaStore which might not be.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
```diff
@@ -70,3 +125,20 @@ class TLSHelper @Inject constructor(
...
+    val userCaAliases = Collections.list(androidCaStore.aliases())
+        .filter { it.startsWith(USER_CA_ALIAS_PREFIX) }
+    if (userCaAliases.isEmpty()) return null
+    return KeyStore.getInstance(KeyStore.getDefaultType()).apply {
```
TimoPtr, line 140: Why getting another instance of a KeyStore?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/TLSHelper.kt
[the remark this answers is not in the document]
TimoPtr: FYI our start time is not great already today, if we can avoid increasing it would be best.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/TLSHelperTest.kt
```diff
@@ -0,0 +1,75 @@
...
+    private fun keyStoreOf(vararg entries: Pair<String, X509Certificate>): KeyStore {
+        return KeyStore.getInstance(KeyStore.getDefaultType()).apply {
+            load(null, null)
+            entries.forEach { (alias, certificate) -> setCertificateEntry(alias, certificate) }
+        }
+    }
+
+    companion object {
+        private val USER_CA = loadCertificate("/tls/user-ca.b64")
+        private val SYSTEM_CA = loadCertificate("/tls/system-ca.b64")
```
TimoPtr, line 65: Why storing the cert in base64? Let's keep the cert in const val string in the test file.

TimoPtr (changes requested):
Thanks for the in-depth explanation. Let's also have @jpelgrom look at it. I do have some trouble reproducing the issue. But the PR seems logical from a high level perspective (still some issues to address).

jpelgrom:
While I can understand the suggested fix, I'm actually unable to consider it fixed with the url you added as an example (which shouldn't require additional certificates to be installed and work on 2026.4.5, if I understand correctly?). On 2026.4.5, OkHttp fails to connect to for me with the same `javax.net.ssl.SSLHandshakeException: java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.`. We (maintainers) need a good example to reproduce the issue and verify any fixes.

The comments in the code are very verbose and clearly AI generated. Please try to keep the information to the minimum required. Describe the current state and why we need this, instead of referring to previous implementations and issues. @TimoPtr added some good comments already.

PR #7031, opened 2026-06-16
Introduce ProvidesSensor annotation + KSP + Lint

TimoPtr:
## Summary

Introduce a new KSP module and a lint rule to force us to annotate all the `BasicSensor` instances. By doing so we ensure that we can build a list of all basic sensor and inject it in DI. It's the base to be able to implement the `SensorRepository` that will wrap the DB and the codebase to always retrieve all the sensors without having to pre-populate the DB.

It is quite verbose, but I made it this way so we don't forget to annotate a new sensor.

## Checklist

## Any other notes

I initially created a hilt module but I found out that the KSP processor of hilt is not using the second run to check if there is a new module generated because of that you cannot generate hilt module with KSP because we cannot priority over the KSP run, so we cannot guarantee that the SensorModule would have been created. Instead I've decided to build a static list that we then provide into the graph.


app/src/test/kotlin/io/homeassistant/companion/android/sensors/BasicSensorCatalogSetTest.kt
```diff
@@ -0,0 +1,36 @@
...
+@HiltAndroidTest
+class BasicSensorCatalogSetTest {
+    @get:Rule
+    val hilt = HiltAndroidRule(this)
+
+    @Inject
+    lateinit var sensors: Set<@JvmSuppressWildcards SensorManager.BasicSensor>
+
+    @Test
+    fun `Given hilt graph then catalog set is non-empty with unique ids and includes catalog sensors`() {
```
jpelgrom: Do we want a full/minimal flavor specific test, to verify that sensors in their modules are included correctly?

app/build.gradle.kts
```diff
@@ -51,6 +51,11 @@ firebaseAppDistributionDefault {
 dependencies {
     // Most of the dependencies are coming from the convention plugin to avoid duplication with `:automotive` module.
     "fullImplementation"(libs.car.projected)
+    ksp(project(":sensor-catalog-processor"))
+}
+
+ksp {
+    arg("catalogModuleSuffix", project.name)
```
jpelgrom: Any particular reason not to add this to the `build-logic` like other dependencies/config?

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/CatalogSensor.kt
```diff
@@ -0,0 +1,13 @@
+package io.homeassistant.companion.android.common.sensors
+
+/**
+ * Marks a [SensorManager.BasicSensor] val for inclusion in the generated `Set<BasicSensor>` catalog
+ * (see the sensor-catalog KSP processor). Every `BasicSensor` val must be annotated; an
+ * error-severity lint rule (`CatalogSensorMissing`) flags omissions. To intentionally exclude a
+ * `BasicSensor` from the catalog, leave it un-annotated and `@Suppress("CatalogSensorMissing")`.
+ *
+ * The annotated val must be statically referenceable (top-level, `object`, or `companion object`).
```
jpelgrom: \+ not private, considering your changes

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/CommonSensorCatalogModule.kt
```diff
@@ -0,0 +1,24 @@
...
+import dagger.multibindings.ElementsIntoSet
+import io.homeassistant.companion.android.sensorcatalog.generated.GeneratedSensorCatalogCommon
+
+/**
+ * Contributes the `:common` module's `@CatalogSensor` sensors into the `Set<BasicSensor>` catalog.
+ *
+ * This Hilt module is hand-written (not generated) on purpose: Hilt's KSP processor does not
+ * reliably aggregate an `@InstallIn` module that is itself *generated* by another KSP processor (the
+ * KSP2 round model doesn't feed it to Hilt). So the sensor-catalog processor generates only a plain
+ * data object ([GeneratedSensorCatalogCommon]), and this source module turns it into the binding.
```
jpelgrom: This description is different from the one in `app` and `wear`, I 'd re-use that one here for consistency.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/SensorRepository.kt
```diff
@@ -0,0 +1,14 @@
+package io.homeassistant.companion.android.common.sensors
+
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Single access point for sensor state and the sensor catalog.
+ *
+ * Stub: for now it only receives the generated [SensorManager.BasicSensor] catalog (built by the
```
jpelgrom:
```suggestion
 * TODO: For now it only receives the generated [SensorManager.BasicSensor] catalog (built by the
```

gradle/libs.versions.toml
```diff
@@ -53,6 +53,7 @@ improv = "0.1.1"
 javaVersion = "11" # https://developer.android.com/studio/write/java11-minimal-support-table
 junit = "6.1.0"
 kotlin = "2.4.0"
+kotlinpoet = "1.11.0" # https://github.com/square/kotlinpoet/pull/2320 we need this otherwise
```
jpelgrom: otherwise ...? If we need it why is the referenced version this specific one?

lint/src/test/kotlin/io/homeassistant/lint/annotation/CatalogSensorMissingDetectorTest.kt
```diff
@@ -0,0 +1,72 @@
...
+                        @CatalogSensor val alpha = SensorManager.BasicSensor("alpha")
+                    }
+                }
+                """,
+            ).indented(),
+        ).issues(CatalogSensorMissingDetector.ISSUE).run().expectClean()
+    }
+
+    @Test
+    fun `Given un-annotated val with Suppress then clean`() {
```
jpelgrom:
```suggestion
    fun `Given un-annotated val when suppressed then clean`() {
```

sensor-catalog-processor/src/main/kotlin/io/homeassistant/companion/android/sensorcatalog/SensorCatalogProcessor.kt
```diff
@@ -0,0 +1,130 @@
...
+import com.squareup.kotlinpoet.MemberName
+import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
+import com.squareup.kotlinpoet.PropertySpec
+import com.squareup.kotlinpoet.TypeSpec
+import com.squareup.kotlinpoet.ksp.KotlinPoetKspPreview
+import com.squareup.kotlinpoet.ksp.toClassName
+import com.squareup.kotlinpoet.ksp.writeTo
+
+private const val CATALOG_SENSOR_FQN = "io.homeassistant.companion.android.common.sensors.CatalogSensor"
+private const val GENERATED_PACKAGE = "io.homeassistant.companion.android.sensorcatalog.generated"
```
jpelgrom: Not sure about this but if you want a different package from `sensors` and it's just for managing them:

```suggestion
private const val GENERATED_PACKAGE = "io.homeassistant.companion.android.settings.sensorcatalog.generated"
```

But the new repository and modules are in `sensors`.

sensor-catalog-processor/src/main/resources/META-INF/services/com.google.devtools.ksp.processing.SymbolProcessorProvider
```diff
@@ -0,0 +1 @@
+io.homeassistant.companion.android.sensorcatalog.SensorCatalogProcessorProvider
```
jpelgrom:
```suggestion
io.homeassistant.companion.android.settings.sensorcatalog.SensorCatalogProcessorProvider
```

sensor-catalog-processor/src/test/kotlin/io/homeassistant/companion/android/sensorcatalog/SensorCatalogProcessorTest.kt
```diff
@@ -0,0 +1,281 @@
...
+        val result = workingDir.runProcessor()
+
+        assertEquals(KotlinSymbolProcessing.ExitCode.OK, result.exitCode, result.messages.joinToString("\n"))
+        val generated = workingDir.generatedFile("GeneratedSensorCatalogTest.kt").readText()
+        // Golden source documenting the generated module. Object and companion-object vals are
+        // referenced qualified by their declaring type (a member import of a companion-object member
+        // does not compile), so a companion resolves to `BarManager.Companion.beta`; only the
+        // top-level gamma is imported directly.
+        val expected = """
+            package io.homeassistant.`companion`.android.sensorcatalog.generated
```
jpelgrom:
```suggestion
            package io.homeassistant.`companion`.android.settings.sensorcatalog.generated
```

jpelgrom (commented):
Took me a while to get started on this because I wanted to understand custom annotation processors, but after some research it does look OK. It works which is what matters most :D

A question on naming, as you'll likely expect from me by now: you're marking this as a sensor for a "catalog". Is this something you intend to surface later? Core does not use "catalog" anywhere so I do think we shouldn't create new terms if not necessary.

TimoPtr:
> Took me a while to get started on this because I wanted to understand custom annotation processors, but after some research it does look OK. It works which is what matters most :D
> 
> A question on naming, as you'll likely expect from me by now: you're marking this as a sensor for a "catalog". Is this something you intend to surface later? Core does not use "catalog" anywhere so I do think we shouldn't create new terms if not necessary.

Do you have some proposal? I wanted to avoid `Sensor` to not mess with the existing Sensor data class.

app/src/test/kotlin/io/homeassistant/companion/android/sensors/BasicSensorCatalogSetTest.kt
TimoPtr: I wanted to, but our current setup is focused on full and if you try to run minimal tests it is failing. We should invest time to make it work if we want to add a test like this one for minimal

app/build.gradle.kts
TimoPtr: Because of `automotive` I dropped a comment about it in the `AppSensorCatalogModule` but maybe it should be here too.

TimoPtr: It's already in automotive. I guess we could move it to the convention plugin and only override in automotive.

sensor-catalog-processor/src/main/kotlin/io/homeassistant/companion/android/sensorcatalog/SensorCatalogProcessor.kt
TimoPtr: I don't see this only for settings, would you prefer to have this in `sensors`?

sensor-catalog-processor/src/main/resources/META-INF/services/com.google.devtools.ksp.processing.SymbolProcessorProvider
TimoPtr: It's the KSP class I'm not sure I want this in settings too.

sensor-catalog-processor/src/main/kotlin/io/homeassistant/companion/android/sensorcatalog/SensorCatalogProcessor.kt
jpelgrom: In that case, yes I think so to avoid introducing another package which only has one class. Every usage is already in the `sensors` package.

jpelgrom:
> > A question on naming, as you'll likely expect from me by now: you're marking this as a sensor for a "catalog". Is this something you intend to surface later? Core does not use "catalog" anywhere so I do think we shouldn't create new terms if not necessary.
> 
> Do you have some proposal? I wanted to avoid `Sensor` to not mess with the existing Sensor data class.

After thinking about it for a while: `@SensorDefinition`? Not sold on it though.

My question was more to indicate that we should avoid ever using "catalog" in a user-facing place, to avoid confusion. The differences between our `BasicSensor` definition, `Sensor` database value and `SensorWithAttributes` which combines data isn't exposed, and this new thing aren't a user concern - it's all part of the sensor.

TimoPtr:
@jpelgrom I've removed any mention to Catalog from the PR.

jpelgrom:
For future reference, `@ProvidesSensor` was agreed upon after [discussion on Discord](https://discord.com/channels/330944238910963714/1346948551892009101/1520882562728460349)

jpelgrom (approved):
Let's go with this, seems solid enough

PR #7041, opened 2026-06-17
Make Sensor immutable

TimoPtr:
## Summary

The `Sensor` data class should be immutable since it comes directly from the repository, if we want to modify it we should make a copy of the object. This PR replace the `var` with `val` in the class.

It is based on #7107

### DIFF
--- a/common/lint-baseline.xml
+++ b/common/lint-baseline.xml
@@ -118,7 +118,7 @@
         errorLine2="        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
         <location
             file="src/main/kotlin/io/homeassistant/companion/android/common/sensors/SensorUpdater.kt"
-            line="370"
+            line="421"
             column="9"/>
     </issue>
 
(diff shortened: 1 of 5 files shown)


jpelgrom (approved):

jpelgrom:
I think the rebase is wrong here, it also has commits from the other PRs.

jpelgrom (approved):
Edge case with sensor updating during sync also looks good, interesting that we didn't encounter it before.

jpelgrom:
Found one more issue, I think it's because of the change here but am not sure as I'm testing with the latest in stack to have sensors work. Changing a setting doesn't update Compose immediately, I need to scroll it out of view to have it update:

https://github.com/user-attachments/assets/a1e555a5-1056-4b43-95b4-b6511f80e500

TimoPtr:
> Found one more issue, I think it's because of the change here but am not sure as I'm testing with the latest in stack to have sensors work. Changing a setting doesn't update Compose immediately, I need to scroll it out of view to have it update:
> settings.mp4

Issue found, I'm going to open a follow-up PR that fixes the issue and a lint rule forbid the usage of `var` in Entity objects.

PR #7070, opened 2026-06-22
Remove runBlocking from MessagingManager

TimoPtr:
## Summary

The `runblocking` is not needed in the `MessagingManager` and even block the current coroutine scope and we shouldn't.


jpelgrom:
Nice catch, this is now no longer needed indeed. Related: #5688.

jpelgrom (approved):

PR #7079, opened 2026-06-23
Apply keep_screen_on change immediately while a dashboard is showing

### LINKED ISSUE #6829: Screen On Notification doesn't apply "keep_screen_on" immediately if app open
2026.4.4
**Android version(s):**
16
**Device model(s):**
Doogee U12
**Home Assistant version:**
2026.5.1

With the "keep_screen_on" attribute, the notification works without any issues; however, without this attribute, the screen timeout no longer activates. This can only be resolved by either opening the app settings to verify that the "Keep Screen On" feature has been disabled, or by triggering a reload of the dashboard using the "Webview" notification, for example.

**Companion App Logs:**

```

```

**Additional information:**

mvanhorn:
## Summary

Fixes #6829.

When a `command_screen_on` notification carries a `command` payload, `MessagingManager` persists the "Keep screen on" preference via `prefsRepository.setKeepScreenOnEnabled(...)`. As diagnosed in the issue, if a `WebViewActivity` dashboard is already in the foreground the new value is never applied to the live window: turning the setting **on** appears to work because acquiring the wake lock lights the screen anyway, but turning it **off** does not clear `FLAG_KEEP_SCREEN_ON` until the activity is recreated. The result is the asymmetry the reporter confirmed: enabling is immediate, disabling is not.

This wires the persisted change through to a live dashboard:

* `MessagingManager` sends an app-internal, package-scoped broadcast (`ACTION_KEEP_SCREEN_ON_CHANGED`) right after it persists the preference in the `COMMAND_SCREEN_ON` branch (only when a command string is present, so the no-command "just wake the screen" path is unchanged).
* `WebViewActivity` registers a non-exported `BroadcastReceiver` (via `ContextCompat.registerReceiver(..., RECEIVER_NOT_EXPORTED)`, matching the existing receiver convention) while it is in the foreground and re-applies/clears `FLAG_KEEP_SCREEN_ON` from the fresh preference value. The flag apply/clear logic is consolidated into a single `applyKeepScreenOn()` helper reused by `onCreate`, `onResume`, and the receiver. The receiver is registered in `onResume` and unregistered in `onPause` (guarded so an unmatched unregister never throws), so nothing leaks while the activity is backgrounded.

Impact: disabling "Keep screen on" via a notification command now takes effect immediately on an open dashboard, with no activity recreation and no behavior change to the wake-lock-only path.

## Checklist

## Screenshots

There is no visible UI change to screenshot: this fix only re-applies the existing `FLAG_KEEP_SCREEN_ON` window flag to a live dashboard, so its effect is whether the screen stays on, not a rendered surface.

## Any other notes

The behavior was verified by reasoning through the issue's four scenarios (open dashboard + enable, open dashboard + disable, no-command wake-only, no foreground activity). I was unable to compile or run the Android test suite locally to add a new automated test, as there is no existing focused JVM/Robolectric harness around `MessagingManager`/`WebViewActivity` to extend without new test infrastructure; the change reuses the same `presenter.isKeepScreenOnEnabled()` flag-apply path that already runs on activity recreation.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
@@ -184,6 +184,7 @@ class MessagingManager @Inject constructor(
         const val COMMAND_ACTIVITY = "command_activity"
         const val COMMAND_WEBVIEW = "command_webview"
         const val COMMAND_KEEP_SCREEN_ON = "keep_screen_on"
+        const val ACTION_KEEP_SCREEN_ON_CHANGED = "io.homeassistant.companion.android.KEEP_SCREEN_ON_CHANGED"
         const val COMMAND_LAUNCH_APP = "command_launch_app"
         const val COMMAND_APP_LOCK = "command_app_lock"
         const val COMMAND_PERSISTENT_CONNECTION = "command_persistent_connection"
@@ -815,6 +816,9 @@ class MessagingManager @Inject constructor(
                 if (!command.isNullOrEmpty()) {
                     prefsRepository.setKeepScreenOnEnabled(
                         command == COMMAND_KEEP_SCREEN_ON,
+                    )
+                    context.sendBroadcast(
+                        Intent(ACTION_KEEP_SCREEN_ON_CHANGED).setPackage(context.packageName),
                     )
                 }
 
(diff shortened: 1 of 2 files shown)


jpelgrom:
The WebViewActivity has been replaced, and the new Compose-based screen already observes the value to respond immediately.

PR #7088, opened 2026-06-24
Force Kotlin version for AbiValidationCompatClasspath

TimoPtr:
## Summary

We started to see failure on CI because lockfile were not updated and the `lockfile` job detected it. It seems something in Kotlin deps changed and we got 
image

This is not something we want we want a fixed version of Kotlin. This PR with the help of codex is addressing this by ensuring we use the right version of Kotlin for the task kotlinAbiValidationCompatClasspath.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,7 @@
 import org.jlleitschuh.gradle.ktlint.reporter.ReporterType
 
+val kotlinVersion = libs.versions.kotlin.get()
+
 plugins {
     alias(libs.plugins.ktlint)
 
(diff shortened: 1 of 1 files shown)


jpelgrom:
Trying to find out where it's coming from with `./gradlew :common:dependencyInsight --dependency org.jetbrains.kotlin:kotlin-build-tools --configuration kotlinAbiValidationCompatClasspath`, I get

```
org.jetbrains.kotlin:kotlin-build-tools-api:2.4.0
+--- org.jetbrains.kotlin:kotlin-build-tools-cri-impl:2.4.0
|    +--- kotlinAbiValidationCompatClasspath (requested org.jetbrains.kotlin:kotlin-build-tools-cri-impl:{strictly 2.4.0})
|    \--- org.jetbrains.kotlin:kotlin-build-tools-impl:2.4.0
|         \--- kotlinAbiValidationCompatClasspath (requested org.jetbrains.kotlin:kotlin-build-tools-impl:{strictly [2.4.0-Beta2, 2.5.0)})
```

I don't really like this change as it might work, but this is an issue that should be prevented with lockfiles (?) and fixed upstream as it doesn't make sense there is a beta version here. Maintaining this extra code shouldn't be on us.

TimoPtr:
> Trying to find out where it's coming from with `./gradlew :common:dependencyInsight --dependency org.jetbrains.kotlin:kotlin-build-tools --configuration kotlinAbiValidationCompatClasspath`, I get
> 
> ```
> org.jetbrains.kotlin:kotlin-build-tools-api:2.4.0
> +--- org.jetbrains.kotlin:kotlin-build-tools-cri-impl:2.4.0
> |    +--- kotlinAbiValidationCompatClasspath (requested org.jetbrains.kotlin:kotlin-build-tools-cri-impl:{strictly 2.4.0})
> |    \--- org.jetbrains.kotlin:kotlin-build-tools-impl:2.4.0
> |         \--- kotlinAbiValidationCompatClasspath (requested org.jetbrains.kotlin:kotlin-build-tools-impl:{strictly [2.4.0-Beta2, 2.5.0)})
> ```
> 
> I don't really like this change as it might work, but this is an issue that should be prevented with lockfiles (?) and fixed upstream as it doesn't make sense there is a beta version here. Maintaining this extra code shouldn't be on us.

I don't like it either, but that allow us to keep the CI check. We should open an issue, merge this until the issue is addressed wdyt?

jpelgrom:
I was looking for issues but the Kotlin issue tracker isn't easy to search (at least for me). We could consider merging this to unblock but then need to reference the issue IN the code so we know when to remove.

TimoPtr:
Issue created https://youtrack.jetbrains.com/issue/KT-87220/Kotlin-Gradle-plugin-resolves-kotlinAbiValidationCompatClasspath-to-newer-beta-Kotlin-artifacts-during-dependency-locking I'm going to add it to the code

PR #7092, opened 2026-06-25
Introduce HaShortcutManager and migration to deep link storage

### LINKED ISSUE #7086: Migrate away from WebViewActivity to open shortcuts
### Description

Today we have shortcuts storing WebViewActivity intent. We need a path to move onto LinkActivity to prepare the removal of the Activity.

### Additional context

We need a migration logic that recreates all the shortcuts, but the application might be started from a shortcut so we still need to register the WebViewActivity for it.
We could always keep a WebViewActivity that if trigger would invoke the migration of the shortcut and forward the request to LinkActivity

LinkActivity needs to support a new param `server_id`

TimoPtr:
## Summary

Closes #7086, by introducing a small manager that handle the creation/migration of the shortcut. In order to make them as much as possible agnostic I've decided to use the deep link path and migrating the Intent to an action + URI that we need to maintain in all case.

I've added also a migration path from the Application class.
I've added also a handling in the WebViewActivity that would only trigger one time then after the shortcut is going to be migrated.

## Checklist

## Any other notes

I've tested the migration and it works, but I might have missed some cases.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/launch/link/LinkHandler.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/launch/link/LinkHandler.kt
@@ -5,6 +5,7 @@ import androidx.core.net.toUri
 import io.homeassistant.companion.android.common.data.servers.ServerManager
 import io.homeassistant.companion.android.common.util.FailFast
 import io.homeassistant.companion.android.database.server.Server
+import io.homeassistant.companion.android.frontend.navigation.FrontendTarget
 import javax.inject.Inject
 import timber.log.Timber
 
@@ -15,12 +16,33 @@ private const val NAVIGATE_URL_PATH = "/navigate/"
 private const val MY_BASE_DOMAIN = "my.home-assistant.io"
 private const val BASE_MY_REDIRECT_URL = "https://$MY_BASE_DOMAIN$REDIRECT_URL_PATH"
 
-private const val DEEP_LINK_SCHEME = "homeassistant"
+internal const val HA_DEEP_LINK_SCHEME = "homeassistant"
 
 private const val INTERNAL_MY_REDIRECT_PREFIX = "_my_redirect/"
 
 private const val MOBILE_PARAM = "mobile"
 private const val MOBILE_VALUE = "1"
+
+private const val SERVER_PARAM = "server"
+private const val SERVER_ID_PARAM = "server_id"
+private const val MORE_INFO_ENTITY_ID_PARAM = "more-info-entity-id"
+
+/**
+ * Builds the `homeassistant://navigate` deep link that [LinkHandler] resolves back to [target] on
+ * the server identified by [serverId]. Used for persisted intents (e.g. shortcuts) so they target
+ * the stable, exported [LinkActivity] entry point rather than an internal activity.
+ */
+internal fun navigateDeepLinkUri(target: FrontendTarget, serverId: Int): Uri {
+    val builder = Uri.Builder()
+        .scheme(HA_DEEP_LINK_SCHEME)
+        .authority(NAVIGATE_URL_PATH.removeSurrounding("/"))
+    when (target) {
+        is FrontendTarget.EntityMoreInfo -> builder.appendQueryParameter(MORE_INFO_ENTITY_ID_PARAM, target.entityId)
+        is FrontendTarget.Path -> builder.path(target.path)
+        FrontendTarget.Default -> Unit
+    }
+    return builder.appendQueryParameter(SERVER_ID_PARAM, serverId.toString()).build()
+}
 
 /**
  * Sealed class that represents the destination of a link.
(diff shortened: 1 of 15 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/launch/link/LinkHandler.kt
```diff
@@ -15,13 +16,34 @@ private const val NAVIGATE_URL_PATH = "/navigate/"
...
+    val builder = Uri.Builder()
+        .scheme(HA_DEEP_LINK_SCHEME)
+        .authority(NAVIGATE_URL_PATH.removeSurrounding("/"))
+    when (target) {
+        is FrontendTarget.EntityMoreInfo -> builder.appendQueryParameter(MORE_INFO_ENTITY_ID_PARAM, target.entityId)
+        is FrontendTarget.Path -> builder.path(target.path)
+        FrontendTarget.Default -> Unit
+    }
+    return builder.appendQueryParameter(SERVER_ID_PARAM, serverId.toString()).build()
+}
```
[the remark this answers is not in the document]
TimoPtr: Good point, I don't know if shortcut can contains other query or fragment. Maybe when targeting a dashboard? 

@jpelgrom any opinion?

app/src/main/kotlin/io/homeassistant/companion/android/launch/link/LinkHandler.kt
```diff
@@ -181,37 +203,57 @@ class LinkHandlerImpl @Inject constructor(private val serverManager: ServerManag
...
+     * otherwise the `server` friendly-name lookup (`default`/absent uses the active server).
+     */
+    private suspend fun resolveNavigateServerId(uri: Uri): Int? {
+        uri.getQueryParameter(SERVER_ID_PARAM)?.toIntOrNull()?.let { return it }
+        val serverName = uri.getQueryParameter(SERVER_PARAM).takeIf { !it.isNullOrBlank() }
+        return when (serverName) {
+            "default", null -> serverManager.getServer()?.id
+            else -> serverManager.servers().find { it.friendlyName.equals(serverName, ignoreCase = true) }?.id
+        }
     }
```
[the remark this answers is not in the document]
TimoPtr: I think it's fine to land in the onboarding if that's really the case. But if this scenario happen something is already wrong on the user side.

app/src/main/kotlin/io/homeassistant/companion/android/launch/link/LinkHandler.kt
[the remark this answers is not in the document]
jpelgrom: The UI for this has always mentioned dashboard or dashboard view or dashboard, so not officially supported as that is a normal `/lovelace`, `/my-custom-dash/bathroom` etc. I also don't remember actual usage of query params here other than Assist before we introduced the Assist shortcut, so putting it in `builder.path` is OK for me.

app/src/main/kotlin/io/homeassistant/companion/android/settings/shortcuts/HaShortcutManager.kt
```diff
@@ -0,0 +1,213 @@
...
+ * rewrites those to the current format so they keep working once `WebViewActivity` is removed.
+ */
+@Singleton
+class HaShortcutManager @Inject constructor(
+    @ApplicationContext private val context: Context,
+    private val iconDialogCompat: IconDialogCompat,
+) {
+
+    /**
+     * Builds a [ShortcutInfoCompat] that launches [path] on [serverId] through [LinkActivity].
```
jpelgrom:
```suggestion
     * Builds a [ShortcutInfoCompat] that launches [path] on [serverId] through a
     * `homeassistant://navigate` deep link.
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/shortcuts/HaShortcutManager.kt
```diff
@@ -0,0 +1,213 @@
...
+            }
+
+            else -> null
+        }
+    }
+
+    /**
+     * Migrates shortcuts created by older app versions (which launched `WebViewActivity` directly)
+     * to the current [LinkActivity] deep-link format, updating dynamic and pinned shortcuts in place
+     * while preserving their label, icon and target.
```
jpelgrom:
```suggestion
     * to the current deep link format, updating dynamic and pinned shortcuts in place while
     * preserving their label, icon and target.
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/shortcuts/HaShortcutManager.kt
```diff
@@ -0,0 +1,213 @@
...
+import io.homeassistant.companion.android.widgets.assist.AssistShortcutActivity
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import timber.log.Timber
+
+/**
+ * Key used by WebViewActivity to store the server used in the shortcut within the Intent.
+ * The value need to stay the same otherwise it's going to break shortcuts.
```
jpelgrom: "The value need**s**", but I actually prefer a stronger worded version:

```suggestion
 * This value cannot be changed, otherwise it's going to break shortcuts.
```

app/src/main/kotlin/io/homeassistant/companion/android/settings/shortcuts/HaShortcutManager.kt
```diff
@@ -0,0 +1,213 @@
...
+ */
+internal const val SHORTCUT_EXTRA_SERVER = "server"
+
+/**
+ * Key used by WebViewActivity to store the path used in the shortcut within the Intent.
+ * The value need to stay the same otherwise it's going to break shortcuts.
```
jpelgrom:
```suggestion
 * This value cannot be changed, otherwise it's going to break shortcuts.
```

jpelgrom (approved):
I like the approach of using our deep link scheme rather than yet another activity.

 - Note we should not document`?server_id=x` for end users as the IDs are internal to the app.
 - I understand agree with the manager concept for now but think it should be integrated into the ViewModel of the pending rewrite once we stop migrating.

Only minor suggestions for documentation, no functional changes. Tested functionality and migration on a real device as well with both shortcut types (dashboard / entity).

PR #7093, opened 2026-06-25
Avoid failing InstrumentationTest when Pull request workflow fail

TimoPtr:
## Summary

Today when the Pull Request workflow fail we fail the Instrumentation Test job. This is annoying because we get two notification (email for instance) for one failure. This PR simply skip the Instrumentation Test job if the workflow failed. 
The downside is that the check is not going to be created but it's fine since the workflow failed anyway. It might just leave the yellow dot on the PR instead of a cross (to be checked).

### DIFF
--- a/.github/workflows/pr-emulator-wtf.yml
+++ b/.github/workflows/pr-emulator-wtf.yml
@@ -18,8 +18,7 @@ jobs:
     environment: ui-test
     if: >
       github.event.workflow_run.event == 'pull_request'
-      && github.event.workflow_run.conclusion != 'cancelled'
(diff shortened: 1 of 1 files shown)


jpelgrom:
I noticed this too but didn't check how to resolve it yet, nice if it works. Waiting to see it in action as there is one failure ⏳

TimoPtr:
> I noticed this too but didn't check how to resolve it yet, nice if it works. Waiting to see it in action as there is one failure ⏳

We need to merge this PR to actually see it in action since it uses the code from `main`.

PR #7094, opened 2026-06-25
Use more precise error and action for the ConnectionError

TimoPtr:
## Summary

This PR is all about making the error screen replicate the old actions we had in the WebViewActivity. I've added screenshot for all the variants.

## Checklist

## Any other notes

I decided to not add the isInternalOverride since it is quite fragile and that the new error screen should help diagnostic better. Also we are always offering a way to reach the settings.


jpelgrom:
Did you consider the actions from the WebViewActivity before re-implementing them? Now is our chance with more room for UI and custom logic. Looking at the screenshots what I notice is:

 - 'Refresh external URL' isn't very descriptive as we actually don't refer to _external_ URL in the app UI
 - If there are 3 actions, do you put settings in the middle or last?
 - Sometimes the action in the text doesn't match the primary action
   - "install the required credential on your [device] and try again" without a system settings deeplink and/or retry option
   - "review the connection settings" should point you to settings or maybe the system's network sheet, not refresh

TimoPtr:
> Did you consider the actions from the WebViewActivity before re-implementing them? Now is our chance with more room for UI and custom logic. Looking at the screenshots what I notice is:

I did I think made a 1:1 mapping between the existing actions and the new one for each kind of error. Except the override to local that I think we should drop.

> 
>     * 'Refresh external URL' isn't very descriptive as we actually don't refer to _external_ URL in the app UI

What is your suggestion just "refresh"?

> 
>     * If there are 3 actions, do you put settings in the middle or last?

I was hoping to discuss this in the PR I doesn't really like the design. If you have suggestions.

> 
>     * Sometimes the action in the text doesn't match the primary action
>       
>       * "install the required credential on your [device] and try again" without a system settings deeplink and/or retry option
>       * "review the connection settings" should point you to settings or maybe the system's network sheet, not refresh

I'm going to look at how we can better look at this kind of issue, maybe we can reproduce the pattern we have in BlockInsecure with a small card with text and action.

TimoPtr:
@jpelgrom I've added more actions on specific errors.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/navigation/FrontendEvent.kt
```diff
@@ -39,6 +39,17 @@ sealed interface FrontendEvent {
...
 
+    /** Open the OS security settings screen (client-certificate installation). */
+    data object OpenSecuritySettings : FrontendEvent
+
+    /** Open the store/app page of the device's current WebView provider so the user can update it. */
+    data object UpdateWebView : FrontendEvent
+
+    /**
+     * Restart the app from scratch.
+     */
```
jpelgrom:
```suggestion
    /** Restart the app from scratch. */
```

jpelgrom (approved):
So many scenarios... I think I tested all but good to do a beta cycle or two in case something was missed after switching the implementation.

common/src/main/res/values/strings.xml
```diff
@@ -1490,6 +1494,8 @@
     <string name="entity_picker_no_entity_found">No entity found</string>
     <string name="entity_picker_no_entity_found_for">No entity found for %s</string>
     <string name="fail_to_navigate_to_uri">No app available to open %s</string>
+    <string name="security_settings">the device security settings</string>
+    <string name="system_webview">the system WebView</string>
```
NoRi2909, line 1498: Asking for translations, because grammar is much more complicated in German, e.g.:

These strings are combined into a single error message, I assume.
Any additional strings that are used for the second half of the sentence?

TimoPtr: It goes with the above string
```
    <string name="fail_to_navigate_to_uri">No app available to open %s</string>
```
So something like 
`No app available to open the device security settings`
`No app available to open the system WebView`

NoRi2909: Thanks for the quick reply. So no other strings or names that are used to create additional messages?

I'm asking because the first part of the sentence was already present in the strings.

TimoPtr: Not at the moment (sorry for the late response).

PR #7040, opened 2026-06-17
Implement SensorRepository on top of SensorDao to handle default

TimoPtr:
## Summary

This PR wrap the access of to the `SensorDao` into the `SensorRepository`. The repository has a new logic that make sure that we always have all the sensors return even if they are not yet in the DAO because they don't need to be there at the moment. It uses the `Set<BasicSensor>` as source of truth.  

Fix #7025 

## Checklist

- [x] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

PR #7107, opened 2026-06-30
Extract SensorReceiverBase update logic into SensorUpdater

TimoPtr:
## Summary

This PR is a wider refactor required to merge at the same time as #7106 and #7040, it fixes issues where we were using `SensorReceiver` instance, we shouldn't because it's android components instantiated by the system. To achieve that I had to modify the bluetooth manager and remove the companion function that are now within the manager directly, so that we can use a new `SensorUpdater`. The `SensorUpdater` is a class that we can inject and use to require an update of sensors. 
I had also to move the stop of the beacon into its own receiver it was mixed within the `SensorReceiver`.

Overall it simplifies a bit the code by removing some parameters and split responsibilities, I did add some very basic tests but it doesn't cover everything in the `SensorUpdater`. 

## Checklist

## Any other notes

We could merge this one onto the base one if we want only one commit.


common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/DeviceCommands.kt
```diff
@@ -196,6 +187,6 @@ suspend fun commandBleTransmitter(
         )
     }
     bluetoothSensorManager.requestSensorUpdate()
-    SensorUpdateReceiver.updateSensors(context)
+    bluetoothSensorManager.requestBluetoothSensorUpdate()
```
jpelgrom: This is confusingly named IMO, especially compared to the call the line above. Updating sensors is the sensor values. Sending the sensor values should be named differently.

jpelgrom (commented):
Already tested functionality with previous PRs, as the other sensor PRs need this to build. Only one minor comment about the use of "update" which now can mean two things.

Copilot/Lint can be ignored for now as you've copied it, I would like to avoid any functional changes we don't need.

common/src/main/kotlin/io/homeassistant/companion/android/common/bluetooth/ble/MonitoringManager.kt
```diff
@@ -90,12 +85,10 @@ class MonitoringManager {
...
         }
-        val stopScanningIntent = Intent(context, SensorUpdateReceiver::class.java)
-        stopScanningIntent.action = SensorReceiverBase.ACTION_STOP_BEACON_SCANNING
         val stopScanningPendingIntent = PendingIntent.getBroadcast(
             context,
             0,
-            stopScanningIntent,
+            StopBeaconScanningReceiver.stopScanningIntent(context),
             PendingIntent.FLAG_MUTABLE,
         )
```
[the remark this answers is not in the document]
TimoPtr: Kept like it was before

wear/src/main/kotlin/io/homeassistant/companion/android/sensors/WearSensorModule.kt
```diff
@@ -48,5 +56,21 @@ abstract class WearSensorModule {
...
+                val intent = Intent(
+                    Intent.ACTION_VIEW,
+                    "$DEEPLINK_SENSOR_MANAGER/$sensorManagerId".toUri(),
+                    context,
+                    HomeActivity::class.java,
+                )
+                TaskStackBuilder.create(context).run {
+                    addNextIntentWithParentStack(intent)
+                    getPendingIntent(notificationId, PendingIntent.FLAG_UPDATE_CURRENT)
+                }
```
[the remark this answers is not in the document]
TimoPtr: Kept like it was

common/src/main/kotlin/io/homeassistant/companion/android/common/notifications/DeviceCommands.kt
TimoPtr: Any suggestions of name?

jpelgrom: sendBluetoothSensorUpdate?

jpelgrom (approved):

PR #7112, opened 2026-07-01
Implement command_volume_level_step notification command for relative volume adjustments

marazmarci:
## Summary
This PR adds a new `command_volume_level_step` notification command (alongside the existing `command_volume_level`) to add support for relative volume adjustments. It enables users to create responsive volume up/down control automations, for example for physical media control buttons through HA (eg. [this Zigbee device](https://www.zigbee2mqtt.io/devices/E2123.html)). Currently the best option for implementing such automation is to first read the current state of the music volume sensor, then add/subtract the desired adjustment value, and then fire a `command_volume_level` notification command to set the new volume. The problem with this approach is slowness. Quick successive button presses would result in most of the button presses to be seemingly ignored, because the music volume sensor doesn't update quickly enough.

The parameters of the new command are exactly the same as for `command_volume_level`. The only difference is in behavior of the `command` key. Here's an example service call YAML to decrease music volume by 2 steps:

```yaml
service: notify.mobile_app_<device_name>
data:
  message: command_volume_level_step
  data:
    media_stream: music_stream
    command: -2
```

## Checklist

- [x] ~~New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).~~
  - No, as `MessagingManager` didn't have any unit tests to begin with.

**I'll open the doc update PR and the PR for adding the new notification command in the [mobile-apps-fcm-push](https://github.com/home-assistant/mobile-apps-fcm-push) repo when this PR is getting close to be merged :hourglass_flowing_sand:**

## Any other notes

This is a reworked & rethought version of a PR I opened (and abandoned) a while ago: #4056 (closed without merging)
In that PR, I implemented the same functionality with a different approach: I extended the existing `command_volume_level` notification command with a `relative: true/false` property. Now I changed my mind, because I think adding a separate notification command feels cleaner, expresses the distinction more clearly, and is more similar to existing patterns in HA, like the `brightness_step` key for the `light.turn_on` command.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/ApplicationModule.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/ApplicationModule.kt
@@ -2,6 +2,7 @@ package io.homeassistant.companion.android
 
 import android.app.DownloadManager
 import android.content.Context
+import android.media.AudioManager
 import androidx.core.content.getSystemService
 import androidx.work.WorkManager
 import dagger.Module
@@ -73,5 +74,11 @@ object ApplicationModule {
     @Singleton
     fun providesDownloadManager(@ApplicationContext context: Context): DownloadManager? {
         return context.getSystemService<DownloadManager>()
+    }
+
+    @Provides
+    @Singleton
+    fun providesAudioManager(@ApplicationContext context: Context): AudioManager? {
+        return context.getSystemService<AudioManager>()
     }
 }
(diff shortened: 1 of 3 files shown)


TimoPtr:
@marazmarci thanks for pushing this through. I've made some changes that you'll have to add to your PR. Is it an option to extend the existing command with more parameters to achieve the same goals as what you are trying to achieve?

PR #7114, opened 2026-07-02
Reduce max heap size to 4G for screenshot test

TimoPtr:
## Summary

CI is failing on screenshot testing again, it seems that we are using more RAM that the runner can allow us. This PR tries to reduce the max heap size to 4g instead of 5g. It passes locally let see if it helps on the CI.


jpelgrom:
🤞

Very odd issue, but hopefully this helps

jpelgrom:
This doesn't appear to have helped enough, screenshot tests jobs are failing even more frequently now with the task being cancelled (system killing it presumably).

PR #7115, opened 2026-07-02
Introduce RoomEntityMutablePropertiesIssue lint rule

TimoPtr:
## Summary

This PR introduce a new lint rules that forbid the usage of `var` within Room Entity and View. It is ready for the Room3 that just got released.

This PR does not address the actual issue we have in the code base but ignore then. Most of them are going to be addressed in follow up PRs.

I wanted to make parametrized test and found out that the lint test API now supports Jupiter so I've migrated to it so I can write the parametrized tests.

## Checklist

### DIFF
--- a/lint/src/main/kotlin/io/homeassistant/lint/room/RoomEntityMutablePropertiesIssue.kt
+++ b/lint/src/main/kotlin/io/homeassistant/lint/room/RoomEntityMutablePropertiesIssue.kt
@@ -1,0 +1,80 @@
+package io.homeassistant.lint.room
+
+import com.android.tools.lint.client.api.UElementHandler
+import com.android.tools.lint.detector.api.Category
+import com.android.tools.lint.detector.api.Detector
+import com.android.tools.lint.detector.api.Implementation
+import com.android.tools.lint.detector.api.Issue
+import com.android.tools.lint.detector.api.JavaContext
+import com.android.tools.lint.detector.api.Scope
+import com.android.tools.lint.detector.api.Severity
+import com.android.tools.lint.detector.api.SourceCodeScanner
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.uast.UClass
+import org.jetbrains.uast.UField
+import org.jetbrains.uast.kotlin.getKotlinMemberOrigin
+
+private val ROOM_ENTITY_ANNOTATIONS = listOf(
+    "androidx.room.Entity",
+    "androidx.room.DatabaseView",
+    "androidx.room3.Entity",
+    "androidx.room3.DatabaseView",
+)
+
+object RoomEntityMutablePropertiesIssue {
+
+    @JvmField
+    val ISSUE = Issue.create(
+        id = "RoomEntityMutableProperty",
+        briefDescription = "Room entity properties should be immutable",
+        explanation = """
+            Room entity classes should use immutable properties to avoid mutating persisted state in place.
+            Use `val` properties and create updated copies instead.
+        """.trimIndent(),
+        category = Category.CORRECTNESS,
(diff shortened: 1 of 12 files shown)


TimoPtr:
It doesn't prevent us to use mutable list but it's already a start.

jpelgrom:
I already merged one of the locations with issues, and the other one is almost ready and just needs a minor tweak. So you can remove the baseline update probably :D

Would be nice if Room offered this built-in, but that doesn't appear to be the case. Google is consistently using `val` in the documentation however.

lint/src/main/kotlin/io/homeassistant/lint/room/RoomEntityMutablePropertiesIssue.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.uast.UClass
+import org.jetbrains.uast.UField
+import org.jetbrains.uast.kotlin.getKotlinMemberOrigin
+
+private val ROOM_ENTITY_ANNOTATIONS = listOf(
+    "androidx.room.Entity",
+    "androidx.room.DatabaseView",
+    "androidx.room3.Entity",
+    "androidx.room3.DatabaseView",
```
jpelgrom, line 22: Views can only be selected, not inserted/updated/removed (see [note box](https://developer.android.com/training/data-storage/room/creating-views)), so mutating views is very unlikely. I'd remove them as it's not expected; if you still want to ban var there because the risk of Compose not getting changes then... (see next comment)

lint/src/main/kotlin/io/homeassistant/lint/room/RoomEntityMutablePropertiesIssue.kt
```diff
@@ -0,0 +1,80 @@
...
+
+object RoomEntityMutablePropertiesIssue {
+
+    @JvmField
+    val ISSUE = Issue.create(
+        id = "RoomEntityMutableProperty",
+        briefDescription = "Room entity properties should be immutable",
+        explanation = """
+            Room entity classes should use immutable properties to avoid mutating persisted state in place.
+            Use `val` properties and create updated copies instead.
```
jpelgrom: ... the issue description and explanation should be updated to also mention it's not views.


```suggestion
        briefDescription = "Room entity/view properties should be immutable",
        explanation = """
            Room entity and view classes should use immutable properties to avoid mutating persisted state
            in place. Use `val` properties and create updated copies instead.
```

lint/src/main/kotlin/io/homeassistant/lint/room/RoomEntityMutablePropertiesIssue.kt
```diff
@@ -0,0 +1,80 @@
...
+                override fun visitClass(node: UClass) {
+                    if (node.isEntity()) {
+                        node.fields
+                            .filter { it.isMutableKotlinProperty() }
+                            .forEach { field ->
+                                context.report(
+                                    ISSUE,
+                                    field,
+                                    context.getNameLocation(field),
+                                    "Room entity properties should be immutable. Use `val` instead of `var`.",
```
jpelgrom: If you accept https://github.com/home-assistant/android/pull/7115/changes#r3516181143

```suggestion
                                    "Room entity/view properties should be immutable. Use `val` instead of `var`.",
```

lint/src/main/kotlin/io/homeassistant/lint/room/RoomEntityMutablePropertiesIssue.kt
TimoPtr: I also want to give implicit meaning that this class is not meant to be mutated. I'll follow your proposal 👍🏻 of changing the description.

PR #7117, opened 2026-07-02
Fix settings update issue by making SensorSetting immutable

TimoPtr:
## Summary

The sensor settings update where not showing as per @jpelgrom found out https://github.com/home-assistant/android/pull/7041#issuecomment-4859534223 with the recent changes we've made.

I've fixed it by making SensorSetting immutable and exposing StateFlow for it. The issue was that now that the sensor is immutable the way we've updated the Settings it was not triggering a change on the compose side because the object itself was still the same reference.

## Checklist

## Any other notes

Based on #7041

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
@@ -70,7 +70,7 @@ class SensorDetailViewModel @Inject constructor(
         private const val SENSOR_SETTING_TRANS_KEY_PREFIX = "sensor_setting_"
 
         // Keep the database-backed flows hot briefly across config changes before stopping collection.
-        private const val STOP_TIMEOUT = 500.milliseconds
+        private val STOP_TIMEOUT = 500.milliseconds
 
         data class PermissionsDialog(val serverId: Int?, val permissions: Array<String>? = null)
         data class LocationPermissionsDialog(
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
```diff
@@ -63,6 +67,9 @@ class SensorDetailViewModel @Inject constructor(
     companion object {
         private const val SENSOR_SETTING_TRANS_KEY_PREFIX = "sensor_setting_"
 
+        // Keep the database-backed flows hot briefly across config changes before stopping collection.
+        private const val STOP_TIMEOUT_MILLIS = 5_000L
```
jpelgrom: You can use a duration here instead of millis which [should be preferred](https://developers.home-assistant.io/docs/android/best_practices?_highlight=duration#time-and-duration). 5 seconds also feels very long for configuration changes.

(This also needs updated imports so simply accepting suggestion won't work)

```suggestion
        private val STOP_TIMEOUT = 1.seconds
```

jpelgrom (commented):
Tested and this correctly fixes the issue. Thanks for the fix!

app/src/main/kotlin/io/homeassistant/companion/android/settings/sensor/SensorDetailViewModel.kt
TimoPtr: I'm ashamed ... I used claude for this and I didn't even see that it was not following the convention I wrote... 🥹

TimoPtr: I also made it 500ms like we had in other places.

Issue #111, opened 2019-12-05
Mobile app invalid webhook payload for vertical accuracy


dshokouhi:
So I had successfully installed 1.1.1 on a pixel 3 and pixel 4 with location working fine and no errors in my HA logs.  I then decided to install it on my Amazon Fire HD8 that has google services and the play store.  App installed and created a device tracker that did not update.  No errors were seen in the logs.  I then installed google maps on the tablet to see if a missing service was found and the following error shows up now.

`2019-12-04 16:50:13 ERROR (MainThread) [homeassistant.components.mobile_app.webhook] Received invalid webhook payload: expected int for dictionary value @ data['vertical_accuracy']. Got None`

This error shows up every few minutes so I suspect the tablet is attempting to send back data but it does not have `vertical_accuracy` to send back. 

I know this is probably a stretch but I think this error should probably be fixed?

JBassett:
Not sure if we can fix that well in app.  Vertical accuracy depends on a certain version of Android and above.  We might need to enhance the mobile app integration to make it optional rather than mandatory.  Or we could hard code an accuracy for devices that don't support it.

dshokouhi:
@JBassett maybe just send back `0` if there is none?  I also wasn't sure where to file this bug.

robbiet480:
@balloob just fixed this in `mobile_app` late last night at https://github.com/home-assistant/home-assistant/pull/29381. Will be in next version of HA, not the app.

balloob:
I fixed allowing `altitude` to be negative. I did not allow `vertical_accuracy` to be None. 

For vertical accuracy, in the Android app can we either send in 0 or not add the key if it's null ?

reharmsen:
When will this be released?
I am noticing this same issue with an Samsung S6, while my Nokia 6 works just fine.

`Received invalid webhook payload: expected int for dictionary value @ data['vertical_accuracy']. Got None`

balloob:
Has been released as 1.1.2

djxela:
`Received invalid webhook payload: required key not provided @ data['device_name']. Got None`

I have this same problem with Samsung S7.  Mobile App version 1.7.2-128

PR #7119, opened 2026-07-02
Adjust SystemUI based on Frontend theme and adjust edge-to-edge

TimoPtr:
## Summary

This is the last feature missing from the WebViewActivity before its removal. I decided also to enable the Edge to Edge support for the frontend while omitting the top since it doesn't work well, by doing so I decided to use a min version of core that was after our last try where we identified the topbar issue where the color of the text and icon were wrong. https://github.com/home-assistant/frontend/issues/29125

## Checklist

## Screenshots

image
image


app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -1015,4 +998,62 @@ internal class FrontendViewModel @VisibleForTesting constructor(
...
+        }
+
+        _viewState.update { state ->
+            if (state is FrontendViewState.Content) {
+                state.copy(statusBarColor = colors.statusBarColor, backgroundColor = colors.backgroundColor)
+            } else {
+                state
+            }
+        }
+    }
```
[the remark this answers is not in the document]
TimoPtr: I'm not sure about this one

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
```diff
@@ -539,27 +570,31 @@ private fun SafeHAWebView(
...
-            contentState?.statusBarColor?.Overlay(
-                modifier = Modifier
-                    .height(insetsPaddingValues.calculateTopPadding())
-                    .fillMaxWidth()
-                    // We don't want the status bar to color the left and right areas
-                    .padding(insets.only(WindowInsetsSides.Horizontal).asPaddingValues()),
-            )
-        }
+        // The top is never drawn edge-to-edge: we always reserve and color the status bar strip until
+        // the frontend addresses https://github.com/home-assistant/frontend/issues/29125
```
jpelgrom: Not sure if it's just this issue - we literally cannot fix the difference in contrast when a menu/drawer opens, it requires more than the linked issue.

```suggestion
        // The top is never drawn edge-to-edge: we always reserve and color the status bar strip.
        // Before considering removal, https://github.com/home-assistant/frontend/issues/29125 and
        // contrast with drawers (left/right status bar cannot be different colors) should be addressed.
```

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
```diff
@@ -121,6 +122,82 @@ sealed interface WebViewAction {
...
+                )},${computedStyleToken("--primary-background-color")}].join('$THEME_COLOR_SPACER')"
+
+            /** Matches the CSS `rgb(r, g, b)` notation the frontend emits for its computed theme tokens. */
+            private val RGB_REGEX = Regex("""rgb\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*\)""")
+
+            /**
+             * Parses the [THEME_COLORS_SCRIPT] result into [ThemeColors]. Returns `null` when [raw]
+             * is absent or does not contain exactly two tokens.
+             *
+             * The result is a JSON string literal, e.g. `"\"rgb(1,2,3)-SPACER-rgb(4,5,6)\""`.
```
jpelgrom: No, the result is a data class. Do you mean the expected raw value? To be honest I think this is detail not needed here, `THEME_COLORS SCRIPT result` is already enough info for a public description.

jpelgrom (commented):
Still a bit disappointing we cannot do the top bar, but this is probably the best for now.

You do like very small functions I see, thankfully there is no longer a real method count limit...

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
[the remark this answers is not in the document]
jpelgrom: `updateThemeColors` is only called from the frontend's external bus when connected or when the theme is updated. I don't realistically see this happening, if there is no webview right now it'll be because it's finishing.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
```diff
@@ -121,6 +122,82 @@ sealed interface WebViewAction {
...
+             * Reads the frontend theme tokens for the status bar (`--app-header-background-color`) and the
+             * page background (`--primary-background-color`) as a single string joined by [THEME_COLOR_SPACER].
+             */
+            private val THEME_COLORS_SCRIPT =
+                "[${computedStyleToken(
+                    "--app-header-background-color",
+                )},${computedStyleToken("--primary-background-color")}].join('$THEME_COLOR_SPACER')"
+
+            /** Matches the CSS `rgb(r, g, b)` notation the frontend emits for its computed theme tokens. */
+            private val RGB_REGEX = Regex("""rgb\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*\)""")
```
jpelgrom, line 174: This is a good reimplementation of what is in WebViewActivity but turns out to be broken for me on core 2026.7.1. The result is `#1c1c1c-SPACER-#111`, so it doesn't match the regex and it falls back to default.

Not sure how long it has been broken, as my theme is a boring standard one. It did work before the top bar changed from blue -> white.

jpelgrom: Sorry, the issue is actually in your parsing function. You are using the regex there, whereas the previous implementation had a fallback if there was a hex color with `toColorInt`

jpelgrom: Alternative function which better mirrors the current implementation and I've tested to fix the issue:

```kotlin
/**
 * Parses a color read from the frontend into a Compose [Color]. Returns `null` when
 * - the value is not a valid `rgb()` triple in the 0-255 range
 * - the value is not a valid hex color
 * - the value is not a supported color name like `red`, `blue`, `fuchsia`, ...
 */
private fun String.toWebViewColorOrNull(): Color? {
    val match = RGB_REGEX.matchEntire(trim())
    return if (match != null) {
        val (r, g, b) = match.destructured
        val red = r.toColorChannelOrNull() ?: return null
        val green = g.toColorChannelOrNull() ?: return null
        val blue = b.toColorChannelOrNull() ?: return null
        Color(red = red, green = green, blue = blue)
    } else {
        try {
            val asInt = trim().toColorInt()
            Color(red = asInt.red, green = asInt.green, blue = asInt.blue)
        } catch (_: IllegalArgumentException) {
            null
        }
    }
}
```
For reproduction, you can keep the device in light mode but manually set the frontend color to dark mode in your user profile. Without this change the status bar keeps the device theme color (white), with this change it updates as expected.

TimoPtr: Thanks for fixing it and sharing the test case.

jpelgrom (approved):

PR #7135, opened 2026-07-06
Pause WebView when activity is stopped

TimoPtr:
## Summary

This is an attempt to pause the WebView when the app is paused instead of unloading like in #5051 to address #979. We need to validate the behavior and validate that the UX remains good for most of our users.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
@@ -381,7 +381,7 @@ internal fun WebViewStopLifecycleEffect(webView: WebView?) {
                     pauseTimers()
                 }
                 Lifecycle.Event.ON_START -> webView?.apply {
-                    Timber.d("Webview resumed")
+                    Timber.d("Webview started")
                     resumeTimers()
                     onResume()
                 }
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
@@ -49,6 +49,7 @@ import androidx.compose.ui.unit.LayoutDirection
 import androidx.compose.ui.viewinterop.AndroidView
 import androidx.hilt.lifecycle.viewmodel.compose.hiltViewModel
 import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleEventObserver
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.compose.LifecycleEventEffect
 import androidx.lifecycle.compose.LocalLifecycleOwner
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
```diff
@@ -359,6 +360,37 @@ private fun FrontendScreenEffects(
...
+    DisposableEffect(webView, lifecycleOwner) {
+        val observer = LifecycleEventObserver { _, event ->
+            when (event) {
+                Lifecycle.Event.ON_STOP -> webView?.apply {
+                    Timber.d("Webview stopped")
+                    onPause()
+                    pauseTimers()
+                }
+                Lifecycle.Event.ON_START -> webView?.apply {
+                    Timber.d("Webview resumed")
```
jpelgrom: Let's not use different lifecycle events in logs to avoid confusion

```suggestion
                    Timber.d("Webview started")
```

jpelgrom (commented):
The linked issue here is not very clear as to what is exactly the cause, as feedback so far also shows. Still, this appears as a better approach to me than completely unloading the view and it is becoming clear this needs some attention given the recent increase in activity.

I wonder if we need a small delay in the`ON_STOP` actions, to help smooth out configuration changes. Did you test with this?

Over on Discord you also mentioned background streaming with Music Assistant. I don't use it myself but guess this will break it...

TimoPtr:
> The linked issue here is not very clear as to what is exactly the cause, as feedback so far also shows. Still, this appears as a better approach to me than completely unloading the view and it is becoming clear this needs some attention given the recent increase in activity.
> 

If you test it I don't feel that I changes anything in terms of UX when opening back the app. Seems like it manage to reconnect right away without even showing the reconnect snackbar of the frontend.

> I wonder if we need a small delay in the`ON_STOP` actions, to help smooth out configuration changes. Did you test with this?
> 

I didn't but I had this idea too but since the UX is working properly I didn't add any delay.

> Over on Discord you also mentioned background streaming with Music Assistant. I don't use it myself but guess this will break it...

I need to test this to see how it behaves.

jpelgrom:
> > I wonder if we need a small delay in the`ON_STOP` actions, to help smooth out configuration changes. Did you test with this?
> 
> I didn't but I had this idea too but since the UX is working properly I didn't add any delay.

I'm also not seeing UX issues or reconnecting yet, but maybe it's different with slower devices and/or unstable connections. I think adding a small (1 second or so) delay could help there if there was any issue, while it's not that difficult to add.

TimoPtr:
> > > I wonder if we need a small delay in the`ON_STOP` actions, to help smooth out configuration changes. Did you test with this?
> > 
> > 
> > I didn't but I had this idea too but since the UX is working properly I didn't add any delay.
> 
> I'm also not seeing UX issues or reconnecting yet, but maybe it's different with slower devices and/or unstable connections. I think adding a small (1 second or so) delay could help there if there was any issue, while it's not that difficult to add.

Testing with Web radio within HA does showcase that there is already a small built-in delay, should we still add the delay?

TimoPtr:
After more testing using my Pixel 9 Pro with Android 16 I end up into a weird scenario where even by killing the app from the recent apps the radio keeps playing (even after waiting minutes). 

I think we should not support background playing within a proper media session. It might appear as a regression for users that wants to use the radio in background but to properly support it we need a media session and disable the webView pause if the session is currently playing, like for PiP?

jpelgrom:
OK, I also did some testing with the media browser and don't see pauses on configuration changes. Based on that I agree that no additional delay is needed. If there was an interruption somewhere it should've been noticed with music.

Agreed media playback currently isn't really great without a session, but I have seen some reports of people using it. With the bug you've mentioned seems like it might be dependent on how the JS handles it? I'm OK with the change potentially breaking it as we never officially supported it but don't be surprised if it becomes a feature request from someone :)

jpelgrom:
@TimoPtr If you agree with https://github.com/home-assistant/android/pull/7135#discussion_r3539295816 good to merge for me :)

TimoPtr:
> @TimoPtr If you agree with https://github.com/home-assistant/android/pull/7135#discussion_r3539295816 good to merge for me :)

I agree and from what I saw the only way is for us to detect that the WebView is playing a media and handle the session ourselves. The WebView seems to not expose the session nor exposing an API for it,  we would need to hook to the JS somehow or using external bus.

jpelgrom:
> I agree and from what I saw the only way is for us to detect that the WebView is playing a media and handle the session ourselves. The WebView seems to not expose the session nor exposing an API for it, we would need to hook to the JS somehow or using external bus.

That is annoying, but out of scope for this PR

jpelgrom (approved):

PR #7149, opened 2026-07-09
Introduce GetEntitiesForDisplayUseCase

### LINKED ISSUE #7148: Make an abstraction layer to retrive Entities information from all registries independent from the version of core
### Description

Make a UseCase that is going to be responsible for retrieving all the information needed about entities independently from the version of core. It should expose a generic representation of an Entity that is going to be used for display. All the logic of fallback fields that we have in the frontend and iOS should be implemented there.

### Additional context

_No response_

TimoPtr:
## Summary

This PR adds a new UseCase to act as an adapter layer holding the logic of getting the right information to display an entity. The goal is to centralize the logic and properly test it to avoid regression.

Closes #7148.

## Checklist

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntityDisplayItem.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntityDisplayItem.kt
@@ -1,0 +1,69 @@
+package io.homeassistant.companion.android.common.data.integration.display
+
+import android.content.Context
+import androidx.compose.runtime.Immutable
+import androidx.compose.ui.unit.LayoutDirection
+import com.mikepenz.iconics.typeface.IIcon
+import io.homeassistant.companion.android.common.data.integration.Entity
+import io.homeassistant.companion.android.common.data.integration.friendlyName
+import io.homeassistant.companion.android.common.data.integration.getIcon
+
+private const val CATEGORY_CONFIG = "config"
+private const val CATEGORY_DIAGNOSTIC = "diagnostic"
+
+/**
+ * Category of an entity in the entity registry, used to separate primary entities from
+ * configuration/diagnostic ones.
+ */
+enum class EntityCategory {
+    CONFIG,
+    DIAGNOSTIC,
+    ;
+
+    companion object {
+        /** Maps the registry string value (`config`/`diagnostic`) to an [EntityCategory], or null. */
+        internal fun fromString(value: String?): EntityCategory? = when (value) {
+            CATEGORY_CONFIG -> CONFIG
+            CATEGORY_DIAGNOSTIC -> DIAGNOSTIC
+            else -> null
+        }
+    }
+}
+
+/** Fully resolved display information for an entity. */
+@Immutable
+data class EntityDisplayItem(
+    val entityId: String,
+    val name: String,
+    val icon: IIcon,
+    val areaName: String? = null,
+    val floorName: String? = null,
+    val deviceName: String? = null,
+    val isHidden: Boolean = false,
+    val entityCategory: EntityCategory? = null,
+    val displayPrecision: Int? = null,
+    val labels: List<String> = emptyList(),
+) {
+    val domain: String get() = entityId.substringBefore('.')
+
+    /**
+     * Formatted subtitle combining area and device name, adapting the separator to the
+     * layout direction. Null if the item has neither.
+     */
+    fun subtitle(layoutDirection: LayoutDirection): String? = listOfNotNull(areaName, deviceName)
+        .takeIf { it.isNotEmpty() }
+        ?.joinToString(if (layoutDirection == LayoutDirection.Ltr) " ▸ " else " ◂ ")
+
+    companion object {
+        /**
+         * Builds an item from an [Entity] alone, without any registry metadata (no
+         * area/floor/device names). Meant for callers that cannot reach the websocket API,
+         * such as the Wear favorites settings screen, and for previews.
+         */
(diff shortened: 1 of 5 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntityDisplayItem.kt
```diff
@@ -0,0 +1,69 @@
...
+    val displayPrecision: Int? = null,
+    val labels: List<String> = emptyList(),
+) {
+    val domain: String get() = entityId.substringBefore('.')
+
+    /**
+     * Formatted subtitle combining area and device name, adapting the separator to the
+     * layout direction. Null if the item has neither.
+     */
+    fun subtitle(layoutDirection: LayoutDirection): String? = listOfNotNull(areaName, deviceName)
```
bgoncal, line 49: It can be an iteration but you may need to have different subtitles depending the context, like:

- When your screens display items from multiple server, showing the server is important
- When your screen has in it's title the server name, then server is not important in the subtitle

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntityDisplayItem.kt
```diff
@@ -0,0 +1,69 @@
...
+import io.homeassistant.companion.android.common.data.integration.friendlyName
+import io.homeassistant.companion.android.common.data.integration.getIcon
+
+private const val CATEGORY_CONFIG = "config"
+private const val CATEGORY_DIAGNOSTIC = "diagnostic"
+
+/**
+ * Category of an entity in the entity registry, used to separate primary entities from
+ * configuration/diagnostic ones.
+ */
```
bgoncal:
```suggestion
```
Also understandable just by reading the code right? If thats the pattern you use in android though, ignore my comment

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/GetEntitiesForDisplayUseCase.kt
```diff
@@ -0,0 +1,226 @@
...
+     */
+    private fun resolveEntityDisplayItems(
+        entities: List<Entity>,
+        snapshot: RegistrySnapshot,
+    ): List<EntityDisplayItem> = entities.map { entity ->
+        val displayEntry = snapshot.displayEntries?.get(entity.entityId)
+        val classicEntry = snapshot.classicEntries?.get(entity.entityId)
+
+        val device = (displayEntry?.deviceId ?: classicEntry?.deviceId)?.let { snapshot.devices[it] }
+        // The area set on the entity itself wins over the area inherited from its device
```
bgoncal:
```suggestion
```

If I am not mistaken this is written also somewhere else, no need to explain the priority, if it is in the code it's the priority defined, perhaps add a test to prevent future devs tweaking it

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntityDisplayItem.kt
TimoPtr: Indeed, I'm not sure that at the moment we have such picker. We usually ask the user to pick a server first. let's iterate on this when we need a different subtitle for now I don't think it is needed.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/GetEntitiesForDisplayUseCase.kt
```diff
@@ -0,0 +1,225 @@
...
+    val displayEntries: Map<String, EntityRegistryDisplayEntry>? = null,
+    val entityCategories: Map<Int, String> = emptyMap(),
+    val classicEntries: Map<String, EntityRegistryResponse>? = null,
+    val devices: Map<String, DeviceRegistryResponse> = emptyMap(),
+    val areas: Map<String, AreaRegistryResponse> = emptyMap(),
+    val floors: Map<String, FloorRegistryResponse> = emptyMap(),
+)
+
+/**
+ * Adapter that resolves the display information (name, area, floor, device, icon, hidden,
```
jpelgrom:
```suggestion
 * Use case that resolves the display information (name, area, floor, device, icon, hidden,
```

jpelgrom (approved):
This looks correct, checked with core impl which matches. We do need to be even more careful about keeping up with any core or frontend changes here.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/GetEntitiesForDisplayUseCase.kt
```diff
@@ -0,0 +1,225 @@
...
+            labels = displayEntry?.labels.orEmpty(),
+        )
+    }
+
+    private suspend fun fetchRegistrySnapshot(serverId: Int, version: HomeAssistantVersion?): RegistrySnapshot =
+        coroutineScope {
+            val webSocketRepository = serverManager.webSocketRepository(serverId)
+            val devices = async {
+                fetchOrNull("device") { webSocketRepository.getDeviceRegistry() }
+            }
```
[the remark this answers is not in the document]
jpelgrom: This might be technically correct but if the websocket repository throws you're also not getting entities...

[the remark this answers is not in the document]
TimoPtr: If we don't get entities we wouldn't reach this code, I think we can keep it like this wdyt?

PR #7151, opened 2026-07-09
Introduce HASearchField

TimoPtr:
## Summary

Extracted the debounce search logic within a `HASearchField`, this was initially made in #6567 by @danielgomezrico. Since I'm currently redoing the EntityPicker I took the opportunity to extract it into a smaller PR. I've also added a small state object and a remember helper on top of tests.

It's built on top of our existing `HATextField`, it can be extended later.

## Checklist

- [x] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

## Screenshots

image

## Any other notes

Based on #7149

PR #7152, opened 2026-07-09
Refactor EntityPicker to use EntityDisplayState

### LINKED ISSUE #7173: Use new entity display within the entity picker
### Description

We have a new adapter layer that contains all the information for a display of an entity based on the information available. We need to use this information in the Entity picker so we don't rely on the entity friendly name anymore.

### Additional context

_No response_

TimoPtr:
## Summary

Updated the EntityPicker to use the use case to pull the data and avoid using the friendly name.

## Checklist

## Screenshots

See in the PR

## Any other notes

Based on #7151


app/src/main/kotlin/io/homeassistant/companion/android/settings/qs/ManageTilesViewModel.kt
```diff
@@ -216,34 +211,23 @@ internal class ManageTilesViewModel @Inject constructor(
...
-                    entities = entitiesDeferred.await(),
-                    entityRegistry = entityRegistryDeferred.await(),
-                    deviceRegistry = deviceRegistryDeferred.await(),
-                    areaRegistry = areaRegistryDeferred.await(),
-                )
+            getEntitiesForDisplay(serverId).collect { state ->
+                _state.update { it.copy(entityDisplayState = state) }
             }
         }
         loadEntitiesJob?.join()
```
jpelgrom, line 227: I think this is no longer needed now?

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/entity/EntityPicker.kt
```diff
@@ -278,56 +171,124 @@ internal fun EntityPicker(
...
+            strokeWidth = LOADING_INDICATOR_STROKE_WIDTH,
+        )
+    }
+    Text(
+        text = entityId,
+        style = HATextStyle.Body,
+        color = colorScheme.colorTextPrimary,
+        maxLines = 1,
+        overflow = TextOverflow.Ellipsis,
+        modifier = Modifier.weight(1f),
```
jpelgrom, line 283: This makes them centered which is weird, when loaded it isn't centered so the text jumps from center > start. Also clearly visible in the screenshot tests.

```suggestion
        textAlign = TextAlign.Start,
        modifier = Modifier.weight(1f),
```

Before (with added network delay):

https://github.com/user-attachments/assets/67164c82-62a2-4466-953d-61e7cd42674b

After:

https://github.com/user-attachments/assets/57f932be-fae8-42cc-8b39-2a86bd931d96

jpelgrom (commented):
Some of the refactoring here, around using a state for the entire sheet and not just EntityDisplayState and the gesture handling, and introducing an area registry test, feels a bit like scope creep. They could be submitted separately but I don't want an even larger PR stack.

jpelgrom:
Not strictly related to this PR, but I notice that my device trackers from entries registered by the app only show as the subtitle:

Entity picker on app with a list of device trackers in the subtitle and no main title

(ignore the fallback icon here)

In this situation the frontend appears to show it as only a title, no subtitle:

Entity picker on frontend with a list of device trackers as the main title, no subtitle

(ignore the domain here - this is in the tile card entity list which always shows domains)

app/src/main/kotlin/io/homeassistant/companion/android/settings/qs/ManageTilesViewModel.kt
TimoPtr: It is not I kept it thinking it might be useful but it should be removed as we should not assume this.

TimoPtr: I propose to not change this since the very next PR https://github.com/home-assistant/android/pull/7180 does remove it.

TimoPtr:
> Not strictly related to this PR, but I notice that my device trackers from entries registered by the app only show as the subtitle:
> Entity picker on app with a list of device trackers in the subtitle and no main title
> 
> (ignore the fallback icon here)
> 
> In this situation the frontend appears to show it as only a title, no subtitle:
> Entity picker on frontend with a list of device trackers as the main title, no subtitle
> 
> (ignore the domain here - this is in the tile card entity list which always shows domains)

I'm going to followup on this in another PR good call 👍🏻

jpelgrom (approved):

PR #7157, opened 2026-07-10
Fix server selector sheet positioning

loganrosen:
## Summary
Fixes the server selector jumping when opened with the three-finger gesture.

`ServerChooserFragment` is already hosted in a `BottomSheetDialogFragment`, so it now renders only the chooser content instead of nesting another Compose modal sheet inside the platform bottom sheet. Direct Compose callers still use `ServerChooser`, which keeps the `HAModalBottomSheet` wrapper.

This also restores the visible sheet surface for the fragment-hosted path and adds a regression test that verifies the content path does not create a nested Compose dialog while direct `ServerChooser` usage still does.

## Checklist

User Documentation: N/A

Developer Documentation: N/A

## Any other notes

Validated with:

- `./gradlew :build-logic:convention:ktlintFormat ktlintFormat --quiet`
- `./gradlew :app:testFullDebugUnitTest --tests 'io.homeassistant.companion.android.settings.server.ServerChooserTest'`

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/settings/server/ServerChooser.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/settings/server/ServerChooser.kt
@@ -82,31 +82,49 @@ fun ServerChooser(
         onDismissRequest = onDismissRequest,
         modifier = modifier,
     ) {
-        LazyColumn {
-            item {
-                Text(
-                    text = stringResource(commonR.string.server_select),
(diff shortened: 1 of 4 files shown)


loganrosen:
Closing this for now because we could only reproduce the jump on the older Play build, and current main already has a different server selector implementation. Without a current-main repro, this cleanup does not feel worth carrying as a behavior fix.

PR #7162, opened 2026-07-11
Add command_screen_off to turn the display off while the device stays awake

markfrancisonly:
## Summary

Adds a `command_screen_off` notification command that turns the display off, and extends `command_screen_on` to turn it back on. This is aimed at wall-mounted dashboards where the server decides when the panel should be dark (presence, night mode). Only the display is affected: whether the app stays reachable while the screen is off follows its connection settings, like the persistent connection and its foreground worker.

The display power is cut with `DevicePolicyManager.lockNow()`. This requires the user to activate a new, minimal device admin that holds only the force-lock policy. Activation is a documented two-step flow: when the command is used with the app in the foreground, the system's activation screen is opened (routed through a small transparent activity because some manufacturers, like Samsung, refuse to open that screen from a new task); in the background a missing-permission notification is posted instead, and the command has to be sent again after activating. The command is refused when a secure keyguard is set — and the device admin activation is never requested in that case — since turning the screen off would otherwise lock the device behind credentials. No wake lock is held: background partial wake locks are an Android vitals liability and Doze can suspend networking regardless, so reachability stays the job of the existing connection settings.

## Checklist

User Documentation: home-assistant/companion.home-assistant#1374

## Any other notes
Tested on a Samsung SM-X110 (Android 16): screen turns off and back on via the commands, and with the persistent connection enabled, notifications keep arriving while the screen is off. Covered by unit tests for the helper (including the keyguard refusal and a `SecurityException` when the admin is deactivated mid-call), the admin receiver, the activation activity, and the command routing.


markfrancisonly:
requires device admin, but works inside google's boundaries

jpelgrom:
Hi, thanks for contributing. Did you see #6973?

> Tested on a Samsung SM-X110 (Android 17)

Is this correct? Based on what I can find there is no official Android 17 for any Samsung device yet. The Tab A9 (your device, from 2023) isn't even on lists of Samsung devices getting Android 17.

markfrancisonly:
Hello. I believe you are correct regarding android 17, and I'm unsure why Fable chose that descriptor, it had ADB access to the device, my SM-X110 devices are running 16 and are too old and slow for another major os revision. 

Unfortunately I agree with your thoughts about Device Admin, but I also am **_desperate_**. Only Android devices are available sub-$200 for multiple rooms wallpanel mounting. Since I have a wallpanel in my bedroom, screen dimming is not good enough, and running Fully is worse than running a forked android companion app. 

I would understand your decision either way to merge or close this PR. Home Assistant has undoubtedly countless android phone users who I would say should be the priority. Wallpanel power users are the only users who need this feature, and the need is not going away if you close a second or third similar PR. 

I have an earlier version of this feature that paints a black overlay instead of locking the display. The overlay implementation has the advantage of being quicker and keeping the camera/mic active, I use the tablets' camera in a [WHIP WebRTC stream](https://github.com/markfrancisonly/webcam-publisher/) for person detection, and screen lock cuts the camera. 

Screen off is missing, but I would be happy to modify this PR to only include the black overlay implementation? Since it's in my Claude memory, it would be trivial to upload it. Do you want it?

markfrancisonly:
btw "blacking" the screen could be seen as more of a platform policy violation than adding Device Admin permissions to turn off the screen. Samsung/google cut the cameras when the screen is locked and appears off by design.

app/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -131,6 +133,7 @@ class MessagingManager @Inject constructor(
     private val settingsDao: SettingsDao,
     private val textToSpeechClient: TextToSpeechClient,
     private val flashlightHelper: FlashlightHelper,
+    private val screenOffHelper: ScreenOffHelper,
```
TimoPtr, line 135: Could you extract the whole logic of ON/Off into the helper? So the whole wakeLock logic is in one place properly tested.

app/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
```diff
@@ -1801,6 +1817,13 @@ class MessagingManager @Inject constructor(
         context.startActivity(intent)
     }
 
+    private fun requestDeviceAdminPermission() {
+        // The activation screen only opens from an activity, see ScreenOffAdminRequestActivity
+        val intent = Intent(context, ScreenOffAdminRequestActivity::class.java)
+        intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
+        context.startActivity(intent)
```
TimoPtr: The pattern we use if for the activity to have a newInstance() companion method to get the proper Intent.

app/src/main/kotlin/io/homeassistant/companion/android/util/ScreenOffAdminRequestActivity.kt
```diff
@@ -0,0 +1,49 @@
...
+                ComponentName(this, ScreenOffAdminReceiver::class.java),
+            )
+            .putExtra(
+                DevicePolicyManager.EXTRA_ADD_EXPLANATION,
+                getString(commonR.string.screen_off_admin_description),
+            )
+        try {
+            requestAdmin.launch(intent)
+        } catch (e: ActivityNotFoundException) {
+            // Some devices, like Android Automotive, have no device admin settings
```
TimoPtr: Ending into this catch is last resort, the error should have been presented to the user previously or at least in a toast from this activity IMO. If it doesn't work with Automotive then it needs to be told to the user, if he uses the cmd.

app/src/test/kotlin/io/homeassistant/companion/android/util/ScreenOffAdminRequestActivityTest.kt
```diff
@@ -0,0 +1,48 @@
...
+
+import android.app.admin.DevicePolicyManager
+import android.content.ComponentName
+import android.os.Bundle
+import androidx.test.core.app.ActivityScenario
+import dagger.hilt.android.testing.HiltTestApplication
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
```
TimoPtr, line 11: Use the Jupiter API like in the rest of the project.

app/src/test/kotlin/io/homeassistant/companion/android/notifications/MessagingManagerScreenCommandsTest.kt
```diff
@@ -0,0 +1,115 @@
...
+import io.homeassistant.companion.android.common.data.servers.ServerManager
+import io.homeassistant.companion.android.common.notifications.NotificationData
+import io.homeassistant.companion.android.database.server.Server
+import io.homeassistant.companion.android.util.ScreenOffAdminRequestActivity
+import io.homeassistant.companion.android.util.ScreenOffHelper
+import io.mockk.coEvery
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.verify
+import org.junit.Assert.assertEquals
```
TimoPtr, line 21: Use Jupiter API

app/src/test/kotlin/io/homeassistant/companion/android/util/ScreenOffAdminReceiverTest.kt
```diff
@@ -0,0 +1,26 @@
...
+import io.mockk.mockk
+import io.mockk.verify
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
+@Config(application = HiltTestApplication::class)
+class ScreenOffAdminReceiverTest {
```
TimoPtr: Does it works without Robolectric by giving a mock context? If so please remove Robolectric and use JUnit Jupiter.

app/src/test/kotlin/io/homeassistant/companion/android/util/ScreenOffHelperTest.kt
```diff
@@ -0,0 +1,135 @@
...
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.Shadows.shadowOf
+import org.robolectric.annotation.Config
+import org.robolectric.shadows.ShadowPowerManager
+
+@RunWith(RobolectricTestRunner::class)
+@Config(application = HiltTestApplication::class)
+class ScreenOffHelperTest {
+
```
TimoPtr: Check if it can be written without Robolectric.

app/src/test/kotlin/io/homeassistant/companion/android/util/ScreenOffHelperTest.kt
```diff
@@ -0,0 +1,135 @@
...
+import android.content.Context
+import android.content.ContextWrapper
+import androidx.test.core.app.ApplicationProvider
+import dagger.hilt.android.testing.HiltTestApplication
+import io.mockk.every
+import io.mockk.mockk
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
```
TimoPtr: Jupiter API

automotive/src/main/AndroidManifest.xml
```diff
@@ -117,6 +117,27 @@
             </intent-filter>
         </receiver>
 
+        <!--Device admin holding only the force lock policy, used by the command_screen_off
```
TimoPtr: If it doesn't work on Automotive then let's not even add it to the manifest (both the receiver and activity)

TimoPtr (changes requested):
I would like @jpelgrom opinion too about using the admin API.

TimoPtr:
@jpelgrom what about gating this on minimal only? It's coming back quite often.

app/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
markfrancisonly: Done in be39869 — `ScreenOffHelper` now owns the whole flow: `turnScreenOn()` acquires the screen-on wake lock, releases the screen-off lock while it is held, and releases it again, so all wake lock handling lives in the helper and is covered by `ScreenOffHelperTest`. The command handler only routes.

app/src/main/kotlin/io/homeassistant/companion/android/notifications/MessagingManager.kt
markfrancisonly: Done in be39869 — the Intent now comes from `ScreenOffAdminRequestActivity.newInstance(context)`.

app/src/main/kotlin/io/homeassistant/companion/android/util/ScreenOffAdminRequestActivity.kt
markfrancisonly: Done in be39869 — the catch now shows a toast (`screen_off_unsupported`) before finishing. On Automotive the command handler tells the user directly with the same toast and never reaches this activity, since the admin components are no longer declared there.

app/src/test/kotlin/io/homeassistant/companion/android/util/ScreenOffAdminRequestActivityTest.kt
markfrancisonly: This one drives the real activity lifecycle through `ActivityScenario`/`Robolectric.buildActivity`, so it needs Robolectric, and Robolectric requires the JUnit 4 runner — matching the project guidance to use JUnit 4 only where Robolectric is required. Happy to convert if there is a Jupiter-compatible setup I missed.

app/src/test/kotlin/io/homeassistant/companion/android/notifications/MessagingManagerScreenCommandsTest.kt
markfrancisonly: Kept on Robolectric/JUnit 4: the test relies on idling the main looper, the started-activity capture, and the foreground-process shadow, which have no mock-context equivalent. The two test classes that don't need Robolectric (helper and receiver) are plain Jupiter in be39869.

app/src/test/kotlin/io/homeassistant/companion/android/util/ScreenOffAdminReceiverTest.kt
markfrancisonly: It does — converted to plain JUnit Jupiter with mocks in be39869.

app/src/test/kotlin/io/homeassistant/companion/android/util/ScreenOffHelperTest.kt
markfrancisonly: Done in be39869 — rewritten as a plain Jupiter test with a mocked context/services and stateful wake lock mocks, no Robolectric.

app/src/test/kotlin/io/homeassistant/companion/android/util/ScreenOffHelperTest.kt
markfrancisonly: Done in be39869 (same rewrite).

automotive/src/main/AndroidManifest.xml
markfrancisonly: Done in be39869 — receiver and activity removed from the Automotive manifest along with its changelog entry, and the command now tells the user it is not supported there.

markfrancisonly:
Thanks for the review! Addressed in be39869 (after merging `main`):
- `ScreenOffHelper` owns the whole wake lock flow for both commands; `MessagingManager` only routes
- `ScreenOffAdminRequestActivity.newInstance()`, plus a toast when the activation screen cannot be opened
- Automotive: the admin receiver/activity are no longer declared there, and the command shows a “not supported” toast
- `ScreenOffHelperTest` and `ScreenOffAdminReceiverTest` are plain JUnit Jupiter now; the activity and command-routing tests stay on Robolectric (JUnit 4) since they need the activity lifecycle and looper/activity shadows

On the admin API question — happy to hear @jpelgrom's take; the device admin force-lock policy is the only non-root way we found to actually power the display off while the device stays awake.

app/src/test/kotlin/io/homeassistant/companion/android/util/ScreenOffAdminRequestActivityTest.kt
TimoPtr: The guidance is for the runner not the assert API.

jpelgrom:
> I would like @jpelgrom opinion too about using the admin API.

> @jpelgrom what about gating this on minimal only? It's coming back quite often.

> On the admin API question — happy to hear @jpelgrom's take; the device admin force-lock policy is the only non-root way we found to actually power the display off while the device stays awake.

I don't think the request is that common - it just happened to show up twice in one month.

My main objection is with 1. the uncertainty it creates in store policies and 2. the risk of device admin if we do not lock it down properly. 1 would be mitigated if we limit it to minimal (I don't believe any of the stores we submit minimal to have strong policies on this), 2 would be harder to mitigate but the current iteration appears to be more locked down than earlier ones.

However, in the latest commits there is suddenly a new limitation where the command won't run if the device has a PIN etc. set which doesn't make sense to me. A lock is very common. If the user wants to turn off the screen on their device and they have a lock set that's something they should be aware of, not refused. We shouldn't promote insecure behavior by forcing no lock.

PR #7163, opened 2026-07-11
Navigate and reload the frontend in place for the webview command

### LINKED ISSUE #5381: command_webview should not relaunch application
I'm trying to display the front door camera feed whenever a person is detected on the stairs. This works well, but switching the view via command_webview relaunches the already running companion app and introduces about 5 - 10 seconds latency. After some days the app seems to crash completely and needs to be manually restarted. That did not happen before.

In my context it makes no sense to restart the already running HA companion app ... I just want to switch tabs. Not sure if there is any use case at all that requires multiple instances. If not, this could fix my problem:
```
android:launchMode="singleInstance"
```

alternatively what this guy suggested on stack overflow
```
In general you can use a combination of Intent.FLAG_ACTIVITY_SINGLE_TOP, Intent.FLAG_ACTIVITY_CLEAR_TOP, Intent.FLAG_ACTIVITY_REORDER_TO_FRONT and Intent.FLAG_ACTIVITY_NEW_TASK to accomplish what you want. However, which one (or ones) to use depends on the situation you have
```
https://stackoverflow.com/a/11555704

I tried the solution above, but was not able to test it, because the push notifications require firebase configuration which I did not manage to get running.

**Additional context**
```
alias: Türkamera auf Tablet EG anzeigen wenn Person(en) erkannt
description: ""
triggers:
  - trigger: numeric_state
    entity_id:
      - sensor.frontdoor_person_count
    above: 0
conditions: []
actions:
  - action: notify.mobile_app_tablet_eg
    metadata: {}
    data:
      message: command_screen_on
    alias: Wakeup Tablet
  - alias: Switch to Frontdoor view
    action: notify.mobile_app_tablet_eg
    metadata: {}
    data:
      message: command_webview
      data:
        command: lovelace/frontdoor
  - wait_for_trigger:
      - trigger: numeric_state
        entity_id:
          - sensor.frontdoor_person_count
        for:
          hours: 0
          minutes: 0
          seconds: 5
        below: 1
    timeout:
      hours: 0
      minutes: 1
      seconds: 0
      milliseconds: 0
  - alias: Switch to Frontdoor view
    action: notify.mobile_app_tablet_eg
    metadata: {}
    data:
      message: command_webview
      data:
        command: lovelace/default_view
mode: single
```

markfrancisonly:
## Summary

`command_webview` relaunched the whole app for every request, even when the frontend was already open and showing the targeted server. Each launch created a new task (`FLAG_ACTIVITY_MULTIPLE_TASK`), so repeated commands piled up activity instances, and the page state — camera streams, scroll position, dialogs — was lost every time.

The command is now Intent driven end to end: it always starts `LaunchActivity` with a `NavigateTo` (or the new `ReloadFrontend`) deep link, using `FLAG_ACTIVITY_CLEAR_TOP | FLAG_ACTIVITY_SINGLE_TOP` in the existing task. A running activity receives the deep link through `onNewIntent` and hands it to the frontend shown on top, which acts on it in place; otherwise the resolved destination is opened through the navigation graph. Deep links received while the UI is still starting are held — only the latest one — until the navigation graph is ready, and a running command follows the same Automotive policy as a launch, so the dedicated Automotive UI is never replaced by the WebView.

Acting in place means:
* a path is sent over the external bus using the `navigate` command (Home Assistant 2025.6+, version checked where it is used), normalized to root-relative first since the frontend resolves relative paths against the current page (the `lovelace/frontdoor` case of #5381),
* an `entityId:` target opens the more-info dialog (the prefix is matched ignoring case and surrounding spaces, since the value is typed by hand),
* a new `reload` value reloads the frontend without cached data, so a stale page cannot be re-served from the cache and the reload releases the resources the page holds, like camera streams and WebRTC connections,
* an absolute URL, or a server without navigation support, gets a full page load instead,
* a request still waiting for the frontend handshake is dropped when the user switches servers meanwhile, so a command can never land on the wrong server.

Servers older than 2025.6 and commands received while the app is closed keep the launch behavior, now without accumulating task instances.

Fixes #5381

## Checklist

User Documentation: home-assistant/companion.home-assistant#1375

## Any other notes
Verified on-device (Samsung SM-X110): dashboard-to-dashboard navigation, more-info for `entityId:` targets, and reload all act on the open frontend with no relaunch; commands received while the app is closed launch a single task. The legacy `WebViewActivity` was removed upstream while this PR was in review, so the implementation targets the frontend screen only. The `UrlUtil` case-insensitivity fix initially included here will be proposed separately.


markfrancisonly:
prior versions of this command accumulated in system resource exhaustion and additionally resulted in delayed nav. this version has good manners and only starts/restarts the ha companion app if closed. to compensate and complete the journey, there's a now reload command that performs a webview hard refresh  or loads the app fresh

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt
```diff
@@ -102,7 +102,7 @@ object UrlUtil {
     }
 
     fun isAbsoluteUrl(it: String?): Boolean {
-        return Regex("^https?://").containsMatchIn(it.toString())
+        return Regex("^https?://", RegexOption.IGNORE_CASE).containsMatchIn(it.toString())
```
TimoPtr: This could be in its own PR

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
```diff
@@ -147,6 +148,8 @@ internal fun FrontendScreen(
     val keepScreenOnEnabled by viewModel.keepScreenOnEnabled.collectAsStateWithLifecycle()
     val improvScanRequested by viewModel.improvScanRequested.collectAsStateWithLifecycle()
 
+    FrontendVisibleLifecycleEffect(viewModel::setFrontendVisible)
```
TimoPtr: Look at the pattern in this file we already have a Effect high level composable where to put such Effect.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -358,6 +369,35 @@ internal class FrontendViewModel @VisibleForTesting constructor(
             }
         }
 
+        viewModelScope.launch {
+            reloadRequestMediator.eventFlow.collect {
+                // Dropping the cache while the page is still loading can wedge the load
+                _webViewActions.emit(
+                    if (_viewState.value is FrontendViewState.Content) {
+                        WebViewAction.HardReload()
```
TimoPtr: Why doing a hard reload?

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -358,6 +369,35 @@ internal class FrontendViewModel @VisibleForTesting constructor(
...
+            // Bus messages sent before the frontend handshake are lost, so each request waits for
+            // it and only the latest request is kept while waiting
+            webViewNavigationMediator.navigationRequests.collectLatest { target ->
+                _viewState.first { it is FrontendViewState.Content }
+                when (target) {
+                    is FrontendTarget.EntityMoreInfo -> _webViewActions.emit(
+                        WebViewAction.OpenMoreInfo(target.entityId),
+                    )
+                    is FrontendTarget.Path -> externalBusRepository.send(NavigateToMessage(path = target.path))
+                    FrontendTarget.Default -> externalBusRepository.send(NavigateToMessage(path = "/"))
```
TimoPtr: why not using `navigateToDefaultDashboard`?

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -358,6 +369,35 @@ internal class FrontendViewModel @VisibleForTesting constructor(
...
+                        WebViewAction.HardReload()
+                    } else {
+                        WebViewAction.Reload()
+                    },
+                )
+            }
+        }
+
+        viewModelScope.launch {
+            // Requests are only made for servers supporting navigate, no version check is needed.
```
TimoPtr: You are assuming that the mediator is always going to gate the version, but it might change in the future.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -358,6 +369,35 @@ internal class FrontendViewModel @VisibleForTesting constructor(
...
+            // Bus messages sent before the frontend handshake are lost, so each request waits for
+            // it and only the latest request is kept while waiting
+            webViewNavigationMediator.navigationRequests.collectLatest { target ->
```
TimoPtr: I get your idea here, but I think to be more generic we should not rely on an in memory flow that notify this VM. Instead I would prefer to use a deep link and not restarting the LaunchActivity, and having the notification command open the deep link. This way it is more flexible, to do that we need to handle on the LaunchActivity new intent coming in, I didn't check how feasible it is.

app/src/main/kotlin/io/homeassistant/companion/android/util/WebViewNavigationMediator.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asSharedFlow
+import kotlinx.coroutines.flow.asStateFlow
+
+/**
+ * Carries webview navigation requests from the server commands to the visible frontend.
+ */
+@Singleton
+class WebViewNavigationMediator @Inject constructor() {
```
TimoPtr: Do you really need to know if the frontend is visible??

TimoPtr (changes requested):
I would prefer using Intent to also allow opening a link without reopening the LaunchActivity and it would drop the need of this two mediators. The WebViewActivity has been removed since this morning.

common/src/main/kotlin/io/homeassistant/companion/android/util/UrlUtil.kt
markfrancisonly: Dropped from this PR in b08e2b1; I'll propose it separately.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
markfrancisonly: Gone in b08e2b1 — the visibility tracking was only needed by the mediator, which is removed.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
markfrancisonly: The reload subcommand targets long-running dashboards (wall tablets) where the WebView ends up with stale cached assets after a server update, or holds wedged native resources like WebRTC camera streams. `clearCache(true)` + `reload()` guarantees a fresh page where a plain `reload()` can reuse the broken cache; a plain reload is still used while the page is loading (see the `HardReload` KDoc). Happy to switch to a plain reload if you prefer.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
markfrancisonly: Done in b08e2b1 — the default target now goes through `navigateToDefaultDashboard()`.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
markfrancisonly: Done in b08e2b1 — the ViewModel now checks `NavigateToMessage.isAvailable` itself and falls back to a full page load at the target for servers without navigation support.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
markfrancisonly: Implemented in b08e2b1 — the command always starts `LaunchActivity` with a `NavigateTo` (or the new `ReloadFrontend`) deep link using `CLEAR_TOP | SINGLE_TOP`, `LaunchActivity` forwards `onNewIntent` deep links, and the shown frontend's ViewModel applies them in place. Both mediators are gone, and it turned out feasible without restarting the activity.

app/src/main/kotlin/io/homeassistant/companion/android/util/WebViewNavigationMediator.kt
markfrancisonly: Not anymore — removed with the mediator in b08e2b1.

markfrancisonly:
Thanks — reworked to the Intent approach in b08e2b1 (after merging `main`, which also drops the legacy `WebViewActivity` changes):
- `command_webview` always builds a `LaunchActivity` deep link (`NavigateTo`, or the new `ReloadFrontend`) with `CLEAR_TOP | SINGLE_TOP`
- `LaunchActivity` now handles `onNewIntent`: the deep link is handed to the visible frontend's ViewModel to act in place, otherwise it navigates to the frontend route — no more mediators or visibility tracking
- version gating and the default-dashboard path live in the ViewModel
- the `UrlUtil` fix will move to its own PR

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
markfrancisonly: Simplified in 58e0465 — a reload() already destroys the current document and releases its native resources (camera/WebRTC), so the application-wide cache clear was not needed. HardReload is removed and the companion docs wording is updated accordingly.

markfrancisonly:
Review follow-ups in 58e0465 and 90f3a04: in-place targets are now classified — relative paths are normalized to root-relative before the frontend navigation (the `lovelace/frontdoor` case from #5381) and absolute URLs take a full page load instead of the bus; a navigation waiting for the frontend handshake can no longer land on the wrong server (a server switch cancels it and the shown server is revalidated once the page is ready); and the reload keeps its cache-dropping behavior — rationale in the inline thread. Ready for another look @TimoPtr.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
markfrancisonly: Final decision after weighing the alternatives: the cache-dropping reload stays (restored in 90f3a04, superseding my earlier "plain reload" reply). The reload subcommand exists to recover pages serving stale cached assets, so a reload that honors the HTTP cache can re-serve exactly what it is meant to replace.

On the application-wide scope of `clearCache`: the app has effectively a single WebView, so the cleared cache is precisely the frontend cache the reload refreshes anyway. A cache-mode bypass scoped to the load (`LOAD_NO_CACHE` + restore once the page finishes) was prototyped and rejected — it needs a set/restore pair coupled across an asynchronous page load plus a pending flag to maintain, for no practical difference in this app.

TimoPtr:
> the `UrlUtil` fix will move to its own PR

You mentioned that UrlUtil will move to its own PR, I don't see any. Is that expected?

jpelgrom:
@dshokouhi You might remember this, but wasn't opening a new activity and not replacing the history of any open 'main' app here intentional?

dshokouhi:
> @dshokouhi You might remember this, but wasn't opening a new activity and not replacing the history of any open 'main' app here intentional?

Yes this was deliberately done to handle cases of re-navigation as eventually if you keep going back and forth navigation just stops completely and the last command is just repeated. It was easy to test goin from like from Tab A > Tab B > Tab C and keep going back and forth and it eventually just stops and no matter what command you send its stuck on Tab A. Creating a new task was the only reliable method for consistent navigation using only the webview commands. A lot of users use this command for devices that cant be reached and thus want to automate opening various tabs back to back. There should be a history of issues and PRs on the topic in git.

TimoPtr:
> > @dshokouhi You might remember this, but wasn't opening a new activity and not replacing the history of any open 'main' app here intentional?
> 
> Yes this was deliberately done to handle cases of re-navigation as eventually if you keep going back and forth navigation just stops completely and the last command is just repeated. It was easy to test goin from like from Tab A > Tab B > Tab C and keep going back and forth and it eventually just stops and no matter what command you send its stuck on Tab A. Creating a new task was the only reliable method for consistent navigation using only the webview commands. A lot of users use this command for devices that cant be reached and thus want to automate opening various tabs back to back. There should be a history of issues and PRs on the topic in git.

This was because we were reaching the max size of the backstack? I think we might then have a flag to offer one or the other, some ppl might be interested to keep the current stack and not a new activity?

dshokouhi:
> > > @dshokouhi You might remember this, but wasn't opening a new activity and not replacing the history of any open 'main' app here intentional?
> > 
> > Yes this was deliberately done to handle cases of re-navigation as eventually if you keep going back and forth navigation just stops completely and the last command is just repeated. It was easy to test goin from like from Tab A > Tab B > Tab C and keep going back and forth and it eventually just stops and no matter what command you send its stuck on Tab A. Creating a new task was the only reliable method for consistent navigation using only the webview commands. A lot of users use this command for devices that cant be reached and thus want to automate opening various tabs back to back. There should be a history of issues and PRs on the topic in git.
> 
> This was because we were reaching the max size of the backstack? I think we might then have a flag to offer one or the other, some ppl might be interested to keep the current stack and not a new activity?

Not a backstack when using command webview it's all forward navigation that gets stuck. Or at least got stuck. Was easy to replicate after 5-10 minutes of sending commands to various tabs. Not navigating the left navigation but the top row of tabs in a dashboard.

TimoPtr:
> > > > @dshokouhi You might remember this, but wasn't opening a new activity and not replacing the history of any open 'main' app here intentional?
> > > 
> > > 
> > > Yes this was deliberately done to handle cases of re-navigation as eventually if you keep going back and forth navigation just stops completely and the last command is just repeated. It was easy to test goin from like from Tab A > Tab B > Tab C and keep going back and forth and it eventually just stops and no matter what command you send its stuck on Tab A. Creating a new task was the only reliable method for consistent navigation using only the webview commands. A lot of users use this command for devices that cant be reached and thus want to automate opening various tabs back to back. There should be a history of issues and PRs on the topic in git.
> > 
> > 
> > This was because we were reaching the max size of the backstack? I think we might then have a flag to offer one or the other, some ppl might be interested to keep the current stack and not a new activity?
> 
> Not a backstack when using command webview it's all forward navigation that gets stuck. Or at least got stuck. Was easy to replicate after 5-10 minutes of sending commands to various tabs. Not navigating the left navigation but the top row of tabs in a dashboard.

Back then did you know why it was getting stuck?

dshokouhi:
> > > This was because we were reaching the max size of the backstack? I think we might then have a flag to offer one or the other, some ppl might be interested to keep the current stack and not a new activity?
> > 
> > 
> > Not a backstack when using command webview it's all forward navigation that gets stuck. Or at least got stuck. Was easy to replicate after 5-10 minutes of sending commands to various tabs. Not navigating the left navigation but the top row of tabs in a dashboard.
> 
> Back then did you know why it was getting stuck?

no but it did feel like a caching issue where command were being received but not respected.

PR #7164, opened 2026-07-11
Detect silently dropped WebSocket connections and retry until subscriptions are restored

### LINKED ISSUE #5259: Gracefully handle reconnection failure on WebSocket
_Originally posted by @TimoPtr in https://github.com/home-assistant/android/pull/5219#discussion_r2044438484_

Currently if the connection fail we try to reconnect and re-subscribe to the entity that were previously registered. But if it fails to reconnect then nothing will happen and even if the connection is re-open later it won't resubscribe.

### LINKED ISSUE #5939: Persistent connection is not persistent
2025.9.2-minimal

**Android version(s):**
13

**Device model(s):**
Lenovo M9

**Home Assistant version:**
2025.10.2

unknown

Lenovo M9 tablet set up as a wall panel, persistent connection is set to "Always". After fresh start of the app, this works flawlessly, I can receive notifications (via notify.mydevice) immediately.

However, the connection is lost after some time. The persistent connection notification is still active, no error message is visible. Nevertheless, notifications to the device then start failing silently. Manually pushing a notify action then results in a "device not connected to local push notifications" error message.

This behavior unfortunately will consistently appear after betwenn 1 and 12 hours of running the companion app. Force close solves the issue temporarily.

**Companion App Logs:**

[homeassistant_companion_log_9-15-2025_0-55-27_125548.txt](https://github.com/user-attachments/files/22920501/homeassistant_companion_log_9-15-2025_0-55-27_125548.txt)
Log file was saved after loss of connection.

n/a

**Additional Information**
Battery optimization for the companion app is turned off

### LINKED ISSUE #6926: Local Notifications sometimes failing (device I'd, webhook)

**Android version(s):**
2026.4.4 minimal
**Device model(s):**
Pixel 10
**Home Assistant version:**
2026.5.4

I have had no issued with notifications before switching to android. Now it sometimes says my device can't recieve local notifications remedy the issue is by force stopping the app and restarting it. However it still comes back in an interval I cannot determine. 

**Companion App Logs:**

```

```

**Additional information:**
Device(s) with webhook id(s) DEVICE_ID not connected to local push notifications

markfrancisonly:
## Summary

When the WebSocket dropped and the single reconnection attempt failed — typically because Home Assistant was restarting — active subscriptions were never restored even after the server came back. For devices relying on local push, notifications stayed dead until the app was manually restarted, with `notify` failing as `not connected to local push notifications`. Connections that die silently (a container restart where the address simply disappears, an access point roam, a NAT timeout) were never detected at all: OkHttp keeps buffering writes into the half-open socket, no close callback ever fires, and the periodic ping result was discarded.

* The reconnection is now retried while there are active subscriptions to restore, stopping on authentication failure since retrying would not help. This replaces the TODO referencing this issue. Attempts start 10 seconds apart and back off up to a 2 minute cap; the backoff resets when a subscription is actually restored, so persistent failures — an unreachable server or a subscription the server keeps rejecting — keep retrying indefinitely without hammering, and are never silently abandoned while their flow is still collected.
* A subscription's old registration is only dropped once the new one is acknowledged. A resubscription the server rejects is retried on the live connection with the capped backoff. One that receives no acknowledgement leaves the server state ambiguous — it may have been accepted with the answer lost — so the connection is cancelled and restoration resumes on a clean connection, where no possibly-accepted attempt can linger as a duplicate registration. Closing a subscription also unsubscribes any attempt still in flight for it.
* The application-level ping now cancels the connection when no pong arrives on a connection that is believed to be established, surfacing silently dropped connections to the same close handling and restore path. A connection (re)established while waiting for the pong is never cancelled.
* The `WebsocketManager` worker keeps supervising while the settings require it and restarts collection jobs that stopped (now covered by a test), covering the case where the app starts while the server is restarting and the very first subscribe fails.

Complementary to — but fully independent of — home-assistant/core#176833, the server-side half of the local push story; see the comment below for how the two fit together.

Fixes #5259
Fixes #5939
Likely also resolves #6926

## Checklist

## Any other notes
Added regression tests for the previously uncovered cases: the server staying unavailable through several retries and then recovering (the original Flow receives events under the new subscription ID), an unacknowledged resubscription being restored on the following reconnection, a resubscription rejected on a still-live connection being retried until it succeeds, an unanswered ping cancelling the connection and its subscription being restored, the backoff schedule while the server stays unreachable, rejection retries backing off without ever dropping the subscription, and the worker restarting a collection job that stopped. Verified on-device: with the server stopped, the app retries continuously and notifications resume within seconds of the server coming back; a silently dropped connection is detected within one ping interval.


markfrancisonly:
this pr resolves issues that were preventing notification commands from being received reliability. good stuff, enjoy

app/src/main/res/xml/changelog_master.xml
```diff
@@ -3,14 +3,17 @@
     tools:ignore="MissingDefaultResource">
     <release version="2026.7.2 - Main" versioncode="3">
         <change>Android 17: added assistant volume level sensor + notification command for control</change>
+        <change>Local push notifications now detect silently dropped connections and keep retrying to reconnect while the server is unreachable, like during a server restart, instead of staying disconnected</change>
```
TimoPtr: Let's remove this from the changelog we don't describe bug fixes in there.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -765,16 +792,18 @@ internal class WebSocketCoreImpl(
...
-                            ?.let {
-                                val subscription = it.key
-                                Timber.d("Unsubscribing from $subscribeMessage")
-                                // Unsubscribe must happen before removing from activeMessages to ensure
-                                // the server acknowledges before we stop handling events for this subscription
-                                unsubscribeEvents(subscription)
-                                channel.close()
-                                activeMessages.remove(subscription)
-                            }
+                        // Resubscribing briefly tracks two entries for the same message, remove
```
TimoPtr: I see why the old registration has to be kept until the new one is acknowledged losing it on an unacked resubscribe is exactly the bug you're fixing. What bothers me is that this invariant lives as two Subscription entries in activeMessages for one logical subscription, which is why awaitClose now needs a loop to sweep them all (easy to miss). Could we make the state explicit in the type instead? For example a sealed variant like ActiveMessage.Subscription.Reconnecting(pendingId = …): mark the old entry as reconnecting when the attempt starts, send the resubscribe as a regular answerable message rather than a Subscription command (so processSendCommand doesn't track a second subscription), and on ack move the entry to the new id / back to the normal state. The map would keep a single entry per subscription, the "remove them all" loop in awaitClose could go back to the simple let, and the reconnecting window would be visible in the types instead of implied by duplicate entries.


It's just an idea that needs to be validated, think about it.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -1022,52 +1051,82 @@ internal class WebSocketCoreImpl(
...
-
-            val shouldAttemptReconnect = hasSubscriptions && wasActive
-
-            if (shouldAttemptReconnect && wsScope.isActive) {
-                // Delay before reconnect unless URL changed
-                if (closeReason != WebSocketState.Closed.Reason.CHANGED_URL) {
-                    delay(DELAY_BEFORE_RECONNECT)
+                val hasSubscriptions = activeMessages.any { it.value is ActiveMessage.Subscription }
+
+                if (hasSubscriptions && wasActive && wsScope.isActive) {
```
TimoPtr: Keep the variable `val shouldAttemptReconnect = hasSubscriptions && wasActive` it makes the condition in the if easier to read.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -1022,52 +1051,82 @@ internal class WebSocketCoreImpl(
         if (connectionHolder.get() == null) return
 
         wsScope.launch {
-            val hasSubscriptions: Boolean
-
             connectedMutex.withLock {
```
TimoPtr, line 1077: I understand that you want the mutex to be hold while starting the Job and also storing the Job ref in the attributes. I would like that the new class attribute be annotated like the others `@GuardedBy("connectedMutex")`, even if it is useless at the moment it is still providing information.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -1022,52 +1051,82 @@ internal class WebSocketCoreImpl(
...
-                    }
+        var toRestore: Set<Long> = activeMessages.filterValues { it is ActiveMessage.Subscription }.keys
+        var retryDelay = DELAY_BEFORE_RECONNECT
+        while (toRestore.isNotEmpty()) {
+            if (connect()) {
+                // The server is reachable again, restart the backoff
+                retryDelay = DELAY_BEFORE_RECONNECT
+                toRestore = resubscribeActiveSubscriptions(toRestore)
+                if (toRestore.isEmpty()) return
+                Timber.w("${toRestore.size} subscriptions not restored, retrying in $retryDelay")
```
TimoPtr: If for some reason the server reject the subscription you end up into an infinite loop of trying and wait for 10s.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -1022,52 +1051,82 @@ internal class WebSocketCoreImpl(
         if (connectionHolder.get() == null) return
 
         wsScope.launch {
-            val hasSubscriptions: Boolean
-
             connectedMutex.withLock {
-                val holder = connectionHolder.getAndSet(null)
+                // Another callback of the same closing already handled it
+                val holder = connectionHolder.getAndSet(null) ?: return@launch
```
TimoPtr, line 1097: So if the holder is null we don't try to reconnect anymore? it might be ok but it needs some thinking to be sure of the behavior.

TimoPtr (changes requested):
Shutdown needs to clear the reconnection and close too if I'm not wrong.

I like the idea of the ping you had it's a nice safe guard.

GimliTheGreat:
I hope you guys work it out, I would love to stop rebooting my Amazon Echos every time I restart HA 😀

markfrancisonly:
**Relationship to home-assistant/core#176833** (mobile_app: keep the push channel registered, degrade routing on repeated confirm timeouts)

The two PRs are complementary halves of the same user-visible failure — `not connected to local push notifications` — and are fully independent: neither depends on the other and they can land in any order.

| Failure | Who can detect it | Fixed by |
|---|---|---|
| Socket **alive**, but the server tore down the push channel after one late notification confirmation | Only the server — the app-level ping added here gets its pong (the socket is fine), so no client change can see this | core#176833: the channel stays registered; repeated timeouts degrade routing to cloud with a periodic local probe instead of tearing down |
| Socket **dead** (server restart, NAT timeout, network roam) with the old server process gone | Only the client — no server code can inform a client holding a socket to an endpoint that no longer exists | This PR: the unanswered ping cancels the connection and the restore loop retries with backoff until subscriptions are re-established |

They also compose cleanly: when the restore loop here re-subscribes, the server replaces its push channel object, which resets the degraded/breaker state from core#176833 — a reconnect always starts with a clean local push path.

markfrancisonly:
CI note: the Emulator.wtf APK build hit `D8: java.lang.OutOfMemoryError: Java heap space` while merging external-dependency dex archives. Compilation passed and this branch does not change any dependencies, so this looks like a runner flake — closing/reopening to retrigger.

app/src/main/res/xml/changelog_master.xml
markfrancisonly: Done in 3a4e4b3 — the changelog entry is dropped.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
markfrancisonly: Implemented in 3a4e4b3 along these lines: re-subscription attempts are now an explicit `ActiveMessage.Reconnecting` type and the resubscribe is no longer tracked as a second `Subscription`, so `findSubscription` stays unambiguous and the awaitClose sweep of duplicate entries is gone.

One delta from your sketch: the attempt is tracked as a temporary `Reconnecting` entry keyed by the id the server will use (pointing at the original, which stays under its old id until promotion), rather than marking the original with a `pendingId`. Keying by the new id lets both the acknowledgement and any events that arrive for the new id before the ack is processed go through the existing map lookups — early events are forwarded to the original subscription's flow. On ack the original moves to the new id; on connection loss the attempt entries are simply dropped and the originals are restored by the restore loop.

awaitClose is back to the simple `let`; the one loop left there is intentional and different from the old sweep: closing a subscription also unsubscribes the pending id of any attempt still in flight for it, so a late acknowledgement cannot resurrect a closed subscription.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
markfrancisonly: Kept in 3a4e4b3 — `val shouldAttemptReconnect = hasSubscriptions && wasActive`.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
markfrancisonly: Done in 3a4e4b3 — `reconnectJob` is annotated with `@GuardedBy("connectedMutex")`.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
markfrancisonly: Reworked in 3a4e4b3: retries now back off exponentially (10 s doubling up to a 2 min cap, resetting when progress is made) instead of hammering every 10 s. It still retries indefinitely on purpose: dropping the subscription would leave its collector waiting forever on a flow that can never emit again, silently — retrying keeps it eventually consistent once the server accepts again, which is what lets local push heal (together with home-assistant/core#176833 on the server side). Subscriptions closed while waiting are removed from the retry set, so the loop cannot outlive its consumers.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
markfrancisonly: That early return only deduplicates callbacks of the same closing: OkHttp can deliver more than one terminal callback (e.g. `onFailure` after `onClosing`), and the first one already took the holder, ran the cleanup and made the reconnect decision — a second pass must not start a competing restore loop. During connection establishment the holder is not set yet and connect() owns the lifecycle, so nothing is lost there either. An intentional shutdown meanwhile cancels the restore loop in close() under the same mutex (3a4e4b3).

markfrancisonly:
Thanks for the review! Everything is addressed in 3a4e4b3 (today's push — the review predates it):
- shutdown: `close()` now cancels the restore loop under `connectedMutex` before closing the socket
- the changelog entry is dropped
- `shouldAttemptReconnect` stays as a named variable, and `reconnectJob` carries `@GuardedBy("connectedMutex")`
- rejected subscriptions retry with capped exponential backoff (10 s → 2 min) instead of a fixed 10 s loop
- and the bigger one: your explicit-type idea is implemented — re-subscription attempts are a dedicated `ActiveMessage.Reconnecting` tracked under the pending id, the duplicate `Subscription` entries and the awaitClose sweep are gone (details in the inline replies)

Glad the ping safeguard reads well!

markfrancisonly:
The duplicate-close race is fixed in 3bb983f: the state snapshot, transition, close-reason consumption and holder removal now happen in one `connectedMutex` critical section, so nothing is captured at callback time and only the coroutine winning the holder decides on reconnection. Includes a regression test delivering both terminal callbacks before cleanup runs, asserting exactly one restore loop starts. Ready for another look @TimoPtr.

markfrancisonly:
Delivering on the Home Assistant promise to be local-first, this PR and the corresponding core home-assistant/core#176833 solve a critical android local notification reliability issue.

Without a solid notification communication channel, no android wallpanel/kiosk notification command works reliably. My other PRs #7162 and #7163 are dependent on this PR.

I'm out of time to iterate further — maintainers, please feel free to push any refactoring or polish directly to the branch. I'll be production testing tonight on my SM-X110s and will report back. Hope you enjoy these changes!

TimoPtr:
Hi @markfrancisonly I invite you to read our recently published AI policy https://github.com/home-assistant/android/blob/main/AI_POLICY.md from this point I expect to speak to a human only, you can use LLMs to shape your answer but let's keep it between humans.

app/src/test/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
```diff
@@ -213,4 +219,36 @@ class WebsocketManagerTest {
         verify(exactly = 0) { context.getSystemService(PowerManager::class.java) }
         coVerify(exactly = 0) { entryPoint.serverManager.connectionStateProvider(any()) }
     }
+
+    @Test
+    fun `Given a collector that stopped when supervising then it is restarted on the next interval`() = runTest {
+        mockSetting(WebsocketSetting.ALWAYS)
```
TimoPtr, line 232: I did miss it while writing this file in the first place, but we should do a `unmockkAll` in the TearDown of this test class.

app/src/main/kotlin/io/homeassistant/companion/android/websocket/WebsocketManager.kt
```diff
@@ -116,7 +117,10 @@ class WebsocketManager(appContext: Context, workerParams: WorkerParameters) :
         fun checkLocalNetworkPermission(): CheckLocalNetworkPermissionUseCase
     }
 
-    override suspend fun doWork(): Result = withContext(Dispatchers.IO) {
+    @VisibleForTesting
+    internal var dispatcher: CoroutineDispatcher = Dispatchers.IO
```
TimoPtr: Move this above the EntryPoint definition.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -997,77 +1053,172 @@ internal class WebSocketCoreImpl(
...
+                if (unrestored.size < toRestore.size) {
+                    // Progress was made, only the persistent failures keep backing off
+                    retryDelay = DELAY_BEFORE_RECONNECT
                 }
-                reconnectSubscriptions()
+                toRestore = unrestored
+                Timber.w("${toRestore.size} subscriptions not restored, retrying in $retryDelay")
+            } else {
+                if (getConnectionState() == WebSocketState.ClosedAuth) {
+                    Timber.e("Authentication failed, not retrying to resubscribe to active subscriptions")
```
TimoPtr: When this happens the subscriptions stay in activeMessages but are never resubscribed. Even if auth is fixed later (a new token for instance), the next successful connection comes up with zero registrations on the server side and nothing restores them until the next socket closing.

The more I look at this, the more I think the resubscription trigger belongs at the end of connect() rather than in the closing path: when a fresh connection is established and activeMessages still contains Subscription entries from a previous connection, start the restore job for them. Semantically it reads as "connect, then re-register anything left over from the previous connection". A fresh socket has no registrations by definition, so any tracked subscription at that point needs restoring. This matters especially because any sendMessage can theoretically reconnect the WebSocket (the worker's ping does this every 30s) and leave the subscriptions tracked but not subscribed.

Note: it would need to launch the job (guarded by reconnectJob like today) rather than resubscribe inline: inline it would deadlock on connectedMutex and make unrelated sendMessage callers wait on the restoration. The closing-path loop would still be needed to actively retry while the server is unreachable; the connect() trigger would be the safety net for every other way a connection comes back.

This is an idea to explore.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
```diff
@@ -1445,6 +1446,540 @@ misc
...
+            webSocketListener.onMessage(
+                mockConnection,
+                """{"id":$newId, "type":"event", "event":{"event_type":"state_changed", "time_fired":"2016-11-26T01:37:24.265429+00:00", "data": {"entity_id":"light.bed_light"}}}""",
+            )
+            assertEquals("light.bed_light", awaitItem().entityId)
+        }
+    }
+
+    @Test
+    fun `Given a resubscription rejected on a live connection Then it is retried until restored`() = runTest {
```
TimoPtr: Reformulate to add `When`

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
```diff
@@ -1445,6 +1446,540 @@ misc
...
+            },
+            "No orphan entry should remain after the flow is closed",
+        )
+        verify(atLeast = 2) {
+            mockConnection.send(match<String> { it.contains(""""type":"unsubscribe_events"""") })
+        }
+    }
+
+    @Test
+    fun `Given a subscription the server keeps rejecting Then it is retried with backoff and never abandoned`() = runTest {
```
TimoPtr: Reformulate to add `When`

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -802,6 +844,12 @@ internal class WebSocketCoreImpl(
         }
     }
 
+    /**
+     * Finds the single [ActiveMessage.Subscription] entry matching [subscribeMessage].
+     *
+     * Resubscription attempts are tracked as [ActiveMessage.Reconnecting] and never match, so one
+     * logical subscription always has exactly one matching entry.
+     */
```
TimoPtr, line 857: Nitpick: based on this doc I would change the return value to be <Long, ActiveMessage.Subscription>?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -839,11 +890,16 @@ internal class WebSocketCoreImpl(
...
                 // Null is acceptable because a message could still arrive after unsubscribe like run-end in Assist pipeline
                 "Event should always be associated to a ActiveMessage.Subscription message"
             }
-            it as? ActiveMessage.Subscription
+            when (it) {
+                // An event can arrive for the new id before the resubscription acknowledgement
+                // has been processed, deliver it to the original subscription's flow
+                is ActiveMessage.Reconnecting -> it.original
+                else -> it as? ActiveMessage.Subscription
+            }
```
TimoPtr: (Suggestion is broken in git but here my proposal to simplify the logic and make it more future proof)

```
        val activeMessage = when (val message = activeMessages[subscriptionId]) {
            // An event can arrive for the new id before the resubscription acknowledgement                                                                                                                                                                                                                              
            // has been processed, deliver it to the original subscription's flow                                                                                                                                                                                                                                        
            is ActiveMessage.Reconnecting -> message.original
            is ActiveMessage.Subscription -> message
            is ActiveMessage.Simple -> {
                FailFast.fail { "Event ${response.id} associated to a Simple message" }
                null
            }
            // A message can still arrive after unsubscribe, like run-end in Assist pipeline                                                                                                                                                                                                                             
            null -> null
        }
```
Move the condition out of the FailFast so that a new type it caught by the compiler and force us to decide what to do with it.

TimoPtr: Also this could be its own small function to make the whole function a bit easier to read.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -997,77 +1053,172 @@ internal class WebSocketCoreImpl(
...
-            -> {
-                connectionState = WebSocketState.Closed(closeReason)
+        // Resubscription attempts in flight can no longer be acknowledged on this connection,
+        // drop them; the original subscriptions they point to stay tracked and are restored
+        // by the restore loop
+        activeMessages
+            .filterValues { it is ActiveMessage.Reconnecting }
+            .forEach { (key, activeMessage) ->
+                activeMessage.responseDeferred.completeExceptionally(IOException("Connection closed"))
+                activeMessages.remove(key)
```
TimoPtr: Any reason to not mutualize the logic with the is Simple above by simply updating the filter condition?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -997,77 +1053,172 @@ internal class WebSocketCoreImpl(
...
-                val holder = connectionHolder.getAndSet(null)
+                // Snapshot and transition under the mutex: a duplicate terminal callback of the
+                // same closing serializes behind the first one, sees the Closed state and no
+                // holder, and can neither suppress nor repeat the reconnect decision
+                val previousState = connectionState
+                val closeReason = pendingCloseReason ?: WebSocketState.Closed.Reason.OTHER
+                val wasActive = previousState == WebSocketState.Active
+                pendingCloseReason = null
+
+                // Transition to closed state unless already closed
```
TimoPtr:
```suggestion
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -997,77 +1053,172 @@ internal class WebSocketCoreImpl(
...
 
+    private fun handleClosingSocket() {
         wsScope.launch {
-            val hasSubscriptions: Boolean
-
             connectedMutex.withLock {
-                val holder = connectionHolder.getAndSet(null)
+                // Snapshot and transition under the mutex: a duplicate terminal callback of the
+                // same closing serializes behind the first one, sees the Closed state and no
+                // holder, and can neither suppress nor repeat the reconnect decision
```
TimoPtr: I'm not sure this comments brings a lot of value here. @jpelgrom wdyt?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -1179,6 +1347,21 @@ private sealed interface Command<T> {
...
+         * messages map keeps a single [ActiveMessage.Subscription] entry per logical subscription
+         * while the attempt is in flight.
+         */
+        class Resubscription(request: WebSocketRequest, val original: ActiveMessage.Subscription) :
+            WithAnswer(request) {
+
+            override fun toActiveMessage(): ActiveMessage {
+                return ActiveMessage.Reconnecting(responseDeferred, original)
+            }
+        }
```
TimoPtr:
```suggestion
        class Resubscription(val original: ActiveMessage.Subscription) :
            WithAnswer(original.request) {

            override fun toActiveMessage(): ActiveMessage {
                return ActiveMessage.Reconnecting(responseDeferred, original)
            }
        }
```
Let's avoid mistake by only taking the original as parameter.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -997,77 +1053,172 @@ internal class WebSocketCoreImpl(
...
+        while (remaining.isNotEmpty()) {
+            val oldId = remaining.removeAt(0)
+            val original = activeMessages[oldId] as? ActiveMessage.Subscription ?: continue
+
+            // Capture the connection this attempt is sent on, so a connection (re)established
+            // while waiting for the answer is never cancelled by mistake.
+            val attemptHolder = connectionHolder.get()
+            val response = sendMessage(
+                Command.WithAnswer.Resubscription(request = original.request, original = original),
+            )
```
TimoPtr:
```suggestion
            val response = sendMessage(Command.WithAnswer.Resubscription(original))
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -997,77 +1053,172 @@ internal class WebSocketCoreImpl(
...
+                    if (attemptHolder != null && connectionHolder.get() === attemptHolder) {
+                        attemptHolder.webSocket.cancel()
                     }
+                    return failed
                 }
-        } else {
-            // TODO https://github.com/home-assistant/android/issues/5259 handle re-connection gracefully or terminates the flows
-            Timber.w("Unable to reconnect, cannot resubscribe to active subscriptions")
+
+                response.success != true -> {
```
TimoPtr, line 1233: With this it means that an always failing request for instance if a template fail it will retry with the backoff all the time until it is addressed.

TimoPtr (commented):
Thanks for pushing this through. It does cover edge cases and should improve the stability of the app. I have few requests still to harden it even more. Then I would like @jpelgrom to also look at it since it's a quite important/fragile part of the app.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
jpelgrom: No, we already have comments in this function every couple of lines and mutexes avoiding two calls racing each other is basics we don't need in a comment.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
jpelgrom: > The more I look at this, the more I think the resubscription trigger belongs at the end of connect() rather than in the closing path: when a fresh connection is established and activeMessages still contains Subscription entries from a previous connection, start the restore job for them. Semantically it reads as "connect, then re-register anything left over from the previous connection".

This makes logical sense. However, one of the reasons for putting it in the closing was to ensure a delay before retrying. Most connection failures are temporary due to a network change - immediately retrying will fail but a couple of seconds later there is a network again so it will work. The delay before attempting to setup a new connection should stay.

TimoPtr: I agree, I want to move only the logic of opening again the pending subscription out of this and directly in the connect path. I'm going to propose something within the PR

TimoPtr:
@markfrancisonly I did push to your PR to address my comments and made the logic changes. It would be nice if you have some time to look at my changes (even if it's merged I'll address the comments).

TimoPtr (approved):

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -182,6 +188,13 @@ internal class WebSocketCoreImpl(
     @Volatile
     private var pendingCloseReason: WebSocketState.Closed.Reason? = null
 
+    /**
+     * The running job working to get the tracked subscriptions restored, replaced on each
+     * closing and each establishment so only one runs at a time.
```
```suggestion
     * closing and each establishment of the connecton so only one runs at a time.
```

app/src/test/kotlin/io/homeassistant/companion/android/websocket/WebsocketManagerTest.kt
```diff
@@ -213,4 +226,36 @@ class WebsocketManagerTest {
         verify(exactly = 0) { context.getSystemService(PowerManager::class.java) }
         coVerify(exactly = 0) { entryPoint.serverManager.connectionStateProvider(any()) }
     }
+
+    @Test
+    fun `Given a collector that stopped when supervising then it is restarted on the next interval`() = runTest {
```
jpelgrom: Align terminology with worker

```suggestion
    fun `Given a collector that stopped when listening for notifications then it is restarted on the next interval`() = runTest {
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -795,17 +842,26 @@ internal class WebSocketCoreImpl(
...
             return flow
         }
     }
 
-    private fun findSubscription(subscribeMessage: Map<String, Any?>): Map.Entry<Long, ActiveMessage>? {
-        return activeMessages.entries.firstOrNull {
-            (it.value as? ActiveMessage.Subscription)?.request?.message ==
-                subscribeMessage
+    /**
+     * Finds the single [ActiveMessage.Subscription] entry matching [subscribeMessage].
```
```suggestion
     * Finds the single [ActiveMessage.Subscription] entry and its ID matching [subscribeMessage].
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -1000,74 +1073,192 @@ internal class WebSocketCoreImpl(
...
+        // Subscriptions closed since the snapshot was taken no longer need to be restored
+        var toRestore: Set<Long> =
+            subscriptionsToRestore.filterTo(mutableSetOf()) { activeMessages[it] is ActiveMessage.Subscription }
+        var retryDelay = DELAY_BEFORE_RECONNECT
+        while (toRestore.isNotEmpty()) {
+            if (getConnectionState() != WebSocketState.Active) return
+            val unrestored = resubscribeActiveSubscriptions(toRestore)
+            if (unrestored.isEmpty()) return
+            if (unrestored.size < toRestore.size) {
+                // Progress was made, only the persistent failures keep backing off
```
```suggestion
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -1000,74 +1073,192 @@ internal class WebSocketCoreImpl(
...
+                retryDelay = DELAY_BEFORE_RECONNECT
             }
+            toRestore = unrestored
+            Timber.w("${toRestore.size} subscriptions not restored, retrying in $retryDelay")
+            // Back off so a rejecting server is not hammered; a subscription is never abandoned
+            // while its collector is still listening
```
```suggestion
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -1000,74 +1073,192 @@ internal class WebSocketCoreImpl(
...
                     }
+                    return failed
                 }
-        } else {
-            // TODO https://github.com/home-assistant/android/issues/5259 handle re-connection gracefully or terminates the flows
-            Timber.w("Unable to reconnect, cannot resubscribe to active subscriptions")
+
+                response.success != true -> {
+                    // The server is reachable but actively refused the subscription: drop the
+                    // rejected attempt, keep the original and let the caller retry with backoff
```
jpelgrom: Described in Timber log

```suggestion
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -1000,74 +1073,192 @@ internal class WebSocketCoreImpl(
...
+                    failed += oldId
+                    Timber.e("Subscription ${original.request} rejected, retrying with backoff")
+                }
+
+                else -> {
+                    val newId = checkNotNull(response.id) { "Response without ID" }
+                    eventSubscriptionMutex.withLock {
+                        if (activeMessages.remove(oldId, original)) {
+                            // Promote the acknowledged attempt: the original subscription with
+                            // its flow and channel is now tracked under the id the server uses
```
```suggestion
                            // Use the server's new id for the original subscription, flow and channel
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -1000,74 +1073,192 @@ internal class WebSocketCoreImpl(
...
+                            activeMessages[newId] = original
+                        } else {
+                            // The subscription was closed while the attempt was in flight, the
+                            // acknowledged id must not outlive it on the server
```
```suggestion
                            // Subscription was stopped while attempt was in flight, unsubscribe
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -1128,6 +1319,23 @@ internal sealed interface ActiveMessage {
...
+
+    /**
+     * A resubscription attempt in flight after a reconnection.
+     *
+     * Tracked under the id the server will use once it acknowledges the attempt, while
+     * [original] stays tracked under its old id so nothing is lost if the attempt fails.
+     * Events that arrive for the new id before the acknowledgement is processed are routed
+     * to [original]'s flow. On acknowledgement the attempt is promoted: [original] replaces
+     * this entry under the new id.
+     *
```
jpelgrom: This describes usage outside of the sealed class, which is already described in place

```suggestion
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImpl.kt
```diff
@@ -1179,6 +1387,20 @@ private sealed interface Command<T> {
                 )
             }
         }
+
+        /**
+         * A resubscription attempt for [original] after a reconnection.
+         *
+         * Unlike [Subscription] it is tracked as an [ActiveMessage.Reconnecting], so the active
+         * messages map keeps a single [ActiveMessage.Subscription] entry per logical subscription
+         * while the attempt is in flight.
```
```suggestion
         * When answered successfully, [original] should be tracked instead of this entire message.
```

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
```diff
@@ -1445,6 +1446,603 @@ misc
         }
     }
 
+    @Test
```
jpelgrom: We have headings / comments as sections in this file - this should get its own

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
```diff
@@ -1445,6 +1446,603 @@ misc
...
+            webSocketListener.onMessage(
+                mockConnection,
+                """{"id":$newId, "type":"event", "event":{"event_type":"state_changed", "time_fired":"2016-11-26T01:37:24.265429+00:00", "data": {"entity_id":"light.bed_light"}}}""",
+            )
+            assertEquals("light.bed_light", awaitItem().entityId)
+        }
+    }
+
+    @Test
+    fun `Given an established connection When a ping is answered Then the connection is kept`() = runTest {
```
jpelgrom: These 3 ping tests should be a different section, not hidden between the other tests

common/src/main/kotlin/io/homeassistant/companion/android/common/data/websocket/WebSocketCore.kt
```diff
@@ -51,6 +51,17 @@ internal interface WebSocketCore {
      */
     suspend fun sendBytes(data: ByteArray): Boolean
 
+    /**
+     * Sends an application-level ping and verifies that the connection is alive.
+     *
+     * If no pong is received while the connection is believed to be established, the connection is
+     * cancelled: this surfaces a silently dropped socket (e.g. the server restarted without the
+     * TCP connection being reset) to the close handling, which restores active subscriptions.
```
jpelgrom: Describe why before what

```suggestion
     * If no pong is received while the connection is believed to be established, this indicates a
     * silently dropped connection. The function will properly handle this as a closed connection
     * and try restoring any active subscriptions.
```

jpelgrom (approved):
Thanks for picking this up @TimoPtr. I'm not seeing any issues while testing and your comments make sense, OK to merge this to see if it improves as far as I'm concerned.

Code comments are a bit much in my opinion, especially in the tests which are a distinctly different style, but that's partially your style and the AI gen origins I'm afraid. I've added some suggestions for the worst ones.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/websocket/impl/WebSocketCoreImplTest.kt
TimoPtr: I'll make proper NestedTest instead.

Rudd-O:
~This does not seem to have fixed the issue for me.  Whenever I restart HA, notifications always fail until I force-close the app.  I have to assume this hasn't been released yet?  Where can I test prereleases?~

EDIT: this fully fixed the issue once I installed what Obtainium believes to be a prerelease, which is in fact 2026.8.X.  So thank you @TimoPtr for this amazing work.

jpelgrom:
This change is in version 2026.8.0 and newer of the app, which are indeed still in beta. We hope to make a new stable release next week.

(You can download each beta here from GitHub or use the Play Store beta program, no need to use 3rd party sources.)

PR #7176, opened 2026-07-13
Use sensor display precision for entity states in Android Auto

### LINKED ISSUE #6564: Android Auto part of the companion app not using display precision
2026.3.2-full

**Android version(s):**
16

**Device model(s):**
Samsung S22 Ultra

**Home Assistant version:**
2026.3.1

2026.3.1

I'm using the Android AUTO part of this companion app. I've created a few "Android Auto Favorites", but those favorites are not using the "Display Precision" as been defined in the SENSOR screen itself. Result is that you now have a lot of decimals on that tiny screen. 

**Companion App Logs:**
none
```

```

None

**Additional information:**
Would be fine if the Android Auto is also using the predefined Display Precision like the rest of the app is doing.

pimvanoerle:
## Summary

Fixes #6564

The Android Auto / Android Automotive entity grid (favourites and domain lists) displayed raw sensor states, ignoring the display precision configured on the server — resulting in values like `21.126456` on the car screen while widgets and the frontend showed `21.1`.

The precision formatting already existed in `Entity.friendlyState()`, but `EntityGridVehicleScreen` never passed the entity registry options to it, so the formatting branch never ran. This change indexes the registry (which the screen already receives) by entity ID and passes each entity's options through — the same data flow the entity widget uses. Both the explicit `displayPrecision` and the integration-provided `suggestedDisplayPrecision` fallback are honored, matching frontend behavior.

## Checklist

- [x] Changes are backward compatible whenever feasible. Entities without registry options keep the previous behavior.

## Screenshots

Screenshot: favourites grid on an Android Automotive emulator showing two temperature sensors with different server-configured precisions — one at 1 decimal, one at 0 — both rendered correctly

android-auto-precision-fix

## Any other notes

- `friendlyState()`'s precision logic had no test coverage; this PR adds unit tests for it (precision priority, precision 0, no options, non-sensor domains, unit-of-measurement append).
- The screen wiring itself has no automated test, as there is currently no test infrastructure for the car app `Screen` classes; it was verified manually on an Android Automotive emulator (API 33) against a live Home Assistant instance (see screenshot).
- The entity registry is fetched once per car app session, consistent with how the registry was already used on these screens — a precision change made on the server appears after the car app restarts (the in-app refresh action reloads entity states, not the registry). Making it live-update would require subscribing to registry updates, which felt out of scope for this fix.
- Changelog updated for Main and Automotive.


common/src/test/kotlin/io/homeassistant/companion/android/common/data/integration/EntityTest.kt
```diff
@@ -221,4 +229,73 @@ class EntityTest {
...
+            initialLocale = Locale.getDefault()
+            // The precision formatting uses the default locale, pin it so assertions are stable
+            Locale.setDefault(Locale.US)
+            SdkVersion.sdkInt = TEST_SDK_VERSION
+        }
+
+        @AfterEach
+        fun tearDown() {
+            Locale.setDefault(initialLocale)
+        }
```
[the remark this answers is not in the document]
pimvanoerle: Good shout :) - added in 13cfc1ad just now.

common/src/test/kotlin/io/homeassistant/companion/android/common/data/integration/EntityTest.kt
```diff
@@ -221,4 +229,73 @@ class EntityTest {
...
+
+        @ParameterizedTest(name = "displayPrecision={0}, suggestedDisplayPrecision={1} -> {2}")
+        @CsvSource(
+            "2, , 20.13",
+            "0, , 20",
+            ", 1, 20.1",
+            "3, 1, 20.126",
+        )
```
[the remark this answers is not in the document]
pimvanoerle: I don't agree with this one - unquoted empty values in @CsvSource are converted to null for reference types by JUnit 5 and up, so the nullable Int? parameters receive null as intended. These cases pass locally and in CI. 

See: https://docs.junit.org/6.1.2/writing-tests/parameterized-classes-and-tests.html#sources-CsvSource

"An empty, quoted value ('') results in an empty String unless the emptyValue attribute is set; whereas, an entirely empty value is interpreted as a null reference." and "An unquoted empty value will always be converted to a null reference regardless of any custom values configured via the nullValues attribute."

Happy to switch to an explicit null sentinel if reviewers prefer it for readability, of course!

pimvanoerle:
> Hi @pimvanoerle
> 
> It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant/android#7176).
> 
> 
> Thanks!

Done! Thanks :)

app/src/main/res/xml/changelog_master.xml
```diff
@@ -2,6 +2,7 @@
 <changelog xmlns:tools="http://schemas.android.com/tools"
     tools:ignore="MissingDefaultResource">
     <release version="2026.7.3 - Main" versioncode="3">
+        <change>Android Auto: entity states now use the sensor display precision configured on the server</change>
```
TimoPtr: This is fixing a bug I don't think we should mention it in the changelog. (Same for automotive)

app/src/main/kotlin/io/homeassistant/companion/android/vehicle/EntityGridVehicleScreen.kt
```diff
@@ -65,6 +65,10 @@ class EntityGridVehicleScreen(
     var entities: List<Entity> = listOf()
     private val isFavorites = title == carContext.getString(R.string.favorites)
 
+    // Index the registry by entity ID so each grid item can look up its options (such as the
+    // sensor display precision) without scanning the whole registry on every render
+    private val entityRegistryOptions = entityRegistry?.associate { it.entityId to it.options }
```
TimoPtr, line 70: I'm not a fan of doing this associate here because it's on the main thread but it is the current way of doing things at the moment so let's keep it.

TimoPtr (approved):
Just remove the changes in the changelog and I think we are good to go 👍🏻 thanks for fixing the bug.

app/src/main/res/xml/changelog_master.xml
pimvanoerle: Good point - that's them taken out now

pimvanoerle:
That's the changelog lines back out - thanks for the review :)

jpelgrom (approved):

PR #7179, opened 2026-07-15
Remove custom apply of safe area and rely on WebView

TimoPtr:
## Summary

It seems that we the new FrontendScreen we don't need to send the insets anymore to the Frontend. I decided to keep the handling of the statusbar within the app so we don't see the menu button jumps while rendering the frontend for the first time. I tested on Android 16 and 17 and it works ok with a top notch. I wonder if I've missed something.

## Checklist

## Screenshots

image
image

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
@@ -20,7 +20,6 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.WindowInsetsSides
 import androidx.compose.foundation.layout.asPaddingValues
-import androidx.compose.foundation.layout.displayCutout
 import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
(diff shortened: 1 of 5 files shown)


jpelgrom:
It looks like WebView behavior has changed in version 144 (early this year) to handle this automatically: https://developer.android.com/develop/ui/views/layout/webapps/understand-window-insets#feature-compatibility. Long overdue :)

While I understand you want to remove the custom variables, from experience not every user/device has updated WebView yet (especially when not using Play Services) and some manufacturers ship WebViews with slight differences. Dropping our custom code and relying on the 144+ behavior only could introduce new issues. I think the issue was with not consuming correctly in landscape, that also shouldn't be a very big change to make?

TimoPtr:
> It looks like WebView behavior has changed in version 144 (early this year) to handle this automatically: [developer.android.com/develop/ui/views/layout/webapps/understand-window-insets#feature-compatibility](https://developer.android.com/develop/ui/views/layout/webapps/understand-window-insets#feature-compatibility). Long overdue :)
> 
> While I understand you want to remove the custom variables, from experience not every user/device has updated WebView yet (especially when not using Play Services) and some manufacturers ship WebViews with slight differences. Dropping our custom code and relying on the 144+ behavior only could introduce new issues. I think the issue was with not consuming correctly in landscape, that also shouldn't be a very big change to make?

Thanks for searching for the root cause 👍🏻 it makes sense. Now I'm not sure how to address this properly, I don't think we have a way to know if the WebView we are using is supporting the insets properly (not using chromium for instance). If we want to prevent the double safe area we need to do it in the frontend and wait the next version.

jpelgrom:
We can get the WebView package and version on most devices. I'm not sure I want to have a comparison of the version however, but at the same time using 'normal' CSS variables instead of a custom one is probably better for community developers.

TimoPtr:
> We can get the WebView package and version on most devices. I'm not sure I want to have a comparison of the version however, but at the same time using 'normal' CSS variables instead of a custom one is probably better for community developers.

I have an idea I need to check. Maybe we can manage to consume the insets before it reach the WebView. Doing so would disable the feature from the WebView when it is available.

TimoPtr:
Ok I found the root cause of the issue, It's on the HAApp that does apply the horizontal insets all the time. I'm going to close this PR and open a new one.

PR #7183, opened 2026-07-16
Disable Horizontal insets in HAApp for FrontendScreen

TimoPtr:
## Summary

HAApp was always applying the Horizontal insets and we were also sending the insets to the Frontend to apply the safe area padding. It was causing double padding being applied. To address the issue I've simply skip the insets when we are in the frontend screen since we need the WebView to handle it to have the right color everywhere when a user override the theme.

## Checklist


jpelgrom (approved):
Tested with webview 133 and 150

PR #7184, opened 2026-07-16
Use GetEntitiesForDisplayUseCase in ManageControls

### LINKED ISSUE #7181: Use GetEntitiesForDisplayUseCase in Control settings screen
### Description

We want to use `GetEntitiesForDisplayUseCase` in `ManageControlsSettings` to not rely on friendly name when the version of the server allows it.

### Additional context

_No response_

TimoPtr:
## Summary

This PR closes https://github.com/home-assistant/android/issues/7181 by replacing the usage of friendly name with the `GetEntitiesForDisplayUseCase` in `ManageControls`.

I've added the icon of the Entity in the list too, to be like in the other entity pickers.

## Checklist

## Screenshots

image


app/src/main/kotlin/io/homeassistant/companion/android/settings/controls/ManageControlsViewModel.kt
```diff
@@ -86,24 +101,18 @@ class ManageControlsViewModel @Inject constructor(
...
-                    } catch (e: CancellationException) {
-                        throw e
-                    } catch (e: Exception) {
-                        Timber.e(e, "Failed to get entities")
-                        null
-                    }
-                    if (entities != null) {
-                        entitiesList[server.id] = entities
+                    // The flow completes with a terminal state after Loading, failures surface as Error
+                    // and leave the server out of the list, like the previous behavior.
```
jpelgrom: We don't refer to previous behavior...

leave the server out of the list to not block configuration of other server's entities

jpelgrom:
Relatively simple change here, but CI is impacted by the flaky test.

> I've hidden the entity marked as hidden.

This is a behavior change compared to before. As the list of entities you can add are not limited by hidden right now I wouldn't include it, or that would have to be changed at the same time but then we have to call it out in the changelog in my opinion for users that added a hidden entity to their controls.

TimoPtr:
> Relatively simple change here, but CI is impacted by the flaky test.
> 
> > I've hidden the entity marked as hidden.
> 
> This is a behavior change compared to before. As the list of entities you can add are not limited by hidden right now I wouldn't include it, or that would have to be changed at the same time but then we have to call it out in the changelog in my opinion for users that added a hidden entity to their controls.

Thanks for catching it, I'll remove since we've agreed that they are going to be shown in search only in #7199

jpelgrom:
This interface has no search though

jpelgrom (approved):

PR #7189, opened 2026-07-16
Introduce observe entity for display

### LINKED ISSUE #7190: Offer a observe method to get update of Entities to be displayed
### Description

We should be able to observe changes on the entities and other registries so that the displayed entities are up to date.

### Additional context

_No response_

TimoPtr:
## Summary

In the automotive screen we are going to need to be able to have updates of the entity that we are going to be display. This PR add the `obverse` capability to the `GetEntityForDisplayUseCase` so that a consumer can get the entity updates from the WebSocket when available. I replace the invoke operator with a `snapshot` method.

I tested the `observe` method by simply replacing the snapshot per a observe in the `ManageTileViewModel` and the entities were properly updating within the UI when the state was changing.

## Checklist

## Any other notes

I'm not sure if we should use `observe` instead of `snapshot` in the EntityPicker. @jpelgrom wdyt?

### DIFF
--- a/common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntityDisplayItem.kt
+++ b/common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntityDisplayItem.kt
@@ -4,8 +4,12 @@ import androidx.compose.runtime.Immutable
 import androidx.compose.ui.unit.LayoutDirection
 import com.mikepenz.iconics.typeface.IIcon
 import io.homeassistant.companion.android.common.data.integration.Entity
+import io.homeassistant.companion.android.common.data.integration.FriendlyState
 import io.homeassistant.companion.android.common.data.integration.friendlyName
+import io.homeassistant.companion.android.common.data.integration.friendlyState
 import io.homeassistant.companion.android.common.data.integration.getIcon
+import io.homeassistant.companion.android.common.data.integration.isActive
+import io.homeassistant.companion.android.common.data.integration.isExecuting
 
 private const val CATEGORY_CONFIG = "config"
 private const val CATEGORY_DIAGNOSTIC = "diagnostic"
@@ -25,12 +29,31 @@ enum class EntityCategory {
     }
 }
 
-/** Fully resolved display information for an entity. */
+/** Geographic position of an entity, resolved from its state attributes. */
+@Immutable
+data class EntityCoordinates(val latitude: Double, val longitude: Double)
+
+/**
+ * Everything needed to display an entity, fully resolved from the entity state and the
+ * registries: the single source of truth for entity displays, so anything a display needs
+ * should be resolved into this class at creation (see the [EntityDisplayItem] constructor
+ * that takes a [Entity]).
+ *
+ * The item is a snapshot at resolution time.
+ */
 @Immutable
 data class EntityDisplayItem(
     val entityId: String,
     val name: String,
     val icon: IIcon,
+    val state: FriendlyState = FriendlyState.Literal(""),
+    val rawState: String = "",
+    /** Whether the entity is currently executing an action. */
+    val isExecuting: Boolean = false,
+    /** Whether the entity is in an active state, for state-colored rendering. */
+    val isActive: Boolean = false,
+    /** Geographic position of the entity, null when it has none. */
+    val coordinates: EntityCoordinates? = null,
     val areaName: String? = null,
     val floorName: String? = null,
     val deviceName: String? = null,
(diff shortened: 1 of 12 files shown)


common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntityDisplayItem.kt
```diff
@@ -25,12 +30,39 @@ enum class EntityCategory {
...
+    val state: FriendlyState = FriendlyState.Literal(""),
+    val rawState: String = "",
+    /** Whether the entity is currently executing an action. */
+    val isExecuting: Boolean = false,
+    /** Whether the entity is in an active state, for state-colored rendering. */
+    val isActive: Boolean = false,
+    /** Geographic position of the entity, null when it has none. */
+    val coordinates: EntityCoordinates? = null,
+    /** Alarm panel display information, only set for `alarm_control_panel` entities. */
+    val alarm: AlarmDisplay? = null,
```
TimoPtr: @jpelgrom I was going to introduce a sealed class on top of the `EntityDisplayItem` but I'm not sure at the moment and went with the `null` approach and kept a generic entity.

jpelgrom:
> I'm not sure if we should use `observe` instead of `snapshot` in the EntityPicker. @jpelgrom wdyt?

I think we shouldn't use `observe` unless there is a need to keep things live, to limit data use and recompositions - a picker that just loaded should be fine with `snapshot`, not expecting it to go out of date.

TimoPtr:
> > I'm not sure if we should use `observe` instead of `snapshot` in the EntityPicker. @jpelgrom wdyt?
> 
> I think we shouldn't use `observe` unless there is a need to keep things live, to limit data use and recompositions - a picker that just loaded should be fine with `snapshot`, not expecting it to go out of date.

That's the direction I've followed in this PR but I wanted your opinion. It's nice to see the list updating but I think probably overkill.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/GetEntitiesForDisplayUseCase.kt
```diff
@@ -101,16 +111,83 @@ class GetEntitiesForDisplayUseCase @Inject constructor(
...
+     * subscription is available (for example when the websocket connection is unavailable).
+     */
+    fun observe(serverId: Int, filter: (Entity) -> Boolean = { true }): Flow<EntityDisplayState> = flow {
+        val initial = loadEntities(serverId) ?: return@flow
+
+        val version = serverManager.getServer(serverId)?.version
+        val entities = LinkedHashMap<String, Entity>()
+        initial.filter(filter).forEach { entities[it.entityId] = it }
+
+        val webSocketRepository = webSocketRepositoryOrNull(serverId)
```
jpelgrom, line 139: Keep the version close to where you first use it

```suggestion
        val entities = LinkedHashMap<String, Entity>()
        initial.filter(filter).forEach { entities[it.entityId] = it }

        val webSocketRepository = webSocketRepositoryOrNull(serverId)
        val version = serverManager.getServer(serverId)?.version
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/GetEntitiesForDisplayUseCase.kt
```diff
@@ -150,82 +227,112 @@ class GetEntitiesForDisplayUseCase @Inject constructor(
...
+            classicEntries = classicEntries?.associateBy { it.entityId },
+        )
+    }
+
+    private fun resolveCustomIcon(customIcon: String?): IIcon? = customIcon
+        ?.takeIf { it.startsWith(MDI_PREFIX) }
+        ?.let { CommunityMaterial.getIconByMdiName(it) }
+}
+
+private suspend fun <T> fetchOrNull(registryName: String, fetch: suspend () -> T?): T? = try {
```
jpelgrom, line 331: Why was this moved out of the usecase class?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/GetEntitiesForDisplayUseCase.kt
```diff
@@ -101,16 +111,83 @@ class GetEntitiesForDisplayUseCase @Inject constructor(
...
+     * current display truth (including the state-dependent [EntityDisplayItem.icon] and
+     * [EntityDisplayItem.state]). Updates that leave the items unchanged are skipped. Since
+     * emissions are complete snapshots, slow consumers can safely `conflate()` the flow and
+     * only render the latest one.
+     *
+     * The flow completes with [EntityDisplayState.Error] when the entities cannot be
+     * retrieved, and completes after the initial [EntityDisplayState.Loaded] when no update
+     * subscription is available (for example when the websocket connection is unavailable).
+     */
+    fun observe(serverId: Int, filter: (Entity) -> Boolean = { true }): Flow<EntityDisplayState> = flow {
```
jpelgrom, line 132: A flexible filter is nice, but if the filter is a specific set of entity IDs we can filter server-side and avoid data use and processing locally (see: `getEntityUpdates(entityIds: List<String>)` in `IntegrationRepository`). I want to avoid a full rewrite but it would be nice if this variant could be added down the line with minimal changes as it is quite relevant basically everywhere except the vehicle interface you're adapting now.

jpelgrom (commented):
With the rename there is no longer an invoke option for the use case here, already bending our architecture rules which aren't that old. OK with the exception for now but if we start adding more logic to the use case it might be time to upgrade it.

Noisy diff because you also tweaked a lot of other functions slightly, please try to avoid it unless necessary. But it seems to mostly be correct and is working correctly.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/GetEntitiesForDisplayUseCase.kt
TimoPtr: It was not used in the class, TBH this has been removed in the next PR. I would like to merge this one if you are ok with this small change to avoid another round of rebase.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/GetEntitiesForDisplayUseCase.kt
TimoPtr: Let's discuss this in the next PR as the logic changed quite a lot.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/GetEntitiesForDisplayUseCase.kt
TimoPtr: Same this changed quite a lot in the next PR.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/GetEntitiesForDisplayUseCase.kt
TimoPtr: > if the filter is a specific set of entity IDs we can filter server-side

I don't think we have an API for that at the moment do we?

jpelgrom (approved):
Let's follow up on the entity use case/manager comments in the next PR. Snapshot works.

PR #7192, opened 2026-07-16
Handle FrontendScreen paused

TimoPtr:
## Summary

The ViewModel is surviving the pause of the screen because of that the logic of timeout of the WebView continues in background and if a URL changes occur while the app is in background it might show the timeout screen. With this change it would discard the timeout while in background and restart when opening again the app. It also discard the timeout page if ever the connection happen while the timeout is shown.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
@@ -58,6 +58,7 @@ import androidx.hilt.lifecycle.viewmodel.compose.hiltViewModel
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.compose.LifecycleEventEffect
+import androidx.lifecycle.compose.LifecycleStartEffect
 import androidx.lifecycle.compose.LocalLifecycleOwner
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.lifecycle.repeatOnLifecycle
@@ -192,6 +193,7 @@ internal fun FrontendScreen(
         onDownloadRequested = viewModel::onDownloadRequested,
         webViewActions = viewModel.webViewActions,
         onSafeAreaInsetsChanged = viewModel::onSafeAreaInsetsChanged,
+        onScreenStartedChanged = viewModel::onScreenStartedChanged,
         onGesture = viewModel::onGesture,
         onLeavingApp = viewModel::onLeavingApp,
         onExoPlayerFullscreenChanged = viewModel::onExoPlayerFullscreenChanged,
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewModel.kt
```diff
@@ -729,12 +744,10 @@ internal class FrontendViewModel @VisibleForTesting constructor(
     }
 
     /**
-     * Waits the [CONNECTION_TIMEOUT] in [FrontendViewState.Loading] and emits an
-     * [FrontendConnectionError.ExternalBusTimeout] if the frontend has not completed its external-bus
-     * handshake (transitioned to [FrontendViewState.Content]) by then.
+     * Waits the [CONNECTION_TIMEOUT] and emits an [FrontendConnectionError.ExternalBusTimeout] if the
+     * frontend has not completed its external-bus handshake (left [FrontendViewState.Loading]) by then.
```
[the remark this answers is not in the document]
jpelgrom:
```suggestion
     * Awaits the [CONNECTION_TIMEOUT] and emits a [FrontendConnectionError.ExternalBusTimeout] if the
     * frontend has not completed its external-bus handshake (left [FrontendViewState.Loading]) by then.
```

TimoPtr:
I tested with my production setup that has internal/external URL put the app in background turned off my wifi wait for 30s (the timeout) and open the app again
Previous behavior was showing the timeout screen
New behavior it loads the frontend

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
```diff
@@ -393,6 +400,14 @@ private fun FrontendScreenEffects(
     LeavingAppEffect(webView = webView, onLeavingApp = onLeavingApp)
 }
 
+@Composable
```
jpelgrom, line 407: We document composable functions in this file

```suggestion
/**
 * Reports the lifecycle start/stop events to [onScreenStartedChanged] whenever the
 * [LifecycleStartEffect] changes.
 */
@Composable
```

jpelgrom (approved):
Got enough network switching in without crashing the webview (native, unrelated to this PR) to validate the behavior and the change in this PR. Not the biggest fan of a lot of lifecycle callbacks but this seems like the best available option without switching to an `AndroidViewModel` which would introduce other complications.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendScreen.kt
jpelgrom: I'm not completely sold on this description, feel free to adjust it or write your own but following our own logic we should give a short description.

TimoPtr: I did made it shorter by removing how potentially it could be use to only the facts.

PR #7203, opened 2026-07-17
Adjust loading screen to hide frontend splash screen and fit

### LINKED ISSUE #3142: Improve splashscreen by hiding webview until fully loaded.

TimoPtr:
## Summary

Thanks to @bgoncal we can properly hide the frontend loading screen (we also need https://github.com/home-assistant/frontend/pull/53171). 
This PR also align the color of the splashscreen to fit our colors and made the loading experience smooth.

I've adjust the tests also to properly use `assertInstanceOf`.

## Checklist

## Screenshots

https://github.com/user-attachments/assets/341f6b8c-350c-4ae1-84b9-bf574dfa3691

## Any other notes

We need to wait for the frontend to merge the PR, otherwise we might see sometimes the frontend loading screen.


app/src/main/kotlin/io/homeassistant/companion/android/frontend/FrontendViewState.kt
```diff
@@ -51,6 +51,11 @@ sealed interface FrontendViewState {
         override val serverId: Int,
         override val url: String,
         val target: FrontendTarget = FrontendTarget.Default,
+        /**
+         * The external bus handshake completed but the loading screen is kept until the frontend
+         * reports it finished rendering (servers reporting the `loaded` event).
```
jpelgrom: Description should be moved to the class description as a `@param`

app/src/main/kotlin/io/homeassistant/companion/android/launch/LaunchActivity.kt
```diff
@@ -186,6 +186,17 @@ class LaunchActivity : AppCompatActivity() {
             viewModel.shouldShowSplashScreen()
         }
 
+        // Skip the system's default splash exit animation: the app's loading screen shows the
+        // same centered icon, so removing the splash instantly makes the transition seamless.
+        // Only on API 31+: below that the system draws no exit animation, and asking for the
+        // splash screen view would inflate a compat layout that needs androidx's
+        // Theme.SplashScreen attributes, which Theme.LaunchScreen does not define.
```
jpelgrom: This feels like AI explanation to me, it can be shorter

```suggestion
        // Skip the system's default splash exit animation for a seamless transition to
        // the icon on the loading screen.
```

app/src/main/kotlin/io/homeassistant/companion/android/loading/LoadingScreen.kt
```diff
@@ -1,54 +1,208 @@
...
 @Composable
 private fun LoadingScreenPreview() {
     HAThemeForPreview {
         LoadingScreen()
     }
 }
+
+@HAPreviews
+@Composable
+private fun LoadingScreenPreview_with_branding() {
```
jpelgrom:
```suggestion
private fun LoadingScreenPreviewWithBranding() {
```

jpelgrom (approved):
Minor suggestions but not adding a lot here as the next PR changes more of it

PR #7206, opened 2026-07-18
Fix notification sensor settings initialization

### LINKED ISSUE #6773: Notification sensor enable initially does not allow setting "Allow list"

**Android version(s):**

**Device model(s):**

**Home Assistant version:**

**Companion App Logs:**

```

```

**Additional information:**

loganrosen:
## Summary

Notification allow-list settings were only persisted after a notification callback. As a result, a newly enabled notification sensor could show no configurable settings until a notification was posted or removed.

Initialize the settings during the normal sensor refresh for enabled posted and removed notification sensors. The callback paths use the same settings helper, preserving existing values and filtering behavior. This also adds regression coverage for both sensors and custom allow-list preservation.

Fixes #6773

## Checklist

## Screenshots

Not applicable; this fixes when existing settings become available and does not change their appearance.

No documentation changes are required.

## Any other notes

None.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/sensors/NotificationListenerSensorManager.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/sensors/NotificationListenerSensorManager.kt
@@ -128,6 +128,9 @@ class NotificationListenerSensorManager @Inject constructor(
     }
 
     override suspend fun requestSensorUpdate() {
+        listOf(lastNotification, lastRemovedNotification)
+            .filter { isEnabled(it) }
+            .forEach { getNotificationSettings(it) }
         updateMediaSession()
     }
 
@@ -145,22 +148,11 @@ class NotificationListenerSensorManager @Inject constructor(
                 return@launch
             }
 
-            val allowPackages = getSetting(
-                lastNotification,
-                SETTING_ALLOW_LIST,
-                SensorSettingType.LIST_APPS,
-                default = "",
-            ).split(", ").filter { it.isNotBlank() }
-
-            val disableAllowListRequirement = getToggleSetting(
-                lastNotification,
-                SETTING_DISABLE_ALLOW_LIST,
-                default = false,
-            )
+            val settings = getNotificationSettings(lastNotification)
 
             if (sbn.packageName == applicationContext.packageName ||
-                (allowPackages.isNotEmpty() && sbn.packageName !in allowPackages) ||
-                (!disableAllowListRequirement && allowPackages.isEmpty())
+                (settings.allowPackages.isNotEmpty() && sbn.packageName !in settings.allowPackages) ||
+                (!settings.disableAllowListRequirement && settings.allowPackages.isEmpty())
             ) {
                 return@launch
             }
@@ -208,22 +200,11 @@ class NotificationListenerSensorManager @Inject constructor(
                 return@launch
             }
 
-            val allowPackages = getSetting(
-                lastRemovedNotification,
-                SETTING_ALLOW_LIST,
-                SensorSettingType.LIST_APPS,
-                default = "",
-            ).split(", ").filter { it.isNotBlank() }
-
-            val disableAllowListRequirement = getToggleSetting(
-                lastRemovedNotification,
-                SETTING_DISABLE_ALLOW_LIST,
-                default = false,
-            )
+            val settings = getNotificationSettings(lastRemovedNotification)
 
             if (sbn.packageName == applicationContext.packageName ||
-                (allowPackages.isNotEmpty() && sbn.packageName !in allowPackages) ||
-                (!disableAllowListRequirement && allowPackages.isEmpty())
+                (settings.allowPackages.isNotEmpty() && sbn.packageName !in settings.allowPackages) ||
+                (!settings.disableAllowListRequirement && settings.allowPackages.isEmpty())
             ) {
                 return@launch
             }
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/sensors/NotificationListenerSensorManager.kt
```diff
@@ -128,6 +128,9 @@ class NotificationListenerSensorManager @Inject constructor(
     }
 
     override suspend fun requestSensorUpdate() {
+        listOf(lastNotification, lastRemovedNotification)
+            .filter { isEnabled(it) }
+            .forEach { getNotificationSettings(it) }
```
jpelgrom, line 134: Without knowing the context for this, it doesn't make a lot of sense why you're getting the setting but are not using it.

```suggestion
        // Retrieve settings to create the allowlist setting for the sensors, if needed
        listOf(lastNotification, lastRemovedNotification)
            .filter { isEnabled(it) }
            .forEach { getNotificationSettings(it) }

```

app/src/main/res/xml/changelog_master.xml
```diff
@@ -4,6 +4,7 @@
     <release version="2026.7.4 - Main" versioncode="3">
         <change>Android 17: added assistant volume level sensor + notification command for control</change>
         <change>Health Connect sleep duration sensor now ignores awake and out of bed time</change>
+        <change>Notification sensor allow list can now be configured immediately after enabling the sensor</change>
```
jpelgrom: We usually don't mention bug fixes, only notable new features and changes

app/src/test/kotlin/io/homeassistant/companion/android/sensors/NotificationListenerSensorManagerTest.kt
```diff
@@ -0,0 +1,137 @@
...
+        mockk<ServerManager>(relaxed = true),
+    )
+
+    @After
+    fun tearDown() {
+        unmockkStatic(NotificationManagerCompat::class)
+    }
+
+    @Test
+    fun `Given last notification enabled when requesting update then its settings are initialized`() = runTest {
```
jpelgrom: You can parametrize this and the next test, only the sensor ID is different

loganrosen: Updated these to use JUnit Jupiter `@ParameterizedTest` with a shared `@MethodSource`, so both the initialization and existing-value preservation cases run for each notification sensor ID. This path only tests repository-setting behavior, so mocking `Context` keeps the tests independent of Robolectric.

app/src/main/res/xml/changelog_master.xml
loganrosen: Removed the changelog entry.

app/src/main/kotlin/io/homeassistant/companion/android/sensors/NotificationListenerSensorManager.kt
loganrosen: Added a comment explaining that loading the settings here persists their defaults before the sensor detail screen observes them.

jpelgrom (approved):
Thanks for changing this, nice improvement to the user experience!

app/src/main/kotlin/io/homeassistant/companion/android/sensors/NotificationListenerSensorManager.kt
```diff
@@ -360,6 +342,24 @@ class NotificationListenerSensorManager @Inject constructor(
         )
     }
 
+    private suspend fun getNotificationSettings(sensor: SensorManager.BasicSensor): NotificationSettings {
+        val allowPackages = getSetting(
```
TimoPtr, line 328: I really don't like the pattern this fix has to lean on: a get that initializes and persists the value as a side effect. This PR even makes it explicit, requestSensorUpdate now calls a getter and discards the result purely so the write happens. To be clear, that side effect is not introduced here,       
  getSetting has always worked like this, and given the current architecture this is the correct fix for the bug.                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                   
  I recently removed exactly this behavior for the sensors themselves in https://github.com/home-assistant/android/pull/7040: the repository returns the default when nothing is stored, and only an actual change persists to the DB. I didn't look at the settings back then, but they should follow the same    
  model. The reason it can't be a drop-in change is that the detail screen currently discovers which settings exist from the stored rows (getSettingsFlow), so the write-on-read is what makes settings show up at all. Moving to the #7040 model means sensor managers declare their available settings and the UI
  merges that declaration with the stored values, at which point this initialization path, and the need for this fix, disappears entirely.                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                   
  That refactor is clearly out of scope here, so nothing to change in this PR. See https://github.com/home-assistant/android/issues/7225

app/src/main/kotlin/io/homeassistant/companion/android/sensors/NotificationListenerSensorManager.kt
```diff
@@ -145,22 +149,11 @@ class NotificationListenerSensorManager @Inject constructor(
...
-                SETTING_DISABLE_ALLOW_LIST,
-                default = false,
-            )
+            val settings = getNotificationSettings(lastNotification)
 
             if (sbn.packageName == applicationContext.packageName ||
-                (allowPackages.isNotEmpty() && sbn.packageName !in allowPackages) ||
-                (!disableAllowListRequirement && allowPackages.isEmpty())
+                (settings.allowPackages.isNotEmpty() && sbn.packageName !in settings.allowPackages) ||
+                (!settings.disableAllowListRequirement && settings.allowPackages.isEmpty())
```
TimoPtr: This logic is duplicated line 204-208 and should be merged into one method.

TimoPtr (changes requested):
Let's just avoid the duplication and we can merge it in this current state.

app/src/main/kotlin/io/homeassistant/companion/android/sensors/NotificationListenerSensorManager.kt
loganrosen: Agreed. The current behavior is necessary for the existing settings discovery flow, and #7225 tracks replacing the write-on-read model outside this fix.

app/src/main/kotlin/io/homeassistant/companion/android/sensors/NotificationListenerSensorManager.kt
loganrosen: Extracted the shared enablement and allow-list filtering into `shouldProcessNotification`, which is now used by both posted and removed notification callbacks.

TimoPtr (approved):

PR #7231, opened 2026-07-21
Introduce EntitiesForDisplayManager to retrive entities for display

### LINKED ISSUE #7190: Offer a observe method to get update of Entities to be displayed

TimoPtr:
## Summary

This PR makes the use case smarter and expose a way to only get the entity information outside of the context for display which is cheaper and most of the time we don't need the context (area, device, ...). The idea of this use case is to expose the entity without having to go through the state logic (name, icon, ...) and exposes everything needed for display.

## Checklist


common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntitiesForDisplayManager.kt
```diff
@@ -0,0 +1,472 @@
...
+    /** An area registry entry changed; the floors are refetched along, they resolve through areas. */
+    data object AreaRegistryChanged : ObserveEvent
+
+    /** A device registry entry changed. */
+    data object DeviceRegistryChanged : ObserveEvent
+}
+
+/**
+ * Registry data fetched once per resolution, indexed by id for the merge.
+ */
```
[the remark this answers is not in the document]
jpelgrom: @TimoPtr This is a valid comment, I noticed this as well. I think this is for `RegistrySnapshot` down below, not this class.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntitiesForDisplayManager.kt
```diff
@@ -0,0 +1,472 @@
...
+private data class RegistrySnapshot(
+    val entries: EntityRegistryEntries = EntityRegistryEntries(),
+    val devices: Map<String, DeviceRegistryResponse> = emptyMap(),
+    val areas: Map<String, AreaRegistryResponse> = emptyMap(),
+    val floors: Map<String, FloorRegistryResponse> = emptyMap(),
+)
+
+/**
+ * Use case that resolves the display information (name, area, floor, device, icon, hidden,
```
jpelgrom:
```suggestion
 * Manager that resolves the display information (name, area, floor, device, icon, hidden,
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntitiesForDisplayManager.kt
```diff
@@ -0,0 +1,472 @@
...
+
+            RegistrySnapshot(
+                devices = devices.await(),
+                areas = areas.await(),
+                floors = floors.await(),
+            ).withEntityEntries(webSocketRepository)
+        }
+
+    private suspend fun WebSocketRepository.fetchDevices(): Map<String, DeviceRegistryResponse> =
+        getDeviceRegistryOrNull().orEmpty().associateBy { it.id }
```
jpelgrom: I'm not sold on introducing a `orNull` variant when in the only place it's used you immediately follow it up with `orEmpty` - why not introduce a `getDeviceRegistryOrEmpty` function then?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntitiesForDisplayManager.kt
```diff
@@ -0,0 +1,472 @@
...
+
+    private suspend fun WebSocketRepository.fetchAreas(): Map<String, AreaRegistryResponse> =
+        getAreaRegistryOrNull().orEmpty().associateBy { it.areaId }
```
jpelgrom: I'm not sold on introducing a `orNull` variant when in the only place it's used you immediately follow it up with `orEmpty` - why not introduce a `getAreaRegistryOrEmpty` function then?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntitiesForDisplayManager.kt
```diff
@@ -0,0 +1,472 @@
...
+
+    private suspend fun WebSocketRepository.fetchFloors(): Map<String, FloorRegistryResponse> =
+        getFloorRegistryOrNull().orEmpty().associateBy { it.floorId }
```
jpelgrom: I'm not sold on introducing a `orNull` variant when in the only place it's used you immediately follow it up with `orEmpty` - why not introduce a `getFloorRegistryOrEmpty` function then?

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/display/EntitiesForDisplayManager.kt
```diff
@@ -0,0 +1,472 @@
...
+     * current display truth (including the state-dependent [EntityDisplayWithoutContext.icon] and
+     * [EntityDisplayWithoutContext.state]). Updates that leave the items unchanged are skipped. Since
+     * emissions are complete snapshots, slow consumers can safely `conflate()` the flow and
+     * only render the latest one.
+     *
+     * The flow completes with [EntityDisplayState.Error] when the entities cannot be
+     * retrieved, and completes after the initial [EntityDisplayState.Loaded] when no update
+     * subscription is available (for example when the websocket connection is unavailable).
+     */
+    fun observeInContext(
```
jpelgrom, line 130: A note on order:

 - `observe`
 - `snapshot`
 - `snapshotInContext`
 - `observeInContext`

Either group them by `observe`/`snapshot` or with/without context, now it's a somewhat random order.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -959,13 +1023,17 @@ suspend fun onEntityPressedWithoutState(entityId: String, integrationRepository:
     )
 }
 
+@Deprecated(
+    "The friendly name is not the name to display, it ignores the entity registry. Resolve the " +
+        "display name with EntitiesForDisplayManager, which reads it from EntityDisplayItem.name.",
+)
```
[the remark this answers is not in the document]
jpelgrom:
```suggestion
@Deprecated(
    "The friendly name is no longer used for display, as it ignores the entity registry. Resolve the " +
        "display name with EntitiesForDisplayManager, which reads it from EntityDisplay.name.",
)
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -959,13 +1023,17 @@ suspend fun onEntityPressedWithoutState(entityId: String, integrationRepository:
...
 val Entity.friendlyName: String
     get() = attributes["friendly_name"]?.toString() ?: entityId
 
 /**
  * Formats the entity state for display without registry metadata, so no sensor display
- * precision is applied. Prefer [EntityDisplayItem.state] when the entity has been
- * resolved through `GetEntitiesForDisplayUseCase`.
+ * precision is applied. Prefer `EntityDisplayItem.state` when the entity has been
+ * resolved through `EntitiesForDisplayManager`.
```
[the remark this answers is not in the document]
jpelgrom:
```suggestion
/**
 * Formats the entity state for display without registry metadata, so no sensor display
 * precision is applied. Prefer `EntityItem.state` when the entity has been resolved
 * through `EntitiesForDisplayManager`.
```

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/Entity.kt
```diff
@@ -959,13 +1023,17 @@ suspend fun onEntityPressedWithoutState(entityId: String, integrationRepository:
...
 
+@Deprecated(
+    "The friendly name is not the name to display, it ignores the entity registry. Resolve the " +
+        "display name with EntitiesForDisplayManager, which reads it from EntityDisplayItem.name.",
+)
 val Entity.friendlyName: String
     get() = attributes["friendly_name"]?.toString() ?: entityId
 
 /**
  * Formats the entity state for display without registry metadata, so no sensor display
```
jpelgrom: Possible alternative for `EntityDisplayWithContext` from this description: `EntityDisplayWithMeta`/`EntityDisplayWithMetadata`? If not the comment should be updated to say 'registry context'.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/IntegrationRepository.kt
```diff
@@ -119,3 +121,27 @@ internal class IntegrationRepositoryFactory @Inject constructor(
...
+/** [IntegrationRepository.getEntities], returning null when they cannot be retrieved. */
+suspend fun IntegrationRepository.getEntitiesOrNull(): List<Entity>? = orNull("entities") { getEntities() }
+
+/** [IntegrationRepository.getEntityUpdates], returning null when it cannot be subscribed to. */
+suspend fun IntegrationRepository.getEntityUpdatesOrNull(): Flow<Entity>? =
+    orNull("entity updates") { getEntityUpdates() }
+
+/** [IntegrationRepository.getEntityUpdates], returning null when it cannot be subscribed to. */
+suspend fun IntegrationRepository.getEntityUpdatesOrNull(entityIds: List<String>): Flow<Entity>? =
+    orNull("entity updates") { getEntityUpdates(entityIds) }
```
jpelgrom: Why not define these in the interface?

jpelgrom (commented):
It's a bit annoying that putting everything in one commit means we lose the history for UseCase -> Manager and EntityDisplayItem -> EntityDisplay, as it didn't detect it was a rename. When comparing it with the old UseCase/EntityDisplayItem the list of changes thankfully isn't that large. I did notice you sneaking in more changes for Wear ;)

I think you already addressed most of my comments from the previous PR and discussions, so not a lot to add. Intuitively I was kind of confused by the usage of `Context` in names since Android's `Context` is a very well-known class, but I don't have a good alternative and when ignoring Android's `Context` it does describe the added information.

jpelgrom:
I'm destroying the stack (GitHub feature) to avoid more rewriting of Git history if this one or next ones are merged

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/Entity.kt
TimoPtr: I get your point but I think context is fine since it's always attach to something where on android it's usually only context or applicationContext. I used context because it was the term used in frontend and iOS.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/IntegrationRepository.kt
TimoPtr: I was not convinced it deserved to be in the impl, I see it as a thin wrapper around the public API and any implementation (test included) would work with this extension. It keeps the impl file more focused I would say. I'm not 100% convinced FYI.

TimoPtr: I decided to move them back into the manager for now.

jpelgrom: I'm also not convinced either way, just expressing a thought. Maybe you had a very good reason.

It doesn't need to be in the implementation and can only exist in the interface, with the implementation not overriding it. Wrapping our own public API in the same file did feel wrong to me.

> I decided to move them back into the manager for now.

Also a good solution. Let's revisit if it's needed in more places in the future.

jpelgrom:
Build failure because EntityWidgetConfigureViewModel still has a GetEntitiesForDisplayUseCase

jpelgrom (approved):
I think we're good, looking at the individual change commits. Let's get this merged to avoid more issues with renames here in future PRs.

PR #7237, opened 2026-07-22
Bump wear target API to 36

TimoPtr:
## Summary

I should have done that a long time ago but I completely forgot and now we need this to happen before the end of august if we want to be able to continue to release new version. 

I did test on an emulator with Android API 36.1 and I didn't spot any issue
- Assist shortcut
- Camera tile
- Thermostat tile
- Entity tile
- Complication (not sure if it worked well)
- List of entities
- Favorites
- Template tile (from the phone)

Documentation updated https://github.com/home-assistant/companion.home-assistant/pull/1394


jpelgrom:
Looking at the behavior changes docs there are some changes we need to adjust for.

https://developer.android.com/training/wearables/versions/6/changes

 - Adjust permissions `BODY_SENSORS` / `BODY_SENSORS_BACKGROUND`
 - Test/remove option for updating tiles on tile visible (`onTileEnterEvent`) as we cannot rely on real-time delivery
 - Default system font changes (I can check this later on my Galaxy Watch as Samsung has a different font set, I'm assuming it'll override on the tile)

TimoPtr:
I was not able to test the new permission on emulator and my watch has a dead battery. If this is not merged by Monday my watch will have battery.

TimoPtr:
I've migrated to the new callback, I think we should keep it at least for old version we know it works for new API we need to test and see on a real device.

wear/src/main/kotlin/io/homeassistant/companion/android/home/views/SensorUi.kt
```diff
@@ -29,7 +30,31 @@ import io.homeassistant.companion.android.util.batterySensorManager
 import io.homeassistant.companion.android.views.ThemeLazyColumn
 import kotlinx.coroutines.runBlocking
 
+/**
+ * A permission that must be requested in two steps from [minSdk]: [background] can only be
+ * requested once [foreground] is granted.
+ */
+private data class TwoStepPermission(val foreground: String, val background: String, val minSdk: Int)
```
jpelgrom, line 37: Nice little helper. I think we can use this in `:app` as well, maybe move it to `:common` to allow for easy use in later PRs?

jpelgrom (commented):
My watch isn't cooperating to check on tiles so that will have to come tomorrow

wear/src/main/kotlin/io/homeassistant/companion/android/home/views/SensorUi.kt
TimoPtr: I'm not sure it brings a lot of value to share. A nice flow that handle the permission all the way might but out of the scope of this PR I would say.

jpelgrom:
Testing on my Galaxy Watch 4 with Samsung's flavor of Wear OS 7 (One UI Watch 8.0), the 'in view' setting of tile refreshing with the callback doesn't work reliably at all anymore. Tapping the manual refresh does very quickly, suggesting it is a callback issue and not a network issue.

I'm not seeing the system font on Home Assistant tiles so that is nice, I guess, to limit variation.

TimoPtr:
> Testing on my Galaxy Watch 4 with Samsung's flavor of Wear OS 7 (One UI Watch 8.0), the 'in view' setting of tile refreshing with the callback doesn't work reliably at all anymore. Tapping the manual refresh does very quickly, suggesting it is a callback issue and not a network issue.
> 
> I'm not seeing the system font on Home Assistant tiles so that is nice, I guess, to limit variation.

I guess it was expected no?

jpelgrom:
Yes, but in my opinion that means we should remove the option on Wear OS 7 as it's confusing to end users why it doesn't work.

TimoPtr:
> Yes, but in my opinion that means we should remove the option on Wear OS 7 as it's confusing to end users why it doesn't work.

Ok I did migrate away the existing config and disable the option in the settings. Should we mention this in the Companion app doc?

wear/src/test/kotlin/io/homeassistant/companion/android/tiles/RefreshIntervalMigrationTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        coVerify(exactly = 0) { cameraTileDao.add(timedCamera) }
+        coVerify(exactly = 1) { thermostatTileDao.add(onViewedThermostat.copy(refreshInterval = 0)) }
+        coVerify(exactly = 1) {
+            wearPrefsRepository.setTemplateTile(4, "{{ now() }}", refreshInterval = 0)
+        }
+        coVerify(exactly = 0) { wearPrefsRepository.setTemplateTile(5, any(), any()) }
+    }
+
+    @Test
+    fun `Given a device still supporting when viewed When migrating Then nothing is read or written`() = runTest {
```
jpelgrom:
```suggestion
    fun `Given a device supporting when viewed interval When migrating Then nothing is read or written`() = runTest {
```

wear/src/test/kotlin/io/homeassistant/companion/android/tiles/RefreshIntervalMigrationTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        SdkVersion.sdkInt = Build.VERSION_CODES.UPSIDE_DOWN_CAKE
+
+        migration.migrate()
+
+        confirmVerified(cameraTileDao, thermostatTileDao, wearPrefsRepository)
```
jpelgrom: Make it more specific; if you end up mocking more and still call every mocked function `confirmVerified` would return `true` as I understand it while you want to ensure they weren't touched at all.

```suggestion
        verify {
            listOf(cameraTileDao, thermostatTileDao, wearPrefsRepository) wasNot called
        }
```

Note the suggestion also requires imports.

jpelgrom (approved):
Just one minor note on tests, otherwise good to run in beta and see if anything was missed.

jpelgrom:
> Should we mention this in the Companion app doc?

Yes as there is a small section describing these settings. A simple addition of "not supported on Wear OS 6 or newer" to the "In view" bullet point should be enough in my opinion.

PR #7239, opened 2026-07-23
Report sub SSL error with snackabar instead of a real error

### LINKED ISSUE #7223: onReceivedSslError has no main-frame filter — subresource SSL error blocks entire connection screen (regression since FrontendScreen rewrite)
## Describe the bug

Connecting to a Home Assistant instance behind Cloudflare fails with "Unable to connect to Home Assistant — certificate authority is not trusted," even though the actual server connection is fine. Connectivity checks (DNS, port reachability, TLS certificate, server connection) all pass — only "Home Assistant verification" fails.

Root cause: the reported "primary error" cert is a self-signed `CN=localhost` cert on an unrelated **third-party subresource** the frontend page loads (in our case, Cloudflare's Web Analytics beacon, `static.cloudflareinsights.com/beacon.min.js`) — not the main HA connection at all. `onReceivedSslError()` in [`HAWebViewClient.kt`](https://github.com/home-assistant/android/blob/main/app/src/main/kotlin/io/homeassistant/companion/android/util/HAWebViewClient.kt) has no main-frame filter, unlike its siblings:

```kotlin
override fun onReceivedError(view: WebView?, request: WebResourceRequest?, error: WebResourceError?) {
    super.onReceivedError(view, request, error)
    // Only handle errors for the main URL being loaded
    if (request?.url?.toString() != currentUrlFlow.value) return
    ...
}

override fun onReceivedHttpError(...) {
    super.onReceivedHttpError(view, request, errorResponse)
    // Only handle errors for the main URL being loaded
    if (request?.url?.toString() != currentUrlFlow.value) return
    ...
}

override fun onReceivedSslError(view: WebView?, handler: SslErrorHandler?, error: SslError?) {
    super.onReceivedSslError(view, handler, error)
    // no URL/main-frame check at all — fires for ANY resource on the page
    onFrontendError(FrontendConnectionError.SslError(...))
}
```

So an SSL hiccup on **any** subresource the page loads — a third-party analytics script, an image, anything — gets treated as a fatal failure of the entire connection.

This appears to be a regression introduced by the `WebViewActivity` → Compose `FrontendScreen` rewrite (#7087, #7094, #7120, landing in 2026.7.2–2026.7.3). The legacy `WebViewActivity` had the same missing filter, but its `showError()` presented `ErrorType.SSL` as a **dismissible `AlertDialog`** over an otherwise still-functional, already-loaded page — so a transient/irrelevant subresource SSL error was just an annoying popup you could dismiss. The new `FrontendScreen` instead transitions to `FrontendViewState.Error`, which fully replaces the WebView content with a blocking "Unable to connect" screen and no way to see the underlying (working) page.

## To Reproduce

1. Point the app at a Home Assistant instance whose frontend page loads any third-party subresource that can produce a `SslError` (in our case: Cloudflare Web Analytics enabled on the zone, `static.cloudflareinsights.com/beacon.min.js`).
2. If that subresource ever fails its TLS handshake (edge issue, interception, blocked, whatever) while the main HA connection is otherwise completely healthy, the app shows a full-screen "Unable to connect to Home Assistant — certificate authority is not trusted" error, unrecoverable via Retry as long as the subresource keeps failing.

## Expected behavior

SSL errors on subresources that aren't the main frame being loaded should be ignored (or at most logged), the same way `onReceivedError`/`onReceivedHttpError` already filter by `request.url == currentUrlFlow.value`. Only a main-frame SSL failure should surface as a connection error.

## Workaround

Downgrading to 2026.6.5 (last version on `WebViewActivity`) avoids the hard block, since that version shows the same class of error as a dismissible dialog instead of replacing the whole screen.

## Suggested fix

Add the same main-frame URL check used in `onReceivedError`/`onReceivedHttpError` to `onReceivedSslError`, comparing `error?.url` (available via `SslError.getUrl()`) against `currentUrlFlow.value`, and only calling `onFrontendError` when they match.

## Environment

- Home Assistant Companion app: 2026.7.4 (also reproduced on 2026.7.2, 2026.7.3)
- Confirmed not present on 2026.6.5
- Android, WebView-rendered frontend (not Custom Tabs) — confirmed via `class android.net.http.SslError` in the error details

TimoPtr:
## Summary

Fixes #7223. This PR stop showing a full screen error when a sub ressources has a SSL error and show a snackbar instead with the host where the SSL error occured. I've added a snackbar test that show how a snackbar looks like but also how it renders big text.

## Checklist


jpelgrom (approved):
We previously discussed this on Discourd, good solution to better match the WebViewActivity optionality but not hiding an error users should fix.

Issue #7035, opened 2026-06-17
End to end test (E2E)


TimoPtr:
### Description

We want to make sure that on every release of core and the android mobile app the app still works as expected.

### Additional context

Part of https://github.com/OpenHomeFoundation/roadmap/issues/167

PR #7271, opened 2026-07-29
Avoid crashing when Bluetooth adapter is not available

TimoPtr:
## Summary

While working on the #7035 I found out that we are currently crashing the app when the device doesn't have a Bluetooth adapter which could happen. This PR handle this case.

## Checklist

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/frontend/improv/FrontendImprovModule.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/frontend/improv/FrontendImprovModule.kt
@@ -1,5 +1,6 @@
 package io.homeassistant.companion.android.frontend.improv
 
+import android.bluetooth.BluetoothManager
 import android.content.Context
 import android.content.pm.PackageManager
 import com.wifi.improv.ImprovManager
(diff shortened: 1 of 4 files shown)


jpelgrom:
While this is a good additional check, these calls should never happen from the frontend if there is no Bluetooth adapter because then the device should be sending [`canSetupImprov = false`](https://github.com/home-assistant/android/blob/1a294b0c004bda6cab585467e12a6acb94581db2/app/src/main/kotlin/io/homeassistant/companion/android/frontend/handler/FrontendMessageHandler.kt#L326), preventing the frontend from interacting with the app's Improv implementation. Did you check if the config sends this correctly / there is a frontend issue here as well?

TimoPtr:
> While this is a good additional check, these calls should never happen from the frontend if there is no Bluetooth adapter because then the device should be sending [`canSetupImprov = false`](https://github.com/home-assistant/android/blob/1a294b0c004bda6cab585467e12a6acb94581db2/app/src/main/kotlin/io/homeassistant/companion/android/frontend/handler/FrontendMessageHandler.kt#L326), preventing the frontend from interacting with the app's Improv implementation. Did you check if the config sends this correctly / there is a frontend issue here as well?

The issue was on hilt side, since we inject the manager into the ViewModel and the ImprovManager is in a class attribute it was initiated when created. In the initialization there is an attribute that get the Bluetooth adapter without checking if it's available and causing a NPE.

jpelgrom:
OK that would have been good context to have and obvious if a stacktrace was included :)

Checked the library and indeed we're using the adapter without additional checks. The user (= this app) is supposed to handle permission/availability checks.

jpelgrom (approved):

Issue #7095, opened 2026-06-25
NFC tags written with Android Companion App v2026.6.5 require extra interaction on scan


schlosshogwarts:
**Home Assistant Android app version(s):*2026.6.5-full*

**Android version(s):*16 BP2A.250605.031.A3*

**Device model(s):*Xiaomi14; Xiaomi17T Pro*

**Home Assistant version:*Core 2026.6.3; Supervisor 2026.06.2; Operating System 17.3*

**Last working Home Assistant release (if known):*on Companion App Version 2026.6.0*

**Description of problem, include YAML if issue is related to notifications:*
**Summary:**
NFC tags written with Android Companion App v2026.6.5 no longer trigger Home Assistant automations directly. Instead, scanning the tag shows a "Open link" confirmation dialog before the automation runs. Tags written with iOS Companion App v2026.6.0 do not have this issue and trigger automations immediately without any extra interaction.

**Steps to reproduce:**
1. Write an NFC tag using Android Companion App v2026.6.5
2. Scan the tag with an Android device
3. A dialog appears asking to confirm opening a Home Assistant link

**Expected behavior:**
Scanning the tag triggers the Home Assistant automation immediately, without any confirmation dialog – as was the case with tags written prior to this update.

**Actual behavior:**
A "Open link" confirmation dialog appears on every scan, requiring an additional tap before the automation executes.

**Additional context:**
- Tags written with Android Companion App before the v2026.6.5 update (approximately early June 2026) still work correctly and trigger automations without a dialog
- Tags written with iOS Companion App v2026.6.0 also work correctly
- Only tags written with Android v2026.6.5 exhibit this behavior
- This suggests the NFC tag format or URI scheme written to the tag changed between versions
*

**Companion App Logs:**

```

```

**Screenshot or video of problem:**

**Additional information:*
I am not sure if I am right here with that and I hope you understand all of it. Thank you for reading this and maybe fixing this little bug. Thanks!*

TimoPtr:
iOS latest release also has this dialog. It's on purpose, when you see the dialog at the bottom you have a button to always approve if you don't want to see it next time. It's to protect against unexpected tag scanning.

schlosshogwarts:
> iOS latest release also has this dialog. It's on purpose, when you see the dialog at the bottom you have a button to always approve if you don't want to see it next time. It's to protect against unexpected tag scanning.

[...]

PR #7285, opened 2026-07-31
Adjust NFC writing AAR records to better fit tag size

jpelgrom:
## Summary

When writing NFC tags, the app tries to [include Android Application Records (AAR)](https://developer.android.com/develop/connectivity/nfc/nfc#aar) for `io.homeassistant.companion.android` and `io.homeassistant.companion.android.minimal` alongside the URL. If those don't fit on the tag, it falls back to **no** AARs to reduce the message size. This PR:
- adjusts behavior so that when all AARs don't fit on the tag, it will fall back to one AAR with the current application ID first*, before falling back to no AARs
- updates behavior for tags that supported NDEF records but weren't already formatted to also try messages with less/no AARs when formatting, instead of immediately failing

This might help with later comments in #7095 as some users reported manually writing one application record to the tag as a workaround/fix.

For the functions that were touched documentation (KDoc) and tests were added.

## Checklist

- [x] AI assistance was used for this contribution. _To figure out JUnit 4 doesn't support parametrized tests per function and what was the easiest alternative._

## Screenshots
n/a

n/a, technical implementation detail

## Any other notes

Note that when testing the application ID contains the `.debug` suffix. With the previous implementation that suffix would never be written to NFC tags, as it either had the production IDs hardcoded or none. With these changes, in the situation with 1 AAR you get the current application ID which will contain `.debug` if using a debug variant.

\* Related: the [iOS companion app implementation](https://github.com/home-assistant/iOS/blob/12452d630a7dc5fe6b3c95615ecdc5a4051c7378/Sources/App/Settings/NFC/iOSTagManager.swift#L110) always writes one AAR for `io.homeassistant.companion.android`.

### DIFF
--- a/app/src/main/kotlin/io/homeassistant/companion/android/nfc/NFCUtil.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/nfc/NFCUtil.kt
@@ -100,15 +100,13 @@ object NFCUtil {
     @Throws(IllegalArgumentException::class, IOException::class, Exception::class)
     private fun writeMessageToTag(nfcMessages: List<NdefMessage>, tag: Tag?): Boolean {
         val nDefTag = Ndef.get(tag)
-
-        nDefTag?.let {
+        nDefTag?.use {
             it.connect()
             val messageToWrite = nfcMessages.firstOrNull { message ->
                 message.toByteArray().size <= it.maxSize
             } ?: throw IllegalArgumentException("Message is too large")
             return if (it.isWritable) {
                 it.writeNdefMessage(messageToWrite)
-                it.close()
                 // Message is written to tag
                 true
             } else {
@@ -121,15 +119,16 @@ object NFCUtil {
             var caughtException: IOException? = null
             // Tag wasn't Ndef yet, so we don't know the size. Try all messages until the last one fails.
             for (message in nfcMessages) {
-                try {
-                    it.connect()
-                    it.format(message)
-                    it.close()
-                    // The data is written to the tag
-                    return true
-                } catch (e: IOException) {
-                    // Failed to format tag with message, try next
-                    caughtException = IOException("Failed to format tag", e)
+                it.use { formatableTag ->
+                    try {
+                        formatableTag.connect()
+                        formatableTag.format(message)
+                        // The data is written to the tag
+                        return true
+                    } catch (e: IOException) {
+                        // Failed to format tag with message, try next
+                        caughtException = IOException("Failed to format tag", e)
+                    }
                 }
             }
             caughtException?.let { e -> throw e }
--- a/app/src/main/kotlin/io/homeassistant/companion/android/nfc/NFCUtil.kt
+++ b/app/src/main/kotlin/io/homeassistant/companion/android/nfc/NFCUtil.kt
@@ -46,14 +46,14 @@ object NFCUtil {
     @Throws(IllegalArgumentException::class, IOException::class, Exception::class)
     fun createNFCMessage(url: String, intent: Intent?): Boolean {
         val nfcRecord = NdefRecord.createUri(url)
-        val applicationRecords = BuildConfig.APPLICATION_IDS.map {
+        val applicationFlavorsRecords = BuildConfig.APPLICATION_IDS.map {
             NdefRecord.createApplicationRecord(it)
         }
-        val packageRecord = NdefRecord.createApplicationRecord(BuildConfig.APPLICATION_ID)
+        val thisApplicationRecord = NdefRecord.createApplicationRecord(BuildConfig.APPLICATION_ID)
 
         val nfcMessages = listOf(
-            NdefMessage(arrayOf(nfcRecord) + applicationRecords),
-            NdefMessage(arrayOf(nfcRecord, packageRecord)),
+            NdefMessage(arrayOf(nfcRecord) + applicationFlavorsRecords),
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/io/homeassistant/companion/android/nfc/NFCUtil.kt
```diff
@@ -65,43 +82,60 @@ object NFCUtil {
...
-            if (it.maxSize < nfcMessage.toByteArray().size) {
-                messageToWrite = fallbackMessage
-            }
-            if (it.maxSize < fallbackMessage.toByteArray().size) {
-                // Message to large to write to NFC tag
-                throw Exception("Message is too large")
-            }
+            val messageToWrite = nfcMessages.firstOrNull { message ->
+                message.toByteArray().size <= it.maxSize
+            } ?: throw IllegalArgumentException("Message is too large")
```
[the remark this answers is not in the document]
jpelgrom: Not strictly related to the changes in this PR but I'll address it

app/src/test/kotlin/io/homeassistant/companion/android/nfc/NFCUtilTest.kt
```diff
@@ -0,0 +1,193 @@
...
+)
+
+/**
+ * Tests for writing tags with NFCUtil.
+ *
+ * This test class uses Robolectric (JUnit 4) because anything NFC related and its use of [Uri]
+ * are Android framework classes.
+ */
+@RunWith(RobolectricTestRunner::class)
+class NFCUtilTest {
```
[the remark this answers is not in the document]
jpelgrom: I don't believe it's needed as we're not interacting with any application classes or Hilt here, running as a unit test similar to UriExtensionsTest.

app/src/main/kotlin/io/homeassistant/companion/android/nfc/NFCUtil.kt
```diff
@@ -30,18 +30,35 @@ object NFCUtil {
...
-    @Throws(Exception::class)
+    /**
+     * Write the given [url] to the tag in [intent].
+     *
+     * @param url URL to write to the tag.
+     * @param intent Tag to write to, in an intent which holds the extra [NfcAdapter.EXTRA_TAG].
+     *
+     * @throws IllegalArgumentException if the url doesn't fit on the tag.
+     * @throws IOException if the tag is NDEF and the url would fit, but it cannot be written to.
+     * @throws Exception for other tag operation exceptions.
```
TimoPtr, line 41: I think this one is way too broad, but since we might not change the underlying I'm ok keeping it there, but most consumer would just catch Exception with a doc like this.

jpelgrom: I'm aware but didn't want to change the other code too much. Docs could mention all possible exceptions but I'm not sure if that is helpful in this PR if we're not going to change the activity which is using it.

app/src/test/kotlin/io/homeassistant/companion/android/nfc/NFCUtilTest.kt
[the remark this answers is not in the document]
TimoPtr: The issue with this is that it is using the HomeAssistantApplication and because of that it enables the StrictMode for instance? In the past it back fire at us because it leaked on other tests if I remember right.

app/src/test/kotlin/io/homeassistant/companion/android/nfc/NFCUtilTest.kt
```diff
@@ -0,0 +1,193 @@
...
+        val ndefMock = mockNdef(messageSlot)
+
+        // Loop through cases as JUnit 4 can only parametrize the entire class
+        tagCases.forEach { (tagSize, expectedNumberOfRecords) ->
+            every { ndefMock.maxSize } returns tagSize
+            messageSlot.clear()
+
+            val success = NFCUtil.createNFCMessage(TAG_URL, tagDiscoveredIntent)
+
+            assertEquals(true, success)
```
TimoPtr: Any reason to not use assertTrue?

app/src/test/kotlin/io/homeassistant/companion/android/nfc/NFCUtilTest.kt
```diff
@@ -0,0 +1,193 @@
...
+                if (messageSlot.captured.toByteArray().size <= tagSize) {
+                    Unit
+                } else {
+                    throw IOException()
+                }
+            }
+
+            val success = NFCUtil.createNFCMessage(TAG_URL, tagDiscoveredIntent)
+
+            assertEquals(true, success)
```
TimoPtr: same here

app/src/test/kotlin/io/homeassistant/companion/android/nfc/NFCUtilTest.kt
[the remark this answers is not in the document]
jpelgrom: OK, adjusted to be safe, also on UriExtensionsTest as that seems to be the only other usage like mine.

app/src/test/kotlin/io/homeassistant/companion/android/nfc/NFCUtilTest.kt
jpelgrom: Inexperience and confusing imports :)

TimoPtr (approved):
Thanks for addressing the comments.

Issue #6308, opened 2026-01-22
Rework UI of MediaPlayerControlsWidgetConfigureActivity to compose and Material3


TimoPtr:
### Description

This screen is using XML and we should migrate it to compose and Material3

oungsi2000:
@TimoPtr Hello! can I work on this?

Issue #6306, opened 2026-01-22
Rework UI of CameraWidgetConfigureActivity to compose and Material3


TimoPtr:
### Description

This screen is using XML and we should migrate it to compose and Material3

### Additional context

Use https://github.com/home-assistant/android/issues/6307 as reference on how it should be implemented in terms of architecture.

PR #7288, opened 2026-08-01
Migrate template widget configuration to Compose and Material3

### LINKED ISSUE #6304: Rework UI of TemplateWidgetConfigureActivity to compose and Material3
### Description

This screen is using XML and we should migrate it to compose and Material3

### Additional context

Use https://github.com/home-assistant/android/issues/6307 as reference on how it should be implemented in terms of architecture.

su7ri:
## Summary

Migrates `TemplateWidgetConfigureActivity` from a legacy XML/View Binding
layout to Jetpack Compose with Material 3, following the same pattern
established for the entity widget in #7007 (referenced from #6307 as the
example to follow for this series of widget migrations).

`TemplateWidgetConfigureActivity.kt` is now a thin wrapper (~60 lines,
down from ~250) around:
- `TemplateWidgetConfigureState`: immutable UI state, including a
  `TemplatePreview` sealed type (Empty/Rendered/Error) replacing the old
  single TextView that mixed rendered output, error messages, and the
  "blank" placeholder.
- `TemplateWidgetConfigureViewModel`: the restore/save/render logic,
  moved out of the Activity into a Hilt ViewModel with `StateFlow`.
- `TemplateWidgetConfigureScreen`: the Compose UI, reusing the shared
  `WidgetBackgroundTypeDropdown` / `WidgetTextColorDropdown` composables.

The live template preview still renders basic HTML from the server
response; the HTML→AnnotatedString conversion follows the same approach
already used on the Wear OS template tile screen
(`SettingsWearTemplateTile.kt`), duplicated here rather than extracted
to keep this PR focused on the widget migration. Happy to extract it to
a shared util in a follow-up if maintainers prefer.

Closes #6304

## Checklist

## Screenshots

image

## Any other notes

Part of the ongoing effort to migrate widget configuration screens to
Compose/Material 3 (see #6303, #6305, #6306, #6308, and others).


app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,305 @@
...
+        label = stringResource(commonR.string.server_select),
+        placeholder = stringResource(commonR.string.server_select),
+        modifier = Modifier.formControlWidth(),
+        enabled = items.isNotEmpty(),
+    )
+}
+
+@Composable
+private fun TemplateSection(template: String, preview: TemplatePreview, onTemplateChanged: (String) -> Unit) {
+    HATextField(
```
jpelgrom, line 187: One line for a template is a bit short. Consider it increasing to a couple of lines.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,305 @@
...
+        value = template,
+        onValueChange = onTemplateChanged,
+        label = { Text(stringResource(commonR.string.template)) },
+        placeholder = { Text(stringResource(commonR.string.template_widget_default)) },
+        modifier = Modifier.formControlWidth(),
+    )
+
+    Text(text = preview.toAnnotatedString(), modifier = Modifier.formControlWidth())
```
jpelgrom: This is not very clearly a preview. I know it's similar to the previous screen but in that one it at least looked slightly different. 

Suggestion: put it in a box similar to the 'Home Assistant is your default assistant' in Assist settings, with a small 'Preview' icon/label.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,283 @@
...
+
+        _state.update { it.copy(selectedServerId = serverId) }
+        val template = _state.value.template
+        if (template.isNotEmpty()) renderTemplate(template, serverId)
+    }
+
+    fun onTemplateChanged(value: String) {
+        _state.update { it.copy(template = value) }
+
+        if (value.isEmpty()) {
```
jpelgrom: I think this could be `isBlank()`, rendering a template of only whitespace makes no sense either

jpelgrom (commented):
Thanks for picking this up!

Could you add a screenshot test as well? You already have a Compose preview so it shouldn't be too difficult :)

> The live template preview still renders basic HTML from the server response; the HTML→AnnotatedString conversion follows the same approach already used on the Wear OS template tile screen (SettingsWearTemplateTile.kt), duplicated here rather than extracted to keep this PR focused on the widget migration. Happy to extract it to a shared util in a follow-up if maintainers prefer.

As it is only ~30 lines you can extract it into a util file in this PR instead of duplicating code.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
jpelgrom: (Your test title already calls it blank but uses an empty string...)

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,305 @@
...
+    renderedSpanned.getSpans(0, renderedSpanned.length, CharacterStyle::class.java).forEach { span ->
+        val start = renderedSpanned.getSpanStart(span)
+        val end = renderedSpanned.getSpanEnd(span)
+        when (span) {
+            is AbsoluteSizeSpan -> addStyle(SpanStyle(fontSize = span.size.sp), start, end)
+            is ForegroundColorSpan -> addStyle(SpanStyle(color = Color(span.foregroundColor)), start, end)
+            is RelativeSizeSpan -> {
+                val defaultSize = 12
+                addStyle(SpanStyle(fontSize = (span.sizeChange * defaultSize).sp), start, end)
+            }
```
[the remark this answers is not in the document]
su7ri: Fixed — extracted into a shared util/compose/HtmlText.kt (per jpelgrom's suggestion elsewhere in this PR to avoid duplicating this logic), using exactly this regex/constant extraction.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Added a minLines param to the shared HATextField and set it to 3 for the template field.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
su7ri: Switched to isBlank() everywhere the template is checked (restore, server change, text change), and added a test for whitespace-only input.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Wrapped it in HASettingsCard with a Preview icon/label, matching the default-assistant card pattern in Assist settings.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,288 @@
...
+        value = template,
+        onValueChange = onTemplateChanged,
+        label = { Text(stringResource(commonR.string.template)) },
+        placeholder = { Text(stringResource(commonR.string.template_widget_default)) },
+        minLines = TEMPLATE_FIELD_MIN_LINES,
+        modifier = Modifier.formControlWidth(),
+    )
+
+    TemplatePreviewCard(preview = preview)
+}
```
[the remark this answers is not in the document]
su7ri: Fixed — wrapped the two composables in a Column so TemplateSection emits from a single source.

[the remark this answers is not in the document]
su7ri: Fixed in e440398 — wrapped the two composables in a Column so TemplateSection emits from a single source.

app/src/screenshotTestFullDebug/reference/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenshotTest/TemplateWidgetConfigureContent empty template_tablet_2f22c4ea_0.png
jpelgrom, line 1: You can see how screenshot testing is useful here - the preview card doesn't work correctly in dark mode, with the icon and text being dark on dark

su7ri: Fixed — explicitly set the icon/text colors via LocalHAColorScheme (colorTextPrimary/colorTextSecondary), following the same pattern as DefaultAssistantCard in Assist settings. Regenerated the reference screenshots too.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+            onValueChange = onTemplateChanged,
+            label = { Text(stringResource(commonR.string.template)) },
+            placeholder = { Text(stringResource(commonR.string.template_widget_default)) },
+            minLines = TEMPLATE_FIELD_MIN_LINES,
+            modifier = Modifier.formControlWidth(),
+        )
+
+        TemplatePreviewCard(preview = preview)
+    }
+}
```
jpelgrom: Avoid the nested column

```suggestion
private fun ColumnScope.TemplateSection(
    template: String,
    preview: TemplatePreview,
    onTemplateChanged: (String) -> Unit,
) {
    HATextField(
        value = template,
        onValueChange = onTemplateChanged,
        label = { Text(stringResource(commonR.string.template)) },
        placeholder = { Text(stringResource(commonR.string.template_widget_default)) },
        minLines = TEMPLATE_FIELD_MIN_LINES,
        modifier = Modifier.formControlWidth(),
    )

    TemplatePreviewCard(preview = preview)
}
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,297 @@
+package io.homeassistant.companion.android.widgets.template
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
```
jpelgrom, line 4:
```suggestion
import androidx.compose.foundation.layout.Column
import androidx.compose.foundation.layout.ColumnScope
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+            modifier = Modifier.formControlWidth(),
+        )
+    }
+}
+
+private fun Modifier.formControlWidth(): Modifier = this
+    .widthIn(max = MaxButtonWidth)
+    .fillMaxWidth()
+
+private const val TEMPLATE_FIELD_MIN_LINES = 3
```
jpelgrom: Put constant at the top of the file before the first function

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+import androidx.compose.ui.unit.dp
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import io.homeassistant.companion.android.common.R as commonR
+import io.homeassistant.companion.android.common.compose.composable.HAAccentButton
+import io.homeassistant.companion.android.common.compose.composable.HADropdownItem
+import io.homeassistant.companion.android.common.compose.composable.HADropdownMenu
+import io.homeassistant.companion.android.common.compose.composable.HASettingsCard
+import io.homeassistant.companion.android.common.compose.composable.HATextField
+import io.homeassistant.companion.android.common.compose.composable.HATopBar
+import io.homeassistant.companion.android.common.compose.theme.HADimens
```
jpelgrom, line 43:
```suggestion
import io.homeassistant.companion.android.common.compose.theme.HADimens
import io.homeassistant.companion.android.common.compose.theme.HASize
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+        )
+    }
+}
+
+private fun Modifier.formControlWidth(): Modifier = this
+    .widthIn(max = MaxButtonWidth)
+    .fillMaxWidth()
+
+private const val TEMPLATE_FIELD_MIN_LINES = 3
+private val HAPreviewIconSize = 16.dp
```
jpelgrom:
```suggestion
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+                horizontalArrangement = Arrangement.spacedBy(HADimens.SPACE2),
+            ) {
+                Icon(
+                    imageVector = Icons.Default.Visibility,
+                    contentDescription = null,
+                    tint = colorScheme.colorTextSecondary,
+                    modifier = Modifier.size(HAPreviewIconSize),
+                )
+                Text(
+                    text = stringResource(commonR.string.template_preview_label),
```
jpelgrom: I think this would look slightly better if it's the size of a title in a dropdown, what do you think?

Image


```suggestion
                    modifier = Modifier.size(HASize.M),
                )
                Text(
                    text = stringResource(commonR.string.template_preview_label),
                    style = HATextStyle.BodyMedium,
```

Don't forget to update screenshot tests if you accept.

If you don't agree update the size to use one of our shared constants:


```suggestion
                    modifier = Modifier.size(HASize.L),
                )
                Text(
                    text = stringResource(commonR.string.template_preview_label),
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import io.homeassistant.companion.android.common.R as commonR
+import io.homeassistant.companion.android.common.compose.composable.HAAccentButton
+import io.homeassistant.companion.android.common.compose.composable.HADropdownItem
+import io.homeassistant.companion.android.common.compose.composable.HADropdownMenu
+import io.homeassistant.companion.android.common.compose.composable.HASettingsCard
+import io.homeassistant.companion.android.common.compose.composable.HATextField
+import io.homeassistant.companion.android.common.compose.composable.HATopBar
+import io.homeassistant.companion.android.common.compose.theme.HADimens
+import io.homeassistant.companion.android.common.compose.theme.HAThemeForPreview
```
jpelgrom, line 46: If you agree with https://github.com/home-assistant/android/pull/7288/changes#r3714512883

```suggestion
import io.homeassistant.companion.android.common.compose.theme.HATextStyle
import io.homeassistant.companion.android.common.compose.theme.HAThemeForPreview
```

jpelgrom (commented):
Almost there!

Compared to the old activity text alignment via HTML no longer shows in the preview but as it's quite an advanced feature and still working in the widget it's fine as it is right now so we can avoid complicating it even more. If more users show up with complicated HTML examples we can always fallback to a classic view which takes the HTML.

|Before|After|
|---|---|
|Image|Image(the`p`tag is adding padding at the bottom)|

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+            ) {
+                Icon(
+                    imageVector = Icons.Default.Visibility,
+                    contentDescription = null,
+                    tint = colorScheme.colorTextSecondary,
+                    modifier = Modifier.size(HAPreviewIconSize),
+                )
+                Text(
+                    text = stringResource(commonR.string.template_preview_label),
+                    color = colorScheme.colorTextSecondary,
```
TimoPtr, line 219: It misses a textStyle probabbly BodyMedium.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+                )
+            }
+            Text(text = preview.toAnnotatedString(), color = colorScheme.colorTextPrimary)
```
TimoPtr: Same here

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/HtmlText.kt
```diff
@@ -0,0 +1,56 @@
...
+
+private val LINE_BREAK_REGEX = "(\r\n|\r|\n)|(\\\\r\\\\n|\\\\r|\\\\n)".toRegex()
+private const val DEFAULT_RELATIVE_SIZE_SP = 12
+
+/**
+ * Converts HTML, as returned by a rendered Home Assistant template, into an [AnnotatedString] by
+ * translating the [android.text.Spanned] styles produced by [HtmlCompat.fromHtml] into Compose
+ * [SpanStyle]s. Shared by every screen that displays a rendered template.
+ */
+fun parseHtml(renderedText: String): AnnotatedString = buildAnnotatedString {
```
TimoPtr, line 28: We could write unit tests for this.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+                onTemplateChanged = onTemplateChanged,
+            )
+            HATextField(
+                value = state.textSize,
+                onValueChange = onTextSizeChanged,
+                label = { Text(stringResource(commonR.string.widget_text_size_label)) },
+                keyboardOptions = KeyboardOptions(keyboardType = KeyboardType.Number),
+                maxLines = 1,
+                modifier = Modifier.formControlWidth(),
+            )
```
TimoPtr: Move this part into the AppearanceSection

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
```diff
@@ -0,0 +1,297 @@
...
+    if (!showServerSelector) return
+
+    HADropdownMenu(
+        items = items,
+        selectedKey = selectedServerId,
+        onItemSelected = onServerSelected,
+        label = stringResource(commonR.string.server_select),
+        placeholder = stringResource(commonR.string.server_select),
+        modifier = Modifier.formControlWidth(),
+        enabled = items.isNotEmpty(),
```
TimoPtr: I'm not sure about keeping this logic, the viewState should old the UI logic with the show/not show `showServerSelector`

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
TimoPtr:
```suggestion
			val content = remember (preview) { preview.toAnnotatedString() }
            Text(text = content, color = colorScheme.colorTextPrimary)
```
To avoid recomputing on recomposition.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureState.kt
```diff
@@ -0,0 +1,51 @@
...
+    val dynamicColorAvailable: Boolean = false,
+    val isUpdateWidget: Boolean = false,
+) {
+    val showServerSelector = serversDropdownItems.size > 1 ||
+        serversDropdownItems.none { it.key == selectedServerId }
+
+    // Guards against saving a template that hasn't been (re-)validated yet: without
+    // `!isRenderingPreview`, editing an already-valid template would keep the action enabled
+    // using the *previous* render's result while the new one is still in flight.
+    val isActionEnabled = preview is TemplatePreview.Rendered && !isRenderingPreview
```
TimoPtr: You probably want also to gate on a valid text size.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureState.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    @StringRes
+    val actionButtonLabel = if (isUpdateWidget) commonR.string.update_widget else commonR.string.add_widget
+
+    val textSizeOrDefault: Float
+        get() = textSize.toFloatOrNull()?.takeIf { it.isFinite() && it > 0 } ?: DEFAULT_TEXT_SIZE.toFloat()
+}
+
+internal const val DEFAULT_TEXT_SIZE = "12"
```
TimoPtr:
```suggestion
private const val DEFAULT_TEXT_SIZE = "12"
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,284 @@
...
+                    )
+                }
+            }
+        }
+    }
+
+    /**
+     * Restores the configuration of an existing widget, or falls back to the active server for a new one.
+     */
+    private suspend fun restoreConfiguration() {
```
TimoPtr: Put private functions at the bottom of the files even if it's used in the constructor.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,284 @@
...
+
+    fun onTextSizeChanged(value: String) {
+        _state.update { it.copy(textSize = value.filter(Char::isDigit)) }
+    }
+
+    fun onBackgroundTypeSelected(backgroundType: WidgetBackgroundType) {
+        _state.update { it.copy(selectedBackgroundType = backgroundType) }
+    }
+
+    internal fun onTextColorSelected(colorHex: String) {
```
TimoPtr:
```suggestion
    fun onTextColorSelected(colorHex: String) {
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,284 @@
...
+        _state.update { it.copy(textColorHex = colorHex) }
+    }
+
+    /** Renders [template] against [serverId], cancelling any render already in flight. */
+    private fun renderTemplate(template: String, serverId: Int) {
```
TimoPtr: move at the end

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
```diff
@@ -0,0 +1,284 @@
...
+            Intent(context, TemplateWidget::class.java).apply {
+                action = BaseWidgetProvider.UPDATE_WIDGETS
+            },
+        )
+    }
+
+    /**
+     * Builds the widget to persist from the current configuration, or null when it is incomplete.
+     */
+    internal suspend fun getPendingDaoEntity(): TemplateWidgetEntity? {
```
TimoPtr:
```suggestion
    private suspend fun getPendingDaoEntity(): TemplateWidgetEntity? {
```

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenshotTest.kt
```diff
@@ -0,0 +1,89 @@
...
+import com.android.tools.screenshot.PreviewTest
+import io.homeassistant.companion.android.common.R as commonR
+import io.homeassistant.companion.android.common.compose.composable.HADropdownItem
+import io.homeassistant.companion.android.common.compose.theme.HAThemeForPreview
+import io.homeassistant.companion.android.database.widget.WidgetBackgroundType
+import io.homeassistant.companion.android.util.compose.HAPreviews
+import io.homeassistant.companion.android.util.previewServer1
+import io.homeassistant.companion.android.util.previewServer2
+
+class TemplateWidgetConfigureScreenshotTest {
```
TimoPtr, line 15: One of the test should not have the server selector to show it works.

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenshotTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
```
TimoPtr, line 16: Also could you use a more advance template like Joris did? that contains some new lines and color?

app/src/screenshotTestFullDebug/reference/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenshotTest/TemplateWidgetConfigureContent template error_phone_landscape_9e00b29d_0.png
TimoPtr, line 1: @jpelgrom Would it make more obvious and more real to center the rendered string? (So default from BodyMedium)

app/src/test/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModelTest.kt
```diff
@@ -0,0 +1,225 @@
...
+
+    @Test
+    fun `Given a whitespace-only template when changed then the preview goes back to empty`() = runTest {
+        coEvery { integrationRepository.renderTemplate("{{ 1 }}", emptyMap()) } returns "1"
+        val viewModel = createViewModel()
+        advanceUntilIdle()
+        viewModel.onTemplateChanged("{{ 1 }}")
+        advanceUntilIdle()
+
+        viewModel.onTemplateChanged("   ")
```
TimoPtr, line 184: You should also verify that it doesn't invoke render on the IntegrationRepo.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModelTest.kt
```diff
@@ -0,0 +1,225 @@
...
+    fun `Given a template when changed then it is rendered against the selected server`() = runTest {
+        coEvery { integrationRepository.renderTemplate("{{ 1 + 1 }}", emptyMap()) } returns "2"
+        val viewModel = createViewModel()
+        advanceUntilIdle()
+
+        viewModel.onTemplateChanged("{{ 1 + 1 }}")
+        advanceUntilIdle()
+
+        assertEquals(TemplatePreview.Rendered("2"), viewModel.state.value.preview)
+        assertTrue(viewModel.state.value.isActionEnabled)
```
TimoPtr, line 105: Verify it invoke the template rendering on the IntegrationRepo

app/src/test/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModelTest.kt
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    @Test
+    fun `Given an invalid configuration when configuration is saved then it is rejected`() = runTest {
+        val viewModel = createViewModel()
+        advanceUntilIdle()
+
+        assertFalse(viewModel.updateWidgetConfiguration())
+    }
+
```
TimoPtr, line 242: You could add a test that check the cancelation of the rendering when a new value arrives before the first one finish to render.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModelTest.kt
```diff
@@ -0,0 +1,225 @@
...
+        template = template,
+        textSize = 21F,
+        lastUpdate = "on",
+        backgroundType = WidgetBackgroundType.TRANSPARENT,
+        textColor = BLACK_HEX,
+    )
+
+    companion object {
+        /** Hex of `colorWidgetButtonLabelBlack`, which is what the widget persists. */
+        private const val BLACK_HEX = "#3A3A3A"
```
TimoPtr: Put it at the top of the file, no need of a companion object.

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HATextField.kt
```diff
@@ -44,6 +44,8 @@ import io.homeassistant.companion.android.common.compose.theme.MaxButtonWidth
  * [KeyboardOptions.imeAction]
  * @param maxLines the maximum height in terms of maximum number of visible lines. If [singleLine]
  * is set to `true`, this value will be ignored
+ * @param minLines the minimum height in terms of minimum number of visible lines. Ignored if
+ * [singleLine] is `true`
```
TimoPtr:
```suggestion
 * @param minLines the minimum height in terms of minimum number of visible lines. If [singleLine]
 * is set to `true`, this value will be ignored
```
To match the other param

TimoPtr:
I forgot we should have compose tests that verify that the callbacks are properly triggered and also the content is adjusting properly based on the UI State like showing the server selector, the text size error, the action button has the right text and it is disabled/enabled on the right conditions.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
jpelgrom: Also resolved if you agree with the first suggestion of https://github.com/home-assistant/android/pull/7288#discussion_r3714512883

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
jpelgrom: FWIW this matches EntityWidgetConfigureScreen. Small logic like enabled or not based on the view state is probably fine? Show/not show the entire server selector in the view state is a couple of lines above this one.

app/src/screenshotTestFullDebug/reference/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenshotTest/TemplateWidgetConfigureContent template error_phone_landscape_9e00b29d_0.png
jpelgrom: Yes that could be nice and should only take one line. I don't want to do a full redesign in this PR though or 100% match the widget, that is out of scope.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
TimoPtr: I think it's better to not have it, if I let it in the EntityWidgetConfigureScreen I think it is by mistakes. Here I feel like the logic is different between the show and the enabled.

jpelgrom: In that case show selector = enabled and we should use the default value for `enabled` (`true`).

```suggestion
```

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Applied as suggested in e578ca5.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Applied as suggested in e578ca5.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Applied in bc72086.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Applied in e578ca5 (replaced with HASize.M).

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Applied — HATextStyle.BodyMedium in e578ca5.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Same fix applies here too, e578ca5.

app/src/main/kotlin/io/homeassistant/companion/android/util/compose/HtmlText.kt
su7ri: Added HtmlTextTest.kt covering plain text, bold/italic (incl. nested), underline, font color, <big>/<small> sizing, and both real and literal-escaped line breaks. Pushed in 8c80041.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Moved in e578ca5.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Removed, now using the default (true) for enabled as agreed. e578ca5.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Applied, but slightly differently than the suggestion — wrapping the whole toAnnotatedString() call in remember doesn't compile since it internally calls stringResource() (for the Empty/Error branches) which @DisallowComposableCalls blocks inside remember's lambda. Instead only the actual parseHtml() call is memoized on text, which is the part worth avoiding on recomposition anyway. c0a1e5e.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureState.kt
su7ri: Good catch — isActionEnabled now also requires a valid text size, so an emptied field can't silently save with the default. Added a test for it too. c0a1e5e.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureState.kt
su7ri: Applied — it's only used within this file. d3c5165.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
su7ri: Moved both restoreConfiguration() and renderTemplate() (same private-function convention) to the bottom, after the public/internal API. d3c5165.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
su7ri: Applied. 71d3de4.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
su7ri: Already moved in d3c5165 (previous commit, same fix as restoreConfiguration).

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModel.kt
su7ri: Applied. 7b81b8b

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenshotTest.kt
su7ri: Added TemplateWidgetConfigureContent single server. 7b81b8b.

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenshotTest.kt
su7ri: Used a rendered preview with a color span, bold, and a line break. I couldn't see the exact reference image, so I went with a reasonable equivalent — happy to adjust if you had something specific in mind. 7b81b8b.

common/src/main/kotlin/io/homeassistant/companion/android/common/compose/composable/HATextField.kt
su7ri: Applied. 7b81b8b.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModelTest.kt
su7ri: Already done in bc72086.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModelTest.kt
su7ri: Already covered — added in an earlier commit (e578ca5).

app/src/test/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModelTest.kt
su7ri: Already covered — added in an earlier commit (e578ca5).

app/src/test/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureViewModelTest.kt
su7ri: Added. 7b81b8b.

app/src/screenshotTestFullDebug/reference/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenshotTest/TemplateWidgetConfigureContent template error_phone_landscape_9e00b29d_0.png
su7ri: Sounds good, leaving as-is for this PR.

TimoPtr:
Thanks for taking the time to contribute.

This contribution looks like it doesn't follow our [AI policy](https://developers.home-assistant.io/docs/ai_policy), so I'm flagging it here. If I've got this wrong, please let me know.

We won't close it this time because @jpelgrom and myself already reviewed it, but please keep this in consideration for future contributions to this PR or future PRs.

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenshotTest.kt
jpelgrom: Instead of adding another test, adjust an existing test to only have one server to save on screenshots used.

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenshotTest.kt
jpelgrom: You should update the screenshots generated by tests as well.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenTest.kt
```diff
@@ -0,0 +1,298 @@
...
+                        onTextColorSelected = { textColor = it },
+                        onActionClick = { actionClicked = true },
+                    )
+                }
+            }
+            dsl()
+        }
+    }
+
+    private companion object {
```
jpelgrom: This doesn't need to be in a companion object

su7ri:
> Thanks for taking the time to contribute.
> 
> This contribution looks like it doesn't follow our [AI policy](https://developers.home-assistant.io/docs/ai_policy), so I'm flagging it here. If I've got this wrong, please let me know.
> 
> We won't close it this time because @jpelgrom and myself already reviewed it, but please keep this in consideration for future contributions to this PR or future PRs.

Thanks for flagging this. I'll keep using AI as a tool since I'm still learning this codebase, but I'll make sure to actually understand and review each change before it goes up, and commit things myself instead of letting the tool push directly.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Already there — added alongside HATextStyle.BodyMedium.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreen.kt
su7ri: Went with the first option — HASize.M for the icon, HATextStyle.BodyMedium for both preview texts. Applied in e578ca5.

app/src/test/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenTest.kt
su7ri: Moved to top-level private vals instead. 79dc55d98.

su7ri:
Hey, just noticed the workflow approval has been pending for about a week — could someone take a look when you get a chance?"

jpelgrom:
> Hey, just noticed the workflow approval has been pending for about a week — could someone take a look when you get a chance?"

I believe you have not yet addressed all comments, which is why I didn't trigger the workflow to avoid wasting resources.

 - https://github.com/home-assistant/android/pull/7288#discussion_r3722988331
 - https://github.com/home-assistant/android/pull/7288#discussion_r3722998781

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenshotTest.kt
su7ri: Updated the reference screenshots locally and pushed. 63a2baf7e

app/src/screenshotTest/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureScreenshotTest.kt
su7ri: Updated and pushed. 6ef896fd5

su7ri: Done — removed the separate test and reused the empty template one with a single server instead. 6ef896fd5

su7ri:
> > Hey, just noticed the workflow approval has been pending for about a week — could someone take a look when you get a chance?"
> 
> I believe you have not yet addressed all comments, which is why I didn't trigger the workflow to avoid wasting resources.
> 
> * [Migrate template widget configuration to Compose and Material3 #7288 (comment)](https://github.com/home-assistant/android/pull/7288#discussion_r3722988331)
> * [Migrate template widget configuration to Compose and Material3 #7288 (comment)](https://github.com/home-assistant/android/pull/7288#discussion_r3722998781)

All outstanding comments should be addressed now — happy to have the checks re-triggered whenever convenient.

jpelgrom:
In case you hadn't noticed, unit tests fail with your latest changes. Please fix them to continue.

su7ri:
I believe all the review comments have been addressed now. Please let me know if any further changes are needed.

TimoPtr:
For the next batch of comment, when you've finished to apply the comments put the PR in Ready for review and re-request the review on the maintainers.

app/src/main/kotlin/io/homeassistant/companion/android/widgets/template/TemplateWidgetConfigureState.kt
```diff
@@ -0,0 +1,54 @@
...
+    // using the *previous* render's result while the new one is still in flight.
+    val isActionEnabled = preview is TemplatePreview.Rendered && !isRenderingPreview && validTextSize != null
+
+    @StringRes
+    val actionButtonLabel = if (isUpdateWidget) commonR.string.update_widget else commonR.string.add_widget
+
+    val textSizeOrDefault: Float
+        get() = validTextSize ?: DEFAULT_TEXT_SIZE.toFloat()
+
+    private val validTextSize: Float?
```
TimoPtr, line 53: Let's exposed also a boolean like https://github.com/home-assistant/android/blob/ad25245be514613881100eee83023a72d55f7d2c/app/src/main/kotlin/io/homeassistant/companion/android/widgets/entity/EntityWidgetConfigureState.kt#L42 so that the UI can reflect that the size is wrong. You already handle the logic because if you empty the text size you disable the Add widget but we should also add the error message like on the other screens.

TimoPtr (commented):
Remove `widget_text_size_template` from the translations since it's not used anymore.

Missing entry in the `ChangelogContent.kt` (and generate screenshots).

I'm going to address this comments and push to your branch to have a final review from @jpelgrom

TimoPtr:
@jpelgrom it's ready for you to review :)

app/detekt-baseline-fullRelease.xml
```diff
@@ -428,6 +426,7 @@
     <ID>ReturnCount:SettingsWearViewModel.kt:SettingsWearViewModel$private suspend fun updateServer</ID>
     <ID>ReturnCount:SsidViewModel.kt:SsidViewModel$fun addHomeWifiSsid: Boolean</ID>
     <ID>ReturnCount:TLSWebViewClient.kt:TLSWebViewClient$private fun getActivity: Activity?</ID>
+    <ID>ReturnCount:TemplateWidgetConfigureViewModel.kt:TemplateWidgetConfigureViewModel$@SuppressLint("NewApi") // The API 26 requirement is checked below before touching the pinning APIs. suspend fun requestWidgetCreation: Boolean</ID>
```
TimoPtr, line 429: @jpelgrom this is the copy of the existing code that is raising this. We should make a generic method instead that's why I didn't touch it and ignored the issue.

jpelgrom (approved):
These screen changes have become very formulaic to review, in a good way. I may be missing something but then it's missed in multiple of these screens.

The template render example still isn't as advanced as I highlighted, but I think that doesn't need to be part of this PR. We can do separate screenshot tests for that.

Let's get this change in, thanks for the final push @TimoPtr !

app/detekt-baseline-fullRelease.xml
jpelgrom: That matches my comment about it being formulaic... I don't think we handle it incorrectly but need to get used to Detekt, this can be followed up in another Detekt fixes PR.

PR #7289, opened 2026-08-01
Declare sensor settings instead of creating them on read

### LINKED ISSUE #7225: Sensor settings should be declared, not created as a side effect of reading them
### Description

SensorManager.getSetting persists the default value into the DB when the setting is not stored yet. The sensor detail screen discovers which settings exist from the stored rows (getSettingsFlow), so a setting only becomes visible once some code path happens to read it. This write-on-read side effect is the root cause of #6773 and forced the workaround in #7206, where `requestSensorUpdate` calls a getter and discards the result purely so the write happens.
#7040 already removed this pattern for the sensors themselves: the repository returns the default when nothing is stored, and only an actual change is persisted. Settings should follow the same model. 

## Proposal

  - Sensor managers declare the settings each sensor offers (name, type, default, entries), similar to how available sensors are declared via getAvailableSensors.
  - The detail screen merges the declaration with the stored values, so settings are visible immediately without a DB row. 
  - getSetting returns the stored value or the declared default, without writing. Only user changes persist.                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                   
## Benefits 

- Removes the initialization workaround from #7206. 
- Fixes the same discovery bug for active_notification_count (its SETTING_INCLUDE_CONTENTS_AS_ATTRS toggle only appears after a notification callback) and for any other sensor whose settings are only read in event-driven paths. 
- Settings behave consistently with sensors after #7040.

### Additional context

_No response_

loganrosen:
## Summary

Sensor settings were previously created in the database as a side effect of reading them, which made discovery mutate state and spread default definitions across call sites. This declares each sensor's settings alongside its sensor definition and merges those declarations with persisted value and enabled overrides when settings are read.

Dynamic settings remain persisted normally, while first edits to declared settings and Bluetooth UUID initialization use transactional field-specific operations so concurrent updates cannot overwrite each other. This keeps existing user values compatible while allowing declaration metadata and defaults to evolve with the app.

Fixes #7225.

## Checklist

## Any other notes

The declaration-only settings flow was also exercised on an API 35 emulator by editing a setting through the Compose UI, force-stopping the app, and confirming the persisted override after a fresh process launch.


common/src/main/kotlin/io/homeassistant/companion/android/database/sensor/SensorDao.kt
```diff
@@ -103,10 +106,37 @@ internal interface SensorDao {
...
 
     @Query("UPDATE sensor_settings SET value = :value WHERE sensor_id = :sensorId AND name = :settingName")
-    suspend fun updateSettingValue(sensorId: String, settingName: String, value: String)
+    suspend fun updateSettingValue(sensorId: String, settingName: String, value: String): Int
+
+    @Transaction
+    suspend fun upsertSettingEnabled(setting: SensorSetting, enabled: Boolean) {
+        if (updateSettingEnabled(setting.sensorId, setting.name, enabled) == 0) {
+            add(setting.copy(enabled = enabled))
+        }
```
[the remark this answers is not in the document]
loganrosen: These helpers are both annotated with `@Transaction`, and Room generates them through `performInTransactionSuspending`. SQLite serializes write transactions, so another enabled/value helper cannot insert between this transaction’s initial `UPDATE` and fallback `REPLACE`; it runs after the first transaction commits, sees the row, and updates only its intended column. Because the reported interleaving cannot occur here, I’m leaving the implementation unchanged.

common/src/main/kotlin/io/homeassistant/companion/android/common/sensors/SensorRepositoryImpl.kt
```diff
@@ -96,19 +99,60 @@ internal class SensorRepositoryImpl @Inject constructor(
...
+    override suspend fun updateSettingValue(sensorId: String, settingName: String, value: String) {
+        settingForUpdate(sensorId, settingName)?.let { dao.upsertSettingValue(it, value) }
+    }
+    override suspend fun getOrInitializeSettingValue(
+        sensorId: String,
+        settingName: String,
+        initialValue: String,
+    ): String = settingForUpdate(sensorId, settingName)?.let {
+        dao.getOrInitializeSettingValue(it, initialValue)
+    }.orEmpty()
```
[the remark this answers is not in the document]
loganrosen: Fixed in 9a74fc7b5. The missing-definition path still triggers FailFast, but release builds now fall back to the caller-provided `initialValue` instead of an empty string. I also added regression coverage for that behavior.

TimoPtr:
@loganrosen I took the liberty to push a change on how we define a settings to avoid having to deal with strings for everything and use a dummy sealed class instead that makes things clearer. Please have a look.

app/src/test/kotlin/io/homeassistant/companion/android/sensors/ProvidedSensorSetTest.kt
```diff
@@ -33,4 +33,22 @@ class ProvidedSensorSetTest {
         assertTrue(ids.contains("car_fuel"))
         assertTrue(ids.contains("car_speed"))
     }
+
+    /**
+     * A duplicated setting name resolves inconsistently: [io.homeassistant.companion.android.common.sensors.SensorRepository]
+     * keys declarations by name and keeps the last, while `SensorManager.getSetting` keeps the first.
+     */
```
jpelgrom: This may be correct but

 - there shouldn't be duplicate combinations of sensor ID + setting name (which this test guards against)
 - behavior could be aligned

As it describes why but not the function it could be a normal comment

app/src/testFull/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManagerTest.kt
```diff
@@ -0,0 +1,45 @@
...
+                Setting.Zones("location_ham_only_enter_zone"),
+                Setting.Toggle("location_ham_zone_bt_combined", default = false),
+                Setting.Number("location_ham_trigger_range", 300),
+            ),
+            LocationSensorManager.backgroundLocation.settings,
+        )
+    }
+
+    @Test
+    fun `Given zone location sensor when inspected then minimum accuracy is declared`() {
```
jpelgrom:
```suggestion
    fun `Given zone location sensor when inspected then minimum accuracy setting is declared`() {
```

common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/BluetoothSensorManagerTest.kt
```diff
@@ -0,0 +1,152 @@
...
+            BluetoothSensorManager.bleTransmitter,
+            BluetoothSensorManager.SETTING_BLE_HOME_WIFI_ONLY,
+        )
+
+        assertEquals(0, result)
+        assertNotNull(throwableCaptured)
+    }
+
+    @Test
+    fun `Given the beacon monitor when inspected then the RSSI multiplier is its only decimal setting`() {
```
jpelgrom: For consistency with other test titles

```suggestion
    fun `Given beacon monitor when inspected then only RSSI multiplier is declared as decimal setting`() {
```

jpelgrom (commented):
There seems to be some scope creep in here with the mutex for Bluetooth, but I'll accept it.

No further comments at this time, it seems to work as described and not break anything. I noticed the setting order is different, but it was never supposed to be stable.

app/src/test/kotlin/io/homeassistant/companion/android/sensors/ProvidedSensorSetTest.kt
loganrosen: Changed this to a normal comment in d8bee7d39.

app/src/testFull/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManagerTest.kt
loganrosen: Updated the test name as suggested in d8bee7d39.

common/src/test/kotlin/io/homeassistant/companion/android/common/sensors/BluetoothSensorManagerTest.kt
loganrosen: Updated the test name for consistency in d8bee7d39.

loganrosen:
> @loganrosen I took the liberty to push a change on how we define a settings to avoid having to deal with strings for everything and use a dummy sealed class instead that makes things clearer. Please have a look.

@TimoPtr Looks good, thanks!

jpelgrom (approved):

PR #7209, opened 2026-07-18
Add Detekt static analysis

loganrosen:
## Summary

Adds Detekt static analysis across the main Gradle build and the included build logic. Module-local baselines capture the 641 existing findings, allowing CI to reject new violations without requiring unrelated cleanup in this PR. Detekt produces HTML and SARIF reports, uploads CI findings to code scanning, and includes an Android Studio run configuration.

Ktlint remains responsible for formatting and line length. The project Detekt configuration disables `MaxLineLength` and `TooGenericExceptionCaught` (which does not match the repository's current exception-handling convention), while applying targeted Compose exemptions to `FunctionNaming`, `LongParameterList`, `TooManyFunctions`, and `UnusedPrivateFunction`. Detekt still adds checks that ktlint does not provide. The current baseline includes 189 magic numbers, 114 long methods, 83 cyclomatically complex methods, 70 classes or objects with too many functions, 56 functions with excessive returns, 21 swallowed exceptions, and one `printStackTrace` call.

The integration uses pinned Detekt 2.0.0-alpha.5 because it targets Kotlin 2.4.0 and Android Gradle Plugin 9.2.1, closely matching this repository's Kotlin 2.4.10 and AGP 9.3.1. Stable Detekt 1.23.8 instead targets Kotlin 2.0.21 and AGP 8.8.1, and Detekt has not announced a stable 2.0 release date. As build-only tooling validated in CI, the alpha does not affect the shipped application.

Closes #5135.

## Checklist

- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.
- [x] I have read the [Open Home Foundation AI Policy](https://developers.home-assistant.io/docs/ai_policy).

Select exactly one option that describes AI usage in this contribution:

- [ ] I have not used AI for this contribution.
- [x] AI assistance was used for this contribution.
- [ ] AI fully generated the code for this contribution, but I've reviewed and understood it before submitting and will respond without AI during review.

## Screenshots

N/A — tooling-only change.

## Link to pull request in documentation repositories

User Documentation: N/A

[...]

PR #7298, opened 2026-08-02
Prevent emulator.wtf reruns from failing on expired artifacts

### LINKED ISSUE #7297: Prevent emulator.wtf workflow reruns from using expired artifacts
### Description

The `PR Emulator.wtf` workflow downloads `emulator-wtf-inputs` from the triggering `Pull Request` workflow run, but that artifact is retained for only one day. A partial rerun can complete successfully and trigger `PR Emulator.wtf` again without rerunning the job that produces the artifact. If the original artifact has expired, instrumentation CI fails during artifact download before any tests run.

Update the workflow so a valid input artifact is available whenever the downstream workflow runs—for example, by extending the artifact retention period or otherwise ensuring the producer runs again. Rerunning PR CI should not create a false failed instrumentation check because its inputs expired.

### Additional context

This occurred on PR #7209. Attempt 2 of workflow run `30712233663` reran only the previously cancelled build job more than one day later. The resulting `PR Emulator.wtf` run (`30763712786`) failed with `Unable to download artifact(s): Artifact has expired`.

loganrosen:
## Summary

`PR Emulator.wtf` runs on `workflow_run` and downloads the `emulator-wtf-inputs` artifact from the triggering `Pull Request` run. That artifact had `retention-days: 1`, but GitHub allows reruns for 30 days, and a *partial* rerun re-triggers the downstream workflow without rebuilding the APKs. Past the first day the download fails with `Artifact has expired`, and the PR gets a red "Instrumentation Test app" check that says nothing about the code under review. This happened on #7209: run `30712233663` attempt 2 reran one cancelled job a day later, and downstream run `30763712786` failed at the download.

Retention goes to 3 days, which covers the manual restarts maintainers actually do, including one that spans a weekend. Past that, a new `check_inputs` job skips `emulator_wtf` instead of failing it, so no check appears on the PR at all and the workflow log explains that re-running *all* jobs of `Pull Request` rebuilds the inputs. That mirrors what #7093 settled on for the "triggering workflow didn't succeed" case. I stopped at 3 rather than 30 days because the artifact is ~100 MB and `Pull Request` runs ~19 times a day, so every extra day of retention leaves ~2 GB standing permanently to serve a rare edge case.

Two smaller fixes fell out of the same failure:

- `emulator_wtf` now downloads the inputs *before* creating the check run. The preflight can't close the window entirely, since the artifact can expire between the check and the download. Downloading first means such a failure happens before any check run exists, so the PR still stays clean.
- `publish_test_results` had the same bug through another door: its download step had no `name` and no `pattern`, so it pulled *every* artifact including the expired one. It failed on the run above too, so the unit and screenshot results never published either. It now excludes `emulator-wtf-inputs` by pattern, which is lossless (that artifact holds only APKs and `devices.txt`) and saves 100 MB per run.

Fixes #7297

## Checklist

## Any other notes

No changelog entry (CI-only, not user-visible) and no tests (workflow YAML, no Kotlin surface), so the first box is unticked. Validated with `yamllint --strict`, the same check the `yamllint` job runs, plus checking the artifacts API query and `download-artifact`'s pattern matching against real runs of both workflows.

### DIFF
--- a/.github/workflows/pr-emulator-wtf.yml
+++ b/.github/workflows/pr-emulator-wtf.yml
@@ -12,18 +12,60 @@ concurrency:
 permissions: {}
 
 jobs:
+  # A partial re-run of "Pull Request" re-triggers this workflow without
+  # rebuilding the emulator.wtf inputs, so they may already be gone. Skip
+  # emulator_wtf rather than failing it, so no red check appears on the PR.
+  # No `environment:` here, so an unusable run never costs a ui-test approval.
+  check_inputs:
+    name: "Check emulator.wtf inputs"
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.event == 'pull_request'
+      && github.event.workflow_run.conclusion == 'success'
+    permissions:
+      actions: read  # Needed for listing the artifacts of the triggering run
+    outputs:
+      available: ${{ steps.inputs.outputs.available }}
+    steps:
+      - name: Check whether the inputs artifact is still available
+        id: inputs
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          RUN_ID: ${{ github.event.workflow_run.id }}
+          RUN_URL: ${{ github.event.workflow_run.html_url }}
+        run: |
+          # Expired artifacts stay in the listing with `expired: true` rather than
+          # disappearing, so this covers both expiry and a missing upload.
+          available=$(gh api "repos/$REPO/actions/runs/$RUN_ID/artifacts?name=emulator-wtf-inputs" \
+            --jq '[.artifacts[] | select(.expired == false)] | length > 0')
+          echo "available=$available" >> "$GITHUB_OUTPUT"
+          if [ "$available" != "true" ]; then
+            echo "::warning::The emulator.wtf inputs from $RUN_URL are no longer available, so instrumentation tests were skipped. Re-run all jobs of that workflow to rebuild them."
+          fi
+
   emulator_wtf:
     name: "Instrumentation Test app"
     runs-on: ubuntu-latest
     environment: ui-test
-    if: >
-      github.event.workflow_run.event == 'pull_request'
-      && github.event.workflow_run.conclusion == 'success'
+    needs: [check_inputs]
+    if: needs.check_inputs.outputs.available == 'true'
     permissions:
       id-token: write  # Needed for OIDC authentication
       actions: read  # Needed for retrieving artifacts
       checks: write  # Needed to publish this job's status as a PR check
     steps:
+      # Kept ahead of the check run: the artifact can still expire while this job
+      # waits on the ui-test approval, and failing with no check run created yet
+      # leaves the PR clean, same as a check_inputs skip.
+      - name: Download emulator.wtf inputs
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: emulator-wtf-inputs
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          path: emulator-wtf-inputs
+
       # Surface this job's own success/failure as a check on the PR commit.
       # workflow_run jobs are detached from the PR, so the status is not shown
       # otherwise. A check run created via the API for a PR-head SHA gets grouped
@@ -86,14 +128,6 @@ jobs:
             echo "job_url=$JOB_URL"
           } >> "$GITHUB_OUTPUT"
 
-      - name: Download emulator.wtf inputs
-        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
-        with:
-          name: emulator-wtf-inputs
-          run-id: ${{ github.event.workflow_run.id }}
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          path: emulator-wtf-inputs
-
       - name: Load device list
         id: devices
         # The artifact comes from a PR build that may originate from a fork; validate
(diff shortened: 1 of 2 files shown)


jpelgrom:
> Git history doesn't record why retention was 1 day originally. If that was deliberate for a reason not visible in the diff, let me know and I'll adjust.

I think the idea was that it would always run immediately after a 'normal' `pr` workflow run (which it was originally a part of, but split to allow forks to also use it), so it didn't need long retention. However with some failures recently being due to the runner being out of RAM (perfect example: [this PR's initial run](https://github.com/home-assistant/android/actions/runs/30767902516/job/91551140726?pr=7298) 😅) not an actual build issue, we (maintainers) sometimes manually restart and that might not be within a day.

Changing the retention to be a bit longer is fine but it doesn't need to be very long. Maybe 3 days? What do you think @TimoPtr ?

TimoPtr:
> > Git history doesn't record why retention was 1 day originally. If that was deliberate for a reason not visible in the diff, let me know and I'll adjust.
> 
> I think the idea was that it would always run immediately after a 'normal' `pr` workflow run (which it was originally a part of, but split to allow forks to also use it), so it didn't need long retention. However with some failures recently being due to the runner being out of RAM (perfect example: [this PR's initial run](https://github.com/home-assistant/android/actions/runs/30767902516/job/91551140726?pr=7298) 😅) not an actual build issue, we (maintainers) sometimes manually restart and that might not be within a day.
> 
> Changing the retention to be a bit longer is fine but it doesn't need to be very long. Maybe 3 days? What do you think @TimoPtr ?

The retention of 3 days could make sense for the weekend but I don't think 7d make sense.

loganrosen:
Makes sense, thanks for the background. Dropped it to 3 days in 1f11c76. Restarts past that window still won't go red, since `check_inputs` skips the job instead of failing it.

.github/workflows/pr-emulator-wtf.yml
```diff
@@ -12,18 +12,60 @@ concurrency:
 permissions: {}
 
 jobs:
-  emulator_wtf:
-    name: "Instrumentation Test app"
+  # A partial re-run of "Pull Request" re-triggers this workflow without
+  # rebuilding the emulator.wtf inputs, so they may already be gone. Skip
+  # emulator_wtf rather than failing it, so no red check appears on the PR.
+  # No `environment:` here, so an unusable run never costs a ui-test approval.
```
jpelgrom:
```suggestion
```

.github/workflows/pr-emulator-wtf.yml
```diff
@@ -12,18 +12,60 @@ concurrency:
...
+    needs: [check_inputs]
+    if: needs.check_inputs.outputs.available == 'true'
     permissions:
       id-token: write  # Needed for OIDC authentication
       actions: read  # Needed for retrieving artifacts
       checks: write  # Needed to publish this job's status as a PR check
     steps:
+      # Kept ahead of the check run: the artifact can still expire while this job
+      # waits on the ui-test approval, and failing with no check run created yet
+      # leaves the PR clean, same as a check_inputs skip.
```
jpelgrom: Approval is automatic for this environment, so this comment is incorrect, but it still makes sense to download first to try to avoid a race.

.github/workflows/pr-emulator-wtf.yml
loganrosen: Good catch, dropped it.

.github/workflows/pr-emulator-wtf.yml
loganrosen: Fixed the comment in 81ec3fc. Kept the download first for the narrower window between `check_inputs` listing the artifact and this step downloading it. Also corrected the same claim in the PR description.

jpelgrom (approved):
Changes look good to me, but I'd like @TimoPtr to do a final check when he's back next week before merging

.github/workflows/pr-emulator-wtf.yml
```diff
@@ -12,18 +12,59 @@ concurrency:
 permissions: {}
 
 jobs:
-  emulator_wtf:
-    name: "Instrumentation Test app"
+  # A partial re-run of "Pull Request" re-triggers this workflow without
+  # rebuilding the emulator.wtf inputs, so they may already be gone. Skip
+  # emulator_wtf rather than failing it, so no red check appears on the PR.
```
TimoPtr, line 17: No red check appears but the check is required so we won't be able to merged too.

.github/workflows/pr-emulator-wtf.yml
```diff
@@ -12,18 +12,59 @@ concurrency:
...
+          RUN_ID: ${{ github.event.workflow_run.id }}
+          RUN_URL: ${{ github.event.workflow_run.html_url }}
+        run: |
+          # Expired artifacts stay in the listing with `expired: true` rather than
+          # disappearing, so this covers both expiry and a missing upload.
+          available=$(gh api "repos/$REPO/actions/runs/$RUN_ID/artifacts?name=emulator-wtf-inputs" \
+            --jq '[.artifacts[] | select(.expired == false)] | length > 0')
+          echo "available=$available" >> "$GITHUB_OUTPUT"
+          if [ "$available" != "true" ]; then
+            echo "::warning::The emulator.wtf inputs from $RUN_URL are no longer available, so instrumentation tests were skipped. Re-run all jobs of that workflow to rebuild them."
```
TimoPtr: I would make this an error and not a warning, if this happen we need to rerun the whole workflow so it's ok to have a red check IMO.

@jpelgrom wdyt?

jpelgrom: If it's an error the user submitting the PR might think they have to re-run the workflow, which they might not be able to do. But they still might check the warning, so no strong opinions.

TimoPtr:
```suggestion
            echo "::error title=Instrumentation tests skipped::The emulator.wtf inputs from $RUN_URL are no longer available. A maintainer needs to re-run all jobs of that workflow to rebuild them."
```
@jpelgrom I think that is more accurate?

PR #7320, opened 2026-08-10
Update BLE transmitter "Minor" documentation

d4nuu8:
## Summary

The special meaning of the minor value 40004 never made into core.

See home-assistant/core#104419.

The merged alternative implementation doesn't need any special value.

See home-assistant/core#104790.

## Checklist

- [x] I have not used AI for this contribution.


common/src/main/res/values/strings.xml
```diff
@@ -612,7 +612,7 @@
...
+    <string name="sensor_description_bluetooth_ble_emitter">Send BLE iBeacon with configured interval, used to track presence around house, e.g. together with roomassistant, esp32–mqtt–room or espresence projects.\n\nWarning: this can affect battery life, particularly if the \"Transmitter power\" setting is set to High or \"Advertise Mode\" is set to Low latency.\n\nSettings allow for specifying:\n– \"UUID\" (standard UUID format), \"Major\" and \"Minor\" (should be 0 – 65535), to tailor identifiers and groups\n- \"Transmitter Power\" and \"Advertise Mode\" to help to preserve battery life (use lowest values if possible)\n - \"Measured Power\" to specify power measured at 1m (initial default -59)\n\nIt is also possible to set beacons to only be transmitted when connected to a Home network, which may be desirable for privacy and battery life.\n\nNote:\nAdditionally a separate setting exists (\"Enable Transmitter\") to stop or start transmitting.</string>
```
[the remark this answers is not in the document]
d4nuu8: Removed

jpelgrom (approved):
Thanks for adjusting this!

common/src/main/res/values/strings.xml
```diff
@@ -612,7 +612,7 @@
...
+    <string name="sensor_description_bluetooth_ble_emitter">Send BLE iBeacon with configured interval, used to track presence around house, e.g. together with roomassistant, esp32–mqtt–room or espresence projects.\n\nWarning: this can affect battery life, particularly if the \"Transmitter power\" setting is set to High or \"Advertise Mode\" is set to Low latency.\n\nSettings allow for specifying:\n– \"UUID\" (standard UUID format), \"Major\" and \"Minor\" (should be 0 – 65535), to tailor identifiers and groups\n- \"Transmitter Power\" and \"Advertise Mode\" to help to preserve battery life (use lowest values if possible)\n- \"Measured Power\" to specify power measured at 1m (initial default -59)\n\nIt is also possible to set beacons to only be transmitted when connected to a Home network, which may be desirable for privacy and battery life.\n\nNote:\nAdditionally a separate setting exists (\"Enable Transmitter\") to stop or start transmitting.</string>
```
NoRi2909, line 615: That first dash in the list is an em-dash – is this intentional or should that be a simple dash, too?


```suggestion
    <string name="sensor_description_bluetooth_ble_emitter">Send BLE iBeacon with configured interval, used to track presence around house, e.g. together with roomassistant, esp32–mqtt–room or espresence projects.\n\nWarning: this can affect battery life, particularly if the \"Transmitter power\" setting is set to High or \"Advertise Mode\" is set to Low latency.\n\nSettings allow for specifying:\n- \"UUID\" (standard UUID format), \"Major\" and \"Minor\" (should be 0 – 65535), to tailor identifiers and groups\n- \"Transmitter Power\" and \"Advertise Mode\" to help to preserve battery life (use lowest values if possible)\n- \"Measured Power\" to specify power measured at 1m (initial default -59)\n\nIt is also possible to set beacons to only be transmitted when connected to a Home network, which may be desirable for privacy and battery life.\n\nNote:\nAdditionally a separate setting exists (\"Enable Transmitter\") to stop or start transmitting.</string>
```

d4nuu8: Shouldn't those actually all be standard dashes?

jpelgrom: This PR has been merged, please submit a new one with your suggestion.

It looks like it originally already was which is why I didn't notice the difference between the first and later ones.

NoRi2909: Yes, that em-dash was already present before. It's just a small inconsistency I noticed in Lokalise post-merge.

The Android UI does not render those dashed lists as markup, so we could also use bullets like  "•" (U+2022) instead to create a bulleted list within this and the next string?

jpelgrom: We could use bullet characters, but it will continue treating it like text. I'm also not sure translators will use them correctly if I'm being honest, but you have more experience with that @NoRi2909.

PR #7350, opened 2026-08-19
Fix missing home zone on the first location update after high accuracy mode turns off

### LINKED ISSUE #7349: Location update is failing when coming out of high precision mode
2026.6.5-full

**Android version(s):**
17

**Device model(s):**
Pixel Pro 10

**Home Assistant version:**
2026.8.2

When the home assistant app on my phone is reporting location to home assistant, it fails to report the zone when switching from high precision mode to regular.

For example, say I'm traveling from far outside my home zone to my home.  It works like this:

1. First, there are `state_changed` updates every 30 seconds because I am not in high precision area.
2. When I get into zone.home_expanded, the mobile app then sends `android.zone_entered` to home assistant.
3. Now home assistant receives `state_changed` messages much more frequently, about ever 3 seconds.  They show that I am not in the home zone still.
4. Eventually, the mobile app sends `android.zone_entered` with zone `zone.home`, which is what I expect.
5. Immediately after, the mobile app send another `state_changed`.  It has the same latitude and longitude as in step 4, which is good.  But it **doesn't** say that I'm in the home zone! THIS IS THE BUG.
6. About 30 seconds later, the phone gets a `state_changed` that includes the new zone, `zone.home`.
7. All the future `state_changed` for location are correct.

So it seems like the very first location update after switching from high precision to regular has the wrong zone in it, even though the latitude and longitude are all correct.  I'm able to reproduce this by using a GPS faker tool on my phone.  I also see this problem just driving home with my phone.

I think that there is a bug in the code that is causing it to generate and send that `state_changed` with the old zone from before the `android.zone_entered`.

I will try to investigate this using some AI.

**Companion App Logs:**

```

```

**Additional information:**

eyal0:
## Summary

Fixes #7349 

## Checklist

- [ ] I have read the [Open Home Foundation AI Policy](https://developers.home-assistant.io/docs/ai_policy).

## Any other notes

The first location send after android.zone_entered was keeping the old zone, even when lat/long were already correct.

What was going wrong

High accuracy is sending every few seconds, then you enter zone.home and the mode switches off. The geofence event uses the same GPS fix as the last high-accuracy update: same coordinates, but an older location.time than the wall-clock time of that last send.

Two skip rules then dropped the geofence location update:

location.time < lastLocationSend → treated as stale (SKIPPED_NOT_LATEST)
Same coordinates as the last send → treated as a duplicate for 15 minutes
android.zone_entered still goes out (that path is separate). The state_changed you see immediately after is the last high-accuracy update, still with the old zone. Regular updates then start, the first one is usually the same last-known fix and gets skipped too. About 30 seconds later a fresh sample gets through and Home Assistant finally shows zone.home.

What changed

Geofence enter/exit updates are no longer skipped as stale or duplicate.
Zone membership is GPS plus entered geofences, so zone.home is included even when that shared GPS fix still looks “outside” to the radius check. On Core 2026.6+ that is sent as in_zones.
The payload is rebuilt just before the webhook, so an in-flight high-accuracy send that started before the geofence still picks up the new zone.


eyal0:
I set an Cursor Grok 4.6 High Fast on it and let it cook.  I have no idea if this solves the problem.  I don't know how to build and run android code.  But I'm wiling to learn!

eyal0:
> Hi @esoha-nvidia
> 
> It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant/android#7350).
> 
> 
> Thanks!

This isn't working.  It says:

```
No pending request found for eyal0. Are you signing the CLA with the same GitHub user that created the commits in the Pull Request?
```

PR #6879, opened 2026-05-21
Send location in zones when using SEND_LOCATION_AS_ZONE_ONLY

TimoPtr:
## Summary

To fulfill the new requirements in core we need to send `in_zones` when we send a location update using zone only. See https://github.com/home-assistant/architecture/discussions/1387 for more details. Implemented in core https://github.com/home-assistant/core/pull/171814

## Checklist

- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

## Screenshots

image

PR #7351, opened 2026-08-19
Fix missing home zone on the first location update after high accuracy mode turns off

### LINKED ISSUE #7349: Location update is failing when coming out of high precision mode

eyal0:
This fixes #7349 

## Checklist


eyal0:
I just used AI to fix the issue, I don't know really know what I'm doing here.  If someone can explain to me how to load a new build based on this code into my phone, I would be able to test if this solves the problem.

eyal0:
I figured out how to compile the app and run it on my phone.  I can see that with the new debug image is solving the problem.  The first state_changed after precision mode ends and regular mode begins now **does** have the correct zone.  All the subsequent messages which previous had the correct zone continue to have the correct zone.  This code does seem to solve the problem.

jpelgrom:
> I just used AI to fix the issue, I don't know really know what I'm doing here. If someone can explain to me how to load a new build based on this code into my phone, I would be able to test if this solves the problem.

Hello, thanks for contributing! The [Open Home Foundation's AI policy](https://developers.home-assistant.io/docs/ai_policy/#code-and-documentation-contributions) requires you to review and understand the code you submit. Were you able to review the code since posting this message? If not, I'm afraid we cannot accept it, but will look at your issue report.

eyal0:
> > I just used AI to fix the issue, I don't know really know what I'm doing here. If someone can explain to me how to load a new build based on this code into my phone, I would be able to test if this solves the problem.
> 
> Hello, thanks for contributing! The [Open Home Foundation's AI policy](https://developers.home-assistant.io/docs/ai_policy/#code-and-documentation-contributions) requires you to review and understand the code you submit. We're you able to review the code since posting this message? If not, I'm afraid we cannot accept it, but will look at your issue report.

Since writing that comment, I simplified the code a lot and it's short and understandable. It looks right to me but if you've got experts on this part of the code, I'd feel more comfortable having them review it.

common/src/main/kotlin/io/homeassistant/companion/android/common/data/integration/ZoneAttributes.kt
```diff
@@ -19,3 +19,50 @@ fun Entity.containsWithAccuracy(location: Location): Boolean {
...
+fun resolveInZones(
+    location: Location,
+    configuredZones: List<Entity>,
+    enteredGeofenceRequestIds: Collection<String>,
+    serverId: Int,
+): List<Entity> {
+    val configuredByEntityId = configuredZones.associateBy { it.entityId }
+    val fromGps = configuredZones.filter { zone ->
+        val radius = zone.attributes["radius"] as? Number
+        radius != null && zone.containsWithAccuracy(location)
```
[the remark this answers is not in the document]
eyal0: No longer relevant.

app/src/main/kotlin/io/homeassistant/companion/android/sensors/LocationUpdateSkipEvaluator.kt
```diff
@@ -0,0 +1,47 @@
...
+): LocationHistoryItemResult? {
+    if (now + LOCATION_FUTURE_TOLERANCE_MS < locationTime && !highAccuracyModeEnabled) {
+        return LocationHistoryItemResult.SKIPPED_FUTURE
+    }
+    if (locationTime < lastLocationSend && !isGeofence) {
+        return LocationHistoryItemResult.SKIPPED_NOT_LATEST
+    }
+    if (now - locationTime >= LOCATION_MAX_AGE_MS) {
+        return LocationHistoryItemResult.SKIPPED_OLD
+    }
```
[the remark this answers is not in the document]
eyal0: No longer relevant.

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -1062,13 +963,77 @@ class LocationSensorManager @Inject constructor(
...
+            location = location,
+            configuredZones = getZones(serverId),
+            enteredGeofenceRequestIds = enteredGeofences,
+            serverId = serverId,
+        )
+        val supportsInZones = serverManager.getServer(serverId)?.version?.isAtLeast(
+            IN_ZONES_MIN_SERVER_YEAR,
+            IN_ZONES_MIN_SERVER_MONTH,
+            0,
+        ) == true
```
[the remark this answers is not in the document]
eyal0: No longer relevant.

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -912,128 +916,24 @@ class LocationSensorManager @Inject constructor(
...
+            locationTime = location.time,
+            now = now,
+            lastLocationSend = lastLocationSend[serverId] ?: 0,
+            lastUpdateLocation = lastUpdateLocation[serverId],
+            updateLocationString = updateLocationString,
+            isGeofence = trigger?.isGeofence == true,
+            highAccuracyModeEnabled = highAccuracyModeEnabled,
+        )
+        if (skipReason != null) {
+            Timber.d("Skipping location update ($skipReason) received: ${location.time} last sent: $lastLocationSend")
```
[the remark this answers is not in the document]
eyal0: No longer relevant.

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -924,7 +924,9 @@ class LocationSensorManager @Inject constructor(
             val inZones = zones
                 .filter {
                     val radius = it.attributes["radius"] as? Number
-                    return@filter radius != null && it.containsWithAccuracy(location)
+                    val matchesGps = radius != null && it.containsWithAccuracy(location)
+                    val matchesGeofence = lastEnteredGeoZones.contains("${serverId}_${it.entityId}")
+                    return@filter matchesGps || matchesGeofence
```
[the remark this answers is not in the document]
eyal0: This is already the case for other usage of `lastEnteredGeoZones` in the rest of the file.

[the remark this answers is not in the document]
eyal0: The rest of the file works like this, too, so I won't change this.

eyal0:
I'm doing some refactoring to make it more uniform.

eyal0:
The calculation of inZones is moved out of the branch body so that it can be used in both forks of the branch.  I have tested that this solves the problem.

To test, go to Settings -> Tools -> Events and listen to event `*` filtered on the word "latitude".  Then use a fake gps app on your phone to simulate moving from away from the zone toward it.  You should see that the zone of the device changes immediately after the device goes from the high accuracy zone into the zone.  It used to take 30 seconds or so.

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -941,16 +950,7 @@ class LocationSensorManager @Inject constructor(
...
-                        2026,
-                        6,
-                        0,
-                    ) == true
-                ) {
-                    inZones.map { it.entityId }
-                } else {
-                    null
-                },
+                inZones = inZones,
```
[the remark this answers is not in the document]
eyal0: This is pre-existing behavior.

eyal0:
@jpelgrom is there any progress on this. It's a real issue for me. I find that detecting entering a zone is not working at the right time. This PR fixes the issue. It's also a low risk fix, I don't see it causing any new bugs.

app/src/main/res/xml/changelog_master.xml
jpelgrom, line 1: We usually don't mention detailed bug fixes, I would remove the changes here.

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -1002,7 +1002,7 @@ class LocationSensorManager @Inject constructor(
                 "Received location that is ${now - location.time} milliseconds old, ${location.time} compared to $now with source ${location.provider}",
             )
             if (lastUpdateLocation[serverId] == updateLocationString) {
-                if (now < (lastLocationSend[serverId] ?: 0) + 900000) {
+                if (now < (lastLocationSend[serverId] ?: 0) + 900000 && trigger?.isGeofence != true) {
```
jpelgrom: Duplicate location even if from geofence that are recent do not need to be sent, the result would be the same?

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -989,9 +989,9 @@ class LocationSensorManager @Inject constructor(
             return
         }
 
-        if (location.time < (lastLocationSend[serverId] ?: 0)) {
+        if (location.time < (lastLocationSend[serverId] ?: 0) && trigger?.isGeofence != true) {
```
jpelgrom: I don't follow this change. If a geofence-based location arrived but we have a more recent location update, I don't see why the geofence-based location should be sent? It's older. Your fix is the couple of lines above in `return@filter matchesGps || matchesGeofence`.

app/src/main/res/xml/changelog_master.xml
eyal0: Okay!

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
eyal0: I agree.

The issue is that there are two location-related services registered with Android Play Services, one to receive background GPS updates and one where zones are registered and Android Play Services reports getting in and out of a zone.  It's possible in theory for the geofence notification from Play Service to arrive before *or* after the location update.  So we might want to have the geofence change force the location update.

But now that we're always computing what zones we are in for all messages, I guess that it doesn't matter.  I'll revert this and the one below.

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
eyal0: Same as above.  Done.

eyal0 (commented):
Thanks for your comments.  Please take another look.

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -919,19 +919,28 @@ class LocationSensorManager @Inject constructor(
         val updateLocation: UpdateLocation
         val updateLocationString: String
         val updateLocationAs: String = getSendLocationAsSetting(serverId)
+        val zones = getZones(serverId)
+        val enteredZones = zones
+            .filter {
+                val radius = it.attributes["radius"] as? Number
+                val matchesGps = radius != null && it.containsWithAccuracy(location)
+                val matchesGeofence = lastEnteredGeoZones.contains("${serverId}_${it.entityId}")
+                return@filter matchesGps || matchesGeofence
```
TimoPtr, line 928: The filter is quite different than the one we had can you describe why?

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
```diff
@@ -962,7 +962,7 @@ class LocationSensorManager @Inject constructor(
                 gps = listOf(location.latitude, location.longitude),
                 gpsAccuracy = accuracy,
                 locationName = null,
-                inZones = null,
+                inZones = inZones,
```
TimoPtr, line 965: From what I remember while implementing https://github.com/home-assistant/android/pull/6879 we decided that in this specific case when we have the coordinates we don't send the inZones to core and core does the inZones computation directly. CC @emontnemery

eyal0: I've done some more analysis.  It seems that there is a small region where android has decided that you are in the home zone but home assistant still hasn't.  (In the image below, problem zone.)

image

In android, zones are registered with the android system and android sends updates to the app when a zone is entered or exited.  The app is registering home_expanded and home.

In android, the home assistant companion app is sending location data about once every 30 seconds unless you are in the home_expanded zone.  Imagine  that you are traveling home from far away.  The black dots on the left represent updates every 30 seconds, so they are spaced far apart.  When android reports that you entered home_expanded, the gps reports become every 2-4 seconds, active location monitoring.  Those are the brown dots and they are close together because the reports are frequent.

Eventually, android will get a notification that we have entered the home zone.  That's the purple circle.  This is the "problem zone", where android thinks that you are in the home zone but the math will *not* show that you are in the home zone.  Maybe this is due to some inaccuracy in the android zones.  The app will send a message to home assistant about the location and home assistant computes that you are *not* in the home zone.  The app will now report every 30 seconds, shown as the black dots again being spaced far apart.

By the time the app reports again, it'll be too late and you are already well into the zone.

I have tested this with a fakegps program on my phone.  I can sit in the problem zone for 10 minutes with android not reporting at high accuracy while Home Assistant still reports that I am not home.

Another possible solution is to get the distance calculation that home assistant is doing and put it into the companion app for deciding when you end high accuracy mode.  You would still turn on high accuracy mode using the android geofence but you would only turn it off when you are in the home zone according to home assistant's math, not according to android's geofence.

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
eyal0: It's only different in two lines.  The "matchesGeofence" is new.  That comes from the updates from the android geofence service.  The code used to only report zones based on whether or not you are mathematically in the zone (`matchesGps`).  Now it uses that but also includes zones where android thinks that you are in the zone based on the geofence.

As described below, android thinks that you are in the zone before the math shows that you are in the zone.

TimoPtr: 🤔 I guess the geofencing is less precise and when you are close it says you are in the zone. I would like @emontnemery on this topic before moving with this PR and also @bgoncal to align iOS if needed.

eyal0: In the meanwhile, I will explore a solution that allows Home Assistant to determine whether or not you are in the zone.  I agree that it would be better to use that instead of relying on Android's geofence.  I would like Home Assistant to be the ultimate decider of whether or not I am in the zone, not Android.

The tricky part in the code is that companion decides which zone you are in based on the geofence reports from Android, so all the code that is trying to figure out whether or not to turn on or off the high accuracy mode doesn't know the location.  It just knows what geofences we are in, which is already shown to be inaccurate.

I'll try a different solution.

eyal0: Please look at #7375 for another approach, @TimoPtr .

app/src/full/kotlin/io/homeassistant/companion/android/sensors/LocationSensorManager.kt
eyal0: @TimoPtr Just to let you know, I did find a solution that doesn't require us to send the zone, like you wanted.  It's #7375 .  I think that it's a nice solution.  It also fixes another pre-existing bug along the way.

eyal0:
I find #7375 more effective.

PR #7372, opened 2026-08-24
Select server URL based on any Wi-Fi connection

adamff-dev:
## Summary

This change adds support for selecting the Home Assistant server URL based on the current Wi-Fi connection.

Users can configure whether any Wi-Fi connection should be treated as an internal connection. The server settings and SSID configuration screens expose this option, and the connection state provider uses it when determining which server URL to use. The database schema was updated accordingly, with tests added for the connection state behavior.

## Checklist

## Screenshots

Screenshots are not included.

## Any other notes

The common module tests completed successfully. A Room database schema migration is included for the new connection setting.


jpelgrom:
I do not believe we should add this option as it promotes insecure behavior. What issue are you trying to solve?

TimoPtr:
I share the same concern as @jpelgrom so for now I'm going to close this PR. I don't get how this could be use securely.

adamff-dev:
The problem is that I don't want to have my device location always activated to detect the WiFi SSID

TimoPtr:
> The problem is that I don't want to have my device location always activated to detect the WiFi SSID

I get your issue but accepting this could lead to exposing users on insecure networks and we don't want that. I would suggest you in that case to use only one URL for all the networks and not rely on the internal logic we have in the app. This way you don't have to enable location at all.

PR #7381, opened 2026-08-28
Ensure WebView request mTLS cert to the app on cold start with cache

### LINKED ISSUE #7324: Android Companion App requires multiple Retry attempts before frontend loads when using Cloudflare Tunnel
2026.6.5-full

**Android version(s):**
16

**Device model(s):**
OPPO Find X9

**Home Assistant version:**
2026.8.1

### Checklist
 I updated Home Assistant Core to the latest version.
 I updated the Android Companion App to the latest version.
 I searched existing issues before opening this report.

### Describe the issue
I am experiencing a recurring issue with the Home Assistant Android Companion App.
When launching the app using my external URL exposed through a Cloudflare Tunnel, the frontend often fails to load and displays "Unable to connect to Home Assistant" with a Retry button.
In most cases I must press Retry twice before the dashboard loads successfully.
The issue is reproducible several times per day.

### What works
Direct access using the local URL.
Access to the same external URL through a web browser.
iOS Companion App.
Android Companion App immediately after performing Reset Frontend Cache.

The problem disappears after resetting the frontend cache but returns after approximately 10 application launches.

**Companion App Logs:**
[Logs Companion Android anonymized.txt](https://github.com/user-attachments/files/30950600/Logs.Companion.Android.anonymized.txt)

**Additional information:**
### Environment
Home Assistant Core: 2026.8.1
Home Assistant OS: 18.2
Cloudflared Add-on: 7.0.13
Android Companion App: 2026.6.5-full
Android WebView: 150.0.7871.181

### Troubleshooting performed
**Home Assistant Core**
No errors are logged in Home Assistant Core when the problem occurs.

**Cloudflare Tunnel**
Cloudflare Tunnel logs show successful communication with Home Assistant while the issue is occurring:
GET / returns 200 OK
GET /api/websocket is successful
Companion App webhook calls return 200 OK
GET /sw-modern.js returns 304 Not Modified

No HTTP 4xx or 5xx errors are present in the tunnel logs.
Example : 
GET https://my-ha-instance.example.com/?external_auth=1
200 OK
GET https://my-ha-instance.example.com/api/websocket
GET https://my-ha-instance.example.com/sw-modern.js
304 Not Modified

This suggests that:

- The tunnel is operational
- Home Assistant is reachable
- WebSocket connectivity appears functional
- Authentication requests successfully reach Home Assistant

### Companion App logs

Companion App logs show:

```text
Using external URL
```

multiple times during startup.

The logs also show:

```text
frontend/loaded
```

which suggests the frontend eventually loads successfully.

However, I also noticed repeated messages such as:

```text
Invalid resource ID 0x00000000
```

and:

```text
Cannot set background color. Skipping coloring...
```

during the loading sequence.

## Additional observations

- The problem only affects Android
- iOS works correctly against the same Home Assistant instance
- Browser access works correctly against the same Cloudflare Tunnel
- Reset Frontend Cache consistently fixes the issue temporarily
- After a frontend cache reset, the application loads instantly for several launches before the issue gradually returns
- No matching errors are present in Home Assistant Core logs
- Cloudflare Tunnel logs indicate successful communication with Home Assistant during the issue

## Possible area of investigation

The behavior appears related to frontend loading, WebView caching, service worker caching, or Android-specific frontend resource handling.

The strongest indicator seems to be that performing **Reset Frontend Cache** immediately resolves the problem, but only temporarily.

TimoPtr:
## Summary

On servers requiring a TLS client certificate (mTLS), the first frontend load after a process start always failed and showed the retry screen. The frontend's app shell is served by its service worker without any network fetch, so the frontend's WebSocket is the first connection to reach the server, and Chromium seems to not invoke `onReceivedClientCertRequest` during a WebSocket handshake: the connection fails instead of asking for the certificate.

This PR primes the certificate before loading the frontend when a client certificate is configured. A new `WebViewAction.PingUrl` sends a `HEAD` request for `/manifest.json` through the WebView network stack, which triggers the certificate negotiation on a regular request. Chromium caches the selection per host, so the frontend's WebSocket reuses it and connects on the first try. Priming is capped by a timeout and every failure mode degrades to loading without priming, keeping the previous behavior at worst.

Along the way, `WebViewAction.AwaitableAction` became a `sealed class` that owns its completion (await() returns the outcome or throws), so the deferred is no longer part of each action's public API.

The downside of this approach is that any app with at least one server with mTLS adds a new HEAD call before reaching the server and we need to wait for the answer so it could adds some delay before opening the URL one round trip to the server and potentially +5ms of polling since we need a synchronous call that we can cancel.

Fixes #7324.

## Checklist

## Any other notes

I was able to reproduce the issue by force stopping the app with an instance using Cloudflare and mTLS configured, with the PR I'm not able to reproduce the issue anymore.


app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
```diff
@@ -103,24 +118,87 @@ sealed interface WebViewAction {
...
+                }
+            }
+        }
+
+        companion object {
+            /** Maximum time [await] waits for the request to finish before giving up. */
+            internal val PING_TIMEOUT = 5.seconds
+
+            /** Interval between checks of [PING_COMPLETED_FLAG], also the detection lag after completion. */
+            private val PING_POLL_INTERVAL = 5.milliseconds
```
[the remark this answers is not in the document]
TimoPtr: @jpelgrom I think we are on the critical path here I would like to minimize the impact of this polling.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
```diff
@@ -103,24 +119,96 @@ sealed interface WebViewAction {
...
+                fetch($urlJson, { method: 'HEAD', cache: 'no-store', mode: 'no-cors' })
+                    .then(function() { window.$completedFlag = $urlJson; }, function() { window.$completedFlag = $urlJson; });
+            """.trimIndent()
+            webView.evaluateJavascript(script) { pollCompletion(webView, urlJson) }
+        }
+
+        override suspend fun await() {
+            try {
+                val elapsed = withTimeout(PING_TIMEOUT) { measureTime { result.await() } }
+                Timber.d("Ping of ${sensitive { url }} completed in $elapsed")
```
jpelgrom:
```suggestion
                Timber.d("Ping of ${sensitive(url)} completed in $elapsed")
```

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
```diff
@@ -103,24 +119,96 @@ sealed interface WebViewAction {
...
+
+        override fun run(webView: WebView) {
+            val urlJson = Json.encodeToString(url)
+
+            // The completion variable holds the pinged URL rather than a boolean so a stale completion
+            // from a previous ping cannot be mistaken for this one.
+            val script = """
+                window.$completedFlag = null;
+                fetch($urlJson, { method: 'HEAD', cache: 'no-store', mode: 'no-cors' })
+                    .then(function() { window.$completedFlag = $urlJson; }, function() { window.$completedFlag = $urlJson; });
```
jpelgrom:
```suggestion
                    .finally(function() { window.$completedFlag = $urlJson; });
```

finally requires WebView 63+ (Dec 2017) - the frontend won't work with a WebView that old so I think it's safe to use this function

app/src/test/kotlin/io/homeassistant/companion/android/frontend/WebViewActionPingUrlTest.kt
```diff
@@ -0,0 +1,141 @@
...
+
+/** [TEST_URL] as the JSON-encoded value evaluateJavascript reports for the completion variable. */
+private const val TEST_URL_JSON = "\"$TEST_URL\""
+
+/**
+ * Tests [WebViewAction.PingUrl] by capturing the scripts it evaluates and feeding results back
+ * through the captured callbacks, standing in for the WebView.
+ */
+@OptIn(EvaluateJavascriptUsage::class)
+class WebViewActionPingUrlTest {
```
jpelgrom: Why not add this to WebViewActionTest (potentially as a nested block)? It is bigger than most actions tested in WebViewActionTest so far but not hundreds of lines.

jpelgrom (commented):
Testing this I was able to reproduce the issue as you mentioned by force stopping the app _most_ of the time but not all of the time, it seems there is some caching / my device was fast enough to not encounter the issue every time. It did seem to wait less often and I don't see a major downside so good to try this.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/WebViewAction.kt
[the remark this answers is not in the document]
jpelgrom: > I would like to minimize the impact of this polling

could be read as

 - I want to keep the startup time as small as possible
 - I want to not create disproportionate startup CPU work

When testing this in real life the latency is >100ms for the first fetch for me as it is cold starting everything. I think we can lower the interval a bit to 10-20 ms.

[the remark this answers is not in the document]
TimoPtr: I went with 15ms as a middle ground.

app/src/test/kotlin/io/homeassistant/companion/android/frontend/WebViewActionPingUrlTest.kt
TimoPtr: Moving

jpelgrom (approved):

PR #7400, opened 2026-09-01
Add monitoring action and tweak screenshot test job

TimoPtr:
## Summary

In order to tweak our RAM settings on the CI, I've added a monitor action that logs every 10s the current state of the VM. Remove the hard constraint on the numbers of worker for the screenshot tests.

## Checklist

- [ ] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.
- [x] I have read the [Open Home Foundation AI Policy](https://developers.home-assistant.io/docs/ai_policy).

Select exactly one option that describes AI usage in this contribution:

- [ ] I have not used AI for this contribution.
- [x] AI assistance was used for this contribution.
- [ ] AI fully generated the code for this contribution, but I've reviewed and understood it before submitting and will respond without AI during review.

PR #7435, opened 2026-09-08
Address zizmor reports on E2E workflow

TimoPtr:
## Summary

Zizmor reported 2 potentials issues in the E2E workflow. This PR address them.


TimoPtr:
Started https://github.com/home-assistant/android/actions/runs/34207062477 to check if it works properly.

TimoPtr:
> Started https://github.com/home-assistant/android/actions/runs/34207062477 to check if it works properly.

Everything works as expected

jpelgrom (approved):
This was also what I did after noting how there could be injection on #7400 😅

PR #7444, opened 2026-09-11
Fix unavailable SSID comparisons

loganrosen:
## Summary

Remove the Detekt findings around unavailable SSID handling without relying on `String` object identity.

`WifiHelper.getWifiSsid()` now checks Android's raw SSID value before removing quotes. Android documents that a real decoded SSID is quoted, while `WifiManager.UNKNOWN_SSID` is returned unquoted when the value is unavailable. This preserves a legitimate network literally named `<unknown ssid>` while filtering the unavailable sentinel on Android 11 and newer, including OEM implementations that return an equal but distinct string.

The normalized nullable SSID is reused by network matching, Improv provisioning, SSID settings, and home-network onboarding. The SSID suggestion condition is also expressed as a nullable eligible candidate, removing the existing `ComplexCondition` finding without changing blank or configured-network behavior.

Focused tests cover raw sentinel filtering, the quoted same-name edge case, pre-Android-11 compatibility, BSSID fallback, and all affected consumers. The affected Detekt baselines shrink by seven entries across app release variants and common.

## Checklist

## Screenshots

N/A — this changes SSID filtering behavior without changing the UI appearance.

User Documentation: home-assistant/companion.home-assistant# N/A

Developer Documentation: home-assistant/developers.home-assistant# N/A

## Any other notes

Validation performed:
- 42 targeted unit/Robolectric tests across the four affected test classes
- `:common:ktlintCheck` and `:app:ktlintCheck`
- `:common:detektMain` and `:app:detektMain`
- `git diff --check`


common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/WifiSsid.kt
```diff
@@ -0,0 +1,15 @@
...
+import io.homeassistant.companion.android.common.util.SdkVersion
+
+/**
+ * Returns whether this value is [WifiManager.UNKNOWN_SSID] on Android 11 or newer.
+ *
+ * The receiver must be an unquoted SSID.
+ *
+ * Earlier Android versions deliberately return `false` to preserve the app's existing SSID handling.
+ */
+fun String.isUnavailableSsid(): Boolean =
```
TimoPtr: I'm not a fan of public extensions on String. It is polluting the autocompletion for every strings that are not a SSID. I would keep this a utils function that takes a string as parameter.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/improv/ui/ImprovSheet.kt
```diff
@@ -152,7 +150,7 @@ private fun ColumnScope.ConfiguringDeviceSection(
  * values when the app lacks the location permission.
  */
 private fun String?.takeIfDisplayable(): String? = takeIf {
-    !it.isNullOrBlank() && (!SdkVersion.isAtLeast(Build.VERSION_CODES.R) || it !== WifiManager.UNKNOWN_SSID)
```
TimoPtr, line 155: > Framework-provided SSID strings are not guaranteed to be the same object as the Android constant

@jpelgrom I think the ref equality was on purpose, it's coming from the same place (library).

common/src/main/kotlin/io/homeassistant/companion/android/common/data/network/WifiSsid.kt
loganrosen: Good point. I moved this into `WifiHelper.getWifiSsid()`, so callers now get a normalized value and there’s no `String` extension.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/improv/ui/ImprovSheet.kt
loganrosen: You were right about why this used reference equality. I checked AOSP, and Android 11–15 return the constant directly, so the old check works there even after `removeSurrounding`. Since the API only guarantees the value, I moved the check before removing quotes. That still allows a real network named `<unknown ssid>` without depending on reference identity.

TimoPtr: Let's remove the extension and do the takeIf directly where it is needed.

TimoPtr:
I'll let @jpelgrom give the final word on this since he has more experience with this API.

app/src/main/kotlin/io/homeassistant/companion/android/frontend/improv/ui/ImprovSheet.kt
loganrosen: Dropped the private extension and inlined the `takeIf` at the call site.

jpelgrom: > > Framework-provided SSID strings are not guaranteed to be the same object as the Android constant
> 
> @jpelgrom I think the ref equality was on purpose, it's coming from the same place (library).

I'm not quite sure it was on purpose or originally a mistake, but it worked and turned out to be a feature as @loganrosen mentions :)

jpelgrom (approved):

PR #7448, opened 2026-09-11
Improve e2e agentic workflow to fit within 20mn

TimoPtr:
## Summary

We had our 2E2 workflow fail this morning because of some accessibility not being able to click on an item that was present. It happens once in a while (maybe we could add a retry). The agentic workflow to write the issue did start properly and did narrow down the issue but
- It has been canceled because it reach the 20mn timeout
- It did fight with the sandbox a lot trying to use tools that are forbidden like `cd`, ...
- It tried to look at the upstream sources where it should have stop earlier

This PR is an attempt to improve the agentic workflow to fit within the 20mn timeslot.

I did use Claude to help me update the skill to try to avoid this situation.


.agents/skills/ha-android-e2e-debugging/SKILL.md
```diff
@@ -27,30 +27,39 @@ gh run download <run-id> --name e2e-artifacts --dir e2e
...
 
 ### 2.3 Home Assistant logs
 
-Check `homeassistant.log` for errors at the same timestamp, and confirm in `homeassistant-config.json` that `mobile_app` is in `components` — the workflow verifies this at startup, but a later integration failure can still break onboarding. `homeassistant-container.json` gives the exact image digest, which matters for the next step.
+Check `homeassistant.log` for errors at the same timestamp, and confirm in `homeassistant-config.json` that `mobile_app` is in `components`: the workflow verifies this at startup, but a later integration failure can still break onboarding. `homeassistant-container.json` gives the exact image digest for the next step.
 
 ### 2.4 Upstream: core and frontend
 
+Only when 2.1 to 2.3 explain nothing. A typical upstream symptom is a clean split by API level: every shard below some API level fails at the same step while every shard above it passes. The frontend ships one bundle to all WebViews, so a change that relies on a newer web feature breaks the older WebViews together.
```
jpelgrom, line 64: > Only when 2.1 to 2.3 explain nothing.

Is this needed if you say at the top "Once a source explains it, skip the rest."? If not, shouldn't 2.3 also start with

> Only when 2.1 and 2.2 explain nothing.

and 2.2 start with

> Only when 2.1 explains nothing.

jpelgrom (commented):
So much handholding. There were some mistakes humans can read over but the AI stumbles on. I don't know if it'll fix things but neither will you, but let's try. It does address where it spent a lot of time in the log.

.agents/skills/ha-android-e2e-debugging/SKILL.md
TimoPtr: It seems that it didn't follow the rule the last time so let's be more direct in this revision.

jpelgrom: OK, but then add it to 2.2 and 2.3 as well

PR #7476, opened 2026-09-15
Force patched Netty on the build classpath

frenck:
## Summary

Forces a patched Netty onto the build classpath, clearing 25 of the 28 open Dependabot alerts on this repository, the one critical among them.

All 25 come from a single path:

```
com.google.firebase:firebase-appdistribution-gradle:5.3.0
  └── io.grpc:grpc-netty:1.69.1
        └── io.netty:* 4.1.110.Final
```

5.3.0 is the newest version Google has published, so there is no plugin bump available to ride. The plugin is genuinely in use, `appDistributionUploadFullRelease` runs from `.github/workflows/onPush.yml`, so it cannot be dropped either. That leaves pinning Netty ourselves until Google ships a newer gRPC.

The affected modules and the versions the advisories want:

| Module | Alerts | Needs |
| --- | --- | --- |
| io.netty:netty-codec-http | 11 | 4.1.137.Final |
| io.netty:netty-codec-http2 | 6 | 4.1.136.Final |
| io.netty:netty-handler | 4 (1 critical) | 4.1.137.Final |
| io.netty:netty-codec | 1 | 4.1.136.Final |
| io.netty:netty-handler-proxy | 1 | 4.1.133.Final |

4.1.138.Final is the current 4.1.x release and covers all of them. The jump is patch level within 4.1.x, which gRPC treats as compatible.

The `eachDependency` shape matches the existing Kotlin ABI validation pin a few lines further down in the same file.

## Impact

None on the app. This is the Gradle build classpath, so nothing here is compiled into or packaged with the APK. The only consumer is the App Distribution upload task in CI.

## Checklist

## Any other notes

**This has not been built locally.** The dependency path above was traced through the published POMs, not through a Gradle resolution, so CI is the first real test of whether the force lands. The "thoroughly tested" box is deliberately left unchecked. If the build is green, the force worked.

The change touches no locked configuration, so no `gradle.lockfile` regeneration should be needed. If CI disagrees, that assumption is where to look first.

Three alerts are left over and are not addressed here:

- 2 for `ch.qos.logback:logback-core` 1.3.14, pulled by `com.pinterest.ktlint:ktlint-cli:1.5.0`. `ktlint-gradle` is already on 14.2.0, the newest, so this needs the ktlint CLI version raised through the extension. That configuration is locked, so it needs `--write-locks` and deserves its own pull request.
- 1 for `org.jetbrains.kotlin:kotlin-gradle-plugin`, wanting 2.4.20-Beta1 or newer. AGP 9.4.0 declares 2.2.10 in its POM, while the version catalog sets Kotlin 2.4.20, so Gradle should already resolve the satisfied version. This looks like Dependabot reading POMs statically rather than a real exposure, but it is worth confirming against a real dependency report before dismissing.


TimoPtr:
I'm in favor of removing App distributions instead https://github.com/home-assistant/android/pull/7477 otherwise a cleaner solution is to add a direct deps to netty in the deps of the project define in the version catalog. This way renovate would be able to track and update the version.

TimoPtr:
Closed in favor of removal in #7477

PR #7477, opened 2026-09-15
Remove Firebase app distributions

TimoPtr:
## Summary

Firebase app distributions rely on old dependencies with vulnerabilities we could fix some versions to not be impacted by those, but I think we should simply drop it since no one uses it. Anyone that wants to test the app can simply use the APK available on each PR or from main or from a release.


.github/workflows/onPush.yml
jpelgrom:
```suggestion
    name: GitHub build and (pre)release
```

jpelgrom (approved):
I don't have insight in the number of users on the 'old' Firebase app distribution, but this hasn't been available since I started contributing to HA on Android (almost 5 years now). In favor of removing.

cc @dshokouhi

PR #6626, opened 2026-03-25
Add native Android media controls for HA media_player entities

FletcherD:
## Summary

I wanted to be able to control a media player entity natively on Android without having to open the app or navigate to a widget. So this feature exposes one or more Home Assistant `media_player` entities as native Android Media Controls ([described here](https://developer.android.com/media/implement/surfaces/mobile)) in the notification shade, the same UI used by other media players on Android. 
The media controls show the currently playing track info and play position with album art. Prev/next track, play/pause and seek controls work and are forwarded to the media_player entity (if the entity supports them). 
A new "Media controls" setting is added under "Companion app" to choose which media_player entities to expose in the media controls, if any. You can choose more than one entity, in which case a notification will be created for each one.
Unit tests are added to test playback state mapping, state flow, settings and everything else I could think of.

## Checklist

- [x] New or updated tests have been added to cover the changes following the testing [guidelines](https://developers.home-assistant.io/docs/android/testing/introduction).
- [x] The code follows the project's [code style](https://developers.home-assistant.io/docs/android/codestyle) and [best_practices](https://developers.home-assistant.io/docs/android/best_practices).
- [x] The changes have been thoroughly tested, and edge cases have been considered.
- [x] Changes are backward compatible whenever feasible. Any breaking changes are documented in the changelog for users and/or in the code for developers depending on the relevance.

## Screenshots

notification_shade_media_controls
settings_media_controls_entry_dark
media_controls_settings_dark

## Link to pull request in documentation repositories

User Documentation PR: https://github.com/home-assistant/companion.home-assistant/pull/1304

PR #7486, opened 2026-09-16
Expose entity subtitle separator

TimoPtr:
## Summary

In https://github.com/home-assistant/android/pull/6626 we need to be able to get the separator use in the subtitle in order to properly modify the subtitle by adding the Server name. I didn't extend the EntityPicker to support multi server but sometimes we need to display a list with entities coming from different servers.

## Checklist

- [x] I have not used AI for this contribution.

## Any other notes

@jpelgrom I wonder if that would make sense somehow to include the serverName within the EntityDisplay all the time we could use it also in the auto favorite that is lacking this information today.


jpelgrom:
> @jpelgrom I wonder if that would make sense somehow to include the serverName within the EntityDisplay all the time we could use it also in the auto favorite that is lacking this information today.

In the past I've included server names only if there is more than one server and the server is maybe not clear. It could be useful, but I'm not sure it should be included in the subtitle like the 'root level'. The information isn't available in EntityDisplay so for now I'd consider it out of scope. We can have this discussion on the other PR where you are including the server name :)

jpelgrom (approved):

PR #7489, opened 2026-09-16
Add when_start to anchor bounded chronometer notifications

DerZade:
## Summary

This PR adds a new `when_start` field for (live) notifications, which can be used for anchoring chronometers.

Currently there is no way to update chronometers (let them run longer / shorter) without them starting to count from the start again. This is true for both counting up and counting down. On iOS this was even more visible, because there a progress bar is displayed with the counter. So if for example my washer re-estimates the remaining time in the middle of the cycle, the progress bar resets to 100% and "counts down" from the new remaining time. 

The only thing which I found was remotely touching on this was [`carryForwardChronometerAnchor` in the iOS App](https://github.com/home-assistant/iOS/blob/edc7153a5e99f637a27a69e8285887114456c985/Sources/Shared/LiveActivity/LiveActivityRegistry.swift#L431-L447), which only works for count-up chronometers if the same relative duration is passed. 

With `when_start` the user has full control over the starting anchor for both count-down and count-up chronometers. On Android it applies to count-ups sent as a negative relative `when`. Count-downs are unaffected, since they are just text and an absolute `when` is already a stable anchor.

`when_start` is always a Unix timestamp. `when_relative` applies to `when` only. Payloads without `when_start` keep the existing carry-forward behavior in iOS unchanged. 

## Checklist

User Documentation: home-assistant/companion.home-assistant#1427

## Any other notes

PR for iOS: home-assistant/ios#5824


app/src/main/kotlin/io/homeassistant/companion/android/changelog/ChangelogContent.kt
```diff
@@ -57,6 +57,10 @@ internal val currentChangelog = Changelog(
             contentRes = commonR.string.changelog_entry_matter_commissioning,
             platforms = setOf(ChangelogPlatform.APP),
         ),
+        ChangelogEntry(
+            contentRes = commonR.string.changelog_entry_chronometer_when_start,
+            platforms = setOf(ChangelogPlatform.APP, ChangelogPlatform.AUTOMOTIVE),
+        ),
```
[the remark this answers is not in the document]
DerZade: The screenshot was updated in this PR with the provided workflow :thinking:

common/src/main/res/values/strings_changelog.xml
```diff
@@ -12,5 +12,6 @@
     <string name="changelog_entry_matter_commissioning">Modernized adding &lt;b&gt;Matter devices&lt;/b&gt;</string>
     <string name="changelog_entry_assist">Modernized the &lt;b&gt;Assist&lt;/b&gt; pop-up</string>
     <string name="changelog_entry_tiles">Modernized settings for &lt;b&gt;tiles&lt;/b&gt;</string>
+    <string name="changelog_entry_chronometer_when_start">Added &lt;b&gt;when_start&lt;/b&gt; to chronometer notifications so a running timer keeps its start time when the notification is updated</string>
```
TimoPtr, line 15: I'm not sure it deserves an entry. If so I would like an action that open the documentation to the proper anchor.

TimoPtr (commented):
The code looks ok but I want first to agree on the doc with iOS.

common/src/main/res/values/strings_changelog.xml
DerZade: I'm happy to remove the entry. I was unsure if it is needed anyway. I figured since it's my first contribution, it was better to include too much than too little!

DerZade:
> The code looks ok but I want first to agree on the doc with iOS.

It's probably a good idea to focus on either android or iOS for reviewing first, and then apply those decisions to the other afterwards :)

PR #7498, opened 2026-09-19
Clean up Detekt exception idioms

loganrosen:
## Summary

Replace five explicit `IllegalStateException`/`IllegalArgumentException` throws with equivalent Kotlin `check`, `error`, and `require` idioms, and remove their Detekt baseline entries across the affected flavors.

The conversions preserve exception types, messages, evaluation order, and existing catch behavior. Focused tests cover the wake-word failure path, nested JSON map-key validation, minimal-flavor Thread behavior, and native MicroWakeWord failure messages.

The `TransmitterManager` `UseRequire` finding remains baselined intentionally. Its explicit exception is caught in the same function and used as routine invalid-parameter control flow, so replacing it with `require` would reduce clarity without improving the API contract.

## Checklist

## Any other notes

Validation completed:

- Focused JVM tests for `JsonUtilTest`, `AssistVoiceInteractionServiceTest` in both flavors, and minimal `ThreadManagerImplTest`
- `detektMain` and `ktlintCheck` for `:app`, `:common`, and `:microwakeword`
- `:microwakeword:compileDebugAndroidTestKotlin`
- `git diff --check`

The MicroWakeWord instrumentation tests were compiled but not executed locally because no device or emulator was available.


app/src/main/kotlin/io/homeassistant/companion/android/widgets/camera/CameraWidget.kt
```diff
@@ -128,7 +128,7 @@ class CameraWidget : AppWidgetProvider() {
                         val baseUrl = urlState.url?.toString()?.removeSuffix("/") ?: ""
                         url = "$baseUrl$entityPictureUrl"
                     } else {
-                        throw IllegalStateException("No URL available to retrieve picture")
+                        error("No URL available to retrieve picture")
```
jpelgrom, line 131: This is a scenario which can be expected (only internal URL and not internal), so not an `error` I believe but rather a `check`? Not too familiar but that is my conclusion after reading Kotlin's docs.

loganrosen: Good point. I changed this to `check(urlState is UrlState.HasUrl)` so the expected insecure/no-URL state is expressed as a failed state precondition. It preserves the existing `IllegalStateException` message and catch/log behavior.

